PR #4, opened 2019-05-28
[KED-917] Add CachedDataSet which caches data in memory to avoid io

tsanikgr:
## Notice

- I submit this contribution under the Apache 2.0 license available [here](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
- I certify that this contribution does not violate the intellectual property rights of anyone else.
 
## Motivation and Context
In many cases, users want to save a dataset to disk/some other filesystem (like Amazon's S3 etc), but want to re-use the dataset in nodes downstream of their pipeline. In these cases, the dataset keeps being reloaded from disk/network for every node in which it is needed.

`CachedDataSet` is an `AbstractDataSet` wrapper which will cache in memory the data the first time it is loaded/saved. Subsequent calls to `.load()` will return the data from memory, and subsequent calls to `.save()` will overwrite both the data on disk/network and the cache.

## How has this been tested?
Unit tests

## Checklist

- [x] Added `Type` label to the PR

### DIFF
--- a/kedro/contrib/io/cached/cached.py
+++ b/kedro/contrib/io/cached/cached.py
@@ -31,7 +31,7 @@ This module contains ``CachedDataSet``, a dataset wrapper which caches in memory
 so that the user avoids io operations with slow storage media
 """
 import pickle
-from typing import Any, Dict
+from typing import Any, Dict, Union
 
 from kedro.io import (
     AbstractDataSet,
@@ -62,7 +62,10 @@ class CachedDataSet(AbstractDataSet, ExistsMixin, FilepathVersionMixIn):
     """
 
     def __init__(
-        self, dataset: AbstractDataSet, max_loads: int = None, version: Version = None
+        self,
+        dataset: Union[AbstractDataSet, Dict],
+        max_loads: int = None,
+        version: Version = None,
     ):
         super().__init__()
         if isinstance(dataset, Dict):
--- a/kedro/contrib/io/cached/cached.py
+++ b/kedro/contrib/io/cached/cached.py
@@ -74,7 +74,7 @@ class CachedDataSet(AbstractDataSet, ExistsMixin, FilepathVersionMixIn):
             self._dataset = dataset
         else:
             raise ValueError(
-                "The argument type of `dataset` should be either a dict/YML "
+                "The argument type of `dataset` should be either a dict/YAML "
                 "representation of the dataset, or the actual dataset object."
             )
         self._cache = MemoryDataSet(max_loads=max_loads)
@@ -121,5 +121,5 @@ class CachedDataSet(AbstractDataSet, ExistsMixin, FilepathVersionMixIn):
         # in the future this can be made pickleable by erasing the cache.
         # A slight challenge would be to see how to propagate max_loads
         raise pickle.PicklingError(
-            "{}: `CachedDataSet`s cannot be " r"serialised".format(str(self))
+            "{}: `CachedDataSet`s cannot be serialised".format(str(self))
         )
--- a/kedro/contrib/io/cached/cached.py
+++ b/kedro/contrib/io/cached/cached.py
@@ -69,7 +69,7 @@ class CachedDataSet(AbstractDataSet, ExistsMixin, FilepathVersionMixIn):
     ):
         super().__init__()
         if isinstance(dataset, Dict):
-            self._dataset = self._from_config(dataset, version or Version(None, None))
+            self._dataset = self._from_config(dataset, version)
         elif isinstance(dataset, AbstractDataSet):
             self._dataset = dataset
         else:
@@ -87,8 +87,9 @@ class CachedDataSet(AbstractDataSet, ExistsMixin, FilepathVersionMixIn):
                 "`CachedDataSet`, not in the wrapped dataset."
             )
         if version:
-            config[VERSIONED_FLAG_KEY] = version
-        return AbstractDataSet.from_config("Cached", config, version.load, version.save)
+            config[VERSIONED_FLAG_KEY] = True
+            return AbstractDataSet.from_config("Cached", config, version.load, version.save)
+        return AbstractDataSet.from_config("Cached", config)
 
     def _describe(self) -> Dict[str, Any]:
         return {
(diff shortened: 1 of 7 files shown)


kedro/contrib/io/cached/cached.py
```diff
@@ -0,0 +1,121 @@
...
+          dataset:
+              type: CSVLocalDataSet
+              filepath: example.csv
+
+    Please notice that if your dataset is versioned, this should be indicated in the wrapper class
+    as shown above
+    """
+
+    def __init__(
+        self, dataset: AbstractDataSet, max_loads: int = None, version: Version = None
```
tolomea: the type annotation for dataset should be something like Union[AbstractDataSet, Dict]

kedro/contrib/io/cached/cached.py
```diff
@@ -0,0 +1,121 @@
...
+            if hasattr(self._dataset, "exists"):
+                return self._dataset.exists()
+            return False  # pragma: no cover
+
+        return self._cache.exists() or _wrapped_exists()
+
+    def __getstate__(self):
+        # in the future this can be made pickleable by erasing the cache.
+        # A slight challenge would be to see how to propagate max_loads
+        raise pickle.PicklingError(
```
tolomea: if you used managed memory datasets like the parallel runner then this would be parallel runner compatible, and I'd guess anyone who cares about caching datasets is also going to want parallel
it'd be nice if had some way to indicate that a thing can be used with managed memory, but is otherwise not generally serializable, at the moment airflow explicitly checks for memory datasets which isn't great and wouldn't catch this

kedro/contrib/io/cached/cached.py
```diff
@@ -0,0 +1,121 @@
...
+        if isinstance(dataset, Dict):
+            self._dataset = self._from_config(dataset, version or Version(None, None))
+        elif isinstance(dataset, AbstractDataSet):
+            self._dataset = dataset
+        else:
+            raise ValueError(
+                "The argument type of `dataset` should be either a dict/YML "
+                "representation of the dataset, or the actual dataset object."
+            )
+        self._cache = MemoryDataSet(max_loads=max_loads)
```
tolomea: I like how you just delegated this

kedro/contrib/io/cached/cached.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def __init__(
+        self,
+        dataset: Union[AbstractDataSet, Dict],
+        max_loads: int = None,
+        version: Version = None,
+    ):
+        super().__init__()
+        if isinstance(dataset, Dict):
+            self._dataset = self._from_config(dataset, version or Version(None, None))
```
DmitriiDeriabinQB:
```suggestion
            self._dataset = self._from_config(dataset, version)
```
otherwise the wrapped dataset is always going to be versioned even if `CachedDataSet` is not versioned

kedro/contrib/io/cached/cached.py
```diff
@@ -0,0 +1,124 @@
...
+
+    @staticmethod
+    def _from_config(config, version):
+        if VERSIONED_FLAG_KEY in config:
+            raise ValueError(
+                "Cached datasets should specify that they are versioned in the "
+                "`CachedDataSet`, not in the wrapped dataset."
+            )
+        if version:
+            config[VERSIONED_FLAG_KEY] = version
```
DmitriiDeriabinQB:
```suggestion
            config[VERSIONED_FLAG_KEY] = True
            return AbstractDataSet.from_config("Cached", config, version.load, version.save)
```

kedro/contrib/io/cached/cached.py
```diff
@@ -0,0 +1,124 @@
...
+        return AbstractDataSet.from_config("Cached", config, version.load, version.save)
```
DmitriiDeriabinQB:
```suggestion
        return AbstractDataSet.from_config("Cached", config)
```

kedro/contrib/io/cached/cached.py
```diff
@@ -0,0 +1,121 @@
...
+        self._cache.save(data)
+        return data
+
+    def _save(self, data: Any) -> None:
+        self._dataset.save(data)
+        self._cache.save(data)
+
+    def _exists(self) -> bool:
+        def _wrapped_exists():
+            if hasattr(self._dataset, "exists"):
```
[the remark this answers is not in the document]
tsanikgr: +1

Completely agree, `ExistsMixin` seems an overkill. Maybe it can have a default implementation that returns `False` (or raise?)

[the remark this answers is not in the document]
tolomea: I will look into this

kedro/contrib/io/cached/cached.py
tsanikgr: On one hand I agree, on the other hand if you have to serialise to move to a separate process caching kind of defeats the purpose here tho, no? Although I agree that this can be much faster for slow formats or network datasets.

> at the moment airflow explicitly checks for memory datasets which isn't great and wouldn't catch this

Maybe inherit from `MemoryDataSet`? But this feels wrong as it conveys a completely wrong message..

Shall I take a stab converting this to a managed memory dataset, or would that be an overkill? What I think I should probably do is just drop the cache when serialising..

tolomea: I'd leave it as is and address it in a followup

tests/contrib/io/cached/test_cached_dataset.py
```diff
@@ -0,0 +1,160 @@
...
+@pytest.fixture
+def cached_ds():
+    wrapped = MemoryDataSet()
+    return CachedDataSet(wrapped)
+
+
+class TestCachedDataset:
+    def test_load_empty(self, cached_ds):
+        with pytest.raises(DataSetError, match=r"has not been saved yet"):
+            _ = cached_ds.load()
```
deepyaman: @tsanikgr Was reading the PR just out of curiosity--what's the purpose of assigning to `_` across your tests? I assume it's because the most frequent usage is to assign to a variable, but that's neither required nor always going to be the case.

Minyus:
@tsanikgr Since I want CachedDataSet feature, I sent a PR to your `feature/cached-dataset` branch to make it ready for merge to the current `develop` branch.

tsanikgr:
[Merged] (https://github.com/quantumblacklabs/private-kedro/commit/3f89a6a5a353d5e0ed94f7b58df2b7e251fae64a)

PR #5, opened 2019-05-28
Refactor how max_loads is set for MemoryDataSets

tsanikgr:
This change allows the custom implementation of `AbstractDataSet`s that
can set a limit on the number of times something can be loaded. For
example, a `CachedDataSet` might want to clear its cache after the data
has been loaded a certain number of times

## Notice

- I submit this contribution under the Apache 2.0 license available [here](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
- I certify that this contribution does not violate the intellectual property rights of anyone else.
 
## Motivation and Context
This change allows the custom implementation of `AbstractDataSet`s that can set a limit on the number of times something can be loaded. For example, a `CachedDataSet` might want to clear its cache after the data has been loaded a certain number of times

As a result, I believe this change makes the code more extensible. To achieve the same effect now, one needs to create a custom `Runner`, which is probably too much work.

Possibly this needs to be deprecated in the future and moved to be a built-in functionality of datasets (like `versioned`).

## How has this been tested?
Existing unit tests - no API breaking changes

## Checklist

- [x] Added `Type` label to the PR

### DIFF
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -290,6 +290,18 @@ class AbstractDataSet(abc.ABC):
             "it must implement the `_describe` method".format(self.__class__.__name__)
         )
 
+    def set_max_loads(self, max_loads: int):
+        """Set how many times this dataset can be loaded. ``DataSet``s that can set a limit to
+        the number of times they can be loaded need to implement this method and the relevant
+        logic.
+
+        Args:
+            max_loads: Maximum number of times ``load`` method of the
+                data set is allowed to be invoked. Any number of calls
+                is allowed if the argument is not set.
+        """
+        pass
+
 
 class ExistsMixin(abc.ABC):
     """Mixin class which provides an exists() method."""
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -45,7 +45,6 @@ from kedro.utils import load_obj
 MAX_DESCRIPTION_LENGTH = 70
 VERSIONED_FLAG_KEY = "versioned"
 VERSION_KEY = "version"
-FORCE_UNLIMITED_LOADS = -1
 
 
 class DataSetError(Exception):
@@ -291,11 +290,10 @@ class AbstractDataSet(abc.ABC):
             "it must implement the `_describe` method".format(self.__class__.__name__)
         )
 
-    def set_max_loads(self, max_loads: int = FORCE_UNLIMITED_LOADS):
+    def set_max_loads(self, max_loads: int):
         """Set how many times this dataset can be loaded. ``DataSet``s that can set a limit to
         the number of times they can be loaded need to implement this method and the relevant
-        logic, or the can specify that they can be loaded unlimited times by using
-        ``kedro.io.core.FORCE_UNLIMITED_LOADS``
+        logic.
 
         Args:
             max_loads: Maximum number of times ``load`` method of the
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -45,6 +45,7 @@ from kedro.utils import load_obj
 MAX_DESCRIPTION_LENGTH = 70
 VERSIONED_FLAG_KEY = "versioned"
 VERSION_KEY = "version"
+FORCE_UNLIMITED_LOADS = -1
 
 
 class DataSetError(Exception):
@@ -290,10 +291,11 @@ class AbstractDataSet(abc.ABC):
             "it must implement the `_describe` method".format(self.__class__.__name__)
         )
 
-    def set_max_loads(self, max_loads: int):
+    def set_max_loads(self, max_loads: int = FORCE_UNLIMITED_LOADS):
         """Set how many times this dataset can be loaded. ``DataSet``s that can set a limit to
         the number of times they can be loaded need to implement this method and the relevant
-        logic.
+        logic, or the can specify that they can be loaded unlimited times by using
+        ``kedro.io.core.FORCE_UNLIMITED_LOADS``
 
         Args:
             max_loads: Maximum number of times ``load`` method of the
--- a/kedro/io/data_catalog.py
+++ b/kedro/io/data_catalog.py
@@ -444,7 +444,7 @@ class DataCatalog:
         return self._data_sets == other._data_sets  # pylint: disable=protected-access
 
     def set_max_loads(self, ds_name: str, max_loads: int):
-        """Set the maximum number of times this dataset can be loaded. ``DataSet``s like
+        """Set the maximum number of times the given dataset can be loaded. ``DataSet``s like
         ``MemoryDataSet`` use this to clear data from memory. In most other cases, this has no
         effect.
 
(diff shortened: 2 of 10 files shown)


tsanikgr:
This will help properly implement #4

kedro/io/data_catalog.py
```diff
@@ -442,3 +442,16 @@ def shallow_copy(self) -> "DataCatalog":
 
     def __eq__(self, other):
         return self._data_sets == other._data_sets  # pylint: disable=protected-access
+
+    def set_max_loads(self, ds_name: str, max_loads: int):
+        """Set the maximum number of times this dataset can be loaded. ``DataSet``s like
```
tolomea: "this dataset" -> "the given dataset"

tsanikgr:
By the way I have not put much though to this - open to alternative suggestions if you think this is something in the right direction

kedro/io/core.py
```diff
@@ -290,6 +290,18 @@ def _describe(self) -> Dict[str, Any]:
             "it must implement the `_describe` method".format(self.__class__.__name__)
         )
 
+    def set_max_loads(self, max_loads: int):
```
DmitriiDeriabinQB: Shouldn't `max_loads` default to `None` if we allow it to not be specified?
(It seems I missed it in the initial implementation of `create_default_data_set`)

kedro/runner/runner.py
```diff
@@ -77,9 +77,12 @@ def run(self, pipeline: Pipeline, catalog: DataCatalog) -> Dict[str, Any]:
         free_outputs = pipeline.outputs() - set(catalog.list())
         unregistered_ds = pipeline.data_sets() - set(catalog.list())
         for ds_name in unregistered_ds:
+            catalog.add(ds_name, self.create_default_data_set(ds_name))
+
+        for ds_name in catalog.list():
             num_loads = len(pipeline.only_nodes_with_inputs(ds_name).nodes)
             num_loads = num_loads if num_loads > 0 else None
-            catalog.add(ds_name, self.create_default_data_set(ds_name, num_loads))
+            catalog.set_max_loads(ds_name, num_loads)
```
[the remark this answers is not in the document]
DmitriiDeriabinQB: Would it make sense to move it under `for ds_name in unregistered_ds`? This, however, means that all MemoryDataSets specified in the config will be ignored, which if I remember, was a primary use case for this.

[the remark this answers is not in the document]
tolomea: maybe anything that we write to, so Pipeline.all_outputs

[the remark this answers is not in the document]
tolomea: Pipeline.all_outputs & Pipeline.all_inputs will give all the intermediate stuff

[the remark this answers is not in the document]
tolomea: having said that, why would you have a memory dataset in the catalog if you didn't want to either prefill it, or have it's content available after the run?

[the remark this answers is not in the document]
tsanikgr: > Would it make sense to move it under for ds_name in unregistered_ds

We want this to work also for `CachedDataSet`s, this is what triggered this PR.

> having said that, why would you have a memory dataset in the catalog if you didn't want to either prefill it, or have its content available after the run?

Agreed - but let's try and cope with this if they do it.

> Pipeline.all_outputs & Pipeline.all_inputs will give all the intermediate stuff

I was almost convinced that this is the right thing to do, until I realised it won't work for top-level cached datasets which are being consumed by multiple nodes in the pipeline..

I got an idea, taking a stab at it now!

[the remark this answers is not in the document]
tolomea: you could make max loads on memory data set only work if it's set before the first load

tsanikgr:
Thanks both for the very insightful reviews. Can you have another look - I think I implemented a solution that allows preventing clearing top-level inputs

tolomea:
I don't like the additional complexity of FORCE_UNLIMITED_LOADS

tsanikgr:
Removed and implemented @tolomea 's suggestion: the call to `.set_max_loads()` is (silently) ignored if the `MemoryDataSet` contains already data

kedro/io/core.py
tsanikgr: Not sure what you mean? Why would you call this method without specifying the `max_loads`?

DmitriiDeriabinQB: In the original implementation if `max_loads is None`, it allows unlimited number of loads

kedro/io/memory_data_set.py
```diff
@@ -121,3 +121,18 @@ def _exists(self) -> bool:
...
+        memory. Calling this method on a ``MemoryDataSet`` that already contains data has *no
+        effect*.
+
+        Args:
+            max_loads: Maximum number of times ``load`` method of the
+                data set is allowed to be invoked, before clearing the
+                data to release memory. Any number of calls is allowed
+                if the argument is not set.
+        """
+        if self._data is not None:
```
DmitriiDeriabinQB: 1. Shouldn't it be `if self._data is None`?
2. If `max_loads` has been passed to the `__init__` then the counter will continue regardless.

kedro/io/memory_data_set.py
```diff
@@ -121,3 +121,18 @@ def _exists(self) -> bool:
...
+
+    def set_max_loads(self, max_loads):
+        """Set how many times this dataset can be loaded before its data being cleared to release
+        memory. Calling this method on a ``MemoryDataSet`` that already contains data has *no
+        effect*.
+
+        Args:
+            max_loads: Maximum number of times ``load`` method of the
+                data set is allowed to be invoked, before clearing the
+                data to release memory. Any number of calls is allowed
```
DmitriiDeriabinQB: If `max_loads` does not accept `None` anymore, then the sentence about the argument not being set should probably be removed

kedro/io/memory_data_set.py
```diff
@@ -121,3 +121,18 @@ def _exists(self) -> bool:
         if self._data is None:
             return False
         return True
+
+    def set_max_loads(self, max_loads):
+        """Set how many times this dataset can be loaded before its data being cleared to release
```
Pet3ris: Would like to avoid the counting if possible.

Could we perhaps create an interface that allows each data set to be used as a python context, e.g.,

```python
with memory_ds.use(): # something like this
    memory_ds.load()
    # process, load some more...
```

Then the disposal would happen automatically, we could reconfigure the pipeline to use these contexts when "accessing" data sources.

**Key benefit**: User doesn't have to do annoying counting to figure out how many times a data set will be called in the pipeline.

tolomea: The interface is a bit clunky but the target `users` here are the runners and for them this interface is the simplest to use. Integrating context manager calls into the runners, particularly releasing the contexts mid way through the run, would be quite complicated.

tsanikgr:
[Merged](https://github.com/quantumblacklabs/kedro/commit/69127a0943ba6a0b55db080fca1b1fdbe4ddb3b8)

PR #15, opened 2019-06-09
 Modularize default load and save argument handling

### LINKED ISSUE #14: Modularize default argument handling for datasets
## Description
Near-identical code to handle default arguments is replicated in almost every dataset implementation. Worse still, functionality across said datasets is the same, but implementation is inconsistent.

## Context
When I want to implement a new dataset, I look at existing datasets as a baseline for implementing my own. However, there are inconsistencies between these datasets, from the more minor (`save_args` handled after `load_args` for some datasets), to the slightly more significant (special casing where there are no default arguments on some datasets but not others) and worse (one case where arguments are evaluated for truthiness instead of `is not None`) (see https://github.com/quantumblacklabs/kedro/blob/0.14.1/kedro/contrib/io/azure/csv_blob.py#L109-L113 as an example representing several of the above). I don't know which one to follow to maintain consistency across the codebase.

## Possible Implementation
#15 

By having `DEFAULT_LOAD_ARGS`/`DEFAULT_SAVE_ARGS` attributes, users can also see the defaults programmatically (with the caveat that this is a drawback if you consider the few cases where such arguments don't apply, like no save on `SqlQueryDataSet` or in general on `LambdaDataSet`/`MemoryDataSet`).

## Possible Alternatives
* Create an intermediate abstract dataset class (or mixin?) so as to not modify `AbstractDataSet` and thereby only apply to those with `load_args`/`save_args`
* Move default argument handling into a utility function and call it from each individual `__init__` method (not preferred)

## Checklist
Include labels so that we can categorise your issue:

- [ ] Add a "Component" label to the issue
- [ ] Add a "Priority" label to the issue

deepyaman:
## Notice

## Motivation and Context
Why was this PR created?

Close #14 

## How has this been tested?

Unit tests still pass, plus limited manual testing. More testing of edge cases possible.

image

## Checklist

- [ ] Added `Type` label to the PR


kedro/contrib/io/pyspark/spark_jdbc.py
```diff
@@ -140,8 +140,7 @@ def __init__(
 
         self._url = url
         self._table = table
-        self._load_args = load_args if load_args is not None else {}
-        self._save_args = save_args if save_args is not None else {}
+        super().__init__(load_args, save_args)
```
deepyaman: This fails a lot of unit tests locally; I have no idea how/why (yet)...

tsanikgr:
See my comment in #14 for an alternative proposition

kedro/io/core.py
```diff
@@ -189,6 +192,22 @@ def from_config(
...
 
+    def __init__(
+        self,
+        load_args: Optional[Dict[str, Any]] = None,
+        save_args: Optional[Dict[str, Any]] = None,
+    ) -> None:
+        self._load_args = (
+            {**self.DEFAULT_LOAD_ARGS, **load_args}
+            if load_args is not None
+            else self.DEFAULT_LOAD_ARGS
```
tolomea: I appreciate that it's copied from some of the existing datasets but this should probably copy to prevent accidental mutation of the defaults, perhaps
```
self._load_args = dict(self.DEFAULT_LOAD_ARGS)
if load_args:
    self._load_args.update(**load_args)
```
which also has the nice effect of removing inline conditional, inline conditionals split across lines should generally be avoided

deepyaman: @tolomea Thanks for the catch! This actually fixes the failing `e2e-tests` and another "weird" case where `super().__init__` was causing failing behavior.

kedro/contrib/io/pyspark/spark_jdbc.py
deepyaman: Resolved by @tolomea's suggestion.

deepyaman:
@idanov @tolomea @tsanikgr Updated with emojis (and changes discussed with Ivan in #14). Let me know if it looks good!

I'm pretty confident in the implementation, less so in the added tests (I generally picked arguments that don't affect anything to cover `load_args`/`save_args`, but I haven't been using most of these datasets).

kedro/io/core.py
```diff
@@ -101,6 +101,9 @@ class AbstractDataSet(abc.ABC):
         >>>         return dict(param1=self._param1, param2=self._param2)
     """
 
+    DEFAULT_LOAD_ARGS = {}
```
idanov: Is this something we can leave to the classes extending `AbstractDataSet`?

kedro/contrib/io/azure/csv_blob.py
```diff
@@ -35,10 +35,11 @@
 import pandas as pd
 from azure.storage.blob import BlockBlobService
 
+from kedro.contrib.io import DefaultArgumentsMixIn
 from kedro.io import AbstractDataSet
 
 
-class CSVBlobDataSet(AbstractDataSet):
+class CSVBlobDataSet(AbstractDataSet, DefaultArgumentsMixIn):
```
tolomea: Base class method resolution proceeds from left to right, so Mixins should always be to the left and base/abstract classes to the right.

kedro/contrib/io/core.py
```diff
@@ -0,0 +1,51 @@
...
+
+    DEFAULT_LOAD_ARGS = {}
+    DEFAULT_SAVE_ARGS = {}
+
+    def __init__(
+        self,
+        load_args: Optional[Dict[str, Any]] = None,
+        save_args: Optional[Dict[str, Any]] = None,
+    ) -> None:
+        self._load_args = self.DEFAULT_LOAD_ARGS.copy()
```
tolomea: this function should call super in case anything else in the hierarchy implements init

kedro/io/core.py
deepyaman: Oh, sorry, I forgot to take this out. 🤦‍♂

kedro/contrib/io/core.py
deepyaman: @tolomea `super().__init__()` without args so that it would fail (not be silent) if base class implemented super with arms?

tolomea: you can either do that or pass *args and **kwargs along, either is fine

kedro/contrib/io/azure/csv_blob.py
deepyaman: @tolomea Thanks for pointing this out! Resolved, although I assume we should also do the same for `FilepathVersionMixIn`/`S3PathVersionMixIn` in another PR.

deepyaman:
@idanov @tolomea @tsanikgr Made the aforementioned changes, and tests pass locally, but running into Java errors on the build. I see some StackOverflow answers around forcing Java 1.8 (i.e. Spark not supporting Java 11), but since lot of other build are failing on this and I feel you all must have encountered this, putting it off till morning. :)

tolomea:
The Spark issue will be resolved when you update to latest develop.

Regarding the ordering of base classes, you are correct about the version mixins, there is work underway to fix that as part of merging the two different mixins into a proper base class.

deepyaman:
Types annotated as required by 3.5. :( That should be everything!

kedro/io/hdf_s3.py
```diff
@@ -103,30 +106,26 @@ def __init__(
...
-            if load_args is not None
-            else default_load_args
-        )
-        self._save_args = (
-            {**default_load_args, **save_args}
-            if save_args is not None
-            else default_save_args
-        )
+
+        # Handle default load and save arguments
```
tsanikgr: Why this (and all other datasets in `kedro.io` that are eligible) are not inheriting from the new `MixIn`? (maybe I missed something here, sorry about that!)

The code feels inconsistent now (+ all this duplication can go away!)

deepyaman: @tsanikgr https://github.com/quantumblacklabs/kedro/issues/14#issuecomment-507703582

tsanikgr: OK thanks! @idanov, since the mix in was introduced, and is solving the problem "of the wrong abstraction", should we just leverage it in the core datasets as well?

deepyaman: @idanov @tsanikgr @tolomea Let me know what you all decide as the core team here; I’m OK with it as is (a current marginal improvement to `core` with potential to move the mix-in from `contrib` after proving value in the future) or pushing the mix-in to `core` now. I’m more keen on merging to `develop` sooner than later due to the number of datasets touched, if possible. :)

kedro/contrib/io/bioinformatics/sequence_dataset.py
```diff
@@ -95,18 +96,7 @@ def __init__(
...
-            {**default_load_args, **load_args}
-            if load_args is not None
-            else default_load_args
-        )
-        self._save_args = (
-            {**default_save_args, **save_args}
-            if save_args is not None
-            else default_save_args
-        )
+        super().__init__(load_args, save_args)
```
LorenaBalanQB: I prefer calling `super` at the top of the constructor, so the subclass would overwrite stuff from the parent, as a "specialisation" of the superclass.

kedro/io/hdf_s3.py
idanov: I think the current version is good for now. Thanks @deepyaman for accommodating for all the comments!

kedro/contrib/io/bioinformatics/sequence_dataset.py
deepyaman: > I prefer calling `super` at the top of the constructor, so the subclass would overwrite stuff from the parent, as a "specialisation" of the superclass.

Fair argument. I just left it in the same place where default arguments were previously handled (as close to the original as I could), but that makes sense.

PR #34, opened 2019-06-25
[KED-921] Add Matplotlib dataset

williamashfordQB:
## Notice

- [X] I acknowledge and agree that, by checking this box and clicking “Submit Pull Request”:

## Motivation and Context
In the process of making documentation of features, programmatically making plots of all features in a standardised way and embedding links in .md files has been a big timesaver. In the hope of helping others to benefit from this automation, this PR has been raised. 

This PR contains a dataset class to save matplotlib figures/plt objects to images locally, MatplotlibWriter. Currently I cannot think of a usecase for reading images into matplotlib, and as such, I've not included any support for load.

## How has this been tested?
The PR contains a unit test suite. 

## Checklist

- [X] Read the [contributing](/CONTRIBUTING.md) guidelines
- [] Updated the documentation to reflect the code changes
- [ ] Added `Type` label to the PR


kedro/contrib/io/matplotlib/matplotlib_writer.py
```diff
@@ -0,0 +1,139 @@
...
+        return {**default_args, **user_args} if user_args else default_args
+
+    def _load(self) -> str:
+        raise DataSetError("Loading not supported for MatplotlibWriter")
+
+    def _save(self, data) -> None:
+
+        if self._mutlifile_mode:
+
+            if not os.path.isdir(self._filepath):
```
tolomea: pathlib.Path is preferred for all path manipulation

kedro/contrib/io/matplotlib/matplotlib_writer.py
```diff
@@ -0,0 +1,139 @@
...
+                        os.path.join(self._filepath, str(index)), **self._save_args
+                    )
+
+            elif isinstance(data, dict):
+                for plot_name, plot in data.items():
+                    plot.savefig(
+                        os.path.join(self._filepath, plot_name), **self._save_args
+                    )
+
+            else:
```
tolomea: do you actually need the multi_file flag? can't you just do the single file stuff in this else condition?

tests/contrib/io/matplotlib/test_simple_io.py
```diff
@@ -0,0 +1,174 @@
...
+
+    expected_load_error = "Loading not supported for MatplotlibWriter"
+
+    assert str(excinfo.value) == expected_load_error
+
+
+def test_exists_functionality(tmpdir):
+    import matplotlib.pyplot as plt
+
+    plot_object = plt.figure()
```
921kiyo: These 3 lines can be in a pytest fixture. (See [test_json_local.py](https://github.com/quantumblacklabs/kedro/blob/develop/tests/io/test_json_local.py) as an example)

921kiyo:
Could you also update `kedro.contrib.io.rst` so this dataset will appear in the doccumentation?

kedro/contrib/io/matplotlib/matplotlib_writer.py
```diff
@@ -0,0 +1,139 @@
...
+        self._filepath = filepath
+        self._load_args = self._handle_default_args(load_args, default_load_args)
+        self._save_args = self._handle_default_args(save_args, default_save_args)
+        self._mutlifile_mode = self._save_args.pop("multiFile")
+
+    @staticmethod
+    def _handle_default_args(user_args: dict, default_args: dict) -> dict:
+        return {**default_args, **user_args} if user_args else default_args
+
+    def _load(self) -> str:
```
921kiyo:
```suggestion
    def _load(self) -> None:
```

kedro/contrib/io/matplotlib/matplotlib_writer.py
williamashfordQB: A single file wouldn't be present as a dict or list as this would imply multiple objects

tests/contrib/io/matplotlib/test_simple_io.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+import matplotlib
+import numpy as np
+import pytest
+
+from kedro.contrib.io.matplotlib import MatplotlibWriter
+from kedro.io import DataSetError
+
+matplotlib.use('TkAgg')  # used to facilitate simple inclusion into kedro CI/CD
```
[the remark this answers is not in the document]
williamashfordQB: This is just an alternative engine to use. I was trying without and was getting some environment issues in pytest which would have required a different means of execution of tests for this module. 

In order to avoid making any inappropriate changes to higher level operations, I decided to conform this section.

tests/contrib/io/matplotlib/test_simple_io.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+def test_should_write_to_image_in_single_mode(tmpdir):
```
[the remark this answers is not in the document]
williamashfordQB: I don't see the benefit in consolidating these tests in a class and think that it would probably be a deviation to how the other tests are written within the package. 

The name is deliberately specific as I believe tests should serve as a sort of secondary living documentation. If users were interested in how this class works, they could come in, look at this tests and easily how a matplotlib image is written in single mode.

This is based a little on the idea of specifications in behaviour driven development and the reasoning behind them too.

tests/contrib/io/matplotlib/test_simple_io.py
```diff
@@ -0,0 +1,174 @@
...
+    plt.savefig(trusted_filepath)
+
+    experimental_filepath = str(tmpdir.join("image_we_write.png"))
+    plot_writer = MatplotlibWriter(filepath=experimental_filepath)
+    plot_writer.save(plt)
+
+    plt.close()
+
+    assert (
+        open(experimental_filepath, "rb").read() == open(trusted_filepath, "rb").read()
```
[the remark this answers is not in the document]
williamashfordQB: What is the benefit of explicitly closing the open?

tests/contrib/io/matplotlib/test_simple_io.py
```diff
@@ -0,0 +1,174 @@
...
+    )
+
+
+def test_should_write_to_image_in_list_multi_mode(tmpdir):
+    import matplotlib.pyplot as plt
```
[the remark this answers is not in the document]
williamashfordQB: I agree, but I couldn't make this work with the linting requirements unfortunately.

kedro/contrib/io/matplotlib/matplotlib_writer.py
```diff
@@ -0,0 +1,139 @@
...
+        raise DataSetError("Loading not supported for MatplotlibWriter")
+
+    def _save(self, data) -> None:
```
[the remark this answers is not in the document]
williamashfordQB: This would require `from matplotlib.pyplot import figure` which would add a non-testing dependency for matplotlib. Is this OK?

tests/contrib/io/matplotlib/test_simple_io.py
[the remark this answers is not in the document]
tolomea: It's considered good practice. If you don't explicitly close (or use a mechanism that does like `with open` or `Path.read_text_text`) then the file gets closed when the garbage collector finalizes it. While that is usually relatively soon, it's only actually guaranteed to happen before the process exits (and then only if the Python VM comes down cleanly). If you are writing to the file that's a quick path to data corruption. If you are reading it's not usually an issue unless you are on a filesystem that read locks, which is anything on Windows and some network things on other systems.

kedro/contrib/io/matplotlib/matplotlib_writer.py
tolomea: right, so a dict or a list implies multi file and so the _multifile_mode flag is unnecessary

tests/contrib/io/matplotlib/test_simple_io.py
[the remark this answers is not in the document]
williamashfordQB: fair enough, will implement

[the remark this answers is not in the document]
tolomea: Further if you don't close it because you know it's safe not to, reviewers have to think about whether it's safe, where as if you close it anyway they don't. Good practice is often about making the reviewers think less.

kedro/contrib/io/matplotlib/matplotlib_writer.py
williamashfordQB: hmm yeah fair enough, will implement

kedro/contrib/io/matplotlib/matplotlib_writer.py
```diff
@@ -0,0 +1,139 @@
...
+                        "multiFile is True but data type "
+                        "not dict or list. Rather, {}".format(plot_type)
+                    )
+                )
+
+        else:
+            data.savefig(self._filepath, **self._save_args)
+
+    def _exists(self) -> bool:
+        return os.path.isfile(self._filepath)
```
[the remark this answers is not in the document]
williamashfordQB: Seems like a bit of overkill to check that the file exists no?

tests/contrib/io/matplotlib/test_simple_io.py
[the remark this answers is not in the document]
williamashfordQB: Sounds good :)

tests/contrib/io/matplotlib/test_simple_io.py
```diff
@@ -0,0 +1,174 @@
...
+
+    plot_object = plt.figure()
+    plt.plot(np.random.rand(1, 5)[0], np.random.rand(1, 5)[0])
+    plt.close()
+
+    plot_writer = MatplotlibWriter(
+        filepath=str(tmpdir.join("some_path")), save_args={"multiFile": True}
+    )
+
+    with pytest.raises(DataSetError) as excinfo:
```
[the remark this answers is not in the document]
williamashfordQB: Awesome, will implement :)

tests/contrib/io/matplotlib/test_simple_io.py
williamashfordQB: I was going to do a fixture but the implementation of the figures is slightly different and I'm not sure of the value of having something generating an object for reuse when a higher level state is concerned (.e.g. with the plt module). 

For this reason I don't think there is a value in reuse of implementing fixtures here.

kedro/contrib/io/matplotlib/matplotlib_writer.py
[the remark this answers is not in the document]
lorenabalan: @921kiyo `self._get_load_path` is only available in `AbstractVersionedDataSet`, which this dataset is not inheriting from, so that's okay I reckon. 
However, as per @tolomea 's comments, `pathlib.Path` is preferred for all path manipulation - @williamashfordQB could you please update this accordingly? Something like `Path(self._filepath).is_file()` should do the trick.

kedro/contrib/io/matplotlib/matplotlib_writer.py
```diff
@@ -0,0 +1,139 @@
...
+        default_save_args = {"multiFile": False}
+        default_load_args = {}
+
+        self._filepath = filepath
+        self._load_args = self._handle_default_args(load_args, default_load_args)
+        self._save_args = self._handle_default_args(save_args, default_save_args)
+        self._mutlifile_mode = self._save_args.pop("multiFile")
+
+    @staticmethod
+    def _handle_default_args(user_args: dict, default_args: dict) -> dict:
```
lorenabalan: Can we make this class extend `DefaultArgsMixin` instead of having this static method, ensuring `super().__init__()` is called at the top? That will take care of save_args & default_save_args, etc., and is consistent with other datasets in contrib.

kedro/contrib/io/matplotlib/matplotlib_writer.py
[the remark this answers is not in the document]
lorenabalan: Had a chat with @tolomea about this, it won't affect anything if you include the import. It's a similar situation for the `SparkDataSet` - as long as the file is not actually imported anywhere else, it's fine. `SparkDataSet` is only imported in the `__init__.py` of the same folder, and that is not imported anywhere else in the codebase, so the requirements don't include `pyspark`. Given that, it'd probably be nice to include the type hint.

lorenabalan:
I've updated the PR title with our internal ticket number to keep track of this more easily. :)

921kiyo:
@williamashfordQB I am wondering if you had time to look at the above comments. We would like to move this forward and merge it, and let us know how we can help:)

tests/contrib/io/matplotlib/test_matplotlib_writer.py
```diff
@@ -0,0 +1,122 @@
...
+            plot.savefig(str(trusted_filepath))
+            experimental_filepath = tmp_path / "dict_images" / filename
+
+            assert trusted_filepath.read_bytes() == experimental_filepath.read_bytes()
+
+    def test_load_fail(self, tmp_path):
+        plot_writer = MatplotlibWriter(filepath=str(tmp_path / "some_path"))
+
+        expected_load_error = "Loading not supported for MatplotlibWriter"
+        with pytest.raises(DataSetError, match=expected_load_error):
```
lorenabalan:
```suggestion
        with pytest.raises(DataSetError, match=pattern):
```

tests/contrib/io/matplotlib/test_matplotlib_writer.py
```diff
@@ -0,0 +1,122 @@
...
+            )
+            plot.savefig(str(trusted_filepath))
+            experimental_filepath = tmp_path / "dict_images" / filename
+
+            assert trusted_filepath.read_bytes() == experimental_filepath.read_bytes()
+
+    def test_load_fail(self, tmp_path):
+        plot_writer = MatplotlibWriter(filepath=str(tmp_path / "some_path"))
+
+        expected_load_error = "Loading not supported for MatplotlibWriter"
```
lorenabalan:
```suggestion
        pattern = r"Loading not supported for `MatplotlibWriter`"
```

kedro/contrib/io/matplotlib/matplotlib_writer.py
```diff
@@ -0,0 +1,114 @@
...
+            >>> multi_plot_writer.save(plots)
+
+    """
+
+    def _describe(self) -> Dict[str, Any]:
+        return dict(
+            filepath=self._filepath,
+            load_args=self._load_args,
+            save_args=self._save_args,
+        )
```
lorenabalan: Small thing, but I'd rather have `__init__` the first function the user sees when they dig out the source code.

kedro/contrib/io/matplotlib/matplotlib_writer.py
```diff
@@ -0,0 +1,114 @@
...
+            save_args: Save args passed to `plt.savefig`. See
+                https://matplotlib.org/api/_as_gen/matplotlib.pyplot.savefig.html
+        """
+        self._filepath = Path(filepath)
+        self._load_args = load_args if load_args else dict()
+        self._save_args = save_args if save_args else dict()
+
+    def _load(self) -> None:
+        raise DataSetError(
+            "Loading not supported for {}".format(self.__class__.__name__)
```
lorenabalan:
```suggestion
            "Loading not supported for `{}`".format(self.__class__.__name__)
```

tests/contrib/io/matplotlib/test_matplotlib_writer.py
```diff
@@ -0,0 +1,122 @@
...
+import pytest
+
+from kedro.contrib.io.matplotlib import MatplotlibWriter
+from kedro.io import DataSetError
+
+matplotlib.use("Agg")  # Disable interactive mode
+
+
+class TestMatplotlibWriter:
+    def test_simgle_image(self, tmp_path):
```
lorenabalan:
```suggestion
    def test_save_simgle_image(self, tmp_path):
```

tests/contrib/io/matplotlib/test_matplotlib_writer.py
```diff
@@ -0,0 +1,122 @@
...
+        plt.savefig(str(trusted_filepath))
+
+        experimental_filepath = tmp_path / "image_we_write.png"
+        plot_writer = MatplotlibWriter(filepath=str(experimental_filepath))
+        plot_writer.save(plt)
+        plt.close()
+
+        assert trusted_filepath.read_bytes() == experimental_filepath.read_bytes()
+
+    def test_list_images(self, tmp_path):
```
lorenabalan:
```suggestion
    def test_save_list_images(self, tmp_path):
```

tests/contrib/io/matplotlib/test_matplotlib_writer.py
```diff
@@ -0,0 +1,122 @@
...
+            plot.savefig(str(trusted_filepath))
+
+            full_experimental_filepath = experimental_filepath / "{}.png".format(
+                str(index)
+            )
+            assert (
+                trusted_filepath.read_bytes() == full_experimental_filepath.read_bytes()
+            )
+
+    def test_dict_images(self, tmp_path):
```
lorenabalan:
```suggestion
    def test_save_dict_images(self, tmp_path):
```

tests/contrib/io/matplotlib/test_matplotlib_writer.py
```diff
@@ -0,0 +1,122 @@
...
+        experimental_filepath = tmp_path / "image_we_write.png"
+        plot_writer = MatplotlibWriter(filepath=str(experimental_filepath))
+        plot_writer.save(plt)
+        plt.close()
+
+        assert trusted_filepath.read_bytes() == experimental_filepath.read_bytes()
+
+    def test_list_images(self, tmp_path):
+        # generate plots
+        plots = list()
```
lorenabalan:
```suggestion
        plots = []
```

kedro/contrib/io/matplotlib/matplotlib_writer.py
```diff
@@ -0,0 +1,114 @@
...
+        )
+
+    def _save(self, data: figure) -> None:
```
lorenabalan: The annotation says we expect only one type for `data`, but below we handle 2 more cases (list, dict), without exceptioning, which means they are also valid. Can we update the signature to reflect this?

tests/contrib/io/matplotlib/test_matplotlib_writer.py
```diff
@@ -0,0 +1,122 @@
...
+        # write and compare
+        trusted_filepath = tmp_path / "image_we_expect.png"
+        plt.savefig(str(trusted_filepath))
+
+        experimental_filepath = tmp_path / "image_we_write.png"
+        plot_writer = MatplotlibWriter(filepath=str(experimental_filepath))
+        plot_writer.save(plt)
+        plt.close()
+
+        assert trusted_filepath.read_bytes() == experimental_filepath.read_bytes()
```
lorenabalan: I would avoid using terms like `trusted` and `experimental`, as they only make sense to people with the business knowledge behind them. It probably makes a lot of sense to a DE, but not immediately obvious to other software engineers. Standard terminology like "actual" and "expected" is imo much clearer at conveying the message in tests and making them more readable (I don't have to think to translate).

tests/contrib/io/matplotlib/test_matplotlib_writer.py
```diff
@@ -0,0 +1,122 @@
...
+                str(index)
+            )
+            assert (
+                trusted_filepath.read_bytes() == full_experimental_filepath.read_bytes()
+            )
+
+    def test_dict_images(self, tmp_path):
+        plots = dict()
+        # generate plots
+        for index in ["boo", "far"]:
```
lorenabalan:
```suggestion
        for filename in ["boo.png", "far.png"]:
```

tests/contrib/io/matplotlib/test_matplotlib_writer.py
```diff
@@ -0,0 +1,122 @@
...
+            filename = "{}.png".format(index)
```
lorenabalan:
```suggestion

```

tests/contrib/io/matplotlib/test_matplotlib_writer.py
```diff
@@ -0,0 +1,122 @@
...
+
+        expected_load_error = "Loading not supported for MatplotlibWriter"
+        with pytest.raises(DataSetError, match=expected_load_error):
+            plot_writer.load()
+
+    def test_exists(self, tmp_path):
+        plot_object = plt.figure()
+        # pylint: disable=unsubscriptable-object,useless-suppression
+        plt.plot(np.random.rand(1, 5)[0], np.random.rand(1, 5)[0])
+        plt.close()
```
lorenabalan: Do we need to add this in the above tests too (test_save_list, test_save_dict)? I don't know if/what bad things happen.

tests/contrib/io/matplotlib/test_matplotlib_writer.py
921kiyo: Agree, I reused the original variables, but will update them :)

tests/contrib/io/matplotlib/test_matplotlib_writer.py
921kiyo: Good catch! `_exists()` does behave unexpectedly (returning False even after `.save` was called) with list and dict now, so I will fix it :)

kedro/contrib/io/matplotlib/matplotlib_writer.py
```diff
@@ -0,0 +1,123 @@
...
+        elif isinstance(data, dict):
+            self._filepath.mkdir(exist_ok=True)
+            for plot_name, plot in data.items():
+                plot.savefig(str(self._filepath / plot_name), **self._save_args)
+        else:
+            data.savefig(str(self._filepath), **self._save_args)
+
+    def _exists(self) -> bool:
+        return self._filepath.is_file() or (
+            self._filepath.is_dir() and len(list(self._filepath.iterdir())) != 0
```
lorenabalan: This can be further simplified to be:
```suggestion
            self._filepath.is_dir() and list(self._filepath.iterdir())
```

921kiyo: Seems like I have to use `len` because mypy doesn't like it and returns. 
` Incompatible return value type (got "Union[bool, List[Path]]", expected "bool")`

PR #41, opened 2019-06-30
[KED-519] Implement templated config

JoostDuisters:
## Notice

## Motivation and Context
This change introduces a TemplatedConfigLoader class with a resolve method. This resolve method allows for template variables, wrapped in '${..}', to be replaced with default values.

I chose for an implementation that requires a dict as input to provide the parameters. This allows for simplicity, lack of side-effects (no files get loaded in the background), and fewer parameters that the user needs to worry about (i.e. user doesn't need to specify whether they want to search config folder for a values.yml file, what the values.yml file should be called, etc.). 

If the user still wants to use a parameter file, they can simply use the normal ConfigLoader class to first load the parameter file in the same way they do for normal config, and then pass it into the TemplatedConfigLoader.

This also makes it easier for the user to mix environment variables into their template variable dictionary (I imagine the $USER environment variable can be very useful), using `os.environ()` syntax.

If the user wants to use templates in their parameter file, they can load the parameter file first using ConfigLoader, and then load the parameters file with TemplatedConfigLoader, passing the very same file in as a dictionary (this is admittedly very hacky but I feel like it would work for people that insist on this feature).

In future iterations we can perhaps work advanced logic like booleans or loops into this, if there is still demand for it.

## How has this been tested?
I created unittests for all the newly introduced functionality (100% coverage).

I ran linting (```make lint```) and unittests (```make test```).

I have not updated documentation yet.

## Checklist

- [X] Read the [contributing](/CONTRIBUTING.md) guidelines
- [X] Assigned myself to the PR
- [X] Added `Type` label to the PR


JoostDuisters:
This should now be linked to this issue: https://github.com/quantumblacklabs/kedro/issues/42

It implements a solution for templated config.

lorenabalan:
I've updated the PR title with our internal ticket number to keep track of this more easily. :)

kedro/contrib/config/templated_config.py
```diff
@@ -0,0 +1,151 @@
...
+
+class TemplatedConfigLoader(ConfigLoader):
+    """
+    Extension of the ConfigLoader class that allows for template values, wrapped in brackets like:
+     ${..}, to be replaced by default values.
+
+    Default values are provided in a dictionary.
+    """
+
+    def resolve(self, patterns: Union[str, List[str]],
```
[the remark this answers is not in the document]
JoostDuisters: I think the first option gives some nice properties, as you are forcing the user to explicitly define the `arg_values` argument, and if they don't it will just behave as the parent `get` method.

The second one is problematic because you want the `patterns` and `arg_values` argument are inherently tied together. Put differently, you want to allow the user to specify different `arg_values` arguments for different `patterns`.

kedro/contrib/config/templated_config.py
```diff
@@ -0,0 +1,151 @@
...
+        val: If this is a string of the format ${param_name}, it gets replaced by a parameter
+        defaults: A lookup from string to string with replacement values
+
+    Returns:
+        either the replacement value, if input val is a string
+
+    """
+
+    if isinstance(val, str):
+        regex_pattern = r'\$\{([^\}]*)\}'
```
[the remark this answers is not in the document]
JoostDuisters: I named them `pattern_full` and `pattern_partial` and added comment to explain why I need both.

tests/contrib/config/test_templated_config.py
```diff
@@ -0,0 +1,340 @@
...
+        assert catalog["boats"]["columns"]["top_speed"] == "${float_type}"
+        assert catalog["boats"]["users"] == ["fred", "${write_only_user}"]
+
+    @pytest.mark.usefixtures("proj_catalog_advanced")
+    def test_catlog_advanced(self, tmp_path, conf_paths):
+        """Test whether it responds well to advanced yaml values (i.e. nested dicts, booleans,
+        lists, etc.)"""
+        (tmp_path / "local").mkdir(exist_ok=True)
+
+        catalog = TemplatedConfigLoader(conf_paths).get("catalog*.yml")
```
DmitriiDeriabinQB: I'd suggest to pass globals dict even if there is nothing to replace since it would break the current flow as it can't handle non-string values (see my comment above)

kedro/contrib/config/templated_config.py
```diff
@@ -0,0 +1,185 @@
...
+            globals_pattern: Optional keyword-only argument specifying a glob pattern.
+                Files that match the pattern will be loaded as a dictionary with default
+                values used for replacement.
+            globals_dict: Optional keyword-only specifying an additional dictionary with
+                default values used for replacement. This dictionary will get merged with
+                the globals dictionary obtained from the globals_pattern. In case of
+                duplicate keys, the arg_values keys take precedence.
+        """
+
+        super(TemplatedConfigLoader, self).__init__(conf_paths)
```
DmitriiDeriabinQB:
```suggestion
        super().__init__(conf_paths)
```
For consistency with `get()` below :)

kedro/contrib/config/templated_config.py
```diff
@@ -0,0 +1,185 @@
...
+        if self._arg_dict:
+            return _replace_vals(config_raw, self._arg_dict)
+
+        return config_raw
+
+
+def _replace_vals(val: Any, defaults: Dict[str, Any]) -> Any:
+    """
+    Recursive function that loops through the values of a map. In case another map or a list is
+    encountered, it calls itsself. When a string is encountered it will use the default dict to
```
DmitriiDeriabinQB:
```suggestion
    encountered, it calls itself. When a string is encountered it will use the default dict to
```

kedro/contrib/config/templated_config.py
```diff
@@ -0,0 +1,185 @@
...
+
+        self._arg_dict = (
+            super(TemplatedConfigLoader, self).get(globals_pattern)
+            if globals_pattern
+            else {}
+        )
+
+        globals_dict = globals_dict or {}
+
+        self._arg_dict = dict(**self._arg_dict, **globals_dict)
```
DmitriiDeriabinQB:
```suggestion
        self._arg_dict = {**self._arg_dict, **globals_dict}
```
As `dict()` will complain is the keys of `self._arg_dict` and `globals_dict` overlap

kedro/contrib/config/templated_config.py
```diff
@@ -0,0 +1,185 @@
...
+
+import jmespath
+
+from kedro.config import ConfigLoader
+
+
+class TemplatedConfigLoader(ConfigLoader):
+    """
+    Extension of the ConfigLoader class that allows for template values, wrapped in brackets like:
+     ${..}, to be replaced by default values.
```
DmitriiDeriabinQB: I'd also mention that jmespath notation for retrieving nested values is also supported

tests/contrib/config/test_templated_config.py
```diff
@@ -0,0 +1,341 @@
...
+        assert catalog["boats"]["columns"]["top_speed"] == "${float_type}"
+        assert catalog["boats"]["users"] == ["fred", "${write_only_user}"]
+
+    @pytest.mark.usefixtures("proj_catalog_advanced")
+    def test_catlog_advanced(self, tmp_path, conf_paths):
+        """Test whether it responds well to advanced yaml values (i.e. nested dicts, booleans,
+        lists, etc.)"""
+        (tmp_path / "local").mkdir(exist_ok=True)
+
+        catalog = TemplatedConfigLoader(conf_paths, globals_dict={}).get("catalog*.yml")
```
DmitriiDeriabinQB: Maybe pass some globals in to make sure that they don't change the config?

DmitriiDeriabinQB:
@JoostDuisters can you please add the description of this PR along with your name to the `RELEASE.md`, and we're ready to merge! 🚀

klesouza:
I know it's already closed. But I was also interested in having similar templating options, although I thought of using Jinja builtin `anyconfig`. 
I started it [here](https://github.com/quantumblacklabs/kedro/compare/develop...klesouza:develop) 
Do you think it's worth considering reusing it instead of writing a custom rendering? We could join the two solutions, to also provide a way to have the context in a separate file and pass it to jinja (my solution for now, just get it from args in the command line)

PR #100, opened 2019-09-23
[KED-928] [Draft] Add support for a CSV Google Cloud Storage Data Connector

plauto:
## Notice

## Motivation and Context
This was created as a draft PR for [issue](https://github.com/quantumblacklabs/kedro/issues/11). It adds support for Google Cloud storage CSV datasets. Tests for VersionedDataset has to be added.
Would be great to get your early feedback (especially on how tests have been implemented).

## How has this been tested?
Testing needs pyvcr. API requests to google cloud services are intercepted and replayed using a cassette. This approach is similar to the one used by dask/gcsfs for testing their package

## Checklist

- [ ] Read the [contributing](/CONTRIBUTING.md) guidelines
- [] Assigned myself to the PR


921kiyo:
@plauto Thank you for opening PR. As an initial comment, I can see lots of large Yaml files were added. Could you please remove them (you can mock the GCP requests in unit tests)?

plauto:
@921kiyo the approach I have followed is the same used in the dask/gcsfs package (see [here](https://github.com/dask/gcsfs/tree/master/gcsfs/tests)). The yaml files contain the recording of API interactions between GCS and the client library. Unfortunately you have to mock a series of HTTP requests from the google APIs, while filtering out token refresh requests etc... An alternative approach could be to write an in memory simulator and replay those http requests from there (this might take a while to write though). 
From what I have seen around people either record these API requests/responses or use a test bucket in GCS. Unfortunately I could not see anything like what you have for testing S3. If you have other ideas I am totally up to try them

mzjp2:
Butting in here, what about doing something like what they do here: https://github.com/googleapis/google-cloud-python/blob/master/storage/tests/unit/test_client.py as an example, instead of all the `yaml` files?

Flid:
Hello @plauto 
I've checked for the alternative testing approaches (sort of what `moto` does for testing `boto`), but I can't fund anything useful. Looks like what you did is just what `gcsfs` creators suggest https://gcsfs.readthedocs.io/en/latest/developer.html#testing-and-vcr
It's a bit sad to keep massive files in the repo, and this approach is completely different from what we do in other places.
We could potentially write our own small mock class for `GCSFileSystem`, to make our tests more "unit-testy".

I'm OK with both approaches really. What's our strategic vision on it? @tolomea

tolomea:
I don't mind much either way. Given that this is done and works and fits with gcsfs practice it's fine.

yetudada:
@plauto We picked up this PR and thanks to you we'll have a `CSVGCSDataSet` in the next Kedro release. Look out for `kedro 0.15.5` and thank you so much for helping us with this. I'm going to close this PR in anticipation of the release.

PR #107, opened 2019-09-28
[KED-1076] Add open flag to build-docs that opens the documentation on build

mzjp2:
## Notice

## Motivation and Context
Why was this PR created?
Closes https://github.com/quantumblacklabs/kedro/issues/86

## How has this been tested?
Similar testing strategy as `kedro.cli` with `docs`.

## Checklist


mzjp2:
Turns out the issue was using pathlib's resolve, in Python3.5 if the filepath doesn't exist (as it doesn't when being mocked), an error is thrown, necessitating the use of mocking that method on the Path object. 

Incidentally, whilst researching this issue, I came across a few articles about the smell and danger of using `.assert_called_once` which the existing Kedro test that mocks `webbrowser` uses (and nowhere  else) here: 

https://github.com/quantumblacklabs/kedro/blob/d2c9648794cdda5ad28ea99cfc2661c076876932/tests/cli/test_cli.py#L185-L188

I think this would be better if `patched_browserr.assert_called_once()` was replaced with `patched_browser.call_count == 1` (I didn't think raising an issue or a PR was necessary for such a a tiny thing).

mzjp2:
Not sure it's worth editing the documentation (https://github.com/quantumblacklabs/kedro/blob/develop/docs/source/03_tutorial/05_package_a_project.md or https://github.com/quantumblacklabs/kedro/blob/develop/docs/source/06_resources/01_faq.md) so I haven't, but if you think it should - I'm happy to.

kedro/template/{{ cookiecutter.repo_name }}/kedro_cli.py
```diff
@@ -249,6 +253,11 @@ def build_docs():
         ]
     )
     call(["sphinx-build", "-M", "html", "docs/source", "docs/build", "-a"])
+    if open_docs:
+        p = Path("docs/build/html/index.html").resolve().as_uri()
+        secho("Opening {}".format(p))
+        webbrowser.open(str(p))
```
lorenabalan: Can we replace "p" here with a more descriptive variable name? :) "docs_page", "docs_home" or anything like that would suffice.

lorenabalan: Also I think the return of `as_uri()` is a `string` anyway, so no need for casting here. If you want to be on the safe side regardless, you could add it to line 257, so that would get passed through to `format` and `open` calls.

kedro/template/{{ cookiecutter.repo_name }}/kedro_cli.py
```diff
@@ -249,6 +253,11 @@ def build_docs():
         ]
     )
     call(["sphinx-build", "-M", "html", "docs/source", "docs/build", "-a"])
+    if open_docs:
+        p = Path("docs/build/html/index.html").resolve().as_uri()
```
lorenabalan: Given the difference in path separators, I suspect this will not play nice with Windows. How about going for something like this instead:
```(Path.cwd() / "docs" / "build" / "html" / "index.html").as_uri()```
and let `pathlib` handle the separators?
This might also help you get rid of the mock object in your test, which is nice.

File: kedro/template/{{ cookiecutter.repo_name }}/kedro_cli.py (hunk 1 above)
Comment (mzjp2) [reply kedro/template/{{ cookiecutter.repo_name }}/kedro_cli.py]: Thanks for the feedback, super valuable! 🚀

Should be fixed now. :)

File: kedro/template/{{ cookiecutter.repo_name }}/kedro_cli.py (hunk 2 above)
Comment (mzjp2) [reply kedro/template/{{ cookiecutter.repo_name }}/kedro_cli.py]: Ah, that's nice. Thanks, should be sorted now. :)

File: tests/template/test_kedro_cli.py [hunk 3]
```diff
@@ -317,6 +317,20 @@ def test_happy_path(self, call_mock, python_call_mock, fake_kedro_cli, mocker):
         )
         fake_rmtree.assert_called_once_with("docs/build", ignore_errors=True)
 
+    @pytest.mark.parametrize("open_flag", ["-o", "--open"])
+    def test_open_docs(
+        self, open_flag, fake_repo_path, call_mock, python_call_mock, fake_kedro_cli, mocker
```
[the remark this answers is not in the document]
mzjp2: Removed those, my bad 🙈 and now I don't need `fake_repo_path` anymore! 🎉

lorenabalan:
> Not sure it's worth editing the documentation (https://github.com/quantumblacklabs/kedro/blob/develop/docs/source/03_tutorial/05_package_a_project.md or https://github.com/quantumblacklabs/kedro/blob/develop/docs/source/06_resources/01_faq.md) so I haven't, but if you think it should - I'm happy to.

Just saw this, sorry. Maybe a note in `05_package_a_project.md` would be nice ("to quickly inspect this, run `my awesome command`"), but I wouldn't worry about it, it's far from being a blocker on merging this.

lorenabalan:
>  Maybe a note in `05_package_a_project.md` would be nice 

Actually `template/{{cookieutter.repo_name}}/README.md` would be more relevant.

PR #129, opened 2019-10-16
[KED-1121] Add MatplotlibWriter S3 DataSet

sheldontsen-qb:
## Notice

## Motivation and Context
Why was this PR created?

## How has this been tested?

## Checklist

- [ ] Read the [contributing](/CONTRIBUTING.md) guidelines
- [ ] Added a description of this change and added my name to the list of supporting contributions in the [`RELEASE.md`](/RELEASE.md) file


kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,173 @@
...
+        plot.savefig(bytes_object, **self._save_args)
+
+        session = boto3.Session(**self._boto_session_args)
+
+        session.client("s3", **self._s3_client_args).put_object(
+            Bucket=self._bucket,
+            Key=key_name,
+            **self._s3_put_object_args,
+            Body=bytes_object.getvalue(),
+        )
```
lorenabalan: The way I read this, it create a new boto client and a new boto session for every figure we save from the list/dict. Is there a specific reason for this, or can we just move these to live on the class (i.e. in the `__init__`) instead?

lorenabalan: We've also moved all our S3 datasets to use `s3fs` under the hood - would be ideal if we could keep it consistent across the board. This PR's approach customises a lot more kwargs than we do currently by default (see for instance `CSVS3DataSet`), I wonder what's the motivation behind it?
If they're all definitely required, it seems like all the different types of kwargs could be passed in on instantiating an `S3FileSystem`:
`s3_client_args` -> `client_kwargs`
`boto_session_args` -> `kwargs`
`s3_put_object_args` -> could perhaps be rolled into `s3_additional_kwargs`

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,173 @@
...
+        self._credentials = copy.deepcopy(credentials) or {}
+        self._s3_put_object_args = s3_put_object_args or {}
+
+        if self._credentials:
+            self._s3_client_args["aws_access_key_id"] = self._credentials[
+                "aws_access_key_id"
+            ]
+            self._s3_client_args["aws_secret_access_key"] = self._credentials[
+                "aws_secret_access_key"
+            ]
```
lorenabalan: Since this is the only place where the credentials are used, we don't have to store them on `self`, but use `_credentials` like suggested on line above.

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,173 @@
...
+                Must contain ``aws_access_key_id`` and ``aws_secret_access_key``.
+                Updates ``s3_client_args`` if provided.
+            load_args: Currently ignored as loading is not supported.
+            save_args: Save args passed to `plt.savefig`. See
+                https://matplotlib.org/api/_as_gen/matplotlib.pyplot.savefig.html
+        """
+
+        self._boto_session_args = boto_session_args or {}
+        self._s3_client_args = s3_client_args or {}
+        self._credentials = copy.deepcopy(credentials) or {}
```
lorenabalan:
```suggestion
       _credentials = copy.deepcopy(credentials) or {}
```

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,173 @@
...
+        s3_client_args: Optional[Dict[str, Any]] = None,
+        s3_put_object_args: Optional[Dict[str, Any]] = None,
+        credentials: Optional[Dict[str, Any]] = None,
+        load_args: Dict[str, Any] = None,
+        save_args: Dict[str, Any] = None,
+    ) -> None:
+        """Creates a new instance of ``MatplotlibWriter``.
+
+        Args:
+            bucket: Name of the bucket without "s3://" prefix
```
lorenabalan:
```suggestion
            bucket_name: Name of the bucket without "s3://" prefix
```

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,173 @@
...
+                bucket="my-super-great-bucket", filepath="matplotlib_list"
+            )
+            list_plot_writer.save(plots_list)
+
+    """
+
+    # pylint: disable=too-many-arguments
+    def __init__(
+        self,
+        bucket: str,
```
lorenabalan:
```suggestion
        bucket_name: str,
```
for consistency with our other S3 datasets :)

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
sheldontsen-qb: Agree on consistency

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
sheldontsen-qb: Agree on consistency

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
sheldontsen-qb: Makes sense!

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
sheldontsen-qb: Regarding comment 1: 
No specific reason, I was just trying to put all the s3 related stuff under 1 function, so that if someone else needs to modify for Azure or GCP (or who knows what else), the only modification they need to do is the `_save_to_s3` method and possibly the line above where `_save_to_s3` is called (for the default filenames). 

For comment 2:
Yes I noticed that it was different. Definitely agree to make things consistent. Let me try. I hacked this quite quickly together on a project using `boto` instead of `s3fs` hence why the PR uses the package.

Could you clarify on "customises lots of defaults"? There's only `_credentials`. Do you mean why so many options for args? ie `s3_client_args`, `boto_session_args`, `s3_put_object_args`? On occasion, this level of kwargs are required. You'll see there's a specific test for encryption because it's something I faced on client environment. On NC1, we don't face this level of security/issues, our bucket policy is very relaxed. So definitely a way is needed to pass all other possible arguments into `S3FileSystem`.

sheldontsen-qb:
@lorenabalan I could not figure out how to mock an encrypted bucket on moto. It seems to do nothing. I guess moto only mocks the calls..?

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,162 @@
...
+
+        _s3 = S3FileSystem(**self._s3fs_args)
+        self._s3 = _s3
+
+    def _describe(self) -> Dict[str, Any]:
+        return dict(
+            bucket_name=self._bucket_name,
+            filepath=self._filepath,
+            s3fs_args=self._s3fs_args,
+        )
```
lorenabalan: This is exposing sensitive information, like `aws_secret_access_key`, via public API (`describe`), which is not a good idea! I would remove this from the description and add the `save_args` instead.

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,162 @@
...
+        if _credentials:
+            if "client_kwargs" not in self._s3fs_args.keys():
+                self._s3fs_args["client_kwargs"] = {}
+
+            self._s3fs_args["client_kwargs"]["aws_access_key_id"] = _credentials[
+                "aws_access_key_id"
+            ]
+            self._s3fs_args["client_kwargs"]["aws_secret_access_key"] = _credentials[
+                "aws_secret_access_key"
+            ]
```
lorenabalan: Credentials should not be provided in `catalog.yml`, via `s3fs_args`, so I would strongly encourage to keep the 2 options separate. I'd remove this bit of code completely, leaving AWS keys delegated to `credentials`, and other optional arguments to `s3fs_args`.

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,162 @@
...
+            credentials: A dictionary of s3 access and secret keys.
+                Must contain ``aws_access_key_id`` and ``aws_secret_access_key``.
+                Updates ``s3_client_args`` if provided.
+            savefig_args: Save args passed to `plt.savefig`. See
+                https://matplotlib.org/api/_as_gen/matplotlib.pyplot.savefig.html
+        """
+
+        _credentials = copy.deepcopy(credentials) or {}
+
+        self._s3fs_args = s3fs_args or {}
```
lorenabalan:
```suggestion
        self._s3fs_args = copy.deepcopy(s3fs_args) or {}
```

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,162 @@
...
+            ]
+            self._s3fs_args["client_kwargs"]["aws_secret_access_key"] = _credentials[
+                "aws_secret_access_key"
+            ]
+
+        self._filepath = filepath
+        self._savefig_args = savefig_args if savefig_args else dict()
+        self._bucket_name = bucket_name
+
+        _s3 = S3FileSystem(**self._s3fs_args)
```
lorenabalan:
```suggestion
        _s3 = S3FileSystem(client_kwargs=_credentials, **self._s3fs_args)
```

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,162 @@
...
+        """Creates a new instance of ``MatplotlibWriter``.
+
+        Args:
+            bucket_name: Name of the bucket without "s3://" prefix
+            filepath: Path to a matplot object file.
+            s3fs_args: Dictionary of arguments
+            credentials: A dictionary of s3 access and secret keys.
+                Must contain ``aws_access_key_id`` and ``aws_secret_access_key``.
+                Updates ``s3_client_args`` if provided.
+            savefig_args: Save args passed to `plt.savefig`. See
```
lorenabalan:
```suggestion
            save_args: Save args passed to `plt.savefig`. See
```

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,162 @@
...
+    """
+
+    # pylint: disable=too-many-arguments
+    def __init__(
+        self,
+        bucket_name: str,
+        filepath: str,
+        s3fs_args: Optional[Dict] = None,
+        credentials: Optional[Dict[str, Any]] = None,
+        savefig_args: Dict[str, Any] = None,
```
lorenabalan:
```suggestion
        save_args: Dict[str, Any] = None,
```

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,162 @@
...
+    ) -> None:
+        """Creates a new instance of ``MatplotlibWriter``.
+
+        Args:
+            bucket_name: Name of the bucket without "s3://" prefix
+            filepath: Path to a matplot object file.
+            s3fs_args: Dictionary of arguments
```
lorenabalan: Not very clear where these go, a link to the `S3FileSystem` docs would be useful, or a mention that this is where you'd specify the encryption type.

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,162 @@
...
+
+        else:
+            self._save_to_s3(key_name=self._filepath, plot=data)
+
+    def _save_to_s3(self, key_name, plot):
+
+        bytes_object = io.BytesIO()
+        plot.savefig(bytes_object, **self._savefig_args)
+
+        full_key_path = "/".join([self._bucket_name, key_name])
```
lorenabalan:
```suggestion
        full_key_path = "{}/{}".format(self._bucket_name, key_name)
```

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,162 @@
...
+
+        elif isinstance(data, dict):
+            for plot_name, plot in data.items():
+                key_path = self._filepath + "/" + plot_name
+                self._save_to_s3(key_name=key_path, plot=plot)
+
+        else:
+            self._save_to_s3(key_name=self._filepath, plot=data)
+
+    def _save_to_s3(self, key_name, plot):
```
lorenabalan: Type annotations here would be nice. :)

tests/contrib/io/matplotlib/test_matplotlib_s3_writer.py
```diff
@@ -0,0 +1,242 @@
...
+        conn.create_bucket(Bucket=BUCKET_NAME)
+        conn.put_bucket_policy(Bucket=BUCKET_NAME, Policy=bucket_policy)
+
+        yield conn
+
+
+# @mock_s3
+# def single_plot_writer():
+#     return MatplotlibWriterS3(bucket_name=BUCKET_NAME, filepath=KEY_PATH)
+
```
lorenabalan: Dead code.

tests/contrib/io/matplotlib/test_matplotlib_s3_writer.py
```diff
@@ -0,0 +1,242 @@
...
+    assert actual_filepath.read_bytes() == expected_path.read_bytes()
+
+
+def test_list_save(tmp_path, mock_list_plot, plot_writer, mocked_s3_bucket):
+    """Test saving list of plots to S3."""
+    # iterable_plot_writer = MatplotlibWriterS3(
+    #     bucket_name=BUCKET_NAME, filepath=KEY_PATH
+    # )
+
```
lorenabalan: Dead code.

tests/contrib/io/matplotlib/test_matplotlib_s3_writer.py
```diff
@@ -0,0 +1,242 @@
...
+            bucket_name=BUCKET_NAME, filepath=KEY_PATH, credentials=AWS_CREDENTIALS
+        )
+
+    return _matplotlibwriters3()
+
+
+def test_save_data(tmp_path, mock_single_plot, plot_writer, mocked_s3_bucket):
+    """Test saving single matplotlib plot to S3."""
+
+    # single_plot_writer = MatplotlibWriterS3(bucket_name=BUCKET_NAME, filepath=KEY_PATH)
```
lorenabalan: Dead code.

tests/contrib/io/matplotlib/test_matplotlib_s3_writer.py
```diff
@@ -0,0 +1,242 @@
...
+
+        assert actual_filepath.read_bytes() == expected_path.read_bytes()
+
+
+def test_dict_save(tmp_path, mock_dict_plot, plot_writer, mocked_s3_bucket):
+    """Test saving dictionary of plots to S3."""
+
+    # iterable_plot_writer = MatplotlibWriterS3(
+    #     bucket_name=BUCKET_NAME, filepath=KEY_PATH
+    # )
```
lorenabalan: Dead code.

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,162 @@
...
+        credentials: Optional[Dict[str, Any]] = None,
+        savefig_args: Dict[str, Any] = None,
+    ) -> None:
+        """Creates a new instance of ``MatplotlibWriter``.
+
+        Args:
+            bucket_name: Name of the bucket without "s3://" prefix
+            filepath: Path to a matplot object file.
+            s3fs_args: Dictionary of arguments
+            credentials: A dictionary of s3 access and secret keys.
```
lorenabalan:
```suggestion
            credentials: A dictionary of S3 access and secret keys.
```

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,162 @@
...
+#     or use the QuantumBlack Trademarks in any other manner that might cause
+# confusion in the marketplace, including but not limited to in advertising,
+# on websites, or on software.
+#
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""
+``MatplotlibWriterS3`` saves matplotlib objects as image files to s3.
```
lorenabalan:
```suggestion
``MatplotlibWriterS3`` saves matplotlib objects as image files to S3.
```

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
sheldontsen-qb: This is related to the change suggested below as well, putting here for easier reference.
`_s3 = S3FileSystem(client_kwargs=_credentials, **self._s3fs_args)`

I'm a bit confused. `credentials` are provided separately:
```
    def __init__(
        self,
        bucket_name: str,
        filepath: str,
        s3fs_args: Optional[Dict] = None,
        credentials: Optional[Dict[str, Any]] = None,
        savefig_args: Dict[str, Any] = None,
    ) -> None:
```
I just added them to the `client_kwargs` so that I do not lose the ability to provide other options to `client_kwargs`.

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
sheldontsen-qb: Done

tests/contrib/io/matplotlib/test_matplotlib_s3_writer.py
sheldontsen-qb: Done

tests/contrib/io/matplotlib/test_matplotlib_s3_writer.py
sheldontsen-qb: Done

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
sheldontsen-qb: Done

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
sheldontsen-qb: Done

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
sheldontsen-qb: Done

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
sheldontsen-qb: Depending on your response below

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
sheldontsen-qb: Depending on your response above

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
lorenabalan: I get you - would these extra options qualify to be passed in through credentials directly, like the aws keys? Looking at the API of `Session.client` in `boto3`, where the `client_kwargs` are passed, that would include: region_name, api_version, verify, endpoint_url, aws_session_token. The last one for instance should definitely be treated as credential.

sheldontsen-qb: Hmmm... Yes, I see..

So, should it be:
```
    def __init__(
        self,
        ...
        # docstring tries to point out that for client_kwargs should be fed in here
        credentials: Optional[Dict[str, Any]] = None, 
        ...
    ) -> None:
```

Or
```
So, should it be:
    def __init__(
        self,
        ...
        # docstring tries to point out that keys should be feed in here via credentials
        client_kwargs: Optional[Dict[str, Any]] = None,
        ...
    ) -> None:
```

1st one is more consistent with the rest of the classes. But option 2 I think is a bit clearer and matches the underlying library a bit better.

lorenabalan: The way I see it, the user/interface doesn't need to know about the underlying backend, and the second approach would make the two (interface and backend) too tightly coupled. If we one day decide, for whatever reason, that we don't use `s3fs` but something else new and cool, `client_kwargs` won't make sense anymore and we're left with either tech debt or breaking changes. So I'd prefer the first approach, maybe specifying in docs what you can put in there/where it goes, you think it's useful. It also means it stays consistent with the other datasets, where the way to specify options for the client would still be via `credentials`.

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
sheldontsen-qb: Done

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
sheldontsen-qb: No longer relevant since keys are passed to `credentials` not `s3fs_args`

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
sheldontsen-qb: Yeah makes sense! Done

lorenabalan:
Also before we merge don't forget to tick the checkboxes in the PR description about contribution guide & legal agreement. :)

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
sheldontsen-qb: Done!

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
sheldontsen-qb: Done!

sheldontsen-qb:
> Also before we merge don't forget to tick the checkboxes in the PR description about contribution guide & legal agreement. :)

Ah I missed those.. got buried. Addressing now

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,150 @@
...
+
+        else:
+            self._save_to_s3(key_name=self._filepath, plot=data)
+
+    def _save_to_s3(self, key_name: str, plot: figure):
+
+        bytes_object = io.BytesIO()
+        plot.savefig(bytes_object, **self._save_args)
+
+        full_key_path = "{}/{}".format(self._bucket_name, key_name)
```
921kiyo: Is it necessary to construct the key path in here and within `_save` separately? I would do this line inside `_save`, which is simpler and you can remove `key_name` arg in this func.

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,150 @@
...
+            for plot_name, plot in data.items():
+                key_path = self._filepath + "/" + plot_name
+                self._save_to_s3(key_name=key_path, plot=plot)
+
+        else:
+            self._save_to_s3(key_name=self._filepath, plot=data)
+
+    def _save_to_s3(self, key_name: str, plot: figure):
+
+        bytes_object = io.BytesIO()
```
921kiyo: I think `bytes_buffer` rather than `bytes_object` makes more sense.

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,150 @@
...
+        plot.savefig(bytes_object, **self._save_args)
+
+        full_key_path = "{}/{}".format(self._bucket_name, key_name)
+
+        with self._s3.open(str(full_key_path), mode="wb") as s3_file:
+            s3_file.write(bytes_object.getvalue())
```
921kiyo: Just curious, how would users be able to see the saved plot (in a human readable way, not bytes), since load is not supported for this dataset?

kedro/contrib/io/matplotlib/__init__.py
```diff
@@ -31,4 +31,5 @@
 ``AbstractDataSet`` implementation to save matplotlib objects as image files.
 """
 
+from .matplotlib_s3_writer import MatplotlibWriterS3  # NOQA
```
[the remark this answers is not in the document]
sheldontsen-qb: What about just `MatplotlibLocal` and `MatplotlibS3`? Do we need the `-Writer`?

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+class MatplotlibWriterS3(AbstractDataSet):
+    # pylint: disable=too-many-instance-attributes
+    """
+    ``MatplotlibWriter`` saves matplotlib objects as image files.
+
+    Example:
+    ::
+
```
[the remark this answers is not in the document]
sheldontsen-qb: Is it necessary to add the `>>>`? I prefer without because it means I can copy paste the example and it will work. By adding the `>>>`, I always have to copy paste, and delete `>>>` from every line.

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,150 @@
...
+                and ``aws_secret_access_key`` are provided here.
+            save_args: Save args passed to `plt.savefig`. See
+                https://matplotlib.org/api/_as_gen/matplotlib.pyplot.savefig.html
+        """
+
+        _credentials = copy.deepcopy(credentials) or {}
+
+        self._s3fs_args = copy.deepcopy(s3fs_args) or {}
+
+        self._filepath = filepath
```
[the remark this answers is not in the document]
sheldontsen-qb: In this case `filepath` refers to keypath, so not sure if you want it to be a `Path` object?

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,150 @@
...
+        self._save_args = save_args if save_args else dict()
+        self._bucket_name = bucket_name
+
+        self._s3 = S3FileSystem(client_kwargs=_credentials, **self._s3fs_args)
```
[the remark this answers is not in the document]
sheldontsen-qb: I have definitely faced the issue where I needed to supply more than just `client_kwargs`.

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,150 @@
...
+        elif isinstance(data, dict):
+            for plot_name, plot in data.items():
+                key_path = self._filepath + "/" + plot_name
+                self._save_to_s3(key_name=key_path, plot=plot)
+
+        else:
+            self._save_to_s3(key_name=self._filepath, plot=data)
+
+    def _save_to_s3(self, key_name: str, plot: figure):
+
```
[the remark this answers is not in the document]
sheldontsen-qb: I just let `black` do it's thing though

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
sheldontsen-qb: 👍

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
sheldontsen-qb: You would have to download the plots from s3 to view them.

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,150 @@
...
+        self._save_args = save_args if save_args else dict()
+        self._bucket_name = bucket_name
+
+        self._s3 = S3FileSystem(client_kwargs=_credentials, **self._s3fs_args)
+
+    def _describe(self) -> Dict[str, Any]:
+        return dict(
+            bucket_name=self._bucket_name,
+            filepath=self._filepath,
+            s3fs_args=self._s3fs_args,
```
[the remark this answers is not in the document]
sheldontsen-qb: Oops yes!

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,150 @@
...
+        list_plot_writer = MatplotlibWriterS3(
+            bucket="my-super-great-bucket", filepath="matplotlib_list"
+        )
+        list_plot_writer.save(plots_list)
+    """
+
+    # pylint: disable=too-many-arguments
+    def __init__(
+        self,
+        bucket_name: str,
```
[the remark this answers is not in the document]
sheldontsen-qb: Sure. Though I always thing big to small, so in my mind, bucket comes before key path

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
sheldontsen-qb: Simpler, but still will need a `key_name` or `key_path` argument though

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,150 @@
...
+        import matplotlib.pyplot as plt
+        from kedro.contrib.io.matplotlib_s3_writer import MatplotlibWriterS3
+
+        # Saving single plot
+        plt.plot([1, 2, 3], [4, 5, 6])
+        single_plot_writer = MatplotlibWriterS3(
+            bucket="my-super-great-bucket", filepath="matplot_lib_single_plot.png"
+        )
+        single_plot_writer.save(plt)
+
```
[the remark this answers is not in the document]
sheldontsen-qb: Done!

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,150 @@
...
+        # Saving dictionary of plots (with SSE)
+        plots_dict = {}
```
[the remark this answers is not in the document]
sheldontsen-qb: Done

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,150 @@
...
+from matplotlib.pyplot import figure
+from s3fs import S3FileSystem
+
+from kedro.io import AbstractDataSet, DataSetError
+
+
+class MatplotlibWriterS3(AbstractDataSet):
+    # pylint: disable=too-many-instance-attributes
+    """
+    ``MatplotlibWriter`` saves matplotlib objects as image files.
```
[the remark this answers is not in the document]
sheldontsen-qb: I prefer `MatplotlibS3Writer`

kedro/contrib/io/matplotlib/__init__.py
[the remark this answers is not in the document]
sheldontsen-qb: Renamed to `MatplotlibS3Writer`

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,150 @@
...
+
+    Example:
+    ::
+
+        import matplotlib.pyplot as plt
+        from kedro.contrib.io.matplotlib_s3_writer import MatplotlibWriterS3
```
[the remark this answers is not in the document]
sheldontsen-qb: Done

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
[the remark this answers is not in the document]
sheldontsen-qb: Done

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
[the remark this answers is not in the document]
921kiyo: I understand, I always get annoyed by that actually. `>>>` is more for the docs (see example code. https://kedro.readthedocs.io/en/latest/kedro.io.CSVLocalDataSet.html#kedro.io.CSVLocalDataSet)

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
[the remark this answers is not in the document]
921kiyo: I will raise a ticket to fix other datasets :)

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,153 @@
...
+        else:
+            full_key_path = "/".join([self._bucket_name, self._filepath])
+            self._save_to_s3(full_key_path=full_key_path, plot=data)
+
+    def _save_to_s3(self, full_key_path: str, plot: figure):
+        bytes_buffer = io.BytesIO()
+        plot.savefig(bytes_buffer, **self._save_args)
+
+        with self._s3.open(full_key_path, mode="wb") as s3_file:
+            s3_file.write(bytes_buffer.getvalue())
```
921kiyo: I think `_exists()` method is missing

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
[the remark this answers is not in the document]
sheldontsen-qb: done!

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
sheldontsen-qb: Done - added tests also

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def _exists(self) -> bool:
+        load_path = "/".join([self._bucket_name, self._filepath])
+        return self._s3.isdir(load_path) or self._s3.isfile(load_path)
```
921kiyo: `self._s3.isdir(load_path)` only checks the existence of the directory, but not checking if something is in the directory. You could use `walk` to check if there is anything in the path. https://s3fs.readthedocs.io/en/latest/api.html#s3fs.core.S3FileSystem.walk

tests/contrib/io/matplotlib/test_matplotlib_s3_writer.py
```diff
@@ -0,0 +1,238 @@
...
+            }
+        ],
+    }
+    bucket_policy = json.dumps(bucket_policy)
+
+    with mock_s3():
+        conn = s3fs.core.boto3.client("s3", **AWS_CREDENTIALS)
+        conn.create_bucket(Bucket=BUCKET_NAME)
+        conn.put_bucket_policy(Bucket=BUCKET_NAME, Policy=bucket_policy)
+
```
921kiyo:
```suggestion
```

tests/contrib/io/matplotlib/test_matplotlib_s3_writer.py
```diff
@@ -0,0 +1,238 @@
...
+        yield conn
+
+
+@pytest.fixture
+def plot_writer(mocked_s3_bucket):  # pylint: disable=unused-argument
+    def _matplotlibwriters3():
```
921kiyo: Why there is an inner func? I think you can just return the instance of the class.

tests/contrib/io/matplotlib/test_matplotlib_s3_writer.py
```diff
@@ -0,0 +1,238 @@
...
+
+def test_save_data(tmp_path, mock_single_plot, plot_writer, mocked_s3_bucket):
+    """Test saving single matplotlib plot to S3."""
+
+    plot_writer.save(mock_single_plot)
+
+    expected_path = tmp_path / "downloaded_image.png"
+    actual_filepath = tmp_path / "locally_saved.png"
+
+    plt.savefig(str(actual_filepath))
```
921kiyo: Why do you need to save locally? You can save the plot object to S3, and download it and check? Same for list and dict test.

tests/contrib/io/matplotlib/test_matplotlib_s3_writer.py
```diff
@@ -0,0 +1,238 @@
...
+        mock_dict_plot[colour].savefig(str(actual_filepath))
+
+        _key_path = KEY_PATH + "/" + colour
+
+        mocked_s3_bucket.download_file(BUCKET_NAME, _key_path, str(expected_path))
+
+        assert actual_filepath.read_bytes() == expected_path.read_bytes()
+
+
+def test_bad_credentials(mock_dict_plot):
```
921kiyo: Shouldn't this fail at the construction because `bad_writer.save` does not use credentials?

tests/contrib/io/matplotlib/test_matplotlib_s3_writer.py
```diff
@@ -0,0 +1,238 @@
...
+            "aws_secret_access_key": "definitely_not_for_testing",
+        },
+    )
+    pattern = "InvalidAccessKeyId"
+
+    with pytest.raises(DataSetError, match=pattern):
+        bad_writer.save(mock_dict_plot)
+
+
+def test_credentials(tmp_path, mock_single_plot, mocked_s3_bucket):
```
921kiyo: Looks like this test is doing exactly the same as `test_save_data` above.

921kiyo: I would also test `test_s3fs_args` as an additional test

tests/contrib/io/matplotlib/test_matplotlib_s3_writer.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+def test_load_fail(plot_writer):
+    pattern = r"Loading not supported for `MatplotlibS3Writer`"
+    with pytest.raises(DataSetError, match=pattern):
+        plot_writer.load()
+
+
+def test_exists_single(mock_single_plot, plot_writer):
+    plot_writer.save(mock_single_plot)
```
921kiyo: You can do `assert not plot_writer.exists()` before this line. As for the next test.

tests/contrib/io/matplotlib/test_matplotlib_s3_writer.py
```diff
@@ -0,0 +1,238 @@
...
+    plot_writer.exists()
```
921kiyo:
```suggestion
    assert plot_writer.exists()
```

tests/contrib/io/matplotlib/test_matplotlib_s3_writer.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+def test_exists_multiple(mock_dict_plot, plot_writer):
+    plot_writer.save(mock_dict_plot)
+    plot_writer.exists()
```
921kiyo:
```suggestion
    assert plot_writer.exists()
```

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,157 @@
...
+        >>>
+        >>> # Saving single plot
+        >>> plt.plot([1, 2, 3], [4, 5, 6])
+        >>> single_plot_writer = MatplotlibS3Writer(
+        >>>     bucket="my-super-great-bucket", filepath="matplot_lib_single_plot.png"
+        >>> )
+        >>> single_plot_writer.save(plt)
+        >>> plt.close()
+        >>>
+        >>> # Saving dictionary of plots (with SSE)
```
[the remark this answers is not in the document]
sheldontsen-qb: Good spot. IDK how this got in there - will remove. Pretty sure it's Sum of Squared Error

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,157 @@
...
+        bucket_name: str,
+        s3fs_args: Optional[Dict] = None,
+        credentials: Optional[Dict[str, Any]] = None,
+        save_args: Dict[str, Any] = None,
+    ) -> None:
+        """Creates a new instance of ``MatplotlibS3Writer``.
+
+        Args:
+            bucket_name: Name of the bucket without "s3://" prefix.
+            filepath: Key path to a matplot object file.
```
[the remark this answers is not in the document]
sheldontsen-qb: Changed to `Key path to matplot object file(s).`

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,157 @@
...
+            s3fs_args: Arguments for ``S3FileSystem``. See
+                https://s3fs.readthedocs.io/en/latest/api.html#s3fs.core.S3FileSystem
+            credentials: Argument for ``client_kwargs``. If needed ``aws_access_key_id``
```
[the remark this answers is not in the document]
sheldontsen-qb: Done

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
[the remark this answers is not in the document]
sheldontsen-qb: I read this as no action for this PR - resolving.

tests/contrib/io/matplotlib/test_matplotlib_s3_writer.py
sheldontsen-qb: I guess the credentials are only used when you try to write to s3, not during class construction?

tests/contrib/io/matplotlib/test_matplotlib_s3_writer.py
sheldontsen-qb: When I started, I was copying the pattern from the local matplotlib tests. But to me the current pattern makes sense, you want to check that you still get the same thing using matplotlib's native save method, vs save to buffer and push to s3 and re-download.

tests/contrib/io/matplotlib/test_matplotlib_s3_writer.py
sheldontsen-qb: Yes you are right. Left over code from when I was experimenting with moto

sheldontsen-qb: Fixed

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
sheldontsen-qb: Ah did I not push something for this. I used `iterdir()`:

```
     def _exists(self) -> bool:
         return self._filepath.is_file() or (
-            self._filepath.is_dir() and bool(list(self._filepath.iterdir()))
+            self._filepath.is_dir() and bool(list(self._filepath.walk()))
         )

```

sheldontsen-qb: Anyway, both seem to work!

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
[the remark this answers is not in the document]
sheldontsen-qb: Changed to `Path`

tests/contrib/io/matplotlib/test_matplotlib_s3_writer.py
sheldontsen-qb: `test_s3_encryption` below actually covers using `s3fs_args`. Should I just rename from `test_credentials` to `test_s3fs_args`?

sheldontsen-qb: Renaming from `test_s3_encryption` to `test_s3fs_args`.

Tbh, I cannot think of any other arg to test, nor whether `moto` only mocks the calls and not the actual s3 properties.

tests/contrib/io/matplotlib/test_matplotlib_s3_writer.py
sheldontsen-qb: Done!

kedro/contrib/io/matplotlib/matplotlib_s3_writer.py
```diff
@@ -0,0 +1,158 @@
...
+# BE LIABLE FOR ANY CLAIM, DAMAGES, OR OTHER LIABILITY, WHETHER IN AN
+# ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF, OR IN
+# CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+#
+# The QuantumBlack Visual Analytics Limited ("QuantumBlack") name and logo
+# (either separately or in combination, "QuantumBlack Trademarks") are
+# trademarks of QuantumBlack. The License does not grant you any right or
+# license to the QuantumBlack Trademarks. You may not use the QuantumBlack
+# Trademarks or any confusingly similar mark as a trademark for your product,
+#     or use the QuantumBlack Trademarks in any other manner that might cause
```
sheldontsen-qb: @921kiyo is the tab here intentional?

921kiyo: Not sure, but the ci will fail if you fixed it I think. So you can just leave it.

sheldontsen-qb: okok

PR #156, opened 2019-10-31
[KED-1198] Implement Databricks-compatible dataset versioning

deepyaman:
## Notice

## Motivation and Context
Why was this PR created?

_Copied from internal Slack channel:_

Hi Dmitrii. Thanks for following up. Unfortunately, I’m still facing issues…

To recap, my data is Azure Blob Storage, and we’re running on Databricks. As a result, I’m accessing via `mnt` paths.

Unfortunately, `/mnt` is not accessible by local file APIs. @prasad.gaikwad essentially pointed this out on my other question (use `/dbfs/mnt/...` as the `filepath` for Pandas-based datasets, instead). I see that echoed in https://forums.databricks.com/answers/14051/view.html -> https://docs.databricks.com/data/databricks-file-system.html#local-file-apis

Therefore, a `filepath` like `/mnt/...` for a versioned `SparkDataSet` fails, because it uses `iglob`--a local file API, trying to access something starting with `/mnt`--to find versions. https://github.com/quantumblacklabs/kedro/blob/develop/kedro/io/core.py#L486 <-- this line returns `[]`

If you specify the `filepath` for a versioned `SparkDataSet` starting with `/dbfs/mnt` instead, it actually finds the versions! However… the `_load` method comes back with: `org.apache.spark.sql.AnalysisException: Path does not exist: dbfs:/dbfs/mnt/...` T_T

I assume Databricks is throwing that extra `dbfs:/` up front, even if it’s prefixed by `/dbfs` already. What would you suggest? Standardize on specifying paths with `/dbfs` for Databricks, but special case drop the `/dbfs` prefix up front in the `_load` or something?

## How has this been tested?

Manually tested on my current project (when no versions exist, and with existing versions).

~TODO: Add test coverage for new lines.~ Also, unit tests (see https://github.com/quantumblacklabs/kedro/pull/156#issuecomment-557505826).

## Checklist

- [ ] Added a description of this change and added my name to the list of supporting contributions in the [`RELEASE.md`](/RELEASE.md) file


kedro/contrib/io/pyspark/spark_data_set.py
```diff
@@ -235,19 +235,23 @@ def _get_spark():
 
     def _load(self) -> DataFrame:
         load_path = self._fs_prefix + str(self._get_load_path())
+        if load_path.startswith("/dbfs"):
```
DmitriiDeriabinQB: Since it's already used 3 times, maybe worth moving into a helper function like `_remove_dbfs_prefix`? Similar to `_parse_glob_pattern`, for example.

deepyaman: Great suggestion; will do! :)

deepyaman: Done! Only took me three weeks. :P

deepyaman:
Pushed fbadd0c to demonstrate that the tests cover the negative if branch for all 3 cases: `save`, `load`, and `exists`. Otherwise, build passes without additional tests (or just one additional test).

deepyaman:
> Apart from tests, could you please also document your change, so that the users know that it solves a specific issue on Databricks.

@DmitriiDeriabinQB Where would you suggest documenting this change--in `pyspark/README.md` or in the docs for working with Databricks? I tried explaining it in the release notes, but it seemed kinda long (I can add a parenthesized reference to see comments elsewhere).

DmitriiDeriabinQB:
@deepyaman I would put it into `pyspark/README.md` _and_ probably also into `SparkDataSet.__init__` docstring - as a note for the `filepath` argument, for example.

RELEASE.md
```diff
@@ -9,12 +9,13 @@
 * Fixed an issue where `%reload_kedro` wasn't reloading all user modules.
 * Fixed `pandas_to_spark` and `spark_to_pandas` decorators to work with functions with kwargs.
 * Fixed a bug where `kedro jupyter notebook` and `kedro jupyter lab` would run a different Jupyter installation to the one in the local environment.
+* Implemented Databricks-compatible dataset versioning.
```
lorenabalan:
```suggestion
* Implemented Databricks-compatible dataset versioning for `SparkDataSet`.
```
maybe

kedro/contrib/io/pyspark/spark_data_set.py
```diff
@@ -62,6 +62,10 @@ def _split_filepath(filepath: str) -> Tuple[str, str]:
     return "", split_[0]
 
 
+def _strip_dbfs_prefix(path: str) -> str:
+    return path[5:] if path.startswith("/dbfs") else path
```
lorenabalan: To make it more obvious where the magic 5 is coming from.
```suggestion
    prefix = "/dbfs"
    return path[len(prefix):] if path.startswith(prefix) else path
```

kedro/contrib/io/pyspark/spark_data_set.py
```diff
@@ -152,7 +156,10 @@ def __init__(  # pylint: disable=too-many-arguments
         """Creates a new instance of ``SparkDataSet``.
 
         Args:
-            filepath: path to a Spark data frame.
+            filepath: path to a Spark data frame. When using Databricks
+                and working with data written to mount path points,
+                specify ``filepath``s for (versioned) ``SparkDataSet``s
+                starting with ``/dbfs/mnt``.
```
lorenabalan: How come the prefix we're stripping out is just "/dbfs"?

kedro/contrib/io/pyspark/spark_data_set.py
```diff
@@ -233,20 +240,18 @@ def _get_spark():
...
         return self._get_spark().read.load(
             load_path, self._file_format, **self._load_args
         )
 
     def _save(self, data: DataFrame) -> None:
-        save_path = str(self._get_save_path())
-        data.write.save(
-            self._fs_prefix + save_path, self._file_format, **self._save_args
-        )
+        save_path = _strip_dbfs_prefix(self._fs_prefix + str(self._get_save_path()))
```
lorenabalan: How come we need `_fs_prefix` this time but not previously?

tests/contrib/io/pyspark/test_spark_data_set.py
```diff
@@ -405,6 +411,70 @@ def test_prevent_overwrite(self, tmp_path, version, sample_spark_df):
...
+        versioned_dataset_dbfs.save(sample_spark_df)
+        assert versioned_dataset_dbfs.exists()
+
+        assert mocked_glob.call_count == 3
+        mocked_glob.assert_has_calls(
+            [
+                call("/dbfs" + str(tmp_path / FILENAME / "*" / FILENAME)),
+                call("/dbfs" + str(tmp_path / FILENAME / "*" / FILENAME)),
+                call("/dbfs" + str(tmp_path / FILENAME / "*" / FILENAME)),
+            ]
```
lorenabalan: [call(...)] * 3

tests/contrib/io/pyspark/test_spark_data_set.py
```diff
@@ -405,6 +411,70 @@ def test_prevent_overwrite(self, tmp_path, version, sample_spark_df):
...
+        mocked_glob = mocker.patch.object(versioned_dataset_dbfs, "_glob_function")
+        mocked_glob.return_value = [str(tmp_path / FILENAME / version.save / FILENAME)]
+
+        assert not versioned_dataset_dbfs.exists()
+
+        versioned_dataset_dbfs.save(sample_spark_df)
+        assert versioned_dataset_dbfs.exists()
+
+        assert mocked_glob.call_count == 3
+        mocked_glob.assert_has_calls(
```
lorenabalan: You could actually probably merge these two checks by doing sth like `assert mocked_glob.call_args_list == expected_calls` etc. But at the same time, why do you want to check the glob function calls here?

tests/contrib/io/pyspark/test_spark_data_set.py
```diff
@@ -405,6 +411,70 @@ def test_prevent_overwrite(self, tmp_path, version, sample_spark_df):
...
+        self, mocker, versioned_dataset_dbfs, version, tmp_path, sample_spark_df
+    ):
+        mocked_glob = mocker.patch.object(versioned_dataset_dbfs, "_glob_function")
+        mocked_glob.return_value = [str(tmp_path / FILENAME / version.save / FILENAME)]
+
+        versioned_dataset_dbfs.save(sample_spark_df)
+        reloaded = versioned_dataset_dbfs.load()
+
+        assert mocked_glob.call_count == 2
+        mocked_glob.assert_has_calls(
```
lorenabalan: Same idea with `call_args_list`

lorenabalan:
Had a chat with @DmitriiDeriabinQB and it's much clearer in my head what's going on now. Feel free to ignore my questions, just my comments about test code stand.

deepyaman:
> @deepyaman I would put it into `pyspark/README.md` _and_ probably also into `SparkDataSet.__init__` docstring - as a note for the `filepath` argument, for example.

Done!

kedro/contrib/io/pyspark/spark_data_set.py
deepyaman: In the past, `_fs_prefix` was added in the `save` call; i.e.

```
        data.write.save(
            self._fs_prefix + save_path, self._file_format, **self._save_args	
        )
```

I moved it up into the `save_path` definition, to make it consistent with how `_fs_prefix` is being added in the `load_path` definition (and not in the `load` call).

tests/contrib/io/pyspark/test_spark_data_set.py
deepyaman: The reason I'm checking glob calls is to make sure that version code is getting called for each `_load`/`_save`/`_exists` call... I think? Maybe you're right, it's an implementation detail that has no need to be tested (so what if the code changed to call `iglob` 10 times repeatedly in the future... as long as it still works, and is there for another reason, it shouldn't affect the passing of this test). Shall I remove them @lorenabalan?

deepyaman:
Thanks @DmitriiDeriabinQB @lorenabalan! :D

PR #202, opened 2020-01-15
Fix bug in to_nodes and from_nodes for pipeline with transcoding

PedroAbreuQB:
## Notice

## Motivation and Context
`from_nodes` and `to_nodes` do not deal well with transcoded datasets. A minimal example would be a pipeline 1 -> A -> 2, where node 1 writes `A@1` and node 2 reads `A@2`. Running `from_nodes(1)` or `to_nodes(2)` should both run the entire pipeline. Instead, what happens is that `from_nodes(1)` only runs node 1, and `to_nodes(2)` only runs node 2. This PR fixes this issue.

## How has this been tested?
A new test that wouldn't previously pass was created. All other tests still run.

## Checklist

### DIFF
--- a/tests/pipeline/test_pipeline_with_transcoding.py
+++ b/tests/pipeline/test_pipeline_with_transcoding.py
@@ -274,12 +274,13 @@ class TestComplexPipelineWithTranscoding:
 
     """
 
-    def test_from_to_nodes_transcoded_names(self, complex_pipeline):
-        """New pipelines contain all nodes that depend on node8, node7."""
+    # pylint: disable=too-many-public-methods
+
+    def test_from_nodes_transcoded_names(self, complex_pipeline):
+        """New pipelines contain all nodes that depend on node8 downstream."""
         from_nodes_pipeline = complex_pipeline.from_nodes("node8")
         nodes = {node.name for node in from_nodes_pipeline.nodes}
 
-        assert len(from_nodes_pipeline.nodes) == 9
         assert nodes == {
             "node1",
             "node2",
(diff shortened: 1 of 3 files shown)


tests/pipeline/test_pipeline_with_transcoding.py
```diff
@@ -274,6 +274,30 @@ class TestComplexPipelineWithTranscoding:
...
+            "node8",
+            "node10",
+        }
+
+        to_nodes_pipeline = complex_pipeline.to_nodes("node7")
+        nodes = {node.name for node in to_nodes_pipeline.nodes}
+
+        assert len(to_nodes_pipeline.nodes) == 3
+        assert nodes == {"node7", "node8", "node9"}
+
```
lorenabalan: I would break this down into 2 unit tests, for separation of concerns.

PedroAbreuQB: I actually tried that first but it would bump into pylint's limit of 20 public methods per class and I didn't want to override that limit.

lorenabalan: Ah I see, in that case you can stick a 
```
# pylint: disable=too-many-public-methods
```
We use that in `test_context` already, and it's just test code so pylint is being a bit dramatic. What do you think?

PedroAbreuQB:
> Great catch Pedro, thank you so much! Could you also add a note about this change in `RELEASE.md` and add your name to the contributor list for the next release (same file)?

Done in d120a17.

tests/pipeline/test_pipeline_with_transcoding.py
PedroAbreuQB: Agree, done in 37a381b.

PR #226, opened 2020-02-18
[KED-1425] Specify parameters in kedro run config through YAML dictionary

fdroessler:
## Description
This is a change for the behaviour of using `kedro run` with the `--config` or `-c` option and a config file. If that config file contains a section that either sets or overwrites `params`, the overwriting could only be done by providing the parameters as a string the same way it is used in the cli.
e.g. `kedro run --params foo:baz` in a config file would be something like
```
run:
  - params: 'foo:baz'
```
This would be fine for one or two parameters but ideally one would like to set them as a dict in the config file:
```
run:
  - params:
    - foo: baz
```

This PR adds that functionality as well as the relevant tests.

## Development notes
When using a config file the context uses the click.callback option to push all setting through to the click context and updates them. By checking if the value variable that arrives at the `_split_params` function is a dictionary, we can accept config file settings that use the standard yaml dictionary format.

https://github.com/quantumblacklabs/kedro/blob/25bf2dfc8af7b3fda7f5bb4e6c00f16b3768e2e9/kedro/template/%7B%7B%20cookiecutter.repo_name%20%7D%7D/kedro_cli.py#L125

This has been tested by extending the following function with relevant use-cases.

https://github.com/quantumblacklabs/kedro/blob/25bf2dfc8af7b3fda7f5bb4e6c00f16b3768e2e9/tests/template/test_kedro_cli.py#L208

## Checklist

~~-[ ] Updated the documentation to reflect the code changes~~

## Notice


kedro/template/{{ cookiecutter.repo_name }}/kedro_cli.py
```diff
@@ -122,7 +122,19 @@ def _split_string(ctx, param, value):
     return [item for item in value.split(",") if item]
 
 
+def _convert_text_to_numeric(value):
```
ZainPatelQB: Like that you've factored this out, but not a big fan of the name here, sounds like it expects the value to be a numeric-parseable string rather than only converting to a numeric type _if_ possible and keeping it as a string otherwise. That said, I don't have a better suggestion, so feel free to ignore me...

CONTRIBUTING.md
```diff
@@ -121,6 +121,8 @@ make install-test-requirements
 make install-pre-commit
 ```
 
+> **_NOTE:_** If Spark/PySpark/Hive tests for datasets are failing it might be due to the lack of Java>8 support from Spark. You can try using `export JAVA_HOME=$(/usr/libexec/java_home -v 1.8)` which works under MacOS or other workarounds. [Reference](https://stackoverflow.com/questions/53583199/pyspark-error-unsupported-class-file-major-version-55)
```
ZainPatelQB: I'd personally shorten this to:

```suggestion
> *Note*: If your Spark/PySpark/Hive tests for datasets are failing, make sure you have Java 8 installed. Spark doesn't currently support Java >8.
```

tests/template/test_kedro_cli.py
```diff
@@ -205,15 +205,33 @@ def test_run_successfully_parallel_via_name(
...
+        expected,
+        config_flag,
+        fake_kedro_cli,
+        fake_load_context,
+        fake_run_config,
+        mocker,
     ):
+        config = anyconfig.load(fake_run_config)
+        config["run"].update(config_addition)
+        anyconfig.dump(config, fake_run_config)
```
ZainPatelQB: Can you explain what this is for and what are the changes you've made to this test beyond adding in more cases in the parametrisation? :)

kedro/template/{{ cookiecutter.repo_name }}/kedro_cli.py
fdroessler: I think that is actually a good point. Maybe `_convert_str_type_to_numeric(value):` or even `_try_convert_str_type_to_numeric(value):`?

CONTRIBUTING.md
fdroessler: Agreed this would solve things. However, I actually didn't need to install Java8 but only set that flag. Keeping maybe the link or the flag could prevent some unnecessary digging around and installing things for some?
Maybe take your suggestion and add: Mac users can try the suggestion here: [Reference]()

tests/template/test_kedro_cli.py
fdroessler: What I wanted to ensure with the test is that no matter which way `params` is set in the config it would be tested properly. That makes it necessary to edit the config file that is written to disk by the test fixture. However, when I do that in the fixture I can't test the case where `params` is empty (I think) because I can't propagate the actual expected outcome to the test and it also breaks other tests. Therefore, I opted to read the config from disk and extend the section I wanted then write it back to disk. A bit ugly but I am open to other options :)

ZainPatelQB: Gotcha. Had a proper look at it this morning and makes sense to me. Although I'd probably have factored that out into a fixture (that depended on the `fake_run_config` fixture). Thoughts @DmitriiDeriabinQB?

kedro/template/{{ cookiecutter.repo_name }}/kedro_cli.py
ZainPatelQB: On second thought, maybe just keep the name and add a docstring?

tests/template/test_kedro_cli.py
```diff
@@ -224,6 +242,9 @@ def test_run_with_config(
             load_versions={},
             pipeline_name="pipeline1",
         )
+        fake_load_context.assert_called_once_with(
+            Path.cwd(), env=None, extra_params=expected
```
ZainPatelQB: We don't really care what it was called with `env=None` here, so maybe:

```suggestion
            Path.cwd(), env=mocker.ANY, extra_params=expected
```

CONTRIBUTING.md
ZainPatelQB: Ah, gotcha.Can you change `> **_NOTE:_**` to `> *Note:*` to be consistent with the other notes in the section? :)

kedro/template/{{ cookiecutter.repo_name }}/kedro_cli.py
```diff
@@ -122,7 +122,19 @@ def _split_string(ctx, param, value):
...
+    try:
+        value = float(value)
+    except ValueError:
+        pass
```
DmitriiDeriabinQB:
```suggestion
        return value
```

kedro/template/{{ cookiecutter.repo_name }}/kedro_cli.py
```diff
@@ -122,7 +122,19 @@ def _split_string(ctx, param, value):
...
+    else:
```
DmitriiDeriabinQB:
```suggestion
```

kedro/template/{{ cookiecutter.repo_name }}/kedro_cli.py
```diff
@@ -122,7 +122,19 @@ def _split_string(ctx, param, value):
...
+        value = int(value) if value.is_integer() else value
```
DmitriiDeriabinQB:
```suggestion
    return int(value) if value.is_integer() else value
```

kedro/template/{{ cookiecutter.repo_name }}/kedro_cli.py
```diff
@@ -122,7 +122,19 @@ def _split_string(ctx, param, value):
...
+    return value
```
DmitriiDeriabinQB:
```suggestion
```

kedro/template/{{ cookiecutter.repo_name }}/kedro_cli.py
DmitriiDeriabinQB: `_try_convert_to_numeric`? It technically also works for inputs that are not strings.

tests/template/test_kedro_cli.py
DmitriiDeriabinQB: Agree with @ZainPatelQB - making a new fixture out of it would improve the readability. I would also suggest to put the parameters logic into a separate test.

kedro/template/{{ cookiecutter.repo_name }}/kedro_cli.py
fdroessler: will go with @DmitriiDeriabinQB last suggestion.

CONTRIBUTING.md
fdroessler: Done.

tests/template/test_kedro_cli.py
fdroessler: Created a new fixture with the other one as dependency and the separated the tests. So that the original still tests the `-c` and `--config` and the new one tests the parameters. Learned something new with indirect fixtures here thanks for that :D hope that actually makes sense.

PR #276, opened 2020-03-05
[KED-1271] Add a Tensorflow Model Dataset with versioning

w0rdsm1th:
## Description
… and associated tests

used similar dummy models to TF's own test cases (e.g. 1 hidden layer, 3 nodes)

added TensorFlow2.X to requirements.txt. TF1 not supported because of differences in API and model serialization. TF2 is backwards compatible with TF1 saved models but would require h5py dependency to read.

no updates to docs but followed docstring conventions.

TensorFlowModelVersionedDataset tests had issues with Windows 260 character maximum path length limit. If use full datetime stamp as version name then the nested folder stucture often exceeds 260 chars.

Unsure how package maintainers would like to handle this error: e.g. try/catch in TensorFlowModelVersionedDataset load and save path resolution methods? Error occurs as TF tries to create the directory with model assets.

Typical Windows maxpath error details:
DataSetError due to
Failed to create a NewWriteableFile: C:\Users\aleks\AppData\Local\Temp\pytest-of-aleks\pytest-106\test_prevent_overwrite_None_sa0\test.tf\2020-03-05T12.34.36.535Z\test.tf\variables\variables_temp_d9ec32fefa50443f8ca2168b111b1a3f/part-00000-of-00001.data-00000-of-00001.tempstate15888019707169134298 : The system cannot find the path specified.

related to issues 
https://github.com/quantumblacklabs/kedro/issues/123
https://github.com/quantumblacklabs/kedro/issues/176

## Development notes
New classes TensorFlowModelDataset and TensorFlowModelVersionedDataset and associated tests.

some tests break on Windows due to 260 char maximum path length. Error is a DataSetError due to TF unable to complete model writing. tested them on Unix and they pass.

## Checklist

- [n/a] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [n/a] Updated the documentation to reflect the code changes

## Notice


kedro/io/tensorflow_model_data_set.py
```diff
@@ -0,0 +1,237 @@
...
+from typing import Any, Dict
+
+import fsspec
+import tensorflow as tf
+
+from kedro.io import AbstractDataSet, AbstractVersionedDataSet, Version
+from kedro.io.core import get_protocol_and_path
+
+
+class TensorFlowModelDataset(AbstractDataSet):
```
mzjp2: Can this go inside `kedro/extras/datasets/tensorflow` instead of `kedro.io` which is for core dataset functionality?

requirements.txt
```diff
@@ -15,6 +15,7 @@ requests>=2.20.0, <3.0
 s3fs>=0.3.0, <1.0  # Needs to be at least 0.3.0 to make use of `cachable` attribute on S3FileSystem
 SQLAlchemy>=1.2.0, <2.0
 tables>=3.4.4, <3.6  # 3.6.0 breaks on install, HDF5 library not found
+tensorflow>=2.0.0, <=2.1.0  # only V2 supported for saving and loading. V1 requires HDF5 and serializes differently
```
mzjp2: Can you put this in `setup.py` under `extras_require` under a `tensorflow` key so users don't need to have `tensorflow` installed when installing `kedro`, but can choose to install it with `pip install "kedro[tensorflow]"` if they want to use a `TensorflowDataSet`.

kedro/io/tensorflow_model_data_set.py
```diff
@@ -0,0 +1,237 @@
...
+    def _describe(self) -> Dict[str, Any]:
+        # todo(aleks) - could also expose model.summary()
+        return dict(
+            filepath=self._filepath,
+            load_args=self._load_args,
+            save_args=self._save_args,
+        )
+
+
+class TensorFlowModelVersionedDataset(AbstractVersionedDataSet):
```
mzjp2: This doesn't need to be a separate class, you should only have one `TenorFlowModelDataSet` that inherits from `AbstractVersionedDataset` (which itself inherits from `AbstractDataSet`). Then users can turn versioning on/off with the `versioned` flag in `catalog.yml`.

kedro/io/tensorflow_model_data_set.py
```diff
@@ -0,0 +1,237 @@
...
+                E.g. for ``GCSFileSystem`` class: `{project: 'my-project', ...}`
+        """
+        _fs_args = copy.deepcopy(fs_args) or {}
+        _credentials = copy.deepcopy(credentials) or {}
+        protocol, path = get_protocol_and_path(filepath, version)
+
+        self._protocol = protocol
+        self._fs = fsspec.filesystem(self._protocol, **_credentials, **_fs_args)
+        super().__init__(
+            PurePath(filepath), version, exists_function=lambda x: Path(x).is_dir(),
```
mzjp2: Is there a reason this isn't just `exists_function = self._fs.exists`?

kedro/io/tensorflow_model_data_set.py
```diff
@@ -0,0 +1,237 @@
...
+        load_path = Path(self._get_load_path())
+        return tf.keras.models.load_model(str(load_path), **self._load_args)
+
+    def _save(self, data: tf.keras.Model) -> None:
+        save_path = Path(self._get_save_path())
+        tf.keras.models.save_model(data, str(save_path), **self._save_args)
+
+    def _exists(self) -> bool:
+        path = Path(self._get_load_path())
+        return path.is_dir()
```
mzjp2: Why is this different to what, say, CSVDataSet does? https://github.com/quantumblacklabs/kedro/blob/develop/kedro/extras/datasets/pandas/csv_dataset.py

Will yours work with arbitrary paths like `s3://<...>`?

kedro/io/tensorflow_model_data_set.py
w0rdsm1th: the TF saved model is a directory of artefacts. could use `self._fs.isdir`?

see fsspec docs:
https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.spec.AbstractFileSystem.isdir

kedro/io/tensorflow_model_data_set.py
w0rdsm1th: it's not different, happy to take your guidance to enhance. I assume `_describe()` is for exposing useful user metadata but I don't know how its ingested elsewhere to display to user. I hadn't used kedro before the hackathon.

`_describe()` seems like nice place to expose TF summary model methods. would also be nice place to expose if its been fitted? output of summary() is a formatted chunk of text describing the model. think might just print to stdout as well. see example here https://stackoverflow.com/q/58657003

https://www.tensorflow.org/api_docs/python/tf/keras/Model#summary

will add a test case for a mocked S3. any other such arbitrary file systems that aiming to support/test?

kedro/io/tensorflow_model_data_set.py
w0rdsm1th: yep thats clear.

for own knowledge where is catalog.yml? assume its like a user config after you've setup a kedro workflow?

mzjp2: Yep, once you generate your kedro project with `kedro new`, the catalog can be found in `conf/base/catalog.yml` which is where users specify their datasets.

mzjp2:
Hi @w0rdsm1th, any update here? :)

kedro/io/tensorflow_model_data_set.py
w0rdsm1th: @mzjp2 your input about exposing the TF model summary methods here please?

w0rdsm1th:
hi @mzjp2 sorry fell off my radar. I will add the tests and update the PR this afternoon

kedro/io/tensorflow_model_data_set.py
mzjp2: @DmitriiDeriabinQB any input on this bit about adding TF stuff to `_describe`? I think that's more about a description of the data itself, whereas `_describe` is more for _metadata_ of the dataset.

DmitriiDeriabinQB: As `fsspec` is still used here, I agree that `_describe()` should be identical to, say, `CSVDataSet`.
Also FYI, `_get_load_path()` will raise `VersionNotFoundError` if versioning is enabled and there are no versions available, hence try/except in `CSVDataSet._describe()`.

DmitriiDeriabinQB: > @DmitriiDeriabinQB any input on this bit about adding TF stuff to `_describe`? I think that's more about a description of the data itself, whereas `_describe` is more for _metadata_ of the dataset.

Is it still relevant as I don't see anything TF specific in both `_describe()` methods?

w0rdsm1th: No I haven't put anything TF specific in there yet. I wanted to ask about best place to expose before putting it into code. Seems like it would be useful information to display to the user in the interface. Although Model summary's can get arbitrarily long for each layer in model.

I am not really a Kedro user so very much keen for input about expected behaviour and functionality. Started contributing through pysprints hackathon

w0rdsm1th:
> hi @mzjp2 sorry fell off my radar. I will add the tests and update the PR this afternoon

*this weekend. Started writing the additional tests and noticed a couple changes to class hierarchy and the "nice" test cases ;)

kedro/extras/datasets/tensorflow/tensorflow_model_data_set.py
```diff
@@ -0,0 +1,166 @@
...
+        # Handle default load and save arguments
+        self._load_args = copy.deepcopy(self.DEFAULT_LOAD_ARGS)
+        if load_args is not None:
+            self._load_args.update(load_args)
+        self._save_args = copy.deepcopy(self.DEFAULT_SAVE_ARGS)
+        if save_args is not None:
+            self._save_args.update(save_args)
+
+    def _load(self) -> None:
+        load_path = Path(self._get_load_path())
```
mzjp2, line 140:
```suggestion
        load_path = get_filepath_str(self._get_load_path(), self._protocol)
```

kedro/extras/datasets/tensorflow/tensorflow_model_data_set.py
```diff
@@ -0,0 +1,166 @@
...
+        return tf.keras.models.load_model(str(load_path), **self._load_args)
```
mzjp2, line 141: I'm a little stuck here, because we should be opening `load_path` using `fsspec` and then passing the file descriptor to `load_model` so that users can use any filesystem they want and aren't tied to local, but it looks like `load_model` only wants a string or an `h5py.File` object? Is there anything we can do here?

kedro/extras/datasets/tensorflow/tensorflow_model_data_set.py
```diff
@@ -0,0 +1,166 @@
...
+            version: If specified, should be an instance of
+                ``kedro.io.core.Version``. If its ``load`` attribute is
+                None, the latest version will be loaded. If its ``save``
+                attribute is None, save version will be autogenerated.
+            credentials: Credentials required to get access to the underlying filesystem.
+                E.g. for ``GCSFileSystem`` it should look like `{'token': None}`.
+            fs_args: Extra arguments to pass into underlying filesystem class.
+                E.g. for ``GCSFileSystem`` class: `{project: 'my-project', ...}`
+        """
+        _fs_args = copy.deepcopy(fs_args) or {}
```
mzjp2, line 115:
```suggestion
        _fs_args = copy.deepcopy(fs_args) or {}
        _fs_open_args_load = _fs_args.pop("open_args_load", {})
        _fs_open_args_save = _fs_args.pop("open_args_save", {})
```

Sorry for changing this out under you, but this is the new format we're exposing for letting users specify load and save args. :)

kedro/extras/datasets/tensorflow/tensorflow_model_data_set.py
```diff
@@ -0,0 +1,166 @@
...
+
+        self._data = None
+
+        # Handle default load and save arguments
+        self._load_args = copy.deepcopy(self.DEFAULT_LOAD_ARGS)
+        if load_args is not None:
+            self._load_args.update(load_args)
+        self._save_args = copy.deepcopy(self.DEFAULT_SAVE_ARGS)
+        if save_args is not None:
+            self._save_args.update(save_args)
```
mzjp2, line 137:
```suggestion
            self._save_args.update(save_args)
            
        _fs_open_args_load.setdefault("mode", "r")
        _fs_open_args_save.setdefault("mode", "w")
        self._fs_open_args_load = _fs_open_args_load
        self._fs_open_args_save = _fs_open_args_save
```

setup.py
```diff
@@ -105,6 +105,10 @@
         "tables>=3.6",
     ],
     "spark": ["pyspark>=2.2.0, <3.0", "hdfs>=2.5.8, <3.0", "s3fs>=0.3.0, <0.4.1"],
+    # only TensorFlow V2 supported for saving and loading. V1 requires HDF5 and serializes differently
+    "tensorflow": [
+        "tensorflow>=2.0.0, <=2.1.0",
```
mzjp2, line 110: Can you upper bound `3.0` instead since that'll be a breaking change version and 2.1.0 should be a non-breaking minor release (or does TF not use SemVer?)?

tests/extras/datasets/tensorflow/test_tensorflow_model_data_set.py
```diff
@@ -0,0 +1,355 @@
...
+
+
+@pytest.fixture
+def dummy_tf_subclassed_model(dummy_x_train, dummy_y_train):
+    """
+    demonstrate that own class models cannot be saved using HDF5 format but can using TF format
+    :param dummy_x_train:
+    :param dummy_y_train:
+    :return:
+    """
```
mzjp2, line 131: We use Google docstrings instead, can you switch to those? :) It's a test, so you can just do
```suggestion
    """Demonstrate that own class models cannot be saved using HDF5 format but can using TF format"""
```

or 

```suggestion
    """Demonstrate that own class models cannot be saved
    using HDF5 format but can using TF format
    """
```

if it complains about length of the line. :)

kedro/extras/datasets/tensorflow/tensorflow_model_data_set.py
```diff
@@ -0,0 +1,166 @@
...
+        return self._fs.exists(load_path)
+
+    def _describe(self) -> Dict[str, Any]:
+        # todo(aleks) - could also expose model.summary()
+        return dict(
+            filepath=str(self._filepath),
+            load_args=self._load_args,
+            save_args=self._save_args,
+            version=self._version,
+        )
```
mzjp2, line 166:
```suggestion
        return dict(
            filepath=str(self._filepath),
            load_args=self._load_args,
            save_args=self._save_args,
            version=self._version,
            layer=self._layer,
        )
```

kedro/extras/datasets/tensorflow/tensorflow_model_data_set.py
```diff
@@ -0,0 +1,166 @@
...
+    )  # type: Dict[str, Any]
+
+    def __init__(
+            self,
+            filepath: str,
+            load_args: Dict[str, Any] = None,
+            save_args: Dict[str, Any] = None,
+            version: Version = None,
+            credentials: Dict[str, Any] = None,
+            fs_args: Dict[str, Any] = None,
```
mzjp2, line 86:
```suggestion
            fs_args: Dict[str, Any] = None,
            layer: str = None,
```

kedro/extras/datasets/tensorflow/tensorflow_model_data_set.py
```diff
@@ -0,0 +1,166 @@
...
+                https://www.tensorflow.org/api_docs/python/tf/keras/models/save_model
+                All defaults are preserved, except for "index", which is set to False.
+                In TF this defaults to 'tf' in TF 2.X, and 'h5' in TF 1.X.
+            version: If specified, should be an instance of
+                ``kedro.io.core.Version``. If its ``load`` attribute is
+                None, the latest version will be loaded. If its ``save``
+                attribute is None, save version will be autogenerated.
+            credentials: Credentials required to get access to the underlying filesystem.
+                E.g. for ``GCSFileSystem`` it should look like `{'token': None}`.
+            fs_args: Extra arguments to pass into underlying filesystem class.
```
mzjp2, line 112:
```suggestion
            fs_args: Extra arguments to pass into underlying filesystem class.
            layer: The data layer according to the data engineering convention:
                https://kedro.readthedocs.io/en/stable/06_resources/01_faq.html#what-is-data-engineering-convention
```

kedro/extras/datasets/tensorflow/tensorflow_model_data_set.py
```diff
@@ -0,0 +1,166 @@
...
+        self._fs = fsspec.filesystem(self._protocol, **_credentials, **_fs_args)
+        super().__init__(
+            PurePath(filepath),
+            version,
+            # exists_function=lambda x: Path(x).is_dir(),
+            # exists_function=self._fs.isdir,
+            exists_function=self._fs.exists,
+        )
+
+        self._data = None
```
mzjp2, line 129:
```suggestion
```

I don't think this is needed and/or used?

kedro/extras/datasets/tensorflow/tensorflow_model_data_set.py
```diff
@@ -0,0 +1,166 @@
...
+
+        self._protocol = protocol
+        self._fs = fsspec.filesystem(self._protocol, **_credentials, **_fs_args)
+        super().__init__(
+            PurePath(filepath),
+            version,
+            # exists_function=lambda x: Path(x).is_dir(),
+            # exists_function=self._fs.isdir,
+            exists_function=self._fs.exists,
+        )
```
mzjp2, line 127:
```suggestion
        super().__init__(
            PurePath(filepath),
            version,
            # exists_function=lambda x: Path(x).is_dir(),
            # exists_function=self._fs.isdir,
            exists_function=self._fs.exists,
        )
        
        self._layer = layer
```

kedro/extras/datasets/tensorflow/tensorflow_model_data_set.py
```diff
@@ -0,0 +1,166 @@
...
+        if save_args is not None:
+            self._save_args.update(save_args)
+
+    def _load(self) -> None:
+        load_path = Path(self._get_load_path())
+        return tf.keras.models.load_model(str(load_path), **self._load_args)
+
+    def _save(self, data: tf.keras.Model) -> None:
+        save_path = Path(self._get_save_path())
+        tf.keras.models.save_model(data, str(save_path), **self._save_args)
```
mzjp2, line 145: Same comment here re: fsspec

kedro/extras/datasets/tensorflow/tensorflow_model_data_set.py
```diff
@@ -0,0 +1,166 @@
...
+
+    # original commit
+    # def _exists(self) -> bool:
+    #     path = Path(self._get_load_path())
+    #     return path.is_dir()
```
mzjp2, line 150: You can delete this. :)

kedro/extras/datasets/tensorflow/tensorflow_model_data_set.py
w0rdsm1th: cleaner, nice

setup.py
w0rdsm1th: good point, they do

w0rdsm1th:
@mzjp2 thanks for the detailed feedback! I had planned on doing a bit more work on this. The arbitrary paths test cases weren't passing in this class or in the "reference case" CSV Dataset tests on my Windows machine. In particular test case tests.extras.datasets.tensorflow.test_tensorflow_model_data_set.TestTensorFlowModelDataset.test_protocol_usage. It seemed like an OS path-issue and I then committed to test on unix but they weren't passing for same reason there either!

AssertionError: assert 'gcs:\\bucket\\test_tf' == 'bucket/test_tf'

re: fsspec passing to keras load_model and save_model. I'm not familliar with fsspec. Definitely doesn't seem like there's a quick easy fix, will have to have a look!

w0rdsm1th:
@mzjp2 re: fsspec context manager returning an OpenFile object... it certainly is a brainer! I can totally see why you want to use fsspec throughout but its flexibility just doesn't seem to interface with TFs public load/save methods that expect pathlike argument.

I found a couple SO answers below from TF and AWS contributors that suggest downloading the file locally with stdlib `tempfile`. And we would then pass the tempfile path to the load_model method. Seems hacky but the only way forward! 

If this approach is acceptable to library design/road map then I am happy to write a method to save to/load from a tempfile and then pass the tempfile path to TF save/load methods.

Couple questions to support:
- Are you happy using the OS default tempfile directory (this the default behaviour for `tempfile`)? Or would you prefer to save to/load from a nominated local "kedro workspace" location?
- This seems like it will be useful functionality to make available to other extensions. Could be a single method in kedro/io/core.py? Or would you prefer to put it somewhere else?

https://docs.python.org/3.6/library/tempfile.html#tempfile.TemporaryDirectory

https://stackoverflow.com/a/47902672/3596968
https://stackoverflow.com/a/55961847/3596968

mzjp2:
> @mzjp2 re: fsspec context manager returning an OpenFile object... it certainly is a brainer! I can totally see why you want to use fsspec throughout but its flexibility just doesn't seem to interface with TFs public load/save methods that expect pathlike argument.
> 
> I found a couple SO answers below from TF and AWS contributors that suggest downloading the file locally with stdlib `tempfile`. And we would then pass the tempfile path to the load_model method. Seems hacky but the only way forward!
> 
> If this approach is acceptable to library design/road map then I am happy to write a method to save to/load from a tempfile and then pass the tempfile path to TF save/load methods.
> 
> Couple questions to support:
> 
> * Are you happy using the OS default tempfile directory (this the default behaviour for `tempfile`)? Or would you prefer to save to/load from a nominated local "kedro workspace" location?
> * This seems like it will be useful functionality to make available to other extensions. Could be a single method in kedro/io/core.py? Or would you prefer to put it somewhere else?
> 
> https://docs.python.org/3.6/library/tempfile.html#tempfile.TemporaryDirectory
> 
> https://stackoverflow.com/a/47902672/3596968
> https://stackoverflow.com/a/55961847/3596968

Thanks for following up on this @w0rdsm1th, super appreciated. We're happy to proceed with the `tmpfile` workaround, but will need to be careful with things. I'm cc-ing in @DmitriiDeriabinQB and @idanov to help answer the questions here.

If it's something you'd rather us do instead of you (I appreciate you've sunk a lot of work into this already), we're happy to take it over and continue (with full credit to you, of course).

idanov:
@w0rdsm1th Thank you very much for your amazing work for this PR! You've done a great job and we really appreciate your effort!

From your research it seems that the `tempfile` is our only option for now and that's how we should proceed forward. I would suggest we restrain from making the creation of the tempfile a method of the general class since not all datasets will be filebased as well as adding tempfile helper there would encourage development of more datasets based on the tempfile, which is a suboptimal solution - it's not uncommon that Kedro pipelines are run on systems with limited disk storage, e.g. EC2 instances, and filling up the disk space by accident can be a source of strange and hard to debug problems for the pipeline developers.

w0rdsm1th:
thanks for the input guys! I have updated the TensorFlowModelDataset._load() and TensorFlowModelDataset._save() methods with tempfile.TemporaryDirectory intermediary functionality.

Separate point to the tempfile intermediary: TestTensorFlowModelDataset::()::test_protocol_usage() tests all fail for me on Windows or Unix but I copied them directly from TestCSVDataSet::()::test_protocol_usage which seems like a reference test case so reluctant to change my own TF test cases. They fail because of the path manipulation and protocol stripping within the test case.

kedro/extras/datasets/tensorflow/tensorflow_model_dataset.py
```diff
@@ -0,0 +1,192 @@
...
+            self._load_args.update(load_args)
+        self._save_args = copy.deepcopy(self.DEFAULT_SAVE_ARGS)
+        if save_args is not None:
+            self._save_args.update(save_args)
+
+        self._tmpfile_callable = (
+            tempfile.NamedTemporaryFile
+            if self._save_args.get("save_format") == "h5"
+            else tempfile.TemporaryDirectory
+        )  # type: Callable
```
andrii-ivaniuk, line 136: `mypy` complains here:
`Incompatible types in assignment (expression has type "object", variable has type "Callable[..., Any]")`

kedro/extras/datasets/tensorflow/README.md
```diff
@@ -0,0 +1,27 @@
...
+import tensorflow as tf
+import numpy as np
+data_set = TensorFlowModelDataset("tf_model_dirname")
+model = tf.keras.Model()
+predictions = model.predict([...])
+data_set.save(model)
+loaded_model = data_set.load()
+new_predictions = loaded_model.predict([...])
+np.testing.assert_allclose(predictions, new_predictions, rtol=1e-6, atol=1e-6)
+```
```
mzjp2, line 18: can you black this/lint this? :)

kedro/extras/datasets/tensorflow/tensorflow_model_dataset.py
```diff
@@ -0,0 +1,191 @@
...
+                https://www.tensorflow.org/api_docs/python/tf/keras/models/save_model
+                All defaults are preserved, except for "save_format", which is set to "tf".
+            version: If specified, should be an instance of
+                ``kedro.io.core.Version``. If its ``load`` attribute is
+                None, the latest version will be loaded. If its ``save``
+                attribute is None, save version will be autogenerated.
+            credentials: Credentials required to get access to the underlying filesystem.
+                E.g. for ``GCSFileSystem`` it should look like `{'token': None}`.
+            fs_args: Extra arguments to pass into underlying filesystem class constructor
+                (e.g. `{"project": "my-project"}` for ``GCSFileSystem``).
```
mzjp2, line 107: i think you need `open_args_load` and stuff here :)

andrii-ivaniuk: `open_args_load` as well as `open_args_save` are not required since we don't use `self._fs.open()` here.

mzjp2: ah of course 🤦‍♂️ my bad

kedro/extras/datasets/tensorflow/tensorflow_model_dataset.py
```diff
@@ -109,6 +109,8 @@ def __init__(
         _fs_args = copy.deepcopy(fs_args) or {}
         _credentials = copy.deepcopy(credentials) or {}
         protocol, path = get_protocol_and_path(filepath, version)
+        if protocol == "file":
+            _fs_args["auto_mkdir"] = True
```
mzjp2, line 113: this will override the users choice, right?
```suggestion
            _fs_args.setdefault("auto_mkdir", True)
```

andrii-ivaniuk: Yes, it will override but if it is set to False the exception will be raised if the target path doesn't exist. What is the use case for such behavior? In the previous `fsspec` version it was set to True by default. 

Note: it is related more to the versioned dataset because it requires a versioned path which doesn't exist (e.g. path `/tmp/test.csv/2019-01-02T00.00.00.000Z/`).

mzjp2: yeah so i think to avoid making this a breaking change, we should do the same thing and "set to True by default" -- not "enforce True"

andrii-ivaniuk: Why is this a breaking change? It was set to True in the previous version.
However my main question: what is the use case to set it to `False`. If there is any then I will update it.

Moreover, if you take a look at the previous implementation of e.g. `CSVLocalDataSet` you can see `save_path.parent.mkdir(parents=True, exist_ok=True)`. So we didn't allow users to not create a target path.

mzjp2: Was it? From my understanding, we didn't set `auto_mkdir` at all in the previous version and relied on `fsspec` defaulting it to `True`. But if users specified `fs_args: auto_mkdir: param` in their `catalog`, we passed `param` onwards to `fsspec`. This no longer does that -- which makes it a breaking change, no?

I agree that there's pretty much no use case for setting `auto_mkdir` to `False` within Kedro. Maybe we should scope out the `fsspec` bump to a separate PR, instead of doing it here

andrii-ivaniuk: I believe there is no any use case to `auto_mkdir` to `False` in the scope of Kedro and overriding the user's value (if he/she will ever provide it) will prevent from raising an error.

docs/source/05_api_docs/kedro.extras.datasets.rst
```diff
@@ -35,3 +35,4 @@ Data Sets
     kedro.extras.datasets.spark.SparkJDBCDataSet
     kedro.extras.datasets.text.TextDataSet
     kedro.extras.datasets.yaml.YAMLDataSet
+    kedro.extras.datasets.tensorflow.TensorFlowModelDataset
```
mzjp2, line 38: i believe we order this alphabetically, so can you put it between `spark` and `text`?

kedro/extras/datasets/tensorflow/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+from kedro.extras.datasets.tensorflow import TensorFlowModelDataset
+
+data_set = TensorFlowModelDataset("tf_model_dirname")
+model = tf.keras.Model()
+predictions = model.predict([...])
+data_set.save(model)
+loaded_model = data_set.load()
+new_predictions = loaded_model.predict([...])
+np.testing.assert_allclose(predictions, new_predictions, rtol=1e-6, atol=1e-6)
```
mzjp2, line 19: i would put more blank lines in between the different portions (data / save / load / data) but not a big deal

RELEASE.md
```diff
@@ -3,9 +3,11 @@
 ## Major features and improvements
```
mzjp2, line 3: should add @w0rdsm1th as a contributor in the `RELEASE.md` notes

kedro/extras/datasets/tensorflow/tensorflow_model_dataset.py
```diff
@@ -0,0 +1,193 @@
...
+        >>> import numpy as np
+        >>>
+        >>> data_set = TensorFlowModelDataset("saved_model_path")
+        >>> model = tf.keras.Model()
+        >>> predictions = model.predict([...])
+        >>>
+        >>> data_set.save(model)
+        >>> loaded_model = data_set.load()
+        >>> new_predictions = loaded_model.predict([...])
+        >>> np.testing.assert_allclose(predictions, new_predictions, rtol=1e-6, atol=1e-6)
```
mzjp2, line 68: this is how i would structure the other example code in the `README` :)

PR #296, opened 2020-03-18
[KED-1492] [WIP] Add dry runner

AntonyMilneQB:
## Description
There is a variety of ways in which a user can specify the set of nodes to execute (e.g. through pipeline, tags, node names, from node, to node, etc.). Executing nodes can take a long time, and so it is useful to be able to preview which nodes will be selected without actually running them. This PR creates a `DryRunner` which may be used as an alternative to the existing `SequentialRunner` and `ParallelRunner`.

## Development notes
The runner can be called through CLI by `kedro run --runner=DryRunner`. There is no `--dry` flag, since apparently `--parallel` will also be removed at some point in the relatively near future.

Example output:
```
test_sequential_runner.py F2020-03-18 14:25:30,420 - kedro.runner.dry_runner - INFO - Wet run would execute 5 nodes:
node5: random(None) -> [A]
node3: identity([A]) -> [B]
node4: identity([B]) -> [C]
node2: identity([C]) -> [D]
node1: identity([D]) -> [E]
2020-03-18 14:25:30,421 - kedro.runner.dry_runner - INFO - Pipeline execution completed successfully.
2020-03-18 14:25:30,422 - kedro.io.data_catalog - INFO - Loading data from `E` (MemoryDataSet)...
```

Testing:
- new test `test_run_successfully_dry_via_name` in `test_kedro_cli`
- new test `test_dry_run_arg` in `test_context`

Open questions/things to do/stuff I'd like help with:
- [ ] are these CLI flags/options correct?
- [ ] documentation: what should I write and where?
- [ ] manual testing: I haven't actually tried using this yet, outside the tests...
- [ ] `test_dry_run_arg` in `test_context` doesn't work and I'm not sure why...
- [ ] since `DryRunner` inherits from `AbstractRunner`, when it is used the `run` method runs two things after `self._run(pipeline, catalog)`: 1. output the message "Pipeline execution completed successfully"; 2. `return {ds_name: catalog.load(ds_name) for ds_name in free_outputs}` (hence the "Loading data" message in the above example). These don't really make sense for `DryRunner`. Should we change the implementation so these two lines of code don't run, and if so how?
- [ ] not sure what, if any, tests are needed beyond those given. I started writing some tests based on `test_sequential_runner` but decided they weren't actually testing the new functionality in any meaningful way. Something that uses `caplog` to make sure it's counted the number of nodes correctly? Seems kind of pointless...

## Checklist

## Notice

### DIFF
--- a/kedro/context/context.py
+++ b/kedro/context/context.py
@@ -459,6 +459,7 @@ class KedroContext(abc.ABC):
 
         save_version = self._get_save_version()
         run_id = self.run_id or save_version
+        runner = runner or SequentialRunner()
 
         record_data = {
             "run_id": run_id,
@@ -473,6 +474,7 @@ class KedroContext(abc.ABC):
             "load_versions": load_versions,
             "pipeline_name": pipeline_name,
             "extra_params": self._extra_params,
+            "runner": type(runner).__name__,
         }
         journal = Journal(record_data)
 
@@ -481,7 +483,6 @@ class KedroContext(abc.ABC):
         )
 
         # Run the runner
-        runner = runner or SequentialRunner()
         return runner.run(filtered_pipeline, catalog)
 
     def _get_run_id(
--- a/kedro/runner/dry_runner.py
+++ b/kedro/runner/dry_runner.py
@@ -25,12 +25,43 @@
 #
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-"""``kedro.runner`` provides runners that are able
-to execute ``Pipeline`` instances.
+"""``DryRunner`` is an ``AbstractRunner`` implementation. It can be used to list which
+nodes would be run without actually executing anything.
 """
 
-from .dry_runner import DryRunner  # NOQA
-from .parallel_runner import ParallelRunner  # NOQA
-from .runner import AbstractRunner, run_node  # NOQA
-from .sequential_runner import SequentialRunner  # NOQA
+from kedro.io import AbstractDataSet, DataCatalog, MemoryDataSet
+from kedro.pipeline import Pipeline
+from kedro.runner.runner import AbstractRunner
+
+
+class DryRunner(AbstractRunner):
+    """``DryRunner`` is an ``AbstractRunner`` implementation. It can be used to list which
+    nodes would be run without actually executing anything.
+    """
+
+    def create_default_data_set(self, ds_name: str) -> AbstractDataSet:
+        """Factory method for creating the default data set for the runner.
+
+        Args:
+            ds_name: Name of the missing data set
+
+        Returns:
+            An instance of an implementation of AbstractDataSet to be used
+            for all unregistered data sets.
+
+        """
+        return MemoryDataSet()
+
+    def _run(self, pipeline: Pipeline, catalog: DataCatalog) -> None:
+        """The method implementing dry pipeline running.
+
+        Args:
+            pipeline: The ``Pipeline`` to run.
+            catalog: The ``DataCatalog`` from which to fetch data.
+        """
+        nodes = pipeline.nodes
+        self._logger.info(
+            "Wet run would execute %d nodes:\n%s",
+            len(nodes),
+            "\n".join(map(str, nodes)),
+        )
--- a/tests/context/test_context.py
+++ b/tests/context/test_context.py
@@ -44,7 +44,7 @@ from kedro.context import KedroContext, KedroContextError
 from kedro.extras.datasets.pandas import CSVDataSet
 from kedro.io.core import Version, generate_timestamp
 from kedro.pipeline import Pipeline, node
-from kedro.runner import ParallelRunner, SequentialRunner
+from kedro.runner import DryRunner, ParallelRunner, SequentialRunner
 
 
 def _get_local_logging_config():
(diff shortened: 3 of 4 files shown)


921kiyo:
Hi @AntonyMilneQB thank you for the draft PR, it's a great idea :) I would like to know more about which outputs you are most interested in (is it the final output of the pipeline, or all the dependencies, or something else?). I'm trying to think about what's the best way to return the outputs to the users. While example output above is informative in CLI, I am thinking of the case when your pipeline is very large so the CLI outputs can be very long and complicated. Kedro-Viz (demo example https://quantumblacklabs.github.io/kedro-viz/) highlights the dependencies between nodes in your big pipeline very well without actually running the pipeline, if you hover over any of the upstream node or data like below. 

Screenshot 2020-03-20 at 09 13 59

It would be great if you could describe more about your use case of the dry runner :)

AntonyMilneQB:
Hey @921kiyo, thanks very much for your thoughts! I think the output that I'm really interested in here is just "which nodes would run if I execute this `kedro run` command?". It might also be nice to say "which datasets would be overwritten if I do this run?" and maybe "which datasets will be consumed if I do this run?", but I think those are less important. For now I decided to just list the `str` representation of each node that would be run, since I think that's clear and simple, but we could definitely provide more information if deemed useful.

I love kedro-viz and use it all the time for this sort of thing, but I don't think it fully serves this use case for various reasons. Let's say I have a big, complicated pipeline and want to run `kedro run --pipeline=p --from-node=n1 --to-node=n2 --tags=t`.
1. AFAIK there's no way in kedro viz to exactly replicate this to highlight which nodes would be run by this command (I can do `kedro viz -pipeline` and then select tags and separately do the from-node and to-node but that's not exactly the same)
1. Even if kedro viz could exactly replicate this selection of nodes (I gather that in future it might be more closely aligned to kedro run), I'd like a quick and easy way to be able to do this from the command line without the dependency on opening up a visualisation, manually selecting nodes, etc. With the CLI option, I can use the dry runner to work out what nodes would run, then just remove `runner=dry` and run the command straightaway without having to refer to a big visualisation.
1. Maybe I just don't want to use kedro viz (e.g. I have a huge, unwieldy graph, I can't run it for some reason, etc.)

@vrtsig seemed to like the idea too, so he might have some more thoughts about exactly what he'd like to see in the output or what he thinks the use case would be :)

vrtsig:
awesome, @AntonyMilneQB . I really like where this is going.

Personally, I'd be happy for the output to be a bit more verbose, e.g.
```
Starting dry run with the following parameters: [...]. Total nodes: 3

Step 1 / 3:
    node: my_pkg.nodes.random
    inputs:
        - None
    outputs:
        - A (MemoryDataSet)

Step 2 / 3:
    node: my_pkg.nodes.identity
    inputs:
        - A (MemoryDataSet)
    outputs:
        - B (CSVLocalDataSet)

Step 3 / 3:
    node: my_pkg.nodes.identity
    inputs:
        - B (CSVLocalDataSet)
    outputs:
        - C (MemoryDataSet)

=====================================

data sources consumed:
    - A
    - B

data sources written:
    - B

data sources returned:
    - C
```

I think Antony's motivation is really well phrased. Yes, this is not the ideal tool to check out a whole pipeline. But it should be much faster and less abstract than kedro viz, and its outputs can be consumed and processed by a machine (we could push this aspect even more by e.g. producing sth in yaml syntax). Also, there are situations (containerized development, cloud machines, etc) where kedro viz is more complicated to set up.

Most importantly though, as Antony points out, this lets you explore the effects of different command line calls which can be difficult to get right.

Big fan of this :-)

PS: pleasepleaseplease remove the phrase `wet run`. 🤢

AntonyMilneQB:
Thanks for the comments Fabian, much appreciated!

Just to clarify, the current format of the output `node3: identity([A]) -> [B]` is just what happens when you do `str(node)`, i.e. it's kedro's built in string representation of a node, given by:
```
    def __str__(self):
        def _sorted_set_to_str(xset):
            return "[{}]".format(",".join(sorted(xset)))

        out_str = _sorted_set_to_str(self.outputs) if self._outputs else "None"
        in_str = _sorted_set_to_str(self.inputs) if self._inputs else "None"

        prefix = self._name + ": " if self._name else ""
        return prefix + "{}({}) -> {}".format(self._func_name, in_str, out_str)
```

We could definitely change this to something more detailed, like in your example, although personally I think I quite like just having a short list of node names without all the details (I originally just had node names and not the datasets at all actually). I guess if you have explicit output you can always shorten it through `kedro -runner=dry | grep node`, while you can't go in the other direction to make a short output more verbose. Or we could allow a more and a less verbose option to be set using some flag? I think that is probably overcomplicating things. Just thinking out loud here anyway...

> PS: pleasepleaseplease remove the phrase `wet run`. 🤢

Haha, I was wondering whether people would object to this... I am definitely happy to change this.

limdauto:
@AntonyMilneQB hey, I have extracted your code and included it into documentation here instead: https://github.com/quantumblacklabs/private-kedro/pull/515 -- along with instruction on how to use it in `kedro run`. Would you be able to take a look? Massive thanks for figuring this out.

921kiyo:
Addressed this in the documentation in https://github.com/quantumblacklabs/kedro/commit/e71222cf81dde3b6926b0b667cd48a1050dc306c Thank you for your contribution :)

PR #357, opened 2020-05-07
[KED-1649] Add a decorator to validate node input arguments types and properties

### LINKED ISSUE #356: [KED-1649] Add a decorator for validating inputs other than dataframes
## Description
a decorator to validate types and expected properties of inputs other than dataframes (I leave the validation of dataframe expectations to Great Expectations).

## Context
It is very common to pass parameters to a node via a dictionary as in `params` or `params:train_params`. Usually, you would have to validate (inside the node function) that the dictionary has the fields you expect and that its values have the properties you expect. However, a decorator and the specification of a model for the inputs would be a more elegant way to address this scenario.

## Possible Implementation
wrap `validate_arguments` (from pydantic) so that the decorator takes as argument the inputs variables to validate.

juan-carlos-calvo:
## Description
Resolves #356 

## Development notes
added `validate_inputs.py` to `kedro.extras.decorators` and `test_validate_inputs.py` to `tests.extras.decorators`.

it was tested by `make test`.

## Checklist

## Notice


.pre-commit-config.yaml
```diff
@@ -95,11 +95,11 @@ repos:
             language: system
             pass_filenames: false
             entry: lint-imports
-          - id: secret_scan
-            name: "Secret scan"
-            language: system
-            pass_filenames: false
-            entry: make secret-scan
```
juan-carlos-calvo, line 102: I kept failing this hook even though when I ran `make secret-scan`  manually it passed.

test_requirements.txt
```diff
@@ -25,6 +25,7 @@ Pillow>=7.1.2, <7.2
 pre-commit>=1.17.0, <2.0.0
 psutil==5.6.6
 pyarrow>=0.12.0, <1.0.0
+pydantic
```
deepyaman, line 28: Can you bound this?

kedro/extras/decorators/validate_inputs.py
```diff
@@ -0,0 +1,71 @@
...
+#     or use the QuantumBlack Trademarks in any other manner that might cause
+# confusion in the marketplace, including but not limited to in advertising,
+# on websites, or on software.
+#
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+This module contains the validate decorator, which can be used as
+``Node`` decorators to validate node inputs of choice. See ``kedro.pipeline.node.decorate``
```
deepyaman, line 31:
```suggestion
This module contains the `validate` decorator, which can be used as a
``Node`` decorator to validate node inputs of choice. See ``kedro.pipeline.node.decorate``
```

test_requirements.txt
juan-carlos-calvo: done!

.pre-commit-config.yaml
lorenabalan: You can skip the precommit hooks by running `git commit -n` so you don't have to comment out or remove the source of pain, and it will fail on CI if that's still unhappy. That way at least we rule out whether it's a problem with the local environment or the actual dependency/code.

test_requirements.txt
```diff
@@ -25,6 +25,7 @@ Pillow>=7.1.2, <7.2
 pre-commit>=1.17.0, <2.0.0
 psutil==5.6.6
 pyarrow>=0.12.0, <1.0.0
+pydantic>=1.5, <1.6.0
```
lorenabalan, line 28: This should also be included in `setup.py` under extras require, so users can do sth like `pip install kedro[validation]` and easily install the dependency they need for the decorator to work. Also worth adding a note about that installation step in the docstrings.

kedro/extras/decorators/validate_inputs.py
```diff
@@ -0,0 +1,71 @@
...
+from pydantic import validate_arguments
+
+RT = TypeVar("RT")
+
+
+def validate(*arguments: Any) -> Callable[[Callable[..., RT]], Callable[..., RT]]:
+    """Creates a decorator for validating function inputs according to their types.
+    For a list of supported types please refer to:
+    https://pydantic-docs.helpmanual.io/usage/types/
+    for types which inherit from `pydantic.BaseModel` it evaluates the functions
```
lorenabalan, line 46:
```suggestion
    For types which inherit from `pydantic.BaseModel` it evaluates the functions
```

.pre-commit-config.yaml
juan-carlos-calvo: I think the problem was that I committed some comments to ignore certain pylint warnings using the warning code which the scan must have confused with secrets.

test_requirements.txt
juan-carlos-calvo: thanks, will do

juan-carlos-calvo:
> Thank you for creating this PR! Could you please also add a note in the README in `kedro.extras.decorators`, under "## What decorators are currently supported?"?
> I'm not very familiar with `pydantic` so this is a more generic question - How does this behave if the user tries to decorate a node function that hasn't got any type hinting? Can we add a test for that?

I'll add a test. It shouldn't do anything.

limdauto:
Hi @juan-carlos-calvo, thank you for this Pull Request. Just to play devil advocate: would it be better for this to live as a separate Python library, e.g. `kedro-pydantic-validation`? You could then allow users to use your decorator by importing `from kedro_pydantic_validation.decorators import validate`. You can also provide a [Hooks](https://kedro.readthedocs.io/en/stable/04_user_guide/15_hooks.html) implementation to add this validation as well for Kedro >= 0.16.

This will allow you to add more features to your extension in the future without being locked into the Kedro release process. Let me know what you think.

limdauto:
Hi @juan-carlos-calvo, thank you again for your contribution. After discussing with the team, this is indeed a very use functionality for Kedro users. However, we believe that a better way to distribute this would be as standalone Python library. The reason is so that we could encourage an ecosystem to spring out of Kedro instead of bringing everything in-house. Here is an example of a library that an open source Kedro user has created: https://github.com/BrickFrog/kedro-pandas-profiling, although in your use case it will be simpler as it doesn't need to be a plugin.

After you create your library, we can include it in our list of community-developed plugins / extensions here: https://kedro.readthedocs.io/en/latest/04_user_guide/10_developing_plugins.html?highlight=plugins#community-developed-plugins.

Let me know what you think.

limdauto:
Hi @juan-carlos-calvo, thank you again for the contribution. As per the previous message and after discussing with the team, I will close this PR in favour of an independent Python library instead. Please do re-open it though if you disagree and we can resume the discussion.

PR #372, opened 2020-05-15
[KED-1681] Add regex filtering to the DataCatalog.list function

datajoely:
## Description
In my work I have found my self writing the following lines frequently:

1. `[x for x in catalog.list() if 'string' in x]`
2. `catalog.__datasets__['dataset_name']._filepath`  <- This is feels naughty when I do so

This PR provides a native mechanism of this functionality in the `DataCatalog.list()` function in a non-breaking way as the previous implementation simply took no arguments. 

## Development notes
- Added two extra parameters to the `DataCatalog.list()` function which can be used in tandem, exclusively or not at all
- Introduced new pytests: 
    1. test_multi_catalog_list_regex
    2. test_multi_catalog_list_bad_regex
    3. test_multi_catalog_list_get_objects
## Checklist

- [x] Added a description of this change and added my name to the list of supporting contributions in the [`RELEASE.md`](../RELEASE.md) file

## Notice


kedro/io/data_catalog.py
```diff
@@ -615,15 +616,45 @@ def add_transformer(
...
+        if regex_search is not None:
+            try:
+                pattern = re.compile(regex_search, flags=re.IGNORECASE)
+                working_data_sets = {
+                    k: v for k, v in self._data_sets.items() if pattern.search(k)
+                }
+            except re.error:
+                raise SyntaxError(
+                    f"Invalid pattern regular expression provided: '{regex_search}'",
+                )
```
mzjp2, line 651: Does the error happen on the `re.compile` line? If so, I'd try (pun intended) to only put the offending line in the `try/except` block, the other one can go outside :)

datajoely: Good point, I'll rejig

WaylonWalker:
Related to #6 

💯 I have monkey patched this into our template and use it 10's of times every day.  Kedro **needs** this functionality.  Thank you so much for adding it in.

My feedback would be that regex may be a bit much for ad-hoc searching, I simply used the `in` operator on my implementation.  Another option that I have recently found is the `fnmatch` module, I really like its simplicity.  It's designed to replicate Unix file pattern matching.

I would also recommend supporting multiple `regex_search` like in #6, I often will use `catalog.query('raw', 'int', 'sales')` to return just the `raw_sales` and `int_sales`.  As catalogs grow really large this becomes super useful to slice into them in various ways.  Maybe if regex was more intuitive to me I wouldn't need this and you can do everything with one regex.

Either way, having a way to slice into a big catalog is really cool, thanks for adding this feature!

datajoely:
> Related to #6
> 
> 💯 I have monkey patched this into our template and use it 10's of times every day. Kedro **needs** this functionality. Thank you so much for adding it in.
> 
> My feedback would be that regex may be a bit much for ad-hoc searching, I simply used the `in` operator on my implementation. Another option that I have recently found is the `fnmatch` module, I really like its simplicity. It's designed to replicate Unix file pattern matching.
> 
> I would also recommend supporting multiple `regex_search` like in #6, I often will use `catalog.query('raw', 'int', 'sales')` to return just the `raw_sales` and `int_sales`. As catalogs grow really large this becomes super useful to slice into them in various ways. Maybe if regex was more intuitive to me I wouldn't need this and you can do everything with one regex.
> 
> Either way, having a way to slice into a big catalog is really cool, thanks for adding this feature!

Hi @WaylonWalker thanks for the feedback. I've tried to summarise the tradeoffs in my thought process:

| Option                  	| Pros                                                                                                                                                                                   	| Cons                                                                                          	|
|-------------------------	|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------	|-----------------------------------------------------------------------------------------------	|
| `in` syntax             	| Simplest                                                                                                                                                                               	| Unable to filter on multiple terms or something complex like a prefix/suffix                                                             	|
| `fnmatch` / glob syntax 	| - Intuitive more complex syntax <br>- NOT exclusion syntax much better than regex                                                                                                      	| - New DSL for the user to understand<br>- Documentation needed to get started as a power user 	|
| `regex`                 	| - Robust and exhaustive mechanism for searching, optional branches are one key benefit<br>- Python `re.search` syntax is forgiving and will match a substring without any special syntax 	| - New (although well known) DSL needed<br>- Most features may never be used                   	|

I like your suggestion of `fnmatch` but I'd argue it is still another DSL to learn one not as helpful to people without a Unix background. I think I feel quite strongly that regex is the right call here because doing something like `^(int|raw)` to find all datasets starting with those two prefixes is quite an elegant way of doing so.

WaylonWalker:
I looked through your examples in your tests after posting my post and agree that you are on the right path and it's  a lack of regex being intuitive to me.  It might be worth tossing a few of those common examples like that in your docstring to make it easier for folks that don't often use regex to grab.  I think the examples from your tests would be a great start.

Again, this is a great feature that kedro needs! You are definitely on the right path.

mzjp2:
Thanks again for raising this PR! I'm a fan of the regex search filter, but a little wary about the other argument `get_objects`. It feels like it's bloating the `list` method a little too much, one hint of this is looking at the return type hint annotation `Union[List[str], Dict[str, AbstractDataSet]]`.

I'd prefer keeping this a little smaller, reducing the complexity and return type of it. It's not too much trouble to get the dataset objects themselves as the caller anyway:

```python
dset_names = catalog.list("regex")
dset_objects = {name: getattr(catalog.datasets, name) for name in dset_names}
```

I don't think the above is particularly burdensome to warrant bloating the `list` method and increasing it's complexity and the API surface of the catalog (which, as a principle, we try and keep as "dumb" as possible).

What do you think?

tests/io/test_data_catalog.py
```diff
@@ -266,6 +266,33 @@ def test_multi_catalog_list(self, multi_catalog):
...
+    def test_multi_catalog_list_regex(self, multi_catalog):
+        """Test that regex patterns filter data sets accordingly"""
+        assert multi_catalog.list(regex_search="^a") == ["abc"]
+        assert multi_catalog.list(regex_search="a|x") == ["abc", "xyz"]
+        assert multi_catalog.list(regex_search="^(?!(a|x))") == []
+        assert multi_catalog.list(regex_search="def") == []
+
+    def test_multi_catalog_list_bad_regex(self, multi_catalog):
+        """Test that bad regex is caught accordingly"""
+        pattern = "Invalid pattern regular expression provided"
```
mzjp2, line 278: Can you test that it displays the bad pattern in the error message? You'll probably need to do some funk regex escaping 🙈

kedro/io/data_catalog.py
```diff
@@ -615,15 +616,55 @@ def add_transformer(
...
+            >>> feat_eng_data = io.list(regex_search='^(prm|feat)')
         """
-        return list(self._data_sets.keys())
+
+        if regex_search is not None:
+            try:
+                pattern = re.compile(regex_search, flags=re.IGNORECASE)
+            except re.error:
+                raise SyntaxError(
+                    f"Invalid pattern regular expression provided: '{regex_search}'"
```
mzjp2, line 656:
```suggestion
                    f"Invalid pattern regular expression provided: `{regex_search}`"
```

I believe that as a rule, we try and use backticks in error messages, so to be consistent^

kedro/io/data_catalog.py
```diff
@@ -615,15 +616,55 @@ def add_transformer(
                 )
             self._transformers[data_set_name].append(transformer)
 
-    def list(self) -> List[str]:
-        """List of ``DataSet`` names registered in the catalog.
+    def list(
+        self, regex_search: Optional[str] = None, get_objects: bool = False
```
mzjp2, line 620: Do you think `regex_filter` is a more appropriate arg name? (you're the regex expert here, so up to you!)

maybe `regex_pattern`?

kedro/io/data_catalog.py
```diff
@@ -615,15 +616,55 @@ def add_transformer(
...
+            ``DataSet`` objects in addition to the names as keys
 
         Returns:
-            A List of ``DataSet`` names, corresponding to the entries that are
-            registered in the current catalog object.
+            A list of ``DataSet`` names or a dictionary of name/object pairs
+            depending on the arguments provided
+
+        Raises:
+            SyntaxError: When invalid regex filter is provided
```
mzjp2, line 639:
```suggestion
            SyntaxError: When an invalid regex filter is provided.
```

Can you also add fullstops to all the other parameter descriptions/docstring entries too? We try and write full sentences in the docstring (key word being _try_ haha).

kedro/io/data_catalog.py
datajoely: I think it's fine, pattern doesn't necessarily say what you're going to do with it.

datajoely:
> Thanks again for raising this PR! I'm a fan of the regex search filter, but a little wary about the other argument `get_objects`. It feels like it's bloating the `list` method a little too much, one hint of this is looking at the return type hint annotation `Union[List[str], Dict[str, AbstractDataSet]]`.
> 
> I'd prefer keeping this a little smaller, reducing the complexity and return type of it. It's not too much trouble to get the dataset objects themselves as the caller anyway:
> 
> ```python
> dset_names = catalog.list("regex")
> dset_objects = {name: getattr(catalog.datasets, name) for name in dset_names}
> ```
> 
> I don't think the above is particularly burdensome to warrant bloating the `list` method and increasing it's complexity and the API surface of the catalog (which, as a principle, we try and keep as "dumb" as possible).
> 
> What do you think?

I understand your point - I'd argue two counter points on this one:

- I've found myself needing to do this A LOT on projects
- `getattr` is not much cleaner
- Maybe a compromise would be to make `DataCatalog._get_dataset` public?
- Since the function has obvious defaults I don't think it complicates the API it almost reminds me of `pd.read_csv` where beginners rarely have to change anything.

tests/io/test_data_catalog.py
datajoely: Updated

datajoely:
@mzjp2  another compromise would be to make a mirror method `list_objects`? But it feels like overkill

DmitriiDeriabinQB:
@datajoely Thanks for raising this! Could you please elaborate a bit more on your use case(s) for using the actual dataset objects?

And regarding the regex, are there any common access patterns for this use case (something like "listing all datasets which name starts with a prefix")?

datajoely:
Hi @DmitriiDeriabinQB I'll try and address both points here:

> Could you please elaborate a bit more on your use case(s) for using the actual dataset objects?

This one is very much an advanced usage, but I still think there should be a publicly accessible way of doing so. In my last project I needed to ingest all DataSets of ``ZippedS3CSVDataSet`` en masse, this updated function would allow me to create a dynamic pipeline without much fuss in a controlled way.

Also in an interactive setting it's also super useful to easily get the `._filepath` or in this case, my custom `.bucket` attribute so that we quickly share links around the team for quick and dirty analysis.

>And regarding the regex, are there any common access patterns for this use case (something like "listing all datasets which name starts with a prefix")? I'm asking because the example from the docstring is doing exactly that - lists the dataset from specific layers by specifying the regex for the name prefix.

Maybe I should use a different example in the docstring, a simple one would be to get all `model_input` tables beginning with `use_case_1` i.e. a subset of a layer. This is an advanced usage, but would be appreciated in day to day use. In essence, I would like both and feel simply exposing layer is insufficient.

mzjp2:
Hey Joel,

So we had an internal discussion and for now at least, the `get_objects` bit seems like a separate piece of work and is increasing the scope of this PR. Can we remove that bit for now, and either create a new PR or an issue where we can discuss it so we can focus on the filtering bit of the `list` in this PR? It sounds like there are two separate discussions going on here 😅 

For full transparency, we're seeing a lot of questions and feature requests pop up around an "interactive workflow", things like `catalog.load(<...>, version=<...>)`, now `catalog.list`. We want to get this interactive workflow right, potentially making it a first class citizen rather than piggybacking on `DataCatalog` which was written with `run` in mind. (credits to @lorenabalan for raising this point)

datajoely:
@mzjp2 I've removed the `get_objects` parameter and appreciate you guys are thinking about a holistic "interactive workflow".

Perhaps there is room for a introducing features that are explicitly marked as something like `beta` would be a great way to get way to get these kind of power user features out quicker in a way that users know that they may brea in the future.

mzjp2:
> Perhaps there is room for introducing features that are explicitly marked as something like `beta` would be a great way to get way to get these kind of power user features out quicker in a way that users know that they may brea in the future.

Yeah, absolutely agree with you there. @limdauto was espousing feature flags (which sounds weird for a library, but is seemingly a done thing) which could be co-opted and potentially be helpful for prototyping new features to (power)-users without guaranteeing that the feature is necessarily released and stable.

RELEASE.md
```diff
@@ -54,6 +54,9 @@
 * Improved the error message for `DataSetNotFoundError` to suggest possible dataset names user meant to type.
 * Added the option for contributors to run Kedro tests locally without Spark installation with `make test-no-spark`.
 * Added option to lint the project without applying the formatting changes (`kedro lint --check-only`).
+* Improved the error message for `DataSetNotFoundError` to suggest possible dataset names user meant to type.
+* Replaced `functools.lru_cache` with `cachetools.cachedmethod` in `PartitionedDataSet` and `IncrementalDataSet` for per-instance cache invalidation.
```
lorenabalan, line 58: Looks like a confused merge conflict here.

kedro/io/data_catalog.py
```diff
@@ -615,15 +616,50 @@ def add_transformer(
                 )
             self._transformers[data_set_name].append(transformer)
 
-    def list(self) -> List[str]:
-        """List of ``DataSet`` names registered in the catalog.
+    def list(self, regex_search: Optional[str] = None) -> List[str]:
+        """
+        List of ``DataSet`` names registered in the catalog if called with
+        default arguments, this can be filtered by providing an optional regular
+        expression which will only return matching keys.
```
lorenabalan, line 623:
```suggestion
        List of all ``DataSet`` names registered in the catalog.
        This can be filtered by providing an optional regular expression 
        which will only return matching keys.
```

kedro/io/data_catalog.py
```diff
@@ -615,15 +616,50 @@ def add_transformer(
...
 
+        Args:
+            regex_search: An optional regular expression which can be provided
+                to limit the data sets returned by a particular pattern.
         Returns:
-            A List of ``DataSet`` names, corresponding to the entries that are
-            registered in the current catalog object.
+            A list of ``DataSet`` names available which match the
+            `regex_search` criteria if provided. All data sets are returned
```
lorenabalan, line 630:
```suggestion
            `regex_search` criteria (if provided). All data set names are returned
```

kedro/io/data_catalog.py
```diff
@@ -615,15 +616,50 @@ def add_transformer(
...
+                    f"Invalid pattern regular expression provided: `{regex_search}`"
+                )
+            working_data_sets = {
+                k: v for k, v in self._data_sets.items() if pattern.search(k)
+            }
+
+        else:
+            working_data_sets = self._data_sets
+
+        return list(working_data_sets.keys())
```
lorenabalan, line 662: We can simplify this now that `get_object` is descoped, and I believe a regex of empty string would match everything.
```suggestion

        if not regex_search:
            return list(self._data_sets.keys())

        try:
            pattern = re.compile(regex_search, flags=re.IGNORECASE)
        except re.error:
            raise SyntaxError(
                f"Invalid pattern regular expression provided: `{regex_search}`"
            )

        matching_data_sets = {
            k: v for k, v in self._data_sets.items() if pattern.search(k)
        }
        return list(matching_data_sets.keys())
```

tests/io/test_data_catalog.py
```diff
@@ -266,6 +266,20 @@ def test_multi_catalog_list(self, multi_catalog):
...
+        """Test that regex patterns filter data sets accordingly"""
+        assert multi_catalog.list(regex_search="^a") == ["abc"]
+        assert multi_catalog.list(regex_search="a|x") == ["abc", "xyz"]
+        assert multi_catalog.list(regex_search="^(?!(a|x))") == []
+        assert multi_catalog.list(regex_search="def") == []
+
+    def test_multi_catalog_list_bad_regex(self, multi_catalog):
+        """Test that bad regex is caught accordingly"""
+        escaped_regex = r"\(\("
+        pattern = f"Invalid pattern regular expression provided: `{escaped_regex}`"
```
lorenabalan, line 279: Sounds more natural?
```suggestion
        pattern = f"Invalid regular expression pattern provided: `{escaped_regex}`"
```

tests/io/test_data_catalog.py
```diff
@@ -266,6 +266,20 @@ def test_multi_catalog_list(self, multi_catalog):
         assert "abc" in entries
         assert "xyz" in entries
 
+    def test_multi_catalog_list_regex(self, multi_catalog):
+        """Test that regex patterns filter data sets accordingly"""
+        assert multi_catalog.list(regex_search="^a") == ["abc"]
+        assert multi_catalog.list(regex_search="a|x") == ["abc", "xyz"]
+        assert multi_catalog.list(regex_search="^(?!(a|x))") == []
+        assert multi_catalog.list(regex_search="def") == []
```
lorenabalan, line 274: You could parametrize this unit test to take regex filter and expected output.
Would it be worth including a case for empty string?

RELEASE.md
```diff
@@ -54,6 +54,9 @@
...
+* Introduced regex filtering and `*DataSet` object access via arguments to the `DataCatalog.list` function.
```
lorenabalan, line 59:
```suggestion
* Introduced regex filtering to the `DataCatalog.list()` function.
```

kedro/io/data_catalog.py
mzjp2: In fact, you no longer need to build a dictionary in the `regex` case, something like:

```python
matching_data_sets [name for name in self._data_sets if pattern.search(name)]
```

should suffice 🥳

datajoely: I've taken bits of both your suggestions. @lorenabalan good shout on the empty string edge case, uncaught it will match all datasets. I've caught this and now log a warning too.

tests/io/test_data_catalog.py
datajoely: Updated, good catch on empty string - I'm not sure I'd want it to match everything

tests/io/test_data_catalog.py
```diff
@@ -266,6 +266,27 @@ def test_multi_catalog_list(self, multi_catalog):
...
+        "pattern,expected",
+        [
+            ("^a", ["abc"]),
+            ("a|x", ["abc", "xyz"]),
+            ("^(?!(a|x))", []),
+            ("def", []),
+            ("", []),
+        ],
+    )
+    def test_multi_catalog_list_regex(self, multi_catalog, pattern, expected):
```
mzjp2, line 279: amazing! 👏

datajoely:
> Almost there, I promise! 🚀🙈
> 
> I went in and fixed the merge errors with `RELEASE.md` stemming from the Kedro release myself, hope you don't mind. :)
> 
> I think with the empty string, I have some mixed feelings about this:
> 
> 1. The special checking for the empty string might be a little "magic" for the end-user, if they just expect whatever they put it to be passed to `re.search`
> 2. If we do go with the special checking for the empty string, not sure if logging is the right mechanism here? Maybe a `warning` instead?
> 
> @lorenabalan @DmitriiDeriabinQB @datajoely thoughts?

So I'm not sure about this - 

- To me the empty string is not a filter, the generic regex character is `.`
- Happy to move to a warning, but I don't know which category it falls under, `SyntaxWarning`?

datajoely:
Can someone merge this for me :) 
image

mzjp2:
Sorry, just waiting for the sync between the two repos to happen, which should flag up a merge conflict, that I'll fix and merge this in within an hour!

mzjp2:
Merging! Took the liberty of adding your name to the release notes under supporting contributors as well. Sorry for the delay on this, you got caught between two releases (0.16 and 0.16.1) 😅 

Thank you! 🙌 🚀

kedro/io/data_catalog.py
```diff
@@ -615,15 +616,47 @@ def add_transformer(
...
+            >>> feat_eng_data = io.list(regex_search='^(prm|feat)')
+            >>> # get data sets which end with 'time_series'
+            >>> models = io.list(regex_search='.+time_series$')
         """
-        return list(self._data_sets.keys())
+
+        if regex_search is None:
+            return list(self._data_sets.keys())
+
+        if not regex_search.strip():
```
DmitriiDeriabinQB, line 651: What if they give `regex_search=" "` to list all dataset names that contain a space character?

datajoely: This would work alternatively `\s` should work. I can’t remember if yaml keys can have spaces, but if so I would expect this behaviour

PR #378, opened 2020-05-19
[KED-1689] Allow JSON strings in CLI `--params` value

nikhiltjoy:
## Description
This adds the ability to send JSON string input to the CLI.

## Development notes
I changed the cookiecutter to parse the string to JSON first, and if it doesn't succeed, it continues with the key:value logic. Tested locally with my own project

## Checklist

- [x] Added a description of this change and added my name to the list of supporting contributions in the [`RELEASE.md`](../RELEASE.md) file

## Notice

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -9,6 +9,7 @@
   - `kedro pipeline create` to create a modular pipeline
 * Improved the CLI speed by up to 50%.
 * Improved error handling when making a typo on the CLI. We now suggest some of the possible commands you meant to type, in `git`-style.
+* Modified the `--params` flag for `kedro run` to allow JSON input strings.
 
 ### Framework
 * All modules in `kedro.cli` and `kedro.context` have been moved into `kedro.framework.cli` and `kedro.framework.context` respectively. `kedro.cli` and `kedro.context` will be removed in future releases.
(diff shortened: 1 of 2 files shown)


mzjp2:
Thank you for raising this PR!

I remember having this discussion with @DmitriiDeriabinQB when he first introduced this functionality. He had some good points against introducing `json` parsing in CLI option values that he might like to share here. :)

DmitriiDeriabinQB:
> Thank you for raising this PR!
> 
> I remember having this discussion with @DmitriiDeriabinQB when he first introduced this functionality. He had some good points against introducing `json` parsing in CLI option values that he might like to share here. :)

The main reason for not allowing JSON is to separate the workflows. The most supported workflow is to provide the parameters as JSON, YAML or any other form of configuration in `conf/` folder, which is then handled by `ConfigLoader` identically to any other project configuration.

Passing parameters via CLI _is_ supported, but only as an auxiliary mechanism to the first use case. If a user tries to pass anything that is more complex than a string or a number (including nested parameters) into a parameter via CLI, we believe they are doing something wrong and should rather use the core mechanism, especially since parameter overriding is supported via config environments.

nikhiltjoyQB:
> > Thank you for raising this PR!
> > I remember having this discussion with @DmitriiDeriabinQB when he first introduced this functionality. He had some good points against introducing `json` parsing in CLI option values that he might like to share here. :)
> 
> The main reason for not allowing JSON is to separate the workflows. The most supported workflow is to provide the parameters as JSON, YAML or any other form of configuration in `conf/` folder, which is then handled by `ConfigLoader` identically to any other project configuration.
> 
> Passing parameters via CLI _is_ supported, but only as an auxiliary mechanism to the first use case. If a user tries to pass anything that is more complex than a string or a number (including nested parameters) into a parameter via CLI, we believe they are doing something wrong and should rather use the core mechanism, especially since parameter overriding is supported via config environments.

Yup, makes sense. I needed it for the use-case of passing a list in --params, and perhaps overriding stuff defined in globals.yml.

However, in this case, we don't need these lines:
https://github.com/quantumblacklabs/kedro/blob/88cf4a471f3b6610360aea6f22de62219fb9112c/kedro/templates/project/%7B%7B%20cookiecutter.repo_name%20%7D%7D/kedro_cli.py#L123-L124

The CLI command only accepts str inputs:
https://github.com/quantumblacklabs/kedro/blob/88cf4a471f3b6610360aea6f22de62219fb9112c/kedro/templates/project/%7B%7B%20cookiecutter.repo_name%20%7D%7D/kedro_cli.py#L190-L192

mzjp2:
> > > Thank you for raising this PR!
> > > I remember having this discussion with @DmitriiDeriabinQB when he first introduced this functionality. He had some good points against introducing `json` parsing in CLI option values that he might like to share here. :)
> > 
> > 
> > The main reason for not allowing JSON is to separate the workflows. The most supported workflow is to provide the parameters as JSON, YAML or any other form of configuration in `conf/` folder, which is then handled by `ConfigLoader` identically to any other project configuration.
> > Passing parameters via CLI _is_ supported, but only as an auxiliary mechanism to the first use case. If a user tries to pass anything that is more complex than a string or a number (including nested parameters) into a parameter via CLI, we believe they are doing something wrong and should rather use the core mechanism, especially since parameter overriding is supported via config environments.
> 
> Yup, makes sense. I needed it for the use-case of passing a list in --params, and perhaps overriding stuff defined in globals.yml.
> 
> However, in this case, we don't need these lines:
> 
> https://github.com/quantumblacklabs/kedro/blob/88cf4a471f3b6610360aea6f22de62219fb9112c/kedro/templates/project/%7B%7B%20cookiecutter.repo_name%20%7D%7D/kedro_cli.py#L123-L124
> 
> The CLI command only accepts str inputs:
> 
> https://github.com/quantumblacklabs/kedro/blob/88cf4a471f3b6610360aea6f22de62219fb9112c/kedro/templates/project/%7B%7B%20cookiecutter.repo_name%20%7D%7D/kedro_cli.py#L190-L192

I believe the reason for that there is to support `params` config passed in via the optional `kedro run --config <config_file>` option, which can specify a dictionary of parameters.

mzjp2:
See this PR: https://github.com/quantumblacklabs/kedro/pull/226 :)

deepyaman:
> See this PR: #226 :)

TIL

DmitriiDeriabinQB:
@mzjp2 Very good point 👍 I already completely forgot about that PR 😄

nikhiltjoy:
cool cool, good to know!

mzjp2:
Thank you regardless for the contribution and hopefully you'll still contribute in the future! 🎉

PR #409, opened 2020-06-11
Fix kedro docs html deployment

tamsanh:
## Description
Running `kedro docs` doesn't work in 0.16.1 due to the move to the `framework` module. When running the command, it tries to find the documents located under `framework` where they do not exist. This is because the deployment process in the `setup.py` and `Makefile` files, the `html` was not updated and so the docs were being deployed to the original `kedro/html` folder rather than the new `kedro/framework/html` folder.

## Development notes
After copying docs to the correct folder, tested with `python -m kedro docs`.
Created a virtualenv and ran `python setup.py install` and then used `kedro docs`.

## Checklist

- [X] Read the [contributing](../CONTRIBUTING.md) guidelines
- [X] Added a description of this change and added my name to the list of supporting contributions in the [`RELEASE.md`](../RELEASE.md) file

## Notice

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 install: build-docs
-	rm -rf kedro/html
-	cp -r docs/build/html kedro
+	rm -rf kedro/framework/html
+	cp -r docs/build/html kedro/framework
 	pip install .
 
 clean:
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -15,10 +15,12 @@
 ## Bug fixes and other changes
 * Removed `/src/nodes` directory from the project template and made `kedro jupyter convert` create it on the fly if necessary.
 * Fixed a bug in `MatplotlibWriter` which prevented saving lists and dictionaries of plots locally on Windows.
+* Repair the `kedro docs` cli command
 
 ## Breaking changes to the API
 
 ## Thanks for supporting contributions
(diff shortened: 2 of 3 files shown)


RELEASE.md
```diff
@@ -15,10 +15,12 @@
 ## Bug fixes and other changes
 * Removed `/src/nodes` directory from the project template and made `kedro jupyter convert` create it on the fly if necessary.
 * Fixed a bug in `MatplotlibWriter` which prevented saving lists and dictionaries of plots locally on Windows.
+* Repair the `kedro docs` cli command
```
921kiyo, line 18:
```suggestion
* Repaired the `kedro docs` cli command
```

921kiyo: Or `Fixed`?

Makefile
```diff
@@ -1,6 +1,6 @@
 install: build-docs
-	rm -rf kedro/html
-	cp -r docs/build/html kedro
+	rm -rf kedro/framework/html
```
mzjp2, line 2: I guess the other fix was to edit the CLI command code itself to look an extra parent above, but this works too - not sure what people prefer

tamsanh: @mzjp2 Yes, you're right. I originally made that change, but then found out there are tests in the CLI command code which asserts that the `kedro/framework/html` is the correct location. Decided to do this change rather than change the test.

PR #415, opened 2020-06-18
[KED-1798] Fix DataSet string representation for falsy values

### LINKED ISSUE #413: [KED-1798] AbstractDataSet.__str__ excludes zero-valued integers
_(Apologies for deviating from the template, but this is quite straightforward and probably doesn't need all of those sections)_

### Example

```pycon
>>> from kedro.io import AbstractDataSet
>>> class MyDataSet(AbstractDataSet):
...     def __init__(self, var):
...         self.var = var
...     def _describe(self):
...         return dict(var=self.var)
...     def _load(self):
...         pass
...     def _save(self, data):
...         pass
...
>>> str(MyDataSet(1))
'MyDataSet(var=1)'
>>> str(MyDataSet(0))
'MyDataSet()'
```

`var` isn't displayed when its value is zero, even though that could be a meaningful thing to display. This is due to the `if value` filter below:

https://github.com/quantumblacklabs/kedro/blob/c141467739d696c1f69dfd51be33b8d02d45b501/kedro/io/core.py#L266

Is falsiness an appropriate check, or should numeric values be included even when they're zero?

deepyaman:
## Description
Close #413 

## Checklist

## Notice

- [ x I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":


kedro/io/core.py
```diff
@@ -263,7 +263,7 @@ def _to_str(obj, is_root=False):
                 text = ", ".join(
                     fmt.format(key, _to_str(value))  # 2
                     for key, value in sorted_dict
-                    if value or isinstance(value, bool)
+                    if value != {} and value is not None
```
idanov, line 266: Should we restrict this just to `is not None`? Empty dictionaries and sets should also be provided, since in Python it is not recommended to have emtpy lists, objects or sets as default values in constructors.

deepyaman: I think that makes sense. I modified it to follow whatever is written in the description (didn't know if there was particular design decisions around excluding empty dictionaries--and you would know this best--but I'd figure it had to do with load and save args getting replaced with default dictionaries. Even then, I think what you say makes sense.

Right now:

```python
>>> from kedro.extras.datasets import pandas
>>> pandas.CSVDataSet(filepath="test.txt")
<kedro.extras.datasets.pandas.csv_dataset.CSVDataSet object at 0x10a9cf150>
>>> str(pandas.CSVDataSet(filepath="test.txt"))
"CSVDataSet(filepath=test.txt, protocol=file, save_args={'index': False})"
```

Your suggestion:

```python
>>> str(pandas.CSVDataSet(filepath="test.txt"))
"CSVDataSet(filepath=test.txt, protocol=file, load_args={}, save_args={'index': False})"
```

Alternative (don't process defaults; I think this may have been the intention of the above code, but I still think the outputs make less sense/are less explicit):

```python
>>> str(pandas.CSVDataSet(filepath="test.txt"))
"CSVDataSet(filepath=test.txt, protocol=file)"  # Because no args were passed and it's `None` before the default substitution.
```

deepyaman: @idanov There will be some (trivial) changes to the tests to support this, but the failures also illustrate what the difference would look like; e.g. `SQLDataSet` will include `load_args={}` in the representation, `PartitionedDataSet` will have `dataset_config={}`. Just want to confirm this change is desirable first. :)

deepyaman:
@idanov @DmitriiDeriabinQB Done! :)

tests/extras/datasets/pandas/test_sql_dataset.py
```diff
@@ -309,5 +309,5 @@ def test_save_error(self, query_data_set, dummy_dataframe):
     def test_str_representation_sql(self, query_data_set):
         """Test the data set instance string representation"""
         str_repr = str(query_data_set)
-        assert "SQLQueryDataSet(sql={})".format(SQL_QUERY) in str_repr
+        assert "SQLQueryDataSet(load_args={{}}, sql={})".format(SQL_QUERY) in str_repr
```
DmitriiDeriabinQB, line 312: Maybe also an f-string here?

tests/io/test_core.py
```diff
@@ -26,14 +26,55 @@
...
+
+    def _load(self):
+        pass  # pragma: no cover
+
+    def _save(self, data):
+        pass  # pragma: no cover
 
 
 class TestCoreFunctions:
+    @pytest.mark.parametrize("var", [1, True] + FALSE_BUILTINS[1:])  # None is not shown
```
DmitriiDeriabinQB, line 71: I would remove `None` from that list as it's not used anywhere in the tests

tests/extras/datasets/pandas/test_sql_dataset.py
deepyaman:
```suggestion
        assert f"SQLQueryDataSet(load_args={{}}, sql={SQL_QUERY})" in str_repr
```

DmitriiDeriabinQB:
As this is not a breaking change, it needs to be merged into master. Changing a target branch doesn't work due to commit history difference. So I've created #418 against master and manually applied the diff from here.

PR #521, opened 2020-09-20
[DO NOT MERGE YET] Support versioning of the underlying dataset

deepyaman:
## Checklist

## Notice


kedro/io/partitioned_data_set.py
```diff
@@ -53,6 +54,13 @@
 S3_PROTOCOLS = ("s3", "s3a", "s3n")
 
 
+def _grandparent(path: str) -> str:
+    path_obj = Path(path)
```
DmitriiDeriabinQB, line 58: We should probably either use `pathlib.PurePosixPath` here, or `return grandparent.as_posix()` below (or both) to make sure we don't end up returning Windows paths with `\` separator for S3, for example.

kedro/io/partitioned_data_set.py
```diff
@@ -53,6 +54,13 @@
...
+    grandparent = path_obj.parent.parent
```
DmitriiDeriabinQB, line 59:
```suggestion
    grandparent = path_obj.parents[1]
```

kedro/io/partitioned_data_set.py
```diff
@@ -53,6 +54,13 @@
...
+    assert grandparent.name == path_obj.name
```
DmitriiDeriabinQB, line 60: `assert`s don't provide a particularly informative error message. Can you rather raise a `DataSetError`?

kedro/io/partitioned_data_set.py
deepyaman: Ah! You're probably right. I had `path_obj = Path(path).resolve()` before which messed it up for S3-ish, so I took it out, but this must be why Windows is failing.

deepyaman: Sorry for the delay on this. Seems it fixed the Windows test errors!

kedro/io/partitioned_data_set.py
deepyaman: Done!

tests/io/test_partitioned_dataset.py
```diff
@@ -274,17 +279,56 @@ def test_invalid_dataset_config(self, dataset_config, error_pattern):
     @pytest.mark.parametrize(
         "dataset_config",
         [
-            {"type": CSVDataSet, "versioned": True},
-            {"type": "pandas.CSVDataSet", "versioned": True},
+            {**dataset_config, "versioned": True}
+            for dataset_config in LOCAL_DATASET_DEFINITION
+            if isinstance(dataset_config, dict)
```
deepyaman, line 284: Seems to be failing due to a Pylint bug (https://github.com/PyCQA/pylint/issues/2491, https://github.com/PyCQA/pylint/issues/3877).

```suggestion
            {**ds_config, "versioned": True}
            for ds_config in LOCAL_DATASET_DEFINITION
            if isinstance(ds_config, dict)
```

deepyaman:
@DmitriiDeriabinQB Changes made/test issues resolved; this PR should be good to go! :shipit: 🤞

deepyaman:
@DmitriiDeriabinQB Can you take a look at this when you get a chance? :)

DmitriiDeriabinQB:
Hey @deepyaman, sorry for the delay, I'll have a look asap.

kedro/io/partitioned_data_set.py
```diff
@@ -58,6 +59,18 @@
...
+def _grandparent(path: str) -> str:
+    path_obj = PurePosixPath(path)
+    grandparent = path_obj.parents[1]
+    if grandparent.name != path_obj.name:
+        last_three_parts = path_obj.relative_to(*path_obj.parts[:-3])
+        raise DataSetError(
+            f"`{path}` is not a well-formed versioned path ending with "
+            f"`filename/timestamp/filename` (got `{last_three_parts}`)."
+        )
+    return str(grandparent)
```
DmitriiDeriabinQB, line 71: Maybe?
```suggestion
    return grandparent.as_posix()
```

kedro/io/partitioned_data_set.py
```diff
@@ -217,7 +220,7 @@ def _normalized_path(self) -> str:
     @cachedmethod(cache=operator.attrgetter("_partition_cache"))
     def _list_partitions(self) -> List[str]:
         return [
-            path
+            _grandparent(path) if self._dataset_config.get(VERSION_KEY) else path
```
DmitriiDeriabinQB, line 223: For readability maybe refactor it into a variable?
```python
dataset_versioned = self._dataset_config.get(VERSION_KEY, False)
return [
    _grandparent(path) if dataset_versioned else path
    for path in self._filesystem.find(self._normalized_path, **self._load_args)
    if path.endswith(self._filename_suffix)
]
```

tests/io/test_partitioned_dataset.py
```diff
@@ -328,7 +372,7 @@ def test_no_partitions(self, tmpdir):
     def test_filepath_arg_warning(self, pds_config, filepath_arg):
         pattern = (
             f"`{filepath_arg}` key must not be specified in the dataset definition as it "
-            f"will be overwritten by partition path"
```
DmitriiDeriabinQB, line 331: We usually keep an `f` prefix  for all string parts even if they aren't formatted (at least this is how PyCharm does it by default)

kedro/io/partitioned_data_set.py
```diff
@@ -58,6 +59,18 @@
...
 
 
+def _grandparent(path: str) -> str:
+    path_obj = PurePosixPath(path)
+    grandparent = path_obj.parents[1]
+    if grandparent.name != path_obj.name:
+        last_three_parts = path_obj.relative_to(*path_obj.parts[:-3])
+        raise DataSetError(
+            f"`{path}` is not a well-formed versioned path ending with "
+            f"`filename/timestamp/filename` (got `{last_three_parts}`)."
```
DmitriiDeriabinQB, line 69: (Optional) Maybe also add a hint about that was expected? Something like:
```python
f"`filename/timestamp/filename` (got `{last_three_parts}`, "
f"expected `{path_obj.name}/<version>/{path_obj.name}`)."
```

DmitriiDeriabinQB (approved):
This is awesome @deepyaman! 🏆 
I really like that it keeps the user API unchanged 👍

DmitriiDeriabinQB:
One more consideration - I think we might also need to update `IncrementalDataSet` with this change (since it doesn't inherit from `PartitionedDataSet._list_partitions`).

kedro/io/partitioned_data_set.py
deepyaman: I couldn't remember whether this line had any impact on S3 path handling, so I tried running through an example real quick:

```
>>> from pathlib import PurePosixPath
>>> path_obj = PurePosixPath("s3://bucket-name/path/to/folder/new/partition.csv")
>>> path_obj.parents[1]
PurePosixPath('s3:/bucket-name/path/to/folder')
```

Since `path_obj` is a `PurePosixPath`, `grandparent` is, too. As a result, I don't think this does anything (the `as_posix` call basically adds a path separator substitution that won't match anything).

Still, if this makes it more clear, I'll change it; let me leave it for last.

kedro/io/partitioned_data_set.py
```diff
@@ -217,7 +220,7 @@ def _normalized_path(self) -> str:
     @cachedmethod(cache=operator.attrgetter("_partition_cache"))
     def _list_partitions(self) -> List[str]:
         return [
```
deepyaman, line 222:
```suggestion
        dataset_is_versioned = self._dataset_config.get(VERSION_KEY, False)
        return [
```

kedro/io/partitioned_data_set.py
deepyaman:
```suggestion
            _grandparent(path) if dataset_is_versioned else path
```

kedro/io/partitioned_data_set.py
```diff
@@ -216,8 +219,9 @@ def _normalized_path(self) -> str:
 
     @cachedmethod(cache=operator.attrgetter("_partition_cache"))
     def _list_partitions(self) -> List[str]:
+        dataset_is_versioned = self._dataset_config.get(VERSION_KEY, False)
```
deepyaman, line 222:
```suggestion
        dataset_is_versioned = self._dataset_config.get(VERSIONED_FLAG_KEY, False)
```
Should this actually be `VERSIONED_FLAG_KEY`, a la https://github.com/quantumblacklabs/kedro/blob/774a29d925a4331e70743155963cd9b643a6bd8c/kedro/io/core.py#L435? I don't think it matters functionally, more from a readability perspective.

tests/io/test_partitioned_dataset.py
```diff
@@ -328,7 +372,7 @@ def test_no_partitions(self, tmpdir):
...
+            "will be overwritten by partition path"
```
deepyaman, line 375:
```suggestion
            f"will be overwritten by partition path"
```

kedro/io/partitioned_data_set.py
deepyaman, line 69:
```suggestion
            f"`filename/timestamp/filename` (got `{last_three_parts}`, "
            f"expected `{path_obj.name}/<version>/{path_obj.name}`)."
```

kedro/io/partitioned_data_set.py
deepyaman: JK, https://github.com/quantumblacklabs/kedro/blob/develop/kedro/io/core.py#L430 pops `VERSIONED_FLAG_KEY` out of the parsed config.

deepyaman, line 223:
```suggestion
        dataset_is_versioned = VERSION_KEY in self._dataset_config
```

kedro/io/partitioned_data_set.py
DmitriiDeriabinQB: Cool, `str(grandparent)` makes sense then 👍

tests/io/test_partitioned_dataset.py
```diff
@@ -286,17 +291,56 @@ def test_invalid_dataset_config(self, dataset_config, error_pattern):
...
+            str(local_dir / "path/to/folder"),
+            {"type": "pandas.CSVDataSet", "versioned": True},
         )
-        with pytest.raises(DataSetError, match=re.escape(pattern)):
-            PartitionedDataSet(str(Path.cwd()), dataset_config)
+
+        pattern = re.escape(
+            f"`{path.as_posix()}` is not a well-formed versioned path ending with "
+            "`filename/timestamp/filename` (got `version/partition/"
+            "file`)."
```
deepyaman, line 340:
```suggestion
            "file`, expected `file/<version>/file`)."
```

tests/io/test_partitioned_dataset.py
```diff
@@ -286,17 +291,56 @@ def test_invalid_dataset_config(self, dataset_config, error_pattern):
...
+            str(local_dir / "path/to/folder"),
+            {"type": "pandas.CSVDataSet", "versioned": True},
         )
-        with pytest.raises(DataSetError, match=re.escape(pattern)):
-            PartitionedDataSet(str(Path.cwd()), dataset_config)
+
+        pattern = re.escape(
+            f"`{path.as_posix()}` is not a well-formed versioned path ending with "
+            "`filename/timestamp/filename` (got `version/partition/"
+            "file`, expected `file/<version>/file`)."
```
deepyaman, line 340:
```suggestion
            "file`)."
```

kedro/io/partitioned_data_set.py
```diff
@@ -58,6 +59,19 @@
...
 
+def _grandparent(path: str) -> str:
+    path_obj = PurePosixPath(path)
+    grandparent = path_obj.parents[1]
+    if grandparent.name != path_obj.name:
+        last_three_parts = path_obj.relative_to(*path_obj.parts[:-3])
+        raise DataSetError(
+            f"`{path}` is not a well-formed versioned path ending with "
+            f"`filename/timestamp/filename` (got `{last_three_parts}`, "
+            f"expected `{path_obj.name}/<version>/{path_obj.name}`)."
```
deepyaman, line 70:
```suggestion
            f"`filename/timestamp/filename` (got `{last_three_parts}`)."
```

tests/io/test_partitioned_dataset.py
deepyaman: Switching this back, as the "expected" message is a confusing IMO if somebody is using a folder structure at the bottom level (e.g. as generated by Spark). The would think they already have version in `"path/to/folder/new/partition/version/partition/file"`, so the recommendation to have `file/<version>/file` seems ill-advised (really, they shouldn't be using a folder structure at the bottom, so let's not confuse them).

kedro/io/partitioned_data_set.py
```diff
@@ -58,6 +59,18 @@
...
+    if grandparent.name != path_obj.name:
+        last_three_parts = path_obj.relative_to(*path_obj.parts[:-3])
+        raise DataSetError(
+            f"`{path}` is not a well-formed versioned path ending with "
+            f"`filename/timestamp/filename` (got `{last_three_parts}`)."
+        )
+    return str(grandparent)
+
+
 class PartitionedDataSet(AbstractDataSet):
```
DmitriiDeriabinQB, line 74: Hey @deepyaman! I believe if we want this to be merged before 0.17, we might need these changes to be implemented for the `IncrementalDataSet` too 🤔  (since `_list_partitions` method for the incremental one isn't inherited from the `PartitionedDataSet`). Please let me know whether it makes sense? 🙂

lorenabalan (approved):
LGTM ✨

kedro/io/partitioned_data_set.py
lorenabalan: Also should this merged in `develop` instead?

limdauto (approved):
LGTM! I have also changed the base to `develop`

DmitriiDeriabinQB:
@deepyaman If you don't mind, I'll attempt to quickly resolve the conflicts + address the change for the incremental dataset.

DmitriiDeriabinQB:
While trying to fix this, I've encountered a potentially major shortcoming:

When we instantiate the underlying dataset, we call `parse_dataset_definition` _without_ load and save versions provided. It means that if someone is saving new partitions with the versioning enabled, the new partitions will be generated using the same save version, _but_ this version timestamp will be different than the save version timestamp for any other versioned dataset within the same run. This might become a source of quite some confusion for end users.

The reason is in this `parse_dataset_definition` call, which usually receives load and save versions generated upstream (by the DataCatalog initialiser). But passing the save version into the PartitionedDataSet initialiser turned out to be a rather tricky business (I couldn't quickly come up with anything not awfully hacky). Because of that, PartitionedDataSet generates its own save version, which differs from the "global" save version associated with the run.

Potential options:
1. Move `versioned: true` flag from the underlying dataset into the PartitionedDataSet itself - this way the PartitionedDataSet receive the version in the `__init__` call and will pass it into the underlying dataset.
2. Leave this implementation be with the version inconsistencies pointed out above.
3. Put this change on hold and redesign the solution entirely.

DmitriiDeriabinQB:
After having an internal discussion, it's probably best to hold off from merging it before the 0.17 release for now and give ourselves a bit more time to think about the correct implementation.

The most feasible option so far seems to be the option 1 from my [comment above](https://github.com/quantumblacklabs/kedro/pull/521#issuecomment-744653023), where we would move `versioned: true` flag from the underlying dataset into the PartitionedDataSet itself.

The only question I'd like to ask any interested contributor is:
❓❓  _would the YAML API below be clear for the end users of Kedro and how easy would it be to realise that the versioning flag only applies to the underlying dataset rather than the partitioned dataset as a whole?_ ❓❓

```yaml
my_partitioned_dataset:
  type: PartitionedDataSet
  path: path/to/folder
  dataset:
    type: pandas.CSVDataSet
    load_args: ...
    save_args: ...
  versioned: true  # <- versioning enabled
```

deepyaman:
@limdauto @lorenabalan Any idea on a path forward on this?

I'm happy with Option 1. Putting `versioned: true` at the dataset level but applying to the partition level is fine, as long as it’s documented as such. `PartitionedDataSet` is a bit of a weird case regardless.

datajoely:
Hey @deepyaman do you think you'll get a chance to finish this one?

datajoely:
@idanov  @lorenabalan will we be progressing this or shall we close?

lorenabalan:
@datajoely let's close it for now.

merelcht:
Re-open to be part of research.

stichbury (approved):
The docs change (fixes the broken link in Prefect docs) LGTM. I cannot sign off on the others 🤷 Thanks @deepyaman 🏆 🏅 🌟

noklam:
Excited to add versioning to PartitionedDataset! Thanks @deepyaman 

2023-08-02 Technical Design
cross-post @deepyaman comments
> - We discussed the options laid out by Dmitrii in https://github.com/kedro-org/kedro/pull/521#issuecomment-744653023; it seems that some people (e.g. Yetu, Sajid) would be happy to just go with #2 and gather feedback sooner, while others (e.g. Merel, Ahdra, Nok) would be happy to start with #1 (bit more involved, but hopefully not significantly). Nobody who was in favor of #2 has any objection to #1.
> - A number of people (e.g. Marcin, Ahdra) would be in favor of getting some more user feedback first. I will send out a feeler to [#user-research](https://kedro-org.slack.com/archives/C03R8N2M8KT) with these options + "I don't want any of this; I need versioning of the top-level dataset" to see if anybody has strong opinions they want to voice. I'll try to send it out today or tomorrow.
> - Unless the [#user-research](https://kedro-org.slack.com/archives/C03R8N2M8KT) poll comes out with significant feedback to change our view, will eventually go forward with #1 to get this feature in the hands of users.

astrojuanlu:
Should we close this now that https://github.com/kedro-org/kedro-plugins/pull/447 was opened?

PR #524, opened 2020-09-22
Add minimal, black-compatible flake8 configuration

deepyaman:
## Description
Kedro configures and then uses both Black and Flake8 (use via the `kedro lint` command); however, it fails to provide a Black-compatible Flake8 config. Kedro does provide config for other tools, so it makes not sense to ignore it.

Also, why do the commands [here](https://github.com/quantumblacklabs/kedro/blob/c615639b0612a47f6bab31736a94585611540b8b/kedro/framework/cli/project.py#L128-L132) specify all those options, if Kedro also provides configuration files? I can remove the inline options, if so desired. Otherwise, seems like somebody could update their config and wonder why `kedro lint` doesn't work as expected.

## Checklist

## Notice


lorenabalan:
> I can remove the inline options, if so desired. Otherwise, seems like somebody could update their config and wonder why kedro lint doesn't work as expected.

That's a good point. However that might break behaviour for projects generated with older template versions. Maybe the 2 changes (config file and remove from in-line commands) are worth bundling together and shipping on `develop`, as part of `0.17.0`. 
Unless @mzjp2 @DmitriiDeriabinQB you can remember a reason why we shouldn't? (you guys were closer to the `kedro lint` feature at the time I think)

Also, another heads-up that this is the sort of change that again needs to be replicated across starters. 🙃

mzjp2:
> That's a good point. However that might break behaviour for projects generated with older template versions. Maybe the 2 changes (config file and remove from in-line commands) are worth bundling together and shipping on `develop`, as part of `0.17.0`.
> Unless @mzjp2 @DmitriiDeriabinQB you can remember a reason why we shouldn't? (you guys were closer to the `kedro lint` feature at the time I think)

I don't think I have an objection. I don't remember what my reasoning was - if the config files didn't already exist, I suspect we were trying to not increase the number of files in the template, but if they did then it was probably just an oversight on my part. :)

kedro/templates/project/{{ cookiecutter.repo_name }}/setup.cfg
```diff
@@ -1,3 +1,7 @@
+[flake8]
+max-line-length = 88
+extend-ignore = E203
```
deepyaman, line 3:
```suggestion
max-line-length=88
extend-ignore=E203
```
Copied from https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length, style and all. Would you prefer I remove the surrounding spaces to be consistent with the `isort` section below?

kedro/templates/project/{{ cookiecutter.repo_name }}/setup.cfg
```diff
@@ -1,3 +1,7 @@
...
+
 [isort]
```
deepyaman, line 5: Two more questions here:

1. Is there any reason to set `force_grid_wrap=0`? It is the default.
2. Do we need `known_third_party=kedro`? It should be detected like any other library.

deepyaman:
> Two more questions here:
> 
> 1. Is there any reason to set `force_grid_wrap=0`? It is the default.
> 2. Do we need `known_third_party=kedro`? It should be detected like any other library.

Aside from the above two questions, I think this is ready.

> That's a good point. However that might break behaviour for projects generated with older template versions. Maybe the 2 changes (config file and remove from in-line commands) are worth bundling together and shipping on `develop`, as part of `0.17.0`.

@lorenabalan Do you want me to repoint this to `develop` then?

kedro/templates/project/{{ cookiecutter.repo_name }}/setup.cfg
DmitriiDeriabinQB: Hey @deepyaman!

Re 1: I think the reason was purely to comply with [Black suggestions](https://black.readthedocs.io/en/stable/compatible_configs.html). Otherwise things may break if isort (ever) decide to change its defaults.

Re 2: It might help with running some unit/e2e tests as Kedro might not be installed at that point but rather just be present in the current working dir (so might be considered a 1st party package, I think)

kedro/templates/project/{{ cookiecutter.repo_name }}/setup.cfg
```diff
@@ -1,3 +1,7 @@
+[flake8]
+max-line-length=88
+extend-ignore=E203
```
DmitriiDeriabinQB, line 3: Maybe? As per what [Black suggests](https://black.readthedocs.io/en/stable/compatible_configs.html#flake8)
```suggestion
extend-ignore=E203,W503
```

DmitriiDeriabinQB (approved):
LGTM! 👍

kedro/templates/project/{{ cookiecutter.repo_name }}/setup.cfg
deepyaman: I guess it's optional. Both `W503` and it's counterpart, `W504`, are ignored by default (https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-ignore) as of https://pycodestyle.pycqa.org/en/latest/developer.html#id9, hence the minimal-compatible config here: https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length (which is what I've used)

921kiyo (approved):
LGTM 👍

PR #538, opened 2020-10-03
give name of function inside incorrect parameters error

### LINKED ISSUE #495: better error messages when wrong number of inputs is passed
## Description

When passing the wrong number of inputs into a kedro node it can be very confusing to where this happened, a more customized message that tells the user more information about the node would be quite helpful.  Some things to consider including would be; the name of the node, inputs, outputs, tags, and function name. 

## Context

Today I ran into an issue where I had scaffolded a new pipeline with many nodes before running.  And saw this error.  It took me a bit longer than it should have to figure out where I had gone wrong.

```
TypeError: Inputs of function expected [], but got a_raw_<dataset>
```

## Context

## Possible Implementation

The error occurred in [kedro/pipeline/node.py#L553](https://github.com/quantumblacklabs/kedro/blob/master/kedro/pipeline/node.py#L553), It would be quite helpful to throw a custom error and include as much information about the node as is reasonable in this error message.

## Alternately

A custom error message is not required, but more information would be greatly appreciated.

WaylonWalker:
#  Description

This is a quick and simple step towards the right direction to resolve #495.

## Development notes

I ran `kedro new`, and broke one of the nodes.  There is a comparison of the current error and the updated error below.  This error is also inspected by the test framework.

### Current Error

Does not give much information to what node caused the TypeError.  While this is generally easy to figure out on small pipelines, larger pipelines can be quite intense to figure out where the error came from.  Any additional information would be greatly appreciated.

image

### Updated Error

The updated error will give the name of the function for full functions, but does not provide better messaging for lambdas.

image

### Feedback

* Is the format of the message ok?
* Is there any potential that the passed in function would not have a `func.__code__.co_name` method?

## Checklist

- [X] This PR is on my personal behalf and does not represent any employer

**NOTE** No new tests were created, only modification of existing tests.

## Notice

### DIFF
--- a/tests/pipeline/test_node.py
+++ b/tests/pipeline/test_node.py
@@ -337,15 +337,15 @@ def inconsistent_input_kwargs():
     [
         (
             inconsistent_input_size,
-            r"Inputs of function expected \[\'input1\'\], but got \[\'A\', \'B\'\]",
+            r"Inputs of function 'dummy_func_args' expected \[\'input1\'\], but got \[\'A\', \'B\'\]",
         ),
         (
             inconsistent_input_args,
-            r"Inputs of function expected \[\'args\'\], but got {\'a\': \'A\'}",
+            r"Inputs of function 'dummy_func_args' expected \[\'args\'\], but got {\'a\': \'A\'}",
         ),
         (
             inconsistent_input_kwargs,
-            r"Inputs of function expected \[\'kwargs\'\], but got A",
+            r"Inputs of function 'dummy_func_args' expected \[\'kwargs\'\], but got A",
         ),
     ],
 )
--- a/tests/pipeline/test_node.py
+++ b/tests/pipeline/test_node.py
@@ -359,15 +359,17 @@ def partial_inconsistent_input_size():
         (
             inconsistent_input_args,
             (
-                r"Inputs of '<function dummy_func_args at .*>' "
-                r"expected \[\'args\'\], but got {\'a\': \'A\'}"
+                r"Inputs of '"
+                r"<function inconsistent_input_args.<locals>.dummy_func_args"
+                r" at .*>' expected \[\'kwargs\'\], but got A"
             ),
         ),
         (
             inconsistent_input_kwargs,
             (
-                r"Inputs of '<function dummy_func_args at .*>' "
-                r"expected \[\'kwargs\'\], but got A"
+                r"Inputs of '"
+                r"<function inconsistent_input_kwargs.<locals>.dummy_func_args"
+                r" at .*>' expected \[\'kwargs\'\], but got A"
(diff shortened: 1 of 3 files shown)


tests/pipeline/test_node.py
```diff
@@ -337,15 +337,15 @@ def dummy_func_args(**kwargs):
     [
         (
             inconsistent_input_size,
-            r"Inputs of function expected \[\'input1\'\], but got \[\'A\', \'B\'\]",
+            r"Inputs of function 'identity' expected \[\'input1\'\], but got \[\'A\', \'B\'\]",
```
lorenabalan, line 340: This looks nicer. Can we add some tests for `lambda` and `partial` when they are passed as the callable to the `node`? Not sure if it'll be as pretty.

WaylonWalker:
@lorenabalan any ideas how to get tests passing?  For some reason `58636d9` was only failing on 3.8.

lorenabalan:
Actually looking at the test cases, I would argue the error message looks more complicated now - addresses in memory are not that user-friendly. Could we instead factor out and leverage the same logic we have in `Node._func_name`? 😃

WaylonWalker:
I am not sure that my change should be causing a seg fault in the memory profiler tests???

Could you rerun tests and see if it persists?

I think I got it cleaned up, and it looks much better, than for the `_func_name` suggestion @lorenabalan!

WaylonWalker:
Thanks for re-running!  Tests are passing.  @lorenabalan  how does the updated format look?  I like that it matches up with other kedro function naming methods.

Personally I am not sure that I like that there are more than one place figuring out how to name a function, but that would have started touching more things than I preferred to do from my end without first discussing.

lorenabalan:
> Thanks for re-running! Tests are passing. @lorenabalan how does the updated format look? I like that it matches up with other kedro function naming methods.

The format looks good to me - maybe could be `Inputs of "this_func_name" function` instead of `Inputs of "this_func_name"` in the error message but that's a minor point. 
 
> Personally I am not sure that I like that there are more than one place figuring out how to name a function, but that would have started touching more things than I preferred to do from my end without first discussing.

Yeah I agree - looking through the code base, I also found [this variation of the implementation](https://github.com/quantumblacklabs/kedro/blob/841a747e09855e9d85d97d932647b5ace7f57a09/kedro/pipeline/decorators.py#L37). 😅 The decorators are going to be deprecated in 0.17.0 so I'm not that worried about that one, it just illustrates that we don't have a consistent way to handle it. Could we factor out the logic from `_func_name` into a helper function `_get_readable_func_name` or smth, to call in both places (in _func_name and in _validate_inputs)? And only warn about `update_wrapper` in `Node._func_name`.

PR #559, opened 2020-10-12
Support passing arguments to underlying filesystem

deepyaman:
## Description

Motivating Slack conversation:

Q: Is `fs_args` not supported for the overarching `PartitionedDataSet`? I need to specify `endpoint_url` for my S3 interface, and I can do this on other datasets (e.g. `pandas.ParquetDataSet`) but not on `PartitionedDataSet`. Without it, I can’t discover partitions.

A: Currently you can put those into `credentials` as this is the only set of kwargs passed into the filesystem constructor. Would it be acceptable for your use case? I will create a ticket to add `fs_args` support. Thanks for raising this 👍 

## Development notes
See #522 for prior activity

## Checklist

## Notice


deepyaman:
@DmitriiDeriabinQB Added a test, as requested. When looking into that, was wondering... does it make sense to, by default, pass `fs_args` to the underlying dataset, too? I can't think of any cases where it would make sense to pass `fs_args` to the `PartitionedDataSet` constructor but the underlying dataset wouldn't take it.

If so, can implement this and corresponding tests much like it is for credentials.

DmitriiDeriabinQB:
@deepyaman Thanks very much for this! Yes, I would say that it makes perfect sense to propagate `fs_args` the same way we do for credentials 👍

deepyaman:
> @deepyaman Thanks very much for this! Yes, I would say that it makes perfect sense to propagate `fs_args` the same way we do for credentials 👍

Done and ready to go! :shipit: 🤞

kedro/io/partitioned_data_set.py
```diff
@@ -98,6 +103,7 @@ def __init__(  # pylint: disable=too-many-arguments
         filepath_arg: str = "filepath",
         filename_suffix: str = "",
         credentials: Dict[str, Any] = None,
+        fs_args: Dict[str, Any] = None,
```
DmitriiDeriabinQB, line 106: Sorry for nitpicking, I haven't noticed that the last time I had look, but as `fs_args` is not the last argument in the constructor, then this change technically become a breaking change if someone instantiates the dataset programmatically using positional arguments only. I know it's 99%+ that it's not going to be the case, but just to be on a safe side.

To mitigate this, could you move `fs_args` in the constructor below `load_args` please (for both partitioned and incremental ones)?

DmitriiDeriabinQB (commented):
Awesome stuff! 👏 👏 
Just one little comment below regarding backwards compatibility.

kedro/io/partitioned_data_set.py
deepyaman: > Sorry for nitpicking

Not at all! I noticed that the order was different from that of other datasets, and it bothered me, but I didn't want to make unnecessary changes. Sounds like there's a good reason for it! I'll fix it shortly.

deepyaman: @DmitriiDeriabinQB Should be good now!

kedro/io/partitioned_data_set.py
```diff
@@ -97,8 +102,9 @@ def __init__(  # pylint: disable=too-many-arguments
         dataset: Union[str, Type[AbstractDataSet], Dict[str, Any]],
         filepath_arg: str = "filepath",
         filename_suffix: str = "",
-        credentials: Dict[str, Any] = None,
```
DmitriiDeriabinQB, line 100: Sorry, `credentials` need to stay right after the `filename_suffix` not to break the order of the arguments in the constructor. To keep things 100% backwards compatible when extending the arguments, we should only add new args to the end of the list (and of course keep then optional).

DmitriiDeriabinQB: Let me push a quick commit for that

DmitriiDeriabinQB (approved):
LGTM! ⭐

andrii-ivaniuk (approved):

PR #568, opened 2020-10-17
give name of function inside incorrect parameters error

### LINKED ISSUE #495: better error messages when wrong number of inputs is passed

WaylonWalker:
#  Description
_I messed up my author config, I rebased and opened this new PR similar to #538_

This is a quick and simple step towards the right direction to resolve #495.

## Development notes

I ran `kedro new`, and broke one of the nodes.  There is a comparison of the current error and the updated error below.  This error is also inspected by the test framework.

### Current Error

Does not give much information to what node caused the TypeError.  While this is generally easy to figure out on small pipelines, larger pipelines can be quite intense to figure out where the error came from.  Any additional information would be greatly appreciated.

image

### Updated Error

The updated error will give the name of the function for full functions, but does not provide better messaging for lambdas.

image

### Feedback

* Is the format of the message ok?
* Is there any potential that the passed in function would not have a `func.__code__.co_name` method?

## Checklist

- [X] This PR is on my personal behalf and does not represent any employer

**NOTE** No new tests were created, only modification of existing tests.

## Notice

# conversation from #538  
---

> @WaylonWalker : @lorenabalan any ideas how to get tests passing? For some reason 58636d9 was only failing on 3.8. It appeared to match exactly to me besides the escape characters.

> @lorenabalan : Actually looking at the test cases, I would argue the error message looks more complicated now - addresses in memory are not that user-friendly. Could we instead factor out and leverage the same logic we have in Node._func_name? 😃

> @WaylonWalker : I am not sure that my change should be causing a seg fault in the memory profiler tests???
> Could you rerun tests and see if it persists?
> I think I got it cleaned up, and it looks much better, than for the _func_name suggestion @lorenabalan!

> @WaylonWalker : Thanks for re-running! Tests are passing. @lorenabalan how does the updated format look? I like that it matches up with other kedro function naming methods.
> Personally I am not sure that I like that there are more than one place figuring out how to name a function, but that would have started touching more things than I preferred to do from my end without first discussing.

> @lorenabalan : Thanks for re-running! Tests are passing. @lorenabalan how does the updated format look? I like that it matches up with other kedro function naming methods.
> The format looks good to me - maybe could be Inputs of "this_func_name" function instead of Inputs of "this_func_name" in the error message but that's a minor point.
> Personally I am not sure that I like that there are more than one place figuring out how to name a function, but that would have started touching more things than I preferred to do from my end without first discussing.
> Yeah I agree - looking through the code base, I also found this variation of the implementation. 😅 The decorators are going to be deprecated in 0.17.0 so I'm not that worried about that one, it just illustrates that we don't have a consistent way to handle it. Could we factor out the logic from _func_name into a helper function _get_readable_func_name or smth, to call in both places (in _func_name and in _validate_inputs)? And only warn about update_wrapper in Node._func_name.


WaylonWalker:
@lorenabalan I implemented the change to move `_func_name` logic to a helper function `_get_readable_func_name `. And referenced it in bother places rather than duplicating.  Let me know your thoughts.  Also heads up, I messed up my git author and didn't want duplicate commits so I opened a new PR.

Looks like I ran into the same segfault/timeout error in testing.  Can you rerun them?

kedro/pipeline/node.py
```diff
@@ -696,3 +685,27 @@ def _to_list(element: Union[None, str, Iterable[str], Dict[str, str]]) -> List:
     if isinstance(element, dict):
         return sorted(element.values())
     return list(element)
+
+
+def _get_readable_func_name(
+    func: Callable, outputs: Union[None, str, List[str], Dict[str, str]] = None
+) -> str:
+    """Get a readable name of the function provided.
```
lorenabalan, line 693:
```suggestion
    """Get a user-friendly readable name of the function provided.
```

kedro/pipeline/node.py
```diff
@@ -550,9 +537,11 @@ def _validate_inputs(self, func, inputs):
                 func_args = inspect.signature(
                     func, follow_wrapped=False
                 ).parameters.keys()
+                func_name = _get_readable_func_name(func)
+
                 raise TypeError(
-                    "Inputs of function expected {}, but got {}".format(
-                        str(list(func_args)), str(inputs)
+                    "Inputs of '{}' function expected {}, but got {}".format(
+                        func_name, str(list(func_args)), str(inputs)
```
lorenabalan, line 544: Can f-string this now.

kedro/pipeline/node.py
```diff
@@ -696,3 +685,27 @@ def _to_list(element: Union[None, str, Iterable[str], Dict[str, str]]) -> List:
...
+        str: readable name of the provided callable func.
+    """
+
+    if hasattr(func, "__name__"):
+        return func.__name__
+
+    name = repr(func)
+    if "functools.partial" in name:
+        warn(
+            "The node producing outputs `{}` is made from a `partial` function. "
```
lorenabalan, line 705: Can use f-strings now.

kedro/pipeline/node.py
```diff
@@ -198,20 +198,7 @@ def __call__(self, **kwargs) -> Dict[str, Any]:
...
-        name = repr(self._func)
-        if "functools.partial" in name:
-            warn(
-                "The node producing outputs `{}` is made from a `partial` function. "
-                "Partial functions do not have a `__name__` attribute: consider using "
-                "`functools.update_wrapper` for better log messages.".format(
-                    self.outputs
-                )
-            )
-            name = "<partial>"
```
lorenabalan, line 213: You can leave this bit here and change the condition to look like:
```
name = _get_readable..(func)
if name == "<partial>":
    warn(...)

return name
```

Otherwise if it fails on validating the inputs for a `partial` on line 540, the warning message is inaccurate because it has no knowledge of its outputs. And I would argue it's maybe cleaner to leave the warning just here, and not potentially confuse the error message.

lorenabalan (commented):
Thanks for your patience with this! Nearly there.

kedro/pipeline/node.py
WaylonWalker: At one point I had checked for outputs being none and only raised if outputs were not none.  I like this better.

kedro/pipeline/node.py
WaylonWalker: Awesome!!  Kedro is the only place I have used anything else in years.

lorenabalan (approved):
Awesome! Can you please remove `temp1` file? 😅  I think it's been accidentally committed at some point.

kedro/pipeline/node.py
```diff
@@ -197,20 +197,14 @@ def __call__(self, **kwargs) -> Dict[str, Any]:
...
 
     @property
-    def _func_name(self):
-        if hasattr(self._func, "__name__"):
-            return self._func.__name__
-
-        name = repr(self._func)
-        if "functools.partial" in name:
+    def _func_name(self) -> str:
+        name = _get_readable_func_name(self._func, self.outputs)
```
lorenabalan, line 201:
```suggestion
        name = _get_readable_func_name(self._func)
```

WaylonWalker: Thanks @lorenabalan!

kedro/pipeline/node.py
```diff
@@ -550,10 +544,10 @@ def _validate_inputs(self, func, inputs):
                 func_args = inspect.signature(
                     func, follow_wrapped=False
                 ).parameters.keys()
+                func_name = _get_readable_func_name(func)
+
                 raise TypeError(
-                    "Inputs of function expected {}, but got {}".format(
-                        str(list(func_args)), str(inputs)
-                    )
+                    f"Inputs of '{func_name}' function expected {list(func_args)}, but got {inputs}"
```
DmitriiDeriabinQB, line 550: This line is probably longer than 88 characters, so needs to be broken down in 2

kedro/pipeline/node.py
```diff
@@ -197,20 +197,14 @@ def __call__(self, **kwargs) -> Dict[str, Any]:
...
-
-        name = repr(self._func)
-        if "functools.partial" in name:
+    def _func_name(self) -> str:
+        name = _get_readable_func_name(self._func)
+        if name == "<partial>":
             warn(
-                "The node producing outputs `{}` is made from a `partial` function. "
+                f"The node producing outputs `{self.outputs}` is made from a `partial` function. "
                 "Partial functions do not have a `__name__` attribute: consider using "
```
DmitriiDeriabinQB, line 205: We try follow a convention where all f-string parts are prefixed with `f` even if there are no variables inside. It just for the consistency with how PyCharm, for example, breaks down such strings.
```suggestion
                f"Partial functions do not have a `__name__` attribute: consider using "
```

kedro/pipeline/node.py
```diff
@@ -197,20 +197,14 @@ def __call__(self, **kwargs) -> Dict[str, Any]:
...
-                "`functools.update_wrapper` for better log messages.".format(
-                    self.outputs
-                )
+                "`functools.update_wrapper` for better log messages."
```
DmitriiDeriabinQB, line 206:
```suggestion
                f"`functools.update_wrapper` for better log messages."
```

DmitriiDeriabinQB (approved):
Nice! 👍

RELEASE.md
```diff
@@ -18,13 +18,16 @@
...
+* Improved error messages for incorrect parameters passed into a node.
 * Fixed issue with saving a `TensorFlowModelDataset` in the HDF5 format with versioning enabled.
 * Added missing `run_result` argument in `after_pipeline_run` Hooks spec.
 
 ## Breaking changes to the API
 
 ## Thanks for supporting contributions
-[Deepyaman Datta](https://github.com/deepyaman), [Bhavya Merchant](https://github.com/bnmerchant), [Lovkush Agarwal](https://github.com/Lovkush-A), [Varun Krishna S](https://github.com/vhawk19), [Sebastian Bertoli](https://github.com/sebastianbertoli), [Daniel Petti](https://github.com/djpetti)
+
+
```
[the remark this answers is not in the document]
WaylonWalker: Good Catch

kedro/pipeline/node.py
WaylonWalker: good catch

WaylonWalker:
Thanks for the help @DmitriiDeriabinQB !  All passing now.

PR #573, opened 2020-10-20
Revert "Add minimal, black-compatible flake8 configuration"

DmitriiDeriabinQB:
Reverts quantumblacklabs/kedro#524

Unfortunately, this has to be reverted from `master` since this represents a breaking change for existing projects - if a project was generated using Kedro `0.16.*` without the change from #524 _and_ Kedro is updated to the version with that change, `kedro lint` command will break since flake8 won't be configured (setup.cfg still lacks `[flake8]` section and CLI args are already gone).

I will migrate the changes from #524 into `develop` so that they are released in Kedro `0.17.0`.


limdauto (approved):

921kiyo (approved):
👍

lorenabalan (approved):

PR #601, opened 2020-11-07
Add regex filter to PartitionedDataset

gonzalofrancoceballos:
## Description
Adds functionality to apply a regex filter before loading data using `PartitionedDataSet`

## Development notes
Added a new parameter `regex_filter` to `PartitionedDataSet` and modified `_list_partitions` method accordingly
Added new test cases in `tests.io.test_partitioned_dataset`

## Checklist
- ✅ Read the [contributing](https://github.com/quantumblacklabs/kedro/blob/master/CONTRIBUTING.md) guidelines
- ✅ Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- ✅ Updated the documentation to reflect the code changes
- ✅ Added a description of this change and added my name to the list of supporting contributions in the [`RELEASE.md`](https://github.com/quantumblacklabs/kedro/blob/master/RELEASE.md) file
- ✅ Added tests to cover my changes

## Notice

- ✅ I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":


kedro/io/partitioned_data_set.py
```diff
@@ -220,6 +226,7 @@ def _list_partitions(self) -> List[str]:
             path
             for path in self._filesystem.find(self._normalized_path, **self._load_args)
             if path.endswith(self._filename_suffix)
+            and bool(re.search(self._regex_filter, Path(path).name))
```
DmitriiDeriabinQB, line 229: 1. For a small performance improvement we should probably [compile](https://docs.python.org/3/library/re.html#re.compile) the expression before using it for filtration
2. Shall we maybe apply regex filter to the whole `path` rather than just its filename? I think it will unlock more use cases for the users

kedro/io/partitioned_data_set.py
```diff
@@ -102,6 +104,7 @@ def __init__(  # pylint: disable=too-many-arguments
         dataset: Union[str, Type[AbstractDataSet], Dict[str, Any]],
         filepath_arg: str = "filepath",
         filename_suffix: str = "",
+        regex_filter: str = "",
```
DmitriiDeriabinQB, line 107: I suggest that this should also be implemented for the `IncrementalDataSet`

DmitriiDeriabinQB (commented):
Hi @gonzalofrancoceballos, thank you for the PR! The idea generally LGTM, here are some extra suggestions:
1. If we go for the regular expressions, we should probably make `regex_filter` the only filtration mechanism and get rid of `filename_suffix` since it's relatively easy to provide a regex that filters the file extension (should be documented).
2. The changes should also be reflected in the partitioned dataset docs (`docs/source/05_data/02_kedro_io.md#partitioned-dataset`)

Plus some comments inline below.

kedro/io/partitioned_data_set.py
gonzalofrancoceballos: Thank you @DmitriiDeriabinQB , about the second comment. I am only applying the filter to the files in the path because `path` the same for al files, am I right?

PR #614, opened 2020-11-14
Refine split of node names string with multiple inputs / outputs

torazem:
## Description
To fix https://github.com/quantumblacklabs/kedro/issues/613

## Development notes
Rejoined tokens that are likely to belong to input/output parameters

## Checklist

## Notice

### DIFF
--- a/kedro/framework/cli/utils.py
+++ b/kedro/framework/cli/utils.py
@@ -209,6 +209,25 @@ def split_string(ctx, param, value):  # pylint: disable=unused-argument
     return [item.strip() for item in value.split(",") if item.strip()]
 
 
+def split_node_string(ctx, param, value):  # pylint: disable=unused-argument
+    """Split string of node names separated by comma"""
+    rough_split = value.split(",")
+
+    # NB: There are still edge-cases that can break the following, but
+    # they are unlikely to happen unintentionally.
+    corrected_split = []
+    current_token_group = []
+
+    for token in rough_split:
+        current_token_group.append(token)
+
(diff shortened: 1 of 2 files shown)


kedro/framework/cli/utils.py
```diff
@@ -209,6 +209,25 @@ def split_string(ctx, param, value):  # pylint: disable=unused-argument
...
+
+    # NB: There are still edge-cases that can break the following, but
+    # they are unlikely to happen unintentionally.
+    corrected_split = []
+    current_token_group = []
+
+    for token in rough_split:
+        current_token_group.append(token)
+
+        if token.endswith("-> None") or token.endswith("]"):
```
lorenabalan, line 224: This reads very fragile to me and rather overfitted, because I don't think this would work if there was a node function that returned something (not None). The function also doesn't appear to be used anywhere.
Personally I think it's better to leave the current behaviour as is as there's no easy way to tell when a comma is intentional vs when it isn't? As I mentioned [in this comment](https://github.com/quantumblacklabs/kedro/issues/613#issuecomment-730407903), we recommend users define their own user-friendly names anyway.

torazem: Agreed; this was just enough to stop my use cases from breaking without introducing breaking changes, but it probably won't age well.

lorenabalan:
Closing this as addressed by https://github.com/quantumblacklabs/kedro/commit/e500f999e8ae479631233ac032c2b9d7a267c258

PR #616, opened 2020-11-17
[KED-2254] Add datatable csv dataset (#592)

### LINKED ISSUE #592: Integrate datable to Kedro extras datasets
## Description
Add [datatable](https://github.com/h2oai/datatable) to Kedro's extras datasets

## Context
[datatable](https://github.com/h2oai/datatable) is derived from R's [data.table](https://github.com/Rdatatable/data.table). As of 20/09/2020 ([Version 0.11.0](https://datatable.readthedocs.io/en/latest/releases/v0.11.0.html)), datatable is fully supported on Windows, Linux and MacOS and can be installed through pip. Offering datable in Kedro's extras dataset could offer many advantages in building pipelines:

- It would make the transition for R-users wanting to use Kedro a lot easier.
- Datable supports CSV, XLSX, and plain text as well as .tar, .gz, .zip, .gz2, and .tgz.
- Certain performance advantages over pandas (especially for datasets larger than 50GB): [benchmark](https://h2oai.github.io/db-benchmark/)
- fread is really fast at reading large CSVs. It also automatically detects separators.
- datatable's df.to_csv has an append parameter which appends the dataframe with column names if the file does not exist yet. If the file already exists, the column names are not added. Pandas to_csv offers "mode='a'" and "header=True" parameters which means choosing between rewriting the column names with each append or not having any column names. This is really practical for pipelines where you add new data from an IncrementalDatasets folder to a concatenated CSV.

## Possible Implementation
- Using the code from [pandas.CSVDataSet](https://github.com/quantumblacklabs/kedro/blob/master/kedro/extras/datasets/pandas/csv_dataset.py), we could change the "pd.read_csv(fs_file, **self._load_args)" to "dt.fread(fs_file, **self._load_args)". Saving the file should be similar code albeit different parameter names.
- For ExcelDataSets, we could also use code from [pandas.ExcelDataSet](https://github.com/quantumblacklabs/kedro/blob/master/kedro/extras/datasets/pandas/excel_dataset.py). Datatable also uses the xlrd engine for xlsx. Replace "pd.read_excel(fs_file, **self._load_args)" with "dt.fread(fs_file, **self._load_args)". Datatable does not output xlsx. So either we do not implement ExcelDataSets for datatable or we convert the datatable to pd.DataFrame and write to excel using pd.DataFrame.to_excel.

mlisovyi:
## Description
Adding dataset to read/write csv using [datatable](https://datatable.readthedocs.io/en/latest/index.html) instead of pandas.
Resolves #592 

## Development notes
I have added only CSV and not Excel as a Dataset. The later generates a lot of problems, when operating with `fsspec`. And when i try to use the file name directly I anyway fail the basic tests as for some reason write -> read sequence on Excel file does not yield the same data type of columns `datatable.Frame` as in the original `datatable.Frame`.

**To be discussed:**
What is the purpose of this add-on. While implementing it, i realized that:
* one gets a `datatable.Frame` object instead of `pandas.DataFrame`:
  * for R users, that are familiar with `data.table` in R this is beneficial as it provides a familiar API instead of pandas API
  * but it runs only on CSV as input
  * but it can be confusing that having CSVDataset but with a different prefix one gets an object that has a very different API
* one gets performance gain of datatable over pandas:
  * multi-thread reading of CSV
  * optimised string storage, i.e. smaller memory footprint by `object` column type
  * out-of-core computation (one can in principle process data that is larger than RAM)
  * but only for CSV input and maybe it would rather make sense to make a wrapper around `pandas.XXXDataset` to be able to read all pandas-supported formats and then benefit from datatable performance.

## Checklist

## Notice


lucasjamar (commented):
I think you need to use datatable >=0.11.0 because it was not compatible to windows prior to this release: https://datatable.readthedocs.io/en/latest/releases/v0.11.0.html
Maybe this is what is causing the code to fail the windows pip compile and unit tests in CircleCI?

mlisovyi:
@lucasjamar good point. Thanks for the suggestion!
We will need to have a look into datatable version. The `0.11.0` version on linux gave me a lot of headache :( When i pip-installed it into the nominal conda environment that had all kedro-requested packages included, the environment went into some weird broken state due to version conflicts and i was not able to do  do even `conda list`. Only uninstalling datatable with pip helped to get the environment into working state.
This requires more work to be invested to understand what goes wrong.

lucasjamar:
> 
> 
> @lucasjamar good point. Thanks for the suggestion!
> We will need to have a look into datatable version. The `0.11.0` version on linux gave me a lot of headache :( When i pip-installed it into the nominal conda environment that had all kedro-requested packages included, the environment went into some weird broken state due to version conflicts and i was not able to do do even `conda list`. Only uninstalling datatable with pip helped to get the environment into working state.
> This requires more work to be invested to understand what goes wrong.

@mlisovyi  I already mentionned this [conda issue](https://github.com/h2oai/datatable/issues/2746) to datatable. Its a real bummer.

mlisovyi:
The issue with datatable has been fixed on datatable side. But we need to wait for the next bugfix(?) release (supposedly 0.11.1)

yetudada:
Thank you so much @mlisovyi! We'll get this reviewed!

kedro/extras/datasets/datatable/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+# confusion in the marketplace, including but not limited to in advertising,
+# on websites, or on software.
+#
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""``AbstractDataSet`` implementations that produce datatable Frames."""
+
+__all__ = [
+    "CSVDataSet",
```
MerelTheisenQB, line 32: I'd suggest using a different name to make it easier to distinguish from the pandas `CSVDataSet`. Perhaps `CSVDataTableDataSet`?

kedro/extras/datasets/datatable/csv_dataset.py
```diff
@@ -0,0 +1,178 @@
...
+
+        # Handle default load and save arguments
+        self._load_args = copy.deepcopy(self.DEFAULT_LOAD_ARGS)
+        if load_args is not None:
+            self._load_args.update(load_args)
+        self._save_args = copy.deepcopy(self.DEFAULT_SAVE_ARGS)
+        if save_args is not None:
+            self._save_args.update(save_args)
+
+        _fs_open_args_load.setdefault("mode", "r")
```
MerelTheisenQB, line 134: Any specific reason why you're using `r` mode here? We've recently updated a lot of our datasets to just use the default modes from fsspec, so if that works for this dataset as well I'd suggest removing this line.

kedro/extras/datasets/datatable/csv_dataset.py
```diff
@@ -0,0 +1,178 @@
...
+
+        with self._fs.open(load_path, **self._fs_open_args_load) as fs_file:
+            return dt.fread(file=fs_file, **self._load_args)
+
+    def _save(self, data: dt.Frame) -> None:
+        save_path = get_filepath_str(self._get_save_path(), self._protocol)
+
+        with self._fs.open(save_path, **self._fs_open_args_save) as fs_file:
+            # convert to pandas before saving as otherwise can not use fs_file
+            data.to_pandas().to_csv(path_or_buf=fs_file, **self._save_args)
```
MerelTheisenQB, line 159: Why are you still using `pandas` to save the data?

MerelTheisenQB (commented):
Thanks for the contribution @mlisovyi ! I added a couple of questions about the implementation.

kedro/extras/datasets/datatable/csv_dataset.py
mlisovyi: Indeed, that is inefficient and wierd.
The reason is that datatable doesn't seem to support file-like object as input for writing out a table. Therefore one can not benefit from access to multiple various filesystems that is provided by `fsspec.open()`

kedro/extras/datasets/datatable/csv_dataset.py
mlisovyi: sure, i believe this will work dropping the line. I will try it out and take it down.

There was no good reason apriory- it was inherited from somewhere else, i believe :)

kedro/extras/datasets/datatable/csv_dataset.py
MerelTheisenQB: Right, so the main purpose of this dataset would be that you can load data in `datatable` and then work with it, before saving it again as pandas? If so, I'd suggest to update the description of the class to make this clear.

mlisovyi: yes, correct. the main 2 benefits would be:
* speed-up of csv reading, as `datatable` allows multi-thread reading from csv
* usage of data manipulation API that is familiar to R users, who might not have experience with `pandas`.
There is a nice list of advantages summarised in the original issue (#592).

Ok, i'll add doc-string of the class. This and the other proposed change will take a couple of days due to other tasks on my TODO.

mlisovyi: ok, the changed have been pushed

kedro/extras/datasets/datatable/csv_dataset.py
mlisovyi: ok, the changed have been pushed

kedro/extras/datasets/datatable/csv_dataset.py
```diff
@@ -0,0 +1,185 @@
...
+    """``CSVDataSet`` loads/saves data from/to a CSV file using an underlying
+    filesystem (e.g.: local, S3, GCS). It uses datatable to handle the CSV file.
+
+    The main differences with respect to the ``pandas`` implementation:
+
+    - speed-up of csv reading, as datatable allows multi-thread reading from csv;
+    - usage of data manipulation API that is familiar to R users,
+      who might not have experience with ``pandas``.
+
+    Note that ``pandas.DataFrame`` is used for writing out.
```
MerelTheisenQB, line 58: Thanks, this is a great explanation! 👏

MerelTheisenQB (approved):
This looks good to me! 👍  Can you please add your name to the `RELEASE.md` list of contributors and we will be good to go! 🎉

kedro/extras/datasets/datatable/csv_dataset.py
```diff
@@ -0,0 +1,185 @@
...
+        Args:
+            filepath: Filepath in POSIX format to a CSV file prefixed with a protocol like `s3://`.
+                If prefix is not provided, `file` protocol (local filesystem) will be used.
+                The prefix should be any protocol supported by ``fsspec``.
+                Note: `http(s)` doesn't support versioning.
+            load_args: Datatable options for loading CSV files.
+                Here you can find all available arguments:
+                https://datatable.readthedocs.io/en/latest/api/dt/fread.html
+                All defaults are preserved.
+            save_args: Datatable options for saving CSV files.
```
DmitriiDeriabinQB, line 99:
```suggestion
            save_args: Pandas options for saving CSV files.
```

kedro/extras/datasets/datatable/csv_dataset.py
DmitriiDeriabinQB: According to [`Frame.to_csv()`](https://datatable.readthedocs.io/en/latest/api/frame/to_csv.html) docs, "If no path is given, then the Frame will be serialized into a string, and that string will be returned".

So maybe we should just get such string, `.encode("utf-8")` it and send to fsspec instead of converting to pandas?

test_requirements.txt
```diff
@@ -5,6 +5,7 @@ behave==1.2.6
 biopython~=1.73
 black==v19.10.b0
 dask[complete]~=2.6
+datatable>=0.11.1, <1.0
```
DmitriiDeriabinQB, line 8: Datatable should also be added into `extras_require` in `setup.py`

kedro/extras/datasets/datatable/csv_dataset.py
mlisovyi: Unfortunately, I do not have enough experience to judge on it. I can blindly implement your suggestion, but I will not be able to validate it due to lack of time.

kedro/extras/datasets/datatable/csv_dataset.py
mlisovyi: Resolving, as there has been no feedback since 6 days

kedro/extras/datasets/datatable/csv_dataset.py
mlisovyi: Unfortunately, I do not have enough experience to judge on it. I can blindly implement your suggestion, but I will not be able to validate it due to lack of time.

test_requirements.txt
mlisovyi: Good catch! Thanks.

I'm not really familiar with the `extras_require` configuration for setuptools. I have added requirements as far as I get the logic, but could you please have a look that it is correct? In particular the cross-dependency of entries with datatable and pandas requirements is not fully clear to me.

mlisovyi:
the failures in the py3.7 and py3.8 linting are very unclear to me. After a quick googling, my guess is that it is related to PyCQA/pylint#3318. I have no clue why does it does not appear in linting check with other python versions and my knowledge at this stage does not allow to pursue it further

MerelTheisenQB:
> the failures in the py3.7 and py3.8 linting are very unclear to me. After a quick googling, my guess is that it is related to [PyCQA/pylint#3318](https://github.com/PyCQA/pylint/issues/3318). I have no clue why does it not appear in linting check with other python versions and my knowledge at this stage does not allow to pursue it further

We're seeing these errors on other builds as well, so they're not related to your PR. I'll investigate further.

MerelTheisenQB:
Hi @mlisovyi , will you be able to finish this PR or are you happy with me jumping in and making it ready for merging?

MerelTheisenQB:
Hi @mlisovyi, will you be able to address the comments on this PR? We're looking at cleaning up the open PR's, so we'll be deleting this in 2 weeks. Thanks!

mlisovyi:
Hi @MerelTheisenQB ,

my apologies, unfortunately I'm currently overwhelmed with tasks at work and I do not see me finding time in the nearest future to contribute to the project. So I think this has to be closed to reduce maintenance burden for the package maintainers. If somewhere will come back to the original issue and will work on it one can always find the branch in my fork.

Once again my apologies.

MerelTheisenQB:
Thanks for getting back about this @mlisovyi 🙂  No worries at all! Hopefully we'll see you back on the Kedro repo again in the future 🚀

PR #708, opened 2021-03-03
Feature/skip node hook

pascalwhoop:
## Description
A hook may want to have the power to entirely skip the execution of a node. While this is probably uncommon, it can be very useful when attempting to make a kedro pipeline only run nodes when their logic or inputs have changed.

Also, I'd like some feedback on the way of skipping a node. An alternative concept would be to add a `disabled` property to the `Node` class which this hook could flip. Then later in the flow, the runner checks if the node is disabled before running it. 

### Underlying idea: Skip unchanged node/output tuples
Here are two small images describing the purpose of the change

This is inspired from how `make` or `docker build` behaves. It doesn't execute steps that don't need to be executed because the outputs wouldn't change. In kedro, if we assume code is deterministic, the same is true for a node.

A common situation for large projects is that pipelines contain dozens or even hundreds of nodes and several engineers tweak nodes here and there. When a pipeline run takes several hours, just running `kedro run` is costly. Ideally, there would be a way to "just run" the pipeline but the pipeline itself is smart enough to figure out what needs to be run and what not. To not run a node, it needs to be able to skip a node execution including its output writing. 

We have also discussed this a bit already in the forum a while back
https://discourse.kedro.community/t/speeding-up-pipeline-processing-with-change-detection/90/17

## Development notes

I added an ability for a hook to return a boolean to signal that it wants to skip a node. 

## Checklist

## Notice

- I certify that (a) this contribution is my original creation and / 

or (b) to the extent it is not my original creation, I am authorised to submit this contribution on behalf of the original creator(s) or their licensees.

### DIFF
--- a/kedro/runner/runner.py
+++ b/kedro/runner/runner.py
@@ -32,7 +32,7 @@ implementations.
 import logging
 from abc import ABC, abstractmethod
 from concurrent.futures import ALL_COMPLETED, ThreadPoolExecutor, as_completed, wait
-from typing import Any, Dict, Iterable
+from typing import Any, Dict, Iterable, Tuple
 
 from kedro.framework.hooks import get_hook_manager
 from kedro.io import AbstractDataSet, DataCatalog
@@ -222,7 +222,7 @@ def _collect_inputs_from_hook(
     inputs: Dict[str, Any],
     is_async: bool,
     run_id: str = None,
-) -> Dict[str, Any]:
+) -> Tuple[Dict[str, Any], bool]:
     inputs = inputs.copy()  # shallow copy to prevent in-place modification by the hook
     hook_manager = get_hook_manager()
     hook_response = hook_manager.hook.before_node_run(  # pylint: disable=no-member
@@ -230,7 +230,11 @@ def _collect_inputs_from_hook(
     )
 
     additional_inputs = {}
+    skip = False
     for response in hook_response:
+        if response is not None and isinstance(response, bool):
+            skip = response
+            continue
         if response is not None and not isinstance(response, dict):
             response_type = type(response).__name__
             raise TypeError(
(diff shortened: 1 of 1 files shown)


MerelTheisenQB:
I think this is a great idea and will definitely benefit a lot of Kedro users! 😄  
Just to clarify, would the idea be to add a `before_node_run` implementation to core Kedro that takes care of skipping the node, or would you add separate hook implementation that behave differently depending on the type of nodes/pipeline? And these would either be auto-registered or the user can register them specifically to their project?

pascalwhoop:
The idea is to write a separate hook implementation that users can choose to add to their `HOOKS`. But for that to be possible (I have a working PoC tested against the pandas-iris starter) kedro has to somehow allow a hook to signal that a node shouldn't be executed.

I could replace the `node.func` with an empty lambda and create a "fake" node which is just empty. But kedro also tries to write the outputs of that node and that also shouldn't happen.

The way this PR is now, the hook returns a boolean flag to signal if the node should be skipped. But maybe it would be better if I add a `disabled` flag to the `Node` class? I'd prefer if the kedro core team makes that call. And of course if this is at all something that fits into the philosophy.

MerelTheisenQB:
Thanks for clarifying @pascalwhoop. I think the idea itself is great, and something that would ideally be possible in Kedro. However, hooks is probably not the way to go to achieve it. Originally, Hooks were added as a mechanism to allow a user to extend Kedro by injecting additional behaviour at certain lifecycle points in Kedro’s main execution. But a hook that signals whether a node should be executed or not would be interfering with the execution flow itself. This functionality should probably happen at the runner level. Obviously, this requires more in-depth technical design, so that's something we should take on as the Kedro team.

pascalwhoop:
@MerelTheisenQB thanks for that response. If I were to write this at the runner level, would it maybe make sense that I write a version that utilises multiple inheritance (aka mixins)? Currently, you have a few runners (Thread/Parallel/Sequential) but all of them could/should be able to run with this node skipping functionality. So maybe mixins would be a way to resolve this so you can have a runner that combines these functions (e.g. run in threads + skip nodes)

PR #720, opened 2021-03-11
Feature/build-reqs options can be passed from the install cli command

gcalmettes:
## Description

Kedro uses `pip-compile` to keep the `requirements.txt` file up to date.
Any arguments/options to be passed to `pip-compile` can currently be directly added to the `kedro build-reqs` command.

However, when running the initial `kedro install` command (which then conditionally triggers `kedro build-reqs`), the `pip-compile` options cannot be passed.

We found it problematic as `pip-compile` by default adds any extra configuration found in the local `~/.pip/pip.conf` file to the generated `requirements.txt`, including for example any credentials to private pypi repositories present in that file (see https://github.com/jazzband/pip-tools/issues/1099).
 
The workaround for now is to first run `kedro install` and then immediately run `kedro build-reqs --no-emit-index-url` to prevent those credentials to be leaked into the generated `requirements.txt`, as `kedro build-reqs` already accepts extra arguments to be passed to `pip-compile`.

This PR adds a new `--build-reqs-args` optional flag to the `kedro install` cli command to allow directly passing any extra argument at the initial install phase.  

## Development notes

A new `--build-reqs-args` has been added as option to the `install` cli command, and those options are directly passed to the [_build_reqs](https://github.com/quantumblacklabs/kedro/blob/master/kedro/framework/cli/project.py#L61) function.

## Checklist

## Notice

### DIFF
--- a/docs/source/04_kedro_project_setup/01_dependencies.md
+++ b/docs/source/04_kedro_project_setup/01_dependencies.md
@@ -19,6 +19,8 @@ The `build-reqs` command will:
 
 To further update the project requirements, you should modify `src/requirements.in` (not `src/requirements.txt`) and re-run `kedro build-reqs`.
 
+> Any extra arguments for the `pip-compile` command can be directly passed as extra flags to `kedro build-reqs`. For example, if you need to install private python packages, you can add your private pypi repository to the repositories `pip-compile` will search in using `kedro build-reqs --extra-index-url="https://private.pypi.repository/simple/"` (see `pip-compile -h` for all the available extra arguments).
+
 
 ## `kedro install`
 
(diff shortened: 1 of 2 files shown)


AntonyMilneQB:
Hello @gcalmettes, and thank you very much for your contribution. It's a nice idea, but I've talked it over with the kedro team and we've decided we would prefer not to implement this feature. 
* The notation `kedro install --build-reqs-args="--annotate"` seems a little awkward as it's nesting arguments, and currently none of the other CLI commands have this sort of pattern
* Adding these arguments directly to `kedro install`, as `kedro install --annotate`, would not work well as we might want to add other arguments to `kedro install` in future, and then it's not clear which should be passed to `build_reqs`
* The current approach of doing `kedro install && kedro build-reqs --extra-args` already achieves the desired behaviour, so it does not seem essential to add this option to `install` since it won't achieve any really new functionality

Thank you again for the idea and your very clear PR and description, and sorry we're not adding it! Your work is much appreciated though 🙂

gcalmettes:
Hello @AntonyMilneQB ,

Thanks for the feedbacks !
Yes I understand totally the decision of the Kedro team. As you said building first the dependencies with the appropriate flags before the kedro install definitely works, and indeed adding the complexity of nested arguments might  not be a good idea on the long run for the kedro API

I am closing this PR as a result.
Thanks again!

PR #734, opened 2021-03-23
Feature/custom errors

WaylonWalker:
Struggling to run Tests locally, trying to leverage CI for this idea.


WaylonWalker:
@idanov I do not see why this would fail `unit_test_36`, but pass `unit_test_38` and `unit_test_37`.  I can't see the results, is it related to the change?

MerelTheisenQB (approved):
Thanks for this PR @WaylonWalker I think this is a great suggestions 👍

kedro/errors/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+
+"""
+Expose public exceptions and warnings.
+"""
+
+
+class KedroConfigError(RuntimeError):
+    """
+    Error raised when kedro fails to startup due to configuration.
+    """
```
lorenabalan, line 37:
```suggestion
    """Error raised when kedro fails to startup due to configuration."""
```

kedro/errors/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+class KedroSessionError(RuntimeError):
+    """
+    Error raised when kedro fails to create a new Session.
+    """
+
+
+class KedroRunnerError(RuntimeError):
+    """
+    Error raised when kedro runner hits a runtime error before all nodes are complete
+    """
```
lorenabalan, line 49:
```suggestion
    """Error raised when kedro runner hits a runtime error before all nodes are complete.
    """
```

kedro/errors/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+class KedroConfigError(RuntimeError):
+    """
+    Error raised when kedro fails to startup due to configuration.
+    """
+
+
+class KedroSessionError(RuntimeError):
+    """
+    Error raised when kedro fails to create a new Session.
+    """
```
lorenabalan, line 43:
```suggestion
    """Error raised when kedro fails to create a new Session."""
```

kedro/errors/__init__.py
```diff
@@ -0,0 +1,49 @@
...
+# or use the QuantumBlack Trademarks in any other manner that might cause
+# confusion in the marketplace, including but not limited to in advertising,
+# on websites, or on software.
+#
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Expose public exceptions and warnings.
+"""
```
lorenabalan, line 31:
```suggestion
"""Expose public exceptions and warnings."""
```

lorenabalan (approved):
This look good to me! I only left a few nit picking suggestions to make the docstrings more aligned with our style guide (Google docstrings).

WaylonWalker:
I just searched and replaced all the new `RuntimeErrors` with the appropriate custom error type.  There are other several other common errors that appear like they could use a custom error type, mostly `ValueErrors`.  A cursory glance many others are `AttributeErrors`, but those seem appropriate to stay like they are.  I feel that it might make most sense to keep the scope of this PR to creating the errors module and replacing `RuntimeErrors`.

@idanov , @lorenabalan would you mind if we moved the existing custom errors into kedro.errors in this PR as well?

image

lorenabalan:
> @idanov , @lorenabalan would you mind if we moved the existing custom errors into kedro.errors in this PR as well?

Personally I'm not against having everything in `kedro.errors`. 😃 We should be careful about moving them though if we do decide to do that, because changing imports paths is a breaking change, it might be cleaner if it happens in separate PRs. 
* one on `master`, that moves the exceptions to the new location, and the old locations import from the new module, potentially with a warning that recommends to import from new paths
* one on `develop`, that removes the backwards-compatible import code from above

WaylonWalker:
@lorenabalan I completely overlooked your message and was waiting for a response for a while.  I started diving back into it this morning.  I moved the `KedroContextError` into the custom errors module this morning.

How do I gracefully depreciate the import of `KedroContextError` in kedro/framework/context/__init__.py.  Do we need to redefine the `KedroContextError` in that module and include a deprecation warning.  I will need to play around to see how that works.

## Existing custom Error todo list

I did search for custom errors that kedro defines and found the following todo list that needs moved.  I will keep working through these.

* DataSetError
* DataSetAlreadyExistsError
* DataSetNotFoundError
* ModularPipelineError
* KedroCliError
* VersionNotFoundError
* OutputNotUniqueError
* ConfirmNotUniqueError
* CircularDependencyError
* MissingConfigException
* BadConfigException(

PR #745, opened 2021-04-05
Adds partition support to SparkHiveDataSet

### LINKED ISSUE #725: Partition as args in SparkHiveDataSet
## Description
I can only partition my data with SparkDataSet and not SparkHiveDataSet whereas I want to save my data in a Hive table, and so use the [_save_validate ](https://github.com/quantumblacklabs/kedro/blob/master/kedro/extras/datasets/spark/spark_hive_dataset.py#L247-L257) method to make sure the columns are OK. 

## Context
Saving my data in Hive without having this extra validation on the schema might be risky. 

## Possible Alternatives
Why not using something similar to SparkDataSet: have a "save_args" in "[\_\_init\_\_](https://github.com/quantumblacklabs/kedro/blob/master/kedro/extras/datasets/spark/spark_hive_dataset.py#L128)" method of SparkHiveDataSet and passing the "partitionBy" item. Then use it in the[ _insert_save](https://github.com/quantumblacklabs/kedro/blob/master/kedro/extras/datasets/spark/spark_hive_dataset.py#L213-L218) method, using the partitionBy in the SQL statement: something like  "INSERT INTO [TABLE] [db_name.]table_name [PARTITION partitionBy] select_statement"

brendalf:
## Description
`SparkHiveDataSet` doesn't not support partitions like `SparkDataSet`. As discussed in #725, @AntonyMilneQB proposed to implement this by adding a new argument `partition_by` inside `SparkHiveDataSet` constructor. 
The `partition_by` argument was meant to be the partition where the data should be inserted, eg: `partition_by="ref = 1"`, to be used later, inside `_save`, like this `insert into {table} partition ({partition_by}) select * from tmp`.
I'm opening this PR as a draft for discussion because I ran over some problems while I was implementing this approach.

Resolves #725 

## Development notes
Here are the problems:
**1. If the table that doesn't exist yet**
In this case, the `SparkHiveDataSet` is going to create with `_create_empty_hive_table(self, data)` but the `partition_by` can be used here straight away because `_create_empty_hive_table` needs a list of names and types to make the partition, something like `PARTITIONED BY (ref integer)`.

I tried to replace the `partition_by` argument with the `partitioned_by` argument, to be used inside the create table, and add a new argument `partition`, that handles the partition where the data should be inserted, to `_save`. But to do that, we would need to change the `save` method of `AbstractDataSet`, so I discarded it.

I also tried to increase the @AntonyMilneQB approach by implementing two arguments to the `SparkHiveDataSet` constructor: `partition` (handles where the data should be inserted) and `partitioned_by` (handles the columns where the table should be partitioned). The problem here is that the `_create_empty_hive_table(self, data)` use CREATE TABLE AS SELECT (CTAS) and we can't use PARTITIONED BY with it. The can solve this by replacing the CTAS with the regular CREATE TABLE, but I don't know how to get the list of name and valid Spark SQL data types from the `data` variable that is passed to `_create_empty_hive_table(self, data)`.

What everyone thinks? Do you know how to implement the create table with partition? Can you see a different approach to solve problem 1?

For now, the implementation reflects the @AntonyMilneQB approach and it only works if the partitioned table is already created.
But that leaves us if the second problem.

**2. How can we validate partitioned tables?**
The `SparkHiveDataSet` uses the `.load()` to generate the `self._table_columns` when the table already exists, but the `load` method loads the table with all columns, including the partitioned by columns, and that makes the `self._validate_save` fails because the user table to insert doesn't have the partition columns.

I tried to get around this by getting the partition columns names and remove these columns from `self._table_columns`, but I don't know if this is the best approach.

What everyone thinks?

## Checklist

## Notice


kedro/extras/datasets/spark/spark_hive_dataset.py
```diff
@@ -151,6 +158,13 @@ def __init__(
         self._database = database
         self._table = table
         self._stage_table = "_temp_" + table
+        self._partition = partition
+
+        # get the name of each partition
+        self._partitions: List[str] = []
+        if self._partition is not None:
+            for partition_set in self._partition.split(","):
+                self._partitions.append(partition_set.split("=")[0].strip())
```
AntonyMilneQB, line 167: Would it make sense to just directly make the function argument `partition: List[str] = None`, as is done for `table_pk`? This would do away with the need for string manipulation here.

kedro/extras/datasets/spark/spark_hive_dataset.py
```diff
@@ -209,8 +226,16 @@ def _save(self, data: DataFrame) -> None:
     def _insert_save(self, data: DataFrame) -> None:
         data.createOrReplaceTempView("tmp")
         columns = ", ".join(self._table_columns)
+
+        partition = ""
+        if self._partition is not None:
+            partition = f"partition ({self._partition.strip()})"
+
```
AntonyMilneQB, line 233: I would probably put this logic in a slightly different place.

e.g. If `partition` argument remains a string then make its default argument `""` and do this in `__init__` method to begin with:
```
self._partition = partition.strip()
```
And then here you'd do:
```
        self._get_spark().sql(
            f"""
            insert into {self._database}.{self._table} partition ({self._partition})
            select {columns} from tmp
            """  # nosec
        )
```

kedro/extras/datasets/spark/spark_hive_dataset.py
```diff
@@ -242,7 +267,10 @@ def _overwrite_save(self, data: DataFrame) -> None:
 
     def _validate_save(self, data: DataFrame):
         hive_dtypes = set(self._load().dtypes)
+        if len(self._partitions) > 0:
```
AntonyMilneQB, line 270:
```suggestion
        if self._partitions:
```

AntonyMilneQB (commented):
Thank you very much for your work here and for detailing where you've got to so well! This looks like a great start. I've left some minor comments, but I'm afraid I don't know enough about spark to consider the problems you've raised. Let me try and get some people who are more knowledgeable about spark to take a look 🙂

kedro/extras/datasets/spark/spark_hive_dataset.py
brendalf: Right, I'll fix this.

kedro/extras/datasets/spark/spark_hive_dataset.py
brendalf: I like these. But, when the table is not partitioned, `partition ({self._partition})` isn't gonna work. Maybe we can move that to the `__init__` too?

kedro/extras/datasets/spark/spark_hive_dataset.py
brendalf: That makes sense, but with just the list of column names we can't identify a partition. Maybe we can make `partition` get a list of tuples, where each position of the list identifies the column name and value. What do you think?

AntonyMilneQB: Ah I see. That could indeed work, but now I'm wondering exactly what the `partition` string in the insert could look like. e.g. we can have `partition_1 = x, partition_2 = y` but we could also just have `partition_1, partition_2` right? Do you know what is the full specification of the syntax of this string?

Now I'm actually also wondering why `SparkHiveDataSet` is using dynamically build SQL queries in the first place rather than the Python API, which would presumably make this sort of thing much easier. Is that impossible when doing the hive thing? @jiriklein

kedro/extras/datasets/spark/spark_hive_dataset.py
AntonyMilneQB: Ah, in that case I would keep `self._partition` as an empty string but still keep the logic for building the SQL query next to the SQL query:
```
        partition =  f"partition ({self._partition})" if self._partition else ""
        self._get_spark().sql(
            f"""
            insert into {self._database}.{self._table} {partition}
            select {columns} from tmp
            """  # nosec
        )
```

jiriklein:
Hi @brendalf - I'm looking at this and I think `SparkHiveDataSet` deserves a proper rewrite to:
1. Align with the wider `kedro.extras.datasets` API, especially with `SparkDataSet`
2. Avoid as many SQL query string injections as possible
I'll be looking at this in the next couple of days so please let me know if you have any questions or concerns.

jiriklein:
Hi @brendalf 
After some research, my proposal is to replace the current `_save()` functionality with the following:
```
def _create_hive_table(self, data: DataFrame, mode: str = None):
    _mode: str = mode or self._write_mode
    data.write.saveAsTable(
        self._full_table_address,
        mode=_mode,
        format=self._format,
        partitionBy=self._partition_by,
        **self._save_args,
    )
```
And removing the `StagedDataSet` altogether. This then fully supports your use-case for leveraging `partitionBy`. Does this make sense to you and your use-case?

kedro/extras/datasets/spark/spark_hive_dataset.py
brendalf: Hi @AntonyMilneQB, sorry for the late reply.
That's right, the `partition` string will be something like `partition_1 = x, partition_2 = y` if the partition type is static or `partition_1, partition_2` if the partition is dynamic.

brendalf:
Hi @jiriklein, how are you?
Sorry for the late reply.

I agree with you that `SparkHiveDataSet` deserves a proper rewrite and your proposal LGTM.

jiriklein:
Hi @brendalf, hope you're well!
`SparkHiveDataSet` has now been fully rewritten and `partitionBy` support has been added, including access to other `save_args`. 
You can find the changes in the latest `develop` code. If you wish to wait for a proper release, you can expect these changes to materialise in version `0.18.0`. 
Hope this helps!

brendalf:
Perfect!

WaleedxxxxjJz:
> ## Description
> 
> `SparkHiveDataSet` doesn't not support partitions like `SparkDataSet`. As discussed in #725, @AntonyMilneQB proposed to implement this by adding a new argument `partition_by` inside `SparkHiveDataSet` constructor. 
> 
> The `partition_by` argument was meant to be the partition where the data should be inserted, eg: `partition_by="ref = 1"`, to be used later, inside `_save`, like this `insert into {table} partition ({partition_by}) select * from tmp`.
> 
> I'm opening this PR as a draft for discussion because I ran over some problems while I was implementing this approach.
> 
> 
> 
> Resolves #725 
> 
> 
> 
> ## Development notes
> 
> Here are the problems:
> 
> **1. If the table that doesn't exist yet**
> 
> In this case, the `SparkHiveDataSet` is going to create with `_create_empty_hive_table(self, data)` but the `partition_by` can be used here straight away because `_create_empty_hive_table` needs a list of names and types to make the partition, something like `PARTITIONED BY (ref integer)`.
> 
> 
> 
> I tried to replace the `partition_by` argument with the `partitioned_by` argument, to be used inside the create table, and add a new argument `partition`, that handles the partition where the data should be inserted, to `_save`. But to do that, we would need to change the `save` method of `AbstractDataSet`, so I discarded it.
> 
> 
> 
> I also tried to increase the @AntonyMilneQB approach by implementing two arguments to the `SparkHiveDataSet` constructor: `partition` (handles where the data should be inserted) and `partitioned_by` (handles the columns where the table should be partitioned). The problem here is that the `_create_empty_hive_table(self, data)` use CREATE TABLE AS SELECT (CTAS) and we can't use PARTITIONED BY with it. The can solve this by replacing the CTAS with the regular CREATE TABLE, but I don't know how to get the list of name and valid Spark SQL data types from the `data` variable that is passed to `_create_empty_hive_table(self, data)`.
> 
> 
> 
> What everyone thinks? Do you know how to implement the create table with partition? Can you see a different approach to solve problem 1?
> 
> 
> 
> For now, the implementation reflects the @AntonyMilneQB approach and it only works if the partitioned table is already created.
> 
> But that leaves us if the second problem.
> 
> 
> 
> **2. How can we validate partitioned tables?**
> 
> The `SparkHiveDataSet` uses the `.load()` to generate the `self._table_columns` when the table already exists, but the `load` method loads the table with all columns, including the partitioned by columns, and that makes the `self._validate_save` fails because the user table to insert doesn't have the partition columns.
> 
> 
> 
> I tried to get around this by getting the partition columns names and remove these columns from `self._table_columns`, but I don't know if this is the best approach.
> 
> 
> 
> What everyone thinks?
> 
> 
> 
> ## Checklist
> 
> 
> 
> - [x] Read the [contributing](https://github.com/quantumblacklabs/kedro/blob/master/CONTRIBUTING.md) guidelines
> 
> - [x] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
> 
> - [x] Updated the documentation to reflect the code changes
> 
> - [x] Added a description of this change and added my name to the list of supporting contributions in the [`RELEASE.md`](https://github.com/quantumblacklabs/kedro/blob/master/RELEASE.md) file
> 
> - [x] Added tests to cover my changes
> 
> 
> 
> ## Notice
> 
> 
> 
> - [x] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":
> 
> 
> 
> - I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
> 
> - I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorised to submit this contribution on behalf of the original creator(s) or their licensees.
> 
> - I certify that the use of this contribution as authorised by the Apache 2.0 license does not violate the intellectual property rights of anyone else.
> 
>

PR #782, opened 2021-06-07
Added smb protocol with test

lucasjamar:
This solves https://github.com/quantumblacklabs/kedro/issues/765

## Description
Added else for filesystems that contain "netloc" but are not cloud protocols.

## Development notes
Added a simple test to ensure the username, password, host and port are returned in the path of smb files

## Checklist

## Notice

### DIFF
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -57,7 +57,7 @@ VERSION_KEY = "version"
 HTTP_PROTOCOLS = ("http", "https")
 PROTOCOL_DELIMITER = "://"
 CLOUD_PROTOCOLS = ("s3", "gcs", "gs", "adl", "abfs")
-SMB_PROTOCOLS = ("smb")
+SMB_PROTOCOLS = "smb"
 
 
 class DataSetError(Exception):
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -57,7 +57,7 @@ VERSION_KEY = "version"
 HTTP_PROTOCOLS = ("http", "https")
 PROTOCOL_DELIMITER = "://"
 CLOUD_PROTOCOLS = ("s3", "gcs", "gs", "adl", "abfs")
-SMB_PROTOCOLS = "smb"
+SMB_PROTOCOLS = ("smb")
 
 
 class DataSetError(Exception):
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -57,6 +57,7 @@ VERSION_KEY = "version"
 HTTP_PROTOCOLS = ("http", "https")
 PROTOCOL_DELIMITER = "://"
 CLOUD_PROTOCOLS = ("s3", "gcs", "gs", "adl", "abfs")
+SMB_PROTOCOLS = ("smb")
 
 
 class DataSetError(Exception):
@@ -682,7 +683,7 @@ def _parse_filepath(filepath: str) -> Dict[str, str]:
             host_with_port = parsed_path.netloc.rsplit("@", 1)[-1]
             host = host_with_port.rsplit(":", 1)[0]
             options["path"] = host + options["path"]
-        else:
+        elif protocol in SMB_PROTOCOLS:
             options["path"] = parsed_path.netloc + options["path"]
 
     return options
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -57,7 +57,7 @@ VERSION_KEY = "version"
 HTTP_PROTOCOLS = ("http", "https")
 PROTOCOL_DELIMITER = "://"
 CLOUD_PROTOCOLS = ("s3", "gcs", "gs", "adl", "abfs")
-SMB_PROTOCOLS = ("smb")
+SMB_PROTOCOLS = ("smb", )
 
 
 class DataSetError(Exception):
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -57,7 +57,7 @@ VERSION_KEY = "version"
 HTTP_PROTOCOLS = ("http", "https")
 PROTOCOL_DELIMITER = "://"
 CLOUD_PROTOCOLS = ("s3", "gcs", "gs", "adl", "abfs")
-SMB_PROTOCOLS = "smb"
+SMB_PROTOCOLS = ("smb", )
 
 
 class DataSetError(Exception):
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -682,6 +682,8 @@ def _parse_filepath(filepath: str) -> Dict[str, str]:
             host_with_port = parsed_path.netloc.rsplit("@", 1)[-1]
             host = host_with_port.rsplit(":", 1)[0]
             options["path"] = host + options["path"]
+        else:
+            options["path"] = parsed_path.netloc + options["path"]
 
     return options
 
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -57,7 +57,7 @@ VERSION_KEY = "version"
 HTTP_PROTOCOLS = ("http", "https")
 PROTOCOL_DELIMITER = "://"
 CLOUD_PROTOCOLS = ("s3", "gcs", "gs", "adl", "abfs")
-SMB_PROTOCOLS = ("smb", )
+SMB_PROTOCOLS = "smb"
 
 
 class DataSetError(Exception):
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -19,6 +19,7 @@
 * `EmailMessageDataSet` added to doctree.
 * When node inputs do not pass validation, the error message is now shown as the most recent exception in the traceback ([Issue #761](https://github.com/quantumblacklabs/kedro/issues/761)).
 * `kedro pipeline package` now only packages the parameter file that exactly matches the pipeline name specified and the parameter files in a directory with the pipeline name.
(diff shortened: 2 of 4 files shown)


lucasjamar:
It appears the unit tests are failing because they are timing out the docs_linkcheck is failing because of some url on documentation about AWS that no longer exists.
Should I modify the url related to AWS?

Please let me know if you have any questions. thanks!

kedro/io/core.py
```diff
@@ -57,6 +57,7 @@
 HTTP_PROTOCOLS = ("http", "https")
 PROTOCOL_DELIMITER = "://"
 CLOUD_PROTOCOLS = ("s3", "gcs", "gs", "adl", "abfs")
+SMB_PROTOCOLS = "smb"
```
AntonyMilneQB, line 60:
```suggestion
SMB_PROTOCOLS = ("smb", )
```

AntonyMilneQB: Also, will there be other protocols in addition to samba which should have paths parsed this way? i.e. should this variable be something more generic than `SMB_PROTOCOLS`? Like `NETWORK_PROTOCOLS` or something?

AntonyMilneQB (commented):
This looks great to me, thank you for your contribution @lucasjamar! I'm not very familiar with these protocols so will try and get some people who are more familiar to take a look.

Please could you explain to a dummy why the whole `username:password@myhost:8080` get passed as a path here when it doesn't for s3 etc.?

AntonyMilneQB:
> It appears the unit tests are failing because they are timing out and the docs_linkcheck is failing because of some url on documentation about AWS that no longer exists.
> Should I modify the url related to AWS?
We're fixing the broken link in a different PR already so don't worry about it - when it's merged to master you can update this PR and it should fix it. And the unit tests timing out is probably just a random error, so re-rerunning the job (which will happen if you do another commit) will probably fix it. So probably nothing to worry about here 🙂

kedro/io/core.py
lucasjamar: I used to work in the past with FTP protocols but not anymore but I can imagine that they will need to parsed in a similar way:
https://filesystem-spec.readthedocs.io/en/latest/_modules/fsspec/implementations/ftp.html
There is a whole bunch of other protocols such as git file system but I dont know if they are that useful to Kedro:
https://filesystem-spec.readthedocs.io/en/latest/_modules/index.html

lucasjamar: > 
> 
> This looks great to me, thank you for your contribution @lucasjamar! I'm not very familiar with these protocols so will try and get some people who are more familiar to take a look.
> 
> Please could you explain to a dummy why the whole `username:password@myhost:8080` get passed as a path here when it doesn't for s3 etc.?

I dont know why this doesnt happen for S3. This would be a question for the author of the original code. In general, I dont see what the _parse_filepath is for since it splits the path into a protocol and path and then feeds that into fsspec. This is something that fsspec handles itself but it additionally parses the host, username, port, etc so in the case of Kedro's _parse_filepath function, stuff like the host, port, username, etc get lost. Its even noted in the _parse_filepath code that it has the same functionality as `fsspec.utils.infer_storage_options` so I dont see why that function is not used instead.
To be clear, my current implementation is not absolutely necessary. There is a workaround.  `urlsplit` considers the first string after `smb://` as the network location. Kedro then excludes that from the path (why, Im guessing because of other filesystem storages such as http or hdfs?). So my workaround has been to provide a dummy netloc and feed stuff like the host, port, username and password via fs_args. Here is an example:
```
raw_iris:
  type: pandas.CSVDataSet
  filepath: smb://dummy_url/folder/iris.csv
  fs_args:
    host: <hostname>
    port: 8080
```

However it feels clumsy and it does beg the point of why the netlwork location was removed in the first place.

Thanks for your help @AntonyMilneQB !

lucasjamar: > 
> 
> I believe @AntonyMilneQB 's code suggestion is still a valid one that should be addressed.

Sorry @lorenabalan ! I had integrated the parentheses, but then I applied black which automatically removed them. I just fixed it again

lucasjamar:
> 
> 
> This looks great to me, thank you for your contribution @lucasjamar! I'm not very familiar with these protocols so will try and get some people who are more familiar to take a look.
> 
> Please could you explain to a dummy why the whole `username:password@myhost:8080` get passed as a path here when it doesn't for s3 etc.?

To follow-up on your question @AntonyMilneQB , id be interested to know why the host and port are ignored for hdfs if someone knows the answer. I would have thought these are important pieces of information.

image

kedro/io/core.py
AntonyMilneQB: The problem is that you're missing a trailing comma - it should be:
```
SMB_PROTOCOLS = ("smb", )
```
Notice the `,` before `)`.

What you currently have is this:
```
SMB_PROTOCOLS = ("smb")
```
... which is interpreted by Python as if the parentheses aren't there at all (and so black strips them out). This sets `SMB_PROTOCOLS` to the string `smb`. Including the comma sets it to a tuple with one element (the string `smb`). This means that looking up membership (`if protocol in SMB_PROTOCOLS`) will work. 🙂

AntonyMilneQB: I'm going to try to look your questions about fsspec and the way we're handling different protocols tomorrow btw - sorry for the delay, it's been a busy couple of days!

limdauto:
@lucasjamar @AntonyMilneQB @lorenabalan I think generally our strategy when parsing filepath and instantiate the underlying `fsspec.filesystem` is:

* Parse `path` and `protocol` from `filepath` attribute of the dataset. The reason we have our own `_parse_path` function and not use the underlying `fsspec.utils.infer_storage_options` was due to how `infer_storage_options` was treating `abl/adfs` differently from `s3/gcs`: https://github.com/dask/adlfs/issues/45. So to ensure parsing consistency, we roll our own `_parse_path`. Since this deals with `path` and `protocol` only, everything else is discarded.
* `host` & `port` are added to `fs_args` as @lucasjamar has noticed.
* `username` & `password` should be added to `credentials`.

When we instantiate `fsspec.filesystem`, e.g. [in a dataset](https://github.com/quantumblacklabs/kedro/blob/master/kedro/extras/datasets/json/json_dataset.py#L115), we will pass everything as `kwargs` to `fsspec`, so the net effect would be similar to passing a string containing everything to fsspec and allow fsspec to parse it with `infer_storage_options`.

What I'd suggest is maybe looking into why you need to specify a dummy url and fix that instead so you can write:

```
filepath: smb://my/file/path/
fs_args:
    host:
    port:
credentials: my_smb_dataset_credentials
```

AntonyMilneQB:
Thank you very much for this @limdauto! Super helpful 👍  Just for completeness, when I was looking to this earlier I also found the [original commit](https://github.com/quantumblacklabs/kedro/commit/e178038df16f6219186afc81dd6e37c2c19c462a) where we moved from fsspec's `infer_storage_options` to our own implementation. Looking through `infer_storage_options`, it is indeed very similar to our `_parse_path`.

lucasjamar:
> 
> 
> @lucasjamar @AntonyMilneQB @lorenabalan I think generally our strategy when parsing filepath and instantiate the underlying `fsspec.filesystem` is:
> 
>     * Parse `path` and `protocol` from `filepath` attribute of the dataset. The reason we have our own `_parse_path` function and not use the underlying `fsspec.utils.infer_storage_options` was due to how `infer_storage_options` was treating `abl/adfs` differently from `s3/gcs`: [dask/adlfs#45](https://github.com/dask/adlfs/issues/45). So to ensure parsing consistency, we roll our own `_parse_path`. Since this deals with `path` and `protocol` only, everything else is discarded.
> 
>     * `host` & `port` are added to `fs_args` as @lucasjamar has noticed.
> 
>     * `username` & `password` should be added to `credentials`.
> 
> 
> When we instantiate `fsspec.filesystem`, e.g. [in a dataset](https://github.com/quantumblacklabs/kedro/blob/master/kedro/extras/datasets/json/json_dataset.py#L115), we will pass everything as `kwargs` to `fsspec`, so the net effect would be similar to passing a string containing everything to fsspec and allow fsspec to parse it with `infer_storage_options`.
> 
> What I'd suggest is maybe looking into why you need to specify a dummy url and fix that instead so you can write:
> 
> ```
> filepath: smb://my/file/path/
> fs_args:
>     host:
>     port:
> credentials: my_smb_dataset_credentials
> ```

@limdauto Thanks a lot for the helpful answer! As mentioned in an earlier comment, `urlsplit` interprets the first part of the path after `://` as the network location even though it belongs to the path. For this reason, this information gets lost. Unfortunately, I dont see any alternative than the dummy url I mentioned or the implementation proposed. Or am I getting this wrong and there is a better solution? Thanks :)

limdauto:
@lucasjamar so I think in this case `smb` belongs similar to the the `file` protocol, in which case instead of the dummy URL, you can use triple forward slashes:

```python
In [2]: _parse_filepath("smb:///test-path")
Out[2]: {'protocol': 'smb', 'path': '/test-path'}
```

This is consistent with how urllib treats `file:` protocol, i.e.

```python
In [5]: urlsplit("file:///foo")
Out[5]: SplitResult(scheme='file', netloc='', path='/foo', query='', fragment='')

In [6]: urlsplit("file://foo")
Out[6]: SplitResult(scheme='file', netloc='foo', path='', query='', fragment='')
```

AntonyMilneQB:
Amazing, thanks very much @limdauto.

@lucasjamar I think this means we don't need this PR any more?

PR #820, opened 2021-07-05
pandas.XMLDataSet

lucasjamar:
## Description
Offer a XMLDataSet to read/ write data from/to a XML file into a pandas df.
https://github.com/quantumblacklabs/kedro/issues/818

## Development notes
Simply modified pandas version to 1.3.0 to see failures resulting from tests.

## Checklist

## Notice


lucasjamar:
Would require dropping python 3.6. Python 3.6 was dropped since pandas 1.2.0
https://github.com/pandas-dev/pandas/pull/35214

ignacioparicio:
Hi @lucasjamar and thank you for opening this PR! I'd suggest to repeat the pandas 1.3.0 test branching off `develop` instead of off `master`. In our `develop` branch (which contains breaking changes and will eventually be released as Kedro 0.18.0), we have already dropped Python 3.6 and added support for pandas~=1.2 (see [this PR already merged](https://github.com/quantumblacklabs/kedro/commit/4f5f9c17bb8daf36a57048982c0b40d7a21fd3a0#diff-575849fb1352ea602031ab76cc6fde4cacdcaf1120dc3525da51a6b8b9a22e84R5)).

If CI passes with pandas 1.3.0 in `develop`, we'd be happy to open up the range to `pandas>=1.2, <1.4`, and subsequently add support for `XMLDataSet` through your PR. The only downside would be that the time to be released would be a bit longer than if we merged to `master`.

Thanks again for looking into this!

lorenabalan:
The existing range `~=1.2` would still include 1.3 so there's no need to update from that perspective. It's the equivalent of `>=1.2.0, <2.0`. If we want to add this dataset, because it's only available in 1.3, we would have to bump the minimum version of `pandas`, so `~=1.3`.

lucasjamar:
Closing to branch off dev

PR #840, opened 2021-07-20
Fix broken module style in html doc

Nieju:
This PR fixes the following in the HTML doc

### DIFF
--- a/docs/source/css/qb1-sphinx-rtd.css
+++ b/docs/source/css/qb1-sphinx-rtd.css
@@ -475,6 +475,4 @@ footer span.commit .rst-content tt,
     display: block;
     content: ' ';
     width: 1px;
-    height: 60px;
-    margin-top: -80px;
 }


ignacioparicio:
Thank you @Nieju for the contribution! Much appreciated 😄 

There's [an old, closed PR](https://github.com/quantumblacklabs/kedro/pull/603) tackling this very issue. @mzjp2 (given you opened the old PR) @limdauto (given you approved the old PR) do you think we can merge this? It's not clear to me what was the outcome of the discussion with @richardwestenra.

lorenabalan (approved):
Approving but @mzjp2 & @limdauto would be good to hear if this jogs your memory.

mzjp2:
No qualms, I had to close my PR because I had deleted my Kedro fork and could no longer "update from master" :)

docs/source/css/qb1-sphinx-rtd.css
```diff
@@ -475,6 +475,4 @@ footer span.commit .rst-content tt,
     display: block;
     content: ' ';
     width: 1px;
-    height: 60px;
-    margin-top: -80px;
```
mzjp2, line 479: As per Richard's comment though:

> The best way to fix this is to make both height and margin-top to have the same magnitude, e.g. 60px and -60px, or 80px and -80px, so that they cancel each other out.

Nieju:
Looks like this simple change will "never" hit the road. Should I close it?

PR #862, opened 2021-08-21
Add CITATION.cff file to the repository

mzjp2:
## Description

GitHub [released a [new feature](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-citation-files) where they read `CITATION.cff` files at the root of your repository in the default branch and creates a citation button/section on your repository (see below for an example)

image

or head to https://github.com/mzjp2/kedro to see what it looks like.

## Checklist

## Notice


CITATION.cff
```diff
@@ -0,0 +1,63 @@
...
+- family-names: "Nijholt"
+  given-names: "Bas"
+- family-names: "Vukolov"
+  given-names: "Dmitry"
+- family-names: "Klein"
+  given-names: "Jiri"
+- family-names: "Fischer"
+  given-names: "Kody"
+- family-names: "Koops"
+  given-names: "Marnix"
```
lorenabalan, line 59: How did you generate the list of names? Some of these names I don't recognise at all, presumably because they were contributors at some point. Probably best to include only the main team?

CITATION.cff
```diff
@@ -0,0 +1,63 @@
...
+title: "Kedro"
+version: "0.17.4"
```
lorenabalan, line 61: Would we have to update the version here every time we release?

lorenabalan (commented):
Pretty cool, just had a few questions. Also is there any point in keeping both this file and the Zenodo reference?

CITATION.cff
mzjp2: You could remove the `version` field from here which detracts slightly from the generated citation (no version after Kedro) if you'd prefer to not have to maintain this, wasn't sure which you would prefer given that the release process already includes replacing versions across the documentation (unless that has changed?)

CITATION.cff
mzjp2: I got them from the Zenodo reference that I think @yetudada populated. Also think the main team is probably better!

mzjp2:
> This is really cool! I saw this feature on Twitter. I have two queries:
> 
> * How does this work? Do we need to update the authors every release, similar to how we acknowledge our contributors on the Release notes?
> * And we would not need Zenodo, I would remove the integration and just use this.

It depends on how thorough you want the authors list to be. i would imagine that if you want to put down just the main team, then you wouldn't need to update the list untill you have another core contributor or a contributor that passes some threshold to be referenced as a collaborator on the citation.

Agreed on ditching Zenodo!

AntonyMilneQB:
Thanks for this @mzjp2. Superseded by #1012.

PR #874, opened 2021-09-04
implement __all__ in __init__ modules

WaylonWalker:
## Description

Reducing the number of `# NOQA` comment strings found in the entire repo from 77 to 9 (69 to 0 in the kedro source code).

## Development notes

Added `__all__` list in every `__init__` module.

## Checklist

## Notice


WaylonWalker:
Looks like one of your linters may not like this change after all, I cannot see the circle ci errors, please let me know your thoughts.

AntonyMilneQB:
Seems like the problem is that `isort` wants to change, e.g.
```
from .config import BadConfigException
from .config import ConfigLoader
from .config import MissingConfigException
```

to

```
from .config import BadConfigException, ConfigLoader, MissingConfigException
```

This should fix it:
```
isort kedro/framework/context/__init__.py
isort kedro/config/__init__.py
isort kedro/runner/__init__.py
isort kedro/io/__init__.py
```

As ever, thank you very much for the contribution! 🙂

AntonyMilneQB (approved):

WaylonWalker:
That makes much more sense @AntonyMilneQB.  Made the change.


Would you be open to this same change in the docs, this a much different side of the coin.  Honestly, it feels more important as this would be teaching others the more "correct" way to do things.  I also get the argument that it may be yet another thing for folks to understand.

WaylonWalker:
@AntonyMilneQB It looks like there was a segfault in unit_test_36, but all the other versions passed.  This seems a bit weird, are you familiar with this?

AntonyMilneQB:
> Would you be open to this same change in the docs, this a much different side of the coin. Honestly, it feels more important as this would be teaching others the more "correct" way to do things. I also get the argument that it may be yet another thing for folks to understand.

What do you mean by the same change in the docs? Some on [the page on `kedro lint`](https://kedro.readthedocs.io/en/stable/09_development/04_lint.html) which includes `isort`? Or are you talking about the docs for contributors?

> @AntonyMilneQB It looks like there was a segfault in unit_test_36, but all the other versions passed. This seems a bit weird, are you familiar with this?

Argh yes, I am getting this on one of my PRs actually. As you saw it's usually fixed by rerunning a couple of times, but I've been getting it a lot recently so we might need to try and actually fix whatever the underlying problem is.

WaylonWalker:
> What do you mean by the same change in the docs? Some on the page on kedro lint which includes isort? Or are you talking about the docs for contributors?


I am talking about the changes in the PR, i.e. replacing `#noqa` with `__all__`.

AntonyMilneQB:
Sorry, you've lost me 😅   Where in the docs do we do `#noqa`?

WaylonWalker:
These two pages use `#noqa`

* [aws_batch](https://kedro.readthedocs.io/en/stable/10_deployment/07_aws_batch.html?highlight=aws%20batch)
* [create_pipelines](https://kedro.readthedocs.io/en/stable/03_tutorial/04_create_pipelines.html)

AntonyMilneQB:
Ah, I see. Thanks for showing it. I didn't see these before because I was searching for `#noqa` without a space 🤦 

I would leave these instances as they are. You are right that they also appear in starters, and this is because it is part of the modular pipeline template (i.e. if you do `kedro pipeline create`, you will get this line). So changing it in the docs would mean also changing it in other places that does indeed increase the complexity of a new template slightly. I do agree with you about demonstrating best practices, but I suspect that in the not too distant future these lines would change from relative to absolute imports anyway (thanks to new aliasing capabilities in `kedro pipeline package/pull`). So overall I'd leave all the other instances 🙂

MerelTheisenQB (approved):
Thanks for the contribution @WaylonWalker 🏅

idanov (commented):
Awesome idea, thanks @WaylonWalker 🎉

PR #875, opened 2021-09-04
Reduce type errors

WaylonWalker:
## Description

Working through some of the mypy type-checking errors.

## Development notes

In order to keep a reasonable scope I globbed for .py files in certain directories, and created the following checklist.  The only error still remaining in these modules with the current configuration is `found module but no type hints or library stubs`

Taking Anthony Sotile's approach to [gradual typing](https://www.youtube.com/watch?v=Rk-Y71P_9KE&t=15s), this phase will eliminate most type checking errors with a default mypy config.

## Checklist

## Notice

### DIFF
--- a/kedro/framework/cli/pipeline.py
+++ b/kedro/framework/cli/pipeline.py
@@ -722,10 +722,10 @@ def _get_wheel_name(**kwargs: Any) -> str:
     # https://stackoverflow.com/q/51939257/3364156
     dist = Distribution(attrs=kwargs)
     bdist_wheel_cmd = dist.get_command_obj("bdist_wheel")
-    bdist_wheel_cmd.ensure_finalized()
+    bdist_wheel_cmd.ensure_finalized()  # type: ignore
 
-    distname = bdist_wheel_cmd.wheel_dist_name
-    tag = "-".join(bdist_wheel_cmd.get_tag())
+    distname = bdist_wheel_cmd.wheel_dist_name  # type: ignore
+    tag = "-".join(bdist_wheel_cmd.get_tag())  # type: ignore
     return f"{distname}-{tag}.whl"
 
 
--- a/kedro/framework/cli/pipeline.py
+++ b/kedro/framework/cli/pipeline.py
@@ -887,8 +887,8 @@ def _generate_wheel_file(
         temp_dir_path = Path(temp_dir).resolve()
 
         project = Project(temp_dir_path)  # project where to do refactoring
-        _refactor_code_for_package(  # type: ignore
-            project, package_source, tests_source, alias, metadata
+        _refactor_code_for_package(
+            project, package_source, tests_source, alias, metadata  # type: ignore
         )
         project.close()
 
--- a/kedro/framework/cli/pipeline.py
+++ b/kedro/framework/cli/pipeline.py
@@ -877,7 +877,7 @@ def _generate_wheel_file(
     pipeline_name: str,
     destination: Path,
     source_paths: Tuple[_SourcePathType, ...],
-    version: str,
+    version: Optional[str],
     metadata: ProjectMetadata,
     alias: str = None,
 ) -> None:
@@ -933,7 +933,10 @@ def _generate_wheel_file(
 
 
 def _generate_setup_file(
-    package_name: str, version: str, install_requires: List[str], output_dir: Path
+    package_name: str,
+    version: Optional[str],
+    install_requires: List[str],
+    output_dir: Path,
 ) -> Path:
     setup_file = output_dir / "setup.py"
     package_data = {
--- a/kedro/framework/cli/pipeline.py
+++ b/kedro/framework/cli/pipeline.py
@@ -693,12 +693,13 @@ def _package_pipeline(  # pylint: disable=too-many-arguments
     # Check that pipeline directory exists and not empty
     _validate_dir(artifacts_to_package.pipeline_dir)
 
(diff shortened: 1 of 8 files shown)


WaylonWalker:
Let me know if this is already being worked, or not a desired change.

AntonyMilneQB:
We don't have any ticket to do this currently, and I see no reason why we wouldn't want to do it (everyone else - please do say if this isn't the case though!). Thank you 🙂

WaylonWalker:
Tests are passing, but unit_tests_36, take too long.

kedro/framework/cli/pipeline.py
```diff
@@ -876,7 +876,7 @@ def _generate_wheel_file(
     pipeline_name: str,
     destination: Path,
     source_paths: Tuple[_SourcePathType, ...],
-    version: str,
+    version: Optional[str],
```
AntonyMilneQB, line 879: I don't understand this change and the one to the signature of `_generate_setup_file`. When I run `mypy` on this file I don't get any errors here, and it looks like you're changing it to put an optional argument in front of a compulsory one which shouldn't be allowed?
```
    def f(a=1, b): pass
         ^
SyntaxError: non-default argument follows default argument
```

AntonyMilneQB (commented):
Hi @WaylonWalker, thanks very much for doing this and sorry it's taken so long to review this! I'm a bit of a mypy newbie (thanks for linking to that video) but personally I'm not a big fan of changes to variable names just so that we can remove the `allow-redefinition` option (at least I imagine that's the motivation?).

e.g.
```
def _local_exists(filepath: str) -> bool:  # SKIP_IF_NO_SPARK
    path = Path(filepath)
```
is more confusing to me than just leaving it as `filepath`, since I need to look through the code and figure out if there's a difference between `path` and `filepath`.

Apart from the renamings and the other comment on function signature I like the changes though, thank you! 👍  Let me try and get some others to give their opinions on this since I guess it's largely a matter of personal taste.

kedro/runner/parallel_runner.py
```diff
@@ -79,11 +79,11 @@ def save(self, data: Any):
             # Checks if the error is due to serialisation or not
             try:
                 pickle.dumps(data)
-            except Exception as exc:  # SKIP_IF_NO_SPARK
+            except Exception as inner_exc:  # SKIP_IF_NO_SPARK
                 raise DataSetError(
                     f"{str(data.__class__)} cannot be serialized. ParallelRunner "
                     "implicit memory datasets can only be used with serializable data"
-                ) from exc
+                ) from inner_exc
```
lorenabalan, line 86: L88 can just be `raise` then. I am not sure why we even have the `else` branch at all here, but that's maybe out of scope for this PR.

kedro/framework/cli/pipeline.py
lorenabalan: I don't think this is correct, because on L697 we make sure we don't pass `None` around

```python
version = version or _get_default_version(metadata, pipeline_name)
```

So the type hint should be `str` only.

As for
> it looks like you're changing it to put an optional argument in front of a compulsory one which shouldn't be allowed?

`Optional[str]` is just the equivalent of `Union[str, None]`. It's not the same as an optional argument, i.e. kwargs. That's denoted by giving it a default value, like we do 2 lines below `alias: str = None`, rather than by the type hint.

kedro/framework/cli/pipeline.py
```diff
@@ -932,7 +932,10 @@ def _generate_wheel_file(
 
 
 def _generate_setup_file(
-    package_name: str, version: str, install_requires: List[str], output_dir: Path
+    package_name: str,
+    version: Optional[str],
```
lorenabalan, line 936: Same here, I thought it should be only `str`.

kedro/pipeline/pipeline.py
```diff
@@ -160,36 +160,39 @@ def __init__(
                 "`nodes` argument of `Pipeline` is None. It must be an "
                 "iterable of nodes and/or pipelines instead."
             )
-        nodes = list(nodes)  # in case it's a generator
+        # in case it's a generator explicityly make it a list
+        # Now it can be a list of pipelines or Nodes
+        nodes_pipeline_list = list(nodes)
         _validate_duplicate_nodes(nodes)
```
lorenabalan, line 166:
```suggestion
        _validate_duplicate_nodes(nodes_pipeline_list)
```
If it's a generator, it'd be empty here because it gets consumed on the line above. This is why I prefer leaving it as it was. 😂

lorenabalan (commented):
I agree with @AntonyMilneQB on `--allow-redefinition`. I'm often quite happy to redefine the same variable rather than pile more stuff in memory, as long as the code is still readable - which I think in most cases it already is.

I generally struggle the understand the motivation behind this PR.  Why would anyone want to run `mypy` on its own, with different config to what we do in `precommit`, therefore not quite as the "styleguide"? (by the way we removed `mypy` from test requirements and solely run it via precommit mirror)

kedro/framework/cli/pipeline.py
WaylonWalker: Not sure what I was thinking on this one, looking at the commit message, it looks geared towards the line below that is `alias: str = None`

WaylonWalker:
I must have my editor mis-configured, everytime I open up a kedro module it lights up with mypy errors. I'll read through the docs to get your mypy config setup. Apologies for confusion and not seeing your mypy config inside pre-commit.

PR #889, opened 2021-09-10
Make `_IsSubclassValidator` work with new Dynaconf

deepyaman:
## Description
Kedro breaks with `dynaconf~=3.1.6`, because the subclassed validator doesn't accept `only`/`exclude` arguments.

## Checklist

## Notice

### DIFF
--- a/kedro/framework/project/__init__.py
+++ b/kedro/framework/project/__init__.py
@@ -58,10 +58,9 @@ class _IsSubclassValidator(Validator):
         self,
         settings: Any,
         env: Optional[str] = None,
-        only: Optional[Union[str, Sequence]] = None,
-        exclude: Optional[Union[str, Sequence]] = None,
+        **kwargs: Any,
     ) -> None:
-        super()._validate_items(settings, env, only, exclude)
+        super()._validate_items(settings, env, **kwargs)
 
         default_class = self.default(settings, self)
         for name in self.names:
--- a/kedro/framework/project/__init__.py
+++ b/kedro/framework/project/__init__.py
@@ -31,7 +31,7 @@ configure a Kedro project and access its settings."""
 import importlib
 import operator
 from collections.abc import MutableMapping
-from typing import Any, Dict, Optional, Sequence, Union
+from typing import Any, Dict, Optional
 from warnings import warn
 
 from dynaconf import LazySettings
--- a/kedro/framework/project/__init__.py
+++ b/kedro/framework/project/__init__.py
@@ -54,8 +54,14 @@ def _get_default_class(class_import_path):
 class _IsSubclassValidator(Validator):
     """A validator to check if the supplied setting value is a subclass of the default class"""
 
-    def _validate_items(self, settings, env=None):
-        super()._validate_items(settings, env)
+    def _validate_items(
+        self,
+        settings: Any,
+        env: Optional[str] = None,
+        only: Optional[Union[str, Sequence]] = None,
+        exclude: Optional[Union[str, Sequence]] = None,
+    ) -> None:
+        super()._validate_items(settings, env, only, exclude)
 
         default_class = self.default(settings, self)
         for name in self.names:
--- a/kedro/framework/project/__init__.py
+++ b/kedro/framework/project/__init__.py
@@ -54,11 +54,11 @@ def _get_default_class(class_import_path):
 class _IsSubclassValidator(Validator):
     """A validator to check if the supplied setting value is a subclass of the default class"""
 
-    def _validate_items(
+    def _validate_items(  # pylint: disable=arguments-differ; work with `dynaconf<3.1.6`
         self,
         settings: Any,
         env: Optional[str] = None,
-        **kwargs: Any,  # pylint: disable=arguments-differ; work with ``dynaconf<3.1.6``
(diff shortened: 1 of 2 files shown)


kedro/framework/project/__init__.py
```diff
@@ -54,8 +54,14 @@ def validator_func(settings, validators):
...
-    def _validate_items(self, settings, env=None):
-        super()._validate_items(settings, env)
+    def _validate_items(
+        self,
+        settings: Any,
+        env: Optional[str] = None,
+        only: Optional[Union[str, Sequence]] = None,
+        exclude: Optional[Union[str, Sequence]] = None,
+    ) -> None:
+        super()._validate_items(settings, env, only, exclude)
```
AntonyMilneQB, line 64: Do you think we should just be doing `**kwargs` in case dynaconf change the signature of this again in the future?

AntonyMilneQB: Ideally we wouldn't be overwriting a protected method here at all I guess, but Lim said that was how they recommended doing this... Would be interesting to know what their recommendation is now given that they changed this signature 😬

deepyaman: Yeah, I'm ambivalent. I considered doing `**kwargs` too, happy to change it to that if you all agree that's better. :)

deepyaman:
Also, any suggestions on how to test this? I don't see an existing test for `_IsSubclassValidator`, not sure if necessary or being covered by e2e tests is sufficient. I'm inclined towards the latter.

kedro/framework/project/__init__.py
AntonyMilneQB: Wait, actually surely this won't work with `dynaconf<3.1.6`? Since `_validate_items` will be expecting 2 arguments and you've provided it with 4. So unless we change version requirement to `dynaconf>=3.1.6` we'll need to use `**kwargs`?

deepyaman:
I guess https://github.com/quantumblacklabs/kedro/blob/deepyaman-patch-1/tests/framework/session/conftest.py#L473 has to change to be something that has an `is_error` attribute.

tests/framework/session/conftest.py
```diff
@@ -466,9 +467,15 @@ def mock_import(mocker):
     # KedroSession eagerly validates that a project's settings.py is correct by
     # importing it. settings.py does not actually exists as part of this test suite,
     # so its import is mocked.
-    settings_module = f"{MOCK_PACKAGE_NAME}.settings"
+    settings_module_name = f"{MOCK_PACKAGE_NAME}.settings"
+    settings_module_location = Path(f"src/{MOCK_PACKAGE_NAME}/settings.py").resolve()
```
deepyaman, line 471: There's probably no need to resolve path, I just tried to make it as "realistic" as possible (since I think modules usually have absolute paths); happy to go either way on this.

kedro/framework/project/__init__.py
AntonyMilneQB: In fact, given that we have `**kwargs` shouldn't we also include `*args` for maximum future proof-ness?

tests/framework/session/conftest.py
deepyaman:
```suggestion
    settings_module_location = f"src/{MOCK_PACKAGE_NAME}/settings.py"
```
Maybe we shouldn't resolve...

kedro/framework/project/__init__.py
deepyaman: > In fact, given that we now have `**kwargs` shouldn't we also include `*args` for maximum future proof-ness?

I'm not sure about that; passing along required, positionally arguments is a bit different from passing on these new, parameterized things with defaults IMO.

deepyaman:
Just failing on some Windows tests now...

Let me see if I can resolve today; if not, and assuming this is a priority to get out in 0.17.5, somebody may need to take this on as I'm on PTO next week. This has been less trivial than anticipated. 😅

limdauto:
I think the correct fix for this one is to reimplement the feature using the public method `validate` to avoid them breaking backward compat again. Let's pin it for now and I will re-implement the validator with the public method `validate` instead.

deepyaman:
> I think the correct fix for this one is to reimplement the feature using the public method `validate` to avoid them breaking backward compat again. Let's pin it for now and I will re-implement the validator with the public method `validate` instead.

@limdauto When you say reimplement by overriding `validate`, do you mean as simple as copy-pasting their `validate`/`_validate_items` (with existing modification), or reimplementing to cater to the specific Kedro use cases? If it's the former, feel like can do that easily enough; don't have enough familiarity with how the validation is at the moment to do the latter.

Also, is it better to pin it, or make this minimal modification in this PR (and potentially skip the annoying Windows tests :P)? Understand there's a risk they'll further break it, and pinning is safer. I'm invested for no good reason into figuring out how to make this work. 😅

limdauto:
@deepyaman yea I'm not sure what's going on in this branch but this works 😅 https://github.com/quantumblacklabs/kedro/pulls

limdauto:
Thanks for the PR @deepyaman. I have finished it off in this branch: https://github.com/quantumblacklabs/kedro/pull/893 by overwriting dynaconf public API instead.

PR #893, opened 2021-09-14
Overwrite dynaconf public API instead of private one.

limdauto:
## Description

To avoid incompatibility with future changes in dynaconf private API, switching the implementation of the `IsSubclassValidator` to use their public API instead. I can't remember for the life of me why I had to use the private API before.

## Checklist

## Notice

### DIFF
--- a/kedro/framework/project/__init__.py
+++ b/kedro/framework/project/__init__.py
@@ -54,8 +54,8 @@ def _get_default_class(class_import_path):
 class _IsSubclassValidator(Validator):
     """A validator to check if the supplied setting value is a subclass of the default class"""
 
-    def validate(self, settings):
-        super().validate(settings)
+    def validate(self, settings, *args, **kwargs):
+        super().validate(settings, *args, **kwargs)
 
         default_class = self.default(settings, self)
         for name in self.names:
--- a/kedro/framework/project/__init__.py
+++ b/kedro/framework/project/__init__.py
@@ -54,8 +54,8 @@ def _get_default_class(class_import_path):
 class _IsSubclassValidator(Validator):
     """A validator to check if the supplied setting value is a subclass of the default class"""
 
-    def _validate_items(self, settings, env=None):
-        super()._validate_items(settings, env)
+    def validate(self, settings):
+        super().validate(settings)
 
         default_class = self.default(settings, self)
         for name in self.names:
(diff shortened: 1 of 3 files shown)


kedro/framework/project/__init__.py
```diff
@@ -54,8 +54,8 @@ def validator_func(settings, validators):
 class _IsSubclassValidator(Validator):
     """A validator to check if the supplied setting value is a subclass of the default class"""
 
-    def _validate_items(self, settings, env=None):
-        super()._validate_items(settings, env)
+    def validate(self, settings, *args, **kwargs):
+        super().validate(settings, *args, **kwargs)
```
lorenabalan, line 58: Do we still need to pass `env=None` ourselves or is it a default for `validate()` too?

lorenabalan (approved):
LGTM, just one question. Thanks for the quick fix! 💯

MerelTheisenQB (approved):
LGTM 👍  Does it mean we can also unpin `dynaconf` in requirements again?

kedro/framework/project/__init__.py
limdauto: It's passed internally by dynaconf. It's set as an attribute on the `self` of the `Validator`.

limdauto:
@MerelTheisenQB I suppose so but after implementing `_ProjectPipelines` as a `MutableMapping`, I don't think we really need dynaconf for anything apart from the validators. Maybe we can rewrite this at some point. It isn't that much work. In any case, I have unpinned dynaconf.

requirements.txt
```diff
@@ -2,7 +2,7 @@ anyconfig~=0.10.0
 cachetools~=4.1
 click<8.0
 cookiecutter~=1.7.0
-dynaconf<3.1.6 # Pinned because Dynaconf broke a method signature in 3.1.6 used in Kedro
+dynaconf~=3.1.2
```
limdauto, line 5: I use `3.1.2` here because some users in an attempt to fix their pipeline before might have pinned it to this version.

lorenabalan: Since it's a stable package, can we be a little more liberal with the boundaries? Like >=3.1.2, <4.0? ~3.1.2 will limit it to the 3.1.x series. Can't remember what we had before, off the top of my head.

limdauto: We had `~=3.1.2`. I will change it to be more liberal.

Issue #919, opened 2021-10-01
Extend the PickleDataSet to take any arbitrary backend via classpath


datajoely:
## Description
Currently the [pickle.PickleDataSet](https://kedro.readthedocs.io/en/stable/kedro.extras.datasets.pickle.PickleDataSet.html) explicitly supports `pickle`, `joblib`, `dill`, and `compress_pickle` back-ends. This supports most use-cases but we have observed situations where more flexibility is needed,  [PyTorch](https://pytorch.org/tutorials/beginner/saving_loading_models.html) is one example.

## Possible Implementation
We need to do a few things in this PR: 

- We could introduce a mechanism where the user provides a classpath to a particular back-end 
- In order to use [Custom DataSets](https://kedro.readthedocs.io/en/stable/07_extend_kedro/03_custom_datasets.html) in Kedro we [use importlib (via the load_obj method)](https://github.com/quantumblacklabs/kedro/blob/1dafdc00b11cf0c25c9c4b7412162e4068e70365/kedro/io/core.py#L377) behind the scenes to do something similar in the [DataCatalog](https://kedro.readthedocs.io/en/stable/05_data/01_data_catalog.html). 

To accomplish this, do the following: 
- Follow the process in our [Contribution Guide](https://github.com/quantumblacklabs/kedro/blob/master/CONTRIBUTING.md)
- This is a advanced ticket - so feel free to [jump on to Discord](https://discord.gg/akJDeVaxnB) for coaching through it.

mzjp2:
I'm happy to take this, with a caveat:

`torch` doesn't implement the `pickle` interface -- they use `save` and `load` instead of `dump` and `load` which means that my implementation (where I don't allow the user to configure the name of the `save`/`dump`) will not work with `torch`, but will work with any other backend that does satisfy the interface (as all the existing ones do today).

PR #931, opened 2021-10-04
Extend PickleDataSet to accept arbitrary backends

mzjp2:
## Description
Implements the request in #919, along with the caveat in https://github.com/quantumblacklabs/kedro/issues/919#issuecomment-933914550

This change is backwards compatible with the existing `backend`s we have today.

## Development notes
* Turned the `backend` into a module-string that I use `importlib` to import and call the `load` and `dump` methods on for `load/save` methods respectively.
* Cleaned up the docstrings to reference this new arbitrary backend rather than the hardcoded 4, though left references to the previous 4 as "good examples" of existing backends (e.g linking to their load and dump arguments.
* Updated the tests to match the new functionality, as well as improved one of them (the `ImportError` test) to test the actual exception string. Not sure how that 100% coverage before.

Two questions:

1. I'm assuming you're happy with the caveat I mentioned, otherwise the alternatives are:
a. Instead of `backend`, use `load_method` and `dump_method` and make those import-paths to functions. This is backwards-incompatible with the way `backends` currently work now and I wouldn't recommend this.
b. Allowing the use of a special-case `load_arg` and `save_arg` titled something like `load_name` and `save_name` so that the user can specify the name of the `load ` and `dump` method on the backend, and default them to `load` and `dump` respectively. This would open support for `torch` whilst remaining backwards compatible.

I would recommend b) over a), but will leave the final decision up to you.

2. I've left `dill`, `compress_pickle` and `joblib` in the `test_requirements.txt` as well as left them inside the `test_load` test of the dataset, but perhaps a more pure test here would be:

Creating a tiny custom mocked backend during the test and passing that in (as well as testing the `load_args` and `save_args` functionality) and removing those three backends from the `test_requirements.txt`.

I felt that was a tad overkill here (and I'm lazy), but again -- if you feel this is the way to go, happy to implement this.

## Checklist

## Notice

### DIFF
--- a/kedro/extras/datasets/pickle/pickle_dataset.py
+++ b/kedro/extras/datasets/pickle/pickle_dataset.py
@@ -28,13 +28,13 @@
 
 """``PickleDataSet`` loads/saves data from/to a Pickle file using an underlying
 filesystem (e.g.: local, S3, GCS). The underlying functionality is supported by
-the ``pickle``, ``joblib``, ``dill``, and ``compress_pickle`` libraries, so it
+the specified backend library passed in (defaults to the ``pickle`` library), so it
 supports all allowed options for loading and saving pickle files.
 """
-import pickle
 from copy import deepcopy
 from pathlib import PurePosixPath
 from typing import Any, Dict
+import importlib
 
 import fsspec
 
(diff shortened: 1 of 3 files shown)


datajoely:
Awesome work @mzjp2 !

kedro/extras/datasets/pickle/pickle_dataset.py
```diff
@@ -95,14 +80,8 @@ class PickleDataSet(AbstractVersionedDataSet):
 
     DEFAULT_LOAD_ARGS = {}  # type: Dict[str, Any]
     DEFAULT_SAVE_ARGS = {}  # type: Dict[str, Any]
-    BACKENDS = {
```
datajoely, line 98: I think we may want to keep reference to the libraires we've tested despite allowing for an arbitrary module name? WDYT

MerelTheisenQB: Agreed, I think it would be useful to add a note on which backends we know to work and which don't (e.g. torch).

mzjp2: Added references to the ones that work in the docstring in https://github.com/quantumblacklabs/kedro/pull/931/commits/ed57e7c0cc5927a0b857594f0145c09dbf0ab762

datajoely:
I like it thanks @mzjp2 - but will defer to the team for reviews :)

MerelTheisenQB (approved):
I'm happy with this implementation, thanks @mzjp2 🏆 😄

PR #964, opened 2021-10-14
[KED-2891] Implement `spark.DeltaTable` dataset

jiriklein:
## Description
This PR was created to add the functionality of `databricks DeltaTable` in a `kedro` pipeline. Unlike a regular dataset, which has the concept of `save` that fully overwrites (or creates) the underlying data, a `DeltaTable` is mutable underneath - this means it can be appended to, updated, merged or deleted, as a regular SQL table is (along with commit logs etc.).

This means that unlike with other datasets, `kedro` needs to be extremely careful on defining what the `save` method does. In my opinion, this is, in the case of `DeltaTable`, _modifying an underlying dataset_. The actual implementation is described below.

## Development notes
In my opinion, the `DeltaTable` method calls are best abstracted away from the users. The actual API then comes in through well-designed conf/catalog entry and a _strategy pattern_ (or similar), that resolves into the right call in the `_save` method.
Happy to take onboard any design contributions on this.

## Checklist

## Notice


datajoely:
So I have a few thoughts - this is very related to the SQL conversation I had with @idanov and @MerelTheisenQB a while back which ended with the conclusion that we don't want to encourage many transformation side effects outside of the Kedro DAG.

The central question here is what sort of declarative CRUD functionality we allow in the catalog. Of the two options (1) Catalog definition (2) Within node, option (2) is still possible in both. I don't know if it means we shouldn't do (1), but it highlights the fact that it adds complexity in terms of implementation which we could just delegate to the existing PySpark API.

Upsert example - The syntax doesn't make it especially clear to the reader where exactly A and B come from.
Merge example - Not available yet, but the syntax needed to support merging two catalog entries feels like it will be cumbersome.

The more I write on this comment, the less sure I am that we should expose this sort of functionality in the catalog definition.

IamSandeepGunda:
> @datajoely @limdauto
> 
> I am going to update the PR but I can foresee that the YAML would take a similar shape. Please let me know what do you think about that.
> 
> ```yaml
> _anchor: &anchor
>   type: spark.DeltaTableDataset
> 
> delta_overwrite_dataset:
>   <<: *anchor
>   filepath: /dbfs/path/to/overwrite/dataset
>   delta_options:
>     replaceWhere: "start_date >= '2017-01-01' AND end_date <= '2021-01-01'"
>     overwriteSchema: true
>   save_args:
>     mode: overwrite
> 
> delta_append_dataset:
>   <<: *anchor
>   filepath: /dbfs/path/to/append/dataset
>   save_args:
>     mode: append
> 
> delta_update_dataset:
>   <<: *anchor
>   filepath: /dbfs/path/to/update/dataset
>   delta_options:
>     timestampAsOf: '2019-01-01'
>     versionAsOf: 2
>   save_args:
>     mode: update
>     update_predicates: 
>       - 'col = "foo"'
>     update_values: 
>       'event': 'baz'
>       'another': 'foobar' 
> 
> delta_upsert_dataset:
>   <<: *anchor
>   filepath: /dbfs/path/to/upsert/dataset
>   save_args:
>     mode: upsert
>     key: 'a.id = b.id'
>     whenMatchedUpdate: 
>       'a.data': 'b.new_data'
>     whenNotMatchedInsert:
>       'a.date': 'b.new_date'
>       'a.id': 'b.id'
> 
> delta_delete_dataset:
>   <<: *anchor
>   filepath: /dbfs/path/to/delete/dataset
>   save_args:
>     mode: delete
>     delete_predicates:
>       - 'date < "2017-01-01"'
> ```

The dates in replaceWhere are static. That would be against the purpose of replaceWhere because we'd need it to be updated whenever we're doing incremental load. That dates being set based on whatever incremental load strategy the user is applying.

kedro/extras/datasets/spark/databricks_deltatable_dataset.py
```diff
@@ -0,0 +1,88 @@
...
+            save_args=save_args,
+            version=version,
+            credentials=credentials,
+        )
+        self._delta_options = delta_options
+
+    def _add_options(self, df: DataFrame) -> DataFrameWriter:
+        # DeltaTable specific opts, such as `schemaValidation`
+        if self._delta_options:
+            df_writer = df.write
```
lorenabalan, line 69: Is it `write` not `writer`?

kedro/extras/datasets/spark/databricks_deltatable_dataset.py
```diff
@@ -0,0 +1,88 @@
...
+    ) -> None:
+        super().__init__(
+            filepath=filepath,
+            file_format="delta",
+            load_args=load_args,
+            save_args=save_args,
+            version=version,
+            credentials=credentials,
+        )
+        self._delta_options = delta_options
```
lorenabalan, line 64: Shouldn't we be doing some validation that the mode provided is indeed one of `update`, `upsert`, `delete`, and tell the user to use `SparkDataSet` instead?

kedro/extras/datasets/spark/databricks_deltatable_dataset.py
```diff
@@ -0,0 +1,88 @@
...
+        return DeltaTable.forPath(self._get_spark(), load_path)
+
+    def _save(self, io: NodeStatus):
+        if io == NodeStatus.SUCCESS:
+            logger.info(
+                "`NodeStatus` returned SUCCESS, `save` operation was performed"
+                "within the context of the node."
+            )
+        else:
+            raise DataSetError("`NodeStatus` returned something other than SUCCESS")
```
lorenabalan, line 88: I've been wondering if we could piggyback on the `confirm` method concept that we use in `PartitionedDataSet`. In `PartitionedDataSet` it just updates the checkpoint so that we don't reprocess the same partitions. I think conceptually that's closer than a `save`, and it means the node itself doesn't have to return some special thing. But on the flip side I think it does mean that it's not captured in Kedro Viz as an output..

kedro/extras/datasets/spark/databricks_deltatable_dataset.py
```diff
@@ -0,0 +1,88 @@
...
+        if self._delta_options:
+            df_writer = df.write
+            return functools.reduce(
+                lambda dfw, opt: df_writer.option(opt, self._delta_options[opt]),
+                self._delta_options,
+                df_writer,
+            )
+        return df.write
+
+    def _load(self):
```
lorenabalan, line 77: Don't forget about typehinting here.

lorenabalan (commented):
I think in principle this looks fine, but some of this stuff goes over my head - can you explain again in short why we can't/shouldn't do the `save` operation ourselves rather than leave it to the user in the node?

lorenabalan:
This was a superb diagram to follow, thanks a lot for including it! Looking at this way was very very helpful.

> ## Transcribing and dataset proposal
> user-research

kedro/extras/datasets/spark/databricks_deltatable_dataset.py
datajoely: I like this thought

MerelTheisenQB:
This is some complex stuff, thank you for writing it all up and adding the diagrams. The transcribing and dataset proposal looks good to me, but I'm also not completely following why the `save` operations need to happen in the node? I totally believe you when you say this is how it should be, but am just completely unfamiliar with `DeltaTable` and its inner workings 😬

kedro/extras/datasets/spark/databricks_deltatable_dataset.py
jiriklein: This particular one enriches the `pyspark.sql.DataFrameWriter` with additional options, hence `df_writer` variable on which I reduce the collected opts - basically daisy-chaining `df_writer.option(...).option(...)...`

kedro/extras/datasets/spark/databricks_deltatable_dataset.py
jiriklein: I think that is right - let me try the functionality and how it behaves with the dataset!

kedro/extras/datasets/spark/databricks_deltatable_dataset.py
jiriklein: > This was a superb diagram to follow, thanks a lot for including it! Looking at this way was very very helpful.
> 
> > ## Transcribing and dataset proposal
> > user-research

Thank you! ❤️

kedro/extras/datasets/spark/databricks_deltatable_dataset.py
lorenabalan: Sorry I was very vague - I meant to check it's indeed `df.write` and not `df.writer`. This is me not knowing much about the API.

jiriklein:
> This is some complex stuff, thank you for writing it all up and adding the diagrams. The transcribing and dataset proposal looks good to me, but I'm also not completely following why the `save` operations need to happen in the node? I totally believe you when you say this is how it should be, but am just completely unfamiliar with `DeltaTable` and its inner workings 😬

@MerelTheisenQB @lorenabalan 
In summary - the API is split (and not even self-consistent), but also it "saves" in the new API directly on method calls (i.e. not using `DataFrameWriter`). This means we should somehow validate that the node function has indeed succeeded and inform the user that the saving occurred in-node. 

To go into more detail, unfortunately this all became a lot more complex. The problem with the `DeltaTable`s is following:
1. They only cover "half" of the API - if you wish to overwrite or append to a `DeltaTable`, you call a `toDF` method on the `DeltaTable` instance, get a `pyspark.sql.DataFrame` and save that to a path -- this is perfectly in line with `kedro` and uses existing `pyspark` API.
2. HOWEVER, if you wish to use the new `merge/upsert`, `delete` or `update` functionality, you use this directly on the `DeltaTable` object. Now the `DeltaTable` object has the "filepath" as part a property of the object for one, and does not use `DataFrameWriter` for saving. Instead, the "save" operations happen either on calling `update()` or `delete()` methods, or by calling `execute()` method on a Merge object that gets created by calling `merge()` methods on the `DeltaTable`

I hope this makes sense?

jiriklein:
## Revenge of the Delta: Electric Boogaloo 

@lorenabalan @MerelTheisenQB @AntonyMilneQB @limdauto @datajoely 
Apologies for the wall of text ahead.
Also thank you for the kind words. 
Also additional kudos to @lorenabalan for her time on the discussion we just had. 

To pen down some of the deep dive:
1. The `NodeStatus` does not really work in current implementation. `runner.py` code `outputs = _call_node_run(node, catalog, inputs, is_async, run_id=run_id)` eventually calls `_outputs_to_dictionary` which leads to `_from_list()` which in turn will raise `ValueError` on line 505 of `kedro.pipeline.node` module. This would require a patch/rework. 
2. The `confirm()` method addition might be a good workaround for `logging` that saving has occurred in-node on-object. It is currently only leveraged by `IncrementalDataSet`. Any additional logic could also be handled here - maybe we could think of adding some dataset dependency here? More on this in point 5. 
3. We may allow the user to `return None` from a node that works with `DeltaTable`s - after all, the saving is done in-node on-instance. That being said, this breaks the `kedro` DAG as `None` --> no outputs --> broken DAG. `_outputs_to_dictionary` with create an empty dict here. 
4. Additionally to point 2, if the user returns `None`, then `for name, data in outputs.items()` DOES NOT iterate and therefore 🚨  `hook_manager.hook.before_dataset_saved` and `hook_manager.hook.after_dataset_saved` DO NOT GET CALLED 🚨  @lorenabalan and myself can live with that, since there's no way to distinguish before and after dataset saving (and it's not saved via `kedro`) but this _must be_ communicated.
5. If the user wants a dataset dependency (i.e. "I want to run node B after node a") they would have to return something to a named `MemoryDataSet` that chains the two together. Again something that needs to be documented well. That being said - I would be a fan of extending the `confirm` method (and adding it to `AbstractDataSet` as contract rather than checking for it by `hasattr` reflection). The extension could have access to `DataCatalog` and allow for adding dataset dependency? Just throwing spaghetti at a wall. 

Anyway - please let me know your thoughts.

jiriklein:
> ## Revenge of the Delta: Electric Boogaloo
> @lorenabalan @MerelTheisenQB @AntonyMilneQB @limdauto @datajoely Apologies for the wall of text ahead. Also thank you for the kind words. Also additional kudos to @lorenabalan for her time on the discussion we just had.
> 
> To pen down some of the deep dive:
> 
> 1. The `NodeStatus` does not really work in current implementation. `runner.py` code `outputs = _call_node_run(node, catalog, inputs, is_async, run_id=run_id)` eventually calls `_outputs_to_dictionary` which leads to `_from_list()` which in turn will raise `ValueError` on line 505 of `kedro.pipeline.node` module. This would require a patch/rework.
> 2. The `confirm()` method addition might be a good workaround for `logging` that saving has occurred in-node on-object. It is currently only leveraged by `IncrementalDataSet`. Any additional logic could also be handled here - maybe we could think of adding some dataset dependency here? More on this in point 5.
> 3. We may allow the user to `return None` from a node that works with `DeltaTable`s - after all, the saving is done in-node on-instance. That being said, this breaks the `kedro` DAG as `None` --> no outputs --> broken DAG. `_outputs_to_dictionary` with create an empty dict here.
> 4. Additionally to point 2, if the user returns `None`, then `for name, data in outputs.items()` DOES NOT iterate and therefore 🚨  `hook_manager.hook.before_dataset_saved` and `hook_manager.hook.after_dataset_saved` DO NOT GET CALLED 🚨  @lorenabalan and myself can live with that, since there's no way to distinguish before and after dataset saving (and it's not saved via `kedro`) but this _must be_ communicated.
> 5. If the user wants a dataset dependency (i.e. "I want to run node B after node a") they would have to return something to a named `MemoryDataSet` that chains the two together. Again something that needs to be documented well. That being said - I would be a fan of extending the `confirm` method (and adding it to `AbstractDataSet` as contract rather than checking for it by `hasattr` reflection). The extension could have access to `DataCatalog` and allow for adding dataset dependency? Just throwing spaghetti at a wall.
> 
> Anyway - please let me know your thoughts.

@lorenabalan @MerelTheisenQB @datajoely @limdauto @AntonyMilneQB 
Had couple of thoughts on aligning the API. On `_save` we can either:

- Assume responsibility is on the user, check for `mode` for writing and whether anything was returned from the node function (e.g. `MemoryDataSet` to preserve the `kedro` DAG)
- Do some argument handling on our side:

1. If `_save` method receives `None`, then `pass` as the saving was done on `DeltaTable` in node function, `confirm` method logs this.
2. If `_save` method receives `pyspark.sql.DataFrame` then call to `super()._save(df)` because we're already inheriting from `spark.SparkDataSet`, then we validate against appropriate `mode` (`overwrite, append, ...`), `confirm` logs that saving was delegated to `spark.SparkDataSet`.
3. if `_save` method receives anything else, then treat it as `MemoryDataSet` that is there to preserve the Kedro DAG, assume saving was done on `DeltaTable` in node, `confirm` logs this.  

Thoughts?

lorenabalan:
Hey Jiri! I've been looking at this and drawing some diagrams to work it out in my head. You're gonna hate me but maybe  `confirms()` was not the best idea in the end, given that we still want to `save()` for DAG preservation purposes. SO the way I see it, we have:

a) `DeltaTableDataSet` as input to a node
b) node that does delta-specific operations **only** (_update_, _upsert_ or _delete_)
c) node function must return _something_
d) output of the node gets mapped to a `DeltaTableDataSet`: a different catalog key to the one in a), but with the same target (i.e. looking at the same `filepath` as a))
e) `DeltaTableDataSet.save()` gets called - it just does a `pass` or logs that the operation was done in the node
f) next node that wants to do a delta operation after the one in b) should take dataset from d) as input

Note: catalog keys for a) and d) can be loaded (via transcoding) with `SparkDataSet`.

I don't think we should go back to `NodeStatus` as that feels like over-engineering. 

Does this make sense to you?

lorenabalan:
> If `_save` method receives `pyspark.sql.DataFrame` then call to `super()._save(df)` because we're already inheriting from `spark.SparkDataSet`, then we validate against appropriate `mode` (`overwrite`, `append`, ...),

What if we're conservative and say if `mode` is one of `overwrite`, `append`, `insert` etc, user should've used `SparkDataSet` instead of `DeltaTableDataSet`?

Basically in my head `DeltaTableDataSet` does only delta-specific operations: `update`, `upsert`, `delete` - all of which are done in the node. Therefore `DeltaTableDataSet.save()` is always just a dummy/log statement. 

Let me know if I'm being naive and what's wrong with this thinking.

tests/extras/datasets/spark/test_deltatable_dataset.py
```diff
@@ -0,0 +1,41 @@
...
+    )
+
+    data = [("Alex", 31), ("Bob", 12), ("Clarke", 65), ("Dave", 29)]
+
+    return SparkSession.builder.getOrCreate().createDataFrame(data, schema)
+
+
+class TestDeltaTableDataSet:
+    def test_load(self, tmp_path, sample_spark_df):
+        filepath = (tmp_path / "data.csv").as_posix()
```
jiriklein, line 30: Good start - not sure however why this `csv` one is here? Or is it to be removed?

jiriklein: Actually if you need data you could just `convertToDelta` in place?

tests/extras/datasets/spark/test_deltatable_dataset.py
```diff
@@ -0,0 +1,41 @@
...
+        spark_ds.save(sample_spark_df)
+
+        spark_delta_ds = SparkDataSet(
+            filepath=filepath, file_format="delta", save_args={"mode": "overwrite"}
+        )
+        spark_delta_ds.save(sample_spark_df)
+        delta_ds = DeltaTableDataset(filepath=filepath)
+        delta_table = delta_ds.load()
+
+        assert isinstance(delta_table, DeltaTable)
```
jiriklein, line 41: So personally I think this is exactly what we're looking for.
I would do the following tests:

1. Test loading `dataframewriter_options` from typical yaml (or already parsed into dict) and checking how it resolves - you can see some of the options here https://docs.delta.io/latest/quick-start.html#read-older-versions-of-data-using-time-travel
2. The above + seeing how those options are set on the `spark.sql.DataFrameWriter` object (probs protected properties to check - see here https://spark.apache.org/docs/latest/api/python/_modules/pyspark/sql/readwriter.html#DataFrameWriter)
3. The obvious - `spark.SparkDataSet` with `file_format = "delta"` but incorrect `mode`

jiriklein (commented):
Thank you @lorenabalan for work on this. 
Just realised one thing - for testing to work (without needing to leverage extensive use of `mocker`) it would ideally need to run over a `DeltaLake` to work (the internals of the `DeltaTable` on the server side). 
This might be ok from integration testing perspective + then we mock on our side, but I am happy to sit down tomorrow and discuss what tests (and perhaps resulting mocks) to look for.

kedro/extras/datasets/spark/deltatable_dataset.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def _load(self) -> DeltaTable:
+        load_path = self._fs_prefix + str(self._filepath)
+        return DeltaTable.forPath(self._get_spark(), load_path)
+
+    def _save(self, data: Any) -> None:
+        logger.info(
+            "Saving was performed on `DeltaTable` object within the context of the node function"
+        )
+        # raise DataSetError(f"{self.__class__.__name__} is a read only dataset type")
```
lorenabalan, line 103: I'm tempted to turn into into an actual error instead of a log message. I think that's more consistent with what we have on other datasets (e.g. `APIDataSet`) and also signals to the users that saving doesn't actually do anything (the operation is done in the node already).

docs/source/11_tools_integration/01_pyspark.md
```diff
@@ -115,6 +116,68 @@ df = catalog.load("weather")
...
+  load_args:
+    header: True
+    inferSchema: True
+  save_args:
+    sep: '|'
+    header: True
+
+weather@spark:
+  type: spark.SparkDataSet
+  filepath: data/01_raw/data
```
datajoely, line 140:
```suggestion
  filepath: s3:/my_bucket/03_primary/temperature
```

docs/source/11_tools_integration/01_pyspark.md
```diff
@@ -115,6 +116,68 @@ df = catalog.load("weather")
...
+  file_format: delta
+  save_args:
+    mode: "overwrite"
+    df_writer:
+      versionAsOf: 0
+
+weather@delta:
+  type: spark.DeltaTableDataSet
+  filepath: data/01_raw/data
```
datajoely, line 149:
```suggestion
  filepath: s3:/my_bucket/03_primary/weather
```

docs/source/11_tools_integration/01_pyspark.md
```diff
@@ -115,6 +116,68 @@ df = catalog.load("weather")
...
+```
+
+The `DeltaTableDataSet` does not support `save()` operation, as the updates happen in place inside the node function, i.e. through `DeltaTable.update()`, `DeltaTable.delete()`, `DeltaTable.merge()`.
+
+> Note: `before_dataset_saved` hook will not run at the expected time, as the actual saving happens in the node through the DeltaTable operations.
```
datajoely, line 154:
```suggestion

> Since the save operation happens within the `node` via the  DeltaTable API, the Kedro `before_dataset_saved` hook will not be triggered.
```

docs/source/11_tools_integration/01_pyspark.md
```diff
@@ -115,6 +116,68 @@ df = catalog.load("weather")
...
+
+```python
+Pipeline(
+    [
+        node(func=..., inputs="temperature", outputs="weather@spark"),
```
datajoely, line 159: so I think these need to be named and we need dummy implementation to show how "first_operation_complete" comes to be

docs/source/11_tools_integration/01_pyspark.md
```diff
@@ -115,6 +116,68 @@ df = catalog.load("weather")
...
+        node(
+            func=...,
+            inputs=["first_operation_complete", "weather@delta"],
+            outputs="second_operation_complete",
+        ),
+    ]
+)
+```
+
+`first_operation_complete` is a `MemoryDataSet` and it signals that a set of Delta operations is complete. This can be used as input to a downstream node, to preserve the shape of the DAG. Otherwise, if no downstream nodes need to run after this, the node can simply not return anything:
```
datajoely, line 170:
```suggestion
`first_operation_complete` is a `MemoryDataSet` and it signals that any Delta operations which occur 'outside' the Kedro DAG are complete. This can be used as input to a downstream node, to preserve the shape of the DAG. Otherwise, if no downstream nodes need to run after this, the node can simply not return anything:
```

docs/source/11_tools_integration/01_pyspark.md
datajoely, line 159:
```suggestion
        node(func=process_barometer_data, inputs="temperature", outputs="weather@spark"),
```

docs/source/11_tools_integration/01_pyspark.md
```diff
@@ -115,6 +116,68 @@ df = catalog.load("weather")
...
+
+The `DeltaTableDataSet` does not support `save()` operation, as the updates happen in place inside the node function, i.e. through `DeltaTable.update()`, `DeltaTable.delete()`, `DeltaTable.merge()`.
+
+> Note: `before_dataset_saved` hook will not run at the expected time, as the actual saving happens in the node through the DeltaTable operations.
+
+```python
+Pipeline(
+    [
+        node(func=..., inputs="temperature", outputs="weather@spark"),
+        node(func=..., inputs="weather@delta", outputs="first_operation_complete"),
```
datajoely, line 160:
```suggestion
        node(func=update_meterological_state, inputs="weather@delta", outputs="first_operation_complete"),
```

docs/source/11_tools_integration/01_pyspark.md
```diff
@@ -115,6 +116,68 @@ df = catalog.load("weather")
...
+        node(
+            func=...,
```
datajoely, line 162:
```suggestion
            func=estimate_weather_trend,
```

docs/source/11_tools_integration/01_pyspark.md
datajoely: Perhaps this image is good enough?
image

datajoely (changes requested):
Hi @lorenabalan I've added some comments and an image that I think makes this work. All in all amazing work so far - what do you think?

docs/source/11_tools_integration/01_pyspark.md
lorenabalan: Is it not 
```suggestion
  filepath: s3a://my_bucket/03_primary/temperature
```
for spark?

docs/source/11_tools_integration/01_pyspark.md
lorenabalan:
```suggestion
  filepath: s3a://my_bucket/03_primary/weather
```

docs/source/11_tools_integration/01_pyspark.md
lorenabalan: Thanks, added!

datajoely:
One last point - perhaps it's worth adding a note under the image saying something like:

> Note: This pattern of creating 'dummy' datasets to preserve the data flow also applies to other 'out of DAG' execution operations operations such as SQL operations within a node.

docs/source/11_tools_integration/01_pyspark.md
```diff
@@ -115,6 +116,81 @@ df = catalog.load("weather")
...
+  load_args:
+    header: True
+    inferSchema: True
+  save_args:
+    sep: '|'
+    header: True
+
+weather@spark:
+  type: spark.SparkDataSet
+  filepath: s3a://my_bucket/03_primary/temperature
```
datajoely, line 140:
```suggestion
  filepath: s3://my_bucket/03_primary/temperature
```

docs/source/11_tools_integration/01_pyspark.md
```diff
@@ -115,6 +116,81 @@ df = catalog.load("weather")
...
+
+[Delta Lake](https://delta.io/) is an open-source project that enables building a Lakehouse architecture on top of data lakes. It provides ACID transactions and unifies streaming and batch data processing on top of existing data lakes, such as S3, ADLS, GCS, and HDFS.
+To setup PySpark with Delta Lake, have a look at [the recommendations in Delta Lake's documentation](https://docs.delta.io/latest/quick-start.html#python).
+
+We recommend the following workflow, which makes use of the [Transcoding](../05_data/01_data_catalog.md) feature in Kedro.
+
+```yaml
+temperature:
+  type: spark.SparkDataSet
+  filepath: data/01_raw/data.csv
```
datajoely, line 129:
```suggestion
  filepath: s3://my_bucket/01_raw/data.csv
```

docs/source/11_tools_integration/01_pyspark.md
```diff
@@ -115,6 +116,81 @@ df = catalog.load("weather")
...
+  filepath: s3a://my_bucket/03_primary/temperature
+  file_format: delta
+  save_args:
+    mode: "overwrite"
+    df_writer:
+      versionAsOf: 0
+
+weather@delta:
+  type: spark.DeltaTableDataSet
+  filepath: s3a://my_bucket/03_primary/weather
```
datajoely, line 149:
```suggestion
  filepath: s3://my_bucket/03_primary/weather
```

kedro/extras/datasets/spark/deltatable_dataset.py
```diff
@@ -0,0 +1,118 @@
...
+            >>> data_set.save(spark_df)
+            >>> deltatable_dataset = DeltaTableDataSet(filepath="test_data")
+            >>> delta_table = deltatable_dataset.load()
+            >>>
+            >>> delta_table.update()
+        """
+
+    # this dataset cannot be used with ``ParallelRunner``,
+    # therefore it has the attribute ``_SINGLE_PROCESS = True``
+    # for parallelism within a Spark pipeline please consider
```
MerelTheisenQB, line 66:
```suggestion
    # for parallelism within a Spark pipeline please consider using
```

.circleci/config.yml
```diff
@@ -449,29 +449,29 @@ workflows:
...
+#      - pip_compile_38
+#      - win_unit_tests_36
+#      - win_unit_tests_37
+#      - win_unit_tests_38
+#      - win_pip_compile_36
+#      - win_pip_compile_37
+#      - win_pip_compile_38
+#      - win_e2e_tests_36
+#      - win_e2e_tests_37
+#      - win_e2e_tests_38
```
limdauto, line 474: Should these be put back in?

datajoely (approved):

limdauto (approved):

.circleci/config.yml
```diff
@@ -67,6 +78,9 @@ commands:
       - checkout
       - setup_conda
       - setup_requirements
+      - run:
+          name: Set up SparkSession
+          command: echo """from pyspark.sql import SparkSession;spark = SparkSession.builder.appName('tmp').getOrCreate();""" | python
```
limdauto, line 83: We should take this out

lorenabalan: Done ✨

docs/source/11_tools_integration/01_pyspark.md
```diff
@@ -115,6 +116,81 @@ df = catalog.load("weather")
...
+```
+
+The `DeltaTableDataSet` does not support `save()` operation, as the updates happen in place inside the node function, i.e. through `DeltaTable.update()`, `DeltaTable.delete()`, `DeltaTable.merge()`.
+
+
+> Since the save operation happens within the `node` via the DeltaTable API, the Kedro `before_dataset_saved` hook will not be triggered.
```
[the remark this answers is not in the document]
lorenabalan: Good point, it's the second option. I've rephrased the sentence, hopefully it clearer now.

docs/source/11_tools_integration/01_pyspark.md
```diff
@@ -115,6 +116,81 @@ df = catalog.load("weather")
 assert isinstance(df, pyspark.sql.DataFrame)
 ```
 
+## Spark and Delta Lake interaction
+
+[Delta Lake](https://delta.io/) is an open-source project that enables building a Lakehouse architecture on top of data lakes. It provides ACID transactions and unifies streaming and batch data processing on top of existing data lakes, such as S3, ADLS, GCS, and HDFS.
+To setup PySpark with Delta Lake, have a look at [the recommendations in Delta Lake's documentation](https://docs.delta.io/latest/quick-start.html#python).
+
+We recommend the following workflow, which makes use of the [Transcoding](../05_data/01_data_catalog.md) feature in Kedro.
```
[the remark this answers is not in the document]
lorenabalan: Added a few more sentences in the beginning. ✅

MerelTheisenQB (approved):
Great work everyone! Thanks @lorenabalan for persevering 👏 🏆  The docs look really good as well!

Issue #971, opened 2021-10-19
Add support for Stata datasets


vavalomi:
## Description
Add support for Stata datasets. 

## Context
Stata (www.stata.com) is a popular statistical analysis software used in wide range of fields - economics, biostatistics, education, medicine, and many more.

## Possible Implementation
Pandas already includes read_stata() and to_stata() functions so the implementation could be straightforward based on already implemented pandas data formats.

AntonyMilneQB:
This should now be possible in 0.17.6 (to be released soon) thanks to @datajoely's new [`pandas.GenericDataSet']`(https://kedro.readthedocs.io/en/latest/kedro.extras.datasets.pandas.GenericDataSet.html#kedro.extras.datasets.pandas.GenericDataSet)! Just specify `file_format: stata`. I'm going to close this issue, but do let us know if this doesn't work for some reason 🙂

PR #822, opened 2021-07-06
pandas.XMLDataSet

lucasjamar:
## Description
Create a xml dataset to read xml file intos pandas df and write pandas df to xml.
https://github.com/quantumblacklabs/kedro/issues/818

## Development notes
1. Tested pandas 1.3.0
2. Created XMLDataSet based on pandas.CSVDataSet.
3. Added xml tests based on csv tests.

## Checklist

- [x] Read the [contributing](https://github.com/quantumblacklabs/kedro/blob/master/CONTRIBUTING.md) guidelines
- [x] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [x] Updated the documentation to reflect the code changes
- [x] Added a description of this change and added my name to the list of supporting contributions in the [`RELEASE.md`](https://github.com/quantumblacklabs/kedro/blob/master/RELEASE.md) file
- [x] Added tests to cover my changes

## Notice

- [x] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

- I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
- I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorised to submit this contribution on behalf of the original creator(s) or their licensees.
- I certify that the use of this contribution as authorised by the Apache 2.0 license does not violate the intellectual property rights of anyone else.

Issue #818, opened 2021-07-02
pandas.XMLDataSet


lucasjamar:
## Description
Add a pandas.XMLDataSet to the list of dataset implementations. Newly available in pandas 1.3.0
https://pandas.pydata.org/docs/reference/api/pandas.read_xml.html?highlight=read_xml
https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_xml.html?highlight=xml

## Context
Such a feature would allow users to read data from a url or file straight into a pandas df without needing to create a node to parse the xml.

I'd be happy to have a crack at this one starting next week :) .

datajoely:
@lucasjamar we'd love a PR doing this, in theory you can copy and paste `pandas.CSVDataSet` and simply change the `pandas` call.

AntonyMilneQB:
Sounds like a great idea! But I'd just check with @lorenabalan or @ignacioparicio first since I know we've recently changed the pandas requirement from 1.0 to 1.2 in develop (master is still `>=0.24`). There might be a good reason we can't yet support 1.3?

lorenabalan:
Closing as resolved in #822 , will be available in 0.18.0.

PR #987, opened 2021-10-25
Add generic pandas dataset to extras

datajoely:
## Description

In Kedro the `pandas` datasets have always been defined on an individual basis. This is unlike the Spark (equivalent for example)  since the dataset 'type' is simply a parameter that gets passed in - as such we can live with just one `spark.SparkDataSet` implementation that handles many file formats. The `pandas.GenericDataSet` works in the same way to support some of the more niche formats on a 'best effort basis'. 

Users regularly ask for new pandas datasets to be introduced (#818 #822 #971) and this approach will work for most cases and automatically if pandas themselves introduce more targets.

The dataset works on two assumptions:

1. The io methods provided by pandas target a filesystem read/write operation. This is why SQL, clipboard and in-memory targets are part of an exclusion list which cause the implementation to fail on load/save with an appropriate error message.
2. We using the 'file_format' parameter we are able to retrieve the relevant `pd.read_{file_format}` (on load) or `pd.DataFrame.to_{file_format}` (on save) methods.

Implementation adapted from `pandas.CSVDataSet` implementing `fsspec` and `Versioning` the same way.  Versioning tests identical to CSV implementation as `pandas.GenericDataSet(file_format='csv' ...)` applied.

Note: Not all Pandas methods support both read and write operations. For example, `read_sas` exists, but there is no facility to write it.

## Development notes
- Introduced new dataset class + test + `__init__.py` import
- Introduced new text requirement of lxml to support the HTML test case
- Tested with SAS, CSV and HTML pandas read/write operations.
- This dataset has been written to be as lazy as possible, with failures occurring at load/save time not instantiation. 

## Checklist

## Notice


RELEASE.md
```diff
@@ -12,6 +12,7 @@
 * Upgraded `pip-tools`, which is used by `kedro build-reqs`, to 6.4. This `pip-tools` version requires `pip>=21.2` while [adding support for `pip>=21.3`](https://github.com/jazzband/pip-tools/pull/1501). To upgrade `pip`, please refer to [their documentation](https://pip.pypa.io/en/stable/installing/#upgrading-pip).
 * Extended ``ExcelDataSet`` to support saving Excel files with multiple sheets.
 * `kedro pipeline package <pipeline>` now raises an error if the `<pipeline>` argument doesn't look like a valid Python module path (e.g. has `/` instead of `.`).
+* Introduced `pandas.GenericDataSet` which provides a 'best effort' facility to read / write any format provided by the `pandas` library without a specific Kedro dataset existing beforehand. For a give `kind` parameter, this dataset requires that at least one `pd.read_{kind}` or `pd.DataFrame.to_{kind}` method is available for use.
```
lorenabalan, line 15:
```suggestion
* Introduced `pandas.GenericDataSet` which provides a 'best effort' facility to read / write any format provided by the `pandas` library without a specific Kedro dataset existing beforehand. For a given `kind` parameter, this dataset requires that at least one `pd.read_{kind}` or `pd.DataFrame.to_{kind}` method is available for use.
```

static/jsonschema/kedro-catalog-0.17.json
```diff
@@ -37,7 +37,8 @@
             "pandas.HDFDataSet",
             "pandas.CSVDataSet",
             "pandas.ExcelDataSet",
-            "pandas.GBQTableDataSet"
+            "pandas.GBQTableDataSet",
+            "pandas.GenericDataSet"
```
lorenabalan, line 41: Amazing, thanks for updating this file!!

test_requirements.txt
```diff
@@ -16,6 +16,7 @@ holoviews~=1.13.0
 import-linter[toml]==1.2.6
 ipython~=7.10
 joblib>=0.14
+lxml~=4.6.3
```
lorenabalan, line 19: What's this one needed for?

tests/extras/datasets/pandas/test_generic_dataset.py
```diff
@@ -0,0 +1,399 @@
...
+        )
+        assert msg in str(e.value)
+
+        with pytest.raises(DataSetError) as e:
+            ds.save(pd.DataFrame([1]))
+        msg = (
+            "Unable to retrieve `pandas.DataFrame.to_kedro` method, please ensure that your 'kind' "
+            "parameter has been defined correctly as per the Pandas API https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.html"
+        )
+        assert msg in str(e.value)
```
lorenabalan, line 373: Can we turn all these checks into 

```python
with pyteset.raises(DataSetError, match=pattern):
```

or

```python
with pyteset.raises(DataSetError, match=re.escape(pattern)):
```

for consistency across the codebase?

tests/extras/datasets/pandas/test_generic_dataset.py
```diff
@@ -0,0 +1,399 @@
...
+        fs_mock = mocker.patch("fsspec.filesystem").return_value
+        filepath = "test.csv"
+        data_set = GenericDataSet(kind="sas", filepath=filepath)
+        assert data_set._version_cache.currsize == 0  # no cache if unversioned
+        data_set.release()
+        fs_mock.invalidate_cache.assert_called_once_with(filepath)
+        assert data_set._version_cache.currsize == 0
+
+
+class TestGenericCsvDataSetVersioned:
```
lorenabalan, line 180:
```suggestion
class TestGenericCSVDataSetVersioned:
```

tests/extras/datasets/pandas/test_generic_dataset.py
```diff
@@ -0,0 +1,399 @@
...
+
+        path = filepath.split(PROTOCOL_DELIMITER, 1)[-1]
+
+        assert str(data_set._filepath) == path
+        assert isinstance(data_set._filepath, PurePosixPath)
+
+    def test_catalog_release(self, mocker):
+        fs_mock = mocker.patch("fsspec.filesystem").return_value
+        filepath = "test.csv"
+        data_set = GenericDataSet(kind="sas", filepath=filepath)
```
lorenabalan, line 173: Could we use the fixtures at the top? `filepath_csv` or `csv_data_set`?

tests/extras/datasets/pandas/test_generic_dataset.py
```diff
@@ -0,0 +1,399 @@
...
+
+class TestGenericCsvDataSetVersioned:
+    def test_version_str_repr(self, load_version, save_version):
+        """Test that version is in string representation of the class instance
+        when applicable."""
+        filepath = "test.csv"
+        ds = GenericDataSet(kind="csv", filepath=filepath)
+        ds_versioned = GenericDataSet(
+            kind="csv", filepath=filepath, version=Version(load_version, save_version)
+        )
```
lorenabalan, line 188: Same here, could we reuse some fixtures?

kedro/extras/datasets/pandas/generic_dataset.py
```diff
@@ -0,0 +1,270 @@
...
+        self._fs_open_args_load = _fs_open_args_load
+        self._fs_open_args_save = _fs_open_args_save
+
+    def _ensure_file_system_target(self) -> None:
+        # Fail fast if provided a known non-filesystem target
+        if self._kind in NON_FILE_SYSTEM_TARGETS:
+            raise DataSetError(
+                f"Cannot create a dataset of kind `{self._kind}` as it "
+                f"it does not support a filepath target/source"
+            )
```
lorenabalan, line 210: 👌 👌

kedro/extras/datasets/pandas/generic_dataset.py
```diff
@@ -0,0 +1,270 @@
...
+
+    def _load(self) -> Any:
+
+        self._ensure_file_system_target()
+
+        load_path = get_filepath_str(self._get_load_path(), self._protocol)
+        load_method = getattr(pd, "read_" + self._kind, None)
```
lorenabalan, line 217: This is minor styleguide shenanigans, but we prefer f-strings over string concatenation.
```suggestion
        load_method = getattr(pd, f"read_{self._kind}", None)
```

kedro/extras/datasets/pandas/generic_dataset.py
```diff
@@ -0,0 +1,270 @@
...
+            "parameter has been defined correctly as per the Pandas API "
+            "https://pandas.pydata.org/docs/reference/io.html"
+        )
+
+    def _save(self, data: pd.DataFrame) -> None:
+
+        self._ensure_file_system_target()
+
+        save_path = get_filepath_str(self._get_save_path(), self._protocol)
+        save_method = getattr(data, "to_" + self._kind, None)
```
lorenabalan, line 232:
```suggestion
        save_method = getattr(data, f"to_{self._kind}", None)
```

kedro/extras/datasets/pandas/generic_dataset.py
```diff
@@ -0,0 +1,270 @@
...
+        if save_method:
+            with self._fs.open(save_path, **self._fs_open_args_save) as fs_file:
+                # KEY ASSUMPTION - first argument is path/buffer/io
```
lorenabalan, line 235: Well done for pointing that out! 💡

kedro/extras/datasets/pandas/generic_dataset.py
```diff
@@ -0,0 +1,270 @@
...
+        version: Version = None,
+        credentials: Dict[str, Any] = None,
+        fs_args: Dict[str, Any] = None,
+    ):
+        """Creates a new instance of ``GenericDataSet`` pointing to a concrete data file
+        on a specific filesystem. The appropriate pandas load/save methods are
+        dynamically identified by string matching on a best effort basis.
+
+        Args:
+            kind: String which is used to match the appropriate load/save method on a best
```
lorenabalan, line 132: Is it worth mentioning it's case-insensitive or it's too small of a thing?

kedro/extras/datasets/pandas/generic_dataset.py
```diff
@@ -0,0 +1,270 @@
...
+
+        self._load_args = deepcopy(self.DEFAULT_LOAD_ARGS)
+        if load_args is not None:
+            self._load_args.update(load_args)
+        self._save_args = deepcopy(self.DEFAULT_SAVE_ARGS)
+        if save_args is not None:
+            self._save_args.update(save_args)
+
+        _fs_open_args_save.setdefault("mode", "w")
+        _fs_open_args_save.setdefault("newline", "")
```
lorenabalan, line 200: Do we need these or is it just leftover from a copy-paste from the csv dataset? I don't think they're relevant outside of that.

lorenabalan (commented):
Implementation looks great! 🔥  Left some comments around stylistics and tests structure, happy to approve once they're addressed. 🙂 
There's also a question around some fs args that I don't think we need for this dataset (newline I believe was only relevant for CSVDataSet, and it's even gone in `develop`, and mode should probably be the default one to be able to write binary)

test_requirements.txt
datajoely: The html dataset test

kedro/extras/datasets/pandas/generic_dataset.py
datajoely: These are cut and paste - happy to remove

kedro/extras/datasets/pandas/generic_dataset.py
datajoely: I think better people assume that it's not, but it's nice that it is flexible if they make a mistake.

RELEASE.md
```diff
@@ -12,6 +12,7 @@
 * Upgraded `pip-tools`, which is used by `kedro build-reqs`, to 6.4. This `pip-tools` version requires `pip>=21.2` while [adding support for `pip>=21.3`](https://github.com/jazzband/pip-tools/pull/1501). To upgrade `pip`, please refer to [their documentation](https://pip.pypa.io/en/stable/installing/#upgrading-pip).
 * Extended ``ExcelDataSet`` to support saving Excel files with multiple sheets.
 * `kedro pipeline package <pipeline>` now raises an error if the `<pipeline>` argument doesn't look like a valid Python module path (e.g. has `/` instead of `.`).
+* Introduced `pandas.GenericDataSet` which provides a 'best effort' facility to read / write any format provided by the `pandas` library without a specific Kedro dataset existing beforehand. For a given `kind` parameter, this dataset requires that at least one `pd.read_{kind}` or `pd.DataFrame.to_{kind}` method is available for use.
```
AntonyMilneQB, line 15: Please can we move this to the table in "Major features and improvements" to go with the other new dataset in this release?

kedro/extras/datasets/pandas/generic_dataset.py
```diff
@@ -0,0 +1,270 @@
+# Copyright 2021 QuantumBlack Visual Analytics Limited
```
AntonyMilneQB, line 1: This can disappear now, and can the one in the test file.

kedro/extras/datasets/pandas/generic_dataset.py
```diff
@@ -0,0 +1,270 @@
...
+    ):
+        """Creates a new instance of ``GenericDataSet`` pointing to a concrete data file
+        on a specific filesystem. The appropriate pandas load/save methods are
+        dynamically identified by string matching on a best effort basis.
+
+        Args:
+            kind: String which is used to match the appropriate load/save method on a best
+                effort basis. For example if 'csv' is passed in the `pandas.read_csv` and
+                `pandas.DataFrame.to_csv` will be identified. An error will be raised unless
+                at least matching `read_{kind}` or `to_{kind}` method is identified.
```
AntonyMilneQB, line 135:
```suggestion
                at least one matching `read_{kind}` or `to_{kind}` method is identified.
```

kedro/extras/datasets/pandas/generic_dataset.py
```diff
@@ -0,0 +1,270 @@
...
+        _fs_open_args_save.setdefault("newline", "")
+        self._fs_open_args_load = _fs_open_args_load
+        self._fs_open_args_save = _fs_open_args_save
+
+    def _ensure_file_system_target(self) -> None:
+        # Fail fast if provided a known non-filesystem target
+        if self._kind in NON_FILE_SYSTEM_TARGETS:
+            raise DataSetError(
+                f"Cannot create a dataset of kind `{self._kind}` as it "
+                f"it does not support a filepath target/source"
```
AntonyMilneQB, line 209:
```suggestion
                f"Cannot create a dataset of kind `{self._kind}` as it "
                f"does not support a filepath target/source."
```

AntonyMilneQB (approved):
This is really great 👍  Agree with all of Lorena's comments, but I'd also add:
* I don't much like `kind` as the argument name but also not sure what would be better given that `type` is obviously not possible. Maybe `format`? I was looking at the (pandas docs)[https://pandas.pydata.org/pandas-docs/stable/user_guide/io.html] for inspiration but couldn't immediately see any great options
* should we add `"pandas.GenericDataSet": [PANDAS],`to setup.py? Obviously it won't install all the dependencies since you won't know what's needed in advance but better than nothing I guess

AntonyMilneQB:
Wait, on the `kind` name - how about `file_format`? Seems clear what it means and matches `SparkDataSet`.

datajoely:
> This is really great 👍 Agree with all of Lorena's comments, but I'd also add:
> 
> * I don't much like `kind` as the argument name but also not sure what would be better given that `type` is obviously not possible. Maybe `format`? I was looking at the (pandas docs)[https://pandas.pydata.org/pandas-docs/stable/user_guide/io.html] for inspiration but couldn't immediately see any great options
> * should we add `"pandas.GenericDataSet": [PANDAS],`to setup.py? Obviously it won't install all the dependencies since you won't know what's needed in advance but better than nothing I guess

Excellent push on both points I've mirrored the SparkDataSet style `file_format` and added basic pandas to setup.py

kedro/extras/datasets/pandas/generic_dataset.py
datajoely: Removing these lines caused my Versioning tests to fail with errors similar to this one (those were lifted from CSVDataSet)

image

I can add them back in, but is that the right thing to do?

kedro/extras/datasets/pandas/generic_dataset.py
```diff
@@ -0,0 +1,242 @@
...
+    """
+
+    DEFAULT_LOAD_ARGS = {}  # type: Dict[str, Any]
+    DEFAULT_SAVE_ARGS = {}  # type: Dict[str, Any]
+
+    # pylint: disable=too-many-arguments
+    def __init__(
+        self,
+        file_format: str,
+        filepath: str,
```
lorenabalan, line 91: Nit: Maybe `filepath` can be the first argument, to be consistent with the other datasets.

kedro/extras/datasets/pandas/generic_dataset.py
```diff
@@ -0,0 +1,242 @@
...
+        )
+
+        self._load_args = deepcopy(self.DEFAULT_LOAD_ARGS)
+        if load_args is not None:
+            self._load_args.update(load_args)
+        self._save_args = deepcopy(self.DEFAULT_SAVE_ARGS)
+        if save_args is not None:
+            self._save_args.update(save_args)
+
+        _fs_open_args_save.setdefault("mode", "w")
```
lorenabalan, line 171: Strange that we need this here, but can live with it for now.

tests/extras/datasets/pandas/test_generic_dataset.py
```diff
@@ -0,0 +1,381 @@
...
+
+
+class TestBadGenericDataSet:
+    def test_bad_file_format_argument(self):
+        ds = GenericDataSet(file_format="kedro", filepath="test.kedro")
+
+        pattern = (
+            "Unable to retrieve `pandas.read_kedro` method, please ensure that your 'file_format' "
+            "parameter has been defined correctly as per the Pandas API https://pandas.pydata.org/docs/reference/io.html"
+        )
```
lorenabalan, line 346: These look like long lines. `black` doesn't re-format strings like this but would be nice to split it across at least 3 lines.

tests/extras/datasets/pandas/test_generic_dataset.py
```diff
@@ -0,0 +1,381 @@
...
+
+        with pytest.raises(DataSetError, match=pattern):
+            _ = ds.load()
+
+        pattern2 = (
+            "Unable to retrieve `pandas.DataFrame.to_kedro` method, please ensure that your 'file_format' "
+            "parameter has been defined correctly as per the Pandas API https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.html"
+        )
```
lorenabalan, line 354: same here.

lorenabalan (approved):
LGTM! I like `file_format` better than `kind`. 👌  Thank you for this contribution! 🙌

PR #1024, opened 2021-11-08
[KED-2526] Update logging.yml to write to logs/info.log

AntonyMilneQB:
## Description
**Before this PR**: when you did `kedro run` (or actually any `kedro` command), a file info.log would be created with the following sort of thing:
```
2021-11-08 17:32:57,237 - root - INFO - Registered hooks from 2 installed plugin(s): kedro-inspect-0.1
2021-11-08 17:32:57,337 - root - INFO - Registered CLI hooks from 1 installed plugin(s): kedro-telemetry-0.1.2
2021-11-08 17:32:58,024 - kedro.framework.session.store - INFO - `read()` not implemented for `BaseSessionStore`. Assuming empty store.
2021-11-08 17:32:58,043 - kedro.framework.session.session - WARNING - Unable to git describe /Users/antony_milne/kedro_stuff/projects/spaceflights
```

All log messages after that would be written to logs/info.log (where you'd expect them to go, based on your conf/base/logging.yml file). Over time in a project, you'd just end up with the unwanted info.log file cluttering things up and repeating the above 4 messages. This made me very sad.

**After this PR**: all messages go to logs/info.log where they belong, and I am now happy. Those messages coming from `root` now come from the appropriate `kedro` name also.

## Development notes
The reason these logs were going to the wrong place is that they are emitted before `session._setup_logging()` is called (which sets up logging according to the instructions in conf/base/logging.yml). The reason that they were being written to a file at all (as opposed to just appearing on screen as is the default in Python) is that we have a file kedro/config/logging.yml which is effectively instantiated right at the beginning of the kedro process [here](https://github.com/quantumblacklabs/kedro/blob/master/kedro/framework/cli/cli.py#L18) (has anyone seen this before?!) Turns out that file was directing logs towards info.log instead of logs/info.log, so I've changed that.

## Next steps?

There was a brief discussion a few weeks ago around whether we should still be providing logging.yml as part of conf in the project template or whether we should just be setting defaults on the framework side and allow users to overwrite that. So it turns out that actually we are already doing exactly that! I think I'd be in favour of moving everything in the project logging.yml to the framework default logger and then removing the file from the template entirely (in a separate PR). Thoughts?

## Checklist

## Notice


kedro/framework/session/session.py
```diff
@@ -329,7 +333,8 @@ def run(  # pylint: disable=too-many-arguments,too-many-locals
         """
         # pylint: disable=protected-access,no-member
         # Report project name
-        logging.info("** Kedro project %s", self._project_path.name)
+
```
AntonyMilneQB, line 336: ```suggestion```

kedro/framework/cli/hooks/manager.py
```diff
@@ -42,7 +42,7 @@ def _register_cli_hooks_setuptools(self) -> None:
         }
 
         if plugin_names:
-            logging.info(
+            logging.getLogger(__name__).info(
```
AntonyMilneQB, line 45: This is the pattern used elsewhere in kedro where `self._logger` isn't available. It means the messages will appear to come from `kedro.framework.hooks.manager` rather than `root`.

lorenabalan (commented):
LGTM but some tests are failing.

AntonyMilneQB:
> LGTM but some tests are failing.

The reason for this is that we were only showing messages >= warning from `kedro.framework.cli`. I've removed that in 7bf3d18 which should fix the tests.

This special treatment of logging from `kedro.framework.cli` has been there since the first commit of kedro 0.14. I just tried out some kedro commands (`info`, `run`, `jupyter notebook`, etc.) and there's no unwanted INFO messages that we're trying to suppress so I'm pretty sure this isn't needed any more and was just unintentionally left there (and can always be added into the project logging.yml if a user does for some reason want it).

MerelTheisenQB (approved):
LGTM! 👍 

I don't know too much about Kedro logging, but I like the sound of removing it from the template and adding it to the framework default logger instead.

kedro/framework/hooks/manager.py
```diff
@@ -89,13 +89,13 @@ def _register_hooks_setuptools(
             plugin_names.add(f"{dist.project_name}-{dist.version}")
 
     if disabled_plugin_names:
-        logging.info(
+        logging.getLogger(__name__).info(
```
lorenabalan, line 92: Optional: We can have this factored out on one line at the top.

lorenabalan (approved):
> I think I'd be in favour of moving everything in the project logging.yml to the framework default logger and then removing the file from the template entirely (in a separate PR). Thoughts?

If the users wanted to change anything, would they have to recreate the entire file, or only the customisation they need? Also maybe worth checking in with some users, but otherwise I'm generally not against the idea.

kedro/framework/hooks/manager.py
AntonyMilneQB: Good idea, done!

AntonyMilneQB:
> If the users wanted to change anything, would they have to recreate the entire file, or only the customisation they need? Also maybe worth checking in with some users, but otherwise I'm generally not against the idea.

I think users would just add the customisations they need and it would add to the `logging` object defined on the framework side. I'll check this and do a bit of user research to see what people think of this anyway.

PR #1028, opened 2021-11-11
[KED-2794] Change `kedro build-reqs` to read from `requirements.txt` and save to `requirements.lock`

MerelTheisenQB:
## Description
Make `kedro build-reqs` read requirements from `requirements.txt` and output the compiled, frozen requirements in `requirements.lock`. Make sure the input file and output file can be overwritten from the CLI - ideally through passing directly the options to pip-compile.

## Development notes
- I added two optional click arguments to `build-reqs` for users to specify which files to use as input and output. Previously, users could already specify an output file because `--output-file` is also recognised by `pip-compile`, however, there was no way to specify another input file. I decided to add the two optional arguments to have consistency between those two inputs. 
- Updated the helper method `_get_requirements_in` and removed the `create_empty` flag as argument since this method is now only used in one place where the default for that flag was set as `True`. 

## 🚧  Remaining work to discuss 
In the modular pipelines workflow we still use `requirements.in` when pulling a pipeline. https://github.com/quantumblacklabs/kedro/blob/d63295e7b132cf0cbec9e7b30495e5a42719137d/kedro/framework/cli/pipeline.py#L281-L285

On first glance I'm guessing this should just be changed to use ~~the `requirements.lock` file~~ correction: it should use the `requirements.txt` file instead. I can open a separate PR for that after this is one is merged. 

## Checklist

## Notice


kedro/framework/cli/project.py
```diff
@@ -218,17 +246,33 @@ def build_docs(metadata: ProjectMetadata, open_docs):
 
 
 @forward_command(project_group, name="build-reqs")
+@click.option(
+    "--input-file",
+    "input_file",
+    multiple=False,
+    default=False,
```
lorenabalan, line 253: I don't think `default` is needed for options - I think they're only relevant for flags.

lorenabalan (commented):
Left a couple of suggestions around tidying up some of the code, if possible. 

Otherwise this looks great, the `.lock` files feel more natural than what we had before. Plus the good news is that I heard yesterday at PackagingCon that they're going to introduce some form of `.lock` file as a standard in Python too, so we're on the right track. 🎉 

On the `kedro pipeline pull` question - we should pull the requirements into the `requirements.txt` file. It's meant to add the pipeline requirements into the file that will be _compiled_ for version resolution. I agree that keeping it as a separate PR makes it easier (for you and for reviewing!).

docs/source/09_development/03_commands_reference.md
```diff
@@ -155,9 +155,10 @@ The commands a project supports are specified on the framework side. If you want
 kedro build-reqs
 ```
 
-This command runs [`pip-compile`](https://github.com/jazzband/pip-tools#example-usage-for-pip-compile) on the project's `src/requirements.in` file. If the file doesn't exist, Kedro will create it by copying from `src/requirements.txt`.
+This command runs [`pip-compile`](https://github.com/jazzband/pip-tools#example-usage-for-pip-compile) on the project's `src/requirements.txt` file and will create `src/requirements.lock` with the compiled requirements.
 
-`kedro build-reqs` also accepts and passes through CLI options accepted by `pip-compile`. For example, `kedro build-reqs --generate-hashes` will call `pip-compile --generate-hashes src/requirements.in`.
+`kedro build-reqs` has two optional arguments to specify which file to compile the requirements from and where to safe the compiled requirements to. These arguments are `--input-file` and `--output-file` respectively.
+`kedro build-reqs` also accepts and passes through CLI options accepted by `pip-compile`. For example, `kedro build-reqs --generate-hashes` will call `pip-compile --generate-hashes src/requirements.txt`.
```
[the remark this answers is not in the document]
MerelTheisenQB: Nope, good catch 🦅

kedro/framework/cli/project.py
```diff
@@ -218,17 +246,33 @@ def build_docs(metadata: ProjectMetadata, open_docs):
...
 def build_reqs(
-    metadata: ProjectMetadata, args, **kwargs
+    metadata: ProjectMetadata, input_file, output_file, args, **kwargs
 ):  # pylint: disable=unused-argument
     """Build the project dependency requirements."""
     source_path = metadata.source_dir
-    _build_reqs(source_path, args)
+    _build_reqs(source_path, input_file, output_file, args)
+    _input = input_file if input_file else "requirements.txt"
+    _output = str(output_file) if output_file else "requirements.lock"
```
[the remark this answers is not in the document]
MerelTheisenQB: Good point! `_build_reqs` is indeed only used here so we might as well move the functionality here.

kedro/framework/cli/project.py
```diff
@@ -218,17 +206,53 @@ def build_docs(metadata: ProjectMetadata, open_docs):
 
 
 @forward_command(project_group, name="build-reqs")
+@click.option(
+    "--input-file",
+    "input_file",
+    multiple=False,
```
lorenabalan, line 212: `click` has a handy `click.Path` type we can use for validation. We use it for other file arguments/options as well.

```python
type=click.Path(exists=True, dir_okay=False, resolve_path=True),
```

That way we don't have to do our own error handling, but leave it up to `click`. Can't remember what gets outputted, if it's a nice error message it's probably easier to rely on that instead.

kedro/framework/cli/project.py
```diff
@@ -218,17 +206,53 @@ def build_docs(metadata: ProjectMetadata, open_docs):
...
+            f"File `{input_file}` not found in the project. "
+            "Please specify another input or create the file and try again."
+        )
+
     secho(
-        "Requirements built! Please update requirements.in "
+        f"Requirements built! Please update {input_file.name} "
         "if you'd like to make a change in your project's dependencies, "
-        "and re-run build-reqs to generate the new requirements.txt.",
+        f"and re-run build-reqs to generate the new {output_file.name}.",
```
lorenabalan, line 255:
```suggestion
        f"and re-run build-reqs to generate the new {Path(output_file).name}.",
```

lorenabalan (approved):

kedro/framework/cli/project.py
```diff
@@ -218,17 +206,53 @@ def build_docs(metadata: ProjectMetadata, open_docs):
...
+
+    else:
+        raise FileNotFoundError(
+            f"File `{input_file}` not found in the project. "
+            "Please specify another input or create the file and try again."
+        )
+
     secho(
-        "Requirements built! Please update requirements.in "
+        f"Requirements built! Please update {input_file.name} "
```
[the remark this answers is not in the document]
MerelTheisenQB: aaahh yes 🤦‍♀️  The tests all use `Path` so it was passing fine..

kedro/framework/cli/project.py
MerelTheisenQB: TIL! 👍  
However, this won't catch the case where the user hasn't specified an input file and there isn't a `requirements.txt` in the source directory in the same way. In that case you'd see:
```
Error: Invalid value for '[SRC_FILES]...': Path '/Users/merel_theisen/Projects/Testing/reqs/src/requirements.txt' does not exist.
```

Our own custom error might be a bit more informative. WDYT?

docs/source/04_kedro_project_setup/01_dependencies.md
```diff
@@ -11,15 +11,15 @@ kedro build-reqs
...
+2. [pip compile](https://github.com/jazzband/pip-tools#example-usage-for-pip-compile) the requirements listed in `src/requirements.txt`
+3. Regenerate `src/requirements.lock` to specify a list of pinned project dependencies (those with a strict version)
 
 ```eval_rst
-.. note::  ``src/requirements.in`` contains "source" requirements, while ``src/requirements.txt`` contains the compiled version of those and requires no manual updates.
+.. note::  ``src/requirements.txt`` contains "source" requirements, while ``src/requirements.lock`` contains the compiled version of those and requires no manual updates.
 ```
 
-To further update the project requirements, you should modify `src/requirements.in` (not `src/requirements.txt`) and re-run `kedro build-reqs`.
+To further update the project requirements, you should modify `src/requirements.txt` (not `src/requirements.lock`) and re-run `kedro build-reqs`.
```
AntonyMilneQB, line 22: This can be simplified a lot because the flow is actually much simpler now (we no longer copy requirements.txt to requirements.in), not just renamed files.

For some reason the `suggestion` feature isn't working here, but I'd go for something like this:

```
The `build-reqs` command will [pip compile](https://github.com/jazzband/pip-tools#example-usage-for-pip-compile) the requirements listed in `src/requirements.txt` into a `src/requirements.lock` file that specifies pinned project dependencies (those with a strict version).

```eval_rst
.. note::  ``src/requirements.txt`` contains "source" requirements, while ``src/requirements.lock`` contains the compiled version of those and requires no manual updates.
```
```

File: docs/source/09_development/03_commands_reference.md [hunk 8]
```diff
@@ -155,9 +155,11 @@ The commands a project supports are specified on the framework side. If you want
 kedro build-reqs
 ```
 
-This command runs [`pip-compile`](https://github.com/jazzband/pip-tools#example-usage-for-pip-compile) on the project's `src/requirements.in` file. If the file doesn't exist, Kedro will create it by copying from `src/requirements.txt`.
+This command runs [`pip-compile`](https://github.com/jazzband/pip-tools#example-usage-for-pip-compile) on the project's `src/requirements.txt` file and will create `src/requirements.lock` with the compiled requirements.
 
-`kedro build-reqs` also accepts and passes through CLI options accepted by `pip-compile`. For example, `kedro build-reqs --generate-hashes` will call `pip-compile --generate-hashes src/requirements.in`.
+`kedro build-reqs` has two optional arguments to specify which file to compile the requirements from and where to safe the compiled requirements to. These arguments are `--input-file` and `--output-file` respectively.
```
AntonyMilneQB, line 160:
```suggestion
`kedro build-reqs` has two optional arguments to specify which file to compile the requirements from and where to save the compiled requirements to. These arguments are `--input-file` and `--output-file` respectively.
```

docs/source/09_development/03_commands_reference.md
```diff
@@ -155,9 +155,11 @@ The commands a project supports are specified on the framework side. If you want
...
+
+`kedro build-reqs` also accepts and passes through CLI options accepted by `pip-compile`. For example, `kedro build-reqs --generate-hashes` will call `pip-compile --generate-hashes src/requirements.txt`.
```
AntonyMilneQB, line 162:
```suggestion
`kedro build-reqs` also accepts and passes through CLI options accepted by `pip-compile`. For example, `kedro build-reqs --generate-hashes` will call `pip-compile --output-file=src/requirements.lock --generate-hashes src/requirements.txt`.
```
Is that right?

kedro/framework/cli/project.py
```diff
@@ -218,17 +206,53 @@ def build_docs(metadata: ProjectMetadata, open_docs):
...
+    help=OUTPUT_FILE_HELP,
+)
 @click.pass_obj  # this will pass the metadata as first argument
 def build_reqs(
-    metadata: ProjectMetadata, args, **kwargs
+    metadata: ProjectMetadata, input_file, output_file, args, **kwargs
 ):  # pylint: disable=unused-argument
-    """Build the project dependency requirements."""
+    """Run `pip-compile` on requirements.txt or the user defined input file and save
+    the compiled requirements to requirements.lock or the user defined output file.
```
AntonyMilneQB, line 226:
```suggestion
    """Run `pip-compile` on src/requirements.txt or the user defined input file and save
    the compiled requirements to src/requirements.lock or the user defined output file.
```

kedro/framework/cli/project.py
```diff
@@ -218,17 +206,53 @@ def build_docs(metadata: ProjectMetadata, open_docs):
...
-    _build_reqs(source_path, args)
+    input_file = input_file or source_path / "requirements.txt"
+    output_file = output_file or source_path / "requirements.lock"
+
+    if input_file.is_file():
+        python_call(
+            "piptools",
+            [
+                "compile",
+                "-q",
```
AntonyMilneQB, line 238: I'd get rid of this `-q` flag. I think it was only relevant when `build_reqs` was being called as part of `kedro install`, which it isn't any more.

kedro/framework/cli/project.py
AntonyMilneQB: I think this case is very unlikely given that projects come with `src/requirements.txt` and there's not really any reason someone would remove it.

Even better though, could we use `default` in the `click.option` to set the default input and output file paths? That way click will check the existence of `input-path` even if it's the default path (though obviously you don't want to do that check for `output-path`) and you could remove 
```
    input_file = input_file or source_path / "requirements.txt"
    output_file = output_file or source_path / "requirements.lock"
```

Not a biggy either way though 🙂

AntonyMilneQB (approved):
Nice! When you remove the `_get_requirements_in` it would be good to check if there's any mention left anywhere of requirements.in (in docs and source).

kedro/framework/cli/project.py
```diff
@@ -218,17 +206,53 @@ def build_docs(metadata: ProjectMetadata, open_docs):
...
+    """Run `pip-compile` on requirements.txt or the user defined input file and save
+    the compiled requirements to requirements.lock or the user defined output file.
+    """
+
     source_path = metadata.source_dir
-    _build_reqs(source_path, args)
+    input_file = input_file or source_path / "requirements.txt"
+    output_file = output_file or source_path / "requirements.lock"
+
+    if input_file.is_file():
```
lorenabalan, line 233:
```suggestion
    if Path(input_file).is_file():
```
so that it works when user provides an `input_file` (which is a `str`).

Alternatively we can cast them earlier, L230-231, and revert back to `input_file.name` in the error message.

kedro/framework/cli/project.py
MerelTheisenQB: I'm don't think you can set the default path because we need to calculate the `source_path` first.

kedro/framework/cli/project.py
MerelTheisenQB: I opted for casting earlier in L230-231

SajidAlamQB (approved):
Awesome! 👍

PR #1047, opened 2021-11-17
Fix erroneous plugin count

AntonyMilneQB:
## Description
A couple of small bug fixes from things spotted while working on another ticket. I've been getting log messages like this:
```
INFO - Registered hooks from 2 installed plugin(s): kedro-inspect-0.1
```
... which obviously doesn't make sense (2 installed plugins but only 1 listed).

Turns out there were actually two issues here:
* `self.load_setuptools_entrypoints` returns the total number of times the entrypoint is used rather than the total number of plugins that use the entrypoint. If you have a plugin that uses the same entrypoint multiple times then the count would be wrong
* in CLI hooks we weren't subtracting the number of `already_registered` plugins

Solution is simple: just give the same number as the length of the list you're printing. That way it's always consistent 😀 

## Development notes
Tested manually and all is good now.

## Checklist


datajoely (approved):

lorenabalan (approved):
Nice catch! Feel free to update the release notes as well if you wish.

SajidAlamQB (approved):

PR #1058, opened 2021-11-22
Add new logging config for CLI hooks manager

AntonyMilneQB:
## Description
In #1024, someone thought it was a good idea to tidy up the loggers a bit. Unfortunately, what said person didn't realise is that this resulted in an incredibly annoying message every time you execute a `kedro` command outside a project...
image

The reason for this is that the CLI hooks manager was trying to write to a file `logs/info.log`, which didn't exist. Actually if we're not in a kedro project we shouldn't be trying to log messages to a file at all because that would be really annoying. I've added a special logging handler for this case which means that the message will only be shown on the console and not written to any file.

The other instances where I had changed `logging.info` to the "correct" kedro logger were in:
* `_register_hooks_setuptools`
* `Session.run`
* `DataCatalog.list`

These are all ok because they're only ever called inside a kedro project, when we _do_ want the info.log file.

## Checklist

### DIFF
--- a/kedro/config/logging.yml
+++ b/kedro/config/logging.yml
@@ -37,6 +37,11 @@ loggers:
         handlers: [console]
         propagate: no
 
+    kedro.framework.cli.hooks.manager:
+        level: INFO
+        handlers: [console]
+        propagate: no
+
 root:
     level: INFO
     handlers: [console, info_file_handler, error_file_handler]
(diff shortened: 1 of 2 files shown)


SajidAlamQB (approved):

lorenabalan (approved):
Good catch! Thank you for fixing it so quickly! 👏

MerelTheisenQB (approved):
👍

PR #1059, opened 2021-11-22
Remove alphabetic sort on node inputs/outputs

### LINKED ISSUE #1055: Possible Incorrect Node argument ordering on Exception
## Description
I've noticed that the order of the node arguments that are logged when a node raises an exception does not the order in which they are defined in the pipeline. 

## Context
The logged exception is useful for resuming the pipeline from the node that failed. If the arguments are not in the correct order, then it's likely that resuming the pipeline using 

```bash
kedro run --from-nodes ...
```
will still result in an error. It could be that the arguments are passed in as keyword arguments, but this is unclear from the error log.

## Steps to Reproduce
The following script 
```python
from kedro.io import DataCatalog, MemoryDataSet
from kedro.pipeline import node, Pipeline
from kedro.runner import SequentialRunner

import pandas as pd
from sklearn.dummy import DummyClassifier

import logging

log = logging.getLogger(__name__)
logging.basicConfig(level=logging.INFO)

def split_data(df):
    X = df.iloc[:,0]
    y = df.iloc[:,1]
    return X, y

def create_model():
    return DummyClassifier()

def train_model(model, X_train, y_train):
    raise ValueError #force exception
    model.fit(X_train, y_train)
    return model

data_catalog = DataCatalog({"data": MemoryDataSet(data=pd.DataFrame({"a":[0,1,2],"b":[True, False, True]}))})
pipeline = Pipeline([
    node(split_data, inputs="data", outputs=["X","y"]),
    node(create_model, inputs=None, outputs="model"),
    node(train_model, inputs=["model","X","y"], outputs="trained_model")
])

runner = SequentialRunner()
print(runner.run(pipeline, data_catalog))
```

## Expected Result
On exception, the logged message should preserve the order of the inputs

```
ERROR:kedro.pipeline.node:Node `train_model([X,model,y]) -> [trained_model]` failed with error: 
WARNING:kedro.runner.sequential_runner:There are 1 nodes that have not run.
You can resume the pipeline run by adding the following argument to your previous command:
  --from-nodes "train_model([model,X, y]) -> [trained_model]"
```

## Actual Result
The arguments appear to be ordered alphabetically, with capital letters taking precedence over lowercase

```
ERROR:kedro.pipeline.node:Node `train_model([X,model,y]) -> [trained_model]` failed with error: 
WARNING:kedro.runner.sequential_runner:There are 1 nodes that have not run.
You can resume the pipeline run by adding the following argument to your previous command:
  --from-nodes "train_model([X,model,y]) -> [trained_model]"
```

## Your Environment

* Kedro version used (`pip show kedro` or `kedro -V`): kedro, version 0.17.5
* Python version used (`python -V`): Python 3.8.10
* Operating system and version: Windows 10 Version 20H2 (OS Build 19042.928)

AntonyMilneQB:
## Description
Fixes #1055 and a long-standing gripe of mine... `str(node)` currently gives the inputs/outputs in alphabetical order. This is confusing because it means they do not agree with the node definition. I checked back through kedro history to see if there was any good reason for the sorting and decided there wasn't.

Before this PR:
```
2021-11-22 12:53:25,089 - kedro.pipeline.node - INFO - Running node:
create_model_input_table([preprocessed_companies,preprocessed_shuttles,reviews]) -> [model_input_table]
```

After this PR:
```
2021-11-22 12:48:11,998 - kedro.pipeline.node - INFO - Running node: 
create_model_input_table([preprocessed_shuttles,preprocessed_companies,reviews]) -> [model_input_table]
```

## Checklist


RELEASE.md
```diff
@@ -30,6 +30,7 @@
 * Deprecated the "Thanks for supporting contributions" section of release notes to simplify the contribution process; Kedro 0.17.6 is the last release that includes this. This process has been replaced with the [automatic GitHub feature](https://github.com/quantumblacklabs/kedro/graphs/contributors).
 * Fixed a bug where the version on the tracking datasets didn't match the session id and the versions of regular versioned datasets.
 * Fixed an issue where datasets in `load_versions` that are not found in the data catalog would silently pass.
+* Altered the string representation of node names so that node input/output is preserved rather than being alphabetically sorted.
```
AntonyMilneQB, line 33:
```suggestion
* Altered the string representation of nodes so that node input/output is preserved rather than being alphabetically sorted. 
```

kedro/pipeline/node.py
```diff
@@ -168,11 +168,11 @@ def __hash__(self):
...
     def __str__(self):
-        def _sorted_set_to_str(xset):
-            return f"[{','.join(sorted(xset))}]"
+        def _set_to_str(xset):
+            return f"[{','.join(xset)}]"
 
-        out_str = _sorted_set_to_str(self.outputs) if self._outputs else "None"
-        in_str = _sorted_set_to_str(self.inputs) if self._inputs else "None"
+        out_str = _set_to_str(self.outputs) if self._outputs else "None"
+        in_str = _set_to_str(self.inputs) if self._inputs else "None"
```
lorenabalan, line 175: I think these would still be sorted in the case of dicts. i.e. if `node(func, inputs={x="y", a="b"}, outputs={z="Z"})`, but that might be fine? Not sure if removing the `sorted()` will break other things.

lorenabalan (approved):
Ha didn't even realise we were doing that.

kedro/pipeline/node.py
AntonyMilneQB: This is a great point. Actually it turns out the behaviour was a little weird in that it didn't reorder`inputs` but it did reorder `outputs`... From looking through some old PRs, it seems like the only reason for the ordering was to get a deterministic dictionary order in Python 3.5. But now in Python 3.6+ it's always preserved no need to keep it, so I've removed it.

RELEASE.md
```diff
@@ -30,6 +30,7 @@
 * Deprecated the "Thanks for supporting contributions" section of release notes to simplify the contribution process; Kedro 0.17.6 is the last release that includes this. This process has been replaced with the [automatic GitHub feature](https://github.com/quantumblacklabs/kedro/graphs/contributors).
 * Fixed a bug where the version on the tracking datasets didn't match the session id and the versions of regular versioned datasets.
 * Fixed an issue where datasets in `load_versions` that are not found in the data catalog would silently pass.
+* Altered the string representation of nodes so that node inputs/outputs order is preserved rather than being alphabetically sorted.
```
AntonyMilneQB, line 33:
```suggestion
* Altered the string representation of nodes so that node inputs/outputs order is preserved rather than being alphabetically sorted.
```

SajidAlamQB (approved):
Nice!

MerelTheisenQB (approved):
I didn't know we were doing that either. Nice fix! 👍

PR #1063, opened 2021-11-23
[KED-2565] Deprecate `register_catalog` and `register_config_loader` hooks

MerelTheisenQB:
## Description
The `register_catalog` and `register_config_loader` hooks will be removed in Kedro 0.18.0. This PR adds a deprecation warning to both hooks. 

## Development notes
Originally, [we decided not to add a deprecation warning to the `register_config_loader` hook ](https://github.com/quantumblacklabs/private-kedro/pull/1090#issuecomment-833587489) because users can't get rid of it.  We're now closer to releasing `0.18.0` so these messages won't be showing much longer, but the original argument is still valid. I don't have a very strong opinion either way. For what it's worth this is what the warnings look like (first two lines after `kedro run`:

Screenshot 2021-11-23 at 16 56 31

## Checklist

- [X] Added a description of this change in the [`RELEASE.md`](https://github.com/quantumblacklabs/kedro/blob/master/RELEASE.md) file

### DIFF
--- a/kedro/framework/hooks/specs.py
+++ b/kedro/framework/hooks/specs.py
@@ -301,7 +301,12 @@ class RegistrationSpecs:
         """
         pass
 
-    @hook_spec(firstresult=True)
+    @hook_spec(
+        firstresult=True,
+        warn_on_impl=DeprecationWarning(
+            "The `register_config_loader` hook is deprecated and will be removed in Kedro 0.18.0."
+        ),
+    )
     def register_config_loader(
         self, conf_paths: Iterable[str], env: str, extra_params: Dict[str, Any]
     ) -> ConfigLoader:
@@ -317,7 +322,12 @@ class RegistrationSpecs:
         """
         pass
 
-    @hook_spec(firstresult=True)
+    @hook_spec(
+        firstresult=True,
+        warn_on_impl=DeprecationWarning(
+            "The `register_catalog` hook is deprecated and will be removed in Kedro 0.18.0."
+        ),
+    )
     def register_catalog(
         self,
         catalog: Optional[Dict[str, Dict[str, Any]]],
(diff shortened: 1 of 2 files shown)


lorenabalan:
> We're now closer to releasing 0.18.0 so these messages won't be showing much longer, but the original argument is still valid. 

So I think I still prefer to not have deprecation warnings for the same reasons - 0.17.6 may be the last minor release before 0.18, but it's not necessarily a short time to for users. Big upgrades still happen late/slower than small upgrades. 

We should probably make some noise about `settings.py` mechanism replacing registration hooks (Discord/GH Discussions?), but I worry this is more annoying than helpful as they can't do anything about it or learn about the new way unless the check out `develop` source code (we don't have a readthedocs build for `develop`).

Would love to hear other opinions though, happy to have it merged if others think otherwise.

MerelTheisenQB:
I like the idea of communicating a bit more about what's going to happen to the registration hooks. I think even though it might be a bit annoying, it's better than not having any deprecation warnings at all. My guess is most people won't care, but the ones who do will then at least get a heads-up about things changing before it's actually released and done.

datajoely:
@lorenabalan @MerelTheisenQB  there is probably a medium article in this one image
image

kedro/framework/hooks/specs.py
```diff
@@ -301,7 +301,12 @@ def register_pipelines(self) -> Dict[str, Pipeline]:
         """
         pass
 
-    @hook_spec(firstresult=True)
+    @hook_spec(
+        firstresult=True,
+        warn_on_impl=DeprecationWarning(
+            "The `register_config_loader` hook is deprecated and will be removed in Kedro 0.18.0."
```
datajoely, line 307: Do we want to say something about the activity will live in `settings.py`

AntonyMilneQB:
Also don't have strong feelings on this but tend to agree with @lorenabalan on this one. I don't see the use in adding deprecation warnings here but we should 100% make clear how users can migrate from these registration hooks to settings.py when we release 0.18.

PR #1064, opened 2021-11-23
[KED-2565] Register `DataCatalog` in settings instead of through the registration hook.

MerelTheisenQB:
## Description
`DataCatalog` is currently registered via hooks - however we would like this to be via `settings.py` instead. 

## Development notes
* Followed the same refactoring the `DataCatalog` registration as done for the `ConfigLoader`
* Kept the `DataCatalog` as part of the `KedroContext` for this PR. The removal will happen seperately.
* Default `register_catalog` `hook_impl` has been removed
* `DATA_CATALOG_CLASS` has been added to `_ProjectSettings` class impl
* `DATA_CATALOG_CLASS` has been added (commented) to `settings.py`
* Default `DATA_CATALOG_CLASS` is set to `kedro.io.DataCatalog`
* Left the `ProjectHooks` in place for now as we have a separate task in the backlog to get rid of the registration hooks fully.

## Checklist

- [X] Added a description of this change in the [`RELEASE.md`](https://github.com/quantumblacklabs/kedro/blob/master/RELEASE.md) file


features/steps/test_starter/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/hooks.py
```diff
@@ -1,19 +1,5 @@
...
-        self,
-        catalog: Optional[Dict[str, Dict[str, Any]]],
-        credentials: Dict[str, Dict[str, Any]],
-        load_versions: Dict[str, str],
-        save_version: str,
-    ) -> DataCatalog:
-        return DataCatalog.from_config(
-            catalog, credentials, load_versions, save_version
-        )
+    pass
```
lorenabalan, line 5: I'm thinking, would it be better to delete the file altogether? The alternative is to leave it as example only (with commented out code), but that might be annoying for users to have a file that they'd have to delete?

kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/hooks.py
```diff
@@ -6,14 +6,4 @@
...
-        self,
-        catalog: Optional[Dict[str, Dict[str, Any]]],
-        credentials: Dict[str, Dict[str, Any]],
-        load_versions: Dict[str, str],
-        save_version: str,
-    ) -> DataCatalog:
-        return DataCatalog.from_config(
-            catalog, credentials, load_versions, save_version
-        )
+    pass
```
lorenabalan, line 9: Same here, maybe best to delete it and remove it from `settings.py` as well.

kedro/framework/context/context.py
```diff
@@ -267,18 +268,18 @@ def _get_catalog(
...
+                load_versions=load_versions,
+                save_version=save_version,
+            )
+        except (TypeError, AttributeError) as exc:
             raise KedroContextError(
                 f"Expected an instance of `DataCatalog`, "
-                f"got `{type(catalog).__name__}` instead."
-            )
+                f"got `{type(settings.DATA_CATALOG_CLASS)}` instead."
+            ) from exc
```
lorenabalan, line 282: So actually this is something that @idanov also pointed out and I tend to agree - that we should let `settings.py` handle the validation, rather than add another layer in `KedroContext`. We've set a bad precedent with `config_loader` but we can remediate it as it's not released yet. 

This is what we do for `SESSION_STORE_CLASS` already - the `KedroSession` accesses it straightforwardly with `settings.SESSION_STORE_CLASS` and doesn't wrap it in exception handling.

kedro/framework/project/__init__.py
```diff
@@ -63,6 +63,9 @@ class _ProjectSettings(LazySettings):
         "CONFIG_LOADER_CLASS", default=_get_default_class("kedro.config.ConfigLoader")
     )
     _CONFIG_LOADER_ARGS = Validator("CONFIG_LOADER_ARGS", default={})
+    _DATA_CATALOG_CLASS = Validator(
```
lorenabalan, line 66: Can we actually use `_IsSubclassValidator` like we do for `_SESSION_STORE_CLASS`? The config loader should also follow the same pattern - feel free to include it here or in a separate PR, whichever is easier for you.

lorenabalan (commented):
Thank you for picking this up and making it such a clean PR to review. Made a few suggestions, let me know what you think. Everything looks good otherwise.

features/steps/test_starter/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/hooks.py
MerelTheisenQB: I considered it, but thought it would make more sense to do it as part of [dropping support for registration hooks](https://jira.quantumblack.com/browse/KED-2504)? I don't mind doing it here, but it would increase the scope of the PR.

kedro/framework/context/context.py
MerelTheisenQB: Ah good shout! I didn't realise that. I'll change it and do the `config_loader` as well.

kedro/framework/project/__init__.py
MerelTheisenQB: I changed `_CONTEXT_CLASS` to use `_IsSubclassValidator` as well

tests/framework/session/test_session.py
```diff
@@ -85,34 +92,41 @@ def mock_config_loader_class(mocker):
 @pytest.fixture
 def mock_settings_config_loader_class(mocker, mock_config_loader_class):
     class MockSettings(_ProjectSettings):
-        _CONFIG_LOADER_CLASS = Validator(
+        mocker.patch("kedro.framework.project.issubclass")
```
lorenabalan, line 95: This line surprised me and got a bit confused in the tests - do you remember why we need to mock the config loader class instead of using `ConfigLoader` as is? (L88-89)

lorenabalan (approved):
Awesome stuff! 💥

tests/framework/session/test_session.py
MerelTheisenQB: Very good catch! 🦅   It seems like it was mocked to make `_get_logging_config` work. I've now just mocked the call to `_get_logging_config` instead and removed the mocked config loader.

lorenabalan: Nice thank you! Do we need to do the same in `test_settings.py` as well?

RELEASE.md
```diff
@@ -64,15 +64,17 @@
...
 ## Thanks for supporting contributions
 
 [Deepyaman Datta](https://github.com/deepyaman), [Lucas Jamar](https://github.com/lucasjamar), [Simon Brugman](https://github.com/sbrugman)
 
 ## Migration guide from Kedro 0.17.* to 0.18.*
-* Please remove any existing `hook_impl` of the `register_config_loader` method from `ProjectHooks` (or custom alternatives).
+* Please remove any existing `hook_impl` of the `register_config_loader` and `register_catalog` methods from `ProjectHooks` (or custom alternatives).
 * Populate `settings.py` with `CONFIG_LOADER_CLASS` set to your expected config loader class (for example `kedro.config.TemplatedConfigLoader` or custom implementation). If `CONFIG_LOADER_CLASS` value is not set, it will default to `kedro.config.ConfigLoader` at runtime.
 * Populate `settings.py` with `CONFIG_LOADER_ARGS` set to a dictionary with expected keyword arguments. If `CONFIG_LOADER_ARGS` is not set, it will default to an empty dictionary.
+* Populate `settings.py` with `DATA_CATALOG_CLASS` set to your expected data catalog class. If `DATA_CATALOG_CLASS` value is not set, it will default to `kedro.io.DataCatalog` at runtime.
```
AntonyMilneQB, line 77:
```suggestion
* Populate `settings.py` with `DATA_CATALOG_CLASS` set to your expected data catalog class. If `DATA_CATALOG_CLASS` is not set, it will default to `kedro.io.DataCatalog` at runtime.
```

tests/framework/session/test_session.py
MerelTheisenQB: Done!

AntonyMilneQB (approved):
General 👍 👍 👍  from me but with a few questions...

* Do we want to be exposing `DATA_CATALOG_ARGS` as we do `CONFIG_LOADER_ARGS`? I don't know if this has been considered already. Personally I have no strong feelings on it - we can always add it on in a separate PR or some time in the future if it turns out it's needed. I suspect @lorenabalan or @idanov would know the answer to whether we want to do this anyway
* Maybe I missed it, but I couldn't see any tests to make sure this is actually working ok? You have the `BadCatalog` but I didn't see any happy path tests analogous to the config loader tests that use `MyConfigLoader`

Completely separate from this PR, but before we release 0.18 we should really go through the release notes and reorganise them. e.g. to put all the registration hook change stuff together.

MerelTheisenQB:
> * Do we want to be exposing `DATA_CATALOG_ARGS` as we do `CONFIG_LOADER_ARGS`? I don't know if this has been considered already. Personally I have no strong feelings on it - we can always add it on in a separate PR or some time in the future if it turns out it's needed. I suspect @lorenabalan or @idanov would know the answer to whether we want to do this anyway

Good question. I'm not sure if we've discussed this or not 🤔  The `DataCatalog` is constructed slightly differently from the `ConfigLoader` though, it calls `from_config` which takes specific parameters so we probably need to add some input validation. I haven't really looked at it in depth though so maybe I'm talking nonsense. 

> * Maybe I missed it, but I couldn't see any tests to make sure this is actually working ok? You have the `BadCatalog` but I didn't see any happy path tests analogous to the config loader tests that use `MyConfigLoader`

The happy paths are the already existing tests. Follow up to this task will be to remove the catalog from KedroContext, so I think it makes sense to revisit/rewrite the tests as part of that as well.

AntonyMilneQB:
> Good question. I'm not sure if we've discussed this or not 🤔 The `DataCatalog` is constructed slightly differently from the `ConfigLoader` though, it calls `from_config` which takes specific parameters so we probably need to add some input validation. I haven't really looked at it in depth though so maybe I'm talking nonsense.

I don't think the fact that it's constructed from `from_config` would make a difference here. We could do
```
        catalog = settings.DATA_CATALOG_CLASS.from_config(
            catalog=conf_catalog,
            credentials=conf_creds,
            load_versions=load_versions,
            save_version=save_version,
            **settings.DATA_CATALOG_ARGS
        )
```

...analogous to how we do

```
 return config_loader_class(
                conf_source=str(self._project_path / settings.CONF_SOURCE),
                env=env,
                runtime_params=extra_params,
            )
```

Overall to me, at a glance it makes sense that we _should_ provide `DATA_CATALOG_ARGS` since, as I understand it, data catalog is on an equivalent footing to config loader and hence should have the same options for configurability and extensibility. This is in contrast to `CONTEXT_CLASS`, for which I'm guessing we are deliberately not enabling `CONTEXT_ARGS`. But I guess others may argue that we should not enable this functionality until there's a proven user need for it.

> The happy paths are the already existing tests. Follow up to this task will be to remove the catalog from KedroContext, so I think it makes sense to revisit/rewrite the tests as part of that as well.

Do these test what happens if I overwrite `DATA_CATALOG_CLASS` with my own custom class? I couldn't see that but could definitely just be missing it!

lorenabalan:
My personal preference would be to leave it as is for now because I don't think we know enough information about that workflow, if it even exists. With the `ConfigLoader` we at least could look at two classes (templated vs regular) and infer a pattern, but there's no feasible option for an alternative for `DataCatalog` out of the box - only potentially user-defined options which we have no visibility over.  

If the workflow exists, I would rather make the catalog go through the same exercise as the config loader, and see if we can define a clearer interface, rather than force it now and feel patchy. We can add it later as a feature addition, not a breaking change.

MerelTheisenQB:
Thanks for your input @lorenabalan and @AntonyMilneQB! I will merge this PR now as is and if we find that users need to pass arguments to the catalog we can create a follow up task to investigate and add this functionality.

PR #1066, opened 2021-11-24
[KED-2939] Fix micropackaging inconsistency with kedro pipeline pull --alias

SajidAlamQB:
## Description

This is only a problem on develop branch, `kedro pipeline pull --alias` should accept a simple string only (for renaming)
add a `--destination` flag on `kedro pipeline pull` to accept the path where to unpack the pipeline
Context:

```
[tool.kedro.pipeline.pull]
"dist/data_cleaning-0.1.tar.gz" = {}
"dist/mlops_utilities-0.1.tar.gz" = {alias = "tools.lib.utils"}
```
First line results in `src/package_name/data_cleaning` being created, with the contents of tar file inside that folder.
Second line results in `src/package_name/tools/lib/utils` being created, with the contents of tar file inside that folder. It's doing a rename and unpacking at the same time.

`"dist/mlops_utilities-0.1.tar.gz" = {alias = "tools.lib.mlops_utlities"} `

Throws an error:

```
File "/Users/Lorena_balan/anaconda3/envs/packagingcon_dev/lib/python3.8/shutil.py", line 799, in move
raise Error("Cannot move a directory '%s' into itself"
shutil.Error: Cannot move a directory '/private/var/folders/v2/vv9w6n9j0nbb5vgglb8vgkrm0000gn/T/tmpdi3o_mt8/mlops_utilities-0.1/mlops_utilities' into itself '/private/var/folders/v2/vv9w6n9j0nbb5vgglb8vgkrm0000gn/T/tmpdi3o_mt8/mlops_utilities-0.1/mlops_utilities/mlops_utilities'.
```

https://jira.quantumblack.com/browse/KED-2939

## Checklist


lorenabalan:
New `pylint` version is kinda annoying. While fixing that we should also see if we can use the same version for python 3.9 as well (i.e. have only one line in our `test_requirements.txt`).

Also copy-pasting my comment here:
>FWIW I don't agree with a lot of the `pylint` complaints - `use-implicit-booleaness-not-comparison` is a nice idea in theory but for tests we want to make sure it's of the expected type as well, i.e. empty _dictionary_ not `None` or empty string, etc..

kedro/framework/cli/pipeline.py
```diff
@@ -238,18 +244,26 @@ def pull_package(  # pylint:disable=unused-argument, too-many-arguments
...
+        destination=destination,
+        fs_args=fs_args,
+    )
     as_alias = f" as `{alias}`" if alias else ""
     message = f"Pipeline {package_path} pulled and unpacked{as_alias}!"
     click.secho(message, fg="green")
 
 
-# pylint: disable=too-many-locals
+# pylint: disable=too-many-arguments, disable=too-many-locals
```
MerelTheisenQB, line 260:
```suggestion
# pylint: disable=too-many-arguments,too-many-locals
```

kedro/framework/cli/pipeline.py
```diff
@@ -535,12 +555,13 @@ def _move_package_with_conflicting_name(
     return refactored_package_path, refactored_tests_path
 
 
-def _install_files(  # pylint: disable=too-many-locals
+def _install_files(  # pylint: disable=too-many-arguments, disable=too-many-locals
```
MerelTheisenQB, line 558:
```suggestion
def _install_files(  # pylint: disable=too-many-arguments, too-many-locals
```

kedro/framework/cli/pipeline.py
```diff
@@ -211,9 +211,8 @@ def delete_pipeline(
 @env_option(
     help="Environment to install the pipeline configuration to. Defaults to `base`."
 )
-@click.option(
-    "--alias", type=str, default="", help="Module location to unpackage under."
-)
+@click.option("--alias", type=str, default="", help="Rename the package.")
+@click.option("--destination", type=str, default="", help="Module location.")
```
MerelTheisenQB, line 215:
```suggestion
@click.option("--destination", type=str, default="", help="Module location to unpackage under.")
```

tests/framework/cli/pipeline/test_pipeline_pull.py
```diff
@@ -631,6 +702,7 @@ def test_pipeline_pull_all(  # pylint: disable=too-many-locals
         self, fake_repo_path, fake_project_cli, fake_metadata, mocker
     ):
         # pylint: disable=import-outside-toplevel
+        # pylint: disable=line-too-long
```
MerelTheisenQB, line 705: AFAIK `black` should reformat the file to break up the line, so you shouldn't have to disable this.

tests/framework/cli/pipeline/test_pipeline_pull.py
```diff
@@ -175,6 +187,56 @@ def test_pull_local_sdist_compare(
...
+                "tools",
+                "--alias",
+                f"{PIPELINE_NAME}",
+            ],
+            obj=fake_metadata,
+        )
+        assert result.exit_code == 0, result.stderr
+        assert "pulled and unpacked" in result.output
+
+    def test_pipeline_pull_with_destination_only(
```
MerelTheisenQB, line 219: Isn't it possible to add this test case to the parameterised tests above?

MerelTheisenQB (commented):
Some minor comments from my side, but otherwise looks good 🙂 Don't forget to add this to the release notes as well.

RELEASE.md
```diff
@@ -53,7 +53,7 @@
 * The `settings.py` setting `CONF_ROOT` has been renamed to `CONF_SOURCE` to align the API. Default value of `conf` remains unchanged.
 * Renamed `extra_params` to `runtime_params` in `kedro.config.config.ConfigLoader` and `kedro.config.templated_config.TemplatedConfigLoader`.
 * Switched from packaging pipelines as wheel files to tar archive files compressed with gzip (`.tar.gz`)
-* `kedro pipeline package` and `kedro pipeline pull --alias` now accept a module path to the pipeline or utility module to package, relative to the `src/<package_name>/`.
+* `kedro pipeline package` and `kedro pipeline pull --alias --destination` now accept a module name and path to the pipeline or utility module to package, relative to the `src/<package_name>/`.
```
MerelTheisenQB, line 56: `--destination` is a new option for `kedro pipeline pull` right? I would probably add a new bullet point to describe the now option and describe how the purpose of `--alias` has changed.

MerelTheisenQB (approved):
Nice work! 👍  Just a minor comment on the release notes.

RELEASE.md
SajidAlamQB: Since this is develop and hasn't been released do we need to say how it has changed?

MerelTheisenQB: Ahh I see what you mean, no in that case you don't have to mention how `--alias` has changed, but I would change this bullet or add another to also explain the difference between `--alias` and `--destination`

tests/framework/cli/pipeline/test_pipeline_pull.py
```diff
@@ -630,7 +685,7 @@ class TestPipelinePullFromManifest:
     def test_pipeline_pull_all(  # pylint: disable=too-many-locals
         self, fake_repo_path, fake_project_cli, fake_metadata, mocker
     ):
-        # pylint: disable=import-outside-toplevel
+        # pylint: disable=import-outside-toplevel, line-too-long
```
SajidAlamQB, line 688: Black was not reformatting line 698

AntonyMilneQB (commented):
At the moment I think this PR is ok so long as we consider only the simple case `destination=pipelines`, but as soon as we start looking to pull into a nested directory, things get tricky I don't think the current implementation will work right.

I think there's a couple of things we're missing here:
* `destination` should really be expecting a directory path (i.e. with `/`) rather than Python package name (with `.`)
* `destination` should be relative to `src/project_name` (already the case I think)
* we need to make sure that this is working correctly in the nested case of  `destination=pipelines/nested`) - this means both the files are being created in the right place and that the refactoring of python imports is working correctly

tests/framework/cli/pipeline/test_pipeline_pull.py
```diff
@@ -175,6 +192,35 @@ def test_pull_local_sdist_compare(
         assert not filecmp.dircmp(test_path, test_dest).diff_files
         assert source_params_config.read_bytes() == dest_params_config.read_bytes()
 
+    def test_pipeline_pull_same_alias_package_name(
```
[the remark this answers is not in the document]
SajidAlamQB: Yes, this test with only `alias` as `PIPELINE_NAME` fails on develop

kedro/framework/cli/pipeline.py
```diff
@@ -549,10 +572,10 @@ def _install_files(  # pylint: disable=too-many-locals
     )
 
     if conf_source.is_dir() and alias:
-        alias_name = alias.split(".")[-1]
-        _rename_files(conf_source, package_name, alias_name)
+        _rename_files(conf_source, package_name, alias)
 
-    module_path = alias or package_name
+    destination = destination or ""
+    module_path = f"{destination}.{alias or package_name}"
```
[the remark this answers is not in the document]
AntonyMilneQB: ok, I see this is used later on as
```
    artifacts = (
        Path(package_dir, *module_path.split(".")),
        Path(package_dir.parent, "tests", *module_path.split(".")),
        project_conf_path / env,
    )
```
and so if `destination = ""` then I'll get `module_path = ".alias"` which after splitting becomes `["", "alias"]`, which works ok with `Path`. And if `destination = "something/nested"` it will also work ok because I'll get `module_path = "something/nested.alias"`, which after splitting becomes `["something/nested", "alias"]`, which also works ok with `Path`.

Is that right? Seems kind of confusing either way. Can we simplify this somehow? 😅

AntonyMilneQB (approved):
Thanks for the changes! I've just tried this out and it all seems to work well, but I must admit I don't feel 100% confident about it still - could just be my lack of understanding of the details of what's going on here, but I'd feel more comfortable if we had a new test case that covered doing a nested destination like `pipelines/nested`. Also a couple of questions/confusions left in the comments - would be very grateful if @lorenabalan could have take a look also.

kedro/framework/cli/pipeline.py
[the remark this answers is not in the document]
SajidAlamQB: Yes, exactly how it worked. I will have a look to see if we can simplify this.

tests/framework/cli/pipeline/test_pipeline_pull.py
```diff
@@ -175,6 +195,64 @@ def test_pull_local_sdist_compare(
...
+                str(sdist_file),
+                "--destination",
+                "tools",
+                "--alias",
+                PIPELINE_NAME,
+            ],
+            obj=fake_metadata,
+        )
+        assert result.exit_code == 0, result.stderr
+        assert "pulled and unpacked" in result.output
```
lorenabalan, line 225: We need to add more checks to these tests, to make sure the expected directory is created and the expected contained files exist. If you look at examples above we do `assert_package_files_exist` etc.

kedro/framework/cli/pipeline.py
```diff
@@ -211,8 +211,13 @@ def delete_pipeline(
...
     help="Environment to install the pipeline configuration to. Defaults to `base`."
 )
+@click.option("--alias", type=str, default="", help="Rename the package.")
 @click.option(
-    "--alias", type=str, default="", help="Module location to unpackage under."
+    "-d",
+    "--destination",
+    type=click.Path(file_okay=False, dir_okay=False),
+    default=None,
+    help="Module location where to unpack under.",
```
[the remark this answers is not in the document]
lorenabalan: Let's leave it relative to `src/package_name` and we can try a change in a separate PR cause I'm not 100% `rope` will like the path without `__init__.py` in it. 
The point you raised about `click.Path` checks though is valid, and we should validate that in our testing.

kedro/framework/cli/pipeline.py
```diff
@@ -549,10 +576,13 @@ def _install_files(  # pylint: disable=too-many-locals
...
 
     if conf_source.is_dir() and alias:
-        alias_name = alias.split(".")[-1]
-        _rename_files(conf_source, package_name, alias_name)
+        _rename_files(conf_source, package_name, alias)
+
+    if destination:
+        module_path = f"{destination}.{alias or package_name}"
+    else:
+        module_path = f"{alias or package_name}"
```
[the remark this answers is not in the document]
lorenabalan: And in fact, we can make this easier on ourselves and make `module_path` be an actual `Path` object. Then this code becomes `module_path = Path(alias or package_name)` etc. 

Maybe we can also turn `destination` into a `Path` as quickly as possible, rather than pass it around as a string until the very end.

lorenabalan (commented):
I think that, in some places, we've shoehorned the fix into the old ways of working and it's made the code harder to reason about. If the tests pass (once you add the additional checks), I'm happy to merge this as is and I'll raise a follow-up cleanup PR if I can refactor some things.

As far as I understand, the ideal state is:
1. `kedro pipeline package <> --alias A` and `kedro pipeline pull <> --alias B` symmetry, i.e. A and B are simple strings, used to rename a package and refactor its imports
2. `kedro pipeline package <> --destination A` and `kedro pipeline pull <> --destination B` symmetry, i.e. A and B look like filepaths (`/`-separated), ideally both relative to _the same location_ (so ideally relative to the project root) **if possible**. This destination is the location where to _output_ the result of the relevant command, i.e. the tarfile or the unpacked source code.

It's my bad for not explaining/remembering this better and for letting the scope of this PR be increased mid-development without realising it'd make everything more confusing.

docs/source/nodes_and_pipelines/modular_pipelines.md
```diff
@@ -199,7 +199,7 @@ To pull multiple modular pipelines in bulk, run `kedro pipeline pull --all`. Thi
 ```toml
 [tool.kedro.pipeline.pull]
 "dist/first-pipeline-0.1.tar.gz" = {}
-"https://www.url.to/second-pipeline-0.1.tar.gz" = {alias = "pipelines.aliased_pipeline", fs-args = "pipeline_pull_args.yml"}
+"https://www.url.to/second-pipeline-0.1.tar.gz" = {alias = "aliased_pipeline", destination = "somewhere/else", fs-args = "pipeline_pull_args.yml"}
```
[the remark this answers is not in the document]
SajidAlamQB: `pipeline package` had a similar a similar example so wanted to keep it consistent.

AntonyMilneQB:
> I think that, in some places, we've shoehorned the fix into the old ways of working and it's made the code harder to reason about. If the tests pass (once you add the additional checks), I'm happy to merge this as is and I'll raise a follow-up cleanup PR if I can refactor some things.
> 
> As far as I understand, the ideal state is:
> 
> 1. `kedro pipeline package <> --alias A` and `kedro pipeline pull <> --alias B` symmetry, i.e. A and B are simple strings, used to rename a package and refactor its imports
> 2. `kedro pipeline package <> --destination A` and `kedro pipeline pull <> --destination B` symmetry, i.e. A and B look like filepaths (`/`-separated), ideally both relative to _the same location_ (so ideally relative to the project root) **if possible**. This destination is the location where to _output_ the result of the relevant command, i.e. the tarfile or the unpacked source code.
> 
> It's my bad for not explaining/remembering this better and for letting the scope of this PR be increased mid-development without realising it'd make everything more confusing.

@lorenabalan thanks for looking at this and the comments - agree with all of it. For a separate PR/ticket/discussion, but just for completeness let me note it here. If `destination` is a filepath relative to project root I think the main argument of `kedro pipeline package` looks increasingly weird being `pipelines.data_science` rather than `src/spaceflights/pipelines/data_science`. This seems much more natural and principle of least astonishment to me, especially given how common it is to give a file path by command line (tabbed completion etc.):
* if `alias` and `destination` are symmetric with `kedro pipeline pull` then I'd expect the main argument to be symmetric also, and `kedro pipeline pull` expects a filepath relative to project root as main argument. Nothing else expects a module path with `.`
* we're no longer using `alias` to be a composite `something.blah`, so it's not like we're rewriting `pipelines.data_science` to `something.blah` any more

lorenabalan:
@AntonyMilneQB the asymmetry between main arguments is something we chose to live with for now and act later once we have more data and feedback from our users... We just need to get sth out in front of our users and iterate on it once that's done. Once again, the `.` had good reasons behind it (highlighting that it's a PYTHON package, and not packaging any random paths), but we can see how this evolves over time, there's no point in overdesigning up-front now.

PR #1087, opened 2021-12-02
Correct the new dataset example

daniel-falk:
## Description
This commit makes sure that the examples in the _custom dataset_ documentation can run without error or exceptions.

The diff view in ../meta/images/diffs-graphic.png would also need to be updated, it is incorrect in two ways:
* _get_filepath_str()_ should be used in both __save()_ and __load()_ method even after changing to the _AbstractVersionedDataSet_ base class.
* The _AbstractDataSet_ does not have any __get_load_path()_ method in contrast to the _AbstractVersionedDataSet_.

## Development notes

The example to create a new dataset had some different issues:

* __save()_ and __describe()_ methods was missing before they were implemented which causes an exception when trying to instantiate the class, instead keep the empty definitions.
* The _AbstractDataSet_ does not have any __get_load_path()_ method in contrast to the _AbstractVersionedDataSet_.
* Adds missing imports from typing.
* Remove unused import from typing.
* Imports _AbstractDataSet_ directly from _kedro.io_ instead of _kedro.io.core_ to align it with the diff-view in the docs.

## Checklist

### DIFF
--- a/docs/source/07_extend_kedro/03_custom_datasets.md
+++ b/docs/source/07_extend_kedro/03_custom_datasets.md
@@ -36,7 +36,7 @@ Here is an example skeleton for `ImageDataSet`:
 <summary><b>Click to expand</b></summary>
 
 ```python
-from typing import Any, Dict, List
+from typing import Any, Dict
 
 import numpy as np
 
@@ -99,12 +99,10 @@ Here is the implementation of the `_load` method using `fsspec` and `Pillow` to 
 
 ```python
 from pathlib import PurePosixPath
+from typing import Any, Dict
 
-from kedro.io.core import (
-    AbstractDataSet,
-    get_filepath_str,
-    get_protocol_and_path,
-)
+from kedro.io import AbstractDataSet
+from kedro.io.core import get_filepath_str, get_protocol_and_path
 
 import fsspec
 import numpy as np
@@ -133,10 +131,18 @@ class ImageDataSet(AbstractDataSet):
             Data from the image file as a numpy array
         """
         # using get_filepath_str ensures that the protocol and path are appended correctly for different filesystems
-        load_path = get_filepath_str(self._get_load_path(), self._protocol)
+        load_path = get_filepath_str(self._filepath, self._protocol)
         with self._fs.open(load_path) as f:
             image = Image.open(f).convert("RGBA")
             return np.asarray(image)
+
+    def _save(self, data: np.ndarray) -> None:
+        """Saves image data to the specified filepath"""
+        ...
+
+    def _describe(self) -> Dict[str, Any]:
+        """Returns a dict that describes the attributes of the dataset"""
+        ...
 ```
 </details>
 
@@ -177,7 +183,7 @@ class ImageDataSet(AbstractDataSet):
     def _save(self, data: np.ndarray) -> None:
         """Saves image data to the specified filepath."""
         # using get_filepath_str ensures that the protocol and path are appended correctly for different filesystems
-        save_path = get_filepath_str(self._get_save_path(), self._protocol)
+        save_path = get_filepath_str(self._filepath, self._protocol)
         with self._fs.open(save_path, "wb") as f:
             image = Image.fromarray(data)
             image.save(f)


docs/source/07_extend_kedro/03_custom_datasets.md
```diff
@@ -255,23 +258,23 @@ class ImageDataSet(AbstractDataSet):
...
+        load_path = get_filepath_str(self._filepath, self._protocol)
         with self._fs.open(load_path, mode="r") as f:
             image = Image.open(f).convert("RGBA")
             return np.asarray(image)
 
     def _save(self, data: np.ndarray) -> None:
         """Saves image data to the specified filepath."""
-        # using get_filepath_str ensures that the protocol and path are appended correctly for different filesystems
-        save_path = get_filepath_str(self._get_save_path(), self._protocol)
+        save_path = get_filepath_str(self.filepath, self._protocol)
```
AntonyMilneQB, line 268:
```suggestion
        save_path = get_filepath_str(self._filepath, self._protocol)
```

docs/source/07_extend_kedro/03_custom_datasets.md
```diff
@@ -390,9 +397,85 @@ class ImageDataSet(AbstractVersionedDataSet):
...
+        """
+-        load_path = get_filepath_str(self._filepath, self._protocol)
++        load_path = get_filepath_str(self._get_load_path(), self._protocol)
+        with self._fs.open(load_path, mode="r") as f:
+            image = Image.open(f).convert("RGBA")
+            return np.asarray(image)
+
+    def _save(self, data: np.ndarray) -> None:
+        """Saves image data to the specified filepath."""
+-        save_path = get_filepath_str(self.filepath, self._protocol)
```
AntonyMilneQB, line 465:
```suggestion
-        save_path = get_filepath_str(self._filepath, self._protocol)
```

AntonyMilneQB:
@daniel-falk Thanks very much for pointing this out. I've completed the changes on your branch by making sure everything is consistent. I've also removed the diff graphic and replaced it with a code version which is up to date and will be easier to maintain in future. Here's what it looks like:

image

kedro/extras/datasets/spark/deltatable_dataset.py
```diff
@@ -1,4 +1,4 @@
-"""``AbstractVersionedDataSet`` implementation to access DeltaTables using
+"""``AbstractDataSet`` implementation to access DeltaTables using
```
AntonyMilneQB, line 1: Completely unrelated - just a small fix.

docs/source/07_extend_kedro/03_custom_datasets.md
```diff
@@ -36,11 +36,11 @@ Here is an example skeleton for `ImageDataSet`:
 <summary><b>Click to expand</b></summary>
 
 ```python
-from typing import Any, Dict, List
+from typing import Any, Dict
 
 import numpy as np
 
-from kedro.io import AbstractDataSet
+from kedro.io.core import AbstractDataSet
```
lorenabalan, line 43:
```suggestion
from kedro.io import AbstractDataSet
```

This is a good catch. We should actually prefer the top-level imports so that if we rename or move `core`, this import would still work.. but if it's a faff to recreate the screenshot then I'm not that bothered.

docs/source/07_extend_kedro/03_custom_datasets.md
```diff
@@ -133,10 +132,18 @@ class ImageDataSet(AbstractDataSet):
...
             image = Image.open(f).convert("RGBA")
             return np.asarray(image)
+
+    def _save(self, data: np.ndarray) -> None:
+        """Saves image data to the specified filepath"""
+        ...
+
+    def _describe(self) -> Dict[str, Any]:
+        """Returns a dict that describes the attributes of the dataset"""
+        ...
```
lorenabalan, line 146: I would actually remove these lines. I believe it was intentional to only focus on one function at a time, rather than include the boilerplate at every stage. The very first step is to define the outline of the dataset, which includes empty `load`, `save` and `describe` methods (L46-L77).

docs/source/07_extend_kedro/03_custom_datasets.md
```diff
@@ -168,16 +175,11 @@ Similarly, we can implement the `_save` method as follows:
 
 
 ```python
-import numpy as np
-from PIL import Image
-from kedro.io.core import AbstractDataSet, get_filepath_str
-
```
lorenabalan, line 174: Did you remove these because they are already included in the code snippet above?

docs/source/07_extend_kedro/03_custom_datasets.md
```diff
@@ -255,16 +256,14 @@ class ImageDataSet(AbstractDataSet):
         Returns:
             Data from the image file as a numpy array
         """
-        # using get_filepath_str ensures that the protocol and path are appended correctly for different filesystems
```
lorenabalan, line 258: Why remove these comments?

docs/source/07_extend_kedro/03_custom_datasets.md
```diff
@@ -255,16 +256,14 @@ class ImageDataSet(AbstractDataSet):
...
-        load_path = get_filepath_str(self._get_load_path(), self._protocol)
+        load_path = get_filepath_str(self._filepath, self._protocol)
         with self._fs.open(load_path, mode="r") as f:
```
lorenabalan, line 260: Just realised, it'd be good to update these `open` calls as well - with newer version of `pylint` or `flake8` they make you explicitly use `encoding`, so we can add an `encoding="utf-8"` while we're here.

docs/source/07_extend_kedro/03_custom_datasets.md
AntonyMilneQB: This is a good point, I'll change it. It's no faff to recreate the screenshot any more since I made it a code `diff` rather than a picture 😀

docs/source/07_extend_kedro/03_custom_datasets.md
AntonyMilneQB: Agree with this - I will revert this change.

docs/source/07_extend_kedro/03_custom_datasets.md
AntonyMilneQB: Yes, exactly. We weren't being consistent about whether we repeated the imports in every example or not. I simplified it by just giving them in the first example and the complete example.

docs/source/07_extend_kedro/03_custom_datasets.md
AntonyMilneQB: To clear up the diff compared to the versioned example (which is now _sort of_ automatically generated). They are already given as comments in the individual examples and so don't don't really add anything when showing the complete example.

docs/source/07_extend_kedro/03_custom_datasets.md
AntonyMilneQB: Good call - will do!

AntonyMilneQB: Actually looking through our codebase it looks like we don't put `encoding` on `self._fs.open` calls, just on `open` function calls?

docs/source/07_extend_kedro/03_custom_datasets.md
```diff
@@ -390,9 +393,87 @@ class ImageDataSet(AbstractVersionedDataSet):
...
+             image.save(f)
+
+     def _describe(self) -> Dict[str, Any]:
+         """Returns a dict that describes the attributes of the dataset."""
+-        return dict(filepath=self._filepath, protocol=self._protocol)
++        return dict(
++            filepath=self._filepath, version=self._version, protocol=self._protocol
++        )
+```
+</details>
```
MerelTheisenQB, line 476: This is so much better than an image! 👏

docs/source/07_extend_kedro/03_custom_datasets.md
lorenabalan: Doesn't matter then, thanks for checking.

docs/source/07_extend_kedro/03_custom_datasets.md
AntonyMilneQB: Done in 7b1a46b

docs/source/07_extend_kedro/03_custom_datasets.md
AntonyMilneQB: Done in 7b1a46b

daniel-falk: I understand your intent to focus on the changes and that's good. I do however find it very confusing in this specific example since the code block looks like a complete file. Some of the other code blocks are more obvious that they are only small snippets, but not this one. In the last code snippet, the one on handling credentials, the ellipsis object ("...") is used to indicate that there are more methods in the class. That could be used here too?

daniel-falk (commented):
Excellent changes, thanks for spending time on easy to understand documentation!

docs/source/07_extend_kedro/03_custom_datasets.md
AntonyMilneQB: Fair point, thank you for explaining @daniel-falk. I like your proposal of `...` - seems like a good compromise between focusing on one function at a time and indicating that it's not complete code. I've done this in 7db9b72481664cf3f01ad52eae93cabf295e7229.

lorenabalan (approved):

MerelTheisenQB (approved):
LGTM 👍

PR #1093, opened 2021-12-03
[e2e] Fix e2e tests on windows

limdauto:
## Description

Currently master is never green. This fixes the issue :) 

Also added a fix to accommodate recent pylint release

## Checklist


kedro/io/partitioned_data_set.py
```diff
@@ -439,7 +439,7 @@ def _parse_checkpoint_config(
         if self._credentials:
             default_config[CREDENTIALS_KEY] = deepcopy(self._credentials)
 
-        if CREDENTIALS_KEY in default_config.keys() & checkpoint_config.keys():
+        if CREDENTIALS_KEY in (default_config.keys() & checkpoint_config.keys()):
```
arnaldog12, line 442: I may be wrong, but I believe this condition will never be true. I did some tests and looks like the `&` operator returns an empty set. Please, see the screenshot below:

Screen Shot 2021-12-03 at 15 16 41

limdauto: @arnaldog12 Hey these are set-like views, so `&` only returns non empty set if the intersection is not empty.

Screenshot 2021-12-03 at 18 21 14

limdauto: I think there is a test that covers this case. In this PR I just want to make it compatible with the pylint release 40 minutes ago but I give up. I will file a ticket on pylint.

features/environment.py
```diff
@@ -110,10 +110,10 @@ def _setup_minimal_env(context):
             "pip>=20.0",
             "setuptools>=38.0",
             "wheel",
-            ".",
         ],
         env=context.env,
     )
+    call([context.python, "-m", "pip", "install", "."], env=context.env)
```
lorenabalan, line 116: Why does this work?... 😂

lorenabalan (approved):
Thank you!! 🎊

MerelTheisenQB (approved):
Wow that's the simplest and smartest fix 😂 🥇

PR #1094, opened 2021-12-05
Ensure path is of type `str` if `overwrite` is set

### LINKED ISSUE #1085: MatplotlibWriter overwrite list of figures does not work.
## Description
The latest MatplotlibWriter (which has not been released as stable yet) has an error here:
(https://github.com/quantumblacklabs/kedro/blob/b76707669cde0b860389cd24a4397cbce001d13c/kedro/extras/datasets/matplotlib/matplotlib_writer.py#L149)

Where self._filepath is an object of 'PurePosixPath' which is not iterable.

## Context
I wanted to output a list of matplotlib figures using this dataset and using the new attribute **overwrite=True**.

## Steps to Reproduce
Here is a snippet that reproduces the error:
```
from kedro.extras.datasets.matplotlib import MatplotlibWriter
import matplotlib.pyplot as plt

# We run save twice so that the figure is overwritten the second time:
for i in range(2):

    dataset = MatplotlibWriter(filepath='test', overwrite=True)
    figs = []
    for i in range(3):
        fig,ax=plt.subplots()
        ax.plot([1,2,3])
        figs.append(fig)
        
    dataset.save(figs)
```

```
TypeError: 'PurePosixPath' object is not iterable
```

# Solution suggestion:
Convert the PurePosixPath object to string solves the problem:
`self._fs.rm(str(self._filepath), recursive=True)`

## Your Environment

* Kedro (latest) (current stable 0.17.5 does not have **overwrite** attribute):
* Python version used 3.7
* Operating system and version: Win10

deepyaman:
## Description
Closes #1085 

## Checklist

### DIFF
--- a/kedro/extras/datasets/matplotlib/matplotlib_writer.py
+++ b/kedro/extras/datasets/matplotlib/matplotlib_writer.py
@@ -146,7 +146,9 @@ class MatplotlibWriter(AbstractVersionedDataSet):
             and self._overwrite
             and self._fs.exists(self._filepath)
         ):
-            self._fs.rm(self._filepath, recursive=True)
+            self._fs.rm(
+                get_filepath_str(self._filepath, self._protocol), recursive=True
+            )
 
         if isinstance(data, list):
             for index, plot in enumerate(data):
--- a/kedro/extras/datasets/matplotlib/matplotlib_writer.py
+++ b/kedro/extras/datasets/matplotlib/matplotlib_writer.py
@@ -5,6 +5,7 @@ import io
 from copy import deepcopy
 from pathlib import PurePosixPath
 from typing import Any, Dict, List, Union
+from warnings import warn
 
 import fsspec
 import matplotlib.pyplot as plt
@@ -123,6 +124,13 @@ class MatplotlibWriter(AbstractVersionedDataSet):
         if save_args is not None:
             self._save_args.update(save_args)
 
+        if overwrite and version is not None:
+            warn(
+                "Setting `overwrite=True` is ineffective if versioning "
+                "is enabled, since the versioned path must not already "
+                "exist; overriding flag with `overwrite=False` instead."
+            )
+            overwrite = False
         self._overwrite = overwrite
 
     def _describe(self) -> Dict[str, Any]:
--- a/kedro/extras/datasets/matplotlib/matplotlib_writer.py
+++ b/kedro/extras/datasets/matplotlib/matplotlib_writer.py
@@ -152,7 +152,7 @@ class MatplotlibWriter(AbstractVersionedDataSet):
         if (
             isinstance(data, (list, dict))
             and self._overwrite
-            and self._fs.exists(save_path)
+            and self._exists()
         ):
             self._fs.rm(get_filepath_str(save_path, self._protocol), recursive=True)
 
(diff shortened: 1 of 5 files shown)


kedro/extras/datasets/matplotlib/matplotlib_writer.py
```diff
@@ -146,7 +146,9 @@ def _save(
             and self._overwrite
             and self._fs.exists(self._filepath)
         ):
-            self._fs.rm(self._filepath, recursive=True)
+            self._fs.rm(
+                get_filepath_str(self._filepath, self._protocol), recursive=True
```
AntonyMilneQB, line 150: Following my comment in #1085 I think what we're interested in here (in both lines 147 and 150) would be `get_filepath_str(save_path, self._protocol)`. Does that make sense to you also?

AntonyMilneQB (commented):
Thanks for fixing this! 

Looks like Lim fixed the pylint stuff in #1093 so if you get the latest `master` then those changes should no longer be needed.

kedro/extras/datasets/matplotlib/matplotlib_writer.py
```diff
@@ -144,9 +152,9 @@ def _save(
         if (
             isinstance(data, (list, dict))
             and self._overwrite
-            and self._fs.exists(self._filepath)
+            and self._fs.exists(save_path)
```
AntonyMilneQB, line 155:
```suggestion
            and self._fs.exists(get_filepath_str(save_path, self._protocol))
```
OR
```suggestion
            and self._exists()
```
since we now know that if `overwrite` is enabled, versioning can't be and hence `self._filepath == self._get_save_path() == self._get_load_path()`.

AntonyMilneQB (approved):
Looks good - thanks very much for fixing this 🙇  I've tested this manually and all seems good now. Just one small comment.

kedro/io/core.py
```diff
@@ -704,7 +704,7 @@ def get_protocol_and_path(filepath: str, version: Version = None) -> Tuple[str,
     protocol = options_dict["protocol"]
 
     if protocol in HTTP_PROTOCOLS:
-        if version:
+        if version is not None:
```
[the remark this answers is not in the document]
deepyaman: Either way technically works. I did it for consistency, because I used `if version is not None` in the `MatplotibWriter`.

I think we also do explicitly mean to check `is not None` (that's what the docstring also says), and PEP-8 says, "Also, beware of writing if x when you really mean if x is not None." Even though a `Version` object can't evaluate to False, felt this may be more explicit/safer.

lorenabalan (approved):
LGTM

kedro/extras/datasets/matplotlib/matplotlib_writer.py
```diff
@@ -141,12 +149,8 @@ def _save(
     ) -> None:
         save_path = self._get_save_path()
 
-        if (
-            isinstance(data, (list, dict))
-            and self._overwrite
-            and self._fs.exists(self._filepath)
-        ):
-            self._fs.rm(self._filepath, recursive=True)
+        if (isinstance(data, (list, dict)) and self._overwrite and self._exists()):
```
deepyaman, line 152:
```suggestion
        if isinstance(data, (list, dict)) and self._overwrite and self._exists():
```

PR #1102, opened 2021-12-08
[KED-2961] Tidy CircleCI config

AntonyMilneQB:
## Description
I started moving the `private-kedro-sync` scripts over to this repo but it got very messy very quickly. So I've decided to do a tidy up of what we've got first before we add even more stuff to it. The result is we now have 150 fewer lines in the CircleCI config and it's (I hope) easier to follow and add the new jobs to.

Functionally this is almost identical to what we had before. These are the checks we were doing:
image

The only difference is that I've removed the one marked as **removed** since I don't think it was there on purpose. Aside from that, all the same checks as before are still there and run identically to before (small exception: I removed install of `pre-push` but this shouldn't change anything - see inline comments).

## Development notes
General principles of the refactoring:
- I've reduced the number of "layers of indirection" by moving stuff from `commands` to `jobs`
- I've used `matrix` so we don't need a separate job for every version of Python we're testing
- this extends the use of `parameters`, which we were already using for Windows, to also choose the docker image

Tested by checking through all the CircleCI jobs - all seem to be working correctly. 
## Checklist


.circleci/config.yml
```diff
@@ -3,23 +3,19 @@ version: 2.1
...
-      - image: 350138855857.dkr.ecr.eu-west-2.amazonaws.com/kedro-builder:3.7
-    resource_class: medium+
-  py38:
+  docker:
+    parameters:
+      python_version:
+        type: string
     docker:
-      - image: 350138855857.dkr.ecr.eu-west-2.amazonaws.com/kedro-builder:3.8
+      - image: 350138855857.dkr.ecr.eu-west-2.amazonaws.com/kedro-builder:<<parameters.python_version>>
```
AntonyMilneQB, line 14: These are exactly the same `executors` as before, just labelled differently so that we can use the `matrix` notation.

.circleci/config.yml
```diff
@@ -3,23 +3,19 @@ version: 2.1
...
     resource_class: medium+
 
 commands:
   setup_conda:
-    description: Activate conda environment
```
AntonyMilneQB, line 22: These descriptions weren't really consistent throughout the file and don't appear anywhere on CircleCI (the human readable `name` does instead, which everything has already) and the names of all the commands are now pretty self-explanatory so I removed them.

.circleci/config.yml
```diff
@@ -62,92 +57,14 @@ commands:
...
 
-  setup_pre_commit:
-    description: Install pre-commit hooks
-    steps:
-      - run:
-          name: Install pre-commit hooks
-          command: pre-commit install --install-hooks
-      - run:
-          name: Run pre-commit hooks
-          command: pre-commit install --hook-type pre-push
```
AntonyMilneQB, line 73: I removed this `pre-push` bit since we don't have any such hooks. Everything else from this removed code chunk is still there, just moved to `jobs`.

.circleci/config.yml
```diff
@@ -62,92 +57,14 @@ commands:
...
-          command: make lint
-
-  e2e_tests:
-    description: Run all end to end tests
+  setup:
     steps:
       - checkout
       - setup_conda
       - setup_requirements
-      - run:
```
AntonyMilneQB, line 102: All this has been moved to `jobs`, nothing changed functionally.

.circleci/config.yml
```diff
@@ -239,41 +209,68 @@ commands:
...
           command: setx /m HDF5_DISABLE_VERSION_CHECK 1
-      - run:
-          name: Run unit tests
-          command: |
-            conda activate kedro_builder
-            pytest .\tests --ignore .\tests\extras\datasets\spark --ignore .\tests\extras\datasets\tensorflow --ignore tests\framework\session\test_session_hooks.py --no-cov
+      - when:
+          # Run `test_parallel_runner.py` separately because of `Windows fatal exception: stack overflow`
+          condition:
+            equal: ["3.8", <<parameters.python_version>>]
```
AntonyMilneQB, line 215: This conditional behaviour was there before since `win_unit_tests_38` was a bit different from `win_unit_tests_36` and `win_unit_tests_37`.

.circleci/config.yml
```diff
@@ -239,41 +209,68 @@ commands:
...
-          key: kedro-deps-v1-win-{{ checksum "requirements.txt" }}-{{ checksum "test_requirements.txt" }}
-      - win_setup_requirements
+      - win_setup:
+          python_version: <<parameters.python_version>>
       - when:
-          # Cache when `parameters.cache_save` is True
-          condition: << parameters.cache_save >>
+          # Save cache only for Python 3.7. There is no need to save it for each Python.
+          condition:
+            equal: ["3.7", <<parameters.python_version>>]
```
AntonyMilneQB, line 273: Again the same behaviour as before, just it's a bit simpler now since we don't need the `parameters.cache_save` variable any more (previously it was just a proxy for when python version = 3.7).

.circleci/config.yml
```diff
@@ -285,234 +282,101 @@ commands:
...
-      - win_pip_compile_37
-      - win_pip_compile_38
-      - win_e2e_tests_36
-      - win_e2e_tests_37
-      - win_e2e_tests_38
-      - run_kedro_viz:
+      - e2e_tests:
+          matrix:
+            parameters:
+              python_version: ["3.6", "3.7", "3.8"]
```
AntonyMilneQB, line 334: This effectively does the following:
```
jobs:
  - e2e_tests:
     name: e2e_tests-3.6
     python_version: 3.6
  - e2e_tests:
     name: e2e_tests-3.7
     python_version: 3.7
  - e2e_tests:
     name: e2e_tests-3.8
     python_version: 3.8
```
i.e. invokes the `e2e_tests` job three times, once for each `python_version`.

.circleci/config.yml
```diff
@@ -285,234 +282,101 @@ commands:
...
-            - win_unit_tests_37
-            # Skipped due to `pywin32 is in an unsupported or invalid wheel`
-            # - win_e2e_tests_36
-            # Skipped due to Windows fatal exception: stack overflow
-            # - win_unit_tests_38
-            - win_e2e_tests_37
-            - win_e2e_tests_38
+            # Should just have e2e_tests and win_e2e_tests but we need to skip
+            # win_e2e_tests-3.6 due to `pywin32 is in an unsupported or invalid wheel`
+            - e2e_tests
```
AntonyMilneQB, line 370: This means that all tests in the `e2e_tests` matrix are required, i.e. it's equivalent to
```
- e2e_tests-3.6
- e2e_tests-3.7
- e2e_tests-3.8
```

MerelTheisenQB (commented):
This does look better to me, but it is a bit hard to tell. All the builds pass so I trust it's all working as expected.

The only thing you have to be aware of is that you might have to do it manually for `develop` as well because I can imagine the merge conflicts will be hard to solve when master has to be merged into develop.

.circleci/config.yml
```diff
@@ -285,234 +282,101 @@ commands:
...
+            - win_e2e_tests-3.7
+            - win_e2e_tests-3.8
+            # Should just have unit_tests and win_unit_tests but we need to skip
+            # win_unit_tests-3.8 due to `Windows fatal exception: stack overflow`
```
lorenabalan, line 374: We can put back `win_e2e_tests-3.6` as @limdauto fixed them. And actually I haven't seen the windows unit test failure on 3.8 in a long time, so I'm happy to have both back in, and simplify this to be
```yaml
- unit_tests
- win_unit_tests
- win_e2e_tests
```

lorenabalan (approved):
This looks great, I love how much we've cut down on the fluff! I had some minor comments but otherwise happy with this, thank you for looking into this! 👏 🧹 ✨

.circleci/config.yml
AntonyMilneQB: Awesome! Done in 005ef88 - much tidier now.

.circleci/config.yml
```diff
@@ -157,11 +74,12 @@ commands:
           command: conda init powershell
       - run:
           name: Create 'kedro_builder' conda environment
-          command: |
-            conda create --name kedro_builder python=<< parameters.python_version >> -y
+          command: conda create -n kedro_builder python=<<parameters.python_version>> -y
+      - run:
+          name: Install 'make' command
+          command: choco install make
```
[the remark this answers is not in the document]
AntonyMilneQB: Very fair comment! Done in 999f867

AntonyMilneQB:
I've managed to [simplify `win_unit_tests` a bit more](https://github.com/quantumblacklabs/kedro/pull/1102/files/999f867e5cc525b64e50e211d9fad689d0a547f9..79e0c3156a32adf16c4a879f0b30ac7c6c63f89c).

* The special case for Python = 3.8 wasn't needed; tests passed fine with `pytest .\tests --ignore .\tests\extras\datasets\spark --ignore .\tests\extras\datasets\tensorflow --ignore tests\framework\session\test_session_hooks.py --no-cov` on all versions
* I've altered this to `make test-no-spark`, removed pyproject_no_spark.toml and passed `--no-cov` to `make test-no-spark` instead (which is what we were using anyway in the above command)
* Tensorflow tests actually were working on 3.7 and 3.8 but not 3.6 on windows. I've added a pytest skip on that version, which is consistent with how we skip windows tests elsewhere (e.g. tests_session_extension_hooks) and nicer than conditionals in the CircleCI config

.circleci/config.yml
```diff
@@ -239,41 +212,62 @@ commands:
...
+            - run:
+                name: Run unit tests without spark
+                command: conda activate kedro_builder; make test-no-spark
+      - when:
+          condition:
+            equal: ["3.6", <<parameters.python_version>>]
+          steps:
+            - run:
+                name: Run unit tests without spark or tensorflow
+                command: conda activate kedro_builder; pytest tests --no-cov --ignore tests/extras/datasets/spark --ignore tests/extras/datasets/tensorflow --numprocesses 4 --dist loadfile
```
lorenabalan, line 228: > Tensorflow tests actually were working on 3.7 and 3.8 but not 3.6 on windows. I've added a special case for 3.6 to do what is effectively make test-no-spark but also ignoring tensorflow tests. When I do a follow up PR to develop this special conditional bit can be removed

So on Windows we *never* run spark tests, and additionally on python 3.6 specifically we also don't run tensorflow tests. But previously we were not running neither spark nor tensorflow tests on *all* python versions on Windows? Is that right?

lorenabalan (commented):
This generally looks good to me but it was very hard to follow the latest set of changes! Moving forward can we please separate functional changes from refactoring changes even for CI config? I expected the change in builds to be either in a separate PR or in one compact commit so I can see the diff better.

.circleci/config.yml
AntonyMilneQB: Yes, exactly. So now we are running tensorflow tests on 3.7 and 3.8 since they (seem to) work.

AntonyMilneQB:
> 

Yes, agreed - sorry! Unfortunately by its nature this CI stuff requires a lot of different pushes to test stuff out, hence the many, many commits. I tried to make it clearer by linking to just the diff that the latest changes produced: [simplify `win_unit_tests` a bit more](https://github.com/quantumblacklabs/kedro/pull/1102/files/999f867e5cc525b64e50e211d9fad689d0a547f9..d5df507247f095fcb33be4f7e3340038fa496d2d).

idanov (approved):
This is awesome 🚀 It is a bit of a large diff for one PR indeed, but it will make our life so much easier going forward and especially when adding and removing python versions from the build. The "matrix" functionality in CCI is quite nice.

PR #1103, opened 2021-12-08
[KED-2933] Keep the logging config in a global variable in kedro.framework.project

SajidAlamQB:
## Description

Configure logging at the top of the framework setup workflow, and make it accessible as a global variable `from kedro.framework.project import LOGGING`, similar to how PACKAGE_NAME is configured:

https://github.com/quantumblacklabs/kedro/blob/master/kedro/framework/project/__init__.py#L194-L199

As a result, the ParallelRunner shouldn't need access to the session to get the logging configuration (in order to bootstrap each parallel subprocess), but simply import global variable. See the code for ParallelRunner, especially

https://github.com/quantumblacklabs/kedro/blob/master/kedro/runner/parallel_runner.py#L287-L308

Related to: https://jira.quantumblack.com/browse/KED-2780
https://jira.quantumblack.com/browse/KED-2933

## Checklist


SajidAlamQB:
The logging config is quite tightly related to the `ConfigLoader`, in this PR we tried to re-create the logic used by `config.py` in `project` by importing `_get_config_from_patterns`. Although this solution somewhat works `_get_config_from_patterns` and the like are intended as methods/interfaces for `ConfigLoader` so i'm unsure if this is a suitable solution? We also import `_convert_paths_to_absolute_posix` from `context` but @MerelTheisenQB agreed this could probably be moved to `utils`.

Some discussions are needed on the best approach for this.

MerelTheisenQB:
To add to @SajidAlamQB 's comment above: the tricky part here is that to instantiate logging, the `ConfigLoader` is needed to fetch the logging configuration files. We looked at how we can achieve the same thing without instantiating a `ConfigLoader` but instead moving/calling the necessary methods from `project/__init__.py`. But as Sajid explained the methods called are still very tightly coupled to the config loader and so the question is whether we should be calling those methods at this point in the flow at all.

kedro/framework/session/session.py
```diff
@@ -209,6 +210,7 @@ def _setup_logging(self) -> None:
         """Register logging specified in logging directory."""
         conf_logging = self._get_logging_config()
         logging.config.dictConfig(conf_logging)
+        configure_logging(conf_logging)
```
lorenabalan, line 213: This to me feels weird. If we're getting rid of `get_current_session` altogether, it'll mean everytime someone creates a session (even just to get the store or the context - if that'll still be a thing), they'll re-configure logging. In my head this was done once, at the top of the workflow, when we `bootstrap_project()`.

lorenabalan: @idanov

idanov: That's ok though, since this is the current behaviour anyways if you aren't using the `get_current_session`. We cannot do it at `bootstrap_project()` time because that's done only for non-packaged projects and at the time `bootstrap_project()` is called, we have no idea yet how and where to load the logging config from.

The logging config is setup only later at runtime on purpose, e.g. when deploying your code, theoretically you might need to change the logging config to suit your set of tools in production that consume your logs. The only lifecycle that is guaranteed to exist at runtime is the session initialisation, so it feels natural to put the config setup there.

kedro/framework/session/session.py
```diff
@@ -209,6 +210,7 @@ def _setup_logging(self) -> None:
         """Register logging specified in logging directory."""
         conf_logging = self._get_logging_config()
         logging.config.dictConfig(conf_logging)
```
idanov, line 212: This needs to move inside `configure_logging()`.

kedro/runner/parallel_runner.py
```diff
@@ -270,7 +270,6 @@ def _run(  # pylint: disable=too-many-locals,useless-suppression
```
idanov, line 270: Could we leave this out as a seprate PR addressing KED-2780? I know it will end up being two very small PRs, but this way our git history will tell a more comprehensive story.

pyproject.toml
```diff
@@ -72,8 +72,7 @@ layers = [
     "config"
```
idanov, line 72: Same comment as the Parallel Runner change. See above ☝️

kedro/framework/session/session.py
SajidAlamQB: Do you mean `logging.config.dictConfig(conf_logging)` has to be called in `configure_logging()`?

kedro/framework/session/session.py
lorenabalan: Ah fair enough thanks for the explanation!

kedro/framework/project/__init__.py
```diff
@@ -2,6 +2,8 @@
 configure a Kedro project and access its settings."""
 # pylint: disable=redefined-outer-name,unused-argument,global-statement
 import importlib
+import logging
+import logging.config
```
lorenabalan, line 6: We shouldn't need both imports, so we can remove one of them.

kedro/framework/project/__init__.py
```diff
@@ -195,6 +198,13 @@ def configure_project(package_name: str):
     PACKAGE_NAME = package_name
 
 
+def configure_logging(conf_logging):
+    """Configure config logging to make it available as a global variable."""
```
lorenabalan, line 202:
```suggestion
    """Configure logging to make it available as a global variable."""
```

kedro/framework/project/__init__.py
```diff
@@ -195,6 +198,13 @@ def configure_project(package_name: str):
     PACKAGE_NAME = package_name
 
 
+def configure_logging(conf_logging):
```
lorenabalan, line 201: Can we add some typehints here so we know what to expect?
```suggestion
def configure_logging(conf_logging: Dict[str, Any]) -> None:
```

not sure if the argument is Dict[str, Any] or Dict[str, str]

lorenabalan: Also, this is a minor thing, but can we think of a better name than `conf_logging`? 😅 config, configuration, logging_config etc. sound better too me but it completely optional.

lorenabalan (approved):

idanov (approved):
Awesome 🎉

PR #1105, opened 2021-12-08
`APIDataSet` ease of use changes to `auth` interactions

### LINKED ISSUE #1101: `APIDataSet` ease of use
## Description

Howdy team!

I was working with the `APIDataSet` recently and had two issues out of the box.

#### 1. Specifying the `auth` keyword argument in yaml

The `requests` library expects the `auth` parameter of a request to be either a `HTTPBasicAuth` or a `tuple` (lists are not allowed, see [here](https://github.com/ianwhale/kedro-kaggle-starter/blob/master/%7B%7B%20cookiecutter.repo_name%20%7D%7D/src/%7B%7B%20cookiecutter.python_package%20%7D%7D/api.py) in requests). At the moment, neither are possible to specify in my `catalog.yml`. 

From what I hear, you're already working on this (#1011). So maybe this point is moot.

#### 2. The `auth` keyword argument and `credentials.yml`

I would like to specify my `(username, password)` tuple inside `credentials.yml`. However, the `APIDataSet`'s `auth` keyword wouldn't get filled in by the config loader. 

To get this working, you'd have to extend `APIDataSet` to have a `credentials` keyword that is filled in for `auth` in an upcall.

It would be great to either have this by default, or even have the loader fill `auth` keywords in addition to `credentials`. Although that might have unintended consequences. 

## Context

Hopefully this would unify the experience a bit. Right now, the `credentials` keyword in a dataset and `credentials.yml` are the main points of access to secrets. Which is probably good.

## Possible Implementation

I whipped up [my own `APIDataSet`](https://github.com/ianwhale/kedro-kaggle-starter/blob/master/%7B%7B%20cookiecutter.repo_name%20%7D%7D/src/%7B%7B%20cookiecutter.python_package%20%7D%7D/api.py) to solve both the problems above.

## Possible Alternatives

To get this working with no changes to `APIDataSet`, we'd have to implement the changes in #1011 so we can specify tuples in `credentials.yml` and have the config loader fill in `auth` as well.

ianwhale:
## Description

Closes #1101.

## Development notes

Updated `APIDataSet` to:
- Accept `credentials` as a keyword argument that is filled in for `auth`. Now the `auth` tuple can be stored in `credentials.yml`.
- Casts whatever iterable that is passed for `auth` to a `tuple` as expected by `requests`.

## Checklist

- [x] Added a description of this change in the [`RELEASE.md`](https://github.com/quantumblacklabs/kedro/blob/main/RELEASE.md) file


AntonyMilneQB (approved):
Beautiful PR, thank you for raising! 

I just have one small hesitation: `credentials` is usually a kwargs dictionary rather than a list. I think this is fine and what you've done makes sense, but since it's a bit unusual would it be possible to modify [Example 13](https://kedro.readthedocs.io/en/stable/05_data/01_data_catalog.html#using-the-data-catalog-with-the-yaml-api) to show how this dataset works with a credentials entry?

ianwhale:
> Beautiful PR, thank you for raising!
> 
> I just have one small hesitation: `credentials` is usually a kwargs dictionary rather than a list. I think this is fine and what you've done makes sense, but since it's a bit unusual would it be possible to modify [Example 13](https://kedro.readthedocs.io/en/stable/05_data/01_data_catalog.html#using-the-data-catalog-with-the-yaml-api) to show how this dataset works with a credentials entry?

Thanks @AntonyMilneQB! Not sure if anyone has called a PR of mine :sparkles:beautiful:sparkles:. 

I updated the USDA corn example you had, though I don't believe that API requires credentials. If we're fine with that, we can call it done.

Otherwise, I can add a second example of using `APIDataSet` with Kaggle, which was my original use case and how I found these issues. This would be more realistic, but their API doesn't take any query params like the USDA one does. So a second example might make more sense.

docs/source/05_data/01_data_catalog.md
```diff
@@ -262,7 +265,13 @@ us_corn_yield_data:
     year: 2000
 ```
 
-When using [`pandas.SQLTableDataSet`](/kedro.extras.datasets.pandas.SQLTableDataSet) or [`pandas.SQLQueryDataSet`](/kedro.extras.datasets.pandas.SQLQueryDataSet) you must provide a database connection string. In the example above we pass it using `scooters_credentials` key from the credentials (see the details in [Feeding in credentials](#feeding-in-credentials) section below). `scooters_credentials` must have a top-level key `con` containing [SQLAlchemy compatible](https://docs.sqlalchemy.org/en/13/core/engines.html#database-urls) connection string. As an alternative to credentials, you could explicitly put `con` into `load_args` and `save_args` (`pandas.SQLTableDataSet` only).
+Note that `usda_credientials` will be passed as the `auth` argument in the `requests` library. Specify the username and password as a list in your `credentials.yml` like the following:
```
AntonyMilneQB, line 268:
```suggestion
Note that `usda_credientials` will be passed as the `auth` argument in the `requests` library. Specify the username and password as a list in your `credentials.yml` as follows:
```

AntonyMilneQB:
Sorry for the slow response @ianwhale - have been away for a few days.

I think the updated USDA corn example is fine, even if it doesn't actually require credentials. Thanks for updating the json schema too - well remembered!

Let's try and get a second review of this so we can hopefully get it merged 🙂

SajidAlamQB (approved):
LGTM! 👍  Thank you for the contribution!

PR #1130, opened 2021-12-19
Improve the readability of `runner` code :pencil2:

deepyaman:
## Checklist

### DIFF
--- a/kedro/runner/parallel_runner.py
+++ b/kedro/runner/parallel_runner.py
@@ -24,7 +24,7 @@ _MAX_WINDOWS_WORKERS = 61
 
 
 class _SharedMemoryDataSet:
-    """``_SharedMemoryDataSet`` a wrapper class for a shared MemoryDataSet in SyncManager.
+    """``_SharedMemoryDataSet`` is a wrapper class for a shared MemoryDataSet in SyncManager.
     It is not inherited from AbstractDataSet class.
     """
 
@@ -162,10 +162,10 @@ class ParallelRunner(AbstractRunner):
         """Factory method for creating the default data set for the runner.
 
         Args:
-            ds_name: Name of the missing data set
+            ds_name: Name of the missing data set.
 
         Returns:
-            An instance of an implementation of _SharedMemoryDataSet to be used
+            An instance of an implementation of ``AbstractDataSet`` to be used
             for all unregistered data sets.
 
         """


kedro/runner/thread_runner.py
```diff
@@ -52,10 +52,10 @@ def create_default_data_set(self, ds_name: str) -> AbstractDataSet:
         """Factory method for creating the default data set for the runner.
 
         Args:
-            ds_name: Name of the missing data set
+            ds_name: Name of the missing data set.
 
         Returns:
-            An instance of an implementation of AbstractDataSet to be used
+            An instance of an implementation of ``AbstractDataSet`` to be used
```
deepyaman, line 58: How do you decide whether to use single vs double vs. no backticks? Line 31 has 1, and line 27 of parallel_runner.py has a mix of 0 and 2.

kedro/runner/thread_runner.py
```diff
@@ -52,10 +52,10 @@ def create_default_data_set(self, ds_name: str) -> AbstractDataSet:
...
             for all unregistered data sets.
```
deepyaman, line 59:
```suggestion
            An instance of ``MemoryDataSet`` to be used for all
            unregistered data sets.
```

kedro/runner/parallel_runner.py
```diff
@@ -162,10 +162,10 @@ def create_default_data_set(  # type: ignore
         """Factory method for creating the default data set for the runner.
 
         Args:
-            ds_name: Name of the missing data set
+            ds_name: Name of the missing data set.
 
         Returns:
-            An instance of an implementation of _SharedMemoryDataSet to be used
+            An instance of an implementation of ``AbstractDataSet`` to be used
             for all unregistered data sets.
```
deepyaman, line 169:
```suggestion
            An instance of ``_SharedMemoryDataSet`` to be used for all
            unregistered data sets.
```

kedro/runner/thread_runner.py
AntonyMilneQB: There's no good reason for this inconsistency. Double backticks are always best because
* if the class/function/whatever being enclosed is exposed in `__init__.py` files  then it gets automatically  linked to the right page in the docs ([code that does this](https://github.com/quantumblacklabs/kedro/blob/main/docs/conf.py#L340), [click on `Pipeline` on this page for example](https://kedro.readthedocs.io/en/stable/kedro.runner.AbstractRunner.html#kedro.runner.AbstractRunner))
* if it's not a kedro object it will get formatted as code in the docs, whereas single backticks or no backticks will just do it as plain text

MerelTheisenQB (approved):
Thanks for these improvements @deepyaman ! 🎖️

lorenabalan (approved):
✨

kedro/runner/thread_runner.py
```diff
@@ -49,14 +49,14 @@ def __init__(self, max_workers: int = None, is_async: bool = False):
...
+        """Factory method for creating the default dataset for the runner.
 
         Args:
-            ds_name: Name of the missing data set
+            ds_name: Name of the missing dataset.
 
         Returns:
-            An instance of an implementation of AbstractDataSet to be used
-            for all unregistered data sets.
+            An instance of ``MemoryDataSet`` to be used for all
```
idanov, line 58: I don't think this is true, there's no requirement for it to be a `MemoryDataSet` and in fact for some runners it shouldn't be a `MemoryDataSet`.

deepyaman: This is specifically on the `ThreadRunner.create_default_data_set` docstring, similar to how `ParallelRunner.create_default_data_set` already had a custom docstring mentioning `_SharedMemoryDataSet`.

Actually, I think the one oversight (in keeping consistent with `ParallelRunner`) is that the return type here should be `-> MemoryDataSet`. WDYT?

lorenabalan: Yeah I agree with the docstrings. I almost flagged the same thing until I saw that in the abstract class we keep it generic, and in the implementations it's concrete. I agree we should make the typehint here be `MemoryDataSet`.

PR #1131, opened 2021-12-27
Document distribution of Kedro pipelines with Dask

deepyaman:
## Checklist


docs/source/10_deployment/01_deployment_guide.md
```diff
@@ -15,9 +15,10 @@ We also provide information to help you deploy to the following:
...
 * to [AWS Batch](07_aws_batch.md)
 * to [Databricks](08_databricks.md)
+* to [Dask](12_dask.md)
 
 <!--- There has to be some non-link text in the bullets above, if it's just links, there's a Sphinx bug that fails the build process-->
 
 In addition, we also provide instructions on [how to integrate a Kedro project with Amazon SageMaker](09_aws_sagemaker.md).
 
-![](../meta/images/deployments.png)
+![](../meta/images/deployments.png)  <!-- TODO(deepyaman): Add Dask to deployment flowchart. -->
```
deepyaman, line 24: Not sure how to recreate the flowchart; any help would be much appreciated!

deepyaman: On an unrelated note, why do all these images not have alt text? I'm not expert on this, but I feel like that's bad for accessibility.

docs/source/10_deployment/04_argo.md
```diff
@@ -1,6 +1,6 @@
 # Deployment with Argo Workflows
 
-This page explains how to convert your Kedro pipeline to use [Argo Workflows](https://github.com/argoproj/argo-workflows), an open source container-native workflow engine for orchestrating parallel jobs on [Kubernetes](https://kubernetes.io/).
+This page explains how to convert your Kedro pipeline to use [Argo Workflows](https://github.com/argoproj/argo-workflows), an open-source container-native workflow engine for orchestrating parallel jobs on [Kubernetes](https://kubernetes.io/).
```
deepyaman, line 3: Following English grammar rules; see also https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/o/open-source#:~:text=Hyphenate%20open%2Dsource%20as%20an,abbreviation%20for%20open%2Dsource%20software..

docs/source/10_deployment/07_aws_batch.md
```diff
@@ -286,13 +288,13 @@ def _track_batch_job(job_id: str, client: Any) -> None:
...
 
-You'll need to set the Batch-related configuration that the runner will use. Add a `parameters.yml` file inside the `conf/aws_batch/` directory created as part of the prerequistes steps, which will include the following keys:
+You'll need to set the Batch-related configuration that the runner will use. Add a `parameters.yml` file inside the `conf/aws_batch/` directory created as part of the prerequistes with the following keys:
 
 ```yaml
 aws_batch:
-    job_queue: "spaceflights_queue"
-    job_definition: "kedro_run"
-    max_workers: 2
+  job_queue: "spaceflights_queue"
```
deepyaman, line 295: To be consistent with Prettier + all the other YAML config (including in this file).

docs/source/10_deployment/12_dask.md
```diff
@@ -0,0 +1,311 @@
...
+```
+
+where the helper function `_instantiate_runner()` looks like this:
+
+```python
+def _instantiate_runner(runner, is_async, project_context):
+    runner_class = load_obj(runner, "kedro.runner")
+    runner_kwargs = dict(is_async=is_async)
+
+    if runner.endswith("DaskRunner"):
```
deepyaman, line 264: Did this in line with how it's done for AWS Batch, but why isn't it an equality condition? Do we expect people to extend it in a very particular way... but relying on naming convention sounds like it would be hacky AF.
```suggestion
    if runner == "DaskRunner":
```

docs/source/10_deployment/12_dask.md
```diff
@@ -0,0 +1,311 @@
...
+
+You're now ready to trigger the run. Without any further configuration, the underlying Dask [`Client`](http://distributed.dask.org/en/stable/api.html#distributed.Client) creates a [`LocalCluster`](http://distributed.dask.org/en/stable/api.html#distributed.LocalCluster) in the background and connects to that:
+
+```bash
+kedro run --runner kedro_tutorial.runner.DaskRunner
+```
+
+#### Set up Dask and related configuration
+
+To connect to an existing Dask cluster, you'll need to set the Dask-related configuration that the runner will use. Create the `conf/dask/` directory and add a `parameters.yml` file inside of it with the following keys:
```
deepyaman, line 281: Following this env-based approach to be consistent with AWS Batch, but honestly don't like it. Feel like you'd want to use env for dev/staging/prod/etc., and deploy to Dask in all of these situations, not just deploy to Dask as a single environment. Maybe it still makes sense for this "guide" purpose, though?

AWS Batch is probably a bit different here, since it is more a prod deployment thing.

Open to thoughts.

docs/source/10_deployment/12_dask.md
```diff
@@ -0,0 +1,311 @@
...
+  address: 127.0.0.1:8786
+```
+
+Next, [set up scheduler and worker processes on your local computer](http://distributed.dask.org/en/stable/quickstart.html#setup-dask-distributed-the-hard-way):
+
+```bash
+$ dask-scheduler
+Scheduler started at 127.0.0.1:8786
+
+$ PYTHONPATH=$PWD/src dask-worker 127.0.0.1:8786
```
deepyaman, line 294: A bit hacky, but maybe right for demo purposes? Or should this use `client.upload_file` somehow?

docs/source/10_deployment/12_dask.md
```diff
@@ -0,0 +1,311 @@
...
+.. note::  The above code snippet assumes each worker is started from the root directory of the Kedro project in a Python environment where all required dependencies are installed.
+```
+
+You're once again ready to trigger the run. Execute the following command:
+
+```bash
+kedro run --env dask --runner kedro_tutorial.runner.DaskRunner
+```
+
+You should start seeing tasks appearing on [Dask's diagnostics dashboard](http://127.0.0.1:8787/status):
```
deepyaman, line 309: I'm guessing this needs to be added to `linkcheck_ignore`, maybe once the build runs lol.

docs/source/10_deployment/12_dask.md
```diff
@@ -0,0 +1,311 @@
...
+
+    with KedroSession.create(env=env, extra_params=params) as session:
+        runner_instance = _instantiate_runner(runner, is_async, context)
+        session.run(
+            tags=tag,
+            runner=runner_instance,
+            node_names=node_names,
+            from_nodes=from_nodes,
+            to_nodes=to_nodes,
+            from_inputs=from_inputs,
```
deepyaman, line 251:
```suggestion
            from_inputs=from_inputs,
            to_outputs=to_outputs,
```
This seems to be missing when compared to latest https://github.com/quantumblacklabs/kedro/blob/master/kedro/framework/cli/project.py.

docs/source/10_deployment/12_dask.md
```diff
@@ -0,0 +1,311 @@
...
+
+        return self.run(to_rerun, catalog)
+```
+
+### Update CLI implementation
+
+You're nearly there! Before being able to use the new runner, update the `run()` function in your `cli.py` file to make sure the runner class is instantiated correctly:
+
+```python
+def run(tag, env, parallel, ...):
```
deepyaman, line 229: I understand that this saves space, but I feel like it would be very confusing to somebody less familiar with Kedro to figure out what to do here. This is especially true given `cli.py` has not been included as part of the project template since 0.17.5, and I don't see step-by-step documentation on how to set something like this up (e.g. I know I can go back to 0.17.4 template to find something like what I need).

docs/source/10_deployment/12_dask.md
```diff
@@ -0,0 +1,311 @@
...
+        )
+    runner = runner or "SequentialRunner"
+    if parallel:
+        runner = "ParallelRunner"
+
+    tag = _get_values_as_tuple(tag) if tag else tag
+    node_names = _get_values_as_tuple(node_names) if node_names else node_names
+
+    with KedroSession.create(env=env, extra_params=params) as session:
+        runner_instance = _instantiate_runner(runner, is_async, context)
```
deepyaman, line 244: `context` doesn't exist; something seems to have been lost in the refactoring.

deepyaman:
```suggestion
        context = session.load_context()
        runner_instance = _instantiate_runner(runner, is_async, context)
```

docs/source/10_deployment/12_dask.md
```diff
@@ -0,0 +1,313 @@
...
+def run(tag, env, parallel, ...):
+    """Run the pipeline."""
+    if parallel and runner:
+        raise KedroCliError(
+            "Both --parallel and --runner options cannot be used together. "
+            "Please use either --parallel or --runner."
+        )
+    runner = runner or "SequentialRunner"
+    if parallel:
+        runner = "ParallelRunner"
```
deepyaman, line 238: Should we include [this deprecation message](https://github.com/quantumblacklabs/kedro/blob/0.17.6/kedro/framework/cli/project.py#L377-L384) in there? I understand if it's overkill for this guide.

docs/source/10_deployment/12_dask.md
```diff
@@ -0,0 +1,313 @@
...
+        """Instantiates the runner by creating a ``distributed.Client``.
+
+        Args:
+            client_args: Arguments to pass to the ``distributed.Client``
+                constructor.
+            is_async: If True, the node inputs and outputs are loaded and saved
+                asynchronously with threads. Defaults to False.
+        """
+        super().__init__(is_async=is_async)
+        Client(**client_args or dict(address="127.0.0.1:8786"))
```
deepyaman, line 93:
```suggestion
        Client(**client_args)
```

docs/source/10_deployment/12_dask.md
```diff
@@ -0,0 +1,313 @@
...
+
+    def _run(
+        self, pipeline: Pipeline, catalog: DataCatalog, run_id: str = None
+    ) -> None:
+        nodes = pipeline.nodes
+        load_counts = Counter(chain.from_iterable(n.inputs for n in nodes))
+        node_dependencies = pipeline.node_dependencies
+        node_futures = {}
+
+        client = Client.current()
```
deepyaman, line 146: [DASK EXPERTISE REQUESTED] If somebody has familiarity with Dask, a review of how I get the `Client` would be very helpful. I think `worker_client` in `_DaskDataSet` is correct, but not sure if I should be using `Client.current()` the way I am in `DaskRunner`. I think `worker_client` is unnecessary here, since it all runs on the scheduler, and `Client.as_current` seems to be for a use case where you have a client object already and want to use it, but I don't find much documentation around this and most of my understanding is from reading the `dask.distributed` source.

docs/source/10_deployment/12_dask.md
```diff
@@ -0,0 +1,313 @@
...
+            # Decrement load counts, and release any datasets we
+            # have finished with. This is particularly important
+            # for the shared, default datasets we created above.
+            for data_set in node.inputs:
+                load_counts[data_set] -= 1
+                if load_counts[data_set] < 1 and data_set not in pipeline.inputs():
+                    catalog.release(data_set)
+            for data_set in node.outputs:
+                if load_counts[data_set] < 1 and data_set not in pipeline.outputs():
+                    catalog.release(data_set)
```
deepyaman, line 170: [DASK EXPERTISE REQUESTED] I've done this in the simplest way possible, in the `as_completed` loop. However, this feels a bit inefficient, as it really could've been released on the final load (rather than waiting for the node to finish running). I think this would require a distributed counter that `_DaskDataSet` instances could modify.. is this even smart?

docs/source/10_deployment/01_deployment_guide.md
datajoely: @limdauto has the draw.io diagram been committed to the repo?

docs/source/10_deployment/12_dask.md
datajoely: @deepyaman I would ask for a review on one of these channels
https://docs.dask.org/en/stable/support.html

docs/source/10_deployment/12_dask.md
datajoely: Overkill, maybe put an inline comment if you really care

datajoely:
This is excellent excellent work - I think once we've got clarification on the Dask specifics we should also make a Dask starter like we have for pyspark

docs/source/10_deployment/01_deployment_guide.md
MerelTheisenQB: I think for the alt text, we just haven't been consistent/forgot to add it.

MerelTheisenQB (commented):
Thanks for writing up these docs @deepyaman ⭐ ⭐

docs/source/10_deployment/12_dask.md
deepyaman: @datajoely I created a thread for the below question; I feel more confident about this one, more I think about it.

docs/source/10_deployment/01_deployment_guide.md
limdauto: @datajoely sorry I missed this ping. This flowchart was made by @stichbury iirc, not me :(

docs/source/10_deployment/12_dask.md
deepyaman: Upon further consideration/reading into it last night, I've decided to leave things as is, at least for now/until receiving further feedback.

1. With regards to releasing on final load, I can't see any straightforward implementation. Any logic to further segregate I/O and node execution in this runner gets more low level, which is not ideal from a tutorial perspective. Also want to avoid the case where submitting separate jobs for I/O (rather than the overarching `_run_node`, because that could mean more communication/large data transfer between workers). Overall, lot of complexity for limited-to-no benefit, IMO.
2. The suggestion to use `delayed` instead of `Future`s from the above Dask Discourse thread doesn't seem best to me, upon further consideration. That would delegate the whole DAG to Dask, and make it harder for Kedro to control anything in between. http://matthewrocklin.com/blog/work/2017/04/19/dask-glm-2 also talks a bit about the tradeoff between simplicity and flexibility between `delayed` and `Future`s, respectively.

lorenabalan:
Closing this as resolved in #1248

PR #1139, opened 2022-01-06
Pin pytables to match version installed through conda

MerelTheisenQB:
## Description
The `win_unit_tests_3.8` are failing on several PRs with a `Window stackoverflow` error which seems to be related to pytables. Pytables have released a new version 9 days ago, but it’s not on conda-forge, so in our windows CI we install pytables `3.6.1` through `conda install -c conda-forge pytables -y` and then in a later step when all requirements are installed pip installs the latest pytables version, which is now `3.7.0`. 

## Development notes
I’m not entirely sure why the errors are only happening on 3.8 (this might be related: https://github.com/PyTables/PyTables/issues/933), but pinning `tables` in `test_requirements.txt` seems to fix it for now. 

## Checklist


lorenabalan (approved):
We can pin it just for windows but it's only a test requirement so I'm not that bothered.

AntonyMilneQB:
Thanks for doing this! Looks like a good fix for the windows tests, but should we be changing this requirement too (just for windows like @lorenabalan suggested)? https://github.com/quantumblacklabs/kedro/blob/main/setup.py#L80

I'm not sure whether we should or not. On the one hand, would be nice to keep things consistent, but on the other since this wouldn't be released for a while maybe we should just wait until pytables have come out with a fix than then change to that new version. I don't know.

MerelTheisenQB:
Yep good points, I'll specify the operating system on the requirements and update `setup.py` as well.

rashidakanchwala (approved):

idanov (approved):

PR #1140, opened 2022-01-07
[KED-2419] Make pipeline and Pipeline consistent

AntonyMilneQB:
## Description
Currently we have `node` and `Node`, which are exactly equivalent. We encourage users to do `node`, which just instantiates `Node`.

We also have `pipeline` and `Pipeline`, which are quite different. `Pipeline` is the underlying class used to create a pipeline; then `pipeline` is then used to _transform_ that according to a namespace and dataset input/output mapping. The current use would be:
```
initial_pipeline = Pipeline([node1, node2, node3], tags=...)
transformed_pipeline = pipeline(initial_pipeline, inputs=..., outputs=..., namespace=...)
```
(Of course you don't have to transform the pipeline if the `initial_pipeline` is all you need.)

To be consistent with `node/Node` and reduce confusion we want to make `pipeline` and `Pipeline` equivalent. We would then encourage users to just use `pipeline`. This would mean no need for the `pipeline(Pipeline())` construction as users would instead just do `pipeline([node1, node2, node3], tag=..., input=..., outputs=..., namespace=...)`.

## Implementation
This is a bit tricky because:
* Both `pipeline` and `Pipeline` do various bits of validation, which I want to keep working exactly the same way for now so that we're still flagging the same errors as before
* the pipeline transformation itself depends on `Pipeline`. So we can't just immediately make `pipeline` a wrapper for `Pipeline` like we do for `node/Node`, because that would be circular

Let me try to explain what I've done using diagrams...

### Before
image

### After
image

Crucially the `Pipeline()` flow is the same as before, and the `pipeline(Pipeline())` flow is equivalent to before - it just has an extra "Validation A, set instance attributes" at the beginning which is unnecessary.
| Flow                                                    | Before                                                                                                | After                                                                                                                                        |
|---------------------------------------------------------|-------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------|
| `Pipeline()` without its new arguments (`inputs` etc.)  | Validation A, set instance attributes                                                                 | Validation A, set instance attributes                                                                                                        |
| `Pipeline()` with new arguments                         | Not possible                                                                                          | Validation A, set instance attributes, Validation B, transform, Validation A, set instance attributes                                        |
| `pipeline(Pipeline())` without `Pipeline` new arguments | Validation A, set instance attributes, Validation B, transform, Validation A, set instance attributes | Validation A, set instance attributes, Validation A, set instance attributes, Validation B, transform, Validation A, set instance attributes |

**I know this is confusing and I'm not totally happy with it! Please do say if there's an easier way of doing this or if nothing I wrote here makes sense. However, note that the last row of that table will no longer be encouraged (instead you'd just do `pipeline()`), and that this is meant to be a temporary state of affairs before future tickets can hopefully simplify the validation.**

Note that everything here is actually a non-breaking change, since the only arguments added to `pipeline/Pipeline` are keyword only and optional.

### This PR
- [ ] Update docstrings, tests

### This ticket, new PR(s)
- [ ] Update our docs to do `pipeline` rather than `Pipeline`
- [ ] Update our project template and starters similarly

### Separate tickets

- [ ] Simplify the validation - do we actually need to repeat Validation A after Validation B like we currently do?
- [ ] Refactor pipeline.py since it's huge
- [ ] Move `pipeline` from `modular_pipeline.py` to `pipeline.py` and delete `modular_pipeline.py`. This would break any imports `from kedro.pipeline.modular_pipeline import pipeline` but not `from kedro.pipeline import pipeline`

## Checklist

### DIFF
--- a/kedro/pipeline/node.py
+++ b/kedro/pipeline/node.py
@@ -599,7 +599,7 @@ def node(
     outputs: Union[None, str, List[str], Dict[str, str]],
     *,
     name: str = None,
-    tags: Iterable[str] = None,
+    tags: Union[str, Iterable[str]] = None,
     confirms: Union[str, List[str]] = None,
     namespace: str = None,
 ) -> Node:
--- a/kedro/pipeline/pipeline.py
+++ b/kedro/pipeline/pipeline.py
@@ -70,6 +70,11 @@ class ConfirmNotUniqueError(Exception):
     pass
 
 
+# factory function.
+# user should just do pipeline, not Pipeline
+# Make Pipeline and pipeline directly interchangeable. If make them similar then no benefit in not going all the way.
+
+
 class Pipeline:  # pylint: disable=too-many-public-methods
     """A ``Pipeline`` defined as a collection of ``Node`` objects. This class
     treats nodes as part of a graph representation and provides inputs,
@@ -81,6 +86,8 @@ class Pipeline:  # pylint: disable=too-many-public-methods
         nodes: Iterable[Union[Node, "Pipeline"]],
         *,
         tags: Union[str, Iterable[str]] = None,
+        # TODO? could also put inputs/outputs/parameters in here
+        # TODO: namespace should be here
     ):
         """Initialise ``Pipeline`` with a list of ``Node`` instances.
 
@@ -157,7 +164,7 @@ class Pipeline:  # pylint: disable=too-many-public-methods
                 self._nodes_by_input[_strip_transcoding(input_)].add(node)
 
         # output -> node with output
-        self._nodes_by_output = {}  # type: Dict[str, Node]
+        self._nodes_by_output = {}  # type: Dict[str, # Node]
         for node in nodes:
             for output in node.outputs:
                 self._nodes_by_output[_strip_transcoding(output)] = node
--- a/kedro/pipeline/pipeline.py
+++ b/kedro/pipeline/pipeline.py
@@ -135,8 +135,10 @@ class Pipeline:  # pylint: disable=too-many-public-methods
         self._set_pipeline(nodes, tags)
         if inputs or outputs or parameters or namespace:
             new_nodes = self._transform_pipeline(inputs, outputs, parameters, namespace)
-            self._set_pipeline(new_nodes, tags)
-            # TODO: don't really need to carry around tags here?
+            self._set_pipeline(
+                new_nodes, tags
+            )  # TODO: don't really need to carry around tags here?
+            # Or neater to do self.__init__(nodes, tags)?
 
     def _set_pipeline(self, nodes, tags):
         if nodes is None:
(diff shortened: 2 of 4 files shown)


kedro/pipeline/node.py
```diff
@@ -599,7 +599,7 @@ def node(
     outputs: Union[None, str, List[str], Dict[str, str]],
     *,
     name: str = None,
-    tags: Iterable[str] = None,
+    tags: Union[str, Iterable[str]] = None,
```
AntonyMilneQB, line 602: This should always have been this way so that `node` and `Node` match identically.

kedro/pipeline/pipeline.py
```diff
@@ -165,6 +178,119 @@ def __init__(
         self._nodes = nodes
         self._topo_sorted_nodes = _topologically_sorted(self.node_dependencies)
 
+    def _transform_pipeline(
```
AntonyMilneQB, line 181: This is exactly the same as it was in modular_pipeline.py, but with (a) the signature changed from
```
def pipeline(
    pipe: Pipeline,
    *,
    inputs: Union[str, Set[str], Dict[str, str]] = None,
    outputs: Union[str, Set[str], Dict[str, str]] = None,
    parameters: Dict[str, str] = None,
    namespace: str = None,
) -> Pipeline:
```
and (b) instances of `pipe` changed to `self` and (c) returns `new_nodes` rather than `Pipeline(new_nodes)`.

kedro/pipeline/pipeline.py
```diff
@@ -839,3 +965,71 @@ class CircularDependencyError(Exception):
     """
 
     pass
+
```
AntonyMilneQB, line 968: All this is the same as in modular_pipeline.py

kedro/pipeline/pipeline.py
datajoely: If you make this public would that be nicer for users?

AntonyMilneQB: Hmm, I don't think so. Any user who wants to do `pipe2 = pipe1._transform_pipeline(args)` should instead just do `pipe2 = pipeline(pipe1, args)`.

MerelTheisenQB:
Thanks for explaining this issue so thoroughly @AntonyMilneQB ! The diagram definitely helps. But I'm still a bit confused about what's going on 😅 

In the last row in the table it says for `pipeline(Pipeline())` without Pipeline new arguments 
the new flow would end up being ```Validation A, set instance attributes, Validation A, set instance attributes, Validation B, transform, Validation A, set instance attributes```

I'm not following why `Validation A, set instance attributes` will happen twice in a row in this case?  


I think I would prefer having a simpler flow over a complete match between `Pipeline` and `pipeline`.

AntonyMilneQB:
Thank you for looking through it and the feedback Merel! Following comments from Ivan also, I've reworked this into something much simpler in #1147.

PR #1147, opened 2022-01-10
[KED-2419] Make pipeline and Pipeline consistent, take 2

AntonyMilneQB:
## Description
Currently we have `node` and `Node`, which are exactly equivalent. We encourage users to do `node`, which just instantiates `Node`.

We also have `pipeline` and `Pipeline`, which are quite different. `Pipeline` is the underlying class used to create a pipeline; then `pipeline` is then used to _transform_ that according to a namespace and dataset input/output mapping. The current use would be:
```
initial_pipeline = Pipeline([node1, node2, node3], tags=...)
transformed_pipeline = pipeline(initial_pipeline, inputs=..., outputs=..., namespace=...)
```
(Of course you don't have to transform the pipeline if the `initial_pipeline` is all you need.)

To be more consistent with `node/Node` and reduce confusion we want to make all the options available in `Pipeline` also available to `pipeline`. We would then encourage users to just use `pipeline`. This would mean no need for the `pipeline(Pipeline())` construction as users would instead just do `pipeline([node1, node2, node3], tag=..., input=..., outputs=..., namespace=...)`.

Note that everything here is actually a non-breaking change, since the only arguments added to `pipeline/Pipeline` are keyword only and optional.

## Question
Do you like this change? @lorenabalan was concerned that it conflates the notion of pipeline and modular pipeline, and the result will be more rather than less confusing to users.

Looking at the arguments that `pipeline` now takes, it does seem a bit confusing to me also. As a user, when I create my first pipeline it makes sense to me that I can specify the `namespace` and `tags.` But why do I have the option of specifying the `inputs` and `outputs`? Shouldn't those be taken from the structure of DAG built up from the nodes provided? (Yes they are - `inputs`/`outputs`/`parameters` are only needed if you want to transform those to something else... but somehow these being arguments to the same function that takes `tags` as an argument doesn't feel quite right to me, like we're mixing two things up).

But also I do think it's nicer to be able to do `pipeline()` rather than `pipeline(Pipeline())`. Overall I think this is an improvement but I'm not totally convinced yet. 

### This PR
- [x] Update tests
- [x] Update release notes

### This ticket, new PR(s)
- [x] Update starters

### Separate tickets

- [ ] Update our docs to do `pipeline` rather than `Pipeline`
- [ ] **Maybe** ❓ Move `pipeline` from `modular_pipeline.py` to `pipeline.py` and delete `modular_pipeline.py`. This would break any imports `from kedro.pipeline.modular_pipeline import pipeline` but not `from kedro.pipeline import pipeline`

## Checklist


kedro/pipeline/node.py
```diff
@@ -599,7 +599,7 @@ def node(
     outputs: Union[None, str, List[str], Dict[str, str]],
     *,
     name: str = None,
-    tags: Iterable[str] = None,
+    tags: Union[str, Iterable[str]] = None,
```
AntonyMilneQB, line 602: This should always have been this way so that `node` and `Node` match identically.

kedro/pipeline/pipeline.py
```diff
@@ -692,10 +692,13 @@ def decorate(self, *decorators: Callable) -> "Pipeline":
...
-        """
-        Return a copy of the pipeline, with each node tagged accordingly.
-        :param tags: The tags to be added to the nodes.
-        :return: New `Pipeline` object.
+        """Tags all the nodes in the pipeline.
+
+        Args:
+            tags: The tags to be added to the nodes.
+
+        returns:
```
AntonyMilneQB, line 700:
```suggestion
        Returns:
```

MerelTheisenQB:
I think this change makes sense, and after our chat I can say that I also like this change 🙂  

I don't necessarily think that this change makes the notion of `pipeline` and `modular pipeline` and how to create them more confusing than the current state. It's not immediately clear from `Pipeline` and `pipeline` which one is used to create a modular pipeline. However, we should help users understand how to use `pipeline` to create a "regular" or "modular" pipeline, probably through doc string and extra documentation. And on top of that explain when/why they need to provide the other arguments.

kedro/pipeline/modular_pipeline.py
```diff
@@ -110,6 +115,12 @@ def pipeline(
     Returns:
         A new ``Pipeline`` object with the new nodes, modified as requested.
     """
+    if not isinstance(pipe, Pipeline):
+        pipe = Pipeline(pipe, tags=tags)
+
+    if inputs is None and outputs is None and parameters is None and namespace is None:
```
lorenabalan, line 121: Optional: We can maybe simplify this condition with an `all`, sth like 

```
all(x is None for x in (inputs, outputs...))
```

lorenabalan (approved):
LGTM,  think the more I look at it the more it grows on me. Would be good to add some tests.

RELEASE.md
```diff
@@ -1,3 +1,14 @@
+# Release 0.17.7
```
AntonyMilneQB, line 1: Will this exist? 🤔

kedro/pipeline/modular_pipeline.py
```diff
@@ -108,8 +114,17 @@ def pipeline(
             any of the expected types (str, dict, list, or None).
 
     Returns:
-        A new ``Pipeline`` object with the new nodes, modified as requested.
+        A new ``Pipeline`` object.
     """
+    if isinstance(pipe, Pipeline):
+        # To ensure that we are always dealing with a *copy* of pipe.
+        pipe = Pipeline([pipe], tags=tags)
```
AntonyMilneQB, line 121: `Pipeline` only takes an iterable of pipelines/nodes, hence wrapping this in `[]`.

AntonyMilneQB (commented):
I've finished updating the docstrings, project template and added some tests, so please do re-review and let me know what you think. I'll update docs and starters in separate PRs.

I wasn't sure exactly how many tests to add for this or how many existing ones to change from `Pipeline` to `pipeline` so let me know if you think I should do some more there. I did manually do a find and replace for all instances of `Pipeline()` and replace them with `pipeline()` and all tests still passed, which is a good sign that people can now always use `pipeline`.

kedro/pipeline/modular_pipeline.py
```diff
@@ -69,34 +69,40 @@ def _validate_datasets_exist(
...
     *,
     inputs: Union[str, Set[str], Dict[str, str]] = None,
     outputs: Union[str, Set[str], Dict[str, str]] = None,
     parameters: Dict[str, str] = None,
+    tags: Union[str, Iterable[str]] = None,
     namespace: str = None,
 ) -> Pipeline:
-    """Create a copy of the pipeline and its nodes,
-    with some dataset names and node names modified.
+    """Create a ``Pipeline`` from a collection of nodes and/or ``Pipelines``s.
```
AntonyMilneQB, line 80:
```suggestion
    """Create a ``Pipeline`` from a collection of nodes and/or ``Pipeline``s.
```

RELEASE.md
```diff
@@ -1,3 +1,14 @@
...
+
+## Major features and improvements
+* `pipeline` now accepts `tags` and a collection of `Node`s and/or `Pipeline`s rather than just a single `Pipeline`. `pipeline` should be used in preference to `Pipeline`.
```
AntonyMilneQB, line 4:
```suggestion
* `pipeline` now accepts `tags` and a collection of `Node`s and/or `Pipeline`s rather than just a single `Pipeline` object. `pipeline` should be used in preference to `Pipeline` when creating a Kedro pipeline.
```

RELEASE.md
MerelTheisenQB: Yes I think it makes sense to make a small release before `0.18.0`

kedro/pipeline/modular_pipeline.py
```diff
@@ -69,34 +69,40 @@ def _validate_datasets_exist(
...
 
     Args:
-        pipe: Original modular pipeline to integrate
+        pipe: The nodes the ``Pipeline`` will be made of. If you
+            provide pipelines among the list of nodes, those pipelines will
+            be expanded and all their nodes will become part of this
+            new pipeline.
         inputs: A name or collection of input names to be exposed as connection points
-            to other pipelines upstream.
+            to other pipelines upstream. This is optional; if not provided, the
```
MerelTheisenQB, line 88: 👍

kedro/pipeline/modular_pipeline.py
```diff
@@ -69,34 +69,40 @@ def _validate_datasets_exist(
...
-            to other pipelines downstream.
+            to other pipelines downstream. This is optional; if not provided, the
+            pipeline inputs are automatically inferred from the pipeline structure.
             When str or Set[str] is provided, the listed output names will stay
             the same as they are named in the provided pipeline.
             When Dict[str, str] is provided, current output names will be
             mapped to new names.
             Can refer to both the pipeline's free outputs, as well as
             intermediate results that need to be exposed.
         parameters: A map of existing parameter to the new one.
```
MerelTheisenQB, line 104: Should we also explicitly say that `parameters` are optional, now it's mentioned for `inputs`, `outputs` and `tags`?

tests/pipeline/test_pipeline_helper.py
```diff
@@ -355,3 +355,16 @@ def test_bad_outputs_mapping(self):
         pattern = "Outputs can't contain free inputs to the pipeline"
         with pytest.raises(ModularPipelineError, match=pattern):
             pipeline(raw_pipeline, outputs={"A": "C"})
+
+    def test_pipeline_always_copies(self):
+        original_pipeline = pipeline([node(constant_output, None, "A")])
+        new_pipeline = pipeline(original_pipeline)
+        assert new_pipeline is not original_pipeline
```
MerelTheisenQB, line 362: To me it seems like the title of this test `test_pipeline_always_copies` doesn't fully match the `assert`. Perhaps add another `assert` that verifies that `new_pipeline` contains what we expect?

MerelTheisenQB (commented):
Nice work! 👍  Just some small comments from my side.

kedro/pipeline/modular_pipeline.py
AntonyMilneQB: I think this is ok without because, unlike the others, `parameters` is not a required argument to `node` so there's no real possible confusion here 🙂

tests/pipeline/test_pipeline_helper.py
AntonyMilneQB: Good point! Done in ea4e090.

MerelTheisenQB (approved):
🌟

idanov (approved):
Awesome 🎉 I love the additional changes which clean up some docs/typing debt ❤️

PR #1154, opened 2022-01-14
Update 01_logging.md

ggerogiokas:
Adding some documentation so one knows how to point to handler names defined within a logging.yml configuration when using the python logger within kedro.


docs/source/08_logging/01_logging.md
```diff
@@ -18,6 +18,8 @@ log.warning("Issue warning")
 log.info("Send information")
 ```
 
+Note: Log names like the example above, `__name__`, can be handler names defined within the `handlers` section keys.  The default handler names are `console, info_file_handler, error_file_handler` found within `conf/base/logging.yml`.
```
AntonyMilneQB, line 21:
````suggestion
```eval_rst
.. note:: The name of a logger corresponds to the keys of the ``loggers``  sections in ``logging.yml`` (e.g. ``kedro.io``). See `Python's logging documentation <https://docs.python.org/3/library/logging.html#logger-objects>`_ for more information.
```
````

AntonyMilneQB (commented):
Hello and thank you for your contribution @ggerogiokas! What you wrote isn't quite correct - the argument to the function should be the name of a _logger_ rather than a _handler_. This isn't obvious - I had to try this out to check it.

I've suggested a change to correct this and also format the note a bit differently so it appears like this in our docs:
image

ggerogiokas:
Looks great to me!  Yea, I was a bit confused with whether it was handler or logger name thanks for checking it out.  I guess all set to push after that.

AntonyMilneQB (approved):

datajoely:
Thanks for doing this @ggerogiokas !

MerelTheisenQB (approved):
LGTM! 👍

PR #1163, opened 2022-01-18
[KED-2865] Make sql datasets use a singleton pattern for connection

lorenabalan:

## Description

Part of the story about Making Remove Datasets Use a Singleton Pattern. Splitting it into 2 PRs, one for SQL datasets, one for Spark dataset(s).

## Checklist


kedro/extras/datasets/pandas/sql_dataset.py
```diff
@@ -223,26 +235,32 @@ def _describe(self) -> Dict[str, Any]:
...
         except NoSuchModuleError as exc:
             raise _get_sql_alchemy_missing_error() from exc
 
     def _exists(self) -> bool:
-        eng = create_engine(self._load_args["con"])
+        eng = self.engine  # type: ignore
         schema = self._load_args.get("schema", None)
         exists = self._load_args["table_name"] in eng.table_names(schema)
-        eng.dispose()
+        # eng.dispose()
```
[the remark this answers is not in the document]
lorenabalan: As per team discussion, leaving this out for now as it might be the case that the connections are closed when the python process is closed (therefore after `kedro run` is done, we are in a decent state), and the garbage collector cleans it up.

tests/extras/datasets/pandas/test_sql_dataset.py
```diff
@@ -77,61 +80,67 @@ def test_load_sql_params(self, mocker, table_data_set):
...
         then the error should contain the original error message"""
         mocker.patch(
-            "pandas.read_sql_table",
+            "kedro.extras.datasets.pandas.sql_dataset.create_engine",
             side_effect=ImportError("No module named 'unknown_module'"),
         )
         pattern = ERROR_PREFIX + r"No module named \'unknown\_module\'"
         with pytest.raises(DataSetError, match=pattern):
-            table_data_set.load()
+            SQLTableDataSet(table_name=TABLE_NAME, credentials=dict(con=CONNECTION))
```
lorenabalan, line 116: Some of these tests are now duplicated, because we're checking for errors at instantiation, rather than on `load` AND `save`. TODO: Deduplicate these and move under `TestSQLTableDataSet`.

kedro/extras/datasets/pandas/sql_dataset.py
```diff
@@ -374,32 +385,48 @@ def __init__(  # pylint: disable=too-many-arguments
...
+        self.create_connection(self._connection_str)
+
+    @classmethod
+    def create_connection(cls, connection_str: str) -> None:
+        """Given a connection string, create singleton connection
+        to be used across all instances of `SQLQueryDataSet` that
+        need to connect to the same source.
+        """
+        if connection_str in getattr(cls, "engines", {}):
+            return  # pragma: no cover
```
lorenabalan, line 398: I added `test_create_connection_only_once` unit test specifically to cover this but it still doesn't get captured in coverage, so added a `pragma: no cover` too.

datajoely:
Really nice work

kedro/extras/datasets/pandas/sql_dataset.py
MerelTheisenQB: Isn't that because the test uses the `SQLTableDataSet` and not the `SQLQueryDataSet` ?

lorenabalan: 🤦‍♀️  I didn't pay close attention, probably... I'll check, thanks!

kedro/extras/datasets/pandas/sql_dataset.py
```diff
@@ -207,42 +207,53 @@ def __init__(
...
+    @classmethod
+    def create_connection(cls, connection_str: str) -> None:
+        """Given a connection string, create singleton connection
+        to be used across all instances of `SQLTableDataSet` that
+        need to connect to the same source.
+        """
+        if connection_str in getattr(cls, "engines", {}):
+            return
+
+        engines = cls.engines if hasattr(cls, "engines") else {}  # type:ignore
```
AntonyMilneQB, line 222: This feels a bit weird to me 🤔  Probably I'm missing something here... but why not just set `engines = {}` with other class attributes above `def __init__`? It means setting a mutable default argument but I think that's fine here (and we already do it with the other class attributes there, although that is arguably a bit different).

Then we could just simplify like:
```
if connection_str in cls.engines:
    return
...
cls[engines] = engine # no need for local `engines` object at all
```

kedro/extras/datasets/pandas/sql_dataset.py
```diff
@@ -207,42 +207,53 @@ def __init__(
...
         )
 
     def _load(self) -> pd.DataFrame:
-        try:
-            return pd.read_sql_table(**self._load_args)
-        except ImportError as import_error:
-            raise _get_missing_module_error(import_error) from import_error
-        except NoSuchModuleError as exc:
-            raise _get_sql_alchemy_missing_error() from exc
+        engine = self.engines.get(self._connection_str)  # type:ignore
```
AntonyMilneQB, line 246: Why the need for `get` here? The entry should always be there in the dictionary I think?

```suggestion
        engine = self.engines[self._connection_str]  # type:ignore
```

tests/extras/datasets/pandas/test_sql_dataset.py
```diff
@@ -53,10 +60,6 @@ def query_file_data_set(request, sql_file):
 
 
 class TestSQLTableDataSetLoad:
-    @staticmethod
-    def _assert_pd_called_once():
-        pd.read_sql_table.assert_called_once_with(table_name=TABLE_NAME, con=CONNECTION)
-
```
AntonyMilneQB, line 59: I love that this is being put inline 👍

tests/extras/datasets/pandas/test_sql_dataset.py
```diff
@@ -224,6 +224,46 @@ def test_table_exists_mocked(self, mocker, table_data_set):
...
+        for d in datasets:
+            d.save(dummy_dataframe)
+
+        start = time.time()
+        for d in datasets:
+            d.load()
+        end = time.time()
+
+        delta = end - start
+        print(delta)
```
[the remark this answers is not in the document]
AntonyMilneQB: Do you intend to keep this as a unit test or is it just you trying stuff out now? I think it's a bit of a weird one to keep.

tests/extras/datasets/pandas/test_sql_dataset.py
```diff
@@ -224,12 +233,71 @@ def test_table_exists_mocked(self, mocker, table_data_set):
...
+        first = SQLTableDataSet(table_name=TABLE_NAME, credentials=dict(con=CONNECTION))
+        assert len(first.engines) == 1
+
+        second = SQLTableDataSet(
+            table_name="other_table", credentials=dict(con=CONNECTION)
+        )
+        assert len(second.engines) == 1
+        assert len(first.engines) == 1
+
+        mock_engine.assert_called_once_with(CONNECTION)
```
AntonyMilneQB, line 297: Do we need a test to check that two different connection strings create two different engines?

AntonyMilneQB (commented):
Just left a few comments for now - I'm going to read through that GitHub issue and ponder a bit more but basically I think looks great 💯 ⭐

kedro/extras/datasets/pandas/sql_dataset.py
lorenabalan: I don't follow your suggestion sorry. I've never seen syntax like `cls[engines]` before either, it doesn't look like idiomatic python to me.

tests/extras/datasets/pandas/test_sql_dataset.py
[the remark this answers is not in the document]
lorenabalan: No, I won't be keeping this, it was just for visibility and testing in the meantime.

kedro/extras/datasets/pandas/sql_dataset.py
AntonyMilneQB: Oh sorry, that was a typo. I mean `cls.engines`. What I meant was this:
```
class SQLTableDataSet(AbstractDataSet):
    engines = {} # like we do with DEFAULT_LOAD/SAVE_ARGS

    def __init__():
        # as you have it now

    @classmethod
    def create_connection(cls, connection_str: str) -> None:
        if connection_str in cls.engines: 
            return

        try:
            engine = create_engine(connection_str)
        except ImportError as import_error:
            raise _get_missing_module_error(import_error) from import_error
        except NoSuchModuleError as exc:
            raise _get_sql_alchemy_missing_error() from exc

        cls.engines[connection_str] = engine
```

So no need for the `hasattr`/`getattr` stuff.

kedro/extras/datasets/pandas/sql_dataset.py
```diff
@@ -374,32 +385,48 @@ def __init__(  # pylint: disable=too-many-arguments
...
-        desc["filepath"] = str(self._filepath)
-        del load_args["con"]
-        desc["load_args"] = str(load_args)
-
-        return desc
+        return dict(
+            sql=str(load_args.pop("sql", None)),
+            filepath=str(self._filepath),
+            load_args=str(load_args),
+        )
```
idanov, line 418: Pretty nice change, surprised we had the previous version for so long 😱

kedro/extras/datasets/pandas/sql_dataset.py
```diff
@@ -374,32 +385,48 @@ def __init__(  # pylint: disable=too-many-arguments
...
             with self._fs.open(load_path, mode="r") as fs_file:
                 load_args["sql"] = fs_file.read()
 
-        try:
-            return pd.read_sql_query(**load_args)
-        except ImportError as import_error:
-            raise _get_missing_module_error(import_error) from import_error
-        except NoSuchModuleError as exc:
-            raise _get_sql_alchemy_missing_error() from exc
+        return pd.read_sql_query(con=engine, **load_args)
```
idanov, line 429: So much simpler now!

tests/extras/datasets/pandas/test_sql_dataset.py
```diff
@@ -224,12 +233,71 @@ def test_table_exists_mocked(self, mocker, table_data_set):
...
+            d.load()
+        end = time.time()
+
+        delta = end - start
+        print(delta)
+        cons = datasets[0].engines
+        for con in cons.values():
+            con.dispose()
+
+    def test_single_connection(self, dummy_dataframe, mocker):
```
idanov, line 259: Maybe worth looking into testing the number of connections for the engine as well: https://stackoverflow.com/questions/34775501/how-could-i-check-the-number-of-active-sqlalchemy-connections-in-a-pool-at-any-g#comment114580726_34781703

tests/extras/datasets/pandas/test_sql_dataset.py
```diff
@@ -224,12 +233,71 @@ def test_table_exists_mocked(self, mocker, table_data_set):
...
+
+        expected_call = mocker.call(name=TABLE_NAME, con=unique_connection, index=False)
+        dummy_to_sql.assert_has_calls([expected_call] * 10)
+
+        for ds in datasets:
+            ds.load()
+            engine = ds.engines[CONNECTION]
+            assert engine is unique_connection
+
+    def test_create_connection_only_once(self, mocker):
```
idanov, line 281: Quite a nice test 👍

idanov (commented):
Looks quite nice, it needs some refactoring of the tests (as in your own comments suggest) and maybe look into create the class attribute similarly to the default arguments, but overall I think we'll be ready to merge this soon. Hopefully it makes it harder for people to run out of connections when having many SQL-based datasets in their pipelines.

kedro/extras/datasets/pandas/sql_dataset.py
lorenabalan: Fixed, was being overly cautious.

tests/extras/datasets/pandas/test_sql_dataset.py
lorenabalan: Hmm I looked into this but we seem to be getting a `NullPoll` in our tests, which doesn't have the nice `status()` return. 🤔

kedro/extras/datasets/pandas/sql_dataset.py
lorenabalan: Makes sense, I was probably just overthinking. Done!

MerelTheisenQB (approved):
I like this a lot! Very elegant solution 🌟

AntonyMilneQB (approved):
Really nice work, especially with the testing! ⭐

idanov (approved):
🎉

PR #1182, opened 2022-01-27
Pin upper pandas version requirement

### LINKED ISSUE #1179: ExcelDataSet not working after pandas 1.4.0 release (python 3.8)
## Description

On a fresh new project, after installing the extra dependency `kedro[pandas.ExcelDataSet]`, the user receives a runtime error when trying to load a xslx file dataset via the data catalog. The cause of the issue seems to be an incompatibility with the new pandas `v1.4.0` and `xlrd~=1.0`, the latter being enforced by kedro `17.6.0`.

## Context

The error occurred while I was following the exact steps of the spaceflights tutorial. In order to load the `shuttles` dataset (stored as a xslx file), I was instructed to install the extra dependency `kedro[pandas.ExcelDataSet]`. After installing the dependency and trying to load the dataset via the catalog, I received the following error: `Pandas requires version '2.0.1' or newer of 'xlrd' (version '1.2.0' currently installed).`

After some debugging, I discovered that pandas released version `1.4.0` two days ago, and it bumped the mininum version of optional dependency xlrd to `2.0.1` ([source](https://pandas.pydata.org/docs/whatsnew/v1.4.0.html#increased-minimum-versions-for-dependencies), [source2](https://github.com/pandas-dev/pandas/blob/v1.4.0/pandas/compat/_optional.py#L40)). As kedro enforces `xlrd~=1.0` via extra dependency `kedro[pandas.ExcelDataSet]` ([source](https://github.com/kedro-org/kedro/blob/0.17.6/setup.py#L75)), pandas checks that the currently installed version of xlrd is `1.2.0` and raises the error before importing the package ([source](https://github.com/pandas-dev/pandas/blob/v1.4.0/pandas/compat/_optional.py#L133-L152)).

I repeated the same steps using python 3.7, but because pandas 1.4.0 only supports python 3.8+ ([source](https://github.com/pandas-dev/pandas/blob/v1.4.0/setup.cfg#L40)), the installed version of pandas was 1.3.5. When I tried to load the dataset, I received no errors (although I got a future warning saying xlrd will not support xlsx files in version >= 2.0)

## Steps to Reproduce
1. Create a virtualenv with python 3.8 and activate it
2. Run `pip install kedro==0.17.6`
3. Run `kedro new`
4. `cd` to project dir
5. Run `kedro install`
6. Replace the line `kedro==0.17.6` with `kedro[pandas.ExcelDataSet]==0.17.6`  in src/requirements.in
7. Run `kedro build-reqs && kedro install`
8. Add file `shuttles.xlsx` (from spaceflights tutorial) to folder `data/01_raw/`
9. Add the dataset information to `catalog.yml`:
    ```yml
    shuttles:
      type: pandas.ExcelDataSet
    ```
10. Run `kedro ipython`
11. Run python code: `shuttles = catalog.load("shuttles")`

## Expected Result
The xlsx file should've been loaded into memory as a pandas Dataframe

## Actual Result
I received the following error:

```
DataSetError: Failed while loading data from data set ExcelDataSet(filepath=/.../data/01_raw/shuttles.xlsx, load_args={'engine': xlrd}, protocol=file, save_args={'index': False}, writer_args={'engine': xlsxwriter}).
Pandas requires version '2.0.1' or newer of 'xlrd' (version '1.2.0' currently installed).
```

## Your Environment

* Kedro version used (`pip show kedro` or `kedro -V`): 0.17.6
* Python version used (`python -V`): 3.8.12
* Operating system and version: WSL: Ubuntu 20.04
* Pandas version: 1.4.0
* xlrd version: 1.2.0

AntonyMilneQB:
## Description
Fixes #1179. pandas have released 1.4.0, which requires `xlrd>=2.0.1` (and is only compatible with Python >=3.8). This is incompatible with our `xlrd~=1.0`.

## Development notes
Note that pandas was unpinned in August 2020 in https://github.com/quantumblacklabs/private-kedro/pull/772. In hindsight we probably should have reinstated an upper bound some time after that. I've done so now to restrict to `<1.4`.

The alternative to this was to allow pandas 1.4 and relax our requirement for xlrd to include 2.0.1. I opted not to do this since xlrd 2.0 introduced [breaking changes](https://xlrd.readthedocs.io/en/latest/changes.html#id1).

## Questions
1. Does this seem like the right thing to do? It stops people from using `pandas>=1.4` which is maybe annoying for users who want to use any new features in those releases. We could do something to make the version of xlrd installed depend on the version of pandas so that not everyone gets `xlrd>=2.0`, but that seemed like overkill to me.
1. Looking at pandas [release notes](https://pandas.pydata.org/pandas-docs/stable/whatsnew/index.html#), every 1.x release (like 1.1, 1.2, 1.3, etc.) appears to break backwards compatibility. So I wonder whether we should change the specification in develop also (currently `pandas~=1.3`)?

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ parameters:
     default: false
 
 orbs:
-  win: circleci/windows@2.2.0
+  win: circleci/windows@2.4.1
 
 # No windows executor is listed here since windows builds use win/default and modify
 # the Python version through the conda environment.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -4,8 +4,9 @@
 * `pipeline` now accepts `tags` and a collection of `Node`s and/or `Pipeline`s rather than just a single `Pipeline` object. `pipeline` should be used in preference to `Pipeline` when creating a Kedro pipeline.
 
 ## Bug fixes and other changes
-* Added tutorial documentation for experiment tracking in Kedro docs (`03_tutorial/07_set_up_experiment_tracking.md`).
-* Added Plotly documentation in Kedro docs (`03_tutorial/06_visualise_pipeline.md`).
+* Added tutorial documentation for experiment tracking (`03_tutorial/07_set_up_experiment_tracking.md`).
+* Added Plotly dataset documentation (`03_tutorial/06_visualise_pipeline.md`).
+* Added the upper limit `pandas<1.4` to maintain compatibility with `xlrd~=1.0`.
 
 ## Minor breaking changes to the API
 
--- a/features/windows_reqs.txt
+++ b/features/windows_reqs.txt
@@ -2,7 +2,7 @@
 # e2e tests on Windows are slow but we don't need to install
 # everything, so just this subset will be enough for CI
 behave==1.2.6
-pandas>=0.24.0, <1.0.4
+pandas>=0.24, <1.4
 psutil==5.6.7
 requests~=2.20
 toml~=0.10.1
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,7 @@ name = "kedro"
 here = path.abspath(path.dirname(__file__))
 
 
-PANDAS = "pandas>=0.24"
+PANDAS = "pandas>=0.24, <1.4"
 SPARK = "pyspark>=2.2, <4.0"
 HDFS = "hdfs>=2.5.8, <3.0"
 S3FS = "s3fs>=0.3.0, <0.5"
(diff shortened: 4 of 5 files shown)


lorenabalan (approved):
Tbh I think this is perfectly fine. At the end of the day, users can still install newer `pandas` themselves with `pip install pandas` rather than through `pip install kedro[pandas]`, and they would have to be careful to use `openpyxl` instead of `xlrd`. This won't be as much of a problem in 0.18 cause we default to `openpyxl` then.

datajoely (approved):

SajidAlamQB (approved):

MerelTheisenQB (approved):
👍

PR #1189, opened 2022-01-31
Upper bound pip in CI

AntonyMilneQB:
## Description
[`pip 22` was released ](https://pip.pypa.io/en/stable/news/) and everything broke. So I put in the bound `pip<22`, which fixed the pip compile issues but broke the Windows e2e tests 🤦

After much digging, I figured out that the warning you get to upgrade pip on the Windows e2e tests for some reason gets interpreted as an error by behave. So in order for these tests to pass we need to use `PIP_DISABLE_PIP_VERSION_CHECK`.

Everything is now working again apart from (a) Windows 3.6 e2e tests which are still broken with the mysterious Rust error, (b) the forked PRs environment variable issue that we have a ticket for.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -180,7 +180,6 @@ jobs:
       - run:
           name: Install dependencies
           command: conda activate kedro_builder; pip install -r features/windows_reqs.txt
-      - run: Write-Host "BLAH $PIP_DISABLE_PIP_VERSION_CHECK"
       - run:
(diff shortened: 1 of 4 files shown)


MerelTheisenQB (approved):
We should probably add another ticket to use the latest version of pip.

SajidAlamQB (approved):

idanov (approved):

PR #1128, opened 2021-12-17
Update Modular Pipeline documentation and split Micropackaging into its own page

datajoely:
Signed-off-by: datajoely <joel.schwarzmann@quantumblack.com>

## Description
- Modular pipeline docs streamlined and brought to life with Kedro-Viz integration
- Micro-packaging docs previously interwoven with introduction to modular pipelines and now made its own standalone page

## Development notes
- `make devserver` review of rendered content

## Checklist

- [x] Read the [contributing](https://github.com/quantumblacklabs/kedro/blob/main/CONTRIBUTING.md) guidelines
- [x] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [x] Updated the documentation to reflect the code changes
- [x] Added a description of this change in the [`RELEASE.md`](https://github.com/quantumblacklabs/kedro/blob/main/RELEASE.md) file
- [x] Added tests to cover my changes

PR #1194, opened 2022-02-01
[KED-2970] Improve micropackaging CLI experience

SajidAlamQB:
## Description

Current problems: * The kedro pipeline command group from 0.18.0 will also package things that are not pipelines and this current naming will lead users to think the kedro pipeline package and kedro pipeline pull commands only work on pipelines

The kedro pipeline command group from 0.18.0 also has commands that don't work on things that are not pipeline modules e.g. kedro pipeline delete only works on pipelines
 
Solution: * Create a new command group called `kedro micropkg` to capture being able to package and pull things that are not pipelines.

@yetudada agreed to go with `kedro micropkg` as the new command group on the cli.

https://jira.quantumblack.com/browse/KED-2970

## Development notes

Documentation and testing WiP

## Checklist


kedro/framework/cli/pipeline.py
```diff
@@ -101,6 +101,11 @@ def pipeline():
     """Commands for working with pipelines."""
 
 
+@pipeline_cli.group()
+def micropkg():
+    """Commands for micropackaging."""
```
lorenabalan, line 106: We should separate this in its own file. We tend to have one `.py` file per CLI group (e.g. jupyter, pipeline. etc.), so this would keep it nice and tidy.

lorenabalan: It would also not be piggy-backing on "pipeline_cli" which makes it confusing to read.

kedro/framework/cli/pipeline.py
```diff
@@ -263,6 +268,9 @@ def pull_package(  # pylint:disable=unused-argument, too-many-arguments
     click.secho(message, fg="green")
 
 
+micropkg.add_command(pull_package)
```
lorenabalan, line 271: I don't know how well that works, given we have `@command_with_verbosity(pipeline, "pull")` around the pull command for instance? Can you check what happens if you run `kedro micropkg package -V` for instance?

kedro/framework/cli/pipeline.py
```diff
@@ -413,6 +421,9 @@ def package_pipeline(
     click.secho(message, fg="green")
 
 
+micropkg.add_command(package_pipeline)
```
lorenabalan, line 424: Also, related to what was raised earlier this week, I think we should show a deprecation warning that `kedro pipeline pull` is being deprecated in favour of `kedro micropkg pull` etc. So the two implementations would _slightly_ differ.

lorenabalan (commented):
It's a good start but we have a bit more to do. I'm not clear on the direction that you wanted to go in terms of breaking change either. In situations like this what normally happens looks like this:
a) have `micropkg` and `pipeline` live alongside each other for a while for the two commands
b) issue a deprecation warning on relevant `pipeline` commands that users should refer to `micropkg` for packaging and pulling
c) remove `pipeline package` and `pipeline pull`, as the functionality has been moved

We can either
- make them live next to each other in 0.18 , and do the removal after, perhaps in 0.19
- make them live next to each other in 0.17.7, and do the removal in 0.18

I personally prefer option 2. Even if `micropkg` in 0.17.7 isn't feature-complete (because some of it is only in `develop`), it's an early signal to our users that this is coming and it will be new way to package modular pipelines (and more) in the future.

We should also update documentation so that it recommends this new cli group instead of `pipeline package` for instance. And of course release notes at the end.

Hope that makes sense, let me know if you have any questions. :)

SajidAlamQB:
> It's a good start but we have a bit more to do. I'm not clear on the direction that you wanted to go in terms of breaking change either. In situations like this what normally happens looks like this: a) have `micropkg` and `pipeline` live alongside each other for a while for the two commands b) issue a deprecation warning on relevant `pipeline` commands that users should refer to `micropkg` for packaging and pulling c) remove `pipeline package` and `pipeline pull`, as the functionality has been moved
> 
> We can either
> 
> * make them live next to each other in 0.18 , and do the removal after, perhaps in 0.19
> * make them live next to each other in 0.17.7, and do the removal in 0.18
> 
> I personally prefer option 2. Even if `micropkg` in 0.17.7 isn't feature-complete (because some of it is only in `develop`), it's an early signal to our users that this is coming and it will be new way to package modular pipelines (and more) in the future.
> 
> We should also update documentation so that it recommends this new cli group instead of `pipeline package` for instance. And of course release notes at the end.
> 
> Hope that makes sense, let me know if you have any questions. :)

I believe @yetudada wanted them to live next to each other per her comment on the [Jira ticket](https://jira.quantumblack.com/browse/KED-2970) so we can monitor the adoption of `micropkg`.

kedro/framework/cli/pipeline.py
SajidAlamQB: `-V `is not an option for this command, there is a `-h` option that prints the other options available for `kedro micropkg package`

yetudada:
> > It's a good start but we have a bit more to do. I'm not clear on the direction that you wanted to go in terms of breaking change either. In situations like this what normally happens looks like this: a) have `micropkg` and `pipeline` live alongside each other for a while for the two commands b) issue a deprecation warning on relevant `pipeline` commands that users should refer to `micropkg` for packaging and pulling c) remove `pipeline package` and `pipeline pull`, as the functionality has been moved
> > We can either
> > 
> > * make them live next to each other in 0.18 , and do the removal after, perhaps in 0.19
> > * make them live next to each other in 0.17.7, and do the removal in 0.18
> > 
> > I personally prefer option 2. Even if `micropkg` in 0.17.7 isn't feature-complete (because some of it is only in `develop`), it's an early signal to our users that this is coming and it will be new way to package modular pipelines (and more) in the future.
> > We should also update documentation so that it recommends this new cli group instead of `pipeline package` for instance. And of course release notes at the end.
> > Hope that makes sense, let me know if you have any questions. :)
> 
> I believe @yetudada wanted them to live next to each other per her comment on the [Jira ticket](https://jira.quantumblack.com/browse/KED-2970) so we can monitor the adoption of `micropkg`.

I'm happy to let them live next to each other for 0.17.x but in 0.18.0, the `pipeline package` and` pipeline pull` should be removed. It just bloats the CLI.

kedro/framework/cli/pipeline.py
yetudada, line 106:
```suggestion
    """Commands for working with micro-packages."""
```

yetudada, line 106:
```suggestion
    """Commands for working with micro-packages."""
```

datajoely:
Make sure you keep aligned or even maybe take over #1128

SajidAlamQB:
Opening new PR on Main: https://github.com/kedro-org/kedro/pull/1224

Issue #1193, opened 2022-02-01
[KED-3080] Add support for click==8.x


Galileo-Galilei:
## Description

I use various packages in my kedro projects, and some raises conflicts on install due to the kedro requirements ``click<8.0.0``. I wish you unpin the requirements to support ``click==0.8.x`` to enable the use of other packages.

## Context

I tried to install ``black=22.1.0`` (the first official stable version of black :tada:), and I got this error from the pip resolver:

```
ERROR: Cannot install <your-project >and kedro==0.16.5 because these package versions have conflicting dependencies.

The conflict is caused by:
    black 22.1.0 depends on click>=8.0.0
    kedro 0.16.5 depends on click<8.0
```

Notice that I use ``kedro==0.16.5``, but the harcoded ``click<8.0`` is still present in all version, including [develop](https://github.com/kedro-org/kedro/blob/73880985b23b6d42b0bbaea329249499c5c7589c/requirements.txt#L3)`so the problem will remain in ``kedro==0.18.x``

## Possible Implementation

Relax requirements to ``click<9.0``. I don't know what are the impacts on the code though: did you pin the version just to be safe that code won't be broken, or were there actual issues with upgrading pip ?

lorenabalan:
Hi @Galileo-Galilei , this has been enabled in `develop` and will be available in 0.18.0.

Galileo-Galilei:
Thank you very much! Is there any plan to upgrade black too? The [template is still using a pre-release version](https://github.com/kedro-org/kedro/blob/4e648779be0dc5abd40b5b78d0e94ccd11910081/kedro/templates/project/%7B%7B%20cookiecutter.repo_name%20%7D%7D/src/requirements.txt#L1), while a [first official no prelease version has been released recently](https://github.com/psf/black/releases/tag/22.1.0), which comes with [a stability policy](https://black.readthedocs.io/en/latest/the_black_code_style/index.html#stability-policy).
I discovered kedro was using an old click when I tried to upgrade black for our internal template to be honest ^^'

PR #1196, opened 2022-02-01
Bump click version

datajoely:
Bump click version - main difference between 7.x and 8.x is they "dropped support for Python 2 and 3.5" which wont affect Kedro anyway 

https://click.palletsprojects.com/en/8.0.x/changes/

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 anyconfig~=0.10.0
 cachetools~=4.1
-click<8.0
+click<9.0
 cookiecutter~=1.7.0
 dynaconf>=3.1.2,<4.0.0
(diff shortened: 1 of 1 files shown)


datajoely:
Created this PR to see if CI passes

datajoely:
Referencing this ticket which notes that popular libraries like Black now use newer version of Click
https://github.com/kedro-org/kedro/issues/1193

datajoely:
Since the behave tests still work I think? this is unproblematic

SajidAlamQB:
Thanks @datajoely we will add this task to Backlog we want to ensure Kedro Telemetry doesn't break from this.

PR #1202, opened 2022-02-02
Fix 3.9 win builds

MerelTheisenQB:
## Description
The windows python 3.9 builds were failing because `pip install tables` doesn't work. 

## Development notes
I changed the test_requirements to just not install `tables` when we're using python 3.9. This is fine because we have a separate step in our setup to install it through conda. 

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -455,47 +455,47 @@ workflows:
   regular:
     unless: <<pipeline.parameters.release_kedro>>  # don't run if 'release_kedro' flag is set
     jobs:
-      - e2e_tests:
-          matrix:
-            parameters:
-              python_version: ["3.7", "3.8", "3.9"]
+#      - e2e_tests:
+#          matrix:
+#            parameters:
+#              python_version: ["3.7", "3.8", "3.9"]
       - win_e2e_tests:
           matrix:
             parameters:
               python_version: ["3.7", "3.8", "3.9"]
-      - unit_tests:
-          matrix:
-            parameters:
-              python_version: ["3.7", "3.8", "3.9"]
+#      - unit_tests:
+#          matrix:
+#            parameters:
+#              python_version: ["3.7", "3.8", "3.9"]
       - win_unit_tests:
           matrix:
             parameters:
               python_version: ["3.7", "3.8", "3.9"]
-      - lint:
-          matrix:
-            parameters:
-              python_version: ["3.7", "3.8", "3.9"]
-      - pip_compile:
-          matrix:
-            parameters:
-              python_version: ["3.7", "3.8", "3.9"]
+#      - lint:
+#          matrix:
+#            parameters:
+#              python_version: ["3.7", "3.8", "3.9"]
+#      - pip_compile:
+#          matrix:
+#            parameters:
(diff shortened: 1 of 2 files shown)


SajidAlamQB (approved):

AntonyMilneQB:
Thanks for fixing this! But I'm actually not sure whether it's just hiding the real problem that `pandas.HDFDataSet` might not work on Windows, Python 3.9 🤔  It works on our CI because we're doing `conda install pytables`, but what about if I'm actually trying to install kedro as a user?

[We have this](https://github.com/kedro-org/kedro/blob/main/setup.py#L80-L84):
```
    "pandas.HDFDataSet": [
        PANDAS,
        "tables~=3.6.0; platform_system == 'Windows'",
        "tables~=3.6; platform_system != 'Windows'",
    ],
```
So if I'm installing `kedro[pandas.HDFDataSet]` on Windows, Python 3.9 what should happen? 

This actually strikes me as a mistake in the [original PR](https://github.com/quantumblacklabs/private-kedro/pull/1093/files) that added Python 3.9 support to CI so happy to deal with it in a separate ticket/PR.

AntonyMilneQB:
Ah, I see this was actually discussed already in https://github.com/quantumblacklabs/private-kedro/pull/1093/files (check the top two resolved conversations). Sounds like a user needs to `conda install pytables` anyway to use this dataset? In which case it doesn't really matter what goes in setup.py.

MerelTheisenQB:
> Thanks for fixing this! But I'm actually not sure whether it's just hiding the real problem that `pandas.HDFDataSet` might not work on Windows, Python 3.9 🤔 It works on our CI because we're doing `conda install pytables`, but what about if I'm actually trying to install kedro as a user?
> 
> So if I'm installing `kedro[pandas.HDFDataSet]` on Windows, Python 3.9 what should happen?

This is a really good point! It would be great if we could get someone to test it for us. Or do another attempt at setting up a window environment ourselves and test it there. 

> This actually strikes me as a mistake in the [original PR](https://github.com/quantumblacklabs/private-kedro/pull/1093/files) that added Python 3.9 support to CI so happy to deal with it in a separate ticket/PR. Or, if our advice is just to `conda install tables` instead of pip install then we should probably change the lines in setup.py to match these ones in test_requirements.txt.

I'll create a proper ticket for this issue, so we can unblock the CI for now. Totally agree we should look into this in more detail.

MerelTheisenQB:
> Ah, I see this was actually discussed already in https://github.com/quantumblacklabs/private-kedro/pull/1093/files (check the top two resolved conversations). Sounds like a user needs to `conda install pytables` anyway to use this dataset? In which case it doesn't really matter what goes in setup.py.

Nice digging! 🕵️  I'll still create a ticket and will add that a solution could just be better documentation.

AntonyMilneQB:
@MerelTheisenQB - just in case you missed it, check my comment just before yours - looks like you and Lorena and Ignacio talked about this before actually.

As a general point actually if we're considering in a new ticket what is the correct ultimate solution here: why do we want `test_requirements.txt` to contain these datasets dependencies? Wouldn't it be better if we just did `pip install kedro[all]` so that we're actually testing the stuff in setup.py rather than whatever is in our CI environment?

Issue #1134, opened 2021-12-29
PickleDataset is not deepcopiable for kedro>=0.17.6


Galileo-Galilei:
## Description

``PickleDataset`` was  modified in ``kedro==0.17.6`` to support any arbitrary backend.

The backend module is now stored as an attribute of the dataset: 

https://github.com/quantumblacklabs/kedro/blob/319a9177c45fac4fe55a0805bd2d83e6d4fa59f6/kedro/extras/datasets/pickle/pickle_dataset.py#L178

## Context

I tried to deepcopy the catalog  (in order to use it in a custom mlflow model : https://github.com/Galileo-Galilei/kedro-mlflow/blob/master/kedro_mlflow/mlflow/kedro_pipeline_model.py#L204), and got the following error : ``TypeError: can't pickle module objects``. 

After investigation, it turns out that the error comes from the ``PickleDataSet`` which is no longer deepcopiable.

## Steps to Reproduce
1. Install kedro 0.17.6 in a new environment  ``conda create -n bug_dc python=3.7 kedro==0.17.6 -y``
2. ``conda activate bug_dc``
3. Run this script:
```python
from kedro.extras.datasets.pickle import PickleDataSet
from copy import deepcopy

pkl_ds = PickleDataSet(
    filepath=r"path/to/my_data.pkl"
)
pkl_ds2=deepcopy(pkl_ds) # raise the error
pkl_ds3=pkl_ds._copy() # same error with the builtin function
```

## Expected Result

The dataset should be deecopied.

## Actual Result

Get the following error
```python 
TypeError: can't pickle module objects
```

## Your Environment

* Kedro version used (`pip show kedro` or `kedro -V`): 0.17.6
* Python version used (`python -V`): 3.7.10
* Operating system and version: Windows 10

## Potential solutions

- For now, I will use the built-in ``DataCatalog.shallow_copy()`` function, but this is not a deepcopy and the underlying datasets have the same memory id.
- It may be useful to make _backend a property so that the import is made on the fly when using the attributes, rather than storing the entire module,e.g. :

````python

class PickleDataSet(AbstractVersionedDataSet):

    # self._backend is the module, self.backend is the string
    @property
    def _backend(self):
        # move the import here and remove it from __init__
        try:
            imported_backend = importlib.import_module(self.backend) # self.backend is the string
        except ImportError as exc:
            raise ImportError(
                f"Selected backend '{backend}' could not be imported. "
                "Make sure it is installed and importable."
[...]

PR #1213, opened 2022-02-04
Make PickleDataSet deepcopy-able

AntonyMilneQB:
## Description
Fixes https://github.com/kedro-org/kedro/issues/1134.

## Development notes
@Galileo-Galilei I've kept the `ImportError` to being raised in `__init__` rather than in `save`/`load`. This is consistent with how other datasets work and means any import errors get flagged as soon as possible.

In hindsight, `self._backend` should have remained as a string in [this PR](https://github.com/kedro-org/kedro/pull/931) which first extended `PickleDataSet` to accept an arbitrary backend because `backend` is also used in the `_describe` method. I've changed it back to a string.

Doing the import every `save`/`load` method shouldn't have any significant performance hit because [Python doesn't reload the whole package on repeated import](https://stackoverflow.com/questions/37067414/python-import-multiple-times).

In addition to the two new tests I've also checked that this now works with parallel runner now (which it didn't before) and the script that @Galileo-Galilei provided.

## Checklist


MerelTheisenQB (approved):
LGTM! 👍  

Minor: How did you check it now works with the `ParallelRunner`? Do we need a test for that?

kedro/extras/datasets/pickle/pickle_dataset.py
```diff
@@ -203,14 +203,16 @@ def _load(self) -> Any:
         load_path = get_filepath_str(self._get_load_path(), self._protocol)
 
         with self._fs.open(load_path, **self._fs_open_args_load) as fs_file:
-            return self._backend.load(fs_file, **self._load_args)  # type: ignore
+            imported_backend = importlib.import_module(self._backend)
```
lorenabalan, line 206: Optional, very minor: naming this the same to what we use in `__init__` made me instantly think "why not just save it on the instance as an attribute", which would make us end up in the same bad spot we were before. I think it's possible that other people would have the same thought reading the code. Maybe we can use a different variable name, `pickle_lib`, `pickling_lib`, `pickle_backend` etc.

lorenabalan (approved):
Looks good! I know we added some tests now that should document this behaviour, but I don't think it's immediately obvious - I think it'd be worth adding some in-line comments about why we import twice rather than saving it as an attribute, or why we import on `__init__` and not just on `load/save` (which btw I like that we do that, this is just documenting it for future maintainers).

kedro/extras/datasets/pickle/pickle_dataset.py
AntonyMilneQB: I deliberately made it `imported_backend` (rather than `backend`) exactly for this reason 🙂  Don't know whether you misread it or whether you think the difference in variable names is not clear enough yet?

AntonyMilneQB:
@MerelTheisenQB Testing with the `ParallelRunner` was just a manual `kedro run` on a pipeline that includes a PickleDataSet and did not run before the fix. I considered adding a new test for this like you suggest, but decided it wasn't required since we don't do special ParallelRunner tests for other datasets, and the thing that made this incompatible with ParallelRunner (not being able to copy) is now covered by new tests.

@lorenabalan Very good point about adding comments, will do so now!

Issue #1208, opened 2022-02-03
Kedro build-reqs breaks with pip>=22.0


luizvbo:
## Description
If I run `kedro build-reqs` with `pip>=22.0`, an error is raised by piptools. I assume this bug was introduced by the release of the version `22.x` of `pip` on 30/01/2022.

## Context
I update the version of my `pip` package to `22.0.3` and tried to build the requirements of my kedro package with `kedro build-reqs`, which raises the following error:

```
/tmp/test/kedro-example/example-venv/bin/python3 -m piptools compile -q /tmp/kedro-example/spaceflights-project/src/requirements.in
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/tmp/test/kedro-example/example-venv/lib/python3.8/site-packages/piptools/__main__.py", line 17, in <module>
    cli()
  File "/tmp/test/kedro-example/example-venv/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/tmp/test/kedro-example/example-venv/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/tmp/test/kedro-example/example-venv/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/tmp/test/kedro-example/example-venv/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/tmp/test/kedro-example/example-venv/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/tmp/test/kedro-example/example-venv/lib/python3.8/site-packages/click/decorators.py", line 21, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/tmp/test/kedro-example/example-venv/lib/python3.8/site-packages/piptools/scripts/compile.py", line 342, in cli
    repository = PyPIRepository(pip_args, cache_dir=cache_dir)
  File "/tmp/test/kedro-example/example-venv/lib/python3.8/site-packages/piptools/repositories/pypi.py", line 106, in __init__
    self._setup_logging()
[...]

PR #1215, opened 2022-02-04
Bump pip-tools version

AntonyMilneQB:
## Description
Fixes https://github.com/kedro-org/kedro/issues/1208. In short, `pip-tools==6.4.0` is not compatible with the recently released `pip>=22` . `pip-tools` have now released a fix version `6.5.0` which does work with `pip>=22`, but it's only compatible with Python 3.7+.

I've also reversed the changes to CI made in https://github.com/kedro-org/kedro/pull/1189 so that the pip requirement is now open-ended again. I actually think this is a good system because if pip releases a new version that breaks our CI it gives us an immediate warning that other stuff will likely break also.

## Checklist


MerelTheisenQB (approved):

RELEASE.md
```diff
@@ -9,6 +9,7 @@
 * Added Plotly dataset documentation (`03_tutorial/06_visualise_pipeline.md`).
 * Added the upper limit `pandas<1.4` to maintain compatibility with `xlrd~=1.0`.
 * Bumped the `Pillow` minimum version requirement to 9.0 (Python 3.7+ only) following [CVE-2022-22817](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-22817).
+* Upgraded `pip-tools`, which is used by `kedro build-reqs`, to 6.5 (Python 3.7+ only). This `pip-tools` version is compatible with `pip>=21.2`, including the most recent releases of `pip`.
```
lorenabalan, line 12: Can we make a mention that for users still on python 3.6, they need to use older version of `pip`, i.e. <22? Or did I misunderstand this?

lorenabalan (approved):

Makefile
```diff
@@ -10,7 +10,7 @@ clean:
 	pre-commit clean || true
 
 install-pip-setuptools:
-	python -m pip install -U "pip~=21.2" "setuptools>=38.0" wheel
+	python -m pip install -U "pip>=21.2" "setuptools>=38.0" wheel
```
lorenabalan, line 13: I'm still not clear how this is fine on python 3.6 if we install piptools 6.4.

AntonyMilneQB: This is a great question... The reason (which I figured out last week but then completely forgot about) is that [pip 22 drops support for Python 3.6](https://pip.pypa.io/en/stable/news/#v22-0). So on the 3.6 environment we'll get piptools 6.4 and pip 21.x, which is all good.

lorenabalan: Aaah that makes sense! Wow nice one, great digging. 👏

RELEASE.md
AntonyMilneQB, line 12:
```suggestion
* Upgraded `pip-tools`, which is used by `kedro build-reqs`, to 6.5 (Python 3.7+ only). This `pip-tools` version is compatible with `pip>=21.2`, including the most recent releases of `pip`. Python 3.6 users should continue to use `pip-tools` 6.4 and `pip<22`.
```

AntonyMilneQB: Python 3.6 users will be on `pip<22` anyway, but I've added a note to be explicit here (since I also had to think about this for a while to figure out what was going on...😀 )

PR #1226, opened 2022-02-08
Alias `astro-iris` to `astro-airflow-iris`

MerelTheisenQB:

## Description
Creating a project with `kedro new --starter astro-iris` doesn't work since `astro-iris` was renamed to `astro-airflow-iris`

We should support both `astro-iris` and `astro-airflow-iris` or at least give a warning to the user that the starter is now renamed. 

## Development notes
Check if a user has entered `astro-iris` as the starter name and change it to `astro-airflow-iris`. 

While working on this I was thinking whether it would make sense to make a generic change where if no template or known starter is found we print an error message with the official starter names. I'll create another PR for that change. 

## Checklist

### DIFF
--- a/kedro/framework/cli/starters.py
+++ b/kedro/framework/cli/starters.py
@@ -91,6 +91,9 @@ def new(
             "Cannot use the --directory flag without a --starter value."
         )
 
+    starter_name = (
+        "astro-airflow-iris" if starter_name == "astro-iris" else starter_name
+    )
     if starter_name in _STARTER_ALIASES:
         if directory:
             raise KedroCliError(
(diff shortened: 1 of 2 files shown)


kedro/framework/cli/starters.py
```diff
@@ -91,6 +91,9 @@ def new(
             "Cannot use the --directory flag without a --starter value."
         )
 
+    starter_name = (
```
datajoely, line 94: Maybe include a little inline comment why we're doing this?

MerelTheisenQB: Done!

PR #1237, opened 2022-02-10
[KED-2306] Document how **kwargs-only functions can receive a mapping of dataset name and data

lorenabalan:

## Description

This pattern has been sth we wanted to document for absolute ages so here it (finally) is.

## Development notes

Plus the CI is builds docs only! 😍 

## Checklist

### DIFF
--- a/docs/source/06_nodes_and_pipelines/01_nodes.md
+++ b/docs/source/06_nodes_and_pipelines/01_nodes.md
@@ -101,6 +101,69 @@ There is a syntax to describe function inputs and outputs. This allows different
 
 Any combinations of the above are possible, except nodes of the form `node(f, None, None)` (at least a single input or output needs to be provided).
 
+## `**kwargs`-only node functions
+
+Sometimes, when creating reporting nodes for instance, you need to know the names of the datasets that your node receives, but you may not have this information in advance. This can be solved by defining a `**kwargs`-only function, like so:
+
+```python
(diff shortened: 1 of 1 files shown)


docs/source/06_nodes_and_pipelines/01_nodes.md
```diff
@@ -101,6 +101,69 @@ There is a syntax to describe function inputs and outputs. This allows different
...
+            inputs=mapping(["ge_input1", "ge_input2", ...]),
+            outputs="ge",
+        ),
+        ...,
+    ]
+)
+```
+
+```eval_rst
+.. note:: `**kwargs` might require you to replace all special characters with `_` in the dataset names, to ensure they are valid Python variable names.
```
idanov, line 160: Just checked that, it seems that Python doesn't care much about the names. So we might drop this note here. The check I did was the following one:

```
def hello(**kwargs):
    for x in kwargs: print(x)

h = {"hello.world": "value", "%^&hello": "result"}
hello(**h)
```

idanov (approved):

docs/source/06_nodes_and_pipelines/01_nodes.md
lorenabalan: Oh wow I'm surprised that works! 😮  I'll remove the note then, and TIL! ⚡ 💡

Issue #1041, opened 2021-11-15
Universal Kedro deployment (Part 3) - Add the ability to extend and distribute the project running logic


Galileo-Galilei:
## Preamble

This is the third part of my serie of design documentation on refactoring Kedro to make deployment easier:

- The first part is described in the issue #770 and focuses on refactoring configuration to separate external and applicative configuration.
- The second part is described in the issue #904 focuses on ``DataCatalog`` entries which have a compute/storage backend different than "python / in memory operations" (including SQl, Spark...). 
- This third part focuses on the ability to modify the running logic at runtime, outside of the ``KedroSession``.

### Current state of Kedro's extensibility

There are currently several ways to extend Kedro natively, described hereafter:

|What is extended|Use cases example|Kedro object|Registration|Popularity|
|---|---|---|---|---|
|Pipeline execution at runtime|-  change catalog entries on the fly (cache data, change git branch... <br /> - log data remotely (mlflow, neptune, dolt, store kedro-viz static files...) |Hooks (Pipeline, node)| - via an entrypoint <br /> -  OR manual declaration in settings.py|High: [a quick github search]((https://github.com/search?p=1&q=before_pipeline_run&type=Code)) shows that many users use hooks to add custom logic at runtime|
|CLI command|- create a configuration file <br /> - profile a catalog entry <br /> - convert a kedro pipeline to an orchestrator <br /> - visualize the pipeline in a webrowser…|plugin click commands|- via an entrypoint|Medium: Seem to be a more adavced use mainly for plugin developpers|
|Data sources connection|- Create a dataset which can connect to a new data source unsupported by kedro (GBQ, HDF, sklearn pipelines, Databrics, Stata, redis,…. are the most recent ones)|AbstractDataSet|As a module, which can be imported by its path in the DataCatalog|High: [a quick search in Kedro's past issues](https://github.com/quantumblacklabs/kedro/issues?q=is%3Aissue+dataset) shows that it is very common request for users who need to connect to specific data sources|

### Use cases not covered by previous mechanisms

However, I've encountered a bunch of use case where people want to extend the **running logic** (=how to run the pipeline) rather than of the execution logic (=how the pipeline behaves during runtime, which is achieved by hooks). Some examples includes: 
[...]

PR #60, opened 2019-07-18
Add only_missing option in KedroContext class 

Minyus:
## Notice

- [ x] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

- I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
- I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorised to submit this contribution on behalf of the original creator(s) or their licensees.
- I certify that the use of this contribution as authorised by the Apache 2.0 license does not violate the intellectual property rights of anyone else.

## Motivation and Context
`run_only_missing` option was missing in `KedroContext` class introduced by `[KED-826] Use KedroContext class in project template (#111)`

## How has this been tested?
1. Run `kedro new` and create a new kedro project including the iris data pipeline.
2. Modify `context.run(tags, runner)` to `context.run(tags, runner, only_missing=True)` in `run.py`
image
3. Run `kedro run` and confirm the 4 tasks are processed by the logged message.
4. Run `kedro run` again and confirm the 4 tasks are not processed again because of the `only_missing` option. 

image

## Checklist

- [x ] Read the [contributing](/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added new entries to the `RELEASE.md` file
- [ ] Added tests to cover my changes
- [ ] Assigned myself to the PR

Issue #30, opened 2019-06-24
Add only-missing option to kedro run command


gotin:
## Description
While Runner has the run_only_missing method which is very useful especially when a pipeline is being developed incrementally, but kedro run command doesn't have an option for it. If it has the option do Runner#run_only_missing, it may be useful.

## Context
If kedro run has an option for doing Runner#run_only_missing, we can skip steps which have been already done before, it makes pipeline developments' more productive.

## Possible Implementation
Add the run function of kedro_cli.py an argument for only_missing behavior,  and add the run function of main function of run.py an argument for only_missing as well.
(This implementation has already been tried at my local environment, and it's working well)

## Checklist
Include labels so that we can categorise your issue:
- [Type: Enhancement]
- [Priority: Medium]

yetudada:
Hi @gotin! We're so glad you've been able to extend the functionality of CLI. We have left it open for users to extend uses of the CLI. Could you talk to us about how only_missing works for you?

gotin:
When I'm developing a pipeline incrementally with huge size of data, run a whole pipeline from the beginning is time-consuming, and less productive.  So, I want to run only nodes which is just created and haven't been run by using 'only-missing' flag. It can omit running nodes which have already been run.
I think we utilize --tag option of kedro run command for this purpose, but sometimes it's difficult for me to tag each node with a suitable name.

Minyus:
I almost filed a duplicate of this issue. 
I totally agree with @gotin . 

`run_only_missing` feature is a big advantage of kedro and it is inconvenient to modify 2 modules (`run.py` and `kedro_cli.py`) every time after running kedro new to use the feature. 

I would suggest kedro add run_only_missing option to kedro run command too.

idanov:
@gotin @Minyus We're in the process of reorganising some things around how Kedro projects are ran and ideally we would like to add an easy way for people to add their own options for running. We also plan to extend the set of options to the `run` command with options like `kedro run --node <node-name>` or `kedro run --from-inputs <input1>,<input2>`, would that solve the problem for you?

gotin:
I’m sure node option and from inputs option are good for some situations. But those can’t work for my situation. What I want really is nothing but invoking run_only_missing from kedro command. I just want to run new nodes which haven’t run since they added.

Minyus:
Same opinion as @gotin
`run_only_missing` is what we want.

@idanov Regarding the extended options, I would suggest the interface like this.

```
kedro run \
[...]

PR #1248, opened 2022-02-12
Document distribution of Kedro pipelines with Dask

deepyaman:
## Description

Update #1131, to which I no longer have write access. This can be either merged directly or merged into that PR, which can then be merged.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1248"></a>


RELEASE.md
```diff
@@ -3,6 +3,7 @@
 ## Major features and improvements
 * `pipeline` now accepts `tags` and a collection of `Node`s and/or `Pipeline`s rather than just a single `Pipeline` object. `pipeline` should be used in preference to `Pipeline` when creating a Kedro pipeline.
 * `pandas.SQLTableDataSet` and `pandas.SQLQueryDataSet` now only open one connection per database, at instantiation time (therefore at catalog creation time), rather than one per load/save operation.
+* Documented distribution of Kedro pipelines with Dask.
```
lorenabalan, line 6: Apologies, this will have to move to a new Release header.

RELEASE.md
```diff
@@ -406,7 +407,7 @@ Check your source directory. If you defined a different source directory (`sourc
 
 ## Major features and improvements
 
-* Added documentation with a focus on single machine and distributed environment deployment; the series includes Docker, Argo, Prefect, Kubeflow, AWS Batch, AWS Sagemaker and extends our section on Databricks
+* Added documentation with a focus on single machine and distributed environment deployment; the series includes Docker, Argo, Prefect, Kubeflow, AWS Batch, AWS Sagemaker and extends our section on Databricks.
```
lorenabalan, line 410: 😂

docs/source/03_tutorial/05_visualise_pipeline.md
```diff
@@ -17,7 +17,7 @@ kedro viz
 ```
 
 This command will run a server on http://127.0.0.1:4141 that will open up your visualisation on a browser. You should
- be able to see the following:
+be able to see the following:
```
lorenabalan, line 20: can we actually just move it on the same line? We don't break up long lines in our .md files, we just use wrap-around in IDE.

docs/source/index.rst
```diff
@@ -150,6 +150,7 @@ Welcome to Kedro's documentation!
    10_deployment/09_aws_sagemaker
    10_deployment/10_aws_step_functions
    10_deployment/11_airflow_astronomer
+   10_deployment/12_dask
```
lorenabalan, line 153: Can we actually rename this page to just `dask`? :) On `develop` we got rid of numbering in our docs, so that we don't break links as often, and this would be one less conflict to resolve on `merge-main-to-develop`.

docs/source/10_deployment/12_dask.md
```diff
@@ -0,0 +1,313 @@
+# Deployment to a Dask cluster
+
+This page explains how to distribute execution of the nodes composing your Kedro pipeline using [Dask](https://docs.dask.org/en/stable/), a flexible, open-source library for parallel computing in Python.
+
+Dask offers both a default, single-machine scheduler and a more sophisticated, distributed scheduler. The newer [`dask.distributed`](http://distributed.dask.org/en/stable/) scheduler is often preferable, even on single workstations, and is the focus of our deployment guide. For more information on the multitude of ways to set up Dask on varied hardware, see [the official how-to guide](https://docs.dask.org/en/stable/how-to/deploy-dask-clusters.html).
```
lorenabalan, line 5:
```suggestion
Dask offers both a default, single-machine scheduler and a more sophisticated, distributed scheduler. The newer [`dask.distributed`](http://distributed.dask.org/en/stable/) scheduler is often preferable, even on single workstations, and is the focus of our deployment guide. For more information on the various ways to set up Dask on varied hardware, see [the official how-to guide](https://docs.dask.org/en/stable/how-to/deploy-dask-clusters.html).
```

docs/source/10_deployment/12_dask.md
```diff
@@ -0,0 +1,313 @@
...
+
+## Why would you use Dask?
+
+`Dask.distributed` is a lightweight library for distributed computing in Python. It complements the existing PyData analysis stack, which forms the basis of many Kedro pipelines. It's also pure Python, which eases installation and simplifies debugging. For further motivation on why people choose to adopt Dask, and, more specifically, `dask.distributed`, see [Why Dask?](https://docs.dask.org/en/stable/why.html) and [the `dask.distributed` docs](http://distributed.dask.org/en/stable/#motivation), respectively.
```
lorenabalan, line 9:
```suggestion
`Dask.distributed` is a lightweight library for distributed computing in Python. It complements the existing PyData analysis stack, which forms the basis of many Kedro pipelines. It's also pure Python, which eases installation and simplifies debugging. For further motivation on why people choose to adopt Dask, and, more specifically, `dask.distributed`, see [Why Dask?](https://docs.dask.org/en/stable/why.html) and [the `dask.distributed` documentation](http://distributed.dask.org/en/stable/#motivation), respectively.
```

docs/source/10_deployment/12_dask.md
```diff
@@ -0,0 +1,313 @@
...
+
+where the helper function `_instantiate_runner()` looks like this:
+
+```python
+def _instantiate_runner(runner, is_async, project_context):
+    runner_class = load_obj(runner, "kedro.runner")
+    runner_kwargs = dict(is_async=is_async)
+
+    if runner.endswith("DaskRunner"):
+        client_args = project_context.params.get("dask_client") or {}
```
lorenabalan, line 267: Wouldn't this be 
```suggestion
        client_args = project_context.params.get("params.dask_client") or {}
```
?

docs/source/10_deployment/12_dask.md
```diff
@@ -0,0 +1,313 @@
...
+        Args:
+            ds_name: Name of the missing dataset.
+
+        Returns:
+            An instance of ``_DaskDataSet`` to be used for all
+            unregistered datasets.
+        """
+        return _DaskDataSet(ds_name)
+
+    def _run_node(
```
lorenabalan, line 110: Does this need to be a class method, or can it be a standalone function? As far as I can tell it's just a thin wrapper around `run_node` so that you pass `dependencies`, but they're not actually used here. Presumable the dask client is?

lorenabalan (approved):

lorenabalan:
LGTM, thanks for tidying up some other parts of the docs while here! 🙌

docs/source/index.rst
deepyaman: Done!

RELEASE.md
deepyaman: Done!

docs/source/10_deployment/12_dask.md
deepyaman: Don't think so:

```
(kedro-dask-example) WASH-178551-C02X31K9JHD4:kedro-dask-example deepyaman_datta$ kedro run --env dask --runner kedro_dask_example.runner.DaskRunner
2022-02-24 22:03:44,996 - kedro.framework.cli.hooks.manager - INFO - Registered CLI hooks from 1 installed plugin(s): kedro-telemetry-0.1.3
2022-02-24 22:03:45,179 - kedro_telemetry.plugin - INFO - You have opted into product usage analytics.
2022-02-24 22:03:45,492 - kedro.framework.session.store - INFO - `read()` not implemented for `BaseSessionStore`. Assuming empty store.
> /Users/deepyaman_datta/kedro-dask-example/src/kedro_dask_example/cli.py(61)_instantiate_runner()
-> client_args = project_context.params.get("dask_client") or {}
(Pdb) project_context.params
{'example_test_data_ratio': 0.2, 'example_num_train_iter': 10000, 'example_learning_rate': 0.01, 'dask_client': {'address': '127.0.0.1:8786'}}
(Pdb) 
```

This is the same way I've been running github.com/deepyaman/kedro-dask-example.

docs/source/10_deployment/12_dask.md
deepyaman: @lorenabalan It doesn't need to be an instance method. I could make it into a static method, if so desired?  It could also be outside the class, but I think:

1. grouping for functionality is valid reason to make it into a static method
2. for the purpose of a deployment guide, having something outside the class is harder to keep track of (not really, but arguably)

Open to doing whatever with this; let me know if you have a strong preference. I can leave it as is or make it into a static method, with least preference for moving it outside the class, but can roll with anything.

lorenabalan: Yeah fair enough about the grouping aspect. Let's stick a `staticmethod` for this one.

docs/source/10_deployment/12_dask.md
lorenabalan: Oh interesting, good to be reminded of that. 😂  Thanks for checking!

docs/source/10_deployment/dask.md
```diff
@@ -0,0 +1,313 @@
...
+            ds_name: Name of the missing dataset.
+
+        Returns:
+            An instance of ``_DaskDataSet`` to be used for all
+            unregistered datasets.
+        """
+        return _DaskDataSet(ds_name)
+
+    def _run_node(
+        self,
```
deepyaman, line 111:
```suggestion
    @staticmethod
    def _run_node(
```

docs/source/10_deployment/dask.md
```diff
@@ -0,0 +1,313 @@
...
+        node_futures = {}
+
+        client = Client.current()
+        for node in nodes:
+            dependencies = (
+                node_futures[dependency] for dependency in node_dependencies[node]
+            )
+            node_futures[node] = client.submit(
+                self._run_node, node, catalog, self._is_async, run_id, *dependencies
+            )
```
deepyaman, line 153:
```suggestion
                DaskRunner._run_node,
                node,
                catalog,
                self._is_async,
                run_id,
                *dependencies,
            )
```

docs/source/10_deployment/12_dask.md
deepyaman: Done!

AntonyMilneQB (approved):
This is really awesome stuff, thank you very much @deepyaman. 

Very happy for this to be merged - the questions I ask here and in the inline comments are just to satisfy my curiosity or contribute to more general ponderings.

I know this follows the same pattern as the AWS batch runner, which is fine, but there's a couple of things that definitely seem hacky about it (as they do for the AWS batch runner). I don't immediately see any better way of doing it now, but I wonder what it would take on the kedro side to make this sort of extension more elegant in future.
1. Injecting custom arguments into `DaskRunner` instantiation via conf/dask/parameters.yml. This is very cunning, but doesn't seem ideal because it's incompatible with running any other run configuration environment (unless you also modify the ConfigLoader to allow for multiple environments to load on top of each other).
2. `def run` is essentially exactly the same as the default one; the only difference (unless I'm missing something?) is that it enables you to pass custom arguments into the runner instantiation.

I wonder if there's some other way of configuring our `run` command in order to make it easier to do this sort of thing. This issue seems very relevant, and I'd be interested if you had any thoughts to add to it based on what you've seen here: https://github.com/kedro-org/kedro/issues/1041

Naively it seems like we could expose `RUNNER_CLASS` and `RUNNER_CLASS_ARGS` in settings.py to enable this sort of thing more directly. But given that the runner is arguably runtime configuration (that belongs in conf) rather than application settings (that belongs in settings.py) that probably doesn't make sense. Sooo I don't know how we can support custom runner + arguments a bit more naturally.

docs/source/10_deployment/dask.md
```diff
@@ -0,0 +1,313 @@
...
+            # for the shared, default datasets we created above.
+            for data_set in node.inputs:
+                load_counts[data_set] -= 1
+                if load_counts[data_set] < 1 and data_set not in pipeline.inputs():
+                    catalog.release(data_set)
+            for data_set in node.outputs:
+                if load_counts[data_set] < 1 and data_set not in pipeline.outputs():
+                    catalog.release(data_set)
+
+    def run_only_missing(
```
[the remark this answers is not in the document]
deepyaman: 1. I implemented this because I felt it was especially relevant for Dask. The `DaskDataSet` actually publishes data to the cluster, so there's less progress lost than in the case of `MemoryDataSet` in case of an error.
2. I did test it months ago when I wrote the behavior, but I probably hacked it in somewhere for testing. Not aware of an easy call off the top of my head (but it probably should be!). Looks like it was raised a few years ago (#30, #60) but decided against back then.

docs/source/10_deployment/dask.md
```diff
@@ -0,0 +1,313 @@
...
+
+Next, [set up scheduler and worker processes on your local computer](http://distributed.dask.org/en/stable/quickstart.html#setup-dask-distributed-the-hard-way):
+
+```bash
+$ dask-scheduler
+Scheduler started at 127.0.0.1:8786
+
+$ PYTHONPATH=$PWD/src dask-worker 127.0.0.1:8786
+$ PYTHONPATH=$PWD/src dask-worker 127.0.0.1:8786
+$ PYTHONPATH=$PWD/src dask-worker 127.0.0.1:8786
```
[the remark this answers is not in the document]
deepyaman: You need to somehow make the code available to the worker, and in case of a single-machine scheduler, this works.

`Client.upload_file` is cleaner, and something like that would be necessary for distributed deployment.

https://stackoverflow.com/a/39994128/1093967 for some more details.

deepyaman:
> P.S. fun fact, don't know if you're aware of the [recent change](https://github.com/kedro-org/kedro/pull/1225): if you hadn't edited in `RELEASE.md` then this PR would only be running the docs workflows on CircleCI. Possibly we should make that regex just look for .md files (rather than files just in docs/) so that PRs like this don't trigger all the code builds.

I noticed today when looking at the Windows/Python 3.6 errors that you all adopted dynamic config, nice!

docs/source/10_deployment/dask.md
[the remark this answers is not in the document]
AntonyMilneQB: Wow, great work searching the archives to find that! I wasn't even aware that `run_only_missing` existed until a couple of months ago when Nikos mentioned it.

datajoely:
Well done @deepayman :pray:

PR #1252, opened 2022-02-14
Feature/tidy settings

AntonyMilneQB:
## Checklist


kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/settings.py
```diff
@@ -1,40 +1,41 @@
 """Project settings."""
```
datajoely, line 1: I would love to see us elaborate on what `settings.py` is versus `pyproject.toml` and `conf/ `

Basically this image:
image

docs/source/kedro_project_setup/settings.md
```diff
@@ -4,28 +4,36 @@ A Kedro project's `settings.py` file contains the application configuration for
 
 By default, all code in `settings.py` is commented out. In the case that settings are not supplied, Kedro chooses sensible default values. You only need to edit `settings.py` if you wish to change to values other than the defaults.
 
-## Available settings
+## Lifecycle settings
```
AntonyMilneQB, line 7: Don't try to understand this diff... Just look at the picture in the PR description instead.

MerelTheisenQB (approved):
Nice work! I like the division in `lifecycle` and `configuration` and `settings.py` looks a lot cleaner now 🧼 😃 

(Also TIL the word _pukka_  and that this emoji exists 🧇 )

kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/settings.py
```diff
@@ -1,40 +1,43 @@
...
-
-# Define keyword arguments to be passed to `SESSION_STORE_CLASS` constructor
+# Keyword arguments to pass to the `SESSION_STORE_CLASS` constructor.
 # SESSION_STORE_ARGS = {
 #     "path": "./sessions"
 # }
 
-# Define custom context class. Defaults to `KedroContext`
+# Class that manages Kedro's library components.
 # CONTEXT_CLASS = KedroContext
```
lorenabalan, line 25: This is the only one that doesn't have an import above it so maybe it'd be a good idea to include that as well.

lorenabalan (commented):
I don't know if I'm a fan of the lifecycle/configuration naming or split. You could argue the `KedroContext` is neither or both (even if it disappears or not in the future). "configuration" is very vague to me, and first time reading it my first thought was "well everything is configuration. yaml, settings, CLI params".. I understand why you thought of that, cause it's stuff to do with config files. But in the future if we add an entry to make `pipelines_registry.py` configurable, or `register_pipelines()` configurable, that's more to do with _template_ config, right?
I'm not yet convinced we have clear categories on what should go in `settings.py`, at least they're definitely not 100% clear in my head, which is why I'm reluctant to divide them up already. If we do, then this is my chance to learn and consolidate that in my brain. 🧠  Keen to hear other people's thoughts?

lorenabalan (approved):
Sorry forgot to approve, I'm just being pedantic on the naming at the end of the day, the docs and file looks fine to me otherwise. I would still like to leave the question above open though.

AntonyMilneQB:
Thanks for the comments everyone! After more thought I've gone with Lorena's here. I think splitting into sections was a sort of premature optimisation. It might be worth doing in the future if the file gets longer and/or we have a clearer idea about what the sections should actually be. "Configuration" and "Lifecycle" were too handwavey and just made up by me for my liking 😅 

I've also made sure everything has the relevant import above it in settings.py and added some information to the docs about the difference between conf, settings.py and pyproject.toml.

kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/settings.py
```diff
@@ -1,40 +1,37 @@
...
-# Define the project ConfigLoader class here.
-# Defaults to kedro.config.ConfigLoader
+# Class that manages how configuration is loaded.
 # from kedro.config import TemplatedConfigLoader
 # CONFIG_LOADER_CLASS = TemplatedConfigLoader
-
-# Define keyword arguments to be passed to `CONFIG_LOADER_CLASS` constructor.
-# These kwargs depend on the `ConfigLoader` class implementation.
+# Keyword arguments to pass to the `CONFIG_LOADER_CLASS` constructor.
 # CONFIG_LOADER_ARGS = {
```
datajoely, line 31: Why don't we also have the catalog and parameter patterns here?

AntonyMilneQB: Hmm yes, I suppose we could potentially make this configurable. It would need to work as an option for the `CONTEXT_CLASS` rather than `CONFIG_LOADER` I think, since that's where these patterns are actually used. Currently it is possible to do this by inheriting from `KedroContext` but it's a bit awkward since you need to overwrite some methods. We could make it easier by making these patterns class attributes or something.

PR #1254, opened 2022-02-15
Remove code for backwards compatibility

MerelTheisenQB:

## Description
We still had some code in place for backwards compatibility that should be removed in `0.18.0`.

## Development notes
- Removed code in `kedro.framework.project._init_.py` there for backwards compatibility with projects that don't have the `pipeline_registry.py` file.
- Removed `configure_project(package_name)` on session creation.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1254"></a>


kedro/extras/extensions/ipython.py
```diff
@@ -49,6 +51,7 @@ def reload_kedro(path, env: str = None, extra_params: Dict[str, Any] = None):
 
     _remove_cached_modules(metadata.package_name)
 
+    configure_project(metadata.package_name)
```
MerelTheisenQB, line 54: Removed `configure_project` from the session creation, but it's still needed by the ipython extension. So maybe this isn't that much of an improvement..

kedro/framework/project/__init__.py
```diff
@@ -126,12 +126,7 @@ def _load_data(self):
                 self._pipelines_module
             )
         except (ModuleNotFoundError, AttributeError) as exc:
-            # for backwards compatibility with templates < 0.17.2
-            # where no pipelines_registry is defined
-            if self._pipelines_module in str(exc):  # pragma: no cover
-                project_pipelines = {}
-            else:
-                raise
+            raise exc
```
AntonyMilneQB, line 129: I think we can get rid of the whole `try/except/else` here now and just do this:
```
register_pipelines = self._get_pipelines_registry_callable(self._pipelines_module)
project_pipelines = register_pipelines()
```
But check with @lorenabalan since I might have forgotten what we were doing here 😅

tests/framework/project/test_pipeline_registry.py
```diff
@@ -50,6 +46,11 @@ def register_pipelines():
     sys.path.pop(0)
 
 
+def test_pipelines_without_configure_project_throws_exception():
+    with pytest.raises(ModuleNotFoundError, match="No module named 'dummy_package'"):
+        assert pipelines == {}
+
+
```
AntonyMilneQB, line 53: This seems potentially confusing as an error message, though I'm not sure when in reality it might actually occur? Really what's happened here is there hasn't been a call to `configure_project` and hence `pipelines` doesn't exist yet (I think), but a ModuleNotFoundError doesn't give any clue of that or how you might fix it.

What happens now/do we have tests for the following cases?
1. there's no `pipeline_registry.py` file at all, and I do some `kedro` command
2. `pipeline_registry.py` exists but has no `register_pipelines` function
I was kind of expecting to see these in this file already but don't think they're here 🤔 Maybe worth adding them or just manually testing. 

It looks like `_get_pipelines_registry_callable` should already raise a clear error message in these cases but worth double checking since this tripped us up before.

kedro/framework/project/__init__.py
lorenabalan: Yes that's right I was just about to comment that.

tests/framework/project/test_pipeline_registry.py
```diff
@@ -50,6 +46,11 @@ def register_pipelines():
     sys.path.pop(0)
 
 
+def test_pipelines_without_configure_project_throws_exception():
+    with pytest.raises(ModuleNotFoundError, match="No module named 'dummy_package'"):
```
lorenabalan, line 50: Hmm I thought `pipelines` should be an empty dictionary without throwing an error when the project is not configured, i.e. when settings have been loaded. I'm having trouble following but do you know why this wasn't raising before?

kedro/extras/extensions/ipython.py
lorenabalan: I think that's fine, it makes it more explicit that `configure_project` is needed in all 3 main workflows atm: CLI, ipython, package mode.

tests/framework/project/test_pipeline_registry.py
MerelTheisenQB: It's a weird one, and I actually need to spend some more time investigating what's going on. But in short this test passed without changes if only this test file is run, but when run as part of `make test` it fails because of the setup in https://github.com/kedro-org/kedro/blob/main/tests/framework/cli/conftest.py. Several fixtures in that file have `scope=module`. When I change it to be a smaller scope this test passes again, but lots of others fail.

MerelTheisenQB: Thanks for checking @AntonyMilneQB !

AntonyMilneQB (approved):
So much for this being a simple 1 point ticket 😬 Great job getting through it!

Just two small suggestions, not important:
1. If `_ProjectPipelines.configure` is usually called with `pipelines_module = None` then it might be worth giving it a default value of `None`. 
2. Maybe `__init__` should also call `self.configure` just to reduce repetition and make it clear it's doing the same thing.

lorenabalan (approved):
Awesome stuff! 🥇

MerelTheisenQB:
> So much for this being a simple 1 point ticket 😬 Great job getting through it!
> 
> Just two small suggestions, not important:
> 
> 1. If `_ProjectPipelines.configure` is usually called with `pipelines_module = None` then it might be worth giving it a default value of `None` so those calls don't have to specify `None` explicitly (which looks very slightly unnatural to me).
> 2. Maybe `__init__` should also call `self.configure` just to reduce repetition and make it clear it's doing the same thing.

I've addressed point 1, but for point 2 it seems weird to not define the class attributes in `__init__`, and actually my IDE complains about that as well if I make the change.

PR #1260, opened 2022-02-17
[KED-3080] Add support for click==8.x

lorenabalan:
## Description

Closes https://github.com/kedro-org/kedro/issues/1193 

## Development notes

* `kedro-telemetry` also has an explicit dependency on `click<8.0`, along with `kedro~=0.17.3`. This means that if we upgrade `click` in <0.18, users won't be able to install the plugin due to dependency conflict. Created https://github.com/kedro-org/kedro-plugins/pull/5 to fix this.
* Removed deprecated code `get_os_args()` in `kedro.framework.cli.cli` as they will remove that function in 8.1. Used their recommended replacement.
* It was mentioned, during the work on https://github.com/quantumblacklabs/kedro-telemetry/pull/5, that the new `click` version might allow us to more easily create that CLI tree. However I haven't yet found the desired functionality, so I vote it's best we leave it for its own separate task, as it isn't as urgent.
* It seems that the new version of `click` is more diligent in type checking / type casting. 
  ❌  Failing test: `test_run_with_params_in_config`
  Our tests were failing on passing dicts or nested dicts to `--params`, which is what could happen via the config file:
  ```
   kedro run -c config.yml
   ```
   ```yaml
   run:
       pipeline: __default__
       node_names:
           - split
       env: local
       params:
           example_test_data_ratio: 0.3 
   ```
  💻  However, from the CLI users should pass a string that looks like this:
   ```
   --params example_test_data_ratio:0.3,x:5
   ```
   ✍️  In the API we define `--params` as `type=str`, so it was trying to convert the dictionary to a `str` somewhere along the line. 
   💡 Solution inspiration came from https://github.com/pallets/click/issues/1886 and changed the type to `click.UNPROCESSED`. This leaves the value as is, without casting it to a `str`. This looks odd, but in practice it means that users will still be able to declare their extra params as a regular dict in their `config.yml` or `config.json`, or as a `,` and `:` separated string (convention) from CLI.

## Debugging trail

If you're _very_ curious, this was the trail I followed. 
https://github.com/pallets/click/blob/main/src/click/core.py#L2331 
-> https://github.com/pallets/click/blob/55b27155db458ada65a35b9a590956ff256a5df7/src/click/core.py#L2287
-> https://github.com/pallets/click/blob/55b27155db458ada65a35b9a590956ff256a5df7/src/click/core.py#L2246
-> https://github.com/pallets/click/blob/55b27155db458ada65a35b9a590956ff256a5df7/src/click/core.py#L2275
-> https://github.com/pallets/click/blob/55b27155db458ada65a35b9a590956ff256a5df7/src/click/types.py#L212 VS https://github.com/pallets/click/blob/55b27155db458ada65a35b9a590956ff256a5df7/src/click/types.py#L186

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1260"></a>

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -49,9 +49,8 @@ commands:
           name: Install pip setuptools
           command: make install-pip-setuptools
       - run:
-          # Virtualenv 20.0.20 broke pre-commit, capped for now
           name: Install venv for some pre-commit hooks
-          command: conda install -y "virtualenv<20.0"
+          command: conda install -y virtualenv
       - run:
           name: Install requirements and test requirements
           command: pip install --upgrade -r test_requirements.txt
@@ -212,9 +211,21 @@ jobs:
       python_version: <<parameters.python_version>>
     steps:
       - setup
-      - run:
-          name: Run unit tests
-          command: make test
+      - unless:
+          condition:
+            equal: ["3.10", <<parameters.python_version>>]
+          steps:
+            - run:
+                name: Run unit tests in parallel
+                command: make test
+      - when:
+          condition:
+            equal: [ "3.10", <<parameters.python_version>> ]
+          steps:
+            - run:
+                name: Run unit tests sequentially
+                command: pytest tests --cov-config pyproject.toml
+
 
   win_unit_tests:
     parameters:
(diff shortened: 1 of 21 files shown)


lorenabalan:
The fix for the unit tests doesn't actually fix the flow in real life. From manually testing `kedro run -c config.yml` I can see how latest `click` doesn't actually override the parameter value, but `click<8.0` works fine and replaces the parameter value.

```yaml
# config.yml
run:
    pipeline: __default__
    node_names:
        - split
    env: local
    params:
        example_test_data_ratio: 0.3 
```

the logging still says:

```
2022-02-23 13:40:46,712 - root - INFO - example_test_data_ratio: 0.200000
```

lorenabalan:
> The fix for the unit tests doesn't actually fix the flow in real life. 

Ignore this, it was a problem with the virtual environment, not with the tests. In an env where I pip-installed local kedro `pip install .`, the overriding works fine.

AntonyMilneQB (approved):
Very nice work and excellent write-up! I can definitely appreciate that this was much harder than the size of the diff makes it look ⭐

MerelTheisenQB (approved):
Great work @lorenabalan! ⭐  The solution does indeed seem simple, but from the description it's clear this wasn't trivial.

Issue #1040, opened 2021-11-15
[KED-3023] Enable plugins to extend starter-aliases default list  


Galileo-Galilei:
## Description

> Is your feature request related to a problem? A clear and concise description of what the problem is: "I'm always frustrated when ..."

We have at work a few kedro starters templates which enable to create kedro projects with a bunch of internal configuration (pre-commit, gitlab-ci, custom ConfigLoader...). We also have a custom plugin which modifies kedro's internal (especially the CONF_ROOT constant and the CLI). 

When we distribute our template insisde the organisation, we have to give the entire VCS path to the project to our user, and it is not very convenient. I wish we were able to decalre a new starter-alias in the plugin, and when the plugin is installed and auto-discovered by kedro, users can user: 

``kedro new --starter=my_starter`` where ``my_starter`` VSC path was declared in the plugin, maybe like this: 

```python: 
_PLUGIN_STARTER_ALIAS: ​{my_starter: path/to/VCS}
```

## Context

>Why is this change important to you? How would you use it? How can it benefit other users?

The 2 main benefits I expect are: 
- it facilitates the distribution of templates, since it makes installation easier for end users
- it facilitates maintenance: if the template path changes for whatever reason (regrouping all the templates in a single repo, moving to another gitlab/github instances for orgnaisations internal reasons), the modification is transparent for end users

## Possible Implementation

Add a ``_PLUGIN_STARTER_ALIAS`` constant in plugins which will be user to extend the https://github.com/quantumblacklabs/kedro/blob/bce62c824a4786e7cf9cc355b31840879ec851bf/kedro/framework/cli/starters.py#L31-L38 set. 

It likely needs a bit more refactoring to make it a ``dict`` [accepting different _STARTERS_REPO path](https://github.com/quantumblacklabs/kedro/blob/bce62c824a4786e7cf9cc355b31840879ec851bf/kedro/framework/cli/starters.py#L39) which is currently a hardcoded path to a single one github repo.

datajoely:
I like this idea! Let me bring it up with the team

AntonyMilneQB:
I also like this idea but am a little bit unsure about implementation. Thinking out loud here...

[...]

PR #1265, opened 2022-02-19
POC [KED-3023] - Enable plugins to extend starter-aliases default list

Galileo-Galilei:
## Description

This PR implements part of the suggestions described in #1040 and creates a fully functional examples on how kedro starter discovery could be modified to enable plugins to created aliased starters. 

## Development notes

- I modified the _STARTERS_ALIASES to make it a dict with the format discussed in above issue.
- I created a ``kedro.starter`` entrypoint which can is iterated over to update the _STARTERS_ALIASES constant. 
- I  updated the ``kedro new`` and ``kedro starter list`` commands to use the updated _STARTERS_ALIASES constant (with new aliases discovered in the entrypoints). 

As of now, the code is already working. You can test it with : 

```bash
conda create -n fps python=3.8 -y
conda activate fps
pip install git+https://github.com/Galileo-Galilei/kedro.git@plugins-starters
pip install git+https://github.com/Galileo-Galilei/fake-plugin-starter.git
kedro starter list
```

and see the following output:

```bash
(*) my_fake_starter: http://google.com # my extra starter
astro-airflow-iris: https://github.com/kedro-org/kedro-starters/tree/main/astro-airflow-iris
astro-iris: https://github.com/kedro-org/kedro-starters/tree/main/astro-airflow-iris
mini-kedro: https://github.com/kedro-org/kedro-starters/tree/main/mini-kedro
pandas-iris: https://github.com/kedro-org/kedro-starters/tree/main/pandas-iris
pyspark: https://github.com/kedro-org/kedro-starters/tree/main/pyspark
pyspark-iris: https://github.com/kedro-org/kedro-starters/tree/main/pyspark-iris
spaceflights: https://github.com/kedro-org/kedro-starters/tree/main/spaceflights
```

You can **have a look at the [plugin code](https://github.com/Galileo-Galilei/fake-plugin-starter.git) if needed**.

**Some questions are still to be answered:** 
- is this a desirable implementation?
- should we add extra conversion / parsing when loading the entrypoint to check the format of the constant (I think we should, this is likely error prone right now)
- should we enable plugin authors to override "official" starters? My current code prevents this but it may be acceptable.
- should we make a visual distinction when listing the starters ? i add an extra (*) at the beginning to ensure they are listed first and identifiable at first glance.

## Checklist

As of now, **I did not add tests and documentation: I'll update them once you will confirm these changes are desirable**

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1265"></a>

### DIFF
--- a/kedro/framework/cli/starters.py
+++ b/kedro/framework/cli/starters.py
@@ -24,7 +24,8 @@ from kedro.framework.cli.utils import (
     _clean_pycache,
     _filter_deprecation_warnings,
     command_with_verbosity,
-    load_entry_points,
+    # load_entry_points,
+    ENTRY_POINT_GROUPS,
 )
 
 KEDRO_PATH = Path(kedro.__file__).parent
@@ -36,16 +37,41 @@ _STARTERS_REPO = "git+https://github.com/kedro-org/kedro-starters.git"
 # and tutorials still refer to `astro-iris`. We create an alias to check if a user has
 # entered old `astro-iris` as the starter name and changes it to `astro-airflow-iris`.
 _STARTER_ALIASES = {
-    "astro-airflow-iris": {"repo": _STARTERS_REPO, "directory": "astro-airflow-iris"},
-    "astro-iris": {
-        "repo": _STARTERS_REPO,
+    "astro-airflow-iris": {
+        "template_path": _STARTERS_REPO,
         "directory": "astro-airflow-iris",
-    },  # this is an alias name
-    "mini-kedro": {"repo": _STARTERS_REPO, "directory": "mini-kedro"},
-    "pandas-iris": {"repo": _STARTERS_REPO, "directory": "pandas-iris"},
-    "pyspark": {"repo": _STARTERS_REPO, "directory": "pyspark"},
-    "pyspark-iris": {"repo": _STARTERS_REPO, "directory": "pyspark-iris"},
-    "spaceflights": {"repo": _STARTERS_REPO, "directory": "spaceflights"},
+        "checkout": "main",
+    },
+    "astro-iris": {
+        "template_path": _STARTERS_REPO,
+        "directory": "astro-airflow-iris",
+        "checkout": "main",
+    },  # this is an alias name for "astro-airflow-iris"
+    "mini-kedro": {
+        "template_path": _STARTERS_REPO,
+        "directory": "mini-kedro",
+        "checkout": "main",
+    },
+    "pandas-iris": {
+        "template_path": _STARTERS_REPO,
+        "directory": "pandas-iris",
+        "checkout": "main",
+    },
+    "pyspark": {
+        "template_path": _STARTERS_REPO,
+        "directory": "pyspark",
+        "checkout": "main",
+    },
+    "pyspark-iris": {
+        "template_path": _STARTERS_REPO,
+        "directory": "pyspark-iris",
+        "checkout": "main",
+    },
+    "spaceflights": {
+        "template_path": _STARTERS_REPO,
+        "directory": "spaceflights",
+        "checkout": "main",
+    },
 }
 
 CONFIG_ARG_HELP = """Non-interactive mode, using a configuration yaml file. This file
@@ -75,29 +101,36 @@ def _remove_readonly(func: Callable, path: Path, excinfo: Tuple):  # pragma: no 
 def _get_starters_aliases() -> Dict[str, Dict[str, str]]:
     """This functions lists all the starters aliases declared in
     the core repo and in plugins entry points.
+
+    The output looks like:
+    {
+        {
+            "astro-airflow-iris": {"template_path": ..., directory=..., "origin": "kedro"},
+            ...,
+            "my-awesome-starter": {"template_path": ..., directory=..., "origin": "my-awesome-plugin"}
+        }
+    }
     """
-    official_starter_aliases = _STARTER_ALIASES.copy()
-    starters_aliases_from_plugins = load_entry_points("starter")
 
-    # TODO: it would likely be a good idea to indicate in the name
-    # from which plugin this is imported, e.g  we may store the origin
-    # which would return a dict like {pandas-iris: kedro, my_awesome-starter: awesome_plugin}
-    # this is useful for debugging and when listing discovered starters
-
-    # TODO Add the possibility for plugins users to specify {starter_name: git+url}
-    #  instead of {starter_name: {"repo": git+url, "directory": None}}
-    # which is unneedly verbose and error prone
-
-    starters_aliases_from_plugins = {
-        k: v
-        for starter_alias in starters_aliases_from_plugins
-        for k, v in starter_alias.items()
+    # add an extra key to indicate from where the plugin come from
+    starters_aliases = {
+        name: {**config, "origin": "kedro"} for name, config in _STARTER_ALIASES.items()
     }
 
-    # TODO discuss if we should let the possibility to override official starters
-    # i.e. return {**official_starter_aliases, **starters_aliases_from_plugins}.
-    # I don't for now, but maybe it's a good idea?
-    return {**starters_aliases_from_plugins, **official_starter_aliases}
+    for starter_entry_point in pkg_resources.iter_entry_points(
+        group=ENTRY_POINT_GROUPS["starters"]
+    ):
+        module_name = starter_entry_point.module_name.split(".")[0]
+        for starter_name, starter_config in starter_entry_point.load().items():
+            if starter_name in starters_aliases:
+                click.secho(
+                    f"The starter alias {starter_name} from {module_name} was already created by from {starters_aliases[starter_name]['origin']} and is ignored",
+                    fg="yellow",
+                )
+                continue  # break the loop
+            starters_aliases[starter_name] = {**starter_config, "origin": module_name}
+
+    return starters_aliases
 
 
 # pylint: disable=missing-function-docstring
@@ -131,18 +164,18 @@ def new(
 
     starter_aliases = _get_starters_aliases()
 
-    if starter_name in starter_aliases.keys():
+    if starter_name in starter_aliases:
         if directory:
             raise KedroCliError(
                 "Cannot use the --directory flag with a --starter alias."
             )
-        template_path = starter_aliases[starter_name]["repo"]
+        template_path = starter_aliases[starter_name]["template_path"]
         # "directory" is an optional key for starters from plugins, so if the key is not present we will use "None".
         directory = starter_aliases[starter_name].get("directory")
-        checkout = checkout or version
+        checkout = checkout or starter_aliases[starter_name].get("checkout", version)
     elif starter_name is not None:
         template_path = starter_name
-        checkout = checkout or version
+        checkout = checkout or starter_aliases[starter_name].get("checkout", version)
     else:
         template_path = str(TEMPLATE_PATH)
 
@@ -185,26 +218,31 @@ def starter():
 def list_starters():
     """List all official project starters available."""
     starters_aliases = _get_starters_aliases()
-
-    output = {}
+    # we "reverse" the dictionnary so that the origin key in starter config
+    # {name: {template_path; ..., origin: ...}} became the first key:
+    # {kedro: ["astro-iris":{...}], "my-plugin":["my-starter":{...}]}
+    starters_by_origin = {}
     for starter_name, starter_config in starters_aliases.items():
-        directory = starter_config.get("directory")
-        repo_url = starter_config.get("repo")
-        repo_url = repo_url.replace("git+", "")
-        if directory:
-            repo_url = repo_url.replace(".git", f"/tree/main/{directory}")
-        # TODO: should we make "core" starters distinguishable from plugin ones?
-        # I add a "(*)" next to the name to make this distinction visual
-        # and at the beginning of the list
-        # BEWARE: in case we allow official plugins to be overriden, below test is not valid
-        starter_name_displayed = (
-            f"(*) {starter_name}"
-            if not starter_name in _STARTER_ALIASES.keys()
-            else starter_name
-        )
-        output[starter_name_displayed] = repo_url
-
-    click.echo(yaml.safe_dump(output))
+        starter_raw_config = {
+            starter_name: {k: v for k, v in starter_config.items() if k != "origin"}
+        }
+        if starter_config["origin"] in starters_by_origin:
+            starters_by_origin[starter_config["origin"]].append(starter_raw_config)
+        else:
+            starters_by_origin[starter_config["origin"]] = [starter_raw_config]
+    for origin, starter_aliases in starters_by_origin.items():
+        click.secho(f"Starters from '{origin}'\n", fg="yellow")
+        output = []
+        for starter_alias in starter_aliases:
+            for starter_name, starter_config in starter_alias.items():
+                starter_name_displayed = (
+                    f"{starter_name} "
+                    f"(template_path={starter_config.get('template_path')}, "
+                    f"directory={starter_config.get('directory')}, "
+                    f"checkout={starter_config.get('checkout')})"
+                )
+                output.append(starter_name_displayed)
+        click.echo(yaml.safe_dump(output))
 
 
 def _fetch_config_from_file(config_path: str) -> Dict[str, str]:
--- a/kedro/framework/cli/starters.py
+++ b/kedro/framework/cli/starters.py
@@ -3,12 +3,15 @@
 This module implements commands available from the kedro CLI for creating
 projects.
 """
+from itertools import groupby
+from operator import itemgetter
 import os
 import re
 import shutil
 import stat
 import tempfile
 from collections import OrderedDict
+from pandas import value_counts
 import pkg_resources
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Tuple
@@ -36,19 +39,36 @@ _STARTERS_REPO = "git+https://github.com/kedro-org/kedro-starters.git"
 # The `astro-iris` was renamed to `astro-airflow-iris`, but old (external) documentation
 # and tutorials still refer to `astro-iris`. We create an alias to check if a user has
 # entered old `astro-iris` as the starter name and changes it to `astro-airflow-iris`.
-_STARTER_ALIASES = {
-    "astro-airflow-iris": {
+_STARTERS_ALIASES = [
+    {
+        "name": "astro-airflow-iris",
         "template_path": _STARTERS_REPO,
         "directory": "astro-airflow-iris",
     },
     # this is an alias name for "astro-airflow-iris"
-    "astro-iris": {"template_path": _STARTERS_REPO, "directory": "astro-airflow-iris"},
-    "mini-kedro": {"template_path": _STARTERS_REPO, "directory": "mini-kedro"},
-    "pandas-iris": {"template_path": _STARTERS_REPO, "directory": "pandas-iris"},
-    "pyspark": {"template_path": _STARTERS_REPO, "directory": "pyspark"},
-    "pyspark-iris": {"template_path": _STARTERS_REPO, "directory": "pyspark-iris"},
-    "spaceflights": {"template_path": _STARTERS_REPO, "directory": "spaceflights"},
-}
+    {
+        "name": "astro-iris",
+        "template_path": _STARTERS_REPO,
+        "directory": "astro-airflow-iris",
+    },
+    {"name": "mini-kedro", "template_path": _STARTERS_REPO, "directory": "mini-kedro"},
+    {
+        "name": "pandas-iris",
+        "template_path": _STARTERS_REPO,
+        "directory": "pandas-iris",
+    },
+    {"name": "pyspark", "template_path": _STARTERS_REPO, "directory": "pyspark"},
+    {
+        "name": "pyspark-iris",
+        "template_path": _STARTERS_REPO,
+        "directory": "pyspark-iris",
+    },
+    {
+        "name": "spaceflights",
+        "template_path": _STARTERS_REPO,
+        "directory": "spaceflights",
+    },
+]
 
 CONFIG_ARG_HELP = """Non-interactive mode, using a configuration yaml file. This file
 must supply  the keys required by the template's prompts.yml. When not using a starter,
@@ -74,6 +94,27 @@ def _remove_readonly(func: Callable, path: Path, excinfo: Tuple):  # pragma: no 
     func(path)
 
 
+def _check_starter_entrypoint_config(module_name: str, config: Dict[str, str]) -> None:
+    if not isinstance(config, dict):
+        raise ValueError(
+            f"The starter configuration loaded from entrypoint should be a 'dict', got '{config}' instead"
+        )
+    mandatory_keys = {"name", "template_path"}
+    optional_keys = {"directory"}
+
+    missing_keys = mandatory_keys - set(config.keys())
+    if missing_keys:  # mandatory keys
+        raise ValueError(
+            f"Entrypoint kedro.starters from {module_name} must have the following missing keys: '{missing_keys}'"
+        )
+
+    extra_keys = set(config.keys()) - mandatory_keys - optional_keys  # optional keys
+    if extra_keys:  # mandatory keys
+        raise ValueError(
+            f"Entrypoint kedro.starters from {module_name} has keys '{extra_keys}' which are not allowed"
+        )
+
+
 def _get_starters_aliases() -> Dict[str, Dict[str, str]]:
     """This functions lists all the starters aliases declared in
     the core repo and in plugins entry points.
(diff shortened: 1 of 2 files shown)


kedro/framework/cli/starters.py
```diff
@@ -23,20 +24,29 @@
...
+    "astro-airflow-iris": {"repo": _STARTERS_REPO, "directory": "astro-airflow-iris"},
+    "astro-iris": {
+        "repo": _STARTERS_REPO,
+        "directory": "astro-airflow-iris",
+    },  # this is an alias name
+    "mini-kedro": {"repo": _STARTERS_REPO, "directory": "mini-kedro"},
+    "pandas-iris": {"repo": _STARTERS_REPO, "directory": "pandas-iris"},
+    "pyspark": {"repo": _STARTERS_REPO, "directory": "pyspark"},
+    "pyspark-iris": {"repo": _STARTERS_REPO, "directory": "pyspark-iris"},
+    "spaceflights": {"repo": _STARTERS_REPO, "directory": "spaceflights"},
```
AntonyMilneQB, line 48: Let's make this `template_path` so it matches the arguments given in `new`. This also keeps it generic.

kedro/framework/cli/starters.py
```diff
@@ -91,19 +129,16 @@ def new(
...
-    # The `astro-iris` was renamed to `astro-airflow-iris`, but old (external) documentation
-    # and tutorials still refer to `astro-iris`. The below line checks if a user has entered old
-    # `astro-iris` as the starter name and changes it to `astro-airflow-iris`.
-    starter_name = (
-        "astro-airflow-iris" if starter_name == "astro-iris" else starter_name
-    )
-    if starter_name in _STARTER_ALIASES:
+    starter_aliases = _get_starters_aliases()
+
+    if starter_name in starter_aliases.keys():
```
AntonyMilneQB, line 134:
```suggestion
    if starter_name in starter_aliases:
```

kedro/framework/cli/starters.py
```diff
@@ -91,19 +129,16 @@ def new(
...
         if directory:
             raise KedroCliError(
                 "Cannot use the --directory flag with a --starter alias."
             )
-        template_path = _STARTERS_REPO
-        directory = starter_name
+        template_path = starter_aliases[starter_name]["repo"]
+        # "directory" is an optional key for starters from plugins, so if the key is not present we will use "None".
+        directory = starter_aliases[starter_name].get("directory")
         checkout = checkout or version
```
AntonyMilneQB, line 142: I think we should be able to specify `checkout` in `_STARTER_ALIASES` also. The order of precedence should be:
1. command line `checkout` argument
2. `checkout` given in `_STARTER_ALIASES`
3. `version`

kedro/framework/cli/starters.py
```diff
@@ -62,6 +72,34 @@ def _remove_readonly(func: Callable, path: Path, excinfo: Tuple):  # pragma: no
...
+    starters_aliases_from_plugins = load_entry_points("starter")
+
+    # TODO: it would likely be a good idea to indicate in the name
+    # from which plugin this is imported, e.g  we may store the origin
+    # which would return a dict like {pandas-iris: kedro, my_awesome-starter: awesome_plugin}
+    # this is useful for debugging and when listing discovered starters
+
+    # TODO Add the possibility for plugins users to specify {starter_name: git+url}
+    #  instead of {starter_name: {"repo": git+url, "directory": None}}
+    # which is unneedly verbose and error prone
```
AntonyMilneQB, line 89: Agree the user shouldn't need to specify `directory` (it should default to `None`) but not sure about shortcutting out the nested `repo` (or `template_path`) key. It's more verbose but I think clearer than introducing a new shortcut syntax.

kedro/framework/cli/starters.py
```diff
@@ -62,6 +72,34 @@ def _remove_readonly(func: Callable, path: Path, excinfo: Tuple):  # pragma: no
...
+
+    starters_aliases_from_plugins = {
+        k: v
+        for starter_alias in starters_aliases_from_plugins
+        for k, v in starter_alias.items()
+    }
```
AntonyMilneQB, line 95: What happens if there are multiple plugins that use this entry point? Will they all be there? Will they overwrite each other if `starter_name` is repeated? 🤔

kedro/framework/cli/starters.py
```diff
@@ -149,10 +184,26 @@ def starter():
...
-    ]
+    starters_aliases = _get_starters_aliases()
+
+    output = {}
+    for starter_name, starter_config in starters_aliases.items():
+        directory = starter_config.get("directory")
+        repo_url = starter_config.get("repo")
+        repo_url = repo_url.replace("git+", "")
+        if directory:
+            repo_url = repo_url.replace(".git", f"/tree/main/{directory}")
```
AntonyMilneQB, line 195: Again, ideally I'd make this generic to work for things outside git.

kedro/framework/cli/utils.py
```diff
@@ -31,6 +31,7 @@
     "line_magic": "kedro.line_magic",
     "hooks": "kedro.hooks",
     "cli_hooks": "kedro.cli_hooks",
+    "starter": "kedro.starter"
```
AntonyMilneQB, line 34:
```suggestion
    "starters": "kedro.starters"
```
Just seems more natural to me (like `hooks`).

Not sure whether this should go inside some hierarchy like `kedro.constants.` as you suggested before (but then what else might go in there?) or whether it belongs at the top level 🤔

AntonyMilneQB (commented):
Hello @Galileo-Galilei! Thanks very much for your work on this. In general I like this a lot and think that using entrypoints is a good solution - I'm just not sure on some of the details. 

My main comment is that, even though different git repos is the most obvious use of this, it's conceivable that I might want to share a starter by some other means (a shared filepath say). `kedro new --starter` is very flexible and can things like filepaths rather than just git repos, so I think we should make sure that extending starters is equally as generic. Similarly I might want to share two different starters that point to exactly the same place but just have different tags, i.e. different `checkout` arguments. I think we should allow this too just to keep things generic. So my suggested format for the dictionary would be:
```
_STARTER_ALIASES = {"name": {"template_path": ..., "directory": ..., "checkout": ...}}
```
where `directory` and `checkout` are optional (`directory` defaults to `None`).

As for your other questions...
> should we add extra conversion / parsing when loading the entrypoint to check the format of the constant (I think we should, this is likely error prone right now)

No strong feelings here - we don't check other entrypoints very carefully I think, and this is quite an advanced feature where there will just be one simple pattern to copy and paste from a documented example. So I'm not too worried about verifying the input.

> should we enable plugin authors to override "official" starters? My current code prevents this but it may be acceptable.

I think best to leave the official ones unaltered - see next answer for why.

> should we make a visual distinction when listing the starters ? i add an extra (*) at the beginning to ensure they are listed first and identifiable at first glance.

No strong feelings, but my suggestion would be to separate the output into sections according to the plugin source. Like how we do for global/project commands if you just type `kedro`: 
image
This would work nicely if you have starters coming from multiple plugins (assuming that's possible).

Galileo-Galilei:
Hi @AntonyMilneQB, thanks for the review. The more I think about it, the more I think the entrypoint solution is "pythonic" indeed. I even found [a reference](https://amir.rachum.com/blog/2017/07/28/python-entry-points/) doing something similar.

> Similarly I might want to share two different starters that point to exactly the same place but just have different tags, i.e. different checkout arguments

I was wondering if setting the ``checkout`` argument in the template is really something we want. You o not lety this possibility currently, and it may be better to point to ``main/master`` by default and make overriding explicit? I do think that it may be confusing for users to just do kedro new --starter=my_awesome_starter and not have the last version installed in case ``checkout`` argument is set to a fixed reference. Specifiying checkout at instantiation time is not enough for you? What would be the use case? 

> No strong feelings, but my suggestion would be to separate the output into sections according to the plugin source. Like how we do for global/project commands if you just type kedro

I actually had the very same idea at first, but I don't know how to make it work honestly :sweat_smile:  I want to reuse your ``load_entry_points`` function and it just returns a list of the entrypoints, I don't know how I can know from which plugin they come from.

kedro/framework/cli/starters.py
Galileo-Galilei: Done!

kedro/framework/cli/starters.py
Galileo-Galilei: Done

AntonyMilneQB:
> I was wondering if setting the `checkout` argument in the template is really something we want. You o not lety this possibility currently, and it may be better to point to `main/master` by default and make overriding explicit? I do think that it may be confusing for users to just do kedro new --starter=my_awesome_starter and not have the last version installed in case `checkout` argument is set to a fixed reference. Specifiying checkout at instantiation time is not enough for you? What would be the use case?

Fair enough. I was thinking that `checkout` can be used more than just to specify a version of a starter - in theory, someone might want to set up their starters so that there are two different starters defined in the same repo but labelled by their branch (instead of by directory). But I'd forgotten that by default we expect starters to use `checkout` to correspond to kedro version (rather than `main/master`), so this scheme for setting up a starters repo doesn't make much sense after all I think. Happy to go with what you think is best here 🙂 

> I actually had the very same idea at first, but I don't know how to make it work honestly 😅 I want to reuse your `load_entry_points` function and it just returns a list of the entrypoints, I don't know how I can know from which plugin they come from.

Ah yes I see, that is annoying. The [type hint with `click.MultiCommand`](https://github.com/Galileo-Galilei/kedro/blob/main/kedro/framework/cli/utils.py#L335) (and exception message) indicates to me that really the `load_entry_points` function is intended for use with loading CLI commands (even though we also use it for `line_magic` it seems). Given that extracting the package name seems quite simple ([like this for example](https://github.com/kedro-org/kedro/blob/main/kedro/framework/cli/cli.py#L72-L77)) I'd probably just forget about using `load_entry_points` in favour of writing your own thing that does exactly what you want here.

kedro/framework/cli/starters.py
Galileo-Galilei: I've done it, even if I am not sure we should let checkout be part of the config (see below)

kedro/framework/cli/starters.py
Galileo-Galilei: I agree for not shortcutting, it would add extra maitneane burden for very little better user experience

kedro/framework/cli/starters.py
Galileo-Galilei: I've added a warning  which will print: the first one will be registered, and other will not with a warning message

kedro/framework/cli/starters.py
Galileo-Galilei: This is your original implementation (with only your starters) and I did not dare change it, but I agree. I've changed it to make it generic

kedro/framework/cli/utils.py
Galileo-Galilei: I hesitated for the "s", let's go for the suggestions. I know we originally discussed about calling it "constants", but I make very strong assumptions on the format when looping on "starters" entrypoint, so This is clearly not generic enough to accept other constants (+ I do not see any other use case as you mention)

Galileo-Galilei:
I've just managed to make it work but the code is poorly readable. I do not use click.MultiCommand and brutally parse the text which is not very satisfying. The output looks like this: 

image

kedro/framework/cli/utils.py
AntonyMilneQB: Makes sense. I'm happy to have it as a top-level entry point `kedro.starters`.

kedro/framework/cli/starters.py
```diff
@@ -62,6 +98,41 @@ def _remove_readonly(func: Callable, path: Path, excinfo: Tuple):  # pragma: no
...
+    }
+
+    for starter_entry_point in pkg_resources.iter_entry_points(
+        group=ENTRY_POINT_GROUPS["starters"]
+    ):
+        module_name = starter_entry_point.module_name.split(".")[0]
+        for starter_name, starter_config in starter_entry_point.load().items():
+            if starter_name in starters_aliases:
+                click.secho(
+                    f"The starter alias {starter_name} from {module_name} was already created by from {starters_aliases[starter_name]['origin']} and is ignored",
```
AntonyMilneQB, line 127:
```suggestion
                    f"Starter alias `{starter_name}` from `{module_name}` has been ignored as it is already defined by `{starters_aliases[starter_name]['origin']}`",
```

kedro/framework/cli/starters.py
```diff
@@ -91,23 +162,20 @@ def new(
...
-        directory = starter_name
-        checkout = checkout or version
+        template_path = starter_aliases[starter_name]["template_path"]
+        # "directory" is an optional key for starters from plugins, so if the key is not present we will use "None".
+        directory = starter_aliases[starter_name].get("directory")
+        checkout = checkout or starter_aliases[starter_name].get("checkout", version)
     elif starter_name is not None:
         template_path = starter_name
-        checkout = checkout or version
+        checkout = checkout or starter_aliases[starter_name].get("checkout", version)
```
AntonyMilneQB, line 178:
```suggestion
        checkout = checkout or version
```
Whatever we decide to do with `checkout`, this part of the conditional should stay the same because it applies when `starter_name` isn't in `starter_aliases`.

kedro/framework/cli/starters.py
```diff
@@ -149,11 +217,32 @@ def starter():
...
+        for starter_alias in starter_aliases:
+            for starter_name, starter_config in starter_alias.items():
+                starter_name_displayed = (
+                    f"{starter_name} "
+                    f"(template_path={starter_config.get('template_path')}, "
+                    f"directory={starter_config.get('directory')}, "
+                    f"checkout={starter_config.get('checkout')})"
+                )
+                output.append(starter_name_displayed)
+        click.echo(yaml.safe_dump(output))
```
AntonyMilneQB, line 245: There's a few possible simplifications here. I think we can [rely on Python dictionary ordering](https://stackoverflow.com/questions/1867861/how-to-keep-keys-values-in-same-order-as-declared), so you can do it without needing to restructure the dictionary at all:
```
    last_origin = "kedro"
    click.secho(f"Built-in starters\n", fg="yellow")

    for starter_name, starter_config in starters_aliases.items():
        origin = starter_config.pop("origin")
        if origin != last_origin:
            click.secho(f"\nStarters from {origin}\n", fg="yellow")
            last_origin = origin
        click.echo(f"{starter_name}: {starter_config}")
```
If you do want to restructure the dictionary then it's probably possible using something like [this](https://www.geeksforgeeks.org/group-list-of-dictionary-data-by-particular-key-in-python/).

The above renders as follows:
image

... which is not super beautiful I will admit 😅 Could probably be made nicer by using `yaml.dump(default_flow_style=True)` instead of just printing the raw dictionary. But I do like only giving `directory` and `checkout` when they're explicitly defined.

AntonyMilneQB (commented):
Thanks for the changes! I left a couple more suggestions but this is looking really good now! Let's try and get a second pair of eyes on it before it becomes too dominated just by my opinion.

> Yes that's what I figured out in the code. It makes sense for kedro itself to have a default to the starter matching the kedro version, but i don't think it really does for user defined starters. I implemented it however to match your suggestion.

Yeah, this is something I've wondered before also. Just for the purposes of this PR, I think we should leave the current behaviour as it is (i.e. by default `checkout` = kedro version, not `main`) and save possible changes to that for a separate discussion.

Then the only question in my mind then is whether plugin-defined starters should be allowed to define `checkout` themselves or not. Not a big deal either way - I'm happy just to defer to you on it or see what the third opinion says.

kedro/framework/cli/starters.py
```diff
@@ -91,23 +162,20 @@ def new(
...
             raise KedroCliError(
                 "Cannot use the --directory flag with a --starter alias."
             )
-        template_path = _STARTERS_REPO
-        directory = starter_name
-        checkout = checkout or version
+        template_path = starter_aliases[starter_name]["template_path"]
+        # "directory" is an optional key for starters from plugins, so if the key is not present we will use "None".
+        directory = starter_aliases[starter_name].get("directory")
+        checkout = checkout or starter_aliases[starter_name].get("checkout", version)
```
lorenabalan, line 175: I agree with @AntonyMilneQB that we leave `checkout` out of the scope for now, and assume the user-defined starters follow the same tagging/branching convention (default to sth that matches the kedro version). This is currently what happens anyway for custom starters, and I think it's a more complicated discussion to have than we could do justice for here.

kedro/framework/cli/starters.py
```diff
@@ -149,11 +217,32 @@ def starter():
...
-    click.echo(yaml.safe_dump(output))
+    starters_aliases = _get_starters_aliases()
+    # we "reverse" the dictionnary so that the origin key in starter config
+    # {name: {template_path; ..., origin: ...}} became the first key:
+    # {kedro: ["astro-iris":{...}], "my-plugin":["my-starter":{...}]}
+    starters_by_origin = {}
+    for starter_name, starter_config in starters_aliases.items():
+        starter_raw_config = {
+            starter_name: {k: v for k, v in starter_config.items() if k != "origin"}
+        }
```
lorenabalan, line 228: You can create a `deepcopy` of `starters_aliases` and use that instead - that way it's safe to use `.pop()` as well, as @AntonyMilneQB suggested.

lorenabalan (commented):
Thank you @Galileo-Galilei for taking a stab at this! 🙌 I like the general direction of this. :) My main observation is that I'd like us to limit the scope to just "auto-discovery of starters from plugins", and not worry about custom `checkout` and `directory` conventions. If we change anything around there it will automatically make this PR a breaking change. 

I believe we shouldn't allow natively supported starter aliases to be overwritten, that can cause _a lot_ of confusion. Also some minimal checks on stuff returned from the entrypoint would be nice. We don't have checks on the others because hooks and CLI commands are harder to validate than a plain dictionary. If a user-defined plugin defined a bad starter entrypoint, we should wrap it up in a nice exception message.

Would love to also see some documentation added before merging.

kedro/framework/cli/starters.py
Galileo-Galilei: As discussed below, I completely remove the "checkout" option of this PR.

kedro/framework/cli/utils.py
Galileo-Galilei: Oh, and I've just remembered why I choose to call it "starter" instead of "starters": it is the name tof the CLI group command (``kedro starter list``) so I wanted to be consistent. What do you want to do? Keep the "s" or not?

kedro/framework/cli/starters.py
```diff
@@ -62,6 +98,41 @@ def _remove_readonly(func: Callable, path: Path, excinfo: Tuple):  # pragma: no
...
+    ):
+        module_name = starter_entry_point.module_name.split(".")[0]
+        for starter_name, starter_config in starter_entry_point.load().items():
+            if starter_name in starters_aliases:
+                click.secho(
+                    f"The starter alias {starter_name} from {module_name} was already created by from {starters_aliases[starter_name]['origin']} and is ignored",
+                    fg="yellow",
+                )
+                continue  # break the loop
+            starters_aliases[starter_name] = {**starter_config, "origin": module_name}
```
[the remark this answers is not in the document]
Galileo-Galilei: Yes, that's much better

kedro/framework/cli/starters.py
Galileo-Galilei: Yes, sure. I did not add checkout in the first version and it really seems that it needs to be discussed further.

kedro/framework/cli/starters.py
Galileo-Galilei: @AntonyMilneQB I thought dictionnary ordering was only from ``python>=3.7`` and forward and I exclitly avoid relying on dict ordering because I know you still support ``python=3.6`` :) But I was wrong so glad to make it more readable.

@ lorenabalan I think `pop()` should be fine because the ``_get_starter_aliases()`` is called at the beginning of the function each time, and it does itself copy the _STARTERS_ALIASES constant. It would be safer to explictly deepcopy the dict in case of future changes though, do you prefer to enforce it?

I am still experimenting to see if we can have something more readable.

kedro/framework/cli/starters.py
```diff
@@ -149,11 +217,32 @@ def starter():
...
+        output = []
+        for starter_alias in starter_aliases:
+            for starter_name, starter_config in starter_alias.items():
+                starter_name_displayed = (
+                    f"{starter_name} "
+                    f"(template_path={starter_config.get('template_path')}, "
+                    f"directory={starter_config.get('directory')}, "
+                    f"checkout={starter_config.get('checkout')})"
+                )
+                output.append(starter_name_displayed)
```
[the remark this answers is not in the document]
Galileo-Galilei: I've made another attempt with ``yaml.dump``. i am a bit worried about readibilty though, since it displayed inforamtion on several lines for each starter and there are a lot of them, so it may hurt discoverability. Current implementation looks like this: 

image
 I can reorder the key (to have "name" as the first key) a d remove unused key (``origin`` or ``directory`` when it is ``None``), but I still find it harder to read than above suggestion. What do you think?

kedro/framework/cli/starters.py
Galileo-Galilei: It is still safe here since _get_starters_aliases recreate the dict when it is called and already deepcopy _STARTERS_ALAIASES.

Galileo-Galilei:
Hi @lorenabalan and @AntonyMilneQB, I've made a couple of changes sine last review: 

-  I refactored the "configuration" dictionnary one should specify in an entrypoint from this: 
```python 
    {"astro-airflow-iris": {
        "template_path": _STARTERS_REPO,
        "directory": "astro-airflow-iris",
    }, ...}
```

to this

```python 
[
    {
        "name": "astro-airflow-iris",
        "template_path": _STARTERS_REPO,
        "directory": "astro-airflow-iris",
    }, ...]
```

Avoiding nested dictionnary make manipulations easier (reordering by "origin" or "name" depending on what is needed] is easier this way, and I think this makes code more readable in general.

I've also tried to output result with ``yaml. dump`` but it is likely too verbose and less readable. I'd likely suggest to go with Antony first suggestion (e.g my initial "custom" formatting with ``directory`` key present only when filled)

kedro/framework/cli/utils.py
AntonyMilneQB: Ah makes sense. Personally I'd go for `starters` since I wouldn't be too surprised if `kedro starter` at some point in the future somehow gets subsumed into `kedro new`, and the inconsistency here makes grammatical sense to me. But let's see what someone else thinks too - @lorenabalan?

kedro/framework/cli/starters.py
```diff
@@ -149,11 +241,30 @@ def starter():
...
+    }
+
+    # ensure kedro starters are listed first
+    kedro_starters = starters_aliases_by_origin.pop("kedro")
+    click.secho(f"\nBuilt-in starters\n", fg="yellow")
+    click.echo(yaml.dump(kedro_starters))
+
+    for origin, starter_config in starters_aliases_by_origin.items():
+        click.secho(f"\nStarters from {origin}\n", fg="yellow")
+        click.echo(yaml.dump(starter_config))
```
AntonyMilneQB, line 267: After spending WAAAY too long playing around with this (and for some time unknowingly working on something that just reproduced what you had obviously thought of before and commented out...), here's the best I've come up with:
```
    starters_aliases_by_origin = groupby(
        sorted(_get_starters_aliases(), key=itemgetter("origin")),
        key=itemgetter("origin"),
    )
    for origin, starter_config in starters_aliases_by_origin:
        click.secho(f"\nStarters from {origin}\n", fg="yellow")
        for starter in starter_config:
            del starter["origin"]
            name = starter.pop("name")
            click.echo(yaml.dump({name: starter}))
```
image

If we care about ordering so that `kedro` is always first, you could do the below, but personally I don't think it's worth the extra complexity and would just leave it unordered.
```
    starters_aliases_by_origin = {k: list(v) for k, v in groupby(
        sorted(_get_starters_aliases(), key=itemgetter("origin")),
        key=itemgetter("origin"),
    )}
    for origin, starter_config in sorted(starters_aliases_by_origin.items(), key=lambda x: x[0] != "kedro"):
```

AntonyMilneQB (commented):
Thanks for all your work on this @Galileo-Galilei. As soon as there's some tests and docs feel free to move it from a Draft, and I'm sure I'll be happy to approve 🙂 

Your fake plugin repo is a bit out of date given the new dictionary structure for `STARTER_ALIASES` you're using - here's one that's up to date if you want to play around more: https://github.com/AntonyMilneQB/fake-plugin-starter

kedro/framework/cli/starters.py
```diff
@@ -3,12 +3,16 @@
...
 """
+from itertools import groupby
+from operator import itemgetter
 import os
 import re
 import shutil
 import stat
 import tempfile
 from collections import OrderedDict
+from pandas import value_counts
```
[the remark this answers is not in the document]
Galileo-Galilei: This is an auto import from my IDE, this is not used indeed.

kedro/framework/cli/starters.py
```diff
@@ -62,6 +94,63 @@ def _remove_readonly(func: Callable, path: Path, excinfo: Tuple):  # pragma: no
...
+    # add an extra key to indicate from where the plugin come from
+    starters_aliases = [{**config, "origin": "kedro"} for config in _STARTERS_ALIASES]
+
+    existing_names = {}  # dict {name: module_name}
+    for starter_entry_point in pkg_resources.iter_entry_points(
+        group=ENTRY_POINT_GROUPS["starters"]
+    ):
+        module_name = starter_entry_point.module_name.split(".")[0]
+        for starter_config in starter_entry_point.load():
+            _check_starter_entrypoint_config(module_name, starter_config)
```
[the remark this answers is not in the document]
Galileo-Galilei: Well spotted. I guess it should raise a warning. As a user, I would hate having an error I don't understand here and I cannot correct here.

kedro/framework/cli/starters.py
```diff
@@ -288,7 +399,7 @@ def _get_cookiecutter_dir(
                 f" Specified tag {checkout}. The following tags are available: "
                 + ", ".join(_get_available_tags(template_path))
             )
-        official_starters = sorted(_STARTER_ALIASES)
+        official_starters = sorted(_STARTERS_ALIASES)
```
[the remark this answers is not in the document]
Galileo-Galilei: It will return only inbuilt starters: the dictionarry is always deepcopied when ``_get_starters_aliases`` is called, so I think the message is still giving the right result. But maybe it would be better to call explicitly ``_get_starters_aliases()`` here and modify the message accordingly?

kedro/framework/cli/starters.py
Galileo-Galilei: Yes I played a lot with the formatting too and get to the same conclusion as you. I'd personnally prefer showing built-in starters first but it is really not a big deal.

kedro/framework/cli/starters.py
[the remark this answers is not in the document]
AntonyMilneQB: Ah, sorry I wasn't clear here. The problem I see isn't so much that this shows only inbuilt starters (though maybe that is a small problem also) but what the exception message will say. Won't this line print out the whole dictionary, with all its subkeys etc., rather than just showing the alias names of the starters?
```
yaml.safe_dump(official_starters)
```

AntonyMilneQB:
Hello @Galileo-Galilei, I was just wondering if you would mind if we on the kedro team took this over? It feels like a nice feature and I think we're most of the way there so would be nice to get it in🙂 

Am I right in thinking that it's basically just tests and documentation left to do, or did you want to put anything else in here?

Galileo-Galilei:
Hi @AntonyMilneQB, sorry for the late reply. 

There are indeed "only" tests and documentation to write but it's still a couple of hours of work (there are a bunch of edge case scenarii + the test setup is quite complicated to emulate plugin registration). I have very little time for now, so feel free to take over the PR, else I might resume working on it in a couple of weeks.  

Thank you very much for the guidance through the code review and sorry for not being as available as I wish I could.

PR #1275, opened 2022-02-22
[KED-3043] Add support for Python 3.10

MerelTheisenQB:
## Description
Add support for Python 3.10

## Development notes
Significant things to note:

1. `moto` had to be updated to a newer version, however when the tests are run in parallel either `test_incremental_dataset` or `test_partioned_dataset` would fail intermittently, so I changed the tests to not be run in parallel anymore. This might not be an ideal solution, because it adds run time of about 4 mins to each unit test build. The lower version of `moto` could still be used for Python versions < 3.10, but the 3.10 build would then still fail if run in parallel. 
2. For Python 3.10 `matplotlib>=3.5` is needed, which breaks `holoviews`. I've disabled the holoviews tests for Python 3.10, but also had to ignore that dataset from the test coverage check in `pyproject.toml`. 
3. Pylint gives different results for python versions < 3.10 compared to python 3.10. I've opened an issue on the `pylint` repo: https://github.com/PyCQA/pylint/issues/5832 My suggestion for now is just mark the `lint-3.10` build as not required. 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1275"></a>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ lint:
 	pre-commit run -a --hook-stage manual $(hook)
 
 test:
-	pytest tests --cov-config pyproject.toml --numprocesses 4 --dist loadfile
+	pytest tests --cov-config pyproject.toml
 
 test-no-spark:
 	pytest tests --no-cov --ignore tests/extras/datasets/spark --numprocesses 4 --dist loadfile
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ test:
 	pytest tests --cov-config pyproject.toml
 
 test-no-spark:
-	pytest tests --no-cov --ignore tests/extras/datasets/spark --numprocesses 4 --dist loadfile
+	pytest tests --no-cov --ignore tests/extras/datasets/spark
 
 e2e-tests:
 	behave
--- a/Makefile
+++ b/Makefile
@@ -16,10 +16,10 @@ lint:
 	pre-commit run -a --hook-stage manual $(hook)
 
 test:
-	pytest tests --cov-config pyproject.toml
+	pytest tests --cov-config pyproject.toml --numprocesses 4 --dist loadfile
 
 test-no-spark:
-	pytest tests --no-cov --ignore tests/extras/datasets/spark
+	pytest tests --no-cov --ignore tests/extras/datasets/spark --numprocesses 4 --dist loadfile
 
 e2e-tests:
 	behave
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,7 @@ min-public-methods = 1
 [tool.coverage.report]
 fail_under = 100
 show_missing = true
-omit = ["kedro/templates/*", "kedro/framework/cli/hooks/specs.py", "kedro/framework/hooks/specs.py", "kedro/extras/datasets/tensorflow/*"]
+omit = ["kedro/templates/*", "kedro/framework/cli/hooks/specs.py", "kedro/framework/hooks/specs.py", "kedro/extras/datasets/tensorflow/*", "kedro/extras/datasets/holoviews/*"]
 exclude_lines = ["pragma: no cover", "raise NotImplementedError"]
 
 [tool.pytest.ini_options]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,7 @@ min-public-methods = 1
 [tool.coverage.report]
 fail_under = 100
 show_missing = true
-omit = ["kedro/templates/*", "kedro/framework/cli/hooks/specs.py", "kedro/framework/hooks/specs.py", "kedro/extras/datasets/tensorflow/*", "kedro/extras/datasets/holoviews/*"]
+omit = ["kedro/templates/*", "kedro/framework/cli/hooks/specs.py", "kedro/framework/hooks/specs.py", "kedro/extras/datasets/tensorflow/*", "kedro/extras/datasets/holoviews/*", "tests/*"]
 exclude_lines = ["pragma: no cover", "raise NotImplementedError"]
 
 [tool.pytest.ini_options]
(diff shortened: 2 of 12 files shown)


pyproject.toml
```diff
@@ -39,7 +39,7 @@ min-public-methods = 1
 [tool.coverage.report]
 fail_under = 100
 show_missing = true
-omit = ["kedro/templates/*", "kedro/framework/cli/hooks/specs.py", "kedro/framework/hooks/specs.py", "kedro/extras/datasets/tensorflow/*"]
+omit = ["kedro/templates/*", "kedro/framework/cli/hooks/specs.py", "kedro/framework/hooks/specs.py", "kedro/extras/datasets/tensorflow/*", "kedro/extras/datasets/holoviews/*", "tests/*"]
```
MerelTheisenQB, line 42: I've added "tests/*" to be omitted, because the coverage was failing for Python 3.10 because `tests_holoviews_writer.py` is skipped. But it made me wonder why the tests weren't omitted in the first place?

AntonyMilneQB: I think it's because the existing `skipif`s refer to Windows, and in Windows we don't do exactly `make test` - we actually do `make test-no-spark`, which includes `--no-cov`.

Fine by me if we just ignore everything in `tests/*`, but in this case it would be tidy to remove the `--ignore tests/...` options wherever they appear elsewhere (Makefile and CircleCI config I guess).

AntonyMilneQB (approved):
Nice work!

On the parallelisation of unit tests: I don't really mind the slowdown, but it would be possible to run all of them in parallel apart from 3.10. You'd need to add a `when/unless` clause into the `unit_tests` job a bit like we have here (only in `main`, not `develop`): https://github.com/kedro-org/kedro/blob/main/.circleci/continue_config.yml#L231-L244

tests/io/test_incremental_dataset.py
```diff
@@ -350,6 +350,7 @@ def mocked_s3_bucket():
     with mock_s3():
         conn = boto3.client(
             "s3",
+            region_name="us-east-1",
```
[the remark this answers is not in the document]
MerelTheisenQB: When I upgraded `moto` this and `test_partioned_dataset` started failing with errors about not having permission to create a bucket. It looks like the credentials are passed slightly differently for the matplotlib and dask.ParquetDataSet, but I'd have to investigate more to figure out why those didn't need the region 🤔

tests/extras/datasets/holoviews/test_holoviews_writer.py
```diff
@@ -33,6 +34,10 @@ def versioned_hv_writer(filepath_png, load_version, save_version):
     return HoloviewsWriter(filepath_png, version=Version(load_version, save_version))
 
 
+@pytest.mark.skipif(
+    sys.version.split(".")[1] == "10",
```
MerelTheisenQB, line 38:
```suggestion
    sys.version_info == (3, 10),
```

SajidAlamQB (approved):
Awesome LGTM! 🎉

pyproject.toml
MerelTheisenQB: I'm not sure that's correct, `--ignore ...` makes sure not to run the tests, so that should still happen. If we remove `--no-cov` for windows it will fail coverage because certain tests aren't run on windows thus not "covering" that code.

tests/io/test_incremental_dataset.py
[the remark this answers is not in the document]
AntonyMilneQB: Weird. I wouldn't worry about it much though, especially given how much time we've spent on AWS credentials in tests recently 😬

tests/extras/datasets/holoviews/test_holoviews_writer.py
```diff
@@ -112,6 +117,10 @@ def test_protocol_usage(self, filepath, instance_type, credentials):
         assert isinstance(data_set._filepath, PurePosixPath)
 
 
+@pytest.mark.skipif(
+   sys.version_info == (3, 10),
+    reason="Python 3.10 needs matplotlib>=3.5 which breaks holoviews.",
+)
```
lorenabalan, line 123:
```suggestion
   sys.version_info == (3, 10),
   reason="Python 3.10 needs matplotlib>=3.5 which breaks holoviews.",
)
```

Makefile
```diff
@@ -16,10 +16,10 @@ lint:
 	pre-commit run -a --hook-stage manual $(hook)
 
 test:
-	pytest tests --cov-config pyproject.toml --numprocesses 4 --dist loadfile
+	pytest tests --cov-config pyproject.toml
 
 test-no-spark:
-	pytest tests --no-cov --ignore tests/extras/datasets/spark --numprocesses 4 --dist loadfile
+	pytest tests --no-cov --ignore tests/extras/datasets/spark
```
lorenabalan, line 22: These were quite handy in making our tests run slightly faster by parallelising. What's the reason for removing it?

pyproject.toml
lorenabalan: I'm also not entirely sure why we run coverage on tests, though _sometimes_ it handily uncovers some overcomplicated fixtures or tests.

AntonyMilneQB: Oh yes, sorry - forget what I said 🙈

Makefile
MerelTheisenQB: See point 1: https://github.com/kedro-org/kedro/pull/1275#issue-1146834454, I could implement Antony's suggestion: https://github.com/kedro-org/kedro/pull/1275#pullrequestreview-891371768

tests/io/test_incremental_dataset.py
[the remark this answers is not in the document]
MerelTheisenQB: No, it's just random like the fake access key and secret 🙂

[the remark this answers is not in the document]
MerelTheisenQB: and yes, after upgrading `moto` this test and the `test_partioned_dataset.py` starting throwing errors about not having permission to create a bucket. Stackoverflow suggested adding the region and that seemed to work.

PR #1309, opened 2022-03-03
Add runner class name to pipeline hooks + introduce `after_command_run` CLI hook spec

datajoely:
## Description

This PR adds some functionality to 3 existing hooks and introduces one new CLI hook:

### Modified hooks:

- `before_pipeline_run`
- `after_pipeline_run`
- `on_pipeline_error`

The runner class name is now included within the `run_params` dictionary.

### New hooks:

- `after_command_run` (CLI hook)

This has been requested several times over the last year or so the two usecases that come to mind are:

1. Ability to tell telemetry which commands actually worked and which failed 
2. The developer of `kedro-dvc` [specifically asked to expose](https://github.com/kedro-org/kedro/discussions/837#discussioncomment-2216745) this so that they can install their own configuration once `kedro new` is executed.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1309"></a>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -221,6 +221,7 @@ The parameters should look like this:
 * To run a pipeline in parallel, use `kedro run --runner=ParallelRunner` rather than `--parallel` or `-p`.
 * If you call `ConfigLoader` or `TemplatedConfigLoader` directly, update the keyword arguments `conf_root` to `conf_source` and `extra_params` to `runtime_params`.
 * If you use `KedroContext` to access `ConfigLoader`, use `settings.CONFIG_LOADER_CLASS` to access the currently used `ConfigLoader` instead.
+* The signature of `KedroContext` has changed and now needs `config_loader` and `hook_manager` as additional arguments of type `ConfigLoader` and `PluginManager` respectively.
 
 # Release 0.17.7
 
--- a/docs/source/kedro_project_setup/configuration.md
+++ b/docs/source/kedro_project_setup/configuration.md
@@ -15,8 +15,10 @@ Kedro-specific configuration (e.g., `DataCatalog` configuration for IO) is loade
 
 ```python
 from kedro.config import ConfigLoader
+from kedro.framework.project import settings
 
-conf_loader = ConfigLoader(conf_source="conf", env="local")
+conf_path = str(project_path / settings.CONF_SOURCE)
+conf_loader = ConfigLoader(conf_source=conf_path, env="local")
 conf_catalog = conf_loader.get("catalog*", "catalog*/**")
 ```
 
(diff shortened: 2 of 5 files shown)


datajoely:
Opted to provide the name of the runner class to avoid circular import, but the user value is identical

docs/source/extend_kedro/hooks.md
```diff
@@ -48,8 +48,9 @@ The naming convention for error hooks is `on_<noun>_error`, in which:
 Lastly, Kedro defines a small set of CLI hooks that inject additional behaviour around execution of a Kedro CLI command:
 
 * `before_command_run`
+* `after_command_run`
 
-This is what the [`kedro-telemetry`](https://github.com/kedro-org/kedro-plugins/tree/main/kedro-telemetry) plugin relies on under the hood in order to be able to collect CLI usage statistics.
+This is what the [`kedro-telemetry`](https://github.com/kedro-org/kedro-plugins/tree/main/kedro-telemetry) plugin relies on these under the hood in order to be able to collect CLI usage statistics.
```
MerelTheisenQB, line 53:
```suggestion
The [`kedro-telemetry`](https://github.com/kedro-org/kedro-plugins/tree/main/kedro-telemetry) plugin relies on these hooks under the hood in order to be able to collect CLI usage statistics.
```

RELEASE.md
```diff
@@ -78,6 +78,8 @@
 * Removed `RegistrationSpecs` and all registration hooks that belonged to it. Going forward users can register custom library components through `settings.py`.
 * Added the `PluginManager` `hook_manager` argument to `KedroContext` and the `Runner.run()` method, which will be provided by the `KedroSession`.
 * Removed the public method `get_hook_manager()` and replaced its functionality by `_create_hook_manager()`.
+* Added `run_id` and `runner_name` attributes to pipeline hook specs.
```
MerelTheisenQB, line 81: What would a user use the `run_id` for?

datajoely: I think it makes a lot of sense in situations where you are logging runs to another system, we expose it the `node_specs` so I felt it was reasonable to introduce.

MerelTheisenQB: We had a discussion yesterday where we decided that one session = one run, and so we don't need the `run_id` anymore. I'm in the process of writing a summary of the discussion, but for now see: https://github.com/kedro-org/kedro/pull/1329

datajoely: Thanks! Would it be okay to expose the `session_id`?

datajoely: Actually removed because it's already in `run_params` I was being silly/

kedro/framework/session/session.py
```diff
@@ -387,6 +387,7 @@ def run(  # pylint: disable=too-many-arguments,too-many-locals
             "load_versions": load_versions,
             "extra_params": extra_params,
             "pipeline_name": pipeline_name,
+            "runner": getattr(runner, "__name__", str(runner)),
```
MerelTheisenQB, line 390: This is a much neater solution! 👍

tests/framework/cli/test_cli_hooks.py
```diff
@@ -108,3 +120,9 @@ def test_kedro_cli_should_invoke_cli_hooks_from_plugin(
             f"Before command `{command}` run for project {fake_metadata}"
             in result.output
         )
+
+        # 'pipeline list' and 'info' aren't actually in the click structure and
+        # return exit code 2 ('invalid usage of some shell built-in command')
```
MerelTheisenQB, line 125: Do you happen to know why these command aren't in the click structure? Should they be?

MerelTheisenQB, line 125: Do you happen to know why these command aren't in the click structure? Should they be?

kedro/framework/cli/cli.py
```diff
@@ -138,13 +138,21 @@ def main(
...
+                complete_var=complete_var,
+                standalone_mode=standalone_mode,
+                **extra,
+            )
+        except SystemExit as exc:
+            self._cli_hook_manager.hook.after_command_run(  # pylint: disable=no-member
+                project_metadata=self._metadata, command_args=args, exit_code=exc.code
+            )
+            # click.core.main() method exits by default, we capture this and then
+            # exit as originally intended
```
MerelTheisenQB, line 154: I'd suggest moving the comment up just before the `except`.

And just to clarify: `click.core.main()` will always throw a `SystemExit` and so we'll always execute the `except` block?

MerelTheisenQB, line 154: I'd suggest moving the comment up just before the `except`.

And just to clarify: `click.core.main()` will always throw a `SystemExit` and so we'll always execute the `except` block?

MerelTheisenQB (approved):
LGTM! Much cleaner than before 👍  
I've got some minor comments and questions, but nothing to block merging.

tests/framework/cli/test_cli_hooks.py
datajoely: So I think the original test was just checking the command was captured, telemetry sends it even if it's actually invalid. I think it's more of an oversight than anything, but I would argue it would make sense to actually switch telemetry to the new one as we could track what the most common failed commands are.

datajoely: See above

kedro/framework/cli/cli.py
datajoely: Yeah basically every route of this exception handler rises `sys.exit({{code}})`

datajoely: See below

datajoely: Updated comment

idanov (approved):
LGTM! Happy the API surface doesn't increase by much.

Issue #1273, opened 2022-02-21
[KED-2629] Users can't provide custom run_id or save_version to KedroSession


AntonyMilneQB:
(transfer from Jira, created by @lorenabalan)

_This may be acceptable behaviour, but we should document it better in that case._

 

Working on [this PR](https://github.com/quantumblacklabs/private-kedro/pull/1098) (see discussion in the thread too), I discovered that in the new session model, users can't set their own custom `run_id`, or use a different way to generate a save version (for example use a different format). They can modify `KedroContext._get_save_version()` or `KedroContext._get_run_id()` but that's not what will be stored in the session store - instead it will contain only the values equivalent to `session_id`.

When a session is created, a `session_id` is generated (timestamp) and written to the store. During a session run, that `session_id` is loaded from the store and is used for both `save_version` and `run_id`, and it'll be the same timestamp every time, i.e.

1 session ↔️ 1 run ↔️ 1 run_id=session_id=save_version

If that's the case, as Antony pointed out, what's the optimal behaviour in this case: 

```python
 with KedroSession.create(...) as session:
    session.run(pipeline1)
    session.run(pipeline2)
```
Note this isn't as strange a thing to do as it might initially seem. In Jupyter, a user could well do multiple `session.run`.

**This ticket includes both the design and implementing the solution. Setup a team discussion to go through design suggestion.**

Following the discussion the team had about this issue these tasks need to be done:

- [x] #1313
- [x] #1314

After `0.18.0` has been released:

- [x] [Fix the assumption in Kedro Experiment Tracking that the user is in a CLI workflow](https://github.com/kedro-org/kedro-viz/issues/764)

AntonyMilneQB:
[This comment in Kedro-Viz](https://github.com/kedro-org/kedro-viz/blob/main/package/kedro_viz/data_access/repositories/runs.py#L84) makes me think that @limdauto was anticipating that run_id should not necessarily be the same as the timestamp?

Also from [this Discord conversation](https://discord.com/channels/778216384475693066/778998585454755870/945349995924684891) with @shaunc on possible integration with DVC, my gut feeling is that saying run_id = session_id = save_version is maybe too restrictive and we should allow for controlling some (all?) of these independently.

shaunc:
[...]

PR #1329, opened 2022-03-08
Enforce `1 session = 1 run`

MerelTheisenQB:
## Description
After the discussion about https://github.com/kedro-org/kedro/issues/1273, it has been decided that a session can only ever have one run. Currently, we've got the concept of a `run_id` in the codebase, which is mainly a remnant from the `journal` and doesn't serve a purpose. I will write up a summary of the whole discussion, but baseline is that a session can only have 1 run, and so the `session_id` and `run_id` will always be the same. 

## Development notes
* Removed `run_id` where it didn't serve a purpose. 
* Changed the `run_params` pipeline hook specs to refer to `session_id` instead of `run_id`. 
* Enforce/inform users that they shouldn't run multiple runs during 1 session: https://github.com/kedro-org/kedro/pull/1329/commits/55fb343f6dd4682884d49fb719085588d9b44d9f

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1329"></a>


docs/source/extend_kedro/hooks.md
```diff
@@ -411,7 +406,7 @@ class DataValidationHooks:
                 expectation_suite,
             )
             expectation_context.run_validation_operator(
-                "action_list_operator", assets_to_validate=[batch], run_id=run_id
+                "action_list_operator", assets_to_validate=[batch]
```
AntonyMilneQB, line 409: I was wondering whether GE would still work without the `run_id` argument here. It appears to be an optional argument but I think it might be quite useful for users to keep track of validations done over multiple runs.

AntonyMilneQB (commented):
Are we sure that we want to remove `run_id` from `after_catalog_created` rather than replacing it with `session_id`? 🤔  Or is there just not an easy way to access `session_id` at the point we would need it?

MerelTheisenQB:
@AntonyMilneQB I think both of your points are valid and so perhaps instead of removing the `run_id` it makes more sense to just replace it with `session_id`.

kedro/framework/session/session.py
```diff
@@ -379,7 +398,8 @@ def run(  # pylint: disable=too-many-arguments,too-many-locals
         )
 
         try:
-            run_result = runner.run(filtered_pipeline, catalog, hook_manager, run_id)
+            run_result = runner.run(filtered_pipeline, catalog, hook_manager)
+            self._run_called = True
```
lorenabalan, line 402: So this is just in case the run is successful. If there's a problem with the pipeline halfway through, we can call `session.run()` again. Is that the intended course? Makes sense for debugging, if a little confusing for the user (but I've run it multiple times to failure, it allowed me to that then!).

kedro/framework/session/session.py
```diff
@@ -327,7 +339,14 @@ def run(  # pylint: disable=too-many-arguments,too-many-locals
         # Report project name
         self._logger.info("** Kedro project %s", self._project_path.name)
 
-        save_version = run_id = self.store["session_id"]
+        if self._run_called:
+            raise KedroSessionError(
+                "A run has already been executed as part of the"
```
lorenabalan, line 344:
```suggestion
                "A run has already been completed as part of the"
```
Maybe, to ease with the confusion mentioned in the previous comment.

tests/framework/session/test_session.py
```diff
@@ -567,7 +567,74 @@ def test_run(
...
+                "__default__": mocker.Mock(),
+            },
+        )
+        mock_context = mock_context_class.return_value
+        mock_catalog = mock_context._get_catalog.return_value
+        mock_runner = mocker.Mock()
+        mock_pipeline = mock_pipelines.__getitem__.return_value.filter.return_value
+
+        message = (
+            "A run has already been executed as part of the active KedroSession. "
```
lorenabalan, line 608:
```suggestion
            "A run has already been completed as part of the active KedroSession. "
```

tests/framework/session/test_session.py
```diff
@@ -567,7 +567,74 @@ def test_run(
...
+        mock_hook.after_pipeline_run.assert_called_once_with(
+            run_params=record_data,
+            run_result=mock_runner.run.return_value,
+            pipeline=mock_pipeline,
+            catalog=mock_catalog,
+        )
+
+    @pytest.mark.usefixtures("mock_settings_context_class")
+    @pytest.mark.parametrize("fake_pipeline_name", [None, _FAKE_PIPELINE_NAME])
+    def test_run_multiple_times(  # pylint: disable=too-many-locals
```
lorenabalan, line 581: We should add a test for multiple runs with a broken pipeline or broken nodes.

lorenabalan (commented):
I **love** that we're getting rid of all the `run_id` passed around everywhere, code feels so much more breathable now!! 
I'm ambivalent about the error on re-run, I have no strong opinion for or against it. On the one hand it does the job, on the other hand I worry that it's gonna end up throwaway code a bit further down the line. I have an open question about what should happen in case the pipeline run fails and maybe that'll clarify things in my head.

kedro/framework/session/session.py
MerelTheisenQB: Yes, so my thinking was that if you're in an interactive workflow to experiment/debug your pipeline it would be overhead if you need to recreate a new `session` if the pipeline didn't run successfully. At the same time this is a niche case just for those people that would actually do a `run` inside a notebook/ipython session.

idanov:
I wonder whether it's worth keeping the concept of a `run_id`, but only setting it to the `session_id`, in order to have a much less distruptive change in APIs and also concepts. Would that make it compatible with Kedro Viz without the need for a new release?

Also a lot of hooks here were receiving the `run_id` and now no longer will receive it, but the users might still need to know the `run_id` / `session_id` at that particular hook. Have you checked with users if they need the `run_id` (now `session_id`) in those hooks or that's no longer needed?

kedro/framework/session/session.py
```diff
@@ -94,18 +102,20 @@ class KedroSession:
...
     """
 
-    def __init__(
+    def __init__(  # pylint: disable=too-many-arguments
         self,
         session_id: str,
         package_name: str = None,
         project_path: Union[Path, str] = None,
         save_on_close: bool = False,
+        run_called: bool = False,
```
idanov, line 111: Do we need to expose that in the constructor? What would be the usecase of someone creating a session, which is non-runnable from the start?

MerelTheisenQB: What do you mean with `non-runnable`? When the pipeline is completely broken to start with?

AntonyMilneQB: I think the point is that no one would ever do `KedroSession(..., run_called=True)` and so there's no need to put this argument in the constructor. Instead we could just do `self._run_called = False` on L118.

MerelTheisenQB: ah okay non-runnable in that sense. Yep will change it!

kedro/framework/session/session.py
```diff
@@ -75,6 +75,14 @@ def _jsonify_cli_context(ctx: click.core.Context) -> Dict[str, Any]:
     }
 
 
+class KedroSessionError(Exception):
+    """``KedroSessionError`` raised by ``KedroSession``
+    in case of run failure as part of a session.
```
AntonyMilneQB, line 80:
```suggestion
    """``KedroSessionError`` raised by ``KedroSession``
    in the case that multiple runs are attempted in one session.
```

kedro/framework/context/context.py
```diff
@@ -290,7 +291,7 @@ def _get_catalog(
             feed_dict=feed_dict,
             save_version=save_version,
```
AntonyMilneQB, line 292: Can `save_version` ever be different from `session_id` now actually? 🤔  On seconds thoughts maybe we don't need to keep this in the hook spec after all if it's just a duplicate of `save_version`?

AntonyMilneQB (approved):
Nice work! I'm actually now not so sure about the hook specs after I realised we already have `save_version` there, sorry 😬  No strong feelings though - up to you.

kedro/framework/context/context.py
MerelTheisenQB: At the moment they will always be the same, but this goes back to the question whether we should allow users to have a custom `save_version`, which requires user research. With the mindset of "don't add things that could potentially be used in the future", I'll remove it for now.

lorenabalan (approved):
I was a **huge** fan of getting rid of passing the id around everywhere. :( I seriously doubt it was a heavily used feature by any user, I'm worried we're leaving around code that's only hypothetical and probably 90% dead. Having said that, it's not a hill I'm willing to die on anytime soon, so happy to let it go for now. 
We should add something in the release notes though that we are replacing `run_id` with sth else in the hook specs, and that users should update their code accordingly (1 short migration guide note).

MerelTheisenQB:
> I was a **huge** fan of getting rid of passing the id around everywhere. :( I seriously doubt it was a heavily used feature by any user, I'm worried we're leaving around code that's only hypothetical and probably 90% dead. Having said that, it's not a hill I'm willing to die on anytime soon, so happy to let it go for now. 

I hear you, I would've preferred to remove it as well, but I think there's an argument around being able to find back the run through the id when using external monitoring/orchestration systems. Anyway, I think this needs user research before we can make a final decision on removing it. 

> We should add something in the release notes though that we are replacing `run_id` with sth else in the hook specs, and that users should update their code accordingly (1 short migration guide note).

Yes I'll add this!

PR #1342, opened 2022-03-11
Speed up Windows CI

AntonyMilneQB:
## Description
I've reviewed our Windows CircleCI setup.
* the general scheme, including a separate windows_reqs.txt file, already seems pretty good. I experimented with various possible improvements but found that the current setup actually worked best so have left it largely unchanged
* installing dependencies was taking up a lot of time. I've added the conda environment to our cache and this speeds things up _a lot_... 
  * Windows unit tests have gone from ~20mins to ~14mins ⚡ 
  * Windows pip compile from ~17 mins to ~9mins 💥 
  * Windows e2e tests similar to before
* while working on this, I also noticed that the Docker images for the Unix jobs are not actually working as we think (quite aside from the build cron jobs not currently running): they do not have the right requirements pre-installed on them so we are actually installing requirements on to them again in CI. This is ok but kind of defeats the point of the pre-build, and means that there's potential for speedup there. I think I know why this is happening - will follow up in a separate ticket after checking some stuff with @limdauto

Handy refs from the old repo that helped me figure out how the e2e tests were working:
https://github.com/quantumblacklabs/private-kedro/pull/1048
https://github.com/quantumblacklabs/private-kedro/pull/1126
Note that e2e tests use the `kedro_builder` conda environment to control Python version and then venv inside that to install project requirements (see environment.py). This seems like a reasonable setup, and the only thing I could cleanup here was remove an old hacky fix that's no longer required.

### What are we caching on Windows?
1. The conda and pip package caches. This is saved once for Python 3.7 and then used for all Python versions. Nothing has changed here in this PR.
2. The actual `kedro_builder` conda environment files (`c:\tools\miniconda3\envs\kedro_builder`). This is saved separately for each Python version. This is the new thing I've introduced in this PR that speeds things up a lot.

### Will there be contamination between different Python versions if they all have the same conda environment name `kedro_builder`?
No, because the caching key includes the Python version. This means that `c:\tools\miniconda3\envs\kedro_builder` is saved and loaded completely separately for different Python versions.

### Does this caching risk the integrity of our tests?
No, because:
1. the caching key includes the checksum of test_requirements.txt and requirements.txt, so if one of those files change then there will be no cache found and so nothing reused
2. we do `pip install -U test_requirements.txt` anyway, so dependencies in the environment are always what they should be and not something from an old environment

**Overall this approach is now analogous to what we do on Unix, where we have a Docker image for each Python version, with a pre-built `kedro_builder` conda env inside there.**

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1342"></a>

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -50,7 +50,6 @@ commands:
       - run: which -a python
       - run: which -a pip
       - run: pip freeze
-      - run: conda info
       - run: python -m pip freeze
       - run:
           name: Install requirements and test requirements
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -183,7 +183,12 @@ jobs:
       - checkout
       - win_setup_conda:
           python_version: <<parameters.python_version>>
-      - run: conda activate kedro_builder; pip install behave==1.2.6
+#      - restore_cache:
+#          key: kedro-deps-v1-win-{{ checksum "requirements.txt" }}-{{ checksum "test_requirements.txt" }}
+#      - run:
+#          name: Install all requirements
+#          command: conda activate kedro_builder; pip install -r test_requirements.txt -U
+      - run: conda activate kedro_builder; pip install psutil==5.6.7 psutil==5.6.7
       - run:
           name: Install 'make' command
           command: choco install make
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -183,12 +183,12 @@ jobs:
       - checkout
       - win_setup_conda:
           python_version: <<parameters.python_version>>
-#      - restore_cache:
-#          key: kedro-deps-v1-win-{{ checksum "requirements.txt" }}-{{ checksum "test_requirements.txt" }}
-#      - run:
-#          name: Install all requirements
-#          command: conda activate kedro_builder; pip install -r test_requirements.txt -U
-      - run: conda activate kedro_builder; pip install behave==1.2.6 psutil==5.6.7
+      - restore_cache:
+          key: kedro-deps-v1-win-{{ checksum "requirements.txt" }}-{{ checksum "test_requirements.txt" }}
+      - run:
+          name: Install all requirements
+          command: conda activate kedro_builder; pip install -r test_requirements.txt -U
+#      - run: conda activate kedro_builder; pip install behave==1.2.6 psutil==5.6.7
       - run:
           name: Install 'make' command
           command: choco install make
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -135,9 +135,9 @@ commands:
       - run:
           name: Install Fiona
           command: conda activate kedro_builder; conda install -c conda-forge fiona -y
-#      - run:
-#          name: Install all requirements
-#          command: conda activate kedro_builder; pip install -r test_requirements.txt -U
+      - run:
+          name: Install all requirements
+          command: conda activate kedro_builder; pip install -r test_requirements.txt -U
       - run:
           name: Print Python environment
           command: conda activate kedro_builder; make print-python-env
@@ -174,6 +174,7 @@ jobs:
 
   # What if checksum changes?
   # venv in conda?
+  # Need pywin pinning?
   win_e2e_tests:
     parameters:
       python_version:
@@ -187,10 +188,9 @@ jobs:
           python_version: <<parameters.python_version>>
       - restore_cache:
           key: kedro-deps-v1-win-{{ checksum "requirements.txt" }}-{{ checksum "test_requirements.txt" }}
-#      - run:
-#          name: Install all requirements
-#          command: conda activate kedro_builder; pip install -r test_requirements.txt -U
-#      - run: conda activate kedro_builder; pip install behave==1.2.6 psutil==5.6.7
+      - run:
+          name: Install dependencies
+          command: conda activate kedro_builder; pip install -r features/windows_reqs.txt
       - run:
           name: Install 'make' command
           command: choco install make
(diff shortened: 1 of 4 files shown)


features/steps/cli_steps.py
```diff
@@ -174,48 +172,6 @@ def exec_kedro_target_checked(context, command):
         assert False
 
 
-@given('I have created new environment "{}"')
```
AntonyMilneQB, line 177: There were some steps not used in any tests, found using `behave -f steps.usage --dry-run features/`. I've removed them but let me know if there's some reason we want to keep them.

features/environment.py
```diff
@@ -125,13 +125,5 @@ def _install_project_requirements(context):
...
     install_reqs = [req for req in install_reqs if "{" not in req]
     install_reqs.append(".[pandas.CSVDataSet]")
-
-    # JupyterLab indirectly depends on pywin32 on Windows. Newer versions of pywin32
-    # (e.g. 3xx, to which jupyterlab~=3.0 defaults) have a bug that prevents
-    # JupyterLab from running, hence the version is forcefully set to 225.
-    # More details: https://github.com/mhammond/pywin32/issues/1431
-    if sys.platform.startswith("win"):
-        install_reqs.append("pywin32==225")
-
```
AntonyMilneQB, line 135: We now support `jupyter-client` 7 and don't need this annoying fix any more.

Makefile
```diff
@@ -10,7 +10,7 @@ clean:
 	pre-commit clean || true
 
 install-pip-setuptools:
-	python -m pip install -U "pip>=21.2" "setuptools>=38.0" wheel
+	pip install -U "pip>=21.2" "setuptools>=38.0" wheel
```
AntonyMilneQB, line 13: This wasn't wrong before, but it was a bit confusing since our CircleCI conda environment is setup correctly so that `pip` and `python` are in the same environment, and we don't do `python -m` for other commands in the CircleCI setup.

.circleci/continue_config.yml
```diff
@@ -128,13 +128,19 @@ commands:
...
+          name: Restore conda environment cache
+          key: kedro-deps-v1-win-<<parameters.python_version>>-{{ checksum "requirements.txt" }}-{{ checksum "test_requirements.txt" }}
       - run:
-          name: Install GDAL
-          command: conda activate kedro_builder; conda install -c conda-forge gdal -y
-      - run:
-          name: Install Fiona
-          command: conda activate kedro_builder; conda install -c conda-forge fiona -y
+          name: Install GDAL and Fiona
+          command: conda activate kedro_builder; conda install gdal fiona -c conda-forge -y
```
AntonyMilneQB, line 143: This isn't just tidying but actually makes things ~30s faster, since `conda install fiona` was downgrading some of the dependencies installed in the previous `conda install gdal` step. Installing them both at the same time is better.

.circleci/continue_config.yml
```diff
@@ -278,16 +287,23 @@ jobs:
...
-                  # Cache pip cache and conda packages directories
-                  - c:\tools\miniconda3\pkgs
-                  - c:\users\circleci\appdata\local\pip\cache
+          - save_cache:
+              name: Save Python package cache
+              key: kedro-deps-v1-win-{{ checksum "requirements.txt" }}-{{ checksum "test_requirements.txt" }}
+              paths:
+                # Cache pip cache and conda packages directories
+                - c:\tools\miniconda3\pkgs
+                - c:\users\circleci\appdata\local\pip\cache
```
AntonyMilneQB, line 301: This is the same as before, just I gave it a name.

.circleci/continue_config.yml
```diff
@@ -278,16 +287,23 @@ jobs:
...
+      - save_cache:
+          name: Save conda environment cache
+          key: kedro-deps-v1-win-<<parameters.python_version>>-{{ checksum "requirements.txt" }}-{{ checksum "test_requirements.txt" }}
+          paths:
+            - c:\tools\miniconda3\envs\kedro_builder
```
AntonyMilneQB, line 306: This is the new caching of the environment itself.

.circleci/continue_config.yml
```diff
@@ -128,13 +128,19 @@ commands:
...
+    parameters:
+      python_version:
+        type: string
     steps:
+      - restore_cache:
+          name: Restore package cache
+          key: kedro-deps-v1-win-{{ checksum "requirements.txt" }}-{{ checksum "test_requirements.txt" }}
+      - restore_cache:
+          name: Restore conda environment cache
+          key: kedro-deps-v1-win-<<parameters.python_version>>-{{ checksum "requirements.txt" }}-{{ checksum "test_requirements.txt" }}
```
AntonyMilneQB, line 140: This is the new bit of caching.

.circleci/continue_config.yml
lorenabalan: Nice catch! Would be nice to add an inline comment about that, completely optional though.

.circleci/continue_config.yml
```diff
@@ -186,6 +191,10 @@ jobs:
       - run:
           name: Install 'make' command
           command: choco install make
+      # We don't use the `win_setup` command here, which would install the full set
+      # of requirements used by unit tests. Even when those packages are cached
+      # it is faster to just install the minimal set of dependencies needed for e2e
+      # tests in a new empty environment rather than restore the cache.
```
lorenabalan, line 197: 🥇 excellent explanation. 👍

lorenabalan (approved):
LGTM, great write-up too as per usual. 🏆

MerelTheisenQB (approved):
This is great @AntonyMilneQB ! Such a good speed-up on the builds and thanks for the clean up 😄 ⭐

Issue #1344, opened 2022-03-14
Add abfss in the list of cloud protocols


philomine:
## Description
abfss is not in the list of cloud protocols in the module kedro/io/core.py.

## Context
I'm currently testing kedro on my project and I am facing an issue when trying to load an ExcelDataSet from an abfss cloud storage.

This works
```{python}
import pandas as pd
pd.read_excel("abfss://container/path/to/excel/file.xlsx", engine="openpyxl")
```

This doesn't
```{python}
from kedro.extras.datasets.pandas import ExcelDataSet
dataset = ExcelDataSet(filepath="abfss://container/path/to/excel/file.xlsx", load_args={"engine": "openpyxl"})
dataset.load()
```

## Possible Implementation
Simply add abfss to the list of cloud protocols.

Currently, in [kedro/io/core.py, line 31](https://github.com/kedro-org/kedro/blob/main/kedro/io/core.py):
```{python}
CLOUD_PROTOCOLS = ("s3", "gcs", "gs", "adl", "abfs")
```

Update with
```{python}
CLOUD_PROTOCOLS = ("s3", "gcs", "gs", "adl", "abfs", "abfss")
```

This solution worked with my issue.

## Possible Alternatives
If adding abfss to the list of cloud protocols is not an option to fix this issue, I can provide more information on my issue.

MerelTheisenQB:
Hi @philomine, thanks for reporting this! This sounds like a very reasonable solution and we'd very much welcome a contribution PR for it 🙂

MerelTheisenQB:
@datajoely We can definitely look at optimising the list of allowed protocols, but currently we have separation between "cloud" and "http" protocols, so it would require a bigger change to handle a generic list like the one above. Also, it doesn't look like `abfss` is included, so we'd have to add that anyway.

datajoely:
Good point - looking into it, it looks like they only make this available via plug-in mechanism once `adlfs` is installed (which I haven't got in this env) 
https://github.com/fsspec/adlfs/blob/fbe43c601e7b162b7a2bf93f7f8b699c747b70e6/setup.py#L40

philomine:
Thanks for your fast answer! I'll open a PR then :)

PR #1348, opened 2022-03-15
Add abfss to the list of cloud protocols

philomine:
## Description

This PR adds `abfss` to the list of cloud protocols in kedro/io/core.py. This enables users to open an abfss path.

## Development notes

Tried to load an instance of `kedro.extras.datasets.pandas.ExcelDataSet`: 
- on the current release: doesn't work
- with this fix: works

## Checklist

Unsure about the way this should be documented and/or tested.

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1348"></a>


datajoely:
Hi @philomine thanks for raising this. @MerelTheisenQB, @idanov I can't shake the feeling we should be more flexible here, we need this change because currently you'll get the following behaviour:

image
 
Is there any merit on making the default case less destructive? It feels like we'll always be playing catch up with fsspec as vendors introduce new protocols

RELEASE.md
```diff
@@ -1,3 +1,12 @@
+# Release 0.17.9
+
+## Major features and improvements
+* Added abfss to list of cloud protocols, enabling abfss paths.
```
lorenabalan, line 4: We don't have a 0.17.8 release either, could we put these below one section please? :)

lorenabalan (approved):

lorenabalan:
> Is there any merit on making the default case less destructive? It feels like we'll always be playing catch up with fsspec as vendors introduce new protocols

This code was introduced on purpose because `fsspec` wasn't parsing the Azure protocol correctly at the time (0.16.x). If this has been corrected, and it parses all protocols correctly (including `http`), we can remove this and go back to relying on `infer_storage_options` from `fsspec`.

datajoely:
Linking to #1344

MerelTheisenQB:
> @MerelTheisenQB do we not want to take @lorenabalan 's suggestion and remove the check on our side and delegate to fsspec entirely?

Ideally we handle it on our side, but I'd suggest doing that in a separate task and not blocking this PR from going in so the `abfss` protocol can already be used in the meantime.

philomine:
Hi! I added the test case suggested. I didn't add any documentation as I don't think it is necessary.

MerelTheisenQB:
> Hi! I added the test case suggested. I didn't add any documentation as I don't think it is necessary.

Thank you! It looks like the `lint` checks are failing. You should be able to fix those by running `make lint` and fix any failures it shows. Let me know if you need help with that.

MerelTheisenQB (approved):
Thanks for this contribution @philomine ⭐

Issue #1336, opened 2022-03-10
Remove the `init_kedro` line magic from the Kedro IPython extension


idanov:
Currently the Kedro IPython extension has two line magic commands, `reload_kedro` and `init_kedro`. The way their are implemented makes `init_kedro` useless and entirely subsumed by `reload_kedro`, so we should drop it. When doing this, we should make sure:

- All `init_kedro` references in the docs are replaced with `reload_kedro`
- All `init_kedro` use cases are still possible with `reload_kedro`

PR #1349, opened 2022-03-15
Remove the `init_kedro` line magic from the Kedro IPython extension

lorenabalan:
## Description

Fixes https://github.com/kedro-org/kedro/issues/1336

## Checklist

### DIFF
--- a/kedro/extras/extensions/ipython.py
+++ b/kedro/extras/extensions/ipython.py
@@ -34,8 +34,10 @@ def _find_kedro_project(current_dir):  # pragma: no cover
     return None
 
 
-def reload_kedro(path, env: str = None, extra_params: Dict[str, Any] = None):
-    """Line magic which reloads all Kedro default variables."""
+def reload_kedro(path="", env: str = None, extra_params: Dict[str, Any] = None):
+    """Line magic which reloads all Kedro default variables.
+    Setting the path will also make it default for subsequent calls.
+    """
 
     import kedro.config.default_logger  # noqa: F401 # pylint: disable=unused-import
     from kedro.framework.cli import load_entry_points
@@ -43,6 +45,14 @@ def reload_kedro(path, env: str = None, extra_params: Dict[str, Any] = None):
     from kedro.framework.session import KedroSession
     from kedro.framework.session.session import _activate_session
     from kedro.framework.startup import bootstrap_project
+
+    # If a path is provided, set it as default for subsequent calls
+    global project_path
+    if path:
+        project_path = Path(path).expanduser().resolve()
+        logging.info("Updated path to Kedro project: %s", str(project_path))
+    else:
+        logging.info("No path argument was provided. Using: %s", str(project_path))
 
     path = path or project_path
     metadata = bootstrap_project(path)
@@ -77,25 +87,12 @@ def reload_kedro(path, env: str = None, extra_params: Dict[str, Any] = None):
         logging.info("Registered line magic `%s`", line_magic.__name__)  # type: ignore
 
 
-def init_kedro(path=""):
-    """Line magic to set path to Kedro project.
-    `%reload_kedro` will default to this location.
-    """
-    global project_path
-    if path:
-        project_path = Path(path).expanduser().resolve()
-        logging.info("Updated path to Kedro project: %s", str(project_path))
-    else:
-        logging.info("No path argument was provided. Using: %s", str(project_path))
-
-
 def load_ipython_extension(ipython):
     """Main entry point when %load_ext is executed"""
 
     global project_path
     global startup_path  # pylint:disable=global-variable-not-assigned
 
-    ipython.register_magic_function(init_kedro, "line")
     ipython.register_magic_function(reload_kedro, "line", "reload_kedro")
 
     project_path = _find_kedro_project(startup_path)
@@ -105,7 +102,4 @@ def load_ipython_extension(ipython):
     except (ImportError, ModuleNotFoundError):
         logging.error("Kedro appears not to be installed in your current environment.")
     except Exception:  # pylint: disable=broad-except
-        logging.warning(
-            "Kedro extension was registered. Make sure you pass the project path to "
-            "`%reload_kedro` or set it using `%init_kedro`."
-        )
+        logging.warning("Kedro extension was registered.")
--- a/tests/extras/extensions/test_ipython.py
+++ b/tests/extras/extensions/test_ipython.py
@@ -1,11 +1,7 @@
 # pylint: disable=import-outside-toplevel,reimported
 import pytest
 
-from kedro.extras.extensions.ipython import (
-    init_kedro,
-    load_ipython_extension,
-    reload_kedro,
-)
+from kedro.extras.extensions.ipython import load_ipython_extension, reload_kedro
 from kedro.framework.session.session import _deactivate_session
 from kedro.framework.startup import ProjectMetadata
 from kedro.pipeline import Pipeline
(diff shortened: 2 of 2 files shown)


kedro/extras/extensions/ipython.py
```diff
@@ -44,6 +46,14 @@ def reload_kedro(path, env: str = None, extra_params: Dict[str, Any] = None):
...
 
+    # If a path is provided, set it as default for subsequent calls
+    global project_path
+    if path:
+        project_path = Path(path).expanduser().resolve()
+        logging.info("Updated path to Kedro project: %s", str(project_path))
+    else:
+        logging.info("No path argument was provided. Using: %s", str(project_path))
+
     path = path or project_path
```
AntonyMilneQB, line 57: I don't see the need for this line any more - should we remove it and just use `project_path` in all following code?

kedro/extras/extensions/ipython.py
```diff
@@ -105,7 +102,4 @@ def load_ipython_extension(ipython):
     except (ImportError, ModuleNotFoundError):
         logging.error("Kedro appears not to be installed in your current environment.")
     except Exception:  # pylint: disable=broad-except
-        logging.warning(
-            "Kedro extension was registered. Make sure you pass the project path to "
-            "`%reload_kedro` or set it using `%init_kedro`."
-        )
+        logging.warning("Kedro extension was registered.")
```
lorenabalan, line 105:
```suggestion
        logging.warning("Kedro extension was registered, but couldn't find a Kedro project. Make sure you run `%reload_kedro <path_to_kedro_project>`. ")
```

kedro/extras/extensions/ipython.py
```diff
@@ -77,25 +87,12 @@ def reload_kedro(path, env: str = None, extra_params: Dict[str, Any] = None):
...
-        logging.info("Updated path to Kedro project: %s", str(project_path))
-    else:
-        logging.info("No path argument was provided. Using: %s", str(project_path))
-
-
 def load_ipython_extension(ipython):
     """Main entry point when %load_ext is executed"""
 
     global project_path
     global startup_path  # pylint:disable=global-variable-not-assigned
```
lorenabalan, line 94: Can we get rid of this global?

MerelTheisenQB (approved):
This looks great! Nice that we got rid of one of the globals as well 👍

kedro/extras/extensions/ipython.py
```diff
@@ -77,35 +89,21 @@ def reload_kedro(path, env: str = None, extra_params: Dict[str, Any] = None):
...
 
-    global project_path
-    global startup_path  # pylint:disable=global-variable-not-assigned
+    global default_project_path
 
-    ipython.register_magic_function(init_kedro, "line")
     ipython.register_magic_function(reload_kedro, "line", "reload_kedro")
 
-    project_path = _find_kedro_project(startup_path)
+    default_project_path = _find_kedro_project(Path.cwd())
```
AntonyMilneQB, line 99: **See following comment no error handling before this. If you agree with what I say there then this one is redundant...** 

Minor: If `_find_kedro_project` can't find a kedro project and returns `None` then this sets the `default_project_path` to `None` and will give the message "No path argument was provided. Using: None".

To me it would just feel a bit more natural if we used the `None` value to specifically mean "`reload_kedro` should re-use the current `default_project_path`" and instead just passed `Path.cwd()` in here instead. It's the default value of `default_project_path` set on L14 so seems a bit odd to change it to `None` rather than just keeping that default.

```suggestion
    kedro_project path = _find_kedro_project(Path.cwd())
    default_project_path = kedro_project_path or Path.cwd()
```

kedro/extras/extensions/ipython.py
```diff
@@ -44,17 +47,26 @@ def reload_kedro(path, env: str = None, extra_params: Dict[str, Any] = None):
...
-    path = path or project_path
-    metadata = bootstrap_project(path)
+    # If a path is provided, set it as default for subsequent calls
+    global default_project_path
+    if path:
+        default_project_path = Path(path).expanduser().resolve()
+        logging.info("Updated path to Kedro project: %s", str(default_project_path))
+    else:
+        logging.info(
+            "No path argument was provided. Using: %s", str(default_project_path)
```
AntonyMilneQB, line 57: Very minor: I suspect these `str` calls could be removed as the logging `%s` would coerce to string anyway (done in several places throughout file).

kedro/extras/extensions/ipython.py
```diff
@@ -77,35 +89,21 @@ def reload_kedro(path, env: str = None, extra_params: Dict[str, Any] = None):
...
+        reload_kedro(default_project_path)
     except (ImportError, ModuleNotFoundError):
         logging.error("Kedro appears not to be installed in your current environment.")
     except Exception:  # pylint: disable=broad-except
         logging.warning(
-            "Kedro extension was registered. Make sure you pass the project path to "
-            "`%reload_kedro` or set it using `%init_kedro`."
+            "Kedro extension was registered but couldn't find a Kedro project. "
+            "Make sure you run `%reload_kedro <path_to_kedro_project>`."
         )
```
AntonyMilneQB, line 109: Minor: maybe this made sense before but it feels like a weird way of handling the flow now. If I try loading the extension and don't have kedro installed then I'll get "ModuleNotFoundError: No module named 'kedro.extras'" which is pretty clear already so I don't see the need to catch the `ImportError` at all here? And then putting everything else in a broad except and assuming that's because a kedro project can't be found seems odd given we have a special value `None` that signifies exactly that already.

To me this would make more sense (combined it with my above suggestion and tidied the result up a bit) as something like this:
```
kedro_project_path = _find_kedro_project(Path.cwd())
if kedro_project_path is not None:
    reload_kedro(kedro_project_path)
else:
    logging.warning(
            "Kedro extension was registered but couldn't find a Kedro project. "
            "Make sure you run `%reload_kedro <path_to_kedro_project>`."
        )
```

AntonyMilneQB (approved):
Looks great! ⭐  

I've left some minor comments which I know is on stuff outside the scope of this ticket, so please feel totally free to not implement them and merge this without changing anything. It would be useful though if you could say if you agree with the comments or not - if you're happy with the changes but don't want to make them here I can stick them in my next PR touching ipython stuff instead.

kedro/extras/extensions/ipython.py
lorenabalan: If I read this correctly, then it's not actually true. `None` is returned if it finds a kedro project, not if kedro is installed. The import and module errors can still be raised if there IS a kedro project, but kedro is not installed (wrong activated environment for instance)  in which  case we give a more friendly errr message to the user.

kedro/extras/extensions/ipython.py
lorenabalan: Yeah that's actually a fair point, I'll leave it to use cwd over None.

kedro/extras/extensions/ipython.py
lorenabalan: Basically if someone has the kedro ipython extension loaded (e.g.set in the ipython config), it *shouldn't* prevent them from using ipython at all. That's why we catch the exception and move on, but give them a heads up.

lorenabalan: In the spirit of [this principle](https://mtlynch.io/human-code-reviews-2/#aim-to-bring-the-code-up-a-letter-grade-or-two), merging this as is as it's already gone a little out of scope, and final perfecting/refactoring can be done under a different task, as that will also require updating unit tests..

PR #1355, opened 2022-03-18
Remove .ipython dir and introduce kedro Jupyter kernel

### LINKED ISSUE #1338: Make `kedro ipython/jupyter` commands not use the project's `.ipython/` dir
Currently Kedro is using a project specific `.ipython/` in order to preload the Kedro IPython extension when running `kedro ipython` and `kedro jupyter notebook/lab`. This is a hack we've been using to simplify the user experience with Jupyter and IPython, but it's not a widely used practice and often causes troubles.

One of the most common problems with this approach is that the user needs to run the Jupyter instances themselves, which is not applicable in environments with managed notebook instances like SageMaker and Databricks.

In order to pave the way to a state where there's very little distinction between local and managed Jupyter instances, we need to modify our commands to do as follows:

- `kedro ipython` should be changed to simply be an alias of `ipython --ext="kedro.extras.extensions.ipython"`
- `kedro jupyter lab/notebook` should take advantage of Jupyter Kernels
    - it should install a new Jupyter Kernel (if it doesn't exist) named after the value of `from kedro.framework.project import PACKAGE_NAME` with display name `Kedro (<PACKAGE_NAME>)`
    - the new kernel should load the `kedro.extras.extensions.ipython` IPython extension on start
    - the new kernel should use the Kedro icon
    - it should run Jupyter by calling `jupyter notebook --MultiKernelManager.default_kernel_name=<PACKAGE_NAME>`
- make `kedro ipython` and `kedro jupyter` print out a nice error message if we don't have `jupyter_client`, `jupyter` or `ipython` installed (and remove the dependency of Kedro on `jupyter_client`)
- remove all `.ipython/` folders in our official starters
- update all our docs mentioning `.ipython/`

AntonyMilneQB:
## Note for reviewers
The diff here isn't as bad as the number of files changed makes it look! Much of it is just deleting old files. The interesting files to review are just:
- kedro/framework/cli/jupyter.py
- kedro/framework/cli/project.py

The docs have changed substantially though - best way to review that is not to look at the .md file but instead to [check the rendered version](https://antonymilneqb.github.io/tools_integration/ipython.html).

## Description

Closes #1338. In brief, this changes `kedro jupyter notebook/lab` and `kedro ipython` commands so that instead of using the `.ipython` folder, they now use the `kedro.extras.extensions.ipython` extension instead. This ensures that there is one consistent way of using kedro interactively.

`kedro ipython` is basically equivalent to `ipython --ext=kedro.extras.extensions.ipython`. `kedro jupyter notebook/lab` achieve the same thing (i.e. automatically loading the extension) by creating a custom Jupyter kernel which points to an ipython launcher that includes `--ext=kedro.extras.extensions.ipython`.

There's no kedro command to remove the kernel; you just do `jupyter kernelspec remove kernel_name`. Note that all kernels created by kedro have the `kedro_` prefix so you can immediately see where they come from if you do `jupyter kernelspec list`. This stuff will go in the docs when I write them.

One advantage of a custom kernel is we now have a snazzy icon on the UI (update! The icon is now the black version as in the docs but I didn't change these screenshots)

**Jupyter notebook**
image

**Jupyter lab**
image

## Development notes
I've needed to make lots of small decisions here exactly what to change and how big a breaking change this should be. Overall I've tried to balance up the opportunity to do a breaking release + wanting to achieve one consistent workflow + ensure that stuff still works + not being overly zealous in removing things that we're not yet sure should be removed. I'll make a separate ticket for items which could be revisited in future. Please let me know if you disagree with any of these decisions since they're very much open for discussion.

### Things I've changed here
* Removed `ipython_message` message informing users which variables are in scope. This seemed to be very out of date and no longer needed given the very similar message printed by `load_ipython_extension`
* Removed `ip` argument for `kedro jupyter notebook/lab`. This actually won't affect anything because you can still do `kedro jupyter notebook/lab --ip=xxx` and it will get passed through `*args`
* Removed `idle_timeout` argument for `kedro jupyter notebook/lab`. This was a tricky one to decide on because it was [originally introduced for a good reason](https://github.com/quantumblacklabs/private-kedro/pull/418): to fix a bug in which the [`%run_viz` line magic creates a process that blocks port 4141 even after Jupyter is killed](https://jira.quantumblack.com/browse/KED-1355). However, (a) the fix does not seem to work for me [i.e. `%run_viz` still blocks port 4141 even with the fix], (b) this fix can't ever work on managed jupyter instances so is incompatible with our goal of "one consistent workflow", (c) does anyone use `%run_viz` anyway? So overall I decided to remove the fix and then later let's work out whether it's really an issue and if so come up with a different fix (probably on the viz side, a `%kill_viz` command or something).
* Removed the `SingleKernelSpecManager` that ensured the user only had one kernel to choose from. This is incompatible with the one consistent workflow goal and seemed unnecessary now that we are clearly labelling the Kedro kernel and making it the default. If we do want to reintroduce this behaviour then it might now be easier to do through `KernelSpecManager.allowed_kernelspecs` rather than making a whole custom kernel spec manager
* Removed `all_kernels` option which gave the user the possibility of overriding the `SingleKernelSpecManager` to show all available kernels (since this is now how it always behaves)

### Things I've not changed here
* `KEDRO_ENV` environment variable remains. This [provides a mechanism for changing `env` and is picked up when the session is created](https://github.com/kedro-org/kedro/blob/main/kedro/framework/session/session.py#L186). As far as I can see, the interactive workflow doesn't strictly need this since a user can do `%reload_kedro env=xxx` to change environment. However, since removing this would in theory affect the creation of sessions outside the interactive workflow also I left it as is. I think we should check with users whether and how it's used and then hopefully remove it in future.
* Calls to `validate_settings` remain in `kedro jupyter notebook/labs`. I don't know if these are still needed but I didn't want to risk taking them out and breaking stuff. If they are still needed then maybe these calls should move to `kedro.extensions.extras.ipython` in the future

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1355"></a>


kedro/framework/cli/jupyter.py
```diff
@@ -247,46 +238,6 @@ def convert_notebook(
...
-    """Build the environment dictionary that gets injected into the subprocess running
-    Jupyter. Since the subprocess has access only to the environment variables passed
-    in, we need to copy the current environment and add ``KEDRO_ENV``.
-    """
-    if not kedro_env:
-        return {}
-    jupyter_env = os.environ.copy()
-    jupyter_env["KEDRO_ENV"] = kedro_env
-    return {"env": jupyter_env}
-
```
AntonyMilneQB, line 288: This is now done just as `os.environ["KEDRO_ENV"] = env`, the same way it's always been done in `kedro ipython`. I don't think the docstring here was right actually - `subprocess.run` will inherit the current environment's variables. And there's no need to worry about contaminating the environment, since adding something to `os.environ` is not persisted after the session is closed.

kedro/framework/cli/jupyter.py
```diff
@@ -90,87 +47,123 @@ def jupyter():
...
+    if env:
+        os.environ["KEDRO_ENV"] = env
+
+    python_call(
+        "jupyter",
+        ("lab", f"--MultiKernelManager.default_kernel_name={kernel_name}") + args,
+    )
+
+
+def _create_kernel(kernel_name: str, display_name: str):
```
AntonyMilneQB, line 112: I spent a loooong time reading through the ipython and jupyter source to try and figure out which were the right high-level API commands to use here to make this as robust as possible so we're not relying on too much stuff that might change. So🤞  this carries on working in the future, but I decided to the wrap the whole call in `try/except` just in case...

kedro/framework/cli/jupyter.py
```diff
@@ -90,87 +47,123 @@ def jupyter():
...
-    arguments = _build_jupyter_command(
-        "lab",
-        ip_address=ip_address,
-        all_kernels=all_kernels,
-        args=args,
-        idle_timeout=idle_timeout,
-        project_name=metadata.project_name,
+    kernel_name = f"kedro_{metadata.package_name}"
+    try:
+        _create_kernel(kernel_name, f"Kedro ({metadata.package_name})")
```
AntonyMilneQB, line 97: N.B. there's some rules on what characters are allowed in a kernel name, but something that's a valid Python package name should always be a valid.

datajoely:
Nice! Would just add I think the `KEDRO_ENV` env var has a lot of value outside of the interactive workflow

kedro/templates/project/{{ cookiecutter.repo_name }}/src/requirements.txt
```diff
@@ -1,6 +1,5 @@
 black==21.5b1
 flake8>=3.7.9, <4.0
-ipython~=7.10
```
AntonyMilneQB, line 3: This was just left over from a bad merge - the specification on the next line is the right one.

requirements.txt
```diff
@@ -6,7 +6,6 @@ dynaconf>=3.1.2,<4.0.0
 fsspec>=2021.4, <=2022.1
 gitpython~=3.0
 jmespath>=0.9.5, <1.0
-jupyter_client>=5.1, <8.0
```
AntonyMilneQB, line 9: This is already in the project template requirements, which is where it really belongs.

kedro/framework/cli/jupyter.py
```diff
@@ -90,87 +48,125 @@ def jupyter():
...
+    if env:
+        os.environ["KEDRO_ENV"] = env
+
+    python_call(
+        "jupyter",
+        ("lab", f"--MultiKernelManager.default_kernel_name={kernel_name}") + args,
+    )
+
+
+def _create_kernel(kernel_name: str, display_name: str) -> None:
```
AntonyMilneQB, line 113: This is the bit that might eventually become its own `kedro jupyter-init` or something like that.

I spent a loooong time reading through the ipython and jupyter source to try and figure out which were the right high-level API commands to use here to make this as robust as possible so we're not relying on too much stuff that might change. So🤞  this carries on working in the future, but I decided to the wrap the whole call in `try/except` just in case...

idanov (commented):
It's pretty nice to see this change happening finally 🎉 I think we need to get rid of all references to the `.ipython/` directory in our codebase and docs, as well as in all our official starters.

kedro/framework/cli/jupyter.py
```diff
@@ -90,87 +48,125 @@ def jupyter():
...
+    """Creates an ipython kernel for the kedro project, if one does not already exist.
+
+    Installs the default ipython kernel (which points towards `sys.executable`)
```
lorenabalan, line 116:
```suggestion
    Installs the default IPython kernel (which points towards `sys.executable`)
```

kedro/framework/cli/jupyter.py
```diff
@@ -90,87 +48,125 @@ def jupyter():
...
+    and customises it to make the launch command load the kedro extension.
+    This is equivalent to the method recommended for creating a custom ipython kernel
```
lorenabalan, line 118:
```suggestion
    This is equivalent to the method recommended for creating a custom IPython kernel
```

kedro/framework/cli/jupyter.py
```diff
@@ -90,87 +48,125 @@ def jupyter():
...
+    on the CLI: https://ipython.readthedocs.io/en/stable/install/kernel_install.html.
+
+    On linux this creates a directory ~/.local/share/jupyter/kernels/{kernel_name}
```
lorenabalan, line 121:
```suggestion
    On Linux this creates a directory ~/.local/share/jupyter/kernels/{kernel_name}
```

kedro/framework/cli/jupyter.py
lorenabalan: Maybe it was true for older versions of Python, but it's still worth testing that KEDRO_ENV is definitely loading up the config from the correct environment.

kedro/framework/cli/jupyter.py
```diff
@@ -3,79 +3,37 @@
...
 CONVERT_ALL_HELP = """Extract the nodes from all notebooks in the Kedro project directory,
 including sub-folders."""
 
 OVERWRITE_HELP = """If Python file already exists for the equivalent notebook,
 overwrite its contents."""
 
 
-def collect_line_magic():
-    """Interface function for collecting line magic functions from plugin entry points."""
-    return load_entry_points("line_magic")
```
lorenabalan, line 47: What is this going to be replaced by? How would stuff like `%run_viz` work?

lorenabalan (commented):
LGTM, but a little worried about the line magic not working.

kedro/framework/cli/jupyter.py
AntonyMilneQB: Yeah, I checked this already manually and all seemed to work ok. Good point about older versions of Python though  - I will double check with another version.

kedro/framework/cli/jupyter.py
```diff
@@ -90,87 +48,125 @@ def jupyter():
...
-        ip_address=ip_address,
-        all_kernels=all_kernels,
-        args=args,
-        idle_timeout=idle_timeout,
-        project_name=metadata.project_name,
-    )
+    kernel_name = f"kedro_{metadata.package_name}"
+    try:
+        _create_kernel(kernel_name, f"Kedro ({metadata.package_name})")
+    except Exception as exc:
```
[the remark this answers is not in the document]
AntonyMilneQB: Hmm yes, I don't see any drawback to moving it inside. It just looked a bit weird to put a whole function in a `try`/`except` like that but I don't really see any reason not to.

kedro/framework/cli/jupyter.py
AntonyMilneQB: This wasn't actually used anywhere and I think has just been overlooked for deletion before. It's done here instead now: https://github.com/kedro-org/kedro/blob/main/kedro/extras/extensions/ipython.py#L85-L87.

kedro/framework/cli/jupyter.py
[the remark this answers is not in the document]
AntonyMilneQB: Done in e88bd825.

kedro/templates/project/{{ cookiecutter.repo_name }}/src/requirements.txt
```diff
@@ -1,6 +1,5 @@
...
 ipython>=7.31.1, <8.0
```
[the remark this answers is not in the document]
AntonyMilneQB: We could, but ipython 8 requires Python >= 3.8 and we still support 3.7. Happy to bump it though if you like and let `pip` work out which version to install for the user's given version of Python. Or maybe even just remove it from the requirements given it's a core dependency of `jupyter` and `jupyterlab` anyway?

kedro/framework/cli/jupyter.py
```diff
@@ -90,87 +48,125 @@ def jupyter():
...
+        display_name=display_name,
     )
 
-    python_call_kwargs = _build_jupyter_env(env)
-    python_call("jupyter", arguments, **python_call_kwargs)
+    kernel_json = Path(kernel_path) / "kernel.json"
+    kernel_spec = json.loads(kernel_json.read_text())
+    kernel_spec["argv"].extend(["--ext", "kedro.extras.extensions.ipython"])
+    # indent=1 is to match the default ipykernel style (see ipykernel.write_kernel_spec).
+    kernel_json.write_text(json.dumps(kernel_spec, indent=1))
```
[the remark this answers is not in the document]
AntonyMilneQB: Not while using pathlib (which is what we generally use now), because `json.dump` needs a file pointer rather than a path object.

AntonyMilneQB:
OK, we're good to go here! Updating the docs turned into quite a substantial rewrite because a lot of the page was out of date. If you want to review the new docs the easiest way is to look at the rendered version here: https://antonymilneqb.github.io/tools_integration/ipython.html

Still to do:
- [ ] update release notes
- [ ] make new tickets out of some small issues discovered here
- [ ] update kedro-starters, but once this PR has merged

features/steps/test_starter/{{ cookiecutter.repo_name }}/src/requirements.txt
```diff
@@ -3,7 +3,6 @@ flake8>=3.7.9, <4.0
 ipython>=7.31.1, <8.0
 isort~=5.0
 jupyter~=1.0
-jupyter_client>=5.1, <8.0
```
AntonyMilneQB, line 6: This is a core dependence of `jupyter` so I've decided to not list it explicitly so we don't have to keep bumping the version requirements.

requirements.txt
AntonyMilneQB: As noted elsewhere I've actually removed `jupyter_client` altogether:
> This is a core dependence of jupyter so I've decided to not list it explicitly so we don't have to keep bumping the version requirements.

tests/framework/cli/test_jupyter.py
```diff
@@ -22,209 +22,147 @@ def mocked_logging(mocker):
...
+@pytest.fixture
+def cleanup_kernel():
+    yield
+    if "my_kernel_name" in find_kernel_specs():
+        KernelSpecManager().remove_kernel_spec("my_kernel_name")
+
+
+@pytest.mark.usefixtures("cleanup_kernel")
+class TestCreateKernel:
+    def test_create_new_kernel(self):
```
AntonyMilneQB, line 144: I decided to do this as a sort of e2e test that checks what the actual result of calling `_create_kernel` is rather than mocking out all those calls. This means its not tied to a particular implementation and will be easy to modify the tests if and when this becomes its own `kedro jupyter-init` command.

docs/source/tools_integration/ipython.md
```diff
@@ -1,173 +1,168 @@
...
+3             4.6          3.1           1.5          0.2     setosa
+4             5.0          3.6           1.4          0.2     setosa
+..            ...          ...           ...          ...        ...
+145           6.7          3.0           5.2          2.3  virginica
+146           6.3          2.5           5.0          1.9  virginica
+147           6.5          3.0           5.2          2.0  virginica
+148           6.2          3.4           5.4          2.3  virginica
+149           5.9          3.0           5.1          1.8  virginica
+
+In [3]: catalog.load("parameters")
```
yetudada, line 87: Is this the only way that users can access parameters now? 🥴 This looks confusing that you would be using `catalog.load` to find parameters.

docs/source/tools_integration/ipython.md
```diff
@@ -1,173 +1,168 @@
...
-
-```python
-parameters = context.params  # type: Dict
-parameters["example_test_data_ratio"]
-# returns the value of 'example_test_data_ratio' key from 'conf/base/parameters.yml'
-```
-
-```eval_rst
-.. note::  You need to reload Kedro variables by calling `%reload_kedro` and re-run the code snippet above if you change the contents of ``parameters.yml``.
+```ipython
```
yetudada, line 131: 👏

yetudada (approved):
This is great! I left minor comments 🚀

kedro/framework/cli/jupyter.py
```diff
@@ -90,87 +46,128 @@ def jupyter():
...
     env,
-    idle_timeout,
     args,
     **kwargs,
-):  # pylint: disable=unused-argument,too-many-arguments
+):  # pylint: disable=unused-argument
     """Open Jupyter Notebook with project specific variables loaded."""
-    _check_module_importable("jupyter_core")
-
+    _check_module_importable("notebook")
```
MerelTheisenQB, line 58: It makes a lot more sense to me that this is now `notebook`, but I'm just wondering why was it `jupyter_core` before?

docs/source/tools_integration/ipython.md
```diff
@@ -1,173 +1,168 @@
-# Use Kedro with IPython and Jupyter Notebooks/Lab
-
-This section follows the [Iris dataset example](../get_started/example_project.md) and demonstrates how to use Kedro with IPython and Jupyter Notebooks / Lab. We also recommend a video that explains the transition from the use of vanilla Jupyter Notebooks to using Kedro, from [Data Engineer One](https://www.youtube.com/watch?v=dRnCovp1GRQ&t=50s&ab_channel=DataEngineerOne).
+# Use Kedro with IPython and Jupyter
 
+This page demonstrates how to use Kedro with IPython, Jupyter Notebook and JupyterLab. We also recommend [a video by Data Engineer One](https://www.youtube.com/watch?v=dRnCovp1GRQ&t=50s&ab_channel=DataEngineerOne) that explains the transition from the use of vanilla Jupyter notebooks to Kedro.
```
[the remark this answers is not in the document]
lorenabalan: @yetudada has also done a great talk (pycon maybe?) about moving from a notebook project to a kedro project, which might also be relevant here.

docs/source/tools_integration/ipython.md
```diff
@@ -1,173 +1,168 @@
...
 
 <iframe width="560" height="315" style="max-width: 100%" src="https://www.youtube.com/embed/dRnCovp1GRQ" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
 
-## Why use a Notebook?
-There are reasons why you may want to use a Notebook, although in general, the principles behind Kedro would discourage their use because they have some [drawbacks when they are used to create production or reproducible code](https://towardsdatascience.com/5-reasons-why-you-should-switch-from-jupyter-notebook-to-scripts-cb3535ba9c95). However, there are occasions when you'd want to put some code into a Notebook, for example:
+## Why use a notebook?
```
lorenabalan, line 7: I think we were using capitalised words on purpose to distinguish from everyday objects, and I tend to agree. At least that's the vague memory I have from when Jo was working with us.

docs/source/tools_integration/ipython.md
lorenabalan: I think you can also do `context.params`

lorenabalan (approved):
Some small observations but LGTM!

MerelTheisenQB (approved):
Forgot to approve earlier, my question shouldn't be a blocker. Great work 👍

PR #1364, opened 2022-03-23
Make it easier for people to run Kedro projects by simply importing them

idanov:
## Description

In many deployment scenarios users want to be able to easily run their Kedro projects by simply importing them. After the change, here's how a packaged Kedro project can be run:

```
from my_project.__main__ import main

main(['--pipeline', 'my_pipeline']) # or just main() if default pipeline is needed
```

## Development notes

The change is only in the template and need to be done in all other official templates accordingly. This is just a minor change before 0.18, but future releases should hide all the custom logic for finding the run from the user.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1364"></a>


MerelTheisenQB (approved):
Great change! 👍 

I think we should add a short explanation in the docs about how users can then run the project, like you added here in the PR description. But that could also be part of a follow up PR, as long as we have it somewhere on the backlog 🙂

lorenabalan (approved):
Nice one! Agree with Merel that maybe some docs would be nice. Don't forget to follow-up with a PR against starters, preferable against `0.18.0` branch/PR.

PR #1328, opened 2022-03-08
Refactor micropkg and remove pipeline pull and pipeline package

SajidAlamQB:
## Description

Following https://github.com/kedro-org/kedro/pull/1224 we need to refactor some variable names and functions so that they don't talk about pipelines but more generally about micropackages. We also need to remove `pipeline package` and `pipeline pull` and their tests as part of this issue.

## Development notes
- Removed `pipeline package` and `pipeline pull` commands and their tests
- Renamed mentions of `pipeline` to `micropkg` in the docs/notes/code

## Checklist

- [x] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [x] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [x] Updated the documentation to reflect the code changes
- [x] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [x] Added tests to cover my changes

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1328"></a>

PR #1368, opened 2022-03-24
Fix windows 3.10 builds

MerelTheisenQB:
## Description
The windows unit test and pip compile build for python 3.10 are failing on https://github.com/kedro-org/kedro/pull/1328

## Development notes
There's no error on the build, just `exit code 1` and I couldn't ssh into the box. I just tried removing the `conda install` step before the `pip install` step and this seems to solve the issue. I'm a bit suspicious because this fix seems a bit too good to be true 🤔  It does clean up our windows build steps. I'll keep an eye out on the builds, but for now let's merge this in if we don't have any other solutions.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1368"></a>

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -138,10 +138,10 @@ commands:
     steps:
       - restore_cache:
           name: Restore package cache
-          key: kedro-deps-v1-win-{{ checksum "requirements.txt" }}-{{ checksum "test_requirements.txt" }}
+          key: kedro-deps-Merel-win-{{ checksum "requirements.txt" }}-{{ checksum "test_requirements.txt" }}
       - restore_cache:
           name: Restore conda environment cache
-          key: kedro-deps-v1-win-<<parameters.python_version>>-{{ checksum "requirements.txt" }}-{{ checksum "test_requirements.txt" }}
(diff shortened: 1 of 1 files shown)


noklam (approved):

AntonyMilneQB:
Hmmm, this would indeed be a nice fix but I'm also a bit suspicious that it's too good to be true. Since the cache key won't have changed here I think it's unfortunately very possible that it's still effectively got those packages `conda install`ed and actually removing those lines didn't change anything.

Since there's no way of clearing cache on CircleCI, to check this you need to run after changing
```
key: kedro-deps-v1-win-<<parameters.python_version>>-{{ checksum "requirements.txt" }}-{{ checksum "test_requirements.txt" }}
```
to 
```
key: kedro-deps-Merel-win-<<parameters.python_version>>-{{ checksum "requirements.txt" }}-{{ checksum "test_requirements.txt" }}
```
everywhere it appears. If that still works then great 😀

MerelTheisenQB:
@AntonyMilneQB great point, let me try and see what happens! I've tried to re-run the build without changes to see if it was a glitch, but that doesn't help..

MerelTheisenQB:
Closing this for now and doing some more investigation.

PR #1369, opened 2022-03-24
Fix jinja2 dependency for building documentation

noklam:

## Description

Jinja2 3.1.0 is breaking the dependencies with sphinx as it deprecated a few function, 3.0.3 is the last version that still works.

It breaks:
* build-doc
* e2etest 

## Development notes

Pin requirements in corresponding files.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1369"></a>


MerelTheisenQB (approved):
Thank you ⭐

noklam:
Notes:

Currently the sensible thing to do is to pin `jinja2` in `setup.py`, however, it doesn't play well, partly because it crashes with `kedro install` as it produce conflicting version. But even after that we still get build error when `pip install ".[docs]" and failing the e2e test.

AntonyMilneQB (approved):
Thanks for fixing this!

If the only failures are GeoJSON dataset on Windows then should be fine to merge I think. Weird that those have spread to versions other than Python 3.10 though...

PR #1392, opened 2022-03-30
Update deployment diagram to include Dask :merman:

### LINKED ISSUE #1321: Update deployment diagram to include Dask
Follow-up to https://github.com/kedro-org/kedro/pull/1248

We should probably make our deployment diagram in something like drawio and include the raw `.drawio` inside `docs/source/meta/images` so that it's easier to maintain and edit. Modern IDEs have extensions that even let you update in place (e.g. PyCharm definitely has one, VSCode might have one too). We've started doing the same for the architecture diagram already.

Bonus points: convert other diagrams to drawio if you find any that are also only included as static screenshot/png.
More bonus points: find diagram file (might be powerpoint) that Jo used to edit and remove it.

deepyaman:
## Description
Closes #1321 

Highly recommend merging #1489 first, so as to separate out Mermaid from MyST migration.

## Development notes
Throwing up an initial version for feedback; still need to figure out stuff like left-aligning text, rendering a list (or can just use asterisks as before), etc. if it's a requirement to look more like the original one.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1392"></a>


deepyaman:
Ah, will need to set https://github.com/mgaitan/sphinxcontrib-mermaid#markdown-support up...

Maybe somebody can review the diagram first, agree whether or not we want to go this way, and then let's think about accommodating in the build. Sphinx issues usually make my fun time less fun, so rather just do it once it's decided. :D

noklam:
I like this new diagram. :) I can't get the padding right, but it looks better to align the platform left.


```mermaid

flowchart TD
    A{Can your Kedro pipeline run on a single machine?} -- YES --> B[Consult the single-machine deployment guide];
    B --> C{Do you have Docker on your machine?};
    C -- YES --> D[Use a container-based approach];
    C -- NO --> E[Use the CLI or package mode];
    A -- NO --> F[Consult the distributed deployment guide];
    F --> G[What distributed platform are you using?\n\nCheck out the guides for:\n<li>Argo</li><li>Prefect</li><li>Kubeflow Pipelines</li><li>AWS Batch</li><li>Databricks</li><li>Dask</li></ul>]; 
    style G text-align:left
    H["Does (part of) your pipeline integrate with Amazon SageMaker?<br/><br/>Read the SageMaker integration guide"];

```

deepyaman:
> it looks better to align the platform left.

Agreed! How did you get it right, though? I can't get it to render left aligned/with bullets on GitHub the way you did, even trying a bunch of StackOverflow answers. 😅

noklam:
Ah, I thought Github would also show the raw markdown here. I don't know mermaid well so I also just did a random google/StackOverflow search to see which one works.

```

flowchart TD
    A{Can your Kedro pipeline run on a single machine?} -- YES --> B[Consult the single-machine deployment guide];
    B --> C{Do you have Docker on your machine?};
    C -- YES --> D[Use a container-based approach];
    C -- NO --> E[Use the CLI or package mode];
    A -- NO --> F[Consult the distributed deployment guide];
    F --> G[What distributed platform are you using?\n\nCheck out the guides for:\n<li>Argo</li><li>Prefect</li><li>Kubeflow Pipelines</li><li>AWS Batch</li><li>Databricks</li><li>Dask</li></ul>]; 
    style G text-align:left
    H["Does (part of) your pipeline integrate with Amazon SageMaker?<br/><br/>Read the SageMaker integration guide"];

```

deepyaman:
> If we can get this working in sphinx then I'd be very happy to change the diagram.

@AntonyMilneQB Thanks, this is what I was looking for! I'm happy to put in the effort to try and get it working in Sphinx as long as we're good with the diagram; just didn't want to if e.g. using Mermaid was out of the question. I'll try and do this as soon as I get a chance.

AntonyMilneQB:
Using Mermaid would definitely be great if we could get it working. If not then note we've also got at a drawio diagram for the Kedro architecture (which is then just manually saved to png I think) and some puml diagrams (in the draft folder but not actually rendered in the docs anywhere). Basically ease of future maintenance is priority here rather than using some particular system. Something which we write inline in the docs (looks like Mermaid would do this) rather than needing any external files sounds best to me if we can get it working.

noklam:
Hopefully it would work and we can have a diagram for hooks lifecycle later.

avan-sh:
@deepyaman noticed you're having to change to every `eval_rst` block. I remember reading changing it to `{eval_rst}` (braces) would be enough. Modifying every block might be a mammoth task

deepyaman:
> @deepyaman noticed you're having to change to every `eval_rst` block. I remember reading changing it to `{eval_rst}` (braces) would be enough. Modifying every block might be a mammoth task

I was able to script most of the changes for `note`s with a simple `perl` command, so wasn't too painful. :)

deepyaman:
Is there a good way to test whether the diagram renders properly? The HTML looks correct (has the `<div class="mermaid">), the right script is included, but `open docs/build/html/deployment/deployment_guide.html` still shows the text rendering. Here's the `div` block:

```html
 <div class="mermaid">
             flowchart TD
     A{Can your Kedro pipeline run on a single machine?} -- YES --&gt; B[Consult the single-machine deployment guide];
     B --&gt; C{Do you have Docker on your machine?};
     C -- YES --&gt; D[Use a container-based approach];
     C -- NO --&gt; E[Use the CLI or package mode];
     A -- NO --&gt; F[Consult the distributed deployment guide];
     F --&gt; G[&quot;What distributed platform are you using?&lt;br/&gt;&lt;br/&gt;Check out the guides for:&lt;br/&gt;&lt;br/&gt;&lt;li&gt;Argo&lt;/li&gt;&lt;li&gt;Prefect&lt;/ li&gt;&lt;li&gt;Kubeflow Pipelines&lt;/li&gt;&lt;li&gt;AWS Batch&lt;/li&gt;&lt;li&gt;Databricks&lt;/li&gt;&lt;li&gt;Dask&lt;/li&gt;&quot;];
     style G text-align:left
     H[&quot;Does (part of) your pipeline integrate with Amazon SageMaker?&lt;br/&gt;&lt;br/&gt;Read the SageMaker integration guide&quot;];
     style H text-align:left
         </div></div>
 </div>
```

Rendering:

image

I can get it to render properly with `mermaid_output_format = "png"` added to docs/conf.py, but I'd rather the HTML/JS version work, if possible. Anyway:

image

AntonyMilneQB:
I've spent quite a while looking at how to get the svg rendering and here's where things stand. Thanks very much to @tynandebold for helping look into this.

By default the sphinx-mermaid plugin points to the minified package https://unpkg.com/mermaid/dist/mermaid.min.js in the `script` tag ([relevant code in the plugin that handles it](https://github.com/mgaitan/sphinxcontrib-mermaid/blob/a6c1f8b9dc3e84cde8e2658856c0668dc47324db/sphinxcontrib/mermaid.py#L343)). If we change this to point to the unminified version then the svg renders correctly. This is done with the following lines in conf.py:
```
app.add_js_file("https://unpkg.com/mermaid/dist/mermaid.js")
mermaid_version = None
```
It is a complete mystery why it works with the unminified version but not the minified one.

Now the problem with using the unminified version is that it adds a relatively hefty javascript package to the docs page load, and from what I've read even these days with our superfast internets this is not really encouraged. In our case:
* minified version adds 282kB to the page load
* unminified version adds 763kB to the page load

For comparison, the docs page without mermaid at all is about 750kB, so the unminified version roughly doubles the page load. The consequence of this is a slower page loading time (but some stuff is done asynchronously so it's not as bad as doubling page load time). Personally another 700kB sounds ok to me, but the really annoying thing here is that the mermaid package will be loaded on every single docs page, even if there's no mermaid diagram there. See https://github.com/mgaitan/sphinxcontrib-mermaid/issues/80.

So our options are (in my order of preference, with 1st one being best):
1. Use the unminified mermaid package, take the page load hit on every docs page. Optional but would be nice: add support to or even better fix  https://github.com/mgaitan/sphinxcontrib-mermaid/issues/80 ourselves (I don't think it will be fixed any time soon by anyone else)
2. Render the diagram as a png instead. This would make it impossible to embed links, which would be a pity, and it's just less cool than an svg...
3. Use something other than mermaid, like puml, and hope the sphinx plugin works better

P.S. the whitespace issue is also noted on the sphinx-mermaid repo but I don't think will be fixed soon. It's a problem for both png and svg renderings. https://github.com/mgaitan/sphinxcontrib-mermaid/issues/82

deepyaman:
@AntonyMilneQB Thanks for the detailed writeup! I'm on board with the first option.

Re implementing the conditional behavior in Mermaid, https://github.com/sphinx-doc/sphinx/pull/8631/files looks like a good start (would add some `has_mermaid_diagram` condition in https://github.com/mgaitan/sphinxcontrib-mermaid/blob/0.7.1/sphinxcontrib/mermaid.py#L343).

docs/conf.py
```diff
@@ -517,6 +521,17 @@ def _add_jinja_filters(app):
...
+    app: Sphinx,
+    pagename: str,
+    templatename: str,
+    context: Dict,
+    doctree: Optional[nodes.document],
+) -> None:
+    if doctree and doctree.next_node(mermaid):
+        app.add_js_file("https://unpkg.com/mermaid/dist/mermaid.js")
+
+
```
AntonyMilneQB, line 534: Does this only add it to pages that have `{mermaid}`?! :o

deepyaman: Yep! But it still isn't working properly, and it's driving me insane...

I've also tried creating a local copy of the `mermaid.js` (or `mermaid.min.js`) file under `_static`, and it loads fine, but doesn't render. I think it does work if I just `app.add_js_file("js/mermaid.min.js")` without this conditional handler, and it adds to all the files.

I'm going to try reverting back to adding to all files, and then taking the approach of removing the script from all pages that don't have a mermaid node. My current hypothesis is that the rendering of the text that get's converted to a Mermaid diagram is messed up without this, but really not sure...

deepyaman:
@AntonyMilneQB In the end, the solution I landed on after much trials and tribulations is very similar to https://github.com/sphinx-doc/sphinx/issues/6241#issuecomment-705070598. With this solution, we've got the minified mermaid.js file, only on pages with Mermaid directives, and without any extraneous "None" scripts. Tested in GitPod (`make build-docs && cd docs/build/html/ && python -m http.server 8000`):

image

image

AntonyMilneQB:
Ahh amazing, congratulations!! 🎉 🥳 Thank you for persevering with this. That is the perfect solution. 

Do you know why it now works with the minified js? Were the `None` scripts somehow interfering with that but not the unminified one?

noklam:
Nice to see it working! When Mermaid is enabled we should definitely add a chart for the `hooks`.

AntonyMilneQB:
Thanks for the explanation! Some things will forever remain a mystery I suppose...

hugovk:
Hi! Thanks for figuring this out!

I've made an upstream PR, based on this, to only install the Mermaid JavaScript where needed.

Please could you have a look at https://github.com/mgaitan/sphinxcontrib-mermaid/pull/101 and test it out?

Thank you!

Issue #1401, opened 2022-04-01
Git fatal error when using a Kedro project inside directory without an initialised repo


oj-m:
## Description
Kedro Docs contains a [pandas Iris example project](https://kedro.readthedocs.io/en/stable/get_started/example_project.html) which has Python 3.6 in the requirements file, which does not execute on newer Apple M1 chipsets.  Attempting to execute it on an M1-compatible version of Python 3.8.13 via `kedro run` results in:
```sh
kedro.framework.session.store - INFO - `read()` not implemented for `BaseSessionStore`. Assuming empty store.
fatal: Needed a single revision
```

## Context
Attempt the docs instructions on https://kedro.readthedocs.io/en/stable/get_started/example_project.html on an Apple M1.

## Steps to Reproduce
Execute the following on an Apple M1:
```
kedro new --starter=pandas-iris
cd pandas-iris
git init
pip install -r src/requirements.txt
kedro run
```

## Expected Result
Run without `fatal` errors.

## Actual Result
```sh
kedro.framework.session.store - INFO - `read()` not implemented for `BaseSessionStore`. Assuming empty store.
fatal: Needed a single revision
```

## Your Environment
Include as many relevant details about the environment in which you experienced the bug:

* Kedro version used: `0.17.7`
* Python version used (`python -V`): `3.8.13`
* Operating system and version:  `macOS 12.3`

AntonyMilneQB:
Hi @oj-m and thanks for the bug report. Is there anything else in the error message at all? I suspect this is something to do with git. Is the directory you're running kedro in version controlled?

> Kedro Docs contains a [pandas Iris example project](https://kedro.readthedocs.io/en/stable/get_started/example_project.html) which has Python 3.6 in the requirements file
Please could you point out where the Python 3.6 requirement is? I'm a bit surprised to hear this. Kedro 0.17.7 should indeed work with 3.6 , 3.7 and 3.8.

datajoely:
I would also add - does this work if you do not init the repository? That error is from `git` not kedro? This stackoverflow suggests it may be related to [Homebrew](https://stackoverflow.com/a/65493486/2010808) as well.

avan-sh:
Hi @oj-m,
I tried to replicate the error on my M1 machine but couldn't get the error. This error might be related to some other package and not kedro related.

oj-m:
Ok, cleaned everything up and skipped `git init`, and received missing `jupyter_client` dependency:

```sh
[...]

PR #1422, opened 2022-04-07
handle git fatal error on non-git projects

avan-sh:
## Description

Addresses #1401 to remove the fatal error when a project is not a git repo or doesn't have the first commit.

## Development notes

Writes any message returned from subprocess call of git to STDOUT. Any error from the subprocess is not visible in the log out from kedro.

## Checklist

- [] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1422"></a>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -13,7 +13,7 @@
 
 ## Major features and improvements
 
-* Added more detail to YAML ParserError error message
+* Added more detail to YAML ParserError error message.
 
 ## Bug fixes and other changes
 * Removed fatal error from being logged when a Kedro session is created in a directory without git.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -13,6 +13,8 @@
 
 ## Major features and improvements
 
+* Added more detail to YAML ParserError error message
+
 ## Bug fixes and other changes
 * Removes fatal error from git in run log on directories without git
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -17,7 +17,7 @@
 
 ## Bug fixes and other changes
 * Removed fatal error from being logged when a Kedro session is created in a directory without git.
- 
+
 
 ## Upcoming deprecations for Kedro 0.19.0
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,23 @@
+# Upcoming Release 0.19.0
+
+## Major features and improvements
+
+## Bug fixes and other changes
+
+## Breaking changes to the API
+
+## Migration guide from Kedro 0.18.* to 0.19.*
+
+
+# Upcoming Release 0.18.1
+
+## Major features and improvements
+
+## Bug fixes and other changes
+
+## Upcoming deprecations for Kedro 0.19.0
+
+
 # Release 0.18.0
 
 ## TL;DR ✨
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -13,10 +13,10 @@
 
 ## Major features and improvements
 
+* Added more detail to YAML ParserError error message
 
 ## Bug fixes and other changes
 * Removed fatal error from being logged when a Kedro session is created in a directory without git.
-* Added more detail to YAML ParserError error message
  
 
 ## Upcoming deprecations for Kedro 0.19.0
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -13,10 +13,11 @@
 
 ## Major features and improvements
 
-* Added more detail to YAML ParserError error message.
 
 ## Bug fixes and other changes
 * Removed fatal error from being logged when a Kedro session is created in a directory without git.
+* Added more detail to YAML ParserError error message
+ 
 
 ## Upcoming deprecations for Kedro 0.19.0
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -14,6 +14,7 @@
 ## Major features and improvements
 
 ## Bug fixes and other changes
+* Removes fatal error from git in run log on directories without git
 
 ## Upcoming deprecations for Kedro 0.19.0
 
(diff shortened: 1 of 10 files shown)


avan-sh:
@datajoely, This required only a small change to the code and might not need checking for a git folder. LMK if this is enough or if it requires any tests. I couldn't find any existing tests at the moment for the function.

datajoely:
Let's mark it as ready for review and let the team look at it! @MerelTheisenQB do you have a view on what's needed to get this in?

avan-sh:
> @avan-sh can you check whether you are following the CircleCI project associated with your fork of the `kedro` project? This might be preventing your PR triggering builds in the kedro-org/kedro project.

This is strange, I'm already following the project and I also have some runs of the workflow from some old branches, but nothing recently.

Screenshot 2022-04-07 at 10 47 12 PM

MerelTheisenQB:
> > @avan-sh can you check whether you are following the CircleCI project associated with your fork of the `kedro` project? This might be preventing your PR triggering builds in the kedro-org/kedro project.
> 
> This is strange, I'm already following the project and I also have some runs of the workflow from some old branches, but nothing recently. I'm trying to trigger build from the recent branches now.
> 
> Screenshot 2022-04-07 at 10 47 12 PM

Ahh sorry, I wasn't clear! I actually meant that you'll have to **unfollow** your own fork, because that would prevent any of your PRs building on circleci for kedro-org/kedro.

avan-sh:
> Ahh sorry, I wasn't clear! I actually meant that you'll have to **unfollow** your own fork, because that would prevent any of your PRs building on circleci for kedro-org/kedro. 

Done

MerelTheisenQB:
I personally like this solution, because it's simple and cleans up the user logs from the confusing `fatal` messages. The only drawback is of course that you might miss a message that is actually relevant. But I can't really think if a reason why Kedro shouldn't work if something is up with git. 

Can you add this fix to the release notes?

datajoely:
@MerelTheisenQB am I right in saying we only do this to provide the git sha in experiment tracking?

MerelTheisenQB:
> @MerelTheisenQB am I right in saying we only do this to provide the git sha in experiment tracking?

We do this to store git related info such as the sha to the session store. This functionality was included straight away with the first proof of concept for the session store, but wasn't really used up until now to display it in viz with experiment tracking.

avan-sh:
> > @MerelTheisenQB am I right in saying we only do this to provide the git sha in experiment tracking?
> 
> We do this to store git related info such as the sha to the session store. This functionality was included straight away with the first proof of concept for the session store, but wasn't really used up until now to display it in viz with experiment tracking.

@MerelTheisenQB What are the uses of Session Store that were in mind during its design? 
Is there any case where a user can make use of this?

I've used it only once when we wanted to enable restarts on a pipeline that used a versioned dataset. 
One other place, I see this being useful is if we wanted to store the run status in a central location. This might require creating a new SessionStore class though.

MerelTheisenQB:
> @MerelTheisenQB What are the uses of Session Store that were in mind during its design? Is there any case where a user can make use of this?
> 
> I've used it only once when we wanted to enable restarts on a pipeline that used a versioned dataset. One other place, I see this being useful is if we wanted to store the run status in a central location. This might require creating a new SessionStore class though.

The `KedroSession` and store are still relatively new concepts in Kedro and we're still designing both. Currently, we mainly mean to use the store as a way to save the details from each run, so we can visualise it in `kedro-viz` or simply for investigative reasons. So far we've made a small POC implementation called the `ShelveStore` which persists data locally, and for the purpose of experiment tracking we've created the `SQLiteStore`. Mind you that for the time being the `SQLiteStore` is inside `kedro-viz` as it's still part of an alpha version of the feature. 

As a user you can leverage the session store for persisting details about your run, you can do this by using the `ShelveStore` or creating your own custom session store and configuring Kedro to use that in `settings.py`

avan-sh:
@AntonyMilneQB Can you let me know if the failed tests are anything to do with my changes. If I'm right, this is is related to another issue.

RELEASE.md
```diff
@@ -16,6 +16,7 @@
 * Added more detail to YAML ParserError error message
```
AntonyMilneQB, line 16:
```suggestion
* Added more detail to YAML ParserError error message.
```

RELEASE.md
```diff
@@ -16,6 +16,7 @@
...
 
 ## Bug fixes and other changes
+* Removes fatal error from git in run log on directories without git
```
AntonyMilneQB, line 19:
```suggestion
* Removed fatal error from being logged when a Kedro session is created in a directory without git.
```

AntonyMilneQB (approved):
LOVE this change. This annoying message has been bugging me for years, I just never got around to figuring out how to fix it. Thank you so much for doing this!

AntonyMilneQB:
While we're here, do you think we should be making a similar change on this line? It's currently not wrapped in any `try`/`except`, but maybe if the above git command ran then there's no way it can give a confusing error message.
```
res = subprocess.check_output(["git", "status", "--short"], cwd=project_path)
```

avan-sh:
> While we're here, do you think we should be making a similar change on this line? It's currently not wrapped in any `try`/`except`, but maybe if the above git command ran then there's no way it can give a confusing error message.
> 
> ```
> res = subprocess.check_output(["git", "status", "--short"], cwd=project_path)
> ```

I'll check on the behaviour of this later today and make any necessary changes.

avan-sh:
> While we're here, do you think we should be making a similar change on this line? It's currently not wrapped in any `try`/`except`, but maybe if the above git command ran then there's no way it can give a confusing error message.
> 
> ```
> res = subprocess.check_output(["git", "status", "--short"], cwd=project_path)
> ```

You're right that there's very little chance that we'll be reaching there. But to handle this situation, we could possibly move the second `git` call and the logic surrounding it into the first try/except block. Please suggest which way is preferred, I think moving the code might make it less confusing on why second git call is not being error handled.

The new `_describe_git` function would look as below:
```
def _describe_git(project_path: Path) -> Dict[str, Dict[str, Any]]:
    project_path = str(project_path)
    try:
        res = subprocess.check_output(["git", "rev-parse", "--short", "HEAD"],cwd=project_path,stderr=subprocess.STDOUT,)
        git_data = {"commit_sha": res.decode().strip()}  # type: Dict[str, Any]
        res = subprocess.check_output(["git", "status", "--short"],cwd=project_path,stderr=subprocess.STDOUT,)
        git_data["dirty"] = bool(res.decode().strip())

    # `subprocess.check_output()` raises `NotADirectoryError` on Windows
    except (subprocess.CalledProcessError, FileNotFoundError, NotADirectoryError):
        logging.getLogger(__name__).warning("Unable to git describe %s", project_path)
        return {}

    return {"git": git_data}
```

AntonyMilneQB:
Thanks for taking a look @avan-sh. Generally I don't like putting too much stuff in `try` but, as you suggest, here I think it's a bit clearer if we put it all in there. So go for it 👍

kedro/framework/session/session.py
```diff
@@ -49,21 +49,25 @@ def _deactivate_session() -> None:
...
-
     try:
         res = subprocess.check_output(
-            ["git", "rev-parse", "--short", "HEAD"], cwd=project_path
+            ["git", "rev-parse", "--short", "HEAD"],
+            cwd=project_path,
+            stderr=subprocess.STDOUT,
+        )
+        git_data = {"commit_sha": res.decode().strip()}  # type: Dict[str, Any]
+        res = subprocess.check_output(
```
MerelTheisenQB, line 59: Minor: but I would rename this perhaps to `git_status_res` just to avoid confusion with the variable above.
```suggestion
        git_status_res = subprocess.check_output(
```

kedro/framework/session/session.py
```diff
@@ -49,21 +49,25 @@ def _deactivate_session() -> None:
...
+            ["git", "status", "--short"],
+            cwd=project_path,
+            stderr=subprocess.STDOUT,
         )
+        git_data["dirty"] = bool(res.decode().strip())
```
MerelTheisenQB, line 64:
```suggestion
        git_data["dirty"] = bool(git_status_res.decode().strip())
```

MerelTheisenQB (approved):
LGTM! Thanks for tackling both git calls as part of this 🙂

Issue #1335, opened 2022-03-09
Technical design decision record for `KedroSession` 


merelcht:
## The `KedroSession` ✨
The `KedroSession` is the object responsible for managing the lifecycle of a Kedro run. It has two main functions:

1. Run execution: It makes sure that all core components needed by Kedro to execute a run are instantiated and the run is executed properly
2. Persisting run data: `KedroSession` offers a way to persist run data through the session store. The following data gets saved in the session store:
- package_name
- project_path
- session_id
- CLI info: command run, run parameters
- Git info: git sha, git branch, is branch dirty or not 
 
## Usage within Kedro 🏗
The` KedroSession` is a relatively new component within Kedro and at the time of writing, is mainly used to manage run lifecycles and for experiment tracking. The experiment tracking feature makes use of a session store implementation called the `SQLiteStore`, which uses SQLite to persist data. Other implementations of the session store available in Kedro are:

1. `BaseSessionStore`: the base class for all session stores that doesn’t persist any data
2. `ShelveStore`: implementation that uses the [shelve](https://docs.python.org/3/library/shelve.html) package to persist data
 
## Relation of a `run` and a `session` 🧑‍🤝‍🧑
While working on https://github.com/kedro-org/kedro/issues/1273 it was decided that Kedro session and Kedro run have a 1-1 mapping. This means that when a session gets created it will only ever be possible to kick off one full pipeline run during that specific session’s existence. In practice, Kedro manages this for you under the hood when `kedro run` is executed. 
 
## FAQ ❓
_How does a Kedro user use KedroSession?_
As a Kedro user you don’t need to access the session directly. When you execute the `kedro run` command, a new session gets created automatically. This session will then kick off the pipeline run and when that process finishes, the session will be closed again persisting any run data if the project is configured with a persistent session store. 
 
_What about using KedroSession in an interactive workflow?_
When using jupyter or ipython you can access the active `session` object or create a new one. You can then retrieve the `session_id`, the run data that will be stored, load the context, and execute a run. However, we do not encourage users to use the session other than for checking the `session_id` and run data. 
 

### Related Github issues and PRs:

- #1273 
- #1329 
- #1314
- https://github.com/kedro-org/kedro-viz/issues/764

datajoely:
A question that I think myself and others will ask is - if I want to access the data catalog as a live object, do I need to create a session for that? Is that the right way?

MerelTheisenQB:
[...]

PR #1423, opened 2022-04-07
Improve kedro run as a package

AntonyMilneQB:
⚠️ This is a huge description because I spent ages figuring this out and I thought we should have it all spelt out somewhere. Haven't done tests or docs yet - just opening this to get initial feedback.

The actual change to code is very small, but a big improvement I think: you can now do
```
from spaceflights.__main__ import main
outputs = main(pipeline="ds")
print(outputs)
``` 
and it will Just Work, even in IPython/Juptyer. All the command line ways of invoking `main` work exactly as before, i.e. `python -m spaceflights --pipeline=ds` etc. We now have consistency between `main` and CLI ways of launching kedro.

## Description
A user can run their kedro project in several ways. Note that the `run` command executed can be defined on the framework side or overridden in turn by a plugin or a project cli.py (done by `_find_run_command`).

1. Kedro CLI: `kedro run`. This is the only route that doesn't go through the project `__main__.py`; instead it goes through `kedro.framework.main`, which builds the CLI tree and does something like `_find_run_command`
2. CLI but not kedro CLI: `python -m spaceflights`. This hits the project `__main__.py` and will call `main`
3. CLI but not kedro CLI: `python src/spaceflights`. This is just a more unusual way of doing 2 
4. Inside your own python script: `from spaceflights.__main__ import main; main()`, then run the script using `python`
5. Inside IPython/Jupyter with _no_ kedro ipython extension: same as 4 but you run it in the notebook
6. Inside IPython/Jupyter with kedro ipython extension: same as 5 but you can also do `session.run` (which is what we have advertised as the way to do a kedro run in the past)

All the above must be run from within the project root or kedro won't be able to find your conf. Options 2 onwards needs you to have `pip install`ed your project or to have `src` in your `PYTHONPATH`. Note that having the project `pip install`ed could mean first doing `kedro package` and then `pip install` the resulting .whl file or it could mean just `pip install ./src` from your project root; it doesn't make a difference.

## Current problems
1. When using 1, 2, 3 and 4 you specify arguments using the CLI syntax `--pipeline ds`. With `main()` you do `main(["--pipeline", "ds"])`, i.e. the CLI syntax shoehorned into a Python function call. When using `session.run` you do a pure Python function call as `session.run(pipeline_name="ds")` - but note the argument name is different!!
2. Methods 5 and 6 run ok but generate a horrible big mysterious ipython exception after completing
3. Method 4 isn't actually useful if you want to run a kedro project with anything else since (big problem) no code after `main()` will actually execute...
4. ... and (smaller problem) `main()` doesn't return anything like `session.run` does, so you couldn't use any kedro outputs downstream anyway
5. Method 6 using `main` doesn't work unless you call `session.close` first because you'll get some error about there already being an active session, same as https://github.com/kedro-org/kedro-viz/issues/811
 
This PR fixes 1, 2, 3, 4. We should have separate ticket(s) to fix 5 and also the following:
- [ ] move the horrible code out of the project `__main__.py` to somewhere framework side, because no one wants to see that in their project... Possible locations might be `kedro.framework.cli.utils` or `kedro.framework.project`
- [ ] figure out if we can get a user to do a nicer looking import than `from spaceflights.__main__ import main`, which doesn't really look like it should be done
- [ ] fix a bug where commands defined in the project cli.py aren't given their full path when you do `kedro` but instead `Project specific commands from cli`. I think this is a bug anyway and it didn't used to be like this...
- [ ] consider changing the behaviour when the project cli.py exists but is empty/all commented out. I found the current  behaviour a bit annoying when playing around with this
- [ ] consider whether we should recommend `main` over `session.run` in IPython or even stop exposing `session` altogether. Mooted briefly in https://github.com/kedro-org/kedro/issues/1335 but needs more discussion. `main` is a higher-level function than `session.run` and respects the hierarchy of framework < plugins < project cli.py. The arguments to `main` also match the usual `kedro run` arguments, which those in `session.run` do not

## Development notes
There's basically two things that have changed:
- add `return` to the kedro run command. `session.run`, and hence `main`, returns the free outputs of the pipeline after the kedro run. This fixes problem 4
- change how `main` calls the `run` command. This fixes problems 1, 2, 3

## Adventures with click
I opened [an issue on the click repo](https://github.com/pallets/click/issues/2249) (which didn't go down well... 😅) describing the original source of problems 1, 2, 3. In short, when you run a click command it _always_ exits afterwards with an Exception, even if execution was successful. This means it's impossible to run any code after that `main` in a Python script and also massively confuses ipython, which blows up.

For the record, some of the things that don't quite fix this are:
* `run(standalone_mode=True)` like suggested on the click repo. This solves problem 2 and 3 but not 1
* `run.callback`. This nearly gets you there but doesn't fill in the default arguments for you so you need to explicitly give all the arguments in the function call, which is horrible
* `run.main` is equivalent to what we were doing before (`run.__call__`) so doesn't help
* `run.invoke` is _really_ close to doing it, but unlike `run.forward` doesn't fills in default arguments and won't notice CLI arguments when called with `python -m spaceflights --pipeline=ds`

Overall, the code here is I think the unique thing that does exactly what we want both on the CLI and in scripts/ipython 🎉 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1423"></a>

### DIFF
--- a/kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/__main__.py
+++ b/kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/__main__.py
@@ -41,6 +41,13 @@ def main(**kwargs):
     package_name = Path(__file__).parent.name
     configure_project(package_name)
     run = _find_run_command(package_name)
+    # This is what happens under the hood of run.__call__. The click context contains
+    # arguments coming from the CLI command (sys.argv, e.g. ["--pipeline", "ds"]) and
+    # default values of arguments that are not supplied. We forward the context to the
+    # invocation of run. Any **kwargs supplied (e.g. `main(pipeline="ds")` will
+    # overwrite the arguments supplied by the context. Overall this means we have a
+    # system where the same main function can handle both arguments coming from CLI and
+    # a Python API.ß
     with run.make_context("run", sys.argv[1:]) as ctx:
         return ctx.forward(run, **kwargs)
 
--- a/kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/__main__.py
+++ b/kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/__main__.py
@@ -3,6 +3,7 @@ as `{{ cookiecutter.repo_name }}` and `python -m {{ cookiecutter.python_package 
 """
 import importlib
 from pathlib import Path
+import sys
 
 from kedro.framework.cli.utils import KedroCliError, load_entry_points
 from kedro.framework.project import configure_project
@@ -36,11 +37,12 @@ def _find_run_command_in_plugins(plugins):
             return group.commands["run"]
 
 
-def main(*args, **kwargs):
+def main(**kwargs):
     package_name = Path(__file__).parent.name
     configure_project(package_name)
     run = _find_run_command(package_name)
-    run(*args, **kwargs)
+    with run.make_context("run", sys.argv[1:]) as ctx:
+        return ctx.forward(run, **kwargs)
 
 
 if __name__ == "__main__":
--- a/kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/__main__.py
+++ b/kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/__main__.py
@@ -1,56 +1,28 @@
 """{{ cookiecutter.project_name }} file for ensuring the package is executable
 as `{{ cookiecutter.repo_name }}` and `python -m {{ cookiecutter.python_package }}`
 """
-import importlib
-from pathlib import Path
 import sys
+from pathlib import Path
 
-from kedro.framework.cli.utils import KedroCliError, load_entry_points
-from kedro.framework.project import configure_project
-
-
-def _find_run_command(package_name):
-    try:
-        project_cli = importlib.import_module(f"{package_name}.cli")
-        # fail gracefully if cli.py does not exist
-    except ModuleNotFoundError as exc:
-        if f"{package_name}.cli" not in str(exc):
-            raise
-        plugins = load_entry_points("project")
-        run = _find_run_command_in_plugins(plugins) if plugins else None
-        if run:
-            # use run command from installed plugin if it exists
-            return run
-        # use run command from `kedro.framework.cli.project`
-        from kedro.framework.cli.project import run
-
-        return run
-    # fail badly if cli.py exists, but has no `cli` in it
-    if not hasattr(project_cli, "cli"):
-        raise KedroCliError(f"Cannot load commands from {package_name}.cli")
-    return project_cli.run
+from kedro.framework.project import configure_project, run
 
 
-def _find_run_command_in_plugins(plugins):
-    for group in plugins:
-        if "run" in group.commands:
-            return group.commands["run"]
+# def main(args=None, **kwargs):
+#     package_name = Path(__file__).parent.name
+#     configure_project(package_name)
+#     run(args or sys.argv[1:], **kwargs)
+#     return 0
 
 
 def main(**kwargs):
     package_name = Path(__file__).parent.name
     configure_project(package_name)
-    run = _find_run_command(package_name)
-    # This is what happens under the hood of run.__call__. The click context contains
-    # arguments coming from the CLI command (sys.argv, e.g. ["--pipeline", "ds"]) and
-    # default values of arguments that are not supplied. We forward the context to the
-    # invocation of run. Any **kwargs supplied (e.g. `main(pipeline="ds")` will
-    # overwrite the arguments supplied by the context. Overall this means we have a
-    # system where the same main function can handle both arguments coming from CLI and
-    # a Python API.ß
-    with run.make_context("run", sys.argv[1:]) as ctx:
-        return ctx.forward(run, **kwargs)
+    if kwargs:
+        run(**kwargs)
+    else:
+        run(sys.argv[1:])
+    return 0
 
 
 if __name__ == "__main__":
-    main()
+    sys.exit(main())
--- a/features/steps/test_starter/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/__main__.py
+++ b/features/steps/test_starter/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/__main__.py
@@ -41,6 +41,13 @@ def main(**kwargs):
     package_name = Path(__file__).parent.name
     configure_project(package_name)
     run = _find_run_command(package_name)
+    # This is what happens under the hood of run.__call__. The click context contains
+    # arguments coming from the CLI command (sys.argv, e.g. ["--pipeline", "ds"]) and
+    # default values of arguments that are not supplied. We forward the context to the
+    # invocation of run. Any **kwargs supplied (e.g. `main(pipeline="ds")` will
+    # overwrite the arguments supplied by the context. Overall this means we have a
+    # system where the same main function can handle both arguments coming from CLI and
+    # a Python API.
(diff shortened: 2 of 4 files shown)


AntonyMilneQB:
P.S. technically this is as breaking change since the way you call `main` with arguments has changed. But given that the `main(["--pipeline", "ds"])` syntax is so new and not well-known I think we should just go for this in 0.18.1. It would be really nice to have this working for databricks, etc.

noklam:
This is a nice summary, also reminds me I have some challenges to make `kedro` pipeline works with other python code. I think this would help integrate kedro with any downstream tasks.

https://github.com/kedro-org/kedro/discussions/795. I am not sure if the discussion still holds or things can be done more easily now.

I think this doc is also useful to lay out how people can run kedro pipeline.

> Method 4 isn't actually useful if you want to run a kedro project with anything else since (big problem) no code after main() will actually execute...

Why is this happening?



> from spaceflights.__main__ import main

It did look a bit weird, but I think it is because we want to execute `python -m spaceflights` thus we need to put it in a `__main__.py` file? Although I haven't see any project actually use this before.

> consider changing the behaviour when the project cli.py exists but is empty/all commented out. I found the current behaviour a bit annoying when playing around with this

I am not super familiar with click and how `cli.py` and plugins get pick up by kedro, so can't comment on this now.

> consider whether we should recommend main over session.run in IPython or even stop exposing session altogether. Mooted briefly in https://github.com/kedro-org/kedro/issues/1335 but needs more discussion. main is a higher-level function than session.run and respects the hierarchy of framework < plugins < project cli.py. The arguments to main also match the usual kedro run arguments, which those in session.run do not

I do like the idea of a higher-level entry point, 

Misc:
This is not specific to `kedro run`, but I notice when I do kedro run I would get a dependencies error (Pip's error, and it's buried in the log that are not obvious at all), complaining that it doesn't match with the plugins, even I am not running the plugin at all.

noklam:
Just an idea that pops up, do we only expect `kedro run` as a self-contained package? It is possible a `kedro` pipeline behaves just like a normal function where it takes an argument instead of catalogue only?

noklam:
@AntonyMilneQB 

In #1807 this would work for packaged kedro project. What's the benefit of calling the CLI instead of just using `KedroSession`, which is the preferable way of calling kedro pipeline programmatically?

```python
from kedro.framework.session import KedroSession
from kedro.framework.project import configure_project
package_name = <your_package_name>
configure_project(package_name)

with KedroSession.create(package_name) as session:
    session.run()
```

AntonyMilneQB:
@noklam sorry for the incredibly slow response - finally got around to this issue on my big list of things to respond to... 😅 

There's a few problems with doing `session.run` like that:
* `python -m spaceflights` should be possible if you just want to run a packaged pipeline by itself. You should be able to  trigger (part of) a pipeline by itself from command line without needing to write a Python script to do it
* having to do `configure_project` and use a context manager for a `KedroSession` feels overly complex, fiddly and too low-level. There should be a more user-friendly way that wraps all that into a single function call
* as you say, it doesn't respect the cli.py `run` command > plugin `run` command command > kedro `run` command hierarchy. In the future I hope this will become increasingly irrelevant. As you've noted before, the cli.py route is not well used (more of a historical leftover I think) so maybe it should be removed. The use of a plugin to override `run` will hopefully become less common if we do something like https://github.com/kedro-org/kedro/issues/1041 also. But for now (0.18.x), I think maintaining this hierarchy is worth doing

noklam:
> * `python -m spaceflights` should be possible if you just want to run a packaged pipeline by itself. You should be able to  trigger (part of) a pipeline by itself from the command line without needing to write a Python script to do it

I understand this need, but I also doubt if there are any people doing this for a real project, executing a module entirely with `python -m` isn't a common thing that I do. With a `kedro` project it's likely you have more than one pipeline, and will need different parameters, so it feels quite limited. If `package_name` is a useful common (See point 2), then it can be just group in `kedro run --package_name=xxx` and all other arguments, which is much more consistent and powerful.


> * in the context of a Python script, your code works fine but doing `configure_project` and a context manager for a `KedroSession` feels overly complex, fiddly and too low-level. There should be a more user-friendly way that wraps all that into a single function call. It should be more obvious how to run a kedro pipeline programmatically. The code you wrote is fine for our "under the hood" implementation but I feel like there should be more accessible higher-level function to do it

I agree that `configure_project` is too low-level as a user-facing function. I think the problem here is what `KedroSession.create(package_name=xxx)` does, it actually did nothing. The only reason why it works in a kedro project is we did this `configure_project` steps behind the scene, either through `cli` or `reload_kedro`, what you pass into `KedroSession` doesn't matter at all. 

@AntonyMilneQB 

I am not sure if `KedroSession` and context managers are too low level though, given this is the standard way to work with the notebook (Databricks) or any Python API. Currently, I view `session.run` as the closest Python API to `kedro run`. 
If the `package_name` argument is doing its job, then it should be the same for both `project` or `package` mode. This is also quite consistent with the current way. `kedro run` is just using `KedroSession.run` and the only job of the `cli` is parsing the argument.

Why do we need a context manager? Is that a good reason to keep them at all? Note that both `extra_params` and `env` are in the context manager because we are saving them to `session_store`, but it doesn't add anything at the moment. In the past, it has some responsibility of activating/deactivating sessions, but that was removed for a while already. In reality, "closing" a session doesn't do anything. So in theory there can be a `kedro_run` function which basically just do "Create a Session and run a session".


> * as you say, it doesn't respect the cli.py `run` command > plugin `run` command command > kedro `run` command hierarchy. In the future I hope this will become increasingly irrelevant. As you've noted before, the cli.py route is not well used (more of a historical leftover I think) so maybe it should be removed. The use of a plugin to override `run` will hopefully become less common if we do something like [Universal Kedro deployment (Part 3) - Add the ability to extend and distribute the project running logic #1041](https://github.com/kedro-org/kedro/issues/1041) also. But for now (0.18.x), I think maintaining this hierarchy is worth doing

I think this is the only compelling reason to me for keeping backward compatibility. But as you said, it's not something we want to keep and this is also a very uncommon usage.

kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/__main__.py
```diff
@@ -1,47 +1,28 @@
...
-
-
-def main(*args, **kwargs):
+def main(**kwargs):
     package_name = Path(__file__).parent.name
     configure_project(package_name)
-    run = _find_run_command(package_name)
-    run(*args, **kwargs)
+    if kwargs:
+        run(**kwargs)
```
noklam, line 21: I guess this is unfinished? as I can see in the other `__main__.py` the result of `session.run` is return but here it doesn't return the output

MerelTheisenQB:
We discussed this in **Technical Design** and talked about the following:

- The `package_name` argument provided in `KedroSession.create()` doesn't do anything, which is confusing to users. This argument should be removed, but since that is a breaking change it will happen in 0.19
- The workaround to run a run within a session is valid, but it's important to note that `bootstrap_project` is only meant for configuring a project that hasn't been packaged. When working with a packaged project `configure_project` should be used instead.
- The problem with the `main()` method is that it currently returns an exit code, so downstream processes can't do anything with it.

Some steps to take to improve the situation and explain it better to users:

- [ ] Add documentation about running a packaged project
- [ ] Add documentation about what the `bootstrap_project` method does and what it's supposed to be used for
- [ ] Add documentation about what the `configure_project` method does and what it's supposed to be used for

noklam:
> The problem with the main() method is that it currently returns an exit code, so downstream processes can't do anything with it.

Clarify this one a bit. I think these are 2 problems
1. `main` should return the session.run() - so downstream processing is possible
2. `click` generate a `sys.exit()` by default, and since `main` wrap around the CLI so it creates some issue on Databricks or even just IPython - [this link](https://github.com/kedro-org/kedro/issues/1807#issuecomment-1240747948) explains more.

merelcht:
This task isn't done, and is definitely something we want to work on. @AntonyMilneQB will never forget about this, and will make sure it gets discussed properly 😄 This PR will be closed because it has diverged a lot from the `main` branch and isn't mergeable.

PR #1434, opened 2022-04-12
Remove KedroSession activate/deactivate mechanism

### LINKED ISSUE #1431: Remove logic to activate and deactivate `KedroSession`
KedroSession is still keeping a global for the currently active session, which causes problems when trying to create a session in a notebook with the Kedro IPython extension and makes it unable to create/run the session. Here's the functions we need to drop:
https://github.com/kedro-org/kedro/blob/main/kedro/framework/session/session.py#L34-L47

And here is where they are used: https://github.com/kedro-org/kedro/blob/main/kedro/framework/session/session.py#L274-L287

Need to add a regression test:
- Create a test that shows the bug: create a session, then try creating another session and see the failure
- Then fix it in the code
- Then this test should pass

noklam:

## Development Notes
* Remove the activate/deactivate & global `_active_session` all together in favor of the related `after_context_created` hook.

## Description

Close #1431 (Copied from #1431)

I checked the repository the `_activate_session()` and the global `_active_session` are only used by ipython.py and the `KedroSession`'s context manager. It was also used by `get_current_session()` but it no longer exists now.

The activate/deactivate session mechanism doesn't look necessary to me, in fact, if you just do `session.run()` today it runs perfectly fine. This whole activate/deactivate thing only stops you if you are using context manager.

Based on our discussion, it looks like we only want to do 1., but we should just do 2. unless there are some use cases we missed. 

1. Update the logic so a session can be created with the existing active session (this was already possible before, it just doesn't work with the Context Manager style of session. However, this essentially means we do not care if there are pre-exist `_active_session`.

2. Remove the activate/deactivate altogether, since the global `_active_session` is not used by anyone (was used by `get_current_session`). But since we allow multiple sessions and there is no practical use to keep track of "active".

## Development notes

* Add test for making sure multiple session can be created when we use context manager.
* Remove activate/deactivate and global `_active_session` 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1434"></a>


Galileo-Galilei:
Hi @noklam , can you check my comment in #1431 before you merge this PR?

noklam:
@Galileo-Galilei I am checking on the comments and older Github issue. Your concerns are valid and this won't be merged until further discussed. I can see why this is important for plugins. Thank you for your detailed comments.

Galileo-Galilei:
Thank you very much!

noklam:
Notes:

1. The scope of this PR will be limited to an ad-hoc fix so it doesn't block new session in Jupyter/Databricks environment when an active session is not closed.
2. The discussion of removing global `_active_session` will be deferred. Particularly, we will need more information about how the plugins developers are using it and do we need to expose more. Potentially adding new hooks/changing hookspec or some more thinking about the responsibility of KedroSession & KedroContext.

noklam:
Ready for review again, but we need to make sure it is merged only after #1458 and release them both in `0.18.1`

kedro/framework/session/session.py
```diff
@@ -31,14 +31,10 @@
...
+    session: "KedroSession", force: bool = False  # pylint: disable=unused-argument
+) -> None:
     global _active_session
-
-    if _active_session and not force and session is not _active_session:
-        raise RuntimeError(
-            "Cannot activate the session as another active session already exists."
-        )
-
     _active_session = session
```
AntonyMilneQB, line 38: Given that https://github.com/kedro-org/kedro/issues/1458 is a prerequisite for this, will we need any of this stuff at all any more?

I'm a bit confused by whether the plan is to simultaneously do both of these:
1. introduce `after_context_created` hook
2. remove all the `_active_session` stuff

Or whether we want to do:
1. remove `_activate_session` stuff but leave `_active_session` global there
2. introduce `after_context_created` hook
3. remove all the `_active_session` stuff

i.e. do we do a straight swap in 0.18.1 (replace all `_active_session` things with the new hook) or do we do a phased introduction (in 0.18.1 you can use `_active_session` OR the new hook; then in 0.18.2 you can only use the new hook)?

Would be good to know if this affects @Galileo-Galilei 🙂

tests/framework/session/test_session.py
```diff
@@ -285,6 +286,18 @@ def test_create(
...
+        self, fake_project, mock_package_name
+    ):
+        first_session = KedroSession.create(mock_package_name, fake_project)
+        _activate_session(first_session)
+        assert kedro.framework.session.session._active_session is first_session
+
+        new_session = KedroSession.create(mock_package_name, fake_project)
+        _activate_session(new_session)
+        assert kedro.framework.session.session._active_session is new_session
+
```
AntonyMilneQB, line 300: Would it be possible to do the context manager of this just to check it works the same way?
```
with KedroSession.create() as first_session:
    assert kedro.framework.session.session._active_session is first_session
    with KedroSession.create() as new_session:
        assert kedro.framework.session.session._active_session is new_session
    assert kedro.framework.session.session._active_session is new_session ## is that right?
```

kedro/framework/session/session.py
noklam: I think the idea is to do 1&2 together in `0.18.1`.  Doing 1 only would leave the `_active_session` None only? Sorry I found that there are commits in my local repository but never pushed to the remote once as I was switching between these related branches.

Galileo-Galilei: For what it's worth, I am completely ok with scenario (1), i.e. shipping the new hook and removing the session activation mechanism simultaneously in kedro==0.18.1. This is a private method and I think we can safely assume that people using my dirty hack of importing the global variable in their own plugin know what they are doing and accept the inherent risk of doing so. 

There is always a possibility that we miss something and that releasing everything at the same time will eventually break the plugin with no workaround at the plugin. I'll come back complaining in a new issue if it were to happen :). I have some concerns about the current proposal to make the context "frozen", but I'll let a comment in the other PR.

RELEASE.md
```diff
@@ -22,6 +22,7 @@
 * Introduced `after_command_run` CLI hook.
 * Update sections on visualisation, namespacing, and experiment tracking in the spaceflight tutorial to correspond to the complete spaceflights starter.
 * Fixed `Jinja2` syntax loading with `TemplatedConfigLoader` using `globals.yml`.
+* Deprecated global `_active_session`, `_activate_session and `_deactivate_session`
```
AntonyMilneQB, line 25:
```suggestion
* Removed global `_active_session`, `_activate_session` and `_deactivate_session`. Plugins that need to access objects such as the config loader should now do so through `context` in the new `after_context_created` hook.
```

AntonyMilneQB (approved):
Just a small comment on the tests but definitely happy to see this change!

noklam:
@AntonyMilneQB With the global `_active_session` it makes sense to test if it get activated. Maybe what worth testing is `KedroContext` should be identical using the non-contextmanger or contextmanager version?

I think we can now check this thanks to `attrs` actually, before that it would compare their hash instead.

RELEASE.md
```diff
@@ -22,6 +22,7 @@
 * Introduced `after_command_run` CLI hook.
 * Update sections on visualisation, namespacing, and experiment tracking in the spaceflight tutorial to correspond to the complete spaceflights starter.
 * Fixed `Jinja2` syntax loading with `TemplatedConfigLoader` using `globals.yml`.
+* Removed global `_active_session`, `_activate_session` and `_deactivate_session`. Plugins that need to access objects such as the config loader should now do so through `context` in the new `after_context_created` hook.
```
MerelTheisenQB, line 25: 👍

tests/framework/session/test_session.py
MerelTheisenQB: I think it makes sense to check that it's now possible to create two sessions without it failing, like it did previously.

MerelTheisenQB (approved):

Issue #1437, opened 2022-04-13
Jinja syntax no longer works with TemplatedConfigLoader


AntonyMilneQB:
While working on [upgrading the kedro-viz demo-project to 0.18](https://github.com/kedro-org/kedro-viz/pull/827) we found the config loader got stuck on loading https://github.com/kedro-org/kedro-viz/blob/main/demo-project/conf/base/catalog_04_feature.yml, which has Jinja syntax.

This is a regression from 0.17.x where Jinja syntax worked fine.

I did a quick fix by providing a custom version of `TemplatedConfigLoader` that is the same as the built-in kedro version but with `ac_template` changed to `True` on [L127](https://github.com/kedro-org/kedro/blob/main/kedro/config/templated_config.py#L127).

To do:
- [x] understand why our current tests didn't flag this regression, since it's quite a big and bad one I think
- [x] write test(s) to demonstrate that 0.18 doesn't work with Jinja2
- [x] fix `TemplatedConfigLoader` until above tests work - **please don't assume my change to L127 is the right solution**, it was just meant as a quick fix. Especially given my confusion explained in the below post
- [x] check your new fixed `TemplatedConfigLoader` works on the kedro-viz demo-project

AntonyMilneQB:
I'm actually not sure why my quick fix worked here 🤔  `self._config_mapping` just searches through files that match `globals_pattern`, [which is `*globals.yml`](https://github.com/kedro-org/kedro-viz/blob/feature/Upgrade-demo-project-to-kedro-0.18/demo-project/src/demo_project/settings.py). So why is Jinja syntax in a the file `catalog_04_feature.yml`, which doesn't match that pattern, affected by a change to L127?

Issue #1011, opened 2021-11-01
Enable PyYAML special tag in the Kedro config loader


datajoely:
## Description
PyYAML supports a large range of non-standard YAML magic functions which allow the user to pass in special tags that extend the standard YAML functionality beyond what is allowed out of the box + cross platform. Examples include `!!tuple` for casting a list to a python `tuple` or event complex tags such as `!!python/object:module.cls` for a class instance.

Users in the past have asked for [tuple](https://github.com/quantumblacklabs/kedro/discussions/973) and [object support](https://github.com/quantumblacklabs/kedro/discussions/898) so it is something that would be helpful. 

## Possible Implementation

Kedro uses `anyconfig` to parse configuration files and we can coerce it into parsing this special PyYAML syntax like follows:
```python
from yaml import UnsafeLoader
import anyconfig
anyconfig.loads("my_object: !!python/tuple [0.3, 0.4]", ac_parser="yaml", Loader=UnsafeLoader)
>>> {'my_object': (0.3, 0.4)}
```

To enable this it would be nice to put a toggle in the `TemplatedConfigLoader` constructor with a boolean argument like `allow_unsafe` and an extra check at this point to enable this special loading mechanism if the `ConfigLoader` is parsing a file ending with `.yml` or `.yaml`:
https://github.com/quantumblacklabs/kedro/blob/be3022ecf9ea18f6a448ef9d2c57bd8ad089c769/kedro/config/config.py#L156

## Possible Alternatives
Keep as it is, there may be some unintended security risks enabling this functionality.

datajoely:
For reference: https://pyyaml.org/wiki/PyYAMLDocumentation

AntonyMilneQB:
I like this idea. My general opinion on config is that we should let users do whatever they want with it and not be constrained by the strict rules of yaml. And since this functionality is already in pyyaml, great. I think we should either support it natively in kedro or provide documentation on how users can write their own config loader to enable it. The only thing I don't like is that it is specific to yaml (as opposed to the other config types we support), but in reality does anyone use those other config types anyway?

### Implementation
I would suggest three different things:
1. make this as general as possible by allowing to pass whatever arguments they want to `anyconfig.load` rather than using an `allow_unsafe` boolean that fixes certain arguments
2. maybe put this in a whole new `SomethingConfigLoader` rather than in `TemplatedConfigLoader` ❓ Not sure
3. maybe use `FullLoader` rather than `UnsafeLoader` ❓ Not sure

[...]

Issue #1411, opened 2022-04-04
How to pass the `env` to config loader since registration moved to `settings.py` in 0.18.x


datajoely:
## Description

In 0.17.x we would register the config loader like so:

```python
def def register_config_loader(
        self, conf_paths: Iterable[str], env: str, extra_params: Dict[str, Any]
    ) -> ConfigLoader:
    return TemplatedConfigLoader(
        conf_paths,
        globals_pattern="*globals.yml",
        globals_dict= {"env": env},
    )

```

In 0.18.x we can register a configloader like this:

```python
from kedro.config import TemplatedConfigLoader  # new import

CONFIG_LOADER_CLASS = TemplatedConfigLoader
CONFIG_LOADER_ARGS = { 'globals_pattern':"*globals.yml"}
```

However we can no longer retrieve the runtime environment in this manner? (I may be missing something)

## Context
> How has this bug affected you? What were you trying to accomplish?

When using `TemplatedConfigLoader` this is a common pattern to make sure paths are explicit.

## Possible solutions:

- Kedro could automatically include `$kedro_env` in the `globals_dict` / `self._config_mapping` namespace?
- You can get creative with the `KEDRO_ENV` environment variable, but it is a fragile solution

datajoely:
@pascalwhoop

AntonyMilneQB:
You're not missing anything: this is indeed no longer possible. I believe it's a deliberate casualty of moving library component customisation from registration hooks into settings.py. Similarly injecting credentials into `register_data_catalog ` is no longer possible - see https://github.com/kedro-org/kedro/issues/1282.

There are two possible ways to do this now:
1. Just manually define `env` in the globals.yml file in each environment.
2. Write a simple custom config loader which injects the variable:
```
class MyTemplatedConfigLoader(TemplatedConfigLoader):
    def __init__(*args, **kwargs):
        super().__init__(*args, **kwargs)
        self._config_mapping["env"] = self.env
```

Personally I'm not a big fan either of these, and I think your idea of automatically including `env` as a variable in `TemplatedConfigLoader` sounds like a sensible thing to do. The decision in #1282, and I imagine for this issue too, would be that the ultimate solution should be considered as part of the great configuration improvements in #891, which I really hope we will tackle soon.

Galileo-Galilei:
[...]

Issue #891, opened 2021-09-13
Synthesis of user research when using configuration in Kedro


hamzaoza:
## Summary
Configuration overhead is an issue that has arisen time and time again from user feedback – particularly as Kedro projects scale in complexity. From user interviews, it can be observed the three main configurations used were for `kedro run`, the Data Catalog and parameters. The remaining options were seen as “_setup once and forgot_” for the remainder of the project. Overall configuration in Kedro is well received and liked by users who appreciate the approach Kedro has taken so far. 

During this research, it became clear that configuration scaling impacts a small set of use cases where you have multiple environments (e.g. `dev`, `staging` and `prod`) and multiple use cases – maybe you’re using the same or a similar pipeline across different products for different countries. To gather deeper insights participants were presented with two existing options for the Data Catalog, and two possible solutions: [pattern matching]( https://en.wikipedia.org/wiki/Pattern_matching) and [Jinja templating](https://jinja.palletsprojects.com/en/3.0.x/) (favouring the former of the two). Users were also asked about their feelings about moving the Data Catalog entirely to Python. Participants were universally against the idea of moving the Data Catalog into Python as it would fundamentally go against the principles of Kedro.

## Table of Contents

1. Introduction
2. Background
3. Research Approach
4. User Interview Matrix
5. Configuration Synthesis
6. GitHub Analysis
7. Data Catalog Generator
8. Solution Criteria

## 1. Introduction

Configuration overhead is an issue that has arisen time and time again from user feedback – particularly as Kedro projects scale in complexity. It’s also an issue for new users who have never been exposed to this concept i.e., Data Scientists using software engineering principles for the first time. This research aims to understand the key pain points users face when using configuration and test possible solutions for the Data Catalog to develop a specification criterion for any solution.

## 2. Background

Kedro is influenced by the [12 Factor App](https://12factor.net/config) but this results in a lot of duplication of configuration. From users, we have heard that yaml files can become unwieldy with each entry written manually making it error prone. Users also want to apply runtime parameters and want to parameterise runs in complex ways which Kedro doesn’t currently support.

As a result, some teams have tried to solve this independently – most notably by using Jinja2 templating through the template config loader though this has not become widespread across other teams. However, as we continue to grow, it is likely that more users will encounter similar issues and will need a Kedro native solution to support growth.

[...]

PR #1438, opened 2022-04-13
Improve `TemplatedConfigLoader` -  Default `${KEDRO_ENV}` global  & expose `anyconfig.load` kwargs 

datajoely:
*FOR DISCUSSION*

## Description
This is a culmination of several things I've seen users struggle with, both related to `TemplatedConfigLoader`.

### 1. Automatically scope the config environment as a  `${KEDRO_ENV}`

Since 0.18.x it is no longer possible to neatly pass the current configuration environment to the `TemplatedConfigLoader`. This used to be possible via the `register_config_loader` hook that had the `env` argument in scope and was thus easy to pass into scope. @pascalwhoop has raised this internally, but we have no GH issue associated with it yet.

### 2. Allowing the user to define arbitrary arguments to  `anyconfig.load`

We currently leverage `anyconfig.load` to open our configuration files, the `ac_template` argument was exposed in 0.17.0, but there are other arguments available as per the [anyconfig API docs](https://python-anyconfig.readthedocs.io/en/latest/usage.html#common-and-backend-specific-keyword-options-on-load-single-config-file). I see no reason not to expose this in a generic way allowing the user to alter the behaviour to their choosing.

- I'm 99% sure this fixes @AntonyMilneQB 's PR https://github.com/kedro-org/kedro/issues/1437
- It would also make #1011 and more complex Jinja templating possible if the user decides to go down that route

## Development notes

To make this possible I've changed:
- The `ac_template` parameter that's passed down in several places to a dictionary that contains this (and in theory any other arguments) to be passed to `anyconfig.load`.
- The `env` argument is added automatically to `self._config_mapping` within the constructor.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1438"></a>

### DIFF
--- a/kedro/config/common.py
+++ b/kedro/config/common.py
@@ -50,7 +50,7 @@ SUPPORTED_EXTENSIONS = [
     ".xml",
 ]
 
-DEFAULT_ANYCONFIG_ARGS = {"ac_template": True}
+DEFAULT_ANYCONFIG_ARGS: Dict[str, Any] = {"ac_template": False}
 
 _config_logger = logging.getLogger(__name__)
 
@@ -95,6 +95,7 @@ def _get_config_from_patterns(
 
     config = {}  # type: Dict[str, Any]
     processed_files = set()  # type: Set[Path]
+    custom_anyconfig_args = anyconfig_args if anyconfig_args else DEFAULT_ANYCONFIG_ARGS
 
     for conf_path in conf_paths:
         if not Path(conf_path).is_dir():
@@ -107,8 +108,7 @@ def _get_config_from_patterns(
             Path(conf_path), patterns, processed_files, _config_logger
         )
         new_conf = _load_configs(
-            config_filepaths=config_filepaths,
-            anyconfig_args=anyconfig_args if anyconfig_args else DEFAULT_ANYCONFIG_ARGS,
+            config_filepaths=config_filepaths, anyconfig_args=custom_anyconfig_args
         )
 
         common_keys = config.keys() & new_conf.keys()
@@ -161,7 +161,7 @@ def _load_config_file(
                 k: v
                 for k, v in anyconfig.load(
                     file,
-                    **(anyconfig_args if anyconfig_args else DEFAULT_ANYCONFIG_ARGS),
+                    **anyconfig_args,
                 ).items()
                 if not k.startswith("_")
             }
@@ -178,7 +178,7 @@ def _load_config_file(
 
 def _load_configs(
     config_filepaths: List[Path],
-    anyconfig_args: Optional[Dict[str, Any]] = None,
+    anyconfig_args: Dict[str, Any],
 ) -> Dict[str, Any]:
     """Recursively load all configuration files, which satisfy
     a given list of glob patterns from a specific path.
@@ -206,7 +206,7 @@ def _load_configs(
     for config_filepath in config_filepaths:
         single_config = _load_config_file(
             config_filepath,
-            anyconfig_args=anyconfig_args if anyconfig_args else DEFAULT_ANYCONFIG_ARGS,
+            anyconfig_args=anyconfig_args,
         )
         _check_duplicate_keys(seen_file_to_keys, config_filepath, single_config)
         seen_file_to_keys[config_filepath] = single_config.keys()
--- a/kedro/config/common.py
+++ b/kedro/config/common.py
@@ -1,31 +1,3 @@
-# Copyright 2021 QuantumBlack Visual Analytics Limited
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#     http://www.apache.org/licenses/LICENSE-2.0
-#
-# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
-# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES
-# OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE, AND
-# NONINFRINGEMENT. IN NO EVENT WILL THE LICENSOR OR OTHER CONTRIBUTORS
-# BE LIABLE FOR ANY CLAIM, DAMAGES, OR OTHER LIABILITY, WHETHER IN AN
-# ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF, OR IN
-# CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
-#
-# The QuantumBlack Visual Analytics Limited ("QuantumBlack") name and logo
-# (either separately or in combination, "QuantumBlack Trademarks") are
-# trademarks of QuantumBlack. The License does not grant you any right or
-# license to the QuantumBlack Trademarks. You may not use the QuantumBlack
-# Trademarks or any confusingly similar mark as a trademark for your product,
-# or use the QuantumBlack Trademarks in any other manner that might cause
-# confusion in the marketplace, including but not limited to in advertising,
-# on websites, or on software.
-#
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
 """This module contains methods and facade interfaces for various ConfigLoader
 implementations.
 """
--- a/kedro/config/common.py
+++ b/kedro/config/common.py
@@ -33,7 +33,7 @@ implementations.
 import logging
 from glob import iglob
 from pathlib import Path
-from typing import AbstractSet, Any, Dict, Iterable, List, Set
+from typing import AbstractSet, Any, Dict, Iterable, List, Optional, Set
 from warnings import warn
 
 from yaml.parser import ParserError
@@ -49,13 +49,16 @@ SUPPORTED_EXTENSIONS = [
     ".properties",
     ".xml",
 ]
+
+DEFAULT_ANYCONFIG_ARGS = {"ac_template": True}
+
 _config_logger = logging.getLogger(__name__)
 
 
 def _get_config_from_patterns(
     conf_paths: Iterable[str],
     patterns: Iterable[str] = None,
-    ac_template: bool = False,
+    anyconfig_args: Optional[Dict[str, Any]] = None,
 ) -> Dict[str, Any]:
     """Recursively scan for configuration files, load and merge them, and
     return them in the form of a config dictionary.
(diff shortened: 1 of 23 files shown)


datajoely:
Docs look like this when rendered
image

kedro/config/common.py
```diff
@@ -49,13 +49,16 @@
     ".properties",
     ".xml",
 ]
+
+DEFAULT_ANYCONFIG_ARGS = {"ac_template": True}
```
AntonyMilneQB, line 53: This should be `False` to preserve the current default behaviour, right?

kedro/config/common.py
```diff
@@ -103,7 +107,8 @@ def _get_config_from_patterns(
             Path(conf_path), patterns, processed_files, _config_logger
         )
         new_conf = _load_configs(
-            config_filepaths=config_filepaths, ac_template=ac_template
+            config_filepaths=config_filepaths,
+            anyconfig_args=anyconfig_args if anyconfig_args else DEFAULT_ANYCONFIG_ARGS,
```
AntonyMilneQB, line 111: The right logic here would be
```
{**DEFAULT_ANYCONFIG_ARGS, **anyconfig_args}
```
Otherwise you need to always repeat the keys given in `DEFAULT_ANYCONFIG_ARGS` when specifying `anyconfig_args` even if you're not modifying them from their default values.

kedro/config/common.py
```diff
@@ -147,11 +155,14 @@ def _load_config_file(config_file: Path, ac_template: bool = False) -> Dict[str,
...
         # Default to UTF-8, which is Python 3 default encoding, to decode the file
-        with open(config_file, encoding="utf8") as yml:
+        with open(config_file, encoding="utf8") as file:
             _config_logger.debug("Loading config file: '%s'", config_file)
             return {
                 k: v
-                for k, v in anyconfig.load(yml, ac_template=ac_template).items()
+                for k, v in anyconfig.load(
+                    file,
+                    **(anyconfig_args if anyconfig_args else DEFAULT_ANYCONFIG_ARGS),
```
AntonyMilneQB, line 164: I don't think we need this logic in every function - just the top level one  that actually gets called.

kedro/config/templated_config.py
```diff
@@ -113,24 +114,34 @@ def __init__(
                 dictionary. This dictionary will get merged with the globals dictionary
                 obtained from the globals_pattern. In case of duplicate keys, the
                 ``globals_dict`` keys take precedence.
+            anyconfig_kwargs: Keyword arguments to pass to the the ``anyconfig.load``
```
AntonyMilneQB, line 117:
```suggestion
            anyconfig_args: Keyword arguments to pass to the ``anyconfig.load``
```

kedro/config/templated_config.py
```diff
@@ -113,24 +114,34 @@ def __init__(
...
+                method. By default we enabled jinja2 via the ``ac_template`` argument.
         """
         super().__init__(
             conf_source=conf_source, env=env, runtime_params=runtime_params
         )
         self.base_env = base_env
         self.default_run_env = default_run_env
+        self._anyconfig_args = (
+            anyconfig_args if anyconfig_args else {"ac_template": True}
+        )
```
AntonyMilneQB, line 127: Similarly you actually want something like `{"ac_template": True, **anyconfig_args}`.

Also I'm not sure about having `DEFAULT_ANYCONFIG_ARGS` in common.py while putting the defaults inside `__init__` here. I don't think which approach is best but I think we should probably do it the same way in both files.

AntonyMilneQB (commented):
Just some minor comments on the code, let me comment on the higher level idea here separately...

kedro/config/templated_config.py
```diff
@@ -113,24 +114,34 @@ def __init__(
...
+                anyconfig_args=self._anyconfig_args,
             )
             if globals_pattern
             else {}
         )
+        env_dict = (
+            deepcopy({"KEDRO_ENV": env})
+            if env
+            else deepcopy({"KEDRO_ENV": self.base_env})
+        )
```
AntonyMilneQB, line 142: I don't see the need to `deepcopy` here? Also I don't think this logic is quite right here.  Since we can ignore `base`, I think we should have:
```
self._config_mapping = {**self._config_mapping, **globals_dict, "KEDRO_ENV": env or default_run_env}
```

AntonyMilneQB:
I have mixed thoughts on this. I totally understand the need for both these features and think both your solutions are improvements on our current state. The question in my mind is whether they would fit in well with the ultimate solution to https://github.com/kedro-org/kedro/issues/891 and what a realistic timeframe for that ultimate solution would be. e.g. I wouldn't want everyone to start using `${KEDRO_ENV}` in their config files only for that to become impossible in the future.

1. Automatically the config environment as a ${KEDRO_ENV}
Isn't this the github issue that tracks it? https://github.com/kedro-org/kedro/issues/1411. I suggested two current workarounds, but building this into the `TemplatedConfigLoader` natively like you've done here makes sense to me. My hesitation is as above, whether this will continue to be the right solution.

2. Allowing the user to define arbitrary arguments to anyconfig.load
This is basically what I suggested in https://github.com/kedro-org/kedro/issues/1011 and so obviously I like it 😀 Looking at it again I'm not sure it's right to expose `anyconfig_args` in the `TemplatedConfigLoader.__init__` though. I think it's probably lower footprint to make the dictionary accessible to override through a custom `MyTemplatedConfigLoader` like this instead:
```
class MyTemplatedConfigLoader(TemplatedConfigLoader):
    def __init__(*args, **kwargs):
        super().__init__(*args, **kwargs)
        self._anyconfig_args = {"Loader": yaml.UnsafeLoader}
```
This is a bit more effort for users  than being able to specify `anyconfig_args` directly in `CONFIG_LOADER_ARGS` in settings.py but means (a) we keep the signature of `TemplatedConfigLoader` a bit cleaner and (b) we don't need to worry about breaking backwards compatibility in the future  if we wanted to remove the argument and implement this another way. Given how few people would want to change `anyconfig_args` from their defaults, I'm not sure it's worth a whole new user-facing argument. This implementation seems like an easy win to me since it immediately gives users more power for how anyconfig.load operates for very little work and risk of breaking changes on our side.

AntonyMilneQB:
Yeah, that makes sense, but I think it's still possible without altering the signature of `TemplatedConfigLoader.__init__` itself:

```
class MyTemplatedConfigLoader(TemplatedConfigLoader):
    def __init__(*args, **kwargs):
        super().__init__(*args, **kwargs)
        self._anyconfig_args = {"Loader": yaml.UnsafeLoader, "whatever_arg_you_need": kwargs["globals_dict"]} # or `self._config_mapping` or whatever
```

I like the idea of generalising to `anyconfig_args`, just I think it probably should not go in the signature.

datajoely:
Okay I'm sold - will update to match at some point

kedro/config/templated_config.py
datajoely: Since we flipped the booleaan on the other one they are different now...

datajoely: I also didn't know about this dict merging syntax, thought you had to go one level up

datajoely:
Updated docs as per the latest changes
image

kedro/config/common.py
```diff
@@ -63,6 +67,7 @@ def _get_config_from_patterns(
 
     config = {}  # type: Dict[str, Any]
     processed_files = set()  # type: Set[Path]
+    custom_anyconfig_args = anyconfig_args if anyconfig_args else DEFAULT_ANYCONFIG_ARGS
```
AntonyMilneQB, line 70:
```suggestion
    custom_anyconfig_args = anyconfig_args or DEFAULT_ANYCONFIG_ARGS
```

kedro/config/common.py
```diff
@@ -98,13 +103,16 @@ def _get_config_from_patterns(
     return config
 
 
-def _load_config_file(config_file: Path, ac_template: bool = False) -> Dict[str, Any]:
+def _load_config_file(
+    config_file: Path, anyconfig_args: Optional[Dict[str, Any]] = None
```
AntonyMilneQB, line 107:
```suggestion
    config_file: Path, anyconfig_args: Dict[str, Any]]
```
since the argument is always provided now, and the `**anyconfig_args` wouldn't actually work with the default `None` value anyway.

kedro/config/templated_config.py
```diff
@@ -119,18 +119,23 @@ def __init__(
...
             )
             if globals_pattern
             else {}
         )
         globals_dict = deepcopy(globals_dict) or {}
-        self._config_mapping = {**self._config_mapping, **globals_dict}
+        self._config_mapping = {
+            **self._config_mapping,
+            **globals_dict,
+            **{"KEDRO_ENV": env or base_env},
```
AntonyMilneQB, line 137:
```suggestion
            "KEDRO_ENV": env or base_env,
```

AntonyMilneQB (commented):
Hmm, on second thoughts this won't work quite right following my suggestion, sorry... Because `self._anyconfig_args` is already used in L127 of `TemplatedConfigLoader.__init__`,  if you do this:
```
class MyTemplatedConfigLoader(TemplatedConfigLoader):
    def __init__(*args, **kwargs):
        super().__init__(*args, **kwargs)
        self._anyconfig_args = {"Loader": yaml.UnsafeLoader}
```
then `self._anyconfig_args` only gets overwritten _after_ it's used on L127.

I don't know if this really matters since, as I understand it, this is not the main important call to `_get_config_from_patterns` (that is in `conf_paths`). But if it does matter we might need to do something with properties to make this work properly.

Either way, these are implementation details, and my general comments still stand (general approval of at least idea 2). Would be good to hear what other people think about it before we spend longer figuring out the right implementation.

kedro/config/common.py
datajoely: You can't do this because it's `Optional` and may be `None`

sheldontsen-qb:
Noticing the jinja2 and passing any extra `anyconfig_kwargs`, we also do the following:
```
        # add environment variables
        self._ctx["env"] = dict(os.environ)

        # if anyconfig throws a warning from templates,
        # so catch this as an error within a context manager
        # and reset warnings after exiting context manager.
        with warnings.catch_warnings():
            warnings.filterwarnings(action="error", module="anyconfig")
            parsed_result = {
                k: v
                for k, v in anyconfig.load(
                    config_file,
                    ac_template=True,
                    paths=filtered_conf_path_modified,  # <---- location of where to find templates which should be configurable
                    ac_context=self._ctx,
                ).items()
                if not k.startswith("_")
            }

        return parsed_result
```

Notice we just dump everything in `env` by default - because why not, secrets are usually managed by env vars anyway. Now you have a consistent way to reference any env var in catalogs/parameters. 

Notice we're catching warnings and raising them as errors - because `anyconfig` has a `try...except` inside somewhere which gives a red-herring error. This will help surface the actual rendering issue rather than `anyconfig` just passing through silently and giving you a malformed error. 

Regarding the comment of including `globals.yml` into the jinja context, I think it might be better to have 2 separate config loaders and not mix syntax. So a `TemplatedConfigLoader` and a `Jinja2ConfigLoader`, either go with `globals.yml` and the `${}` syntax, or `application_vars.j2` with the `{{ }}` syntax (don't have to settle with that name, but you need something equivalent). 

Note it is possible to build it in a way that you can support both, but you'd need to properly document which comes first (jinja2 first resolves then globals) but I think it would be messy to see both kinds of syntax. Hence the preference to keep them separate. However, it is possible to build this in a backwards compatible way so one can slowly transition away from `${}` to `{{ }}` over time instead of one big PR. But then one perhaps should consider providing guidance on this.  

Just sharing my two cents!

datajoely:
After discussion with core team - will close this PR and raise a couple of issues

PR #1448, opened 2022-04-19
Fix CircleCI triggers

AntonyMilneQB:
## Description
Every hour there's a mysterious CircleCI job being launched on kedro-viz in my name...

After much detective work I think I've figured out the reason: when we changed from scheduled workflows to scheduled pipelines in https://github.com/kedro-org/kedro/pull/1365 I think we missed some conditions. Instead of just the `hourly_pr_merge`, it's actually running all the workflows on main, which include a `viz_build` job which triggers a build on kedro-viz CI.

Hopefully making these conditions stricter will mean that only `hourly_pr_merge` gets triggered hourly and then the extra jobs on kedro-viz (and on kedro) no longer get triggered.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1448"></a>

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -505,13 +505,13 @@ workflows:
   version: 2.1
 
   build_docs_only:
-    when: <<pipeline.parameters.docs_change>>
-    unless:
-      or:
-        - <<pipeline.parameters.code_change>>
-        - <<pipeline.parameters.release_kedro>>
-        - <<pipeline.parameters.run_hourly>>
-        - <<pipeline.parameters.run_nightly>>
+    when:
+      and:
+        - <<pipeline.parameters.docs_change>>
+        - not: <<pipeline.parameters.code_change>>
+        - not: <<pipeline.parameters.release_kedro>>
+        - not: <<pipeline.parameters.run_hourly>>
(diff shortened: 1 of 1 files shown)


tynandebold (approved):
Amazing. I've been wondering for days why this Viz build is happening.

rashidakanchwala (approved):
awesome Sherlock!! :D

noklam (commented):
Thank you for fixing this, didn't notice `kedro-viz` build would be trigger from the `kedro` repo.

The CI conditions are a bit annoying though, basically we have to specify almost everything as "do this when X but not y1 and y2 and y3 and y4 and y5 ...".

AntonyMilneQB:
> Thank you for fixing this, didn't notice `kedro-viz` build would be trigger from the `kedro` repo.
> 
> The CI conditions are a bit annoying though, basically we have to specify almost everything as "do this when X but not y1 and y2 and y3 and y4 and y5 ...".

Yes, it's very annoying! I can see why people are complaining about scheduled pipelines compared to scheduled workflows.

AntonyMilneQB:
@MerelTheisenQB please could you merge? None of the failures here are a result of my PR 🙂

PR #1450, opened 2022-04-19
Use CircleCI context for AWS keys for ECR publish

idanov:
## Description

Using CircleCI context for AWS credentials provides an easy way to use one set of credentials over multiple repositories. Since we have an enforced key rotation scheme for the ECR instance we are using, this will make key rotations a tad easier.


AntonyMilneQB (approved):
Nice change!

MerelTheisenQB (approved):
👍

PR #1453, opened 2022-04-20
Update configuration docs

MerelTheisenQB:
## Description
The documentation about the use of `ConfigLoader` was terribly out of date. I've updated them to reflect the current way of instantiating a `ConfigLoader`.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1453"></a>

### DIFF
--- a/kedro/config/config.py
+++ b/kedro/config/config.py
@@ -51,8 +51,11 @@ class ConfigLoader(AbstractConfigLoader):
 
         >>> import logging.config
         >>> from kedro.config import ConfigLoader
+        >>> from kedro.framework.project import settings
+        >>> from pathlib import Path
         >>>
-        >>> conf_loader = ConfigLoader('conf', 'local')
+        >>> conf_path = str(Path.cwd() / settings.CONF_SOURCE)
+        >>> conf_loader = ConfigLoader(conf_source=conf_path, env="local")
         >>>
         >>> conf_logging = conf_loader.get('logging*')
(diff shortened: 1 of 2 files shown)


kedro/config/config.py
```diff
@@ -51,8 +51,10 @@ class ConfigLoader(AbstractConfigLoader):
 
         >>> import logging.config
         >>> from kedro.config import ConfigLoader
+        >>> from kedro.framework.project import settings
         >>>
-        >>> conf_loader = ConfigLoader('conf', 'local')
+        >>> conf_path = str(project_path / settings.CONF_SOURCE)
```
noklam, line 56: Thank you for updating this!

Is `project_path` something available in settings already? Can we make `conf_path` available in settings so we can import it instead of asking users to compose the path themselves?

noklam (approved):

kedro/config/config.py
MerelTheisenQB: No it's not. I was deciding between `project_path` or `Path.cwd()`, and chose the former because I thought it would be clearer to the user what path they should be providing. The user would only ever need to construct this path themselves if they want to create a custom `ConfigLoader` or access the object directly in a notebook as far as I'm aware. So IMO it's fine to keep it like this because they won't usually need to construct anything at all.

noklam: That makes sense, thank you.

AntonyMilneQB (approved):
Thanks for fixing this!

PR #1454, opened 2022-04-20
Replace `os.subprocess` with python based git library

### LINKED ISSUE #1451: Rewrite `_describe_git` in `KedroSession` to use python based implementation of the git commands
> https://github.com/FriendCode/gittle seems to be the easiest option, although it hasn't been maintained for years and it's probably not a good idea to rely on a seemingly abandoned project. Another option is to implement the two functions we need ourselves as per this example here: https://wyag.thb.lt/#orgd8ab0ff (which is more fun, but probably harder, especially for the `git status` check).

# Background
Traditionally we have been using `git` within kedro run to track metadata about the working directory. The current way we do is using `subprocess` which calls the git executable directly. However, it causes some issues when we are in some read-only / Databricks's dbfs / spark-submit job. 

The idea is to use native python calls to eliminate the need for an external process.

Todo:
- [ ] Find alternatives (open-source library or re-implement the 2 functions we need)
Possible solutions
* https://github.com/jelmer/dulwich (A low-level library that is still actively maintained)
* https://github.com/FriendCode/gittle  (A high-level library, but is no longer maintained)
* Implement ourselves (reference: https://wyag.thb.lt/#orgd8ab0ff)

# Use Case
* `_describe_git` https://github.com/kedro-org/kedro/blob/b3c78e57eab5865ffdc32a5813cb839811705d68/kedro/framework/session/session.py#L50-L71
  * ["git", "rev-parse", "--short", "HEAD"]
  * ["git", "status", "--short"],

noklam:

## Description

Fix #1451

## Development notes

* Replace the `subprocess` call with native Python code

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1454"></a>

### DIFF
--- a/kedro/framework/session/session.py
+++ b/kedro/framework/session/session.py
@@ -4,13 +4,14 @@ import getpass
 import logging
 import logging.config
 import os
-import subprocess
 import traceback
 from copy import deepcopy
 from pathlib import Path
 from typing import Any, Dict, Iterable, Union
 
 import click
+from git import GitError, InvalidGitRepositoryError, Repo
+from sqlalchemy import Boolean
 
 from kedro import __version__ as kedro_version
 from kedro.config import ConfigLoader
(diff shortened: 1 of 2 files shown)


noklam:
Do we want to convert everything to function calls? There is also heavy use of subprocess in various tests.

Issue #506, opened 2020-09-08
[KED-2143] Adding a ConfigLoader instance into hook specs params


takikadiri:
## Description
When developping a kedro plugin, i regularly need to access to configs and potentielly some plugin-specific configs files. Since the plugin use hook mechanism, i no longer can bring whatever context attribute to my hook implemantation (except the parameters defined in the hook specs).

## Context
Here in the [kedro-mlflow](https://github.com/Galileo-Galilei/kedro-mlflow/blob/404da2844964c4454377cf55d71f8ac48d1443b6/kedro_mlflow/framework/context/mlflow_context.py#L11)  plugin we were forced to redefine a ConfigLoader instance inside the plugin. 
That lead to incoherence between the context ConfigLoader property and the new Configloader created inside the hook.

Other plugins will need this functionality, i imagine a kedro-spark plugin that use hook mechanism and access a spark config file from project folder path (spark.yml), or a kedro-sas plugins that do the same thing (getting configs in order to create a parametrized session)
## Possible Implementation
A possible implementation is to pass the context config_loader to the hook.

hook specs
```python
 @hook_spec
    def before_pipeline_run(
        self, run_params: Dict[str, Any], pipeline: Pipeline, catalog: DataCatalog, config_loader: ConfigLoader
    ) -> None:
```

context
```python
hook_manager = get_hook_manager()
hook_manager.hook.before_pipeline_run(  # pylint: disable=no-member
            run_params=record_data, pipeline=filtered_pipeline, catalog=catalog, config_loader=self.config_loader
        )
```

lorenabalan:
Hi @takikadiri , you've highlighted a very good point. We thought about this and we've actually added a set of hooks to register library components, such as pipelines, data catalog, and [config loader](https://github.com/quantumblacklabs/kedro/commit/5484b9b4bed1878c0dc25fc4feae2f2bd731cab4), with a Kedro project. I think might solve your use case.
This functionality will be made available in 0.16.5, which is going to be released very soon. :)

takikadiri:
Thank you @lorenabalan for the quick reply ! It's realy great having the possibility to registrer library component such as the config loader, i will certainly use it.
But my point here is about not having the possibility to pass the config loader instance (created with register_config_loader) to another hook let's say the pipeline_before_run hook. 
There may be something that escapes me about the hook mechanisms :)

Galileo-Galilei:
[...]

PR #1465, opened 2022-04-21
Make `KedroContext` a frozen dataclass 

### LINKED ISSUE #1459: Revisit: Make KedroContext a dataclass and add `config_loader` as a property
# Description
As `KedroSession` now control the lifecycle of Kedro's run, `KedroContext` act like a container and it stores important attributes. Since we now dropped Python 3.6 support, we can make use of Python's `dataclass` to further tidy up `KedroContext`'s constructor code.

todo:
- [x] Make `KedroContext` a dataclass(https://github.com/kedro-org/kedro/pull/1465)
- [x] Add `config_loader` as @property instead of relying the private `context._get_config_loader()` https://github.com/kedro-org/kedro/pull/1505
- [x] Update corresponding document in Starter/Kedro

noklam:
## Description

Fix #1459 

## Context
As `KedroSession` now control the lifecycle of Kedro's run, `KedroContext` acts as a data container and stores important attributes. 
Since we now dropped Python 3.6 support, we can make use of Python's `dataclasses`.

The benefit of it is
1. Signifying `KedroContext` will be just a data container.
2. Simplify Python class construction (less boilerplate code) using the new Python feature.

## Development Changes
This PR is a bit long so I try to keep the summary of changes at the top. 

* Make `KedroContext` a dataclass with `attrs`, Python's `dataclasses` was considered but it does not fit well. See more in the discussion.
  * Any simple public `@property` that we use to do `self._x = x` is now just a one line defintion at the top with `attrs.setters.frozen` 
* `config_loader` is now a public property of `KedroContext`.
* Add tests to ensure important `KedroContext` attributes remains immutable in case we refactor this in future.

The Class init signature is unchanged as evidenced by this:
image

## More Development Notes
* Make `KedroContext` a dataclass, signifying it will be simply a container as we gradually move the logic out from it.
    * `dataclass` doesn't support partial "read-only" property, so there are 2 options to achieve that.
       1. Do it like old fashion, use `@property` to achieve the read-only attribute, just don't use `dataclass` or use alternatives like `attrs` (ancestor of `dataclass`, but it's not inside the standard library).
       3. Set `frozen=True`, but use `__setattr__` for `__post_init__`.  The benefit of using it is to lock the context, so `context.arbitary_attr = 'xyz'` assignment is not possible, it also reduce some boilerplate code to set `@property`, but the tradeoff is the more complicated `__post_init__` method. (See https://stackoverflow.com/questions/59222092/how-to-use-the-post-init-method-in-dataclasses-in-python).
       4. Use `dataclass` but not frozen, mimic the frozen by custom implementation instead.

The `__post_init__` look like this if we follow (b):
```python
        object.__setattr__(self, "_package_name", package_name)
        object.__setattr__(
            self, "project_path", Path(project_path).expanduser().resolve()
        )
        object.__setattr__(self, "_extra_params", deepcopy(extra_params))
        object.__setattr__(self, "_hook_manager", hook_manager)
```
     
Originally I favor method (b) as I am okay with a cleaner class with the complicated bit just in `__post_init__`, however, it seems to create trouble for `mypy` and other linter as it doesn't understand the attribute assignment.

## todo 
- [ ] [PR - Update pyspark doc](https://github.com/kedro-org/kedro-starters/pull/87) (thanks Antony!)

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1465"></a>


AntonyMilneQB:
Great work getting to the bottom of this! I also didn't realise it would be so difficult.

A few quick comments:
* remember to update the pyspark starters with the same change you make in  the docs
* I think your solution (b) is fine, but I'm also not personally so fussed if we just don't use `frozen` at all and then can do a more normal `self._x = x` assignment in `__post_init__`

kedro/framework/context/context.py
```diff
@@ -159,65 +159,47 @@ def _update_nested_dict(old_dict: Dict[Any, Any], new_dict: Dict[Any, Any]) -> N
...
 
-        """
-        return self._project_path
+    def __post_init__(self, package_name, project_path, hook_manager, extra_params):
+        object.__setattr__(self, "_package_name", package_name)
+        object.__setattr__(
+            self, "project_path", Path(project_path).expanduser().resolve()
+        )
+        object.__setattr__(self, "_hook_manager", hook_manager)
+        object.__setattr__(self, "_extra_params", deepcopy(extra_params))
```
MerelTheisenQB, line 202: What's the logic behind which attributes are added here and the ones that aren't? For example, why is `package_name` added but not `env`?

MerelTheisenQB (commented):
I tend to agree with @AntonyMilneQB that it would be okay not to use `frozen`. We'll just need to make it clear to users what the implications of overwriting are.

kedro/framework/context/context.py
noklam: Good questions!

Personally, I am not too worried about it, but it is still a change in API since we now open up the possibility to overwrite it (though we can argue users could overwrite the `self._xxxx` if they are hacky.

noklam: The current implementation is approach b, which works functionally, but it will also break IDE support & linters they seem to have trouble understand `obj.__setattr__(self, 'a', 1)` has the same semantic as `self.a = 1`. Things like code navigation no longer works.

My hesitation is seems we didn't get many benefits with `dataclass` and do we really need it.

noklam: Would love to hear some opinion from @idanov

noklam:
Another possible workaround (c) by implementing the frozen mechanism, the advantage is it doesn't break code completion/linting and no weird `object.__setattr__`, but it starts to feel there is too many effort to fight with `dataclass`. Both (b) & (c) requires deep understanding how `dataclass` implement the immutability and find clever way to get around with it.

```python

@dataclass
class KedroContext:
    """``KedroContext`` is the base class which holds the configuration and
    Kedro's main functionality.
    """

    package_name: InitVar[str]
    project_path: InitVar[Union[Path, str]]
    config_loader: ConfigLoader
    hook_manager: InitVar[PluginManager]
    env: Optional[str] = None
    extra_params: InitVar[Dict[str, Any]] = None


    def __post_init__(self, package_name, project_path, hook_manager, extra_params):        
        self.project_path = Path(project_path).expanduser().resolve()
        self._package_name = package_name
        self._hook_manager = hook_manager
        self._extra_params = deepcopy(extra_params)

        self._proteced_attributes = ["package_name", "project_path", "config_loader"]
        # `__setattr__`, `__delattr__` is the dataclass way to mimic class's property
    
        def __setattr__(self, name, value):
            if name in self._proteced_attributes:
                raise FrozenInstanceError(f"cannot assign to field {name!r}")

        def __delattr__(self, name):
            if name in self._proteced_attributes:
                raise FrozenInstanceError(f"cannot delete field {name!r}")
        self.__setattr__ = __setattr__
        self.__delattr__ = __delattr__
```

kedro/framework/context/context.py
```diff
@@ -1,11 +1,13 @@
+# pylint: disable=no-member
```
MerelTheisenQB, line 1: Do we need to disable this for the whole file or is it possible to add it to the specific line(s) it applies to?

noklam: Yes, I change it back to the specific line(s). It's not clear why pylint complains about "no-member" for `config_loader` only but not the other attributes.

kedro/framework/context/context.py
```diff
@@ -159,65 +160,45 @@ def _update_nested_dict(old_dict: Dict[Any, Any], new_dict: Dict[Any, Any]) -> N
                 old_dict[key] = value
 
 
+def _expand_full_path(project_path: Union[str, Path]) -> Path:
+    return Path(project_path).expanduser().resolve()
+
+
+@define
```
AntonyMilneQB, line 167: Just from reading the docs, I wonder if we should be using https://www.attrs.org/en/stable/api.html#attr.attrs.frozen
```suggestion
@attrs.frozen
```
Maybe cleaner overall since it allows you to remove the `on_setattr=attrs.setters.frozen`? Or would it interfere with `_package_name` etc. attributes and then you have to use `field` for them?

tests/framework/context/test_context.py
```diff
@@ -238,8 +239,20 @@ def test_attributes(self, tmp_path, dummy_context):
...
+        with pytest.raises(AttributeError):
+            dummy_context.catalog = 1
+        with pytest.raises(AttributeError):
+            dummy_context.params = 1
+        with pytest.raises(FrozenAttributeError):
+            dummy_context.config_loader = 1
+        with pytest.raises(FrozenAttributeError):
+            dummy_context.env = 1
+        with pytest.raises(FrozenAttributeError):
+            dummy_context.project_path = 1
```
AntonyMilneQB, line 252: Should we also be testing that you can't set `package_name`, `extra_params` and `hook_manager`? Or can you set those? 🤔

AntonyMilneQB (approved):
Love the `attrs` solution, great work! 🤩 Just a few minor comments, nothing big.

requirements.txt
```diff
@@ -1,4 +1,5 @@
 anyconfig~=0.10.0
+attrs~=21.0
```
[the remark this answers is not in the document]
noklam: Thanks!

kedro/framework/context/context.py
noklam: 2 reasons for this.

1. To keep the API identical as before, so it is a pure refactor work that doesn't change any interface.
2. Initially I find `@frozen` doesn't work with the transforming logic with `__attrs_post_init__`, but I now find that it can works with `converter` so we could make this entire class immutable.

tests/framework/context/test_context.py
noklam: In theory, you can, it's just a Python class afterall, nothing stop you to overwrite `instance._attribute` if you really want.

kedro/framework/context/context.py
noklam: But let's go with `@frozen` now, it is more cleaner and actually simplifies the test as well as we got 1 `FrozenInstsance` error to check only.

noklam:
I love the implementation with `@frozen` which further cleans up the code. The constructor is now very concise and fits in half-page! No more scrolling!

AntonyMilneQB:
> I love the implementation with `@frozen` which further cleans up the code. The constructor is now very concise and fits in half-page! No more scrolling!

Agreed, this definitely looks like the right solution now! 🎉  Much cleaner than before.

RELEASE.md
```diff
@@ -17,12 +17,14 @@
 
 ## Bug fixes and other changes
 * Removed fatal error from being logged when a Kedro session is created in a directory without git.
+* `KedroContext` is now an `attrs`'s dataclass and `config_loader` is available as public attribute.
```
AntonyMilneQB, line 20:
```suggestion
* `KedroContext` is now an `attrs`'s frozen class and `config_loader` is available as public attribute.
```

kedro/framework/context/context.py
```diff
@@ -1,11 +1,12 @@
 """This module provides context for Kedro project."""
-
 from copy import deepcopy
 from pathlib import Path, PurePosixPath, PureWindowsPath
 from typing import Any, Dict, Optional, Union
 from urllib.parse import urlparse
 from warnings import warn
 
+import attrs
+from attrs import field
```
AntonyMilneQB, line 9:
```suggestion
from attrs import field, frozen
```
Super minor nitpicky suggestions, feel free to ignore 😀

kedro/framework/context/context.py
```diff
@@ -159,65 +160,42 @@ def _update_nested_dict(old_dict: Dict[Any, Any], new_dict: Dict[Any, Any]) -> N
                 old_dict[key] = value
 
 
+def _expand_full_path(project_path: Union[str, Path]) -> Path:
+    return Path(project_path).expanduser().resolve()
+
+
+@attrs.frozen
```
AntonyMilneQB, line 167:
```suggestion
@frozen
```

noklam:
Development notes:
CI is failing with probably irrelevant stuff, found one CI did pip-compile successfully 4 days ago [here](https://app.circleci.com/pipelines/github/kedro-org/kedro/7559/workflows/282d2c70-62f8-4f44-806b-b06e6de104ce/jobs/123610).

Galileo-Galilei:
> **@AntonyMilneQB  wrote:** I think your solution (b) is fine, but I'm also not personally worried if we just don't use frozen or use any properties at all, leave the variables as mutable and assume that people who overwrite them know what they are doing..Then we could do a more normal self._x = x assignment in __post_init__

> **@MerelTheisenQB wrote:** I tend to agree with @AntonyMilneQB that it would be okay not to use frozen. We'll just need to make it clear to users what the implications of overwriting are.

Just to make a point : as a user, I really want be able to at least **add** some attributes to the context (especially to enable passing object between hooks, but also to have a clean API to expose objects to user with an interactive workflow). If I were to override one of the existing attributes I would certainly do it in purpose, but I understand your concerns about possible (albeit unlikely) involuntary overriding by a user. 

**Maybe a compromise would be to make the existing attributes frozen, but still enable to add new ones?** I have absolutely no idea if it is easy to impement with ``attrs`` though. 

@noklam What is the rational about making the class frozen (while it is currently not)? Do you think my concerns make sense or do I misunderstand something?

noklam:
@Galileo-Galilei  Thank you for your comments.

> Just to make a point : as a user, I really want be able to at least add some attributes to the context (especially to enable passing object between hooks, but also to have a clean API to expose objects to user with an interactive workflow). If I were to override one of the existing attributes I would certainly do it in purpose, but I understand your concerns about possible (albeit unlikely) involuntary overriding by a user.

How would you retrieve these attributes after adding new attributes to the context? `after_context_created` is the only hook exposing `context` currently.

> Maybe a compromise would be to make the existing attributes frozen, but still enable to add new ones? I have absolutely no idea if it is easy to impement with attrs though.

It would be still possible by just doing `object.__setattr__('ATTR_NAME', value)`, instead of `context.ATTR_NAME = value`.


@noklam What is the rational about making the class frozen (while it is currently not)? Do you think my concerns make sense or do I misunderstand something?

You are correct that it is the `KedroContext` is more conservative now. I can share my thought.
1. We are now exposing the `context` object directly, thus more chance that the user may override or abuse the object for everything (`Context` is a higher-level object in Kedro, similar to the argument why not exposing `session` that gives users access to everything indirectly). So l am being more conservative and try to limit the surface that exposes to users. It's much easier to relax the interface if we find reasonable needs.
2. It's still possible to override, with an arguably less elegant method.
3. From `kedro` framework perspective, `context` should really be just a container created by `session` and thus it feels right to make it "immutable". 

These are my quick thoughts, I may miss something about interactive workflow, would love to hear more from you!

AntonyMilneQB:
Thanks very much for the comments @Galileo-Galilei. Before I comment on whether or not we should make this class frozen let me just come up with some concrete examples to make sure I understand what you're saying here and in your latest comment https://github.com/kedro-org/kedro/issues/506#issuecomment-1117903541. This took quite a bit of careful thought so please do let me know whether I've got these right!

### Example 1: passing objects between hooks

In this case you would like to use `context` as a vehicle to share objects between different hook classes.

```
class HooksA:
    @hook_impl
    def after_context_created(self, context):
        context.new_attribute = "something"
        self.context = context

    @hook_impl
    def before_node_run(self):
        print(self.context.new_attribute)

class HooksB:
    @hook_impl
    def after_context_created(self, context):
        print(context.new_attribute)
        self.context = context
    
    @hook_impl
    def before_node_run(self):
        print(self.context.new_attribute)
```

We know that `HooksA.after_context_created` always runs before `HooksA.before_node_run`, and similarly for `HooksB`, but it's not obvious what the ordering of `HooksA` vs. `HooksB` is. Hence the `HooksA.before_node_run` will work as intended, but for the `HooksB.after_context_created` and `HooksA.before_run_run` to work we rely on `HooksA.after_context_created` running before `HooksB.after_context_created`, which seems very fragile. (I don't even know how the ordering is defined if hooks are auto-registered through `pip install` rather than listed in `HOOKS` in settings.py - do you? Is it just alphabetical?)

In summary, we have:
* `HooksA.after_context_created` - ok
* `HooksA.before_node_run` - ok
* `HooksB.after_context_created` - relies on ordering
* `HooksB.before_node_run` - relies on ordering

The above is currently achieved using `context = _active_session.load_context()` as follows:
```
class HooksA:
    @hook_impl
    def before_pipeline_run(self):
        context = _active_session.load_context()
        context.new_attribute = "something"

    @hook_impl
    def before_node_run(self):
        context = _active_session.load_context()
        print(context.new_attribute)

class HooksB:
    @hook_impl
    def before_pipeline_run(self):
        context = _active_session.load_context()
        print(context.new_attribute)
    
    @hook_impl
    def before_node_run(self):
        context = _active_session.load_context()
        print(context.new_attribute)
```

Here we have:
* `HooksA.before_pipeline_run` - ok
* `HooksA.before_node_run` - ok
* `HooksB.before_pipeline_run` - relies on ordering
* `HooksB.before_node_run` - ok

... which is better than above (since `HooksB.before_node_run` is now ok), but still seems fragile.

### Example 2: passing objects into the interactive workflow

Similar to the above, you want to use `context` as a vehicle, but instead of sharing an object between two different hook classes, you want to pass the object to an ipython session. When I am in an ipython session with the `kedro.extras.extensions.ipython` loaded we do `context = session.load_context()` and make `context` available to the user. Hence `after_context_created` hooks have run, and in the Jupyter notebook you would be able to do `print(context.new_attribute)` if you could add `new_attribute` to the class.

There is no problem with ordering here as far as I can tell.

### Quick comments
Example 2 seems straightforward, but Example 1 seems fundamentally difficult due to the way that hooks work independently. Overall it would be useful to you have some sort of shared state between different hook classes. I'm not sure what the right way to do this is:
* maybe there's some pluggy docs on how this should be achieved, but I suspect it's considered something of an anti-pattern (inter-dependencies between hooks)?
* if there's no way to do in pluggy directly then you're relying on some sort of global kedro object (`session`/`context`/whatever). But I don't see an obvious way around the ordering fragility with any such method here?

Galileo-Galilei:
**TL; DR**

- **Adding attributes to the context is a _nice to have_**, not a must have. There are always workaround to achieve what I want, it's jsut a matter of slightly simplifying the developer experience.
- Even more, **since there is a workaround to add attributes even if the class is frozen, I'm compeletly fine with suggested implementation**

------------

Thank you both for your answers, it helps a lot. 

@noklam: 

> How would you retrieve these attributes after adding new attributes to the context? after_context_created is the only hook exposing context currently.

My idea was exactly to do as what Antony suggests in his example 2. This is "tricky and fragile" as he explains, but it does currently works and I've seen some real world example, albeit this is uncommon.

I am **totally in line with your point 1** (context is now exposed to user and as a consequence it must be more protected than it is now) **and your point 2** (you have a workaround if a user really wants to add extra attributes - I did not know this was possible with ``attrs``, that's good news!). 

Regarding your point 3, I have more mixed feelings: **I think it would feel very natural to kedro users to be able to do ``context.mlflow`` or ``context.spark`` to access custom configuration files**. After all, ``context`` is where configuration belong in kedro projects, even if these configuration files are not "official" ones. 

@AntonyMilneQB : you're reading my mind :smile:  I was not very precise but your made a wonderful summary of the 3 use cases I envision. 

- for **example 1**, I don't really need to modify the ``context`` as you mention since I can store the extra attribute in the ``Hook`` instance  itself to reuse it further, so **I don't really care here**.
- **for example 2**, this is exactly what I suggested and you've described it very precisely. **I've seen this use case in a real project, even if I feel like you this is very "tricky and fragile"**. I am 99% sure there is an ordering between hooks and I've checked it before, but I don't know exactly what the order is. I think it is "hooks from plugins in the order the packages are installed, then ``settings.py`` hooks in the order they are declared", but this should be checked. 
- **for example 3**, this is indeed what I implied. I actually do exactly what you suggest (push extra variable with ``get_ipython().push(new_attribute)`` in ``kedro-mlflow``, but I face **2 challenges**:
  - **most users ignore that these variables exist** when I add them in my own plugins, despite documenting them. If it was **inside the context, I hope that auto completion would help discovering** them.
  - **many users** in my company **don't launch sessions with ``kedro jupyter notebook`` because they use a remote jupyterhub** instance. As a consequence **they create the ``session`` and the ``context`` manually** at the beginning of their notebook, and they always forget to create and setup the mlflow configuration. The new ``after_context_created`` hook will enable to setup the configuration automatically which is great, but **if they need to access extra configuration, they have to recreate it manually**: 
```
from kedro_mlflow.config import get_mlflow_config
mlflow_config=get_mlflow_config(session)
mlflow_config.setup() # this will be done automatically inside the hook, but I'd like to expose the mlflow_config variable too
```

**from a dev experience perspective, it makes much more sense if they can access directly with ``context.mlflow_config``** with no action required from their side.

### On your comments


> - context is actually semantically the right place for the objects you’re trying to pass
> - context is just a convenient container but you can achieve everything you want just by making new variables instead
> - context is actually necessary to pass these objects, i.e. achieves something you can’t do just by making new variables. I don’t think this is true for hooks but maybe for ipython?

It's more about points 1 & 2 : I can put new configuration in custom variables (and I already do), but I think it is much more "user friendly" to expose it through the context (it is where I would naturally look for extra configuration + autocompletion would help to discover it) 

I agree example 1 & 3 are straightforward. Regarding example 2, I don't think this is something kedro should try to enforce specifically (and it indeed sounds like an anti pattern), but since it is currently doable it should be nice to preserve the behaviour if possible.

noklam:
Thanks for the comments. Storing plugins-related in `context` makes sense to me, and I think the workaround should be enough to support that.

I will proceed to merge this PR.

noklam:
Final decision - this is going into `develop` instead of `main`, which should be released in `0.19.0`. The argument for this is **Removing the ability to add attributes to Context is a breaking change**. 

A separate PR will be made for `main`.

AntonyMilneQB:
Thanks @Galileo-Galilei for all your comments, and sorry for the very slow response! What you say about adding attributes to the context being more user- and developer-friendly than independent variables definitely makes sense 👍 

Just a comment on the jupyter notebook point: we have been working on the interactive workflow and are planning to work on it further. There's a [category for such issues](https://github.com/kedro-org/kedro/issues?q=is%3Aopen+is%3Aissue++label%3A%22Component%3A+Jupyter%2FIPython%22+), and there will be more added to this milestone when I get round to it over the next few weeks. As ever we're very interested in hearing your feedback on it and any suggestions you have!

Part of this work which has already happened (released fully in 0.18, but actually the ipython extension has been around for a while before that) is to simplify the flow for exactly the case you describe of managed jupyter instances. Basically users should no longer have to manually set up a `context` etc. in their jupyter notebook; instead, just loading the kedro ipython extension should do everything for them. See https://kedro.readthedocs.io/en/stable/tools_integration/ipython.html#managed-jupyter-instances for details.

PR #1467, opened 2022-04-21
Fix TemplatedConfigLoader raising ValidationError

### LINKED ISSUE #1402: Project templating using TemplatedConfigLoader raises ValidationError in 0.18.0
## Description
When following the guide for [Template configuration](https://kedro.readthedocs.io/en/stable/kedro_project_setup/configuration.html#template-configuration), Kedro 0.18.0 raises ValidationError as `kedro.config.templated_config.TemplatedConfigLoader` is not recognized as a subclass of `kedro.config.config.ConfigLoader`.

## Context
This bug hinders me from using the `TemplatedConfigLoader`, according to the documentation.

## Steps to Reproduce
1. Create new Kedro project
```
kedro new --starter=pandas-iris
cd new-kedro-project
```
2. Create and activate venv. Install requirements.
```
python -m venv my_venv
source my_venv/bin/activate
pip install -r src/requirements.txt
```
3. Open the `settings.py` file.
```
nano src/new_kedro_project/settings.py
```
Uncomment lines 28,29,31,32,33 as suggested in the guide. Expected config:
```
# Class that manages how configuration is loaded.
from kedro.config import TemplatedConfigLoader
CONFIG_LOADER_CLASS = TemplatedConfigLoader
# Keyword arguments to pass to the `CONFIG_LOADER_CLASS` constructor.
CONFIG_LOADER_ARGS = {
    "globals_pattern": "*globals.yml",
}
```
4. Run the default pipeline
```
kedro run
```

## Expected Result
Defeault pipeline should run with global config keys overwritten.

## Actual Result
ValidationError is raised.

```
  File "/usr/local/bin/kedro", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.9/site-packages/kedro/framework/cli/cli.py", line 205, in main
    cli_collection = KedroCLI(project_path=Path.cwd())
  File "/usr/local/lib/python3.9/site-packages/kedro/framework/cli/cli.py", line 114, in __init__
    self._metadata = bootstrap_project(project_path)
  File "/usr/local/lib/python3.9/site-packages/kedro/framework/startup.py", line 155, in bootstrap_project
  File "/usr/local/lib/python3.9/site-packages/kedro/framework/project/__init__.py", line 166, in configure_project
    settings.configure(settings_module)
  File "/usr/local/lib/python3.9/site-packages/dynaconf/base.py", line 182, in configure
    self._wrapped = Settings(settings_module=settings_module, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/dynaconf/base.py", line 235, in __init__
    self.validators.validate(
  File "/usr/local/lib/python3.9/site-packages/dynaconf/validator.py", line 417, in validate
    validator.validate(self.settings, only=only, exclude=exclude)
  File "/usr/local/lib/python3.9/site-packages/kedro/framework/project/__init__.py", line 35, in validate
    raise ValidationError(
dynaconf.validator.ValidationError: Invalid value `kedro.config.templated_config.TemplatedConfigLoader` received for setting `CONFIG_LOADER_CLASS`. It must be a subclass of `kedro.config.config.ConfigLoader`.
```

## Your Environment
* Kedro version used (`pip show kedro` or `kedro -V`): `0.18.0`
* Python version used (`python -V`): `3.9.10`
* Operating system and version: `macOS 12.3`

AntonyMilneQB:
## Description
Fixes #1402 using the approach agreed on in tech design session: by making a new validator. This means that the default value if `CONFIG_LOADER_CLASS` is unspecified is `ConfigLoader`, but that supplying `TemplatedConfigLoader` in settings.py now passes validation.

Tests for settings.py have been made stricter (and failed before the fix). I've also tested manually and all works well now.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1467"></a>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -17,11 +17,9 @@
 
 ## Bug fixes and other changes
 * Removed fatal error from being logged when a Kedro session is created in a directory without git.
-* Fixed `CONFIG_LOADER_CLASS` validation so that `TemplatedConfigLoader` can be specified in settings.py.
-
-
-- Added runner name to pipeline hook spec `run_params`
-- Introduced an `after_command_run` CLI hook spec
+* Fixed `CONFIG_LOADER_CLASS` validation so that `TemplatedConfigLoader` can be specified in settings.py. Any `CONFIG_LOADER_CLASS` must be a subclass of `AbstractConfigLoader`.
+* Added runner name to the `run_params` dictionary used in pipeline hooks.
+* Introduced `after_command_run` CLI hook.
 
 ## Upcoming deprecations for Kedro 0.19.0
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -20,6 +20,9 @@
 * Fixed `CONFIG_LOADER_CLASS` validation so that `TemplatedConfigLoader` can be specified in settings.py.
 
 
+- Added runner name to pipeline hook spec `run_params`
+- Introduced an `after_command_run` CLI hook spec
+
 ## Upcoming deprecations for Kedro 0.19.0
 
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -17,6 +17,7 @@
 
 ## Bug fixes and other changes
 * Removed fatal error from being logged when a Kedro session is created in a directory without git.
+* Fixed `CONFIG_LOADER_CLASS` validation so that `TemplatedConfigLoader` can be specified in settings.py.
 
 
 ## Upcoming deprecations for Kedro 0.19.0
--- a/kedro/framework/project/__init__.py
+++ b/kedro/framework/project/__init__.py
@@ -49,6 +49,16 @@ class _HasSharedParentClassValidator(Validator):
         default_class = self.default(settings, self)
         for name in self.names:
             setting_value = getattr(settings, name)
+            # In the case of ConfigLoader, default_class.mro() will be:
+            # [kedro.config.config.ConfigLoader,
+            # kedro.config.abstract_config.AbstractConfigLoader,
+            # abc.ABC,
+            # object]
+            # We pick out the direct parent and check if it's in any of the ancestors of
+            # the supplied setting_value. This assumes that the direct parent is
+            # the abstract class that must be inherited from.
+            # A more general check just for a shared ancestor would be:
+            # set(default_class.mro()) & set(setting_value.mro()) - {abc.ABC, object}
             default_class_parent = default_class.mro()[1]
             if default_class_parent not in setting_value.mro():
                 raise ValidationError(
--- a/kedro/framework/project/__init__.py
+++ b/kedro/framework/project/__init__.py
@@ -42,6 +42,9 @@ class _IsSubclassValidator(Validator):
 
 
 class _HasSharedParentClassValidator(Validator):
+    """A validator to check that the parent of the default class is an ancestor of
+    the settings value."""
+
     def validate(self, settings, *args, **kwargs):
         super().validate(settings, *args, **kwargs)
 
--- a/kedro/framework/project/__init__.py
+++ b/kedro/framework/project/__init__.py
@@ -1,8 +1,6 @@
 """``kedro.framework.project`` module provides utitlity to
 configure a Kedro project and access its settings."""
 # pylint: disable=redefined-outer-name,unused-argument,global-statement
-import abc
-
 import importlib
 import logging.config
 import operator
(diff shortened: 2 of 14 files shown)


AhdraMeraliQB (approved):
LGTM! 👏

kedro/framework/project/__init__.py
```diff
@@ -39,6 +41,25 @@ def validate(self, settings, *args, **kwargs):
...
 
 
+class _HasSharedParentClassValidator(Validator):
+    """A validator to check that the parent of the default class is an ancestor of
+    the settings value."""
+
+    def validate(self, settings, *args, **kwargs):
+        super().validate(settings, *args, **kwargs)
+
+        default_class = self.default(settings, self)
```
noklam, line 49: so default_class here would be either `ConfigLoader` if not specified or `TemplatedConfigLoader`, correct?

kedro/framework/project/__init__.py
```diff
@@ -39,6 +41,25 @@ def validate(self, settings, *args, **kwargs):
...
+        for name in self.names:
+            setting_value = getattr(settings, name)
+            default_class_parent = default_class.mro()[1]
+            if default_class_parent not in setting_value.mro():
+                raise ValidationError(
+                    f"Invalid value `{setting_value.__module__}.{setting_value.__qualname__}` "
+                    f"received for setting `{name}`. It must be a subclass of "
+                    f"`{default_class_parent.__module__}.{default_class_parent.__qualname__}`."
```
noklam, line 57: Maybe we could add some comments to explain this.

Here is my understanding, `default_class` is `ConfigLoader` if we are talking about config_loader, but this is expected to be a generic validator.

So the assumption here is, **the default_class is something inherit the `AbstractXXXXX` class, thus default_class.mro()[1] will be pointing to the `AbstractXXXXX` class.** If the default aren't directly inherit from the abstract class, this would fail.

kedro/framework/project/__init__.py
AntonyMilneQB: No, `default_class` comes from `default=_get_default_class("kedro.config.ConfigLoader")` so is always `ConfigLoader`. The actual specified value is given in `setting_value = getattr(settings, name)`.

These dynaconf validators are very confusing to read I know but we're following their standard syntax here 😅

kedro/framework/project/__init__.py
AntonyMilneQB: Yes, you're exactly right and I will add a comment to explain this!

`ConfigLoader.mro()` looks like this:
```
[kedro.config.config.ConfigLoader,
 kedro.config.abstract_config.AbstractConfigLoader,
 abc.ABC,
 object]
```
So we are picking out the direct parent of `ConfigLoader`,  whereas we look at all ancestors of the class specified in settings.py. This means that if I make a `CustomConfigLoader` that inherits from `TemplatedConfigLoader`, it will still pass the validation.

I actually originally wrote this to check for a common ancestor anywhere in the inheritance chain, like `set(default_class.mro()) & set(setting_value.mro()) - {abc.ABC, object}` but decided that was more confusing overall because you have to explicitly exclude some classes. Like you say, I think the assumption that `default_class`'s direct parent is `AbstractXXX` is unlikely to be false in the future. And if we do change it then the tests will fail straight away and we can change the validation code then.

RELEASE.md
```diff
@@ -17,6 +17,7 @@
 
 ## Bug fixes and other changes
 * Removed fatal error from being logged when a Kedro session is created in a directory without git.
+* Fixed `CONFIG_LOADER_CLASS` validation so that `TemplatedConfigLoader` can be specified in settings.py.
```
MerelTheisenQB, line 20: Might be helpful for future reference to mention how we fixed it, so something like fixed config_loader_class validation by checking for shared parent class, so that ...

tests/framework/project/test_settings.py
```diff
@@ -3,29 +3,63 @@
...
+    pass
+
+
+class ProjectHooks:
+    pass
+
+
 @pytest.fixture
 def mock_package_name_with_settings_file(tmpdir):
+    """This mock settings file tests everything that can be customised in settings.py.
```
MerelTheisenQB, line 27: This is great to test settings properly 👍

MerelTheisenQB (approved):
This is an excellent solution! Very elegant 👏  I agree it would be useful to add a comment or two explaining what happens exactly with `mro()[1]`, especially for future team members. 

Great work @AntonyMilneQB !

RELEASE.md
AntonyMilneQB: Done!

kedro/framework/project/__init__.py
noklam: Yes, I agree just looking 1 layer up is sufficient.

PR #1480, opened 2022-04-25
Adding a new after context created hook into kedros hooks collection

### LINKED ISSUE #1458: Adding a new `after_context_created` hook into Kedro's hooks collection
# Description
Related: #506 
Prerequisite of: #1431 
The new hook `after_context_created` will be the solution for #506, by exposing the `context`, it will provide a public API for developers to access `catalog`, `config_loader`, `credentials` etc. The `KedroContext` will gradually become a container while the `KedroSession` controls the lifecycle of Kedro's run

todo:
- [ ] Create new `after_context_created` hook
- [ ] Update Runners
- [ ] Update corresponding tests
- [ ] Add the lifecycle hooks chart to illustrate the order of hook being executed (seems datasets specs is missing from test previously)

noklam:
## Description

Close #1458 

## Development notes

* New hook `after_context_created`
* Add test for the new hook
* Patch tests for `DatasetSpecs` 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1480"></a>


noklam:
The linting is breaking but I don't know how to suppress it. This is caused by `KedroContext` being imported as an argument signature and thus breaking the contract.

image

pyproject.toml
```diff
@@ -72,7 +72,8 @@ layers = [
     "config"
 ]
 ignore_imports = [
-    "kedro.runner.parallel_runner -> kedro.framework.project"
+    "kedro.runner.parallel_runner -> kedro.framework.project",
+    "kedro.framework.hooks.specs -> kedro.framework.context.context"
```
noklam, line 76: This is added to avoid the importool linting error.

AhdraMeraliQB (approved):
Nicely done! 👍

MerelTheisenQB:
I think the error for the circular dependency that the linter was throwing is actually legit, and we should be calling the hook at the point in the session where the `context` is loaded from settings.

noklam:
@MerelTheisenQB The linter was throwing an error in `specs.py` where I have `KedroContext` imported as a type hint for `after_context_created`. Is there a way to get rid of it?

I agree with you and I move the hook into `load_context` now. Originally I was thinking that there may be cases where we still want `after_context_created` hook even if we are not inside a `kedro run` (e.g. some hooks/plugins on just data catalog). Now I think even in those cases we should still create a session and use `load_context` instead of manually creating one.

RELEASE.md
```diff
@@ -12,6 +12,7 @@
 # Upcoming Release 0.18.1
 
 ## Major features and improvements
+* Added a new hook `after_context_created`.
```
AntonyMilneQB, line 15:
```suggestion
* Added a new hook `after_context_created` that passes the `KedroContext` instance as `context`.
```

kedro/framework/hooks/specs.py
```diff
@@ -4,6 +4,7 @@
 """
 from typing import Any, Dict, Optional
 
+from kedro.framework.context.context import KedroContext
```
AntonyMilneQB, line 7:
```suggestion
from kedro.framework.context import KedroContext
```

kedro/framework/hooks/specs.py
```diff
@@ -279,3 +280,20 @@ def after_dataset_saved(self, dataset_name: str, data: Any) -> None:
             data: the actual data that was saved to the catalog.
         """
         pass
+
+
+class KedroContextSpecs:
+    """Namespace that defines all specifications for a kedro context's lifecycle hooks."""
```
AntonyMilneQB, line 286:
```suggestion
    """Namespace that defines all specifications for a Kedro context's lifecycle hooks."""
```

kedro/framework/hooks/specs.py
```diff
@@ -279,3 +280,20 @@ def after_dataset_saved(self, dataset_name: str, data: Any) -> None:
...
+
+    @hook_spec
+    def after_context_created(
+        self,
+        context: KedroContext,
+    ) -> None:
+        """Hooks to be invoked after a `KedroContext` is created. This is the earliest
+        hook trigger within a kedro's run. The `KedroContext` stores useful information
+        such as `credentials`, `config_loader` and `env`.
+        .
```
AntonyMilneQB, line 296:
```suggestion
        hook triggered within a Kedro run. The `KedroContext` stores useful information
        such as `credentials`, `config_loader` and `env`.
```

AntonyMilneQB (commented):
I feel like there should be more tests here somewhere? e.g. if you look at tests for `after_catalog_created` I think there's a few more which would probably be suitable to duplicate for this new hook.

AntonyMilneQB:
Also, I think this is indeed the right place for the hook call now and your fix for the import linting looks good to me 👍

kedro/framework/hooks/specs.py
noklam: Ops I think it is actually 
`from kedro.framework.context.context import KedroContext`

https://github.com/kedro-org/kedro/blob/main/kedro/framework/context/context.py

MerelTheisenQB (approved):
I just have one question regarding test setup, but other than that I'm happy with this implementation 👍 

The linting fix makes sense as well. I originally thought it checked for circular dependency, but it checks for a certain hierarchy, which makes sense to ignore in this case.

tests/framework/session/test_session_extension_hooks.py
```diff
@@ -560,3 +563,45 @@ def test_after_dataset_load_hook_async(
...
+    """Test the behavior of `after_context_created` when updating node inputs."""
+
+    def test_attributes_are_accessible(
+        self,
+        mock_session_with_after_context_created_hook,
+        logs_listener,
+    ):
+        mock_session_with_after_context_created_hook.load_context()
+        hooks_log_messages = [r.message for r in logs_listener.logs]
+        assert "After context created" in str(hooks_log_messages).strip("[]")
```
MerelTheisenQB, line 607: I'm a little confused about this setup, do we expect this to use both the `after_context_created` hook implementation from `conftest` as well as the one above here? The name of the test makes me think we should be checking that we can indeed access the attributes, but this check just looks at the log message.

AntonyMilneQB (approved):
Also happy to approve this assuming the tests can be fixed 🙂

Issue #1470, opened 2022-04-22
Alter framework logging.yml configuration


AntonyMilneQB:
Various small changes to make to the framework kedro/config/logging.yml (**not to the project conf/base/logging.yml**):
* delete the section `root`
* add the `kedro.io` and `kedro.pipeline` that are in project conf/base/logging.yml to the framework kedro/config/logging.yml 
* add the following to the loggers:
```
kedro:
  level: INFO
  handlers: [console, info_file_handler, error_file_handler]
  propagate: no
```

Do some kedro runs to make sure things look ok. I'm pretty sure they will.

MerelTheisenQB:
Might fix the `kedro-airflow` problem, double check this is the case.

AntonyMilneQB:
Slightly different, new and improved version of this in https://github.com/kedro-org/kedro/pull/1491/commits/83b074f586932c7c790afde9b174b33607ec50f1:
* keep the `root` logger but no longer change its level to `INFO` (it will keep the default of `WARNING`)
* `kedro` logger just changes the level to `INFO`, which propagates to the root logger
* the only special case is `kedro.framework.cli.hooks.manager`, where we don't want to write to a file

Issue #1474, opened 2022-04-22
Write documentation on how to make a custom logging.yml in your project


AntonyMilneQB:

datajoely:
So much like the project CLI being removed from the template I'm fine with this living docs. 

However what do we think about a `kedro template {thing}` command?

We could use it to create templates for custom CLI, logging, datasets etc. a bit like we do for `kedro pipeline create`.

Thoughts?

noklam:
I think it could help but I feel it is something that falls into the "nice to have" category. Although looking at the statistics seems people do use `kedro pipeline create`? Maybe I am wrong here.

I just found that the `cli.py` is also outdated, maybe another sign that not many people are actually using it.

datajoely:
`kedro pipeline create` is the best way to create pipelines, the ideal above is very much a 'nice to have'.

`cli.py` is a funny one as I think we've made it quite hard to change, I'm not sure it's super well documented how you create your own. Like I would **expect to see it here**, but it's not there.

noklam:
It's actually [here](https://kedro.readthedocs.io/en/stable/extend_kedro/common_use_cases.html?highlight=cli.py#use-case-3-how-to-add-or-modify-cli-commands). Actually, I am only able to search it with the keywords `cli.py`, I suspect a normal user will struggle to find this information.

I am not sure how many projects will need the `cli.py` for a project base usage, plugins are probably the way to go for cross-projects.

Issue #1461, opened 2022-04-20
Improve logging setup in Kedro


idanov:
Currently Kedro has some issues with the logging setup, which can cause some problems from minor annoyances to failures to run a Kedro project in certain environments. Here is a small list of the problems our current logging setup causes:

- Kedro logs at root level, rather than `kedro.*` in many places
- `root` log level is set to `INFO` instead of `WARNING` on all official starters
- on read-only file systems Kedro projects fail to execute, rather than just failing to write the logs
- `KedroSession` logs useless things at `INFO` level, like e.g. that the default session store doesn’t have a `read()` method and `git describe` failures…
- maybe I am mistaken, but I think Kedro logs things before the logging config is setup (to be confirmed).
- CLI commands output too many things, e.g. when you do `kedro catalog list` it should only show the output so it can be machine readable, but now there's lots of `DEBUG` level worthy output
- kedro leaves `info.log` files everywhere for unknown reasons (easiest to notice is when runing a notebook in `notebooks/`)
- https://github.com/kedro-org/kedro-plugins/issues/13

In order to address those issues, we need to work on the following:

- [x] #1472
- [x] #1470
- [x] #1469
- [x] #1473
- [x] #1474
- [x] ~Remove file handlers, special case `kedro.framework.cli.hooks.manager` handler and add documentation on how to setup logging with them (**only if #1472 shows we should remove them**)~
- [x] #1558
- [x] #1512
- [x] #1513
- [x] #1514
- [x] #1575
- [x] #1515
- [x] https://github.com/kedro-org/kedro-plugins/issues/31
- [x] #1683
- [ ] Consider whether to set `spaceflights` log level to `INFO` on framework-side rather than project. Like point 2 of https://github.com/kedro-org/kedro/issues/1469. Needs to use `PACKAGE_NAME`?
- [ ] Work out whether we should have a verbosity flag to change logging level (note we already have a verbosity option on some commands, but it looks like it only changes how detailed `KedroCliError` message is). Otherwise there's no way you will see `DEBUG` level messages emitted by logging calls before `session._setup_logging` is called, even if you alter your project logging.yml to show `DEBUG` level messages. e.g. https://hydra.cc/docs/tutorials/basic/running_your_app/logging/
- [ ] https://github.com/kedro-org/kedro/issues/1464
- [ ] kedro-org/kedro/issues/1728 - should rich stuff be moved to hooks anyway? See also my thoughts on spark hooks

[...]

PR #1491, opened 2022-04-29
Alter framework logging.yml configuration

AhdraMeraliQB:

## Description

Adds the changes outlined in #1470 to update the inbuilt logging setup.

## Development notes

The changes were tested by running some cli commands and noting the changes in the output between this and the unchanged version, and between valid and invalid pipelines. The only divergence in output was noted when running `kedro ipython`

**Original output**
```
...
Python 3.7.11 (default, Jul 27 2021, 07:03:16) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.32.0 -- An enhanced Interactive Python. Type '?' for help.
2022-04-29 13:59:26,965 - root - INFO - Updated path to Kedro project: /Users/ahdra_merali/Desktop/kedrotest
2022-04-29 13:59:27,249 - kedro.framework.session.store - INFO - `read()` not implemented for `BaseSessionStore`. Assuming empty store.
2022-04-29 13:59:27,299 - kedro.framework.session.session - WARNING - Unable to git describe /Users/ahdra_merali/Desktop/kedrotest
2022-04-29 13:59:27,600 - numexpr.utils - INFO - NumExpr defaulting to 8 threads.
2022-04-29 13:59:28,576 - root - INFO - ** Kedro project kedrotest
2022-04-29 13:59:28,576 - root - INFO - Defined global variable `context`, `session`, `catalog` and `pipelines`
2022-04-29 13:59:30,174 - root - INFO - Registered line magic `run_viz`
```

**Updated output**
```
...
Python 3.7.11 (default, Jul 27 2021, 07:03:16) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.32.0 -- An enhanced Interactive Python. Type '?' for help.
WARNING:kedro.framework.session.session:Unable to git describe /Users/ahdra_merali/Desktop/kedrotest
2022-04-29 13:47:55,995 - numexpr.utils - INFO - NumExpr defaulting to 8 threads.
2022-04-29 13:47:56,921 - root - INFO - ** Kedro project kedrotest
2022-04-29 13:47:56,921 - root - INFO - Defined global variable `context`, `session`, `catalog` and `pipelines`
2022-04-29 13:47:58,538 - root - INFO - Registered line magic `run_viz`
```

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1491"></a>


kedro/config/logging.yml
```diff
@@ -33,15 +33,21 @@ handlers:
 
 loggers:
     anyconfig:
-        level: WARNING
-        handlers: [console]
+        level: INFO
```
noklam, line 36: Why `anyconfig` level is changed from `WARNING` to `INFO`?

AhdraMeraliQB: Oops - thanks for the catch!

AhdraMeraliQB:
@AhdraMeraliQB 
> I am a bit surprise the `kedro.framework.session` related log seems missing after the changes?

As am I 🤔 ; converting to draft to address the e2e tests and investigate further

AntonyMilneQB:
Thanks for all very thorough testing and observations noted here.

I've written a comment in https://github.com/kedro-org/kedro/issues/1461 to explain better how kedro logging works so that everyone can understand this better, because it is quite confusing. 

Given the above let me explain the differences that I see here (please do say if I missed any!):
1. `root - INFO - Updated path to Kedro project` message has disappeared
2. ``kedro.framework.session.store - INFO - `read()` not implemented for `BaseSessionStore` `` has disappeared 
3. `kedro.framework.session.session - WARNING - Unable to git describe`  is still there but looks gross

All the above logs are emitted before the project-side logging.yml is loaded and hence just the framework-side logging.yml is used, which is why changing the project-side logging.yml has changed them. When you removed our custom `root` logger, it falls back on the Python default one which only emits messages of level `WARNING` and above. This explains the disappearance of 1 and 2. The Python default logger looks gross which explains 3.

Now, which (if any) of these are problems and do we need to rethink what we're doing here? I think we're on the right lines and this is ok so far, but once I've had a chance to think a bit more about it later this morning I will say if that's not the case!

noklam:
Follow up on @AntonyMilneQB comments about the 3 changes.

1. I think it is ok to not show this message, if we ever want to see this message we need to set the root logger to INFO level
2. This should be disappear but I think not for the right cause. It should be a DEBUG level message.
3. I think we may missed some configuration. The format of this warning message is off. Maybe we need to have a kedro library level logging for anything not specified. Same apply to point 2, I suspect if it is a warning level log it will also be printed in the wrong format.

noklam:
Thanks, I think changing the hook specific logger to `kedro` logger is the right thing to do!

AntonyMilneQB:
Now that https://github.com/kedro-org/kedro/pull/1506 has been merged I've tested this again and all behaves as expected:

```
➜ kedro ipython
2022-05-09 10:41:38,334 - kedro.framework.cli.hooks.manager - INFO - Registered CLI hooks from 1 installed plugin(s): kedro-telemetry-0.2.0
Kedro-Telemetry is installed, but you have opted out of sharing usage analytics so none will be collected.
ipython --ext kedro.extras.extensions.ipython
Python 3.8.13 (default, Mar 28 2022, 06:16:26) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.33.0 -- An enhanced Interactive Python. Type '?' for help.
2022-05-09 10:41:42,257 - kedro.extras.extensions.ipython - INFO - Updated path to Kedro project: /Users/antony_milne/kedro_stuff/projects/spaceflights
2022-05-09 10:41:42,499 - kedro.framework.session.store - INFO - `read()` not implemented for `BaseSessionStore`. Assuming empty store.
2022-05-09 10:41:42,569 - kedro.framework.session.session - WARNING - Unable to git describe /Users/antony_milne/kedro_stuff/projects/spaceflights
2022-05-09 10:41:43,464 - kedro.extras.extensions.ipython - INFO - ** Kedro project spaceflights
2022-05-09 10:41:43,464 - kedro.extras.extensions.ipython - INFO - Defined global variable `context`, `session`, `catalog` and `pipelines`
2022-05-09 10:41:44,762 - kedro.extras.extensions.ipython - INFO - Registered line magic `run_viz`
```

Marking this as ready to review since I think it we're all good here now and it would be nice to get it merged in so we can make progress on the following tickets that depend on it 🙂

noklam (approved):
Thank you! The log now looks good to me. Good to be merged after fixing the tests.

pyproject.toml
```diff
@@ -39,7 +39,7 @@ min-public-methods = 1
 [tool.coverage.report]
 fail_under = 100
 show_missing = true
-omit = ["kedro/templates/*", "kedro/framework/cli/hooks/specs.py", "kedro/framework/hooks/specs.py", "kedro/extras/datasets/tensorflow/*", "kedro/extras/datasets/holoviews/*", "tests/*"]
+omit = ["kedro/templates/*", "kedro/extras/logging/color_logger.py", "kedro/framework/cli/hooks/specs.py", "kedro/framework/hooks/specs.py", "kedro/extras/datasets/tensorflow/*", "kedro/extras/datasets/holoviews/*", "tests/*"]
```
AntonyMilneQB, line 42: I've added the color logger here. I actually have no idea why we were getting 100% coverage before because this file has never been fully tested... But given that we're going to deprecate it I decided just to add it to the coverage ignore rather than write tests for it.

kedro/config/logging.yml
```diff
@@ -32,16 +32,13 @@ handlers:
...
-        propagate: no
-
     kedro.framework.cli.hooks.manager:
         level: INFO
         handlers: [console]
         propagate: no
 
+    kedro:
+        level: INFO
+
```
MerelTheisenQB, line 42: Originally in https://github.com/kedro-org/kedro/issues/1470 we needed to add `handlers` and `propagate` to all loggers. Why is this no longer needed? And what exactly do these things do? I still find the logging behaviour a bit mysterious 😅

MerelTheisenQB (approved):
One question to understand the changes better, but not blocking and I'm happy with merging this!

pyproject.toml
noklam: Isn't there a `test_color_logger.py`? It's probably not related but seen this from Kiyo yesterday. https://921kiyo.com/code-coverage-level-with-pytest-cov/

kedro/config/logging.yml
AntonyMilneQB: Yeah, I only fully understood this after reading the logging docs for the 1000th time last week 😀 This will become clearer with examples when we complete #1474 and I'm happy to present to the team to explain further, but in brief:

* loggers work hierarchically and activate all loggers above them. e.g. when `kedro.x.y` does logging it will also log to any loggers named `kedro.x` and `kedro`. `root` is the at the top of the hierarchy
* the above is the case when `propagate: true` (the default if `propagate` is not specified explicitly). If you set `propagate: false` then a logging call to `kedro.x.y` will not activate any other loggers
* if you propagate to higher up loggers then if they both have the same handler, this handler will be called separately by each logger that uses it. This results in duplicated log messages
* our general strategy according to the Python docs should be to delegate as much as possible to the highest up logger (so `root` if possible), attach handlers to that logger and then allow lower level handlers just to propagate their calls up the chain

The [docs on `propagate`](https://docs.python.org/3/library/logging.html#logging.Logger.propagate) are quite clear and helpful for explaining all this more.

In practice, what happens now is that any logging done anywhere inside kedro will be propagated up to the root logger, which attaches the required handlers. i.e. if `kedro.x.y` then it will check if there's any logger defined explicitly for `kedro.x.y` (there isn't), then it will check for `kedro.x` (there isn't), then `kedro` (which is defined). `kedro` doesn't define any handlers, so it gets propagates up to `root`, which is the thing that actually calls the handlers. The point in defining `kedro` as a logger is just so we can set `level: INFO` so that any info log messages coming from kedro are shown. The default level set by `root` (and hence by all other logging calls outside `kedro`) is WARNING.

`kedro.framework.cli.hooks.manager` is a weird special case that will hopefully disappear in the near future if we remove the file handlers from the root logger.

pyproject.toml
AntonyMilneQB: There is, but it only tests how the color logger handles `INFO` level messages. Ideally it would go through every level of logging and test them all. I don't know how it was passing coverage before given none of the other levels are tested 🤔

AntonyMilneQB: Interesting post on the different coverage levels, I didn't know about that before! I wonder what we'd get if we added `--cov-branch` to our tests.

kedro/config/logging.yml
MerelTheisenQB: Thanks for this explanation @AntonyMilneQB! It makes a lot more sense now 😄

PR #1489, opened 2022-04-27
Migrate deprecated `recommonmark` to `myst-parser`

deepyaman:
## Description

Close #1479 

Also caught a lot of broken links, since MyST-Parser doesn't like them, whereas `recommonmark` seems not to care. The unfixed links don't work on the current Kedro website, so I feel like this is a bit more urgent. :)

## Development notes

`make build-docs && open docs/build/html/index.html` looks beautiful

## Checklist

- [x] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [ ] Added tests to cover my changes

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1489"></a>

Issue #1293, opened 2022-02-25
[KED-2346] Separate out dev requirements in kedro project template


AntonyMilneQB:
As per discussions https://github.com/kedro-org/kedro/issues/1077 and https://github.com/kedro-org/kedro/discussions/844 we want to separate dev requirements into their own file outside requirements.txt. This means _any requirements that are not required for `kedro run` or running a packaged kedro project (which only exposes `kedro run`)._ Remember in v0.18 there is no `kedro install` command.
 * Change project template to move requirements from requirements.txt into the new dev requirements file. Also move docs requirements from setup.py extras_require to there
 * Modify kedro CLI commands that currently install requirements (like `kedro build-docs`) to do the right thing
 * Modify kedro CLI commands that need dev requirements but don't currently install them (like `kedro lint`) do the right thing
 * Change starters templates in same way
 * Update documentation

Things to check and think about:
 * Would all `kedro lint`, `kedro build-docs` etc. commands need to run a pip install to install dev requirements? Is this annoying and slow? Is Waylon's idea of pre-commit a better alternative?
 * Assuming we want a dev requirements file, what is the right name for it? Use whatever is convention these days
 * Do commands like `kedro pipeline`, `kedro catalog` have any dependencies outside kedro, and if so where should those dependencies go (requirements or dev requirements)?
 * What should `kedro build-reqs` do?

@idanov commented:
We might need to add a general recommendation on how to use local wheel files with Kedro, so all CLI commands work properly.

@AntonyMilneQB commented:
See https://github.com/quantumblacklabs/private-kedro/issues/1288

Also worth checking whether any kedro requirements should be moved to project requirements. e.g. pip-tools is there but after deprecation of `kedro install` can presumably be moved to a project requirement? Do we need jupyter_client in kedro requirements? etc.

yetudada:
I have been thinking about this ticket because I think we should base it on how useful these dependencies are for users. Here's how our users have used CLI-commands which have dependencies in the project's `requirements.txt`, since the 1st of September 2021:

|  CLI Command | Unique Runs | Dependencies |
| --- | --- | --- |
| `kedro lint` | 27 | `black==21.5b1`, `flake8>=3.7.9, <4.0` and `isort~=5.0` |
| `kedro test` | 47 | `pytest-cov~=3.0`, `pytest-mock>=1.7.1, <2.0` and `pytest~=6.2` |
| `kedro activate-nbstripout` | 0 | `nbstripout~=0.4` |
[...]

PR #1503, opened 2022-05-04
Migrate starter template docs from recommonmark to  myst_parser

### LINKED ISSUE #1479: Migrate from recommonmark to MyST
## Description
We use recommonmark plugin to write docs in markdown format, but this package has been deprecated and archived. The recommendation is to use [MyST](https://github.com/executablebooks/MyST-Parser) instead.

See https://github.com/readthedocs/recommonmark/issues/221 for more info.

## Context
Using mermaid in our docs as proposed by #1392 and any other future changes might require a switch to `MyST` as recommended by [sphinxcontrib-mermaid](https://github.com/mgaitan/sphinxcontrib-mermaid#markdown-support). 

Trying an older suggestion `sphinxcontrib-mermaid` (version 0.7) doesn't work, so it might still require the migration to MyST.

avan-sh:
## Description

PR #1489 addresses the bulk of the changes required for migrating to myst-parser in docs, this pr modifies `conf.py` in templates which is used by `kedro build-docs` command in new projects.

Closes #1479

**Note**: While this is likely to not be breaking for new projects, docs copied directly from older projects might need to be updated.

## Development notes
- Modified  `docs/conf.py` & `setup.py` to remove references to `recommonmark` and update it to `myst_parser`.
- Updated test for `kedro build docs`.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1503"></a>


AntonyMilneQB (approved):
Ah nice! I was just saying to @MerelTheisenQB yesterday that we should make this change but might not get around to it since probably we should reconsider the whole `kedro build-docs` command, whether to continue using sphinx, update other bits in conf.py because it's very old...

Anyway, this definitely seems like a good improvement to the current setup, thank you! Please can you add a note to RELEASE.md and also make the change to kedro-starters?

avan-sh:
> Ah nice! I was just saying to @MerelTheisenQB yesterday that we should make this change but might not get around to it since probably we should reconsider the whole `kedro build-docs` command, whether to continue using sphinx, update other bits in conf.py because it's very old...

@AntonyMilneQB Is this just about how docs are built within a kedro project? If this is already on cards, this change might not be worth it as introducing two breaking changes in succession might not be worth it.

I could also take a crack at the refactor/clean-up if you could elaborate a bit on what/why we're looking to do.

AntonyMilneQB:
@avan-sh Yeah, this is just about how docs are build within a kedro project. I'd say that future changes are not imminently on the cards, so I think it's still worth making this change now. 

Given that this change doesn't actually touch the `kedro build-docs` command and is just a small change to the template I don't think it's really a breaking change, because `kedro build-docs` will still work the same way on projects using an old template. Also the subset of users who use `kedro build-docs` + use markdown documentation + embed RST in their markdown is so small anyway - it's kind of a weird documentation setup that we have on kedro and just got copied to the project template.

Equally you could argue that the number of affected users is so small that it's just not worth the bother of updating it, since no one has noticed that the project template uses an old sphinx extension...

> I could also take a crack at the refactor/clean-up if you could elaborate a bit on what/why we're looking to do.

Thank you for the offer but I think it's a bit premature to actually do anything about this yet. I was just mentioning it as a vague future plan - we still need to check telemetry data, do user research, etc. Probably needs to be considered as part of the whole vaguely-planned streamlining of the CLI and https://github.com/kedro-org/kedro/issues/1293.

deepyaman:
> I'd say that future changes are not imminently on the cards, so I think it's still worth making this change now.
👍 

> Given that this change doesn't actually touch the `kedro build-docs` command and is just a small change to the template I don't think it's really a breaking change, because `kedro build-docs` will still work the same way on projects using an old template.
:+2:

PR #1507, opened 2022-05-05
Fix CI issues with python 3.10 builds

MerelTheisenQB:
## Description
The python 3.10 builds all started failing because pip couldn't finish resolving dependencies.

## Development notes
I've tried pinning and updating several package versions. Eventually pinning `nbconvert~=6.0` and `nbformat~=5.0` seemed to do the trick.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1507"></a>

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 
-#  the default pipeline parameters, which will be updated according to
+# the default pipeline parameters, which will be updated according to
 # the results of the path-filtering orb
 parameters:
   code_change:
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 
-# the default pipeline parameters, which will be updated according to
+#  the default pipeline parameters, which will be updated according to
 # the results of the path-filtering orb
 parameters:
   code_change:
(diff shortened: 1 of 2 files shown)


AntonyMilneQB (approved):
If it fixes CI it always gets 👍  from me 😀 

But I'm a bit surprised these files are in test_requirements.txt at all since they don't seem to be listed anywhere else, e.g. in the project requirements. Presumably they're dependencies of some other packages already, so would it work just to remove them from test_requirements.txt altogether?

noklam (approved):
Oh wow, now every CI is passing? Amazing...

I am quite surprise it also fix the numpy build issue.

MerelTheisenQB:
> If it fixes CI it always gets 👍 from me 😀
> 
> But I'm a bit surprised these files are in test_requirements.txt at all since they don't seem to be listed anywhere else, e.g. in the project requirements. Presumably they're dependencies of some other packages already, so would it work just to remove them from test_requirements.txt altogether? And then pip can hopefully work out what version to install.

I'll see what happens if I remove these requirements!

MerelTheisenQB:
Even better solution in the end: just remove `nbformat` and `nbconvert` from `test_requirements.txt` and let pip install them as dependencies of other packages if needed.

PR #1508, opened 2022-05-05
Tidy setup.py

AntonyMilneQB:
## Description
A very small bit of housekeeping while I notice it...

`profilers` should really have been removed when we removed the `extras.transformers`. Technically this is breaking because you can no longer do `pip install kedro[xxx]` for these things I've removed, but I'm absolutely sure that no one uses them.

This means we only have dataset dependencies + a slightly weird docs one in setup.py now, so that moving to `kedro-datasets` will be simpler.

## Checklist


SajidAlamQB (approved):
Nice catch. 👍

MerelTheisenQB (approved):
Great cleanup! 🧹 ✨

noklam (approved):
Nice!

Issue #1476, opened 2022-04-25
[spike] Improve Kedro CLI startup time


noklam:
## Description

Currently, Kedro CLI startup time can be slow even if very little work is done, for example, doing `kedro` to show the available command or help can take a few seconds. These commands should feel instant since it is just printing out some help text without any heavy work.

# Useful tool to profile import time
* https://github.com/asottile/importtime-waterfall
* `python -X importtime run.py`

1. The `kedro` command itself takes 0.7s, which isn't fast but still an acceptable response time. Initial investigation suggests that it can take few seconds mainly due to a slow import during plugin registration.

## Kedro startup time with plugin installed
(0.7s, removed screenshot)

## Kedro startup time with no plugin installed (telemetry + viz)
<img width="957" alt="image" src="https://github.com/kedro-org/kedro/assets/18221871/b8a3a5dc-5c24-4847-88c7-63125e4b819d">

Causes:
1. Slow import from plugin
2. Slow import from `kedro` itself
3. Slow import from `pkg_resource` (something that is run with `console_script`)
  * https://stackoverflow.com/questions/53223214/slow-response-of-cli-written-with-python-click
  * https://github.com/pypa/setuptools/issues/510

## Context
Why is this change important to you? How would you use it? How can it benefit other users?

The Kedro CLI can be slow sometimes and lead to a bad user experience. It should be the most common way that users interact with kedro.

## Possible Implementation
(Optional) Suggest an idea for implementing the addition or change.
The first improvement that can be done is to do lazy loading in plugin registration to address (1). (2) & (3) will requires more time to study.

## Possible Alternatives
(Optional) Describe any alternative solutions or features you've considered.
Compile some of the kedro CLI, so it prints out help message without actually loading any python code.

e.g.
`kedro` -> Load some pre-compiled text files and print, it may be tricky with plugins though.

MerelTheisenQB:
Some older issues about this topic: https://github.com/quantumblacklabs/private-kedro/issues/930 & https://github.com/quantumblacklabs/private-kedro/pull/537

znfgnu:
I created PR that should solve this problem in kedro-org/kedro-viz#1196 - any help with making tests passed would be appreciated

astrojuanlu:
(From #3033)

Still an issue. For example, `kedro info` on a `spaceflights` project takes a bit more than 6 seconds on my machine (M1 MacBook Pro from 2022):

```
$ time kedro info
 _            _
[...]

PR #1542, opened 2022-05-16
Handle plugin registration failure ContextualVersionConflict with log instead of raising error

### LINKED ISSUE #1487: ContextualVersionConflict Error Stop Kedro running when dependency clashes
## Description

ContextualVersionConflict error arises on two ocassions when using Kedro :-
- when you install a Kedro plug-in with a dependency that clashes with a kedro, or a kedro-viz dependency
- when you do %reload_kedro on Databricks 

## Steps to Reproduce on Databricks
1. clone a kedro project on github
2. make sure you have kedro-viz installed on Databricks in the same place as kedro (i.e. Cluster Library or using pip command) 
3. load kedro IPython extension 
4. try using %reload_kedro. line magic command

## Steps to Reproduce outside Databricks 
1. for a kedro-plugin, in the setup.py add to install_requires a dependency that clashes with kedro
2. pip install the kedro-plugin
3. do a kedro run when in the project folder 

## Actual Result
image (4)

## Expected Result
Instead of an error message which breaks Kedro, it should just display a warning maybe? 

## Your Environment

* Kedro version used (`pip show kedro` or `kedro -V`): 0.18
* Python version used (`python -V`): 3.8
* Operating system and version:  Linux

noklam:

## Description

Fix #1487 

Currently, when using `kedro` with plugins, the conflict version between libraries will stop kedro completely to load. The more ideal way is to ignore the plugins and throw warnings.

These errors can at least happen in 3 different ways.

1. Conflict library version specified in `kedro` itself. This could happen if you install kedro, then downgrade `pip`. In theory, if you use Python API it will run successfully, but anything involving the entry points will fail. The main reason is that `pkg_resource` always validates the dependencies, and there seems to be no easy way to fix it.
2. Conflict library version coming from `kedro-xxxxx` plugins. This is the direct cause of the creation of this PR, the less intrusive way to deal with it is just to log the error instead of terminating the entire program. I choose to use a higher level of LOG level as for certain plugins, it could mean the run is actually no longer valid. (i.e. `kedro-mlflow` or any plugin that actually has additional behavior for a kedro run. 
3. Load entry points when we want to call the `__main__` of a kedro package programmatically.

## Development notes

1. This PR replaces the usage of `pkg_resource` with `importlib-xxx` for a few reasons.
1.1 The `pkg_resource` is an old library and maybe deprecated [soon](https://github.com/pypa/setuptools/milestone/4), even the official website of `pkg_resource` suggest using `importlib-xxx` alternatives whenever possible. `pluggy` and `pytest` replaced `pkg_resource` with `importlib` too.
1.2. Improve import speed - the `pkg_resource` import lots of thing at the top-level, which slow down the CLI. [#1476](https://github.com/kedro-org/kedro/issues/1476)
1.3. Reduce the chance that the `ContextualVersionConflict` types of error. This is not solved unfortunately, it seems that this validation still happens if one of `kedro`'s dependencies uses `pkg_resource`. `dynaconf` is one of them, and importing `dynaconf` will trigger the same error.
2. When the plugin registry fails, `kedro` will log them instead of terminate the program.
3. Move `setuptools` to `pyproject.toml` ([PEP-518](https://peps.python.org/pep-0518/), not really related to the PR but found out this when studying how the python build system works.)
4. ❌ `micropkg` is still using `pkg_resource` for requirement parsing, it is not handled now since it is out of scope.

# Compatibility issue:
https://bugs.python.org/issue44246
Unfortunately the entry_points() API is not consistent between Python version, `importlib.metadata` is only available with python>=3.8. I spend quite a while to figure this out and make all test passed. For now, the backward compatible `importlib_metadata` seems to be a better option, but it is not with the standard lib `importlib`.

This is how it looks if the plugin's registration fails. For example I explicitly downgrade `plotly` so `kedro-viz` is not registered successfully.
image

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1542"></a>


kedro/framework/cli/utils.py
```diff
@@ -328,13 +331,18 @@ def load_entry_points(name: str) -> Sequence[click.MultiCommand]:
...
     for entry_point in entry_points:
         try:
             entry_point_commands.append(entry_point.load())
-        except Exception as exc:
-            raise KedroCliError(f"Loading {name} commands from {entry_point}") from exc
+        except Exception as exc:  # pylint: disable=broad-except
+            logger.warning(
+                KedroCliError(f"Fail to load {name} commands from {entry_point}")
+            )
+            logger.warning(exc)
```
noklam, line 345: This is the most important part. It's hard to determine from the framework side whether a kedro's run is still valid if plugins fails. `kedro-viz` is defintely ok to just ignore it, but for `kedro-mlflow` type of plugin that actually modify the run behavior,  it may be problematic so I am using `warning` level.

pyproject.toml
```diff
@@ -1,3 +1,8 @@
+# PEP-518 https://peps.python.org/pep-0518/
+[build-system]
+# Minimum requirements for the build system to execute.
+requires = ["setuptools>=38.0", "wheel"]  # PEP 508 specifications.
+
```
noklam, line 5: This is something I just come across when I work on this PR, no related to the PR particularly.

noklam:
I still have to make it work for py37... but feel free to drop comments.

kedro/framework/cli/cli.py
```diff
@@ -65,10 +68,9 @@ def info():
     plugin_versions = {}
     plugin_entry_points = defaultdict(set)
     for plugin_entry_point, group in ENTRY_POINT_GROUPS.items():
-        for entry_point in pkg_resources.iter_entry_points(group=group):
+        for entry_point in importlib_metadata.entry_points().select(group=group):
```
AntonyMilneQB, line 71: Could this also raise some kind of exception? Or is it just the `entry_point.load` that we should wrap in the `try`?

kedro/framework/cli/cli.py
```diff
@@ -98,12 +100,13 @@ def docs():
...
-    for entry_point in pkg_resources.iter_entry_points(group=group):
+    for entry_point in importlib_metadata.entry_points().select(group=group):
         try:
             init_hook = entry_point.load()
             init_hook()
-        except Exception as exc:
-            raise KedroCliError(f"Initializing {entry_point}") from exc
+        except Exception as exc:  # pylint: disable=broad-except
+            logger.warning(KedroCliError(f"Fail to initialize {entry_point}"))
+            logger.warning(exc)
```
AntonyMilneQB, line 109:
```suggestion
            logger.warning(f"Failed to initialise %s. Full exception: %s", entry_point, exc)
```
Unless there's as good reason to use the `KedroCliError` still?

kedro/framework/cli/utils.py
AntonyMilneQB: Similar suggestion here to the other occurrence in cli.py.

kedro/framework/cli/cli.py
```diff
@@ -98,12 +100,13 @@ def docs():
 
 def _init_plugins():
     group = ENTRY_POINT_GROUPS["init"]
-    for entry_point in pkg_resources.iter_entry_points(group=group):
+    for entry_point in importlib_metadata.entry_points().select(group=group):
```
AntonyMilneQB, line 103: Would it make sense to refactor this function a bit so it uses `load_entry_points` in utils.py? At a glance they look very similar.

tests/framework/cli/test_cli.py
```diff
@@ -91,12 +91,12 @@ def test_print_version(self):
...
 
-    def test_info_contains_plugin_versions(self, entry_point, mocker):
-        get_distribution = mocker.patch("pkg_resources.get_distribution")
-        get_distribution().version = "1.0.2"
+    def test_info_contains_plugin_versions(self, entry_point):
+        entry_point.dist.version = "1.0.2"
         entry_point.module_name = "bob.fred"
 
         result = CliRunner().invoke(cli, ["info"])
+        print(result.output)
```
AntonyMilneQB, line 99:
```suggestion
```

pyproject.toml
AntonyMilneQB: We've got `"setuptools>=38.0"` and `wheel` requirements in various other places in the repo (just do a Ctrl+F to see them). Does this mean we can remove them from there?

If yes then that would be nice but a bit of a bigger change, so probably worth doing in a separate PR.

AntonyMilneQB (commented):
Looks good! Another ticket which turned out to be much harder than expected, great job tackling it ⭐ 

I think a more conventional way to deal with a backported package would be
* try to import `importlib.metadata` and only if that fails `import importlib_metadata`
* add `python_version < '3.8'` to `importlib_metadata` in requirements so that it's only installed if necessary

See, for example, https://github.com/pallets/click/pull/1890/files.

But I don't know if this is possible here because of the API change you mention 😬

kedro/framework/cli/cli.py
AntonyMilneQB: Though I guess this is just `kedro info` rather than being triggered on every single `kedro` command like the other instances so it doesn't matter if there's uncaught exceptions anyway.

noklam: This is indeed the solution I go with initially, but later I find out the API was not consistent. So technically only Python 3.10 can use the stdlib.

image

noklam: It could fail, but like you said it should be just `kedro info`, unlike other plugin where the program still run without loading it. I think it is ok to leave it.

pyproject.toml
noklam: good point, created a new PR and will revert this soon.

kedro/framework/cli/cli.py
AntonyMilneQB: Urghh, what a pain. I think what you did in https://github.com/kedro-org/kedro/commit/3c12bccce22127191c9e802ff0f85e0e3afeff8b was ok, although probably it's more usual to just repeat the conditional import in multiple files.

BUT if the standard library API is only right in Python 3.10 then let's just forget about it and go for `importlib_metadata` all the way like you're doing now 👍

noklam:
Couples notes for the reviews:
* refactor the entry_points loop for `load_entry_points()` and `_init_plugins()`
* Revert PEP-518 changes and do it in a separate PR instead.
* Consider just log the error instead of using `KedroCLIError`
* Handle the conditional import for `importlib` better so it does not need to install a library for py310

kedro/framework/cli/cli.py
noklam: I have considered that, but it also means we have to copy the same block in multiple files, and also anywhere that use `mocker.patch`, which is quite hard to read.

For the context, I started this PR with `importlib_metadata`, so I wasn't aware of the inconsistent API between the Python version. In theory, if I avoid using the `select` API, it could be compatible with python3.8-3.10. But for both case, we still need to have the conditional import, so I would rather just stick with `importlib_metadata`, unless this extra dependencies is causing trouble.

AntonyMilneQB: Yeah, that totally makes sense. I'm definitely happy to go with `importlib_metadata` throughout.

kedro/framework/cli/cli.py
noklam: The original test for `_init_plugins` is actually incorrectly testing the `load` exception, it should be testing the error during hook initialisation instead.
```
init_hook = entry_point.load() # previously the the error was triggered here
init_hook()  # this is what we want to test instead
```

kedro/framework/cli/cli.py
```diff
@@ -96,14 +98,15 @@ def docs():
...
-    for entry_point in pkg_resources.iter_entry_points(group=group):
+def _init_plugins() -> None:
+    init_hooks = load_entry_points("init")
+    for init_hook in init_hooks:
         try:
-            init_hook = entry_point.load()
             init_hook()
-        except Exception as exc:
-            raise KedroCliError(f"Initializing {entry_point}") from exc
+        except Exception as exc:  # pylint: disable=broad-except
```
AntonyMilneQB, line 106: Now you've made it clearer exactly what raises the exception here, I actually think we should remove the `try`/`except` altogether and just do a plain
```
for init_hook in init_hooks:
    init_hooks()
```

Reasoning:
* exceptions raised by `entry_point.load()` are now handled by `load_entry_points`
* exceptions raised by `init_hook()` should still be raised, not hidden and logged, and there's not much point wrapping them in `KedroCliError` like we used to. This is analogous to what happens e.g. if a plugin with a hook raises an exception - it doesn't get caught and logged or converted into a kedro error message

All this is kind of minor because realistically no one uses `init_hooks` and it's probably redundant now we have `before_command_run` anyway... But if it's not too annoying to chance I think this would be a small improvement.

requirements.txt
```diff
@@ -5,6 +5,7 @@ cookiecutter~=1.7.0
 dynaconf>=3.1.2,<4.0.0
 fsspec>=2021.4, <=2022.1
 gitpython~=3.0
+importlib_metadata >=4.4 # Compatible with Python 3.10 importlib.metadata
```
AntonyMilneQB, line 8:
```suggestion
importlib_metadata>=4.4 # Compatible with Python 3.10 importlib.metadata
```

AntonyMilneQB (approved):
Really great work! Just one small comment. And also don't forget to add to RELEASE.md.

kedro/framework/cli/cli.py
noklam: Good catch, I think it is actually wrong to catch the error, error should be raised!

PR #1543, opened 2022-05-16
 Ensure `logging.yml` is optional and tidy logging config

### LINKED ISSUE #1469: Ensure `logging.yml` is optional and tidy logging config
**Must be completed after #1470.**

### Step 1
Delete (a) the file logging.yml and (b) the logs folder from all the following places:
* kedro project template
* kedro e2e test
* all kedro starters

**No longer doing the above; see https://github.com/kedro-org/kedro/issues/1461#issuecomment-1127770992.**

I think you'll get an exception like this:
```
kedro.config.abstract_config.MissingConfigException: No files found in ['/Users/antony_milne/kedro_stuff/projects/spaceflights/conf/base', '/Users/antony_milne/kedro_stuff/projects/spaceflights/conf/local'] matching the glob pattern(s): ['logging*', 'logging*/**', '**/logging*']
```
Catch this and emit as a debug level message instead. Write new tests/modify existing ones as needed.

**kedro/config/logging.yml should remain! Don't delete that.**

### Step 2 - No longer doing this; see https://github.com/kedro-org/kedro/issues/1461#issuecomment-1127770992.
Add the following to `configure_logging` after the `logging.config` line:
```
logging.getLogger(PACKAGE_NAME).setLevel(logging.INFO)
```

Add the following logger to `kedro/config/logging.yml`:
```
    # A logger for package_name is set at runtime in kedro.framework.project.__init__.
    # This is equivalent to the following entry.
    # package_name:
    #   level: INFO
```
Test to make sure this works as planned. Ask me to explain what this is all about.

### Step 3
Tidy up tests. Look for all places we write a logging.yml in tests and remove them. `local_logging_config` appears to be the way we do this, but check any other references to logs to make sure they should still be there. `local_local_config` in `tests/session/confest.py` is a special case that should remain.

### Step 4
Manual testing: check that everything works the same in terms of console and file logging when you do a `kedro run`. Make sure you test log messages coming from different places, e.g. something emitted from within the project.

AntonyMilneQB:
## Description
Closes #1469. 

In short, you no longer need to supply a project-side logging.yml file.

Sorry it's a bit of a huge PR to review, but the changes here are actually quite small. I've left lots of comments in the files that hopefully make it easier! Almost all the files changed are just updating tests. These are much simpler now because we don't have to provide a mock logging.yml file for them to run 🎉 

Equivalent changes to starters templates are in https://github.com/kedro-org/kedro-starters/pull/90.

## Testing
Tested with and without a logging.yml the following commands:
- `kedro run`
- `kedro run --runner=ParallelRunner`
- `kedro info`
- `kedro ipython`
- `kedro jupyter lab`
- `kedro run` with an exception thrown in the pipeline
- `kedro package` and then run packaged project

All work as expected, but it turns out there's actually always been a problem with the parallel runner not using the framework-side logging config. This problem is now more prominent because it means that if you don't have a project-side logging.yml the logging will disappear completely (because there's effectively no logging config at all in the runner any more). I'm going to make a separate issue to fix this as part of the point we already have in #1461 to try and improve the parallel runner's logging handling.

## TODO
- [ ] Update release notes
- [x] Make parallel runner logging ticket

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1543"></a>


kedro/config/logging.yml
```diff
@@ -1,44 +1,21 @@
 version: 1
```
AntonyMilneQB, line 1: Some of these changes are just formatting. What's actually changed here is:
* removed `info_file_handler` and `error_file_handler`. These are now only defined on the project-side logging config
* removed special case `kedro.framework.cli.hooks.manager` logger (which was the same as the default, just without the file-based handlers enabled)

features/steps/test_starter/{{ cookiecutter.repo_name }}/conf/base/logging.yml
```diff
@@ -1,55 +1,44 @@
 version: 1
```
AntonyMilneQB, line 1: Changes here are the same as to the template file - see explanation there,

kedro/templates/project/{{ cookiecutter.repo_name }}/conf/base/logging.yml
```diff
@@ -1,55 +1,44 @@
 version: 1
```
AntonyMilneQB, line 1: Most of these changes are just formatting. What's actually changed here is:
* removed `json_formatter` which was never used anywhere
* removed the special case `kedro.x` loggers in favour of the generic `kedro` one, which did the same thing anyway
* altered `root` logger so it no longer shows all `INFO` level messages (it's now the same as the framework-side logging config and doesn't specify this explicitly, so defaults to Python's WARNING level)
* added a special logger for `{{ cookiecutter.python_package }}`. This is to ensure that calls [like this](https://github.com/kedro-org/kedro-starters/blob/main/spaceflights/%7B%7B%20cookiecutter.repo_name%20%7D%7D/src/%7B%7B%20cookiecutter.python_package%20%7D%7D/pipelines/data_science/nodes.py#L55) still emit logs

tests/framework/cli/test_cli.py
```diff
@@ -71,6 +77,16 @@ def fake_session(mocker):
...
 
 
+class TestDefaultLogging:
+    def test_setup_root_logger(self):
+        root_logger = logging.getLogger()
+        assert "console" in {handler.name for handler in root_logger.handlers}
+
+    def test_setup_kedro_logger(self):
+        kedro_logger = logging.getLogger("kedro")
+        assert kedro_logger.level == logging.INFO
```
AntonyMilneQB, line 87: These are updated versions of tests that used to be in `test_cli_logging_setup.py` (tests were in the wrong place before, now that file is deleted).

tests/framework/session/test_session.py
```diff
@@ -380,6 +346,24 @@ def test_broken_config_loader(self, mock_settings_file_bad_config_loader_class):
...
+            "No project logging configuration loaded; "
+            "Kedro's default logging configuration will be used."
+        ]
+        actual_log_messages = [
+            rec.getMessage()
+            for rec in caplog.records
+            if rec.name == SESSION_LOGGER_NAME and rec.levelno == logging.DEBUG
+        ]
+        assert actual_log_messages == expected_log_messages
+
```
AntonyMilneQB, line 366: This is the only really interesting new test.

kedro/framework/session/session.py
```diff
@@ -185,8 +185,15 @@ def _get_logging_config(self) -> Dict[str, Any]:
 
     def _setup_logging(self) -> None:
         """Register logging specified in logging directory."""
-        conf_logging = self._get_logging_config()
-        configure_logging(conf_logging)
+        try:
```
AntonyMilneQB, line 188: This is the key change: we now allow there to be no logging.yml on the project-side.

AhdraMeraliQB (approved):
Nicely done! Great work on the description too, super helpful 🌟

tests/framework/session/test_session.py
MerelTheisenQB: This test is specifically checking that we get that debug message, but the other tests aren't actually using logging config either right? So in theory they would all show that message in debug mode?

MerelTheisenQB (approved):
Great work @AntonyMilneQB ! 🤩  This has cleaned up the logging so much!

One minor question around testing, do we have any tests now that check stuff works correctly when the user has a custom project level logging.yml? I couldn't completely figure it out from the changes made to the tests.

kedro/templates/project/{{ cookiecutter.repo_name }}/conf/base/logging.yml
```diff
@@ -1,55 +1,44 @@
...
+    encoding: utf8
+    delay: True
 
-    kedro.pipeline:
-        level: INFO
-        handlers: [console, info_file_handler, error_file_handler]
-        propagate: no
+loggers:
+  kedro:
+    level: INFO
```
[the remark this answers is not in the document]
AntonyMilneQB: Yes, exactly! There's no handlers explicitly specified here, so it will propagate to `root` and use those.

tests/framework/session/test_session.py
AntonyMilneQB: Yes indeed! Though some of the other tests have `mocker.patch("kedro.framework.session.KedroSession._setup_logging")` so the debug message won't be emitted there anyway.

AntonyMilneQB:
> Great work @AntonyMilneQB ! 🤩 This has cleaned up the logging so much!
> 
> One minor question around testing, do we have any tests now that check stuff works correctly when the user has a custom project level logging.yml? I couldn't completely figure it out from the changes made to the tests.

Thank you! This is covered by `test_setup_logging_using_absolute_path` and also the e2e tests but it's not very explicit at the moment. I will change this so it's better.

PR #1546, opened 2022-05-17
Revisit logging levels

MerelTheisenQB:
## Description
Part of https://github.com/kedro-org/kedro/issues/1461

## Development notes
Changed several logs from `INFO` to `DEBUG` level, because they're not that important and sometimes cause confusion for users. 

Initially I wanted to change the logging level to `DEBUG` for the following lines as well:

1. https://github.com/kedro-org/kedro/blob/main/kedro/framework/hooks/manager.py#L82-L85
```
  logger.info(
            "Hooks are disabled for plugin(s): %s",
            ", ".join(sorted(disabled_plugin_names)),
        )
```

2.https://github.com/kedro-org/kedro/blob/main/kedro/framework/hooks/manager.py#L88-L92
```
logger.info(
            "Registered hooks from %d installed plugin(s): %s",
            len(plugin_names),
            ", ".join(sorted(plugin_names)),
        )

```

3. https://github.com/kedro-org/kedro/blob/main/kedro/framework/cli/hooks/manager.py#L45-L49
```
 logging.getLogger(__name__).info(
                "Registered CLI hooks from %d installed plugin(s): %s",
                len(plugin_names),
                ", ".join(sorted(plugin_names)),
            )
```

However, these lines all get logged before the logging configuration is read in the session (https://github.com/kedro-org/kedro/blob/main/kedro/framework/session/session.py#L172). This makes it impossible to then show these debug logs, because the user only has control over the logging configuration for log lines that come after that config is read.

If we feel that the above 3 logs aren't useful, we can still make them debug level, but that would in reality mean that this isn't ever logged anymore. 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1546"></a>


AhdraMeraliQB (approved):

noklam (approved):
LGTM. This is not really related to the PR and is minor, but wonder why we do the `**` before Kedro project.

 https://github.com/kedro-org/kedro/blob/c6882f64e9f76a352826fd29b4aeb0e8663d4c62/kedro/extras/extensions/ipython.py#L79-L82

https://github.com/kedro-org/kedro/blob/c6882f64e9f76a352826fd29b4aeb0e8663d4c62/kedro/framework/session/session.py#L327

MerelTheisenQB:
> LGTM. This is not really related to the PR and is minor, but wonder why we do the `**` before Kedro project.

Good question! My guess is just to make it stand out a bit more from the other log commands.

AntonyMilneQB (commented):
Just a few comments:
1. I think we should get rid of the `**` and highlight the message a nicer way using rich instead. We don't use `**` anywhere else in kedro and it's always looked weird to me. I'll add it to one of the rich tickets if this sounds good?
2. Ivan and I also realised what you say about the 3 hook log messages when we talked about this a few weeks ago. Our conclusion was that these _should_ become `DEBUG` level, even though as you say it means that realistically there's not currently a way of making them appear at all... The point 3 from bottom on https://github.com/kedro-org/kedro/issues/1461 is one way we might be able to handle this in the future. For the time being I think it's fine to mark them as `DEBUG` and they will become invisible. If anyone complains about this (_really_ unlikely I think) then we can edit the framework-side logging.yml to change the specific `kedro.framework.hooks.manager` logger to show `DEBUG` level.
3. Would be nice to see what the output of `kedro run` and `kedro catalog list` look like after this just so we check it achieves Ivan's vision of a clean output with no preamble.

Also, great job fixing up the tests. I know how fiddly that can be with caplog and everything 😬

MerelTheisenQB:
> Just a few comments:
> 
> 1. I think we should get rid of the `**` and highlight the message a nicer way using rich instead. We don't use `**` anywhere else in kedro and it's always looked weird to me. I'll add it to one of the rich tickets if this sounds good?

Agreed, and done!

> 2. Ivan and I also realised what you say about the 3 hook log messages when we talked about this a few weeks ago. Our conclusion was that these _should_ become `DEBUG` level, even though as you say it means that realistically there's not currently a way of making them appear at all... 

I've changed this. The annoying thing is we actually checked explicitly for these log statements in a couple of e2e tests. For now I've just commented out those lines and added a comment explaining what's going on. If preferred I could just remove it all, but if we in future find a way to make them appear again we probably want to get these tests back as they were. 

> 3. Would be nice to see what the output of `kedro run` and `kedro catalog list` look like after this just so we check it achieves Ivan's vision of a clean output with no preamble.

Output of `kedro run`:

kedro run

Output of `kedro catalog list`:
kedro catalog list

AntonyMilneQB (approved):
This is perfect, thanks for updating it! The logs look much cleaner now.

That kedro telemetry message kind of stands out and, looking at the repo, I'm actually not sure what we want to do about it. I'll make a separate issue for it.

PR #1555, opened 2022-05-18
Fix Windows  Python 3.10 CI issues

SajidAlamQB:

## Description

`win_pip_compile-3.10` and `win_unit_tests-3.10` are failing on the "`Install all requirements`" stage with no useful error message.

This PR will look to investigate and resolve said issue.

## Development notes

In circleci `continue_config.yml` I've slightly narrowed down the issue to the `restore_cache` step in `win_setup_requirements`.

Specifically the offending lines were: 

```
restore_cache:
          name: Restore conda environment cache
          key: kedro-deps-v1-win-<<parameters.python_version>>-{{ checksum "requirements.txt" }}-{{ checksum "test_requirements.txt" }}
```
Removing these lines have resolved the error, but I still need to investigate if this is a viable solution.

**Edit 1 after investigation:**

After further digging, I found that in `C:\tools\miniconda3\envs\kedro_builder` the `zlib.dll` file goes missing after the `install GDAL, Fiona and pytables` step in our CI, this `.dll` file still exists after both `restore_cache` attempts so problem could be dependency clashes with the restored conda environment and the `install GDAL, Fiona and pytables` step. When this file goes missing the environment can no longer run any commands including `pip install` hence the empty error message.

**Edit 2:**

So I am certain now that the `Restore conda environment cache` and `install GDAL, Fiona and pytables` steps in our CI are having dependency clashes for python 3.10. I presume the cache we are getting isn't compatible with the installation of the latest GDAL, Fiona or pytables. A potential fix for this could be incrementing our cache key to v2 essentially resetting our cache.

**Edit 3:**

I tried incrementing the cache key to v2 although the build succeeds on the first run subsequent runs have the same failure as before. Another solution is to run `install GDAL, Fiona and pytables` before the cache step but I want to know your thoughts on this?

**Edit 4:**

After talking with @AntonyMilneQB running `install GDAL, Fiona and pytables` before the cache step would not be ideal as we would be running outdated packages as a result. A final stop-gap solution is to skip the caching of the conda environment for python 3.10.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1555"></a>

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -155,14 +155,12 @@ commands:
         # Dependency resolution works best when installing these altogether in one
         # `conda install` command rather than one at a time in several sequential `conda install`s.
       - run:
-          name: Install GDAL
-          command: conda activate kedro_builder; conda install gdal -c conda-forge -y
+          name: Install GDAL, Fiona and pytables
+          command: conda activate kedro_builder; conda install gdal fiona pytables -c conda-forge -y
+        # The above lines causes zlib.dll to go missing in kedro_builder therefore we need to re-install zlib
       - run:
-          name: Install Fiona
-          command: conda activate kedro_builder; conda install fiona -c conda-forge -y
-      - run:
-          name: Install pytables
-          command: conda activate kedro_builder; conda install pytables -c conda-forge -y
+          name: Install zlib
+          command: conda activate kedro_builder; conda install zlib=1.2.11=*_5
       - run:
           name: Install all requirements
           command: conda activate kedro_builder; pip install -r test_requirements.txt -U
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -144,12 +144,6 @@ commands:
       python_version:
         type: string
     steps:
-      - restore_cache:
-          name: Restore package cache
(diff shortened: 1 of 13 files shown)


MerelTheisenQB:
Great investigative work @SajidAlamQB 🕵️‍♂️ 💥  ! If this makes the builds pass consistently, I'm happy to swap the cache restoring and installation of the packages. Looking at the build times from this PR and the other branches it doesn't add any significant time to install these packages first. It also seems that for other python version the clashes exist, and are gone after your fix, so hopefully this will create a more stable CI for all builds 🤞

AntonyMilneQB (approved):
Amazing work getting to the bottom of this! 🎉  Not an easy one to figure out at all.

This looks like a better solution to me than swapping the order since that would have overwritten the latest (and so "more correct") installation of GDAL/fiona/pytables with the cached version every time, which is probably fine but a bit naughty. Especially when it's working ok in all other Python versions, it seems unnecessary to swap it just to get it working on 3.10.

Ignoring the cache here only seems to add a minute or so to CI. And hopefully in the future some time the dependencies fix themselves and we can remove the `unless`.

MerelTheisenQB (approved):
Agreed that this is an even better solution 👍 Thanks for getting to the bottom of this!

PR #1574, opened 2022-05-25
[Bug Fix] - Kedro info failure, entry_point should use `module` instead of `module_name`

noklam:
…ad of `module_name`

## Description

This is a bug fix which `entry_point.module` should be used instead of `entry_point.module_name`.

This PR should also fix the unittest so that such error should be caught instead of pass.
Currently the Mock is accepting any attribute which is not expected.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1574"></a>


tests/framework/cli/conftest.py
```diff
@@ -33,13 +33,12 @@
...
 def entry_points(mocker):
-    return mocker.patch("importlib_metadata.entry_points")
+    return mocker.patch("importlib_metadata.entry_points", spec=True)
 
 
 @fixture
 def entry_point(mocker, entry_points):
-    ep = mocker.MagicMock()
-    entry_points.return_value = ep
+    ep = mocker.patch("importlib_metadata.EntryPoint", spec=True)
```
noklam, line 41: `spec=True` is needed to trigger `AttributeError` when accessing non-exist attribute. I wonder if we may need to update other tests eventually. 

But I haven't fully understood these parameters, so will do it later instead. I have raise an issue on pytest repo for this.

tests/framework/cli/test_cli.py
```diff
@@ -104,9 +104,13 @@ def test_print_version(self):
         assert result_abr.exit_code == 0
         assert version in result_abr.output
 
+    def test_info(self):
+        result = CliRunner().invoke(cli, ["info"])
+        assert result.exit_code == 0
+
```
noklam, line 110: This will fail properly even if the mocking is wrong.

AntonyMilneQB: I don't think this is doing anything. We already have `test_info_no_plugins` (which actually shouldn't need the `entry_points` fixture it's using), and that does the same as your test. Unfortunately it can't pick up on the bug because the error is only triggered if there are plugins installed, and when this test runs there aren't any so it passes already.

noklam: @AntonyMilneQB Thanks, I missed the test. Actually, the test would fail properly without any change now.

The reason is exactly the problem we discussed. The `MagicMock` should throw `AttributeError` when accessing attributes that do not exist.  `entry_point.module_name` would result in `AttributeError` now due to the change in `conftest.py`.

tests/framework/cli/conftest.py
AntonyMilneQB: This doesn't look right to me 🤔  I don't think we should need `spec=True` here. I think we're probably just mocking the wrong things or don't have the right tests to pick up on the bug.

tests/framework/cli/test_cli.py
AntonyMilneQB: I think best thing is to make a new `info.feature` e2e test. If you search for `I have installed the test plugin` you'll see we already have a step which will help create a project that does has a plugin installed. If you run `kedro info` on such a project that e2e test would pick up on the bug properly.

AntonyMilneQB: @noklam hmmm, how can `test_info_no_plugins` pick up the problem when there's no plugins in that test so it doesn't go into the for loop that triggers the exception?

noklam: See the test result below. It can because it doesn't really load any plugin. We mock the `entry_point` so it always return a list of `[ep]` that goes into the loop. If you look at the test `test_info_with_plugin`, it does the same thing, it just also mock the `entry_point.module` to fake the attribute.

If you are saying the loop was never entered then both test would not be valid.


_______________________________________________________________________________________________________________________________________________ TestCliCommands.test_info_no_plugins ________________________________________________________________________________________________________________________________________________

self = <tests.framework.cli.test_cli.TestCliCommands object at 0x7fdab0b8ecd0>

    @mark.usefixtures("entry_point")
    def test_info_no_plugins(self):
        result = CliRunner().invoke(cli, ["info"])
>       assert result.exit_code == 0
E       assert 1 == 0
E        +  where 1 = <Result AttributeError("Mock object has no attribute 'module_name'")>.exit_code

noklam: We can have a test that mocks for no plugins and one without mocking, but that feels a bit duplicate to me.

Could we just remove the mocking for `test_info_no_plugins` so it basically acts the same as e2etest?

I think it's still valid to add the e2etest to make sure the environment is correct, but a unit test will still be valuable because it's fast to run and easy to debug.

tests/framework/cli/conftest.py
SajidAlamQB: Did you manage to find out if `spec=True` ok here?

SajidAlamQB: Did you manage to find out if `spec=True` ok here?

noklam: @SajidAlamQB  I think using `spec` which mock the real object is better than using `MagicMock`.

image

AhdraMeraliQB: Looks like we'd need to set `spec=True` for both Mock and MagicMock: https://stackoverflow.com/questions/52978673/how-does-magicmock-avoid-throwing-attributeerror-when-a-random-method-is-called

Curious as to how this is handled in other instances without using spec 🤔

AhdraMeraliQB (approved):
Nice fix!

tests/framework/cli/conftest.py
noklam: Here is a mistake that I made initially. The mock only cares about the method/class that it mocks, not the return value.

For example, if we have a function `create_kedrocontext(args) -> KedroContext`. If we mock `create_kedrocontext` with `spec=True`, it will return a generic `MagicMock`.
 If we want the return value to respect `KedroContext`'s spec, we need to specify the return value as `MagicMock(KedroContext`.

SajidAlamQB (approved):
LGTM!🌟

PR #1592, opened 2022-06-06
[KED-3023] Enable plugins to extend starter-aliases default list

### LINKED ISSUE #1040: [KED-3023] Enable plugins to extend starter-aliases default list

noklam:

## Description

Taking over the original PR from #1265, would fix #1040

## Development notes

* modified the _STARTERS_ALIASES to make it a dict with the format discussed in above issue.
* created a kedro.starter entrypoint which can is iterated over to update the _STARTERS_ALIASES constant.
* updated the kedro new and kedro starter list commands to use the updated _STARTERS_ALIASES constant (with new aliases discovered in the entrypoints).
* Refactor with `KedroStarterSpec`, the location is temp, it probably doesn't belong `kedro.framework.cli.starters`

The code can be run with
```
conda create -n fps python=3.8 -y
conda activate fps
pip install -e .  # Assume have the latest commit 
pip install git+https://github.com/noklam/fake-plugin-starter.git
kedro starter list
```

## Todo
- [ ] Unit test + e2etest?
- [ ] Documentions

## Screenshots
image

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1592"></a>


noklam:
It's in a good shape now and the PR should be functioning as it intended. Big thanks to @Galileo-Galilei to draft this PR initially and make it very easy for me to pick up.

I plan to do a bit more refactoring, so this https://github.com/kedro-org/kedro/pull/1592/commits/56ee39cb41ca76725145e3db0813e378ea61410d will be my checkpoint in case I fail to :).

kedro/framework/cli/starters.py
```diff
@@ -38,6 +43,81 @@
...
+    """
+
+    name: str
+    template_path: str
+    directory: Optional[str] = None
+    origin: Optional[str] = field(init=False, default=None)
+
+    def __attrs_post_init__(self):
+        if not self.directory:
+            self.directory = self.name  # By default they are the same unless specified.
```
AntonyMilneQB, line 74: I don't think this is the right behaviour (will explain more in review comment). I think it should just be left as `None`.

kedro/framework/cli/starters.py
```diff
@@ -28,7 +33,7 @@
 KEDRO_PATH = Path(kedro.__file__).parent
 TEMPLATE_PATH = KEDRO_PATH / "templates" / "project"
 
-_STARTER_ALIASES = {
+_OFFICIAL_STARTER_ALIASES = {
```
MerelTheisenQB, line 36: Is this used anywhere apart from in the tests?

kedro/framework/cli/starters.py
```diff
@@ -38,6 +43,81 @@
...
+    name: str
+    template_path: str
+    directory: Optional[str] = None
+    origin: Optional[str] = field(init=False, default=None)
+
+    def __attrs_post_init__(self):
+        if not self.directory:
+            self.directory = self.name  # By default they are the same unless specified.
+
+    def format(self) -> Dict:
```
AntonyMilneQB, line 76: Would doing this as `__str__` be a bit cleaner? So you can do `str(kedro_starter_spec)` directly?

AntonyMilneQB: Oh, I see this actually returns a `Dict` rather than a `str`. In that case I think `to_dict` would be more conventional method name here.

kedro/framework/cli/starters.py
```diff
@@ -38,6 +43,81 @@
...
+        Examples:
+        ::
+            >>> spec = KedroStarterSpec("astro-airflow-iris",
+                                    "git+https://github.com/kedro-org/kedro-starters.git")
+            >>> spec.format()
+                {
+                    "astro-airflow-iris": {
+                        "name": "astro-airflow-iris",
+                        "template_path": "git+https://github.com/kedro-org/kedro-starters.git",
+                                          }
```
AntonyMilneQB, line 91: Related to the `directory` vs. `name` issue, I don't think this is quite the right format.

kedro/framework/cli/starters.py
```diff
@@ -38,6 +43,81 @@
...
+                        name="astro-airflow-iris",
+                        template_path="git+https://github.com/kedro-org/kedro-starters.git",
+                        directory=None
+                                   )
+    """
+
+    name: str
+    template_path: str
+    directory: Optional[str] = None
+    origin: Optional[str] = field(init=False, default=None)
```
AntonyMilneQB, line 70:
```suggestion
    origin: Optional[str] = field(init=False, default="kedro")
```
Just to remove the need for L118?

features/steps/cli_steps.py
```diff
@@ -242,6 +242,25 @@ def create_project_with_starter(context):
     assert res.returncode == OK_EXIT_CODE, res
 
 
+@given("I have run a non-interactive kedro new with custom plugin starter")
+@when("I run a non-interactive kedro new with custom plugin starter")
+def create_project_with_plugin_starter(context):
+    """Behave step to run kedro new given the config I previously created."""
```
MerelTheisenQB, line 248: Maybe we can merge this with `create_project_with_starter` above and pass an optional argument with a starter name. If it's not provided then it would just use `starter_dir = Path(__file__).parent / "test_starter"`

kedro/framework/cli/starters.py
```diff
@@ -38,6 +43,81 @@
 }
 _STARTERS_REPO = "git+https://github.com/kedro-org/kedro-starters.git"
 
+
+@define(order=True)
```
MerelTheisenQB, line 47: What does this do?

kedro/framework/cli/starters.py
```diff
@@ -62,6 +142,69 @@ def _remove_readonly(func: Callable, path: Path, excinfo: Tuple):  # pragma: no
     func(path)
 
 
+def _is_valid_starter_entrypoint_config(
+    module_name: str, config: KedroStarterSpec
```
AntonyMilneQB, line 146:
```suggestion
    module_name: str, spec: KedroStarterSpec
```
I like spec 🙂  `config` is already overused in kedro so let's try to avoid it here and call it a specification throughout.

kedro/framework/cli/starters.py
```diff
@@ -62,6 +142,69 @@ def _remove_readonly(func: Callable, path: Path, excinfo: Tuple):  # pragma: no
...
+        module_name = starter_entry_point.module.split(".")[0]
+        for starter_config in starter_entry_point.load():
+            is_valid_config_status = _is_valid_starter_entrypoint_config(
+                module_name, starter_config
+            )
+            if not is_valid_config_status:
+                click.secho(
+                    f"Starter alias `{starter_config.name}` from `{module_name}` "
+                    f"has been ignored as the config is invalid and cannot be loaded",
+                    fg="yellow",
```
MerelTheisenQB, line 192: How will these `click.secho` commands work together with rich?

MerelTheisenQB (commented):
This is great work @noklam ! 👍 I've left some initial comments and questions.

kedro/framework/cli/starters.py
AntonyMilneQB, line 36: Is this still used anywhere? I feel like it just duplicates `_OFFICIAL_STARTERS_SPECS` so could be removed?

kedro/framework/cli/starters.py
```diff
@@ -62,6 +142,69 @@ def _remove_readonly(func: Callable, path: Path, excinfo: Tuple):  # pragma: no
...
+        ),
+        KedroStarterSpec(
+            name="astro-iris",
+            template_path="git+https://github.com/kedro-org/kedro-starters.git",
+            directory="astro-airflow-iris",
+            origin="kedro",
+        ),
+    ]
+    """
+    # add an extra key to indicate from where the plugin come from
```
AntonyMilneQB, line 176:
```suggestion
```

kedro/framework/cli/starters.py
```diff
@@ -62,6 +142,69 @@ def _remove_readonly(func: Callable, path: Path, excinfo: Tuple):  # pragma: no
...
+    starters_aliases = _OFFICIAL_STARTERS_SPECS
+
+    existing_names: Dict[str, str] = {}  # dict {name: module_name}
+    for starter_entry_point in importlib_metadata.entry_points().select(
```
AntonyMilneQB, line 180: Is there any chance of just calling the new entrypoint function that you modified recently or is this sufficiently different that we need to copy and paste the right bits here? (Not a big problem if we do, just wondering if your new scheme might make it easier to be more DRY now)

kedro/framework/cli/starters.py
```diff
@@ -62,6 +142,69 @@ def _remove_readonly(func: Callable, path: Path, excinfo: Tuple):  # pragma: no
...
+        for starter_config in starter_entry_point.load():
+            is_valid_config_status = _is_valid_starter_entrypoint_config(
+                module_name, starter_config
+            )
+            if not is_valid_config_status:
+                click.secho(
+                    f"Starter alias `{starter_config.name}` from `{module_name}` "
+                    f"has been ignored as the config is invalid and cannot be loaded",
+                    fg="yellow",
+                )
```
AntonyMilneQB, line 193: Doesn't this just repeat the warning from inside `_is_valid_starter_entrypoint_config`? Could you just do the check inline here?
```
if not isinstance(config, KedroStarterSpec):
    click.secho("...")
elif ...
```
and then remove `_is_valid_starter_entrypoint_config` altogether?

kedro/framework/cli/starters.py
```diff
@@ -288,11 +444,12 @@ def _get_cookiecutter_dir(
                 f" Specified tag {checkout}. The following tags are available: "
                 + ", ".join(_get_available_tags(template_path))
             )
-        official_starters = sorted(_STARTER_ALIASES)
-
+        official_starters = sorted(_OFFICIAL_STARTERS_SPECS)
```
AntonyMilneQB, line 447: How does this implement the `sort`? I see you have `order=True` but what does that actually generate for the comparison methods?

kedro/framework/cli/starters.py
```diff
@@ -149,11 +295,21 @@ def starter():
...
-    repo_url = _STARTERS_REPO.replace("git+", "").replace(".git", "/tree/main/{alias}")
-    output = [
-        {alias: repo_url.format(alias=alias)} for alias in sorted(_STARTER_ALIASES)
-    ]
-    click.echo(yaml.safe_dump(output))
+    starters_aliases = _get_starters_aliases()
+
+    # ensure kedro starters are listed first
+    official_starters = sorted(
+        list(filter(lambda x: x.origin == "kedro", starters_aliases))
```
AntonyMilneQB, line 302:
```suggestion
        starter_alias for starter_alias in starters_aliases if starter_alias.origin != "kedro"
```
Generally we avoid too much functional stuff in the kedro codebase in favour of comprehensions 🙂

AntonyMilneQB (commented):
A few small comments here, will post higher level thoughts on the overall data structure in a separate post before github loses all my comments!
1. Let's go for `spec` rather than `config` throughout since `config` is already used for something else in starters
2. What happens if a plugin provides something that is `KedroStarterSpec` but has some validation problem (like missing field)? Will `attr` raise some suitable error message?

kedro/framework/cli/starters.py
AntonyMilneQB: In fact, doesn't `attr` have something built in that would do this? `asdict`?

AntonyMilneQB:
This looks great, I love the new `attr` implementation instead of the nested dictionary structures! Really nice that we get sorting and validation for free with that 👍 

I have two high level suggestions on how to make this even better: 
1. I think the directory vs. name behaviour isn't quite right.
2. I think it could be made even simpler (both in terms of our codebase and how a plugin author writes a spec).

### `directory` and `name`
The fact that the `directory` takes the starter `name` is a bit of a special case for the kedro official starters and should not be the default behaviour for user-defined starters. A plugin author should be able to specify a starter just by giving a `name` and `template_path`, and in that case `directory` should remain `None`. The kedro starters should be defined so that they explicitly do `KedroStarterSpec("spaceflights", "git+https://github.com/kedro-org/kedro-starters.git", "spaceflights").

### Simplifications
I wonder if it would be overall better to turn `_OFFICIAL_STARTERS_SPECS` into a dictionary of the format `{starter_alias: KedroStarterSpec}` rather than a list. `KedroStarterSpec` could then drop the `name` field (especially if we go for a dictionary format like suggested below). My feeling is that this might reduce the number of times you need to iterate through `_OFFICIAL_STARTERS_SPECS`, e.g. to check whether a starter alias exists you just check `alias in _OFFICIAL_STARTERS_SPECS` rather than `alias in {starter_spec.name for starter_spec in _OFFICIAL_STARTERS_SPECS}`. Give it a go and see what you think, but my gut instinct is this might be simpler overall 🙂 

Given that `KedroStarterSpec` is very similar to a dictionary, I feel like it should be simpler to directly convert to and from dictionary also (maybe using something built into attrs). The "to" direction would be useful for `yaml.dump`ing the output.  The "from" direction I think might be a better way for a plugin author to specify their custom starter. e.g. this worked according to the old scheme @Galileo-Galilei wrote:
```
starters = [
    {"name": "fake_name", "template_path": "http://google.com", "directory": "blahblah"},
    {"name": "blah", "template_path": "http://google.com"}
]
```
A plugin author now has to import `KedroStarterSpec` and understand the schema of that, which is a bit trickier than just supplying a dictionary (probably only advanced users will do this and I'm sure could figure out how `KedroStarterSpec` works, but let's make their lives as easy as possible anyway!).

Then we do the conversion to `KedroStarterSpec` ourselves in starters.py via `KedroStarterSpec(**starters_dict)`. Or maybe `{starters_dict["alias"]: `KedroStarterSpec(**starters_dict.pop("alias"))}` if we change the format of `_OFFICIAL_STARTERS_SPECS`.

kedro/framework/cli/starters.py
noklam: @MerelTheisenQB This is the goodies that `dataclasses` or `attrs` offer. 

Imagine if you have a class like this
```python
class MyClass:
   a: int
   b: int

a = MyClass(1,1)
b = MyClass(1,1)
```
If you do ` a == b`, you get a `False` because it is looking at its hash by default, but what you really want is something like a `namedtuple`, which should look at the values of the field. (We did this for `Node`). But with `dataclasses` and `attrs` class, it will return `True` as it generates the correct `__magic__` method under the hood.

In this case, `order=True` will allows you to do comparison like  `MyClass(1,1) < MyClass(2,1)` by look at the fields value, by default the orders go with `a` then `b`, which makes the class `sortable`. I think under the hood it generates a `__lt__` (lesser than) method.

kedro/framework/cli/starters.py
noklam: I thought about this too! And I roughly remember it wasn't straightforward, will revisit it.

kedro/framework/cli/starters.py
noklam: I found `filter` is quite expressive in this case with a straightforward lambda, but no strong preference here, I am okay with switching it to list comprehension.

kedro/framework/cli/starters.py
noklam: See another comments I have for Merel, but it's quite intuitive. You can treat it like a namedtuple basically.

kedro/framework/cli/starters.py
noklam: I think so, this is pretty much a leftover method that doesn't do much now since all the validation logic is removed.

noklam:
I tend to agree it's quite similar to a dict-like structure. I tried a plain dictionary and also `KedroStarterSpecDict` `attrs` class inheriting from `collections.UserDict`.

It's useful to store it as a dict, as the key is unique so we can just remove duplicate alias without using a set that called `existing_name` to keep track of it.

However, I found that it is not easy to just use `asdict` to convert it to a nice format. Using `KedrostarterSpecdict` allow us to convert the class to dictionary recursive with a built-in function, but it will print all attributes instead of the needed one. I think this is quite custom logic, and some kind of `format` method cannot be avoided.

In the end, I give up on the `KedroStarterSpecDict` since the main advantage is enabling the `asdict` recursive call, but we are not using it, so a plain dict would be sufficient.

kedro/framework/cli/starters.py
```diff
@@ -149,11 +233,31 @@ def starter():
...
+        alias for alias, spec in starters_dict.items() if spec.origin != "kedro"
+    )
+    unofficial_starters_dict = {
+        alias: starters_dict[alias] for alias in unofficial_starters_aliases
+    }
+
+    for starters_dict in [official_starters_dict, unofficial_starters_dict]:
+        if not starters_dict:
+            continue
+        _, first_spec = next(iter(starters_dict.items()))
```
AntonyMilneQB, line 256:
```suggestion
        first_spec = next(iter(starters_dict.values()))
```

kedro/framework/cli/starters.py
```diff
@@ -288,11 +392,12 @@ def _get_cookiecutter_dir(
...
             )
-        official_starters = sorted(_STARTER_ALIASES)
-
+        official_starters = sorted(_OFFICIAL_STARTER_SPECS)
         raise KedroCliError(
             f"{error_message}. The aliases for the official Kedro starters are: \n"
-            f"{yaml.safe_dump(official_starters)}"
+            f"""{yaml.safe_dump(
+                [starter.format() for starter in official_starters],
+                sort_keys=False)}"""
```
AntonyMilneQB, line 400: This should just be aliases, so `official_starters.keys()`.

AntonyMilneQB (commented):
I think we are 95% there! Thank you for your perseverance with changing this. It's looking really nice and clean now I think 👍 ⭐ 

A few comments/questions, and then hopefully the next round of reviewing will be it! 😀 
1. Please can we use just `name` or just `alias` throughout?
2. What happens if someone defines a plugin that is `KedroStarterSpec` but doesn't provide a `template_path` for example? I guess it raises some sort of validation error, but at what point in the flow does that happen? Does one bad starter prevent the whole `kedro new` or `kedro starters list` commands from running? 
3. I like the new `_OFFICIAL_STARTER_SPECS` dictionary structure but I'm not sure why we want starter name as both a key and an attribute of the value? They key is useful to have but then the value just seems to duplicate it. Assuming plugin authors are defining a `KedroStarterSpec` with a `name` property then I'd just extract that out into the key and not keep it there.
4. Related: I think it's a bit confusing that in the kedro source code we define starters using a dictionary while a plugin author defines using a list. I think either way is ok, but best to do the same way in both I think. So two possibilities: (a) change the plugin structure to be a dictionary also or (b) define `_OFFICIAL_STARTER_SPECS` as a list and then generate the dictionary via `_OFFICIAL_STARTER_SPECS = {spec.name: spec for spec in _OFFICIAL_STARTER_SPECS}`

kedro/framework/cli/starters.py
```diff
@@ -27,16 +30,45 @@
...
+
+@define(order=True)
+class KedroStarterSpec:  # pylint: disable=too-few-public-methods
+    """Specification of custom kedro starter template
+    Args:
+        name: name of the starter (name of the directory)
+        template_path: the path to git repository
+        directory: by default it will look for `name`, optionally override with custom directory.
+        origin: preserved field used by kedro internally to determine where the starter
+        comes from, users do not need to provide this field.
```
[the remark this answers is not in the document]
noklam: 1. Yes, I think I prefer `alias` 
2. This need to be fixed, you will get an error like `    KedroStarterSpec(
TypeError: __init__() missing 1 required positional argument: 'template_path'`. I think the correct thing to do is to create a custom entrypoint load function for everything in `kedro` and used it for all plugins entrypoint loading. It should try to load the entrypoint - if fail then just throw a warning and continue with the program.
3. To be discussed.
4. The 2nd option sounds better, the 1st option has more room for the user to make mistakes (e.g. the key is different from the alias name)

kedro/framework/cli/starters.py
```diff
@@ -149,11 +233,31 @@ def starter():
...
+    unofficial_starters_aliases = sorted(
+        alias for alias, spec in starters_dict.items() if spec.origin != "kedro"
+    )
+    unofficial_starters_dict = {
+        alias: starters_dict[alias] for alias in unofficial_starters_aliases
+    }
+
+    for starters_dict in [official_starters_dict, unofficial_starters_dict]:
+        if not starters_dict:
+            continue
```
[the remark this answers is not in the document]
noklam: This is great, merging the dicts first is a cleaner solution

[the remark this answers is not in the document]
noklam: Actually this isn't right. It should print the official starter first with the extra colored text stating the origin. The `unofficial_starters_dict` isn't generalized enough if there are more than 1 origin.

kedro/framework/cli/starters.py
noklam: In fact it is now a dictionary, so after sorting it is just a list of key!

MerelTheisenQB (commented):
Agree that this is almost ready! Just a couple more comments and questions from me.

kedro/framework/cli/starters.py
```diff
@@ -22,22 +24,53 @@
...
+    """
+
+    alias: str
+    template_path: str
+    directory: Optional[str] = None
+    origin: Optional[str] = field(init=False)
+
+
+_OFFICIAL_STARTER_SPECS = [
+    KedroStarterSpec("astro-airflow-iris", _STARTERS_REPO, "astro-airflow-iris"),
```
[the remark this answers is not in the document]
noklam: @MerelTheisenQB Antony convinced me that the `kedro-starters` is a special case. For other users, it will be likely just 1 template, so they don't have to have a single repository with multiple templates. By default, the top level directory will be the template, `directory` is an optional argument if they need to organise multiple templates in folders.

kedro/framework/cli/starters.py
```diff
@@ -62,6 +95,61 @@ def _remove_readonly(func: Callable, path: Path, excinfo: Tuple):  # pragma: no
...
 
+def _get_starters_dict() -> Dict[str, KedroStarterSpec]:
+    """This functions lists all the starter aliases declared in
+    the core repo and in plugins entry points.
+    The output looks like:
+    {"astro-airflow-iris":
+        KedroStarterSpec(
+            name="astro-airflow-iris",
+            template_path="git+https://github.com/kedro-org/kedro-starters.git",
+            directory=None
```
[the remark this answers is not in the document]
noklam: Updated

kedro/framework/cli/starters.py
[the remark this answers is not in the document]
MerelTheisenQB: That makes sense! Consider me convinced as well 🙂

kedro/framework/cli/utils.py
```diff
@@ -318,6 +319,27 @@ def _check_module_importable(module_name: str) -> None:
         ) from exc
 
 
+def get_entry_points(name: str) -> importlib_metadata.EntryPoints:
```
AntonyMilneQB, line 322:
```suggestion
def _get_entry_points(name: str) -> importlib_metadata.EntryPoints:
```
The fewer functions we advertise as public the better 😀

kedro/framework/cli/starters.py
```diff
@@ -22,22 +24,53 @@
...
-    "spaceflights",
-}
 _STARTERS_REPO = "git+https://github.com/kedro-org/kedro-starters.git"
 
+
+@define(order=True)
+class KedroStarterSpec:  # pylint: disable=too-few-public-methods
+    """Specification of custom kedro starter template
+    Args:
+        alias: alias of the starter which shows up on `kedro starter list` and `kedro new`
```
AntonyMilneQB, line 41:
```suggestion
        alias: alias of the starter which shows up on `kedro starter list` and is used by the starter argument of `kedro new`
```

kedro/framework/cli/starters.py
```diff
@@ -22,22 +24,53 @@
...
+        template_path: the path to git repository
```
AntonyMilneQB, line 42:
```suggestion
        template_path: path to a directory or a URL to a remote VCS repository supported
by `cookiecutter` 
```
This is actually very flexible (see `STARTER_ARG_HELP` where I copied this from).

kedro/framework/cli/starters.py
```diff
@@ -22,22 +24,53 @@
...
+        directory: by default it will look for `alias`, optionally override with custom directory.
```
AntonyMilneQB, line 43:
```suggestion
        directory: optional directory inside the repository where the starter resides.
```
(copied from `DIRECTORY_ARG_HELP`)

kedro/framework/cli/starters.py
```diff
@@ -22,22 +24,53 @@
...
+        origin: preserved field used by kedro internally to determine where the starter
+        comes from, users do not need to provide this field.
```
AntonyMilneQB, line 45:
```suggestion
        origin: reserved field used by kedro internally to determine where the starter
        comes from, users do not need to provide this field.
```

kedro/framework/cli/starters.py
```diff
@@ -62,6 +95,65 @@ def _remove_readonly(func: Callable, path: Path, excinfo: Tuple):  # pragma: no
...
+    starter_specs = _OFFICIAL_STARTER_SPECS
+
+    for starter_entry_point in get_entry_points(name="starters"):
+        origin = starter_entry_point.module.split(".")[0]
+        specs = _safe_load_entry_point(starter_entry_point) or []
+        for spec in specs:
+            if not isinstance(spec, KedroStarterSpec):
+                click.echo(
+                    f"The starter configuration loaded from module {origin}"
+                    f"should be a 'KedroStarterSpec', got '{type(spec)}' instead"
```
AntonyMilneQB, line 128: Maybe put this in red or whatever we normally do the errors in `click.echo`?

kedro/framework/cli/starters.py
```diff
@@ -62,6 +95,65 @@ def _remove_readonly(func: Callable, path: Path, excinfo: Tuple):  # pragma: no
...
+                click.secho(
+                    f"Starter alias `{spec.alias}` from `{origin}` "
+                    f"has been ignored as it is already defined by"
+                    f"`{starter_specs[spec.alias].origin}`",
+                    fg="yellow",
+                )
+            else:
+                spec.origin = origin
+                starter_specs[spec.alias] = spec
+    print(starter_specs)
```
AntonyMilneQB, line 140:
```suggestion
```

kedro/framework/cli/starters.py
```diff
@@ -149,11 +238,26 @@ def starter():
...
-        {alias: repo_url.format(alias=alias)} for alias in sorted(_STARTER_ALIASES)
-    ]
-    click.echo(yaml.safe_dump(output))
+    starters_dict = _get_starters_dict()
+
+    # Group all specs by origin as nested dict and sort it.
+    sorted_starters_dict: Dict[str, Dict[str, KedroStarterSpec]] = {
+        origin: dict(sorted(starters_dict_by_origin))
+        for origin, starters_dict_by_origin in groupby(
+            starters_dict.items(), lambda item: item[1].origin
```
AntonyMilneQB, line 247: Why is it `item[1]`? Could we just do `starters_dict.values()` and then `lambda value: value.origin`? I guess not because then you lose the keys?

tests/framework/cli/test_starters.py
```diff
@@ -78,10 +82,46 @@ def test_starter_list(fake_kedro_cli):
...
         assert alias in result.output
 
 
+def test_starter_list_with_starter_plugin(fake_kedro_cli, entry_point):
+    """Check that `kedro starter list` prints out the plugin starters."""
+    entry_point.load.return_value = [KedroStarterSpec("valid_starter", "valid_path")]
+    entry_point.module = "valid_starter"
+    result = CliRunner().invoke(fake_kedro_cli, ["starter", "list"])
+    assert result.exit_code == 0, result.output
+    assert "valid_starter" in result.output
```
AntonyMilneQB, line 95: Minor suggestion: this test could be made a bit stricter since `assert "valid_starter" in result.output` is quite weak, especially when "valid_starter" is both the starter alias and the module name.

So maybe set `entry_point.module = "valid_starter_module"` and test the `result.output` a bit more thoroughly.

tests/framework/cli/test_starters.py
```diff
@@ -78,10 +82,46 @@ def test_starter_list(fake_kedro_cli):
...
+            "should be a 'KedroStarterSpec'",
+        ),
+        (
+            [
+                KedroStarterSpec("duplicate", "duplicate"),
+                KedroStarterSpec("duplicate", "duplicate"),
+            ],
+            "has been ignored as it is already defined by",
+        ),
+    ],
```
AntonyMilneQB, line 112: Can we add a case for `KedroStarterSpec` is instantiated incorrectly?

features/starter.feature
```diff
@@ -0,0 +1,9 @@
+Feature: List Kedro Starters
+
+  Scenario: List all starters with custom starters from plugin
+    Given I have prepared a config file
+    And I have installed the test plugin
+    And I have run a non-interactive kedro new with starter
```
AntonyMilneQB, line 6: You shouldn't actually need this step?

features/starter.feature
```diff
@@ -0,0 +1,9 @@
+Feature: List Kedro Starters
```
AntonyMilneQB, line 1:
```suggestion
Feature: List Kedro starters
```

docs/source/extend_kedro/plugins.md
```diff
@@ -44,6 +44,32 @@ Once the plugin is installed, you can run it as follows:
 kedro to_json
 ```
 
+## Extending Starters Template
```
AntonyMilneQB, line 47:
```suggestion
## Extend starter aliases
```

docs/source/extend_kedro/plugins.md
```diff
@@ -44,6 +44,32 @@ Once the plugin is installed, you can run it as follows:
...
+You can extend `kedro starter list` and `kedro new` by introducing your own `cookiecutter` template.
```
AntonyMilneQB, line 48:
```suggestion
It is possible to extend the list of starter aliases built into Kedro. This means that a [custom Kedro starter](create_kedro_starters.md) can be used directly through the `starter` argument in `kedro new` rather than needing to explicitly provide the `template` and `directory` arguments. A custom starter alias behaves in the same way as an official Kedro starter alias and is also picked up by `kedro starter list`. 
```

Also would be good to link something on the "create kedro starters" page point to this new section. Maybe a `{note}` at the end saying you can add an alias by creating a plugin using `kedro.starters` entrypoint, linking to here.

docs/source/extend_kedro/plugins.md
```diff
@@ -44,6 +44,32 @@ Once the plugin is installed, you can run it as follows:
...
+```python
+starters = [
+    KedroStarterSpec(
+        alias="test_plugin_starter",
+        template_path="https://github.com/kedro-org/kedro-starters/",
+        directory="pandas-iris",
+    )
+]
+
+The `directory` argument is optional, it is useful when you have multiple templates like [kedro-starters](https://github.com/kedro-org/kedro-starters). If you only have one template, your top-level directory will be treated as the template. You can take the [pandas-iris](https://github.com/kedro-org/kedro-starters/tree/main/pandas-iris) example for reference.
```
AntonyMilneQB, line 61:
```suggestion
The `directory` argument is optional and should be used when you have multiple templates in one repository as for the [official kedro-starters](https://github.com/kedro-org/kedro-starters). If you only have one template, your top-level directory will be treated as the template. You can take the [pandas-iris](https://github.com/kedro-org/kedro-starters/tree/main/pandas-iris) example for reference.
```

docs/source/extend_kedro/plugins.md
```diff
@@ -44,6 +44,32 @@ Once the plugin is installed, you can run it as follows:
...
+## Extending Starters Template
+You can extend `kedro starter list` and `kedro new` by introducing your own `cookiecutter` template.
+
+You need to extend the starters by providing a list of  `KedroStarterSpec`, in this example it is defined in a file called `plugin.py`.
+
+```python
+starters = [
+    KedroStarterSpec(
+        alias="test_plugin_starter",
+        template_path="https://github.com/kedro-org/kedro-starters/",
```
AntonyMilneQB, line 56: Does this work? We have `git+https://github.com/kedro-org/kedro-starters.git` in the source code?

Edit: just tried it and it does!

kedro/framework/cli/starters.py
```diff
@@ -79,9 +171,7 @@ def create_cli():  # pragma: no cover
 @click.option("--starter", "-s", "starter_name", help=STARTER_ARG_HELP)
 @click.option("--checkout", help=CHECKOUT_ARG_HELP)
 @click.option("--directory", help=DIRECTORY_ARG_HELP)
-def new(
-    config_path, starter_name, checkout, directory, **kwargs
-):  # pylint: disable=unused-argument
+def new(config_path, starter_name, checkout, directory, **kwargs):
```
AntonyMilneQB, line 174:
```suggestion
def new(config_path, starter_alias, checkout, directory, **kwargs):
```
Needs to be changed in a few places.

AntonyMilneQB (approved):
Beautiful work all round! This looks really nice and clean now, I'm very happy with how it turned out. Great work on the tests and docs also!

One small concern I only just thought of. Given what you found before about the slowness of Python entrypoints, I guess this makes our `kedro new` and `kedro starter list` commands even slower than they currently are 😬 Not something to worry about now, but might make dealing with CLI slowness slightly more pressing.

docs/source/extend_kedro/plugins.md
AntonyMilneQB: Maybe we should be consistent about how we do it though. Probably URL to a git repo is a bit nicer than the `git+` address.

kedro/framework/cli/utils.py
noklam: I think the main issue is we can't control what the plugin does, the slowness is mainly due to loading up tons of python libraries when we don't need it. (e.g. `kedro-viz` dependencies are all loaded even though we are just doing the discovery command `kedro`).

This may be one thing we can consider when we are chopping off some kedro CLI and re-organizing some of them into sub-commands, there are rooms to improve the CLI performance.

For the template plugin - I don't think it should carry many libraries, so I am not too concerned this would slow down the CLI too much.

kedro/framework/cli/starters.py
noklam: @AntonyMilneQB I actually just discovered this when I am trying to combine the e2e test steps!

features/steps/cli_steps.py
```diff
@@ -222,19 +222,21 @@ def add_test_jupyter_nb(context):
         test_nb_fh.write(TEST_JUPYTER_ORG)
 
 
-@given("I have run a non-interactive kedro new with starter")
-@when("I run a non-interactive kedro new with starter")
-def create_project_with_starter(context):
+@given('I have run a non-interactive kedro new with starter "{starter}"')
+@when('I run a non-interactive kedro new with starter "{starter}"')
+def create_project_with_starter(context, starter):
```
MerelTheisenQB, line 227: Could the default for `starter` be `None` so we don't have to add "default" to all tests, or does that not work for the behave tests?

MerelTheisenQB (approved):
This is really excellent work @noklam ! I hope you manage to fix the builds soon so we can get it merged 😄 🚀

idanov (approved):
Really loving this addition, great work 🎉

noklam:
The CI was failing because of a bug, once we have the fix the PR should pass again.

The bug was caused by the spawn process (Windows or MacOS >=3.8), because it will re-import everything.

The story roughly goes like this:
1. so basically the import from `kedro.framework.cli.starters` imports `kedro.framework.cli` en route
2. in `__init__.py`, it does `from .cli import main`
3. when cli is imported, it runs this: `import kedro.config.default_logger`, which reset the `logger` to the framework defaults.

noklam:
@AntonyMilneQB Thanks a lot for your reviews!

PR #1594, opened 2022-06-07
Bump the upper bound for Flake8 dependency to <5.0

deepyaman:
Signed-off-by: Deepyaman Datta <deepyaman.datta@utexas.edu>

## Development notes

~Pending #1595 (as is any other code PR).~

FYI relaxed the `importlib_metadata` lower bound to be compatible with https://github.com/PyCQA/flake8/blob/c6882772e1640ff4c59f6d84986377813cf9236a/setup.cfg#L45.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1594"></a>

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ cookiecutter~=1.7.0
 dynaconf>=3.1.2,<4.0.0
 fsspec>=2021.4, <=2022.1
 gitpython~=3.0
-importlib_metadata>=4.4 # Compatible with Python 3.10 importlib.metadata
+importlib_metadata>=3.6; python_version < '3.10' # The "selectable" entry points were introduced in `importlib_metadata` 3.6 and Python 3.10.
 jmespath>=0.9.5, <1.0
 pip-tools~=6.5
 pluggy~=1.0.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ cookiecutter~=1.7.0
 dynaconf>=3.1.2,<4.0.0
 fsspec>=2021.4, <=2022.1
 gitpython~=3.0
-importlib_metadata>=3.6; python_version < '3.10' # The "selectable" entry points were introduced in `importlib_metadata` 3.6 and Python 3.10.
+importlib_metadata>=3.6  # The "selectable" entry points were introduced in `importlib_metadata` 3.6 and Python 3.10.
 jmespath>=0.9.5, <1.0
 pip-tools~=6.5
 pluggy~=1.0.0
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -20,11 +20,12 @@
 * The file `conf/base/logging.yml` is now optional. See [our documentation](https://kedro.readthedocs.io/en/0.18.2/logging/logging.html) for details.
 
 ## Bug fixes and other changes
-* Bumped `pyyaml` upper-bound to make Kedro compatible with the [pyodide](https://pyodide.org/en/stable/usage/loading-packages.html#micropip) stack.
+* Bumped `pyyaml` upper bound to make Kedro compatible with the [pyodide](https://pyodide.org/en/stable/usage/loading-packages.html#micropip) stack.
 * Updated project template's Sphinx configuration to use `myst_parser` instead of `recommonmark`.
 * Reduced number of log lines by changing the logging level from `INFO` to `DEBUG` for low priority messages.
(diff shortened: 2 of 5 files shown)


requirements.txt
```diff
@@ -5,7 +5,7 @@ cookiecutter~=1.7.0
 dynaconf>=3.1.2,<4.0.0
 fsspec>=2021.4, <=2022.1
 gitpython~=3.0
-importlib_metadata>=4.4 # Compatible with Python 3.10 importlib.metadata
+importlib_metadata>=3.6; python_version < '3.10' # The "selectable" entry points were introduced in `importlib_metadata` 3.6 and Python 3.10.
```
deepyaman, line 8:
```suggestion
importlib_metadata>=3.6  # The "selectable" entry points were introduced in `importlib_metadata` 3.6 and Python 3.10.
```
The alternative is to do something like:

```python
try:
    import importlib.metadata as importlib_metadata
except ImportError:
    import importlib_metadata
```

(The `importlib_metadata` or some other name needs to be kept consistent, so that the patching in tests could work.)

This would avoid the "unnecessary" `importlib_metadata` import for Python 3.10, but it seems like this would keep complexity lower, and `importlib_metadata` shouldn't drag in extra requirements (to affect compatibility) anyway.

deepyaman:
No idea what broke the `import-linter` build on Windows using 3.10, as the issue also manifests in https://github.com/kedro-org/kedro/pull/1594/commits/940ffa38f1f5cca77e70b6ea975408895693ec4d, which is without the `importlib_metadata` import altogether.

Furthermore, version of `importlib_metadata` should be the same here as it is on `main`...

requirements.txt
```diff
@@ -5,7 +5,7 @@ cookiecutter~=1.7.0
 dynaconf>=3.1.2,<4.0.0
 fsspec>=2021.4, <=2022.1
 gitpython~=3.0
-importlib_metadata>=4.4 # Compatible with Python 3.10 importlib.metadata
+importlib_metadata>=3.6  # The "selectable" entry points were introduced in `importlib_metadata` 3.6 and Python 3.10.
```
MerelTheisenQB, line 8: I find this comment slightly confusing.. Do you mean selectable entry points only work with Python 3.10, or that `importlib_metadata` 3.6 introduces selectable entry points and is also compatible with python 3.10?

MerelTheisenQB (approved):
Minor comment, but other than that happy to have this merged! 😃

requirements.txt
deepyaman: `importlib_metadata` introduces new features that are later merged into CPython `importlib.metadata`. In this case, the "selectable" entry points feature--used by Kedro--was introduced in `importlib_metadata` 3.6, and thus also included in `importlib.metadata` natively in 3.10 (actually, `importlib_metadata` 4.4 was merged into Python 3.10, but Python 3.8 only has version `importlib_metadata` 1.4 features, and Python 3.9 didn't introduce anything on this front).

The comment is actually just copied from the compatibility note at the end of [this section](https://docs.python.org/3.10/library/importlib.metadata.html#entry-points).

deepyaman: As mentioned in https://github.com/kedro-org/kedro/pull/1594#discussion_r892941129, I don't think it's worth doing a conditional import. We can drop the `importlib_metadata` third-party dependency when Kedro drops support for Python 3.9 lol.

noklam (approved):

requirements.txt
```diff
@@ -5,7 +5,7 @@ cookiecutter>=2.1.1, <3.0
 dynaconf>=3.1.2, <4.0
 fsspec>=2021.4, <=2022.1
 gitpython~=3.0
-importlib_metadata>=4.4 # Compatible with Python 3.10 importlib.metadata
+importlib_metadata>=3.6  # The "selectable" entry points were introduced in `importlib_metadata` 3.6 and Python 3.10.
```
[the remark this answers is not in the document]
deepyaman: @noklam The way it works is that features are brought over to the standard lib in certain releases. Therefore, Python 3.10 `importlib.metadata` contains everything released up through `importlib_metadata==4.4`. This includes stuff from `importlib_metadata==3.6`.

I see you edited out the comment about it affecting the Flake8 dependency, but just for posterity: [3.8.0 introduced the bound dependency `importlib_metadata<4.4`](https://github.com/PyCQA/flake8/blob/3.8.0/setup.cfg#L51).

Issue #1414, opened 2022-03-11
Update to cookiecutter 2.02 


ghoersti:
## Description
**Is your feature request related to a problem? A clear and concise description of what the problem is:** 

I want to be able to use `{{uuid4()}} in cookie cutter prompts.  We have a process which requires it to be part of the template. 
Cookie cutters Docs say that this functionality is available in 1.73 , but this is not true. 
[According to this ](https://github.com/cookiecutter/cookiecutter/pull/1493#issuecomment-898959028) . 

I know that the release is not yet stable for cookie cutter 2.0.2, but I just wanted to get transparency that the cookie cutter 1.7.3 features are not actually available until 2.xx namely `cookiecutter.extensions.UUIDExtension`

## Context
**Why is this change important to you? How would you use it? How can it benefit other users?**
Currently as a stub I am using `"{{ random_ascii_string(24) }}"` but this is not satisfactory. 
We have a script to work around it, but it would be very useful for templates to come out of the box with the ability to have UUID functionality from cookie cutter, without making a custom jinja extension, or custom code.

## Possible Implementation
(Optional) Suggest an idea for implementing the addition or change.
Bump to `cookiecutter==2.02'  eventually 

## Possible Alternatives

- Jinja extension
- Custom script that does all the stuff `prompts.yml` would ask for  then pass to config `kedro new --config myconfig --starter my_starter`
- Its not a stable release so this can likely just be closed, I just wanted to get the awareness out there that it is effecting kedro users in some way.

AntonyMilneQB:
Hi @ghoersti and thanks for raising this very well written issue. Sorry no one responded before - we don't check issues on kedro-starters so much so I think this was inadvertently missed. I've moved this to the main repo for better visibility, and also because `cookiecutter` is a core kedro requirement so if we change anything then we would need to bump it on this repo rather than the starters repo.

I'm curious about how you even know about cookiecutter 2.0.2? I see the latest docs on their website are for 2.0.2, but there's not been any release of it on their repo or PyPI. Is it possible to install 2.0.2 somehow already?

ghoersti:
@AntonyMilneQB  Thank you for moving this for visibility!  Well you can actually see the docs are out for it [here](https://cookiecutter.readthedocs.io/en/2.0.2), but I came across some post some where I cant find anymore. 

Current cookie master is on `2.0.0 `. For usage I just use the plain old.

```
git clone https://github.com/cookiecutter/cookiecutter.git
[...]

PR #1601, opened 2022-06-09
Bump `cookiecutter` to address known vulnerability

### LINKED ISSUE #1600: Security issue with cookiecutter<2.1.1
## Description
Snyk is complaining about a security issue with the ``cookiecutter``  (1.7.0) used in kedro. I don't know if it is valid or not, but it breaks my CI workflow when installing kedro. 

image
 

## Context

``kedro`` cannot be installed in an enterprise context, and my [kedro-mlflow CI](https://github.com/Galileo-Galilei/kedro-mlflow/pull/327) complains. This impacts all kedro versions. 

## Other remarks

I haven't seen a specific process to warn about potential security issues, but it does not feel right to open a public github issue though, did I miss something there? This one does not seem too big a threat (and snyk warns publicly about it), so I choose to report it there.

deepyaman:
Signed-off-by: Deepyaman Datta <deepyaman.datta@utexas.edu>

## Description
Closes https://github.com/kedro-org/kedro/issues/1414 and closes #1600.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1601"></a>

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,8 @@
 anyconfig~=0.10.0
 cachetools~=4.1
 click<9.0
-cookiecutter>=2.1.1
-dynaconf>=3.1.2, <4.0.0
+cookiecutter>=2.1.1, <3.0
+dynaconf>=3.1.2, <4.0
 fsspec>=2021.4, <=2022.1
 gitpython~=3.0
 importlib_metadata>=4.4 # Compatible with Python 3.10 importlib.metadata
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,8 @@
 anyconfig~=0.10.0
 cachetools~=4.1
 click<9.0
-cookiecutter~=1.7.0
-dynaconf>=3.1.2,<4.0.0
+cookiecutter>=2.1.1
+dynaconf>=3.1.2, <4.0.0
 fsspec>=2021.4, <=2022.1
 gitpython~=3.0
 importlib_metadata>=4.4 # Compatible with Python 3.10 importlib.metadata
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -25,6 +25,7 @@
 * Reduced number of log lines by changing the logging level from `INFO` to `DEBUG` for low priority messages.
 * Kedro's framework-side logging configuration no longer performs file-based logging. Hence superfluous `info.log`/`errors.log` files are no longer created in your project root, and running Kedro on read-only file systems such as Databricks Repos is now possible.
 * The `root` logger is now set to the Python default level of `WARNING` rather than `INFO`. Kedro's logger is still set to emit `INFO` level messages.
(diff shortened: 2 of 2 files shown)


AntonyMilneQB:
Looking at the [cookiecutter releases](https://github.com/cookiecutter/cookiecutter/releases), it seems that 2.0.x never made it to PyPI. Releases went 1.7.3, 2.1.0, 2.1.1.

Looking at the breaking changes in 2.1.0 I don't think any should affect us, and all the tests pass here, but given how central cookiecutter is to kedro I think we should do a quick manual test to make sure that various `kedro new` commands are still working.

requirements.txt
```diff
@@ -1,8 +1,8 @@
 anyconfig~=0.10.0
 cachetools~=4.1
 click<9.0
-cookiecutter~=1.7.0
-dynaconf>=3.1.2,<4.0.0
+cookiecutter>=2.1.1
```
AntonyMilneQB, line 4: Should we add an upper bound of `<3`?

AntonyMilneQB: Yeah, after skimming through [this article](https://iscinumpy.dev/post/bound-version-constraints/), similar things elsewhere and from some recent experiences I think we should revisit the way we pin versions also. The bounds on CalVer packages we have should make sense though: black actually has [exactly the stability policy](https://black.readthedocs.io/en/stable/the_black_code_style/index.html#stability-policy) that you mention (although it doesn't really do "breaking" changes the same way at all). Other packages that are CalVer we have a kind of lagging pin where we upper bound to the last version we saw it was working.

So for now I would add `<3` just to be consistent with how we do other requirements, even though I think we should reconsider that general policy.

deepyaman: > black actually has [exactly the stability policy](https://black.readthedocs.io/en/stable/the_black_code_style/index.html#stability-policy) that you mention

TIL!

requirements.txt
```diff
@@ -1,8 +1,8 @@
...
+dynaconf>=3.1.2, <4.0.0
```
deepyaman, line 5:
```suggestion
dynaconf>=3.1.2, <4.0
```

requirements.txt
deepyaman, line 4:
```suggestion
cookiecutter>=2.1.1, <3.0
```

AntonyMilneQB (approved):
Thank you for this! Happy to approve assuming that you manually tested it and it works ok 😀

deepyaman:
> Looking at the [cookiecutter releases](https://github.com/cookiecutter/cookiecutter/releases), it seems that 2.0.x never made it to PyPI. Releases went 1.7.3, 2.1.0, 2.1.1.
> 
> Looking at the breaking changes in 2.1.0 I don't think any should affect us, and all the tests pass here, but given how central cookiecutter is to kedro I think we should do a quick manual test to make sure that various `kedro new` commands are still working.

I did `kedro new` and `kedro new --starter spaceflights` with both the current and latest `cookiecutter` versions, and the output is identical.

Issue #1350, opened 2022-03-15
Consistent nodes execution order with `SequentialRunner`


noklam:
## Description
I'm always frustrated when I execute an identical pipeline it can yield different results. The root cause of this issue is due to the fact that multiple combinations of nodes exists. Kedro only try to solve the DAGs by finding 1 possible solution, but it is not guaranteed to be the same. 

One workaround is to specify input/output of nodes to make sure there is only 1 possible solution, but this is not ideal as users has to maintain arbitrary dummy variables.

### What is missing in the DAGs?
Seed of random number generator. Consider a simple pipeline with 3 nodes:
```
A-- \    
      \   
       C   
      /   
    /   
B--
```
In this pipeline, there are 2 possible execution order with SequentialRunner, 1. A->B->C, 2. B->A->C. Although there are no strong preference whether 1/2 is better, it is better to stick with one of them, as the output can be changed. 

## Context
In data science/machine learning pipeline, setting a seed to ensure **reproducible** result are very common, and currently there are no easy way to achieve this. 

## Possible Implementation
Ensure the resolved nodes are sorted so it always run in the same order with SequentialRunner.

## Possible Alternatives
(Optional) Describe any alternative solutions or features you've considered.

datajoely:
So on this - we use the `toposort~=1.5` external library, since Kedro was released there is now a stdlib https://docs.python.org/3/library/graphlib.html module that does the same thing. I we do look into making this deterministic, it might be a good opportunity to adopt this too.

noklam:
Note: currently we are using a function called `toposort`, but actually there is a function call `toposort_flatten` will guarantee the node order

datajoely:
@noklam awesome!

AntonyMilneQB:
This is interesting - I'd never thought that individual node ordering would make any difference in practice so long as things are toposorted, even if that sort isn't deterministic.

@noklam just for my understanding, can you give a concrete example of something you might do where changing node execution order can alter the final output? Your diagram of nodes didn't come out correctly so it's a bit hard to understand.

datajoely:
I'll just add users do ask for this maybe once every two months, I've even seen people introduce fake nodes to force ordering.

noklam:
@AntonyMilneQB 
This has to deal with the fact that the random number generator do have an specific order.

Imagine a simple pipeline like this
```
train_test_split_data-- \    
                         \   
                         train_model   
                         /   
                        /   
Model initialization--
```
[...]

PR #1604, opened 2022-06-09
Consistent node execution order by sorting node with `Sequentialrunner`

noklam:

## Description

#1350 

## Checklist

### DIFF
--- a/kedro/pipeline/pipeline.py
+++ b/kedro/pipeline/pipeline.py
@@ -870,7 +870,7 @@ def _validate_transcoded_inputs_outputs(nodes: List[Node]) -> None:
         )
 
 
-def _topologically_sorted(node_dependencies) -> List[Set[Node]]:
+def _topologically_sorted(node_dependencies) -> List[List[Node]]:
     """Topologically group and sort (order) nodes such that no node depends on
     a node that appears in the same or a later group.
 
@@ -894,7 +894,11 @@ def _topologically_sorted(node_dependencies) -> List[Set[Node]]:
         return f"Circular dependencies exist among these items: {circular}"
 
     try:
-        return list(toposort(node_dependencies))
+        result = []
+        for dependencies in toposort(node_dependencies):
+            # Make sure it is sorted to have consistent execution order when run with SequentialRunner
+            result.append(sorted(dependencies))
+        return result
     except ToposortCircleError as exc:
         message = _circle_error_message(exc.data)
         raise CircularDependencyError(message) from exc


tests/pipeline/test_pipeline.py
```diff
@@ -588,6 +592,52 @@ def test_connected_pipeline(self, disjoint_pipeline):
...
+
+            Although we are not sure which permutation it is, but it should always output
+            the same permutation.
+
+            A-- \
+            B--- \
+            C---- F
+            D--- /
+            E-- /
+        """
```
AntonyMilneQB, line 612: This is nice 👍

tests/pipeline/test_pipeline.py
```diff
@@ -588,6 +592,52 @@ def test_connected_pipeline(self, disjoint_pipeline):
...
+
+        def multiconcat(*args):
+            return sum(args)
+
+        mock_hash = mocker.patch(f"{__name__}.Node.__hash__")
+        expected_sorted_nodes: List[List[Node]] = None
+
+        # Repeat 10 times so we can be sure it is not purely by chance
+        for _ in range(10):
+            mock_hash.return_value = random.randint(1, 1e20)
```
AntonyMilneQB, line 622: I don't think this is doing what you want it to do. This is currently fixing the hash of every `Node` instance in this pipeline to be the same. We don't want `hash(node1)` to be the same as `hash(node2)`. What we should have is:
```
        n1 = node(constant_output, None, "A")
        n2 = node(constant_output, None, "B")
        n3 = node(constant_output, None, "C")
        n4 = node(constant_output, None, "D")
        n5 = node(constant_output, None, "E")
        n6 = node(multiconcat, ["A", "B", "C", "D", "E"], "F")

       # You don't have to nest these, you can put them all in one with block
       # But actually for Python < 3.10 it's super ugly formatting still: https://stackoverflow.com/questions/3024925/create-a-with-block-on-several-context-managers
        for _ in range(10):
            with mock.patch.object(n1, "__hash__", random.randint(1, 1e20)):
                with mock.patch.object(n2, "__hash__", random.randint(1, 1e20)):
                    with mock.patch.object(n3, "__hash__", random.randint(1, 1e20)):
                        with mock.patch.object(n4, "__hash__", random.randint(1, 1e20)):
                            with mock.patch.object(n5, "__hash__", random.randint(1, 1e20)):
                                with mock.patch.object(n6, "__hash__", random.randint(1, 1e20)):
                                    inverted_fork_dags = Pipeline([n1, n2, n3, n4, n5, n6])
                                    # use inverted_fork_dags.nodes
```
... but this is still not a great test because the current code in `main` doesn't fail it 😬 

After spending a looooong time playing around with this, I think it might just not be worth writing a test for it at all all... So long as it works as it should in manual testing then I think we're fine.

Happy to explain more about what I discovered while playing around with the testing here if you want to hear. It's certainly a tricky one.

AntonyMilneQB (approved):
Very nice fix! I actually think the test here isn't doing what you think and needs to change, but also I'm not sure there's a good way to write a test for this at all. So might just be best overall to remove it.

RELEASE.md
```diff
@@ -25,9 +25,11 @@
 * Reduced number of log lines by changing the logging level from `INFO` to `DEBUG` for low priority messages.
 * Kedro's framework-side logging configuration no longer performs file-based logging. Hence superfluous `info.log`/`errors.log` files are no longer created in your project root, and running Kedro on read-only file systems such as Databricks Repos is now possible.
 * The `root` logger is now set to the Python default level of `WARNING` rather than `INFO`. Kedro's logger is still set to emit `INFO` level messages.
+* Kedro pipeline will have consistent execution order given the same set of nodes when using with `SequentialRunner`.
```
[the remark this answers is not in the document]
MerelTheisenQB: I don't know if this is any better to be honest 😅 "Added sorting of nodes for the `SequentialRunner` to facilitate consistent execution order across multiple runs. "

MerelTheisenQB (approved):
This looks good to me, and even though it would be ideal to have a test for this I recognise it isn't trivial to test this well. I'd be happy to merge this without a specific test if we're satisfied with the results from the manual test 👍

PR #1612, opened 2022-06-12
Generic typings abstract dataset

### LINKED ISSUE #1564: Feature: make AbstractDataset take data type as generic parameter.
## Description

Abstract dataset is currently incompletely typed, as _load and _save use "Any".

## Context

I would like to write typed code.

## Possible Implementation

Use:
```
class AbstractDataset(Generic[DS]): ...
```
If necessary for deprecation, could have something like:
```
class BaseDataset(Generic[DS]): ...

class AbstractDataset(BaseDataset[Any]): ...
```
and deprecate `AbstractDataset` in favor of `BaseDataset`.

daBlesr:
Signed-off-by: Niels Drost <codingdutchman@gmail.com>

## Description
Resolves #1564 

## Development notes
Only generic types have been added for `AbstractDataSet` and `AbstractVersionedDataSet`, to define load and save types.
In all classes in extras that use before mentioned classes, the correct types have been added. 
Checked using mypy.

## Checklist

❌  Dit not add tests, because it has no run time influence.

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1612"></a>


daBlesr:
Hi, could anyone help me fixing the above build errors, as I cannot try for myself with Sphinx? (@AntonyMilneQB 😄)

The error is:
```sphinx.errors.SphinxWarning: /home/circleci/project/kedro/extras/datasets/api/api_dataset.py:docstring of kedro.io.core.AbstractDataSet.load::py:class reference target not found: _DO```
where `_DO` is on of the generic types of `AbstractDataSet`. Not sure, what I am supposed to do here.

File: kedro/io/core.py [hunk 1]
```diff
@@ -66,7 +66,11 @@ class VersionNotFoundError(DataSetError):
     pass
 
 
-class AbstractDataSet(abc.ABC):
+_DI = TypeVar("_DI")
+_DO = TypeVar("_DO")
```
MerelTheisenQB, line 70: What do `_DI` and `_DO` stand for?

RELEASE.md
```diff
@@ -25,6 +25,7 @@
 * Reduced number of log lines by changing the logging level from `INFO` to `DEBUG` for low priority messages.
 * Kedro's framework-side logging configuration no longer performs file-based logging. Hence superfluous `info.log`/`errors.log` files are no longer created in your project root, and running Kedro on read-only file systems such as Databricks Repos is now possible.
 * The `root` logger is now set to the Python default level of `WARNING` rather than `INFO`. Kedro's logger is still set to emit `INFO` level messages.
+* Generic typing for `AbstractDataSet` and `AbstractVersionedDataSet`.
```
MerelTheisenQB, line 28:
```suggestion
* Added generic typing for `AbstractDataSet` and `AbstractVersionedDataSet` as well as typing to all datasets.
```

MerelTheisenQB (commented):
Thanks for taking this up @daBlesr ! 🙂 
A small question from my side, but generally happy with how this looks.

kedro/io/core.py
daBlesr: They refer to `DataInput` (save) and `DataOutput` (load), but I'd be happy with changing it to something else. Generics are usually typed with one- or two-letter acronyms, and these two made the most sense to me.

kedro/extras/datasets/pandas/json_dataset.py
```diff
@@ -22,7 +22,7 @@
 logger = logging.getLogger(__name__)
 
 
-class JSONDataSet(AbstractVersionedDataSet):
+class JSONDataSet(AbstractVersionedDataSet[pd.DataFrame, Any]):
```
AntonyMilneQB, line 25:
```suggestion
class JSONDataSet(AbstractVersionedDataSet[pd.DataFrame, pd.DataFrame]):
```
And change the type signature for `_load` 🙂 It does't look right at the moment.

kedro/extras/datasets/holoviews/holoviews_writer.py
```diff
@@ -21,7 +21,7 @@
 HoloViews = TypeVar("HoloViews")
 
 
-class HoloviewsWriter(AbstractVersionedDataSet):
+class HoloviewsWriter(AbstractVersionedDataSet[HoloViews, str]):
```
AntonyMilneQB, line 24:
```suggestion
class HoloviewsWriter(AbstractVersionedDataSet[HoloViews, None]):
```
```suggestion
class JSONDataSet(AbstractVersionedDataSet[pd.DataFrame, pd.DataFrame]):
```
And change the type signature for `_load` 🙂

kedro/io/core.py
AntonyMilneQB: I like it 👍

AntonyMilneQB (approved):
Awesome PR, thank you so much! I'll check the sphinx build out. Unfortunately I think I know what the problem will be and it might not be so easy to fix 😬 

I've made a few suggestions, mainly based on the fact that our current type hints weren't right... I'm also wondering what would be the best way to handle these two cases:
* `_save` method isn't really implemented and just raises an exception (e.g. `APIDataSet`). Currently it is `Any` which makes sense but maybe there's a better convention here?
* `_load` method isn't really implemented. Should this be `None` or is there some better way to do it?

Happy to go with whatever you think is best for these, so long as it's done consistently throughout.

Also, out of curiosity, why did you pick up this issue? Did you independently find the need to  do #1564 or just happen across the issue and thought it would be a good one to try?

kedro/io/core.py
MerelTheisenQB: Okay that makes sense! I'd just leave a comment that's what they refer to for future reference.

MerelTheisenQB (approved):
Thanks for contributing this @daBlesr ! 😄

AntonyMilneQB:
I've fixed the docs in [c4f50bf](https://github.com/kedro-org/kedro/pull/1612/commits/c4f50bf62df98dcc71f397cecdef7837220cacba). Unfortunately there's a bit of an annoyance with how the docs for the datasets render.
image

Compared to before:
image

To be honest the way these docs render the `load`/`save` methods for datasets was completely useless before anyway, so I'm not too worried about this. It just looks a bit weird.

daBlesr:
Interesting case. Python does have a type `NoReturn` when a method unconditionally raises an exception:  [NoReturn PEPs](https://peps.python.org/pep-0484/#the-noreturn-type). My proposal:

* `_load` and `_save` return type if raises exception: `NoReturn`.
* `_save` data type if raises exception: `None`.

Practically with the generic typing this would look like:
- No input, does have output: `AbstractDataSet[None, request.Response]`
- Input, but no output: `AbstractDataSet[DataFrame, NoReturn]`

Let me know if you agree, and I'll make sure this usage is consistent.

daBlesr:
Wrt to your other question, yea I came across it and I thought it'd be a good one to do my first contribution to Kedro.

> Also, out of curiosity, why did you pick up this issue? Did you independently find the need to do #1564 or just happen across the issue and thought it would be a good one to try?

The latter, I thought it'd be a good issue for a first contribution 😬

AntonyMilneQB:
This sounds like a great proposal, please do go ahead with it! I didn’t know about `NoReturn` before but it definitely fits the bill here. Thank you very much for your contribution. It’s much appreciated.

Issue #1537, opened 2022-05-16
Document what the `ShelveStore` is for


merelcht:
## Description
Users are running into issues when using the `ShelveStore` and other tools that use multiprocessing, e.g. Prefect. See: https://github.com/kedro-org/kedro/issues/1442

This has been reported by several users.

## Implementation

- [X] It seems that even if a user isn't using the `ShelveStore`, kedro imports it and this interferes with multiprocessing used by other tools. Verify this is the case, and if so make sure Kedro _only_ imports `ShelveStore` when it's explicitly set as the session store. This has been solved in : https://github.com/kedro-org/kedro/pull/1614
- [ ] Clarify in docs + tutorials that the `ShelveStore` is not the default. Explain what it does, so hopefully users will only use it if really needed.

avan-sh:
I think I might have found a solution for this, happy to make other necessary changes on this. 
Also please highlight if there are any other instances of conflicts from other libraries other than 1442.

astrojuanlu:
Was the docs/tutorials part done? Sorry, I'm missing some context here

noklam:
@astrojuanlu No, but I guess the priority for that will be really low. We use to put ShelveStore as a commented line in `settings.py`. Recently I update that with #2587, in practice I don't think anyone use it at all.

Issue #1442, opened 2022-04-14
Kedro run blocks creation of new multiprocessing context


avan-sh:
## Description

An import of ShelveStore(which happens automatically during `kedro run`) automatically creates a multiprocessing context blocking users or packages that create a new multiprocessing context of a different start method.
**Note: Documenting an error raised by discord user.**

## Context
A user on discord raised the following error that occurs when adding `from pycbc.waveform import get_fd_waveform` import statement in `nodes.py`. Reference to the [thread](https://discord.com/channels/778216384475693066/963181015822651443)

## Steps to Reproduce
1. Have a running kedro project (eg: spaceflights)
2. Add  `from pycbc.waveform import get_fd_waveform` import statement to your project (any module that is imported at runtime eg: `nodes.py` or `settings.py`).
3. Run `kedro run` 

### Alternative option 1
Execute the following script.
```
from kedro.framework.session.store import ShelveStore
from pycbc.waveform import get_fd_waveform
```

Execute the following script.
```
from kedro.framework.cli import main
import multiprocessing

multiprocessing.set_start_method("fork")
```

## Expected Result

Pipeline should run without any error

## Actual Result
`kedro run` 

```
 from pycbc.waveform import get_fd_waveform
  File "/lib/python3.8/site-packages/pycbc/__init__.py", line 150, in <module>
    multiprocessing.set_start_method('fork')
  File "/lib/python3.8/multiprocessing/context.py", line 243, in set_start_method
    raise RuntimeError('context has already been set')
RuntimeError: context has already been set
```

## Your Environment
Include as many relevant details about the environment in which you experienced the bug:

* Kedro version used (`pip show kedro` or `kedro -V`): kedro, version 0.18.0
* Python version used (`python -V`): python 3.8.13
* Operating system and version: MacOS

## Analysis
An import of `kedro.framework.session.store.ShelveStore` occurs during a run which creates a context of multiprocessing. This raises error when pycbc tries to create a new context during its import (which it does as it uses fork while the default is spawn in MacOS).

## Temporary Solutions
**Option 1:** Comment ShelveStore class in `kedro.framework.session.store` module.

**Option 2:** Adding import to kedro executable. Updated kedro executable would be:
```
#!/opt/anaconda3/envs/k18/bin/python
# -*- coding: utf-8 -*-
import re
import sys
from pycbc.waveform import get_fd_waveform
from kedro.framework.cli import main

if __name__ == "__main__":
[...]

PR #1614, opened 2022-06-13
Refactor `ShelveStore` into a separate module

avan-sh:
## Description
This change refactors `ShelveStore` into a separate module to avoid conflict multiprocessing context.
Related to #1537 & #1442. 
This might possibly be a breaking change, so happy to drop this or merge this to develop.
Currently a draft, will update tests & docs in subsequent commits.

## Development notes
- Source of the error is [this](https://github.com/gwastro/pycbc/blob/d05f9043000f0af79cdbd30264cd986aeb120af3/pycbc/__init__.py#L148) line from pycbc & import of BaseSession(likely from one of  [1](https://github.com/kedro-org/kedro/blob/main/kedro/framework/project/__init__.py) or [2](https://github.com/kedro-org/kedro/blob/e454e16195160611821ab74dad793526a7fe2838/kedro/framework/session/session.py#L26). ) automatically setting multiprocessing context
- Tested this in a kedro project with the conflict due to `pycbc.waveform. get_fd_waveform` as mentioned in 1442

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1614"></a>


AntonyMilneQB:
@avan-sh thank you for this! Technically it's a breaking change, but so few people (anyone?) uses `ShelveStore` that I think we could regard it as a bug fix to be honest. Maybe it's even possible to make it non-breaking by maintaining the old import path through some trickery involving `__all__` or `__init__.py`? Not sure.

With that said, do we even want to keep `ShelveStore` in kedro at all? Realistically I don't think anyone uses it, and if we make `SQLiteStore` the comment-out default in settings.py then no one will even really know that it exists. @MerelTheisenQB

MerelTheisenQB:
Completely agree with @AntonyMilneQB that we can just get this in as a bug fix, because we don't expect many people to be using the `ShelveStore`. 

We might indeed remove it in the future, but I think until we've actually moved `SQLiteStore` to the kedro framework it would be good to keep the `ShelveStore` as an example of how to make a persistent store from the base one. I think it would be great to add a couple of sentences to the docs explaining that the `ShelveStore` is mainly a proof of concept and doesn't serve a very useful purpose.

noklam:
Thanks @avan-sh! Do you think this is ready for review?

noklam:
I marked it as ready for review anyway. 

The root cause is that `ShelveStore` has a class-level `Lock` which is created at import time, this creates some issues when people deploy Kedro on Lambda. #1537 will be a follow-up task but I am not sure if we will ever get it done.

I trace back to the commit when `Lock` is introduced:
Issue:
https://github.com/quantumblacklabs/private-kedro/pull/850

**commit**
https://github.com/quantumblacklabs/private-kedro/pull/863/commits/588356cbe675dbf1490b0cbd2a73127320e57e6e

The `Lock` is probably the right thing to do, it prevents two sessions from trying to save to the same location on disk at the same time. In reality, I don't think anyone is using it so it's not too important.

@MerelTheisenQB @AntonyMilneQB already agree that this is a minor breaking change since no one really use `ShelveStore`. I think we can go ahead and move it to a separate module.

noklam (approved):

MerelTheisenQB (approved):
We need to add this change to the release notes, then it's ready to be merged!

avan-sh:
@noklam @MerelTheisenQB  I've changed the release notes too. I'm guessing it's likely ready to be merged now.

noklam:
@avan-sh awesome! Thanks a lot for your contribution!

PR #1633, opened 2022-06-19
APIDataSet reduce number of arguments and move to Load args

### LINKED ISSUE #711: Additional options for APIDataSet (e.g. proxies)
## Description
Add proxies parameter to APIDataSet. Perhaps other options could also interesting?

## Context
When comparing the list of options available from [requests](https://requests.readthedocs.io/en/master/api/) with those available in the [APIDataSet](https://kedro.readthedocs.io/en/stable/kedro.extras.datasets.api.APIDataSet.html), there are few missing that might be interesting to add.
Personally, the most interesting to have would be the proxies option as discussed [here](https://discourse.kedro.community/t/an-elegant-way-of-setting-up-http-https-proxies/197). Before I used to set my proxies in my global env variables but I stumble upon cases where my pipeline has two APIs requiring two different proxies. Having the proxies options in the data catalog would have the added benefit that the proxies could be loaded using the TemplatedConfigLoader, thereby keeping the proxies confidential in the credentials.yml.

Other than proxies, I was wondering if there are other options that users might find to be of interest? Should we consider adding all options available in requests that are not yet available in the APIDataSet so that users have full flexibility?

## Possible Implementation
I would like to try solving this one if thats ok. Hopefully it requires just adding these options in extras/datasets/api/APIDataSet.py. I am not certain if these options would require extra tests. Any inputs?

daBlesr:
## Description
Resolves #711 

## Development notes
I have changed the implementation of APIDataSet by removing all the specific configuration parameters, and generally passing `load_args` to a request instance as keyword arguments. Also, users are now able to construct different kind of Authenticator objects to use with requests. 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1633"></a>


tests/extras/datasets/api/conftest.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+import requests_mock
+
+
+@pytest.fixture
+def requests_mocker():
```
afuetterer, line 6: Is there a difference between defining your own fixture and using the provided one from `requests_mock`? 
https://requests-mock.readthedocs.io/en/latest/pytest.html

You could just use `requests_mock` as a fixture?

RELEASE.md
```diff
@@ -35,7 +35,7 @@
 * Bumped the upper bound for the Flake8 dependency to <5.0.
 * `kedro jupyter notebook/lab` no longer reuses a Jupyter kernel.
 * Required `cookiecutter>=2.1.1` to address a [known command injection vulnerability](https://security.snyk.io/vuln/SNYK-PYTHON-COOKIECUTTER-2414281).
-
+* Reduced constructor arguments for `APIDataSet`, and made it consistent w.r.t. requests API, automatically enabling the full configuration domain: stream, certificates, proxies, and more.
```
AntonyMilneQB, line 38: Since this is going into `develop` it won't be released until 0.19. Hence the release docs belong in the "Upcoming Release 0.19.0" section above instead.

I'd put this in a new `### DataSets` section under "Breaking changes to the API" and also "Migration guide from Kedro 0.18.* to 0.19.*" that shows how someone should alter their existing `data`/`header` etc. arguments to sit under `load_args`. See https://github.com/kedro-org/kedro/blob/main/RELEASE.md#release-0180 for an example of how we've done this before.

kedro/extras/datasets/api/api_dataset.py
```diff
@@ -38,58 +44,35 @@ def __init__(
...
+            self._credentials.update(credentials)
 
-        if isinstance(auth, Iterable):
-            auth = tuple(auth)
+        self._auth = None
+        if credentials is not None:
+            self._auth = create_authenticator(class_type=auth_type, **credentials)
 
         self._request_args: Dict[str, Any] = {
+            **(load_args or {}),
```
AntonyMilneQB, line 72: Minor: could we put this after the other arguments? Just feels a bit weird to me to put the `**kwargs` first here when at least `url` and `method` are compulsory arguments.

AntonyMilneQB (commented):
Thanks very much for working on this @daBlesr! Really nice that someone has picked this up since it had got a bit stuck.

Moving arguments to `load_args` definitely makes sense to me here given the API of `requests.request` that you posted in the issue. Like you suggest, we aim to keep the interface as close to the underlying library as possible so this seems like a very good improvement to our `APIDataSet` 👍 

Your changes to `auth` also very much make sense and I totally see where you're coming from here, but I think they should not be part of this PR. Let's limit the scope here to just be the `load_args` part. The `auth` question is more of an open question (as are credentials in kedro in general) - I will post more about this in #711.

tests/extras/datasets/api/conftest.py
daBlesr: good one!

RELEASE.md
daBlesr: Yep, thanks for pointing this out!

kedro/extras/datasets/api/api_dataset.py
```diff
@@ -107,12 +90,13 @@ def _execute_request(self) -> requests.Response:
         return response
 
     def _load(self) -> requests.Response:
-        return self._execute_request()
+        with sessions.Session() as session:
+            return self._execute_request(session)
```
AntonyMilneQB, line 94: This is totally fine, I'm just curious why you chose to explicitly create the `session` given this seems to be what's done under the hood already by `requests`. I'm not very familiar with making requests with or without a sessions so would like to learn! Do you think it's just nice to be more explicit?

kedro/extras/datasets/api/api_dataset.py
```diff
@@ -21,83 +21,66 @@ class APIDataSet(AbstractDataSet):
...
 
-        auth = credentials or auth
+        if credentials is not None and "auth" in self._load_args:
+            raise ValueError("Cannot specify both auth and credentials.")
 
-        if isinstance(auth, Iterable):
-            auth = tuple(auth)
+        self._auth = None
+        if credentials is not None:
+            self._auth = tuple(credentials)
```
AntonyMilneQB, line 66: As I understand it, now there are two options for supplying credentials/auth:
1. if it's `auth` in `load_args` then it's assumed to be `AuthBase`. No casting to `tuple` in the case that it's an `Iterable`.
2. if it's `credentials` then it's assumed to be `Iterable` and we cast to `tuple`

This is a little different from now where `auth` and `credentials` can both be either `AuthBase` or `Iterable`. Hence it's a bit more of a breaking change than just moving things to `load_args`. I think this is fine (and an improvement on how we currently do it), but let's also mention it in docs/release notes. I'll make separate comments with suggestions for this.

kedro/extras/datasets/api/api_dataset.py
```diff
@@ -21,83 +21,66 @@ class APIDataSet(AbstractDataSet):
...
-                to the json kwarg in the requests object.
-                https://requests.readthedocs.io/en/master/user/quickstart/#more-complicated-post-requests
-            timeout: The wait time in seconds for a response, defaults to 1 minute.
-                https://requests.readthedocs.io/en/master/user/quickstart/#timeouts
-            credentials: same as ``auth``. Allows specifying ``auth`` secrets in
-                credentials.yml.
-
+            load_args: Additional parameters to be fed to requests.request.
+                https://docs.python-requests.org/en/latest/api/
+            credentials: Allows specifying secrets in credentials.yml.
```
AntonyMilneQB, line 53:
```suggestion
            credentials: Allows specifying secrets in credentials.yml.
                Expected format is ``('login', 'password')``.
```

kedro/extras/datasets/api/api_dataset.py
AntonyMilneQB, line 66: Small simplification?
```suggestion
        if credentials is not None:
            self._load_args["auth"] = tuple(credentials)
```

kedro/extras/datasets/api/api_dataset.py
```diff
@@ -21,83 +21,66 @@ class APIDataSet(AbstractDataSet):
...
         self._request_args: Dict[str, Any] = {
             "url": url,
             "method": method,
-            "data": data,
-            "params": params,
-            "headers": headers,
-            "auth": auth,
-            "json": json,
-            "timeout": timeout,
+            "auth": self._auth,
```
AntonyMilneQB, line 71:
```suggestion
```
part of same suggested simplifcation

kedro/extras/datasets/api/api_dataset.py
```diff
@@ -21,83 +21,66 @@ class APIDataSet(AbstractDataSet):
...
+            **self._load_args,
         }
 
     def _describe(self) -> Dict[str, Any]:
-        return dict(**self._request_args)
+        # prevent auth from logging
+        request_args_cp = self._request_args.copy()
+        request_args_cp.pop("auth", None)
+        return dict(**request_args_cp)
```
AntonyMilneQB, line 79: Good thinking!

tests/extras/datasets/api/test_api_dataset.py
```diff
@@ -17,41 +17,43 @@
...
+    def test_headers_in_request(self, requests_mock):
+        api_data_set = APIDataSet(
+            url=TEST_URL, method=TEST_METHOD, load_args={"headers": TEST_HEADERS}
+        )
+        requests_mock.register_uri(TEST_METHOD, TEST_URL, headers={"pan": "cake"})
+
+        response = api_data_set.load()
+
+        assert response.request.headers["key"] == "value"
+        assert response.headers["pan"] == "cake"
```
AntonyMilneQB, line 47: Please can you explain these lines? Is the first line the request and the second the response? Why isn't it `key=pan` and `value=cake`? 🤔

tests/extras/datasets/api/test_api_dataset.py
```diff
@@ -76,33 +80,75 @@ def test_http_error(self, requests_mocker, method):
...
+        encoded = base64.b64encode(f"{username}:{password}".encode("latin-1"))
+        return f"Basic {encoded.decode('latin-1')}"
+
+    @pytest.mark.parametrize(
+        "auth_kwarg",
+        [{"load_args": {"auth": ("john", "doe")}}, {"credentials": ("john", "doe")}],
+    )
+    def test_auth_sequence(self, requests_mock, auth_kwarg):
+        """
+        ``auth`` and ``credentials`` should be able to be any Iterable.
```
AntonyMilneQB, line 123: This isn't quite right. `auth` now only works with `tuple`: if you do `"auth": ["john", "doe"]` I think this will fail. Which is ok behaviour I think, just the comment is wrong. Probably it's only worth testing the `credentials` case on this test to be honest.

AntonyMilneQB (approved):
Awesome work, especially on the tests!! Thank you so much for this contribution. I have lots of small comments and questions but nothing big so very happy to approve this 😀 

One potential problem I foresee here is that it seems that some `requests` arguments expect `tuple` rather than just a general Iterable, notably `auth` and `cert`. This is a problem because when a user specifies these arguments in yaml, they always come out as a list. This was one of the original motivations for https://github.com/kedro-org/kedro/issues/1101.

The only fix I can think for this is for us to manually go through `load_args` and convert anything that is a `list` to a `tuple` (checking for `Iterable` won't work for at least `cert` because it could also be a string, which we don't want to convert to tuple!). That seems kind of ugly though. Interested what you think! Possibly our great configuration overhaul will provide a better solution here (like a native way of putting Python lists into yaml e.g. https://github.com/kedro-org/kedro/issues/1011).

Also interested in seeing you have any comments on this issue on credentials, partly inspired by the problems you've uncovered here! 🙏  https://github.com/kedro-org/kedro/issues/1646

kedro/extras/datasets/api/api_dataset.py
daBlesr: I did this because subclasses (the API Paginator) can then use this session to do more requests without overhead.

kedro/extras/datasets/api/api_dataset.py
daBlesr: I have reverted the implementation and added more tests: `AuthBase` instances can be provided now.

tests/extras/datasets/api/test_api_dataset.py
daBlesr: There is no difference between `requests_mock.last_request` and `response.request`. For clarity, I have changed all usages into `response.request`

daBlesr: > Please can you explain these lines? Is the first line the request and the second the response? Why isn't it key=pan and value=cake? 🤔

Requests headers (a bit of information you send along with the request) are not the same as response headers (a bit of information sent back by the server). In the above example, the requests headers are literally `{"key": "value"}`. The response headers are `{"pan": "cake"}`.

tests/extras/datasets/api/test_api_dataset.py
daBlesr: You are right, this does fail indeed. Thanks for pointing this out. I have added tests to make sure this does not happen.

tests/extras/datasets/api/test_api_dataset.py
```diff
@@ -111,60 +157,72 @@ def test_exists_http_error(self, requests_mocker, method):
...
-        api_data_set = APIDataSet(**kwargs)
-        requests_mocker.register_uri(
-            method,
-            TEST_URL_WITH_PARAMS,
-            headers=TEST_HEADERS,
-            text=TEST_TEXT_RESPONSE_DATA,
+        api_data_set = APIDataSet(
+            url=TEST_URL,
+            method=TEST_METHOD,
+            load_args={"cert": ("cert.pem", "privkey.pem")},
```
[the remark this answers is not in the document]
daBlesr: Correct, I have added an implementation to convert lists to tuples for some request parameters.

kedro/extras/datasets/api/api_dataset.py
AntonyMilneQB: Sounds good, thank you!

tests/extras/datasets/api/test_api_dataset.py
AntonyMilneQB: Got it, thank you!

AntonyMilneQB (approved):
The conversion from lists to tuples seems good, but I'm slightly confused about what you ended up doing with the types for `credentials`/`auth` - feels like you opted to go for something non-breaking here but didn't revert everything you intended to? I don't mind where we end up here (your original solution vs. this new one) so long as all the code, docstrings, release notes are consistent on it. Please do feel free to merge whenever you're happy with it 🙂 

Thank you again for all your work on this!

kedro/extras/datasets/api/api_dataset.py
```diff
@@ -21,83 +21,81 @@ class APIDataSet(AbstractDataSet):
...
+        """
+        if isinstance(value, List):
+            return tuple(value)
+        return value
+
     def _describe(self) -> Dict[str, Any]:
-        return dict(**self._request_args)
+        # prevent auth from logging
+        request_args_cp = self._request_args.copy()
+        request_args_cp.pop("auth", None)
```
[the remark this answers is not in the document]
daBlesr: I don't think it is necessary, the argument is the path to the certificates. Certificates are usually stored in fixed locations like `/etc/letsencrypt/live/{domainname}/fullchain.pem`

kedro/extras/datasets/api/api_dataset.py
```diff
@@ -21,83 +21,81 @@ class APIDataSet(AbstractDataSet):
...
         method: str = "GET",
-        data: Any = None,
-        params: Dict[str, Any] = None,
-        headers: Dict[str, Any] = None,
-        auth: Union[Iterable[str], AuthBase] = None,
-        json: Union[List, Dict[str, Any]] = None,
-        timeout: int = 60,
-        credentials: Union[Iterable[str], AuthBase] = None,
+        load_args: Dict[str, Any] = None,
+        credentials: Union[Tuple[str, str], List[str]] = None,
```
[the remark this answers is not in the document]
daBlesr: I'm still doubting about this, since technically it works. However, I think it might be confusing to the reader, since `APIDataSet` are automatically constructed from the catalog, and `AuthBases` cannot be inferred this way. So I am wondering, in what cases would this constructor type be used when using Kedro? At this moment, for a user that wants to do a request using an `AuthBase` instance, should imo take a different approach, such as extending `APIDataSet`. However, if you have other ideas, I'll put `AuthBase` in the typing.

afuetterer:
Hi, what is the status of this PR? Do you need more time working on this?

kedro/extras/datasets/api/api_dataset.py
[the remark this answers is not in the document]
AntonyMilneQB: Very slow reply I know, sorry!! 😅 The use case here would be that the `APIDataSet` is constructed directly using the Python API rather than from the data catalog. In this case you can easily pass arbitrary Python objects to the constructor. This is unusual but does occur and is something we support. So I've put `AuthBase` back in the type hint 🙁

tests/extras/datasets/api/test_api_dataset.py
```diff
@@ -13,96 +13,190 @@
...
+        return f"Basic {encoded.decode('latin-1')}"
+
+    @pytest.mark.parametrize(
+        "auth_kwarg",
+        [
+            {"load_args": {"auth": ("john", "doe")}},
+            {"load_args": {"auth": ["john", "doe"]}},
+            {"load_args": {"auth": HTTPBasicAuth("john", "doe")}},
+            {"credentials": ("john", "doe")},
+            {"credentials": ["john", "doe"]},
```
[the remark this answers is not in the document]
AntonyMilneQB: I've added this now.

RELEASE.md
```diff
@@ -6,6 +6,10 @@
 
 ## Breaking changes to the API
 
+### DataSets
+* Reduced constructor arguments for `APIDataSet` by replacing most arguments with a single constructor argument `load_args`. This makes it more consistent with other Kedro DataSets and the underlying `requests` API, and automatically enables the full configuration domain: stream, certificates, proxies, and more.
+
```
AntonyMilneQB, line 11:
```suggestion
```

kedro/extras/datasets/api/api_dataset.py
```diff
@@ -21,83 +22,83 @@ class APIDataSet(AbstractDataSet[None, requests.Response]):
...
+            **self._load_args,
         }
 
+    @staticmethod
+    def _convert_type(value: Any):
+        """
+        From the Catalog, iterations are provided as lists.
+        However, for some Parameters in the python Requests library,
+        only Tuples are allowed.
+        """
```
AntonyMilneQB, line 88:
```suggestion
        From the Data Catalog, iterables are provided as Lists.
        However, for some parameters in the Python requests library,
        only Tuples are allowed.
        """
```

kedro/extras/datasets/api/api_dataset.py
```diff
@@ -21,83 +22,83 @@ class APIDataSet(AbstractDataSet[None, requests.Response]):
...
-                https://requests.readthedocs.io/en/latest/user/quickstart/#more-complicated-post-requests
-            timeout: The wait time in seconds for a response, defaults to 1 minute.
-                https://requests.readthedocs.io/en/latest/user/quickstart/#timeouts
-            credentials: same as ``auth``. Allows specifying ``auth`` secrets in
-                credentials.yml.
-
+            load_args: Additional parameters to be fed to requests.request.
+                https://docs.python-requests.org/en/latest/api/
+            credentials: Allows specifying secrets in credentials.yml.
+                Expected format is ``('login', 'password')``.
```
AntonyMilneQB, line 55:
```suggestion
                Expected format is ``('login', 'password')`` if given as a tuple or list.
                An ``AuthBase`` instance can be provided for more complex cases.
```

AntonyMilneQB:
Hi @afuetterer, thanks for the push here and sorry for the huge delay - this was 99% ready to merge and then seemed to have stalled. I've just polished it off now and I think we should be ready to merge.

Thank you @daBlesr for all your hard work on this, it's much appreciated 🙏

MerelTheisenQB (approved):
This looks great! Thanks for the contribution @daBlesr ⭐ and for finishing the PR @AntonyMilneQB ⭐

kedro/extras/datasets/api/api_dataset.py
```diff
@@ -21,83 +22,84 @@ class APIDataSet(AbstractDataSet[None, requests.Response]):
...
-            json: The request payload, used for POST, PUT, etc requests, passed in
-                to the json kwarg in the requests object.
-                https://requests.readthedocs.io/en/latest/user/quickstart/#more-complicated-post-requests
-            timeout: The wait time in seconds for a response, defaults to 1 minute.
-                https://requests.readthedocs.io/en/latest/user/quickstart/#timeouts
-            credentials: same as ``auth``. Allows specifying ``auth`` secrets in
-                credentials.yml.
-
+            load_args: Additional parameters to be fed to requests.request.
+                https://docs.python-requests.org/en/latest/api/
```
AntonyMilneQB, line 53:
```suggestion
                https://requests.readthedocs.io/en/latest/api/#requests.request
```

PR #1634, opened 2022-06-20
Update docs on how to update Jupyter logging width

### LINKED ISSUE #1575: Disable rich logger in Jupyter
Unfortunately it seems that the rich logging produced when you do a kedro run does not render nicely in Jupyter (and hence also databricks etc.). The message column is very narrow so our logs are wrapped weirdly. See https://github.com/kedro-org/kedro/pull/1569#issuecomment-1137439071 and https://github.com/kedro-org/kedro/pull/1569#issuecomment-1137002651 for examples.

This isn't a show-stopper (a user can always just revert to plain console logging, as explained in our docs) but is quite annoying. It's a known problem with rich (https://github.com/Textualize/rich/issues/504#issuecomment-745992549) and not easy to fix by changing config options for the rich logging handler on our side.

So the solution for now is to just not use the rich logging handler in Jupyter. Possible solutions look quite hacky but might be one of the following:
1. Don't call `session._setup_logging` when `KedroSession.create` is called from `kedro.extras.extensions.ipython`
2. Manually swap the rich handler for console handler in `reload_kedro`. Something like this, but less hacky (also need to be careful so this works with and without project-side logging.yml):
```
rich_handler = [
  handler for handler in logging.getLogger().handlers if handler.name == "rich"
][0]
logging.getLogger().removeHandler(rich_handler)
logging.getLogger().addHandler(console_handler)
```
3. Add a new `jupyter` logging handler (which is the same as the `console` one) and use that in `reload_kedro`
  
N.B. there doesn't seem to be a problem with column wrapping in ipython, but by the nature of how the Kedro IPython extension works this will affect Jupyter and IPython in the same way. This is fine I think.

AntonyMilneQB:
## Description
Closes #1575.

I managed to complete a PR on rich (https://github.com/Textualize/rich/pull/2332), which does two things:
* increases rich Jupyter console size, so our messages appear much less narrowly wrapped now
* introduces two new environment variables so that users can configure this width themselves

This is a very good outcome because it means (a) we can continue doing rich logging in Jupyter and (b) there's no horrible hacking at all on our side 🎉 So all that's left here is a small piece of documentation.

image

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1634"></a>

### DIFF
--- a/docs/source/logging/logging.md
+++ b/docs/source/logging/logging.md
@@ -66,7 +66,7 @@ You must provide a value for both `COLUMNS` and `LINES` even if you only wish to
 
 ### Rich logging in Jupyter
 
-Rich also formats the logs in JupyterLab and Jupyter Notebook. The size of the output console does not adapt to your window but can be controlled through the `JUPYTER_COLUMNS` and `JUPYTER_LINES` environment variables, which default to 115 and 100 respectively.
+Rich also formats the logs in JupyterLab and Jupyter Notebook. The size of the output console does not adapt to your window but can be controlled through the `JUPYTER_COLUMNS` and `JUPYTER_LINES` environment variables. The default values (115 and 100 respectively) should be suitable for most users, but if you require a different output console size then you should alter the values of `JUPYTER_COLUMNS` and `JUPYTER_LINES`.
 
 ## Perform logging in your project
 
--- a/docs/source/logging/logging.md
+++ b/docs/source/logging/logging.md
@@ -64,6 +64,10 @@ export COLUMNS=120 LINES=25
 You must provide a value for both `COLUMNS` and `LINES` even if you only wish to change the width of the log message. Rich's default values for these variables are `COLUMNS=80` and `LINE=25`.
 ```
 
+### Rich logging in Jupyter
+
+Rich also formats the logs in JupyterLab and Jupyter Notebook. The size of the output console does not adapt to your window but can be controlled through the `JUPYTER_COLUMNS` and `JUPYTER_LINES` environment variables, which default to 115 and 100 respectively.
+
 ## Perform logging in your project
 
 To perform logging in your own code (e.g. in a node), you are advised to do as follows:


docs/source/logging/logging.md
```diff
@@ -64,6 +64,10 @@ export COLUMNS=120 LINES=25
 You must provide a value for both `COLUMNS` and `LINES` even if you only wish to change the width of the log message. Rich's default values for these variables are `COLUMNS=80` and `LINE=25`.
 ```
 
+### Rich logging in Jupyter
+
+Rich also formats the logs in JupyterLab and Jupyter Notebook. The size of the output console does not adapt to your window but can be controlled through the `JUPYTER_COLUMNS` and `JUPYTER_LINES` environment variables, which default to 115 and 100 respectively.
```
MerelTheisenQB, line 69: So what's the "ideal" setting for these environment variables is it 115 and 100 or something else?

AntonyMilneQB: 115 and 100 seemed to be ideal, so I don't think many people will want to change them. That's what's shown in the screenshot.

MerelTheisenQB: Okay, so just to clarify do they need to do anything when they want to use Rich in Jupyter? It's not super clear to me now if they need to set those environment variables or not.

AntonyMilneQB: No, they don't need to do anything. They only need to change the environment variables if they don't like the default width that seemed to work well for me. Let me make this clearer!

AntonyMilneQB:
```suggestion
Rich also formats the logs in JupyterLab and Jupyter Notebook. The size of the output console does not adapt to your window but can be controlled through the `JUPYTER_COLUMNS` and `JUPYTER_LINES` environment variables. The default values (115 and 100 respectively) should be suitable for most users, but if you require a different output console size then you should alter the values of `JUPYTER_COLUMNS` and `JUPYTER_LINES`.
```

AntonyMilneQB: @MerelTheisenQB how's that suggestion?

MerelTheisenQB: Great, that's very clear to me now 🙂

SajidAlamQB (approved):
This is awesome! 🌟

MerelTheisenQB (approved):
Nice one! And great work on getting a PR merged in Rich 😄 💯

PR #1639, opened 2022-06-21
Fix WINDOW CI PYTHON 3.10 fail pip-compile

noklam:

## Description

Fixed fail CI due to pip-compile at https://github.com/kedro-org/kedro/pull/1630

PEP517 and PEP518 are relatively new specifications related to the Python build system. It causes some issue with some of the libraries that we depends on ob.
## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1639"></a>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,8 @@
+# PEP-518 https://peps.python.org/pep-0518/
+[build-system]
+# Minimum requirements for the build system to execute.
+requires = ["setuptools>=38.0", "wheel"]  # PEP 518 specifications.
+
 [tool.black]
 exclude = "/templates/|^features/steps/test_starter"
 


noklam:
@MerelTheisenQB Not really sure about this, it could be caused by the dependencies, so maybe some of them didn't support Window Python 3.10 well.

SajidAlamQB (approved):
Awesome! 🎉

deepyaman:
@noklam More because I don't understand PEP 517/518 particularly well, but I don't understand how this resolves the issue. If you look at the `pip install` logs for this build, you'll see that `import-linter` is being built based on `pyproject.toml` rather than `PEP517`--could this be something that is inconsistent because `import-linter` doesn't specify their build system?

I don't quite understand how changing the build system for Kedro here made a difference.

deepyaman:
As another data point, https://app.circleci.com/pipelines/github/kedro-org/kedro/9308/workflows/f57743ea-ed24-4075-b0e3-a50c062d547a/jobs/142356 has it using PEP517 but doesn't fail.

noklam:
@deepyaman Could you try re-trigger your PR to see if it will pass? To be honest, I don't understand the issue fully as well, I have noticed my PR fail with the pip-compile test occasionally and it's more stable when I have the `pyproject.toml`.

I think the problem PEP517 and PEP518 are trying to solve is the `chicken or egg` problem - What versions of `setuptools` and `wheel` does one need to need to read the `setup.py`? Currently there is a circular dependencies, and I have no idea how `pip` or `setuptools` are handling this, it's highly likely depends on the version too.

deepyaman:
> @deepyaman Could you try re-trigger your PR to see if it will pass? To be honest, I don't understand the issue fully as well, I have noticed my PR fail with the pip-compile test occasionally and it's more stable when I have the `pyproject.toml`.

https://app.circleci.com/pipelines/github/kedro-org/kedro/9592/workflows/ca4978bd-38cd-47ec-a1d8-b9adf3eee257/jobs/146155 is re-running now, and it looks like it built the `import-linter` (and `memory-profiler`) wheels without issue.

AntonyMilneQB:
@deepyaman That is a radical thought! It does indeed seem that the pip-compile jobs are there largely for historical reasons, but I have found the pip compile jobs quite useful for immediately flagging where there's version conflicts and the like. I guess this would still be flagged by the `pip install -r test_requirements.txt` stage of unit tests though.

Two things to mention:
1. We should think about our general policy on upper bounding and whether we should use dependabot
2. Handling dependencies should be _much_ easier when we move kedro dataset dependency out of this repo and into the new kedro-datasets one

deepyaman:
> I guess this would still be flagged by the `pip install -r test_requirements.txt` stage of unit tests though.

@AntonyMilneQB this was my understanding!

PR #1642, opened 2022-06-22
Fix/the `getpass.getuser()` call fails when no user is present

### LINKED ISSUE #1624: KedroSession: getpass.getuser() call fails
## Description
First of all, I want to thank you for this splendid work that is Kedro. I believe it makes our life easier and I am glad that we have adopted it at my company.
This is the first time that I post an issue here, so I hope to frame this correctly!

In my company, we are using Kedro pipelines orchestrated in Airflow (we don't use kedro-airflow but we have a custom setup).
I have recently upgraded one pipeline to Kedro 0.18, but it's currently failing here:

```
# Line 167 of kedro/framework/session/session.py
session_data["username"] = getpass.getuser()
```

with the following error:
``KeyError: 'getpwuid(): uid not found: 1000``

I believe this is due to this process running with an uid that is not defined in /etc/passwd.
I am not super familiar with this things, but googling, it seems this is not an uncommon situation when using existing images.

Maybe it might be better to wrap that line in a try/except clause and catch ImportErrors?
This username tag does not seem to serve a critical purpose so I think the code should not crash because of this.

Right now I have solved it temporarily by setting an environment variable (USERNAME). What do you think?

## Steps to Reproduce
1. Run Kedro pipeline in a containerized setting with a UID that does not exist in /etc/passwd

## Expected Result
I expect that the code runs fine, and maybe the KedroSession is initialized with a "none" user or something like that

## Actual Result
The code crashes with a KeyError when setting username in KedroSession

## Your Environment
Our custom Docker Image is ultimately based on python:3.8-buster:

* Kedro version used (`pip show kedro` or `kedro -V`): 0.18.1
* Python version used (`python -V`): 3.8
* Operating system and version: Code running in Airflow on a custom image based on python:3.8-buster

tynandebold:
## Description

Fixes #1624.

## Development notes

We now wrap the call in a `try-except` block to prevent the failure.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1642"></a>

### DIFF
--- a/kedro/framework/session/session.py
+++ b/kedro/framework/session/session.py
@@ -165,7 +165,7 @@ class KedroSession:
 
         try:
             session_data["username"] = getpass.getuser()
-        except:  # pylint: disable=broad-except
+        except:  # pylint: disable=bare-except
             logging.getLogger(__name__).debug("Unable to get username.")
 
         session._store.update(session_data)
--- a/kedro/framework/session/session.py
+++ b/kedro/framework/session/session.py
@@ -163,7 +163,10 @@ class KedroSession:
         if extra_params:
             session_data["extra_params"] = extra_params
 
-        session_data["username"] = getpass.getuser()
+        try:
+            session_data["username"] = getpass.getuser()
+        except:  # pylint: disable=broad-except
+            logging.getLogger(__name__).debug("Unable to get username.")
 
         session._store.update(session_data)
 
--- a/kedro/framework/session/session.py
+++ b/kedro/framework/session/session.py
@@ -165,8 +165,10 @@ class KedroSession:
 
         try:
             session_data["username"] = getpass.getuser()
-        except:  # pylint: disable=bare-except
-            logging.getLogger(__name__).debug("Unable to get username.")
+        except Exception as exc:  # pylint: disable=broad-except
+            logging.getLogger(__name__).debug(
+                "Unable to get username. Full exception: %s", exc
+            )
 
         session._store.update(session_data)
 
--- a/tests/framework/session/test_session.py
+++ b/tests/framework/session/test_session.py
@@ -496,6 +496,27 @@ class TestKedroSession:
         ]
         assert actual_log_messages == expected_log_messages
 
+    def test_get_username_error(
+        self, fake_project, mock_package_name, mocker, caplog
+    ):
+        """Test that username information is not added to the session store
+        if call to getuser() fails
+        """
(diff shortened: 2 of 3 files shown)


AntonyMilneQB (approved):
Congratulations on your first (?) kedro core PR!

The only test failures you need to worry about are the linter fails for `kedro.framework.session.session` (forget about the other lint fails). Hopefully my suggestion will fix that.

kedro/framework/session/session.py
```diff
@@ -163,7 +163,10 @@ def create(  # pylint: disable=too-many-arguments
         if extra_params:
             session_data["extra_params"] = extra_params
 
-        session_data["username"] = getpass.getuser()
+        try:
+            session_data["username"] = getpass.getuser()
+        except:  # pylint: disable=bare-except
+            logging.getLogger(__name__).debug("Unable to get username.")
```
AntonyMilneQB, line 169:
```suggestion
        except Exception as exc:  # pylint: disable=broad-except
            logging.getLogger(__name__).debug(
                "Unable to get username. Full exception: %s",
                exc
            )
```

tests/framework/session/test_session.py
```diff
@@ -496,6 +496,27 @@ def test_git_describe_error(
...
+    def test_get_username_error(
+        self, fake_project, mock_package_name, mocker, caplog
+    ):
+        """Test that username information is not added to the session store
+        if call to getuser() fails
+        """
+        caplog.set_level(logging.DEBUG, logger="kedro")
+
+        mocker.patch("kedro.framework.session.KedroSession._setup_logging")
+        mocker.patch("getpass.getuser", side_effect=FakeException)
```
AntonyMilneQB, line 508:
```suggestion
        mocker.patch("getpass.getuser", side_effect=FakeException("getuser error"))
```

AntonyMilneQB:
FYI I've (hopefully) fixed the linting and also updated the debug message to include the actual exception.

tynandebold:
> FYI I've (hopefully) fixed the linting and also updated the debug message to include the actual exception.

Thanks for that! Some of the tests are still failing. Doesn't seem to be caused by this PR any longer, though I'm not 100% sure.

noklam:
@tynandebold try re-trigger the failed jobs, there are some unstable tests especially for Window fails occasionally.

MerelTheisenQB (approved):
LGTM! 👍 Thanks for picking this up @tynandebold 😄

Issue #1361, opened 2022-03-23
[KED-3025] Simplify kedro new workflow into one question


AntonyMilneQB:
_Transferred from Jira ticket by @datajoely_

Both internal and external survey say that they don't need to fine control of 3 separate questions
![image](https://user-images.githubusercontent.com/49395058/159673318-50e3ab93-b82e-4e35-b23e-80d3c7eea26c.png)
![image](https://user-images.githubusercontent.com/49395058/159673354-df49011f-199f-46e1-812a-782a86f23d9c.png)

AntonyMilneQB:
My opinion here is very much in favour of getting rid of the 2nd and 3rd questions, for exactly the same reason as @WaylonWalker gives here - having 3 different things just creates more confusion. When I first did `kedro new` I didn't really know what the difference between the 3 different things were and it just made using kedro feel more complicated.

![image](https://user-images.githubusercontent.com/49395058/159674015-c491a015-cb9e-4946-9226-5dd8564964c3.png)

datajoely:
So the only thing that's worth aligning on is the pesky `-` vs `_` debate

According to [PEP8](https://peps.python.org/pep-0008/#package-and-module-names) 'use of underscores is discouraged' this is why we `pip install kedro-viz` but if you wanted to use the library you write `from kedro_viz import api` .

Possible solutions:

1) In theory most users wouldn't care/know if we aggressively made everything underscores since its syntactically valid in every case, even if it breaks convention.
2) We dynamically coerce the name of the package to use dashes and module to use underscores. This would follow convention, but I'm unsure if it will introduce unintended side effects.

AntonyMilneQB:
Just a couple of clarifications to add to this - I spent too long reading about package names a while ago and it is pretty confusing so let me share the benefits of what I learnt here:
* strictly speaking, the name used in `pip install` is the _distribution_ name rather than the package name, and PEP8 doesn't say anything about the use of underscores in this (see https://stackoverflow.com/questions/54597212/using-hyphen-dash-in-python-repository-name-and-package-name)
* the PEP8 convention is speaking about package name as in `import ...`. Here underscores are discouraged, but given that dashes are impossible, underscores are realistically the only option if you have a package name with multiple words (or we could just take the first word in the project name, or we could take the first letter of every word in the project name... but probably just makes things more confusing)
* the importable package name doesn't even need to be the same as the distribution name (as in `pip install scikit-learn` vs. `import sklearn` example)
[...]

PR #1643, opened 2022-06-22
Simplify `kedro new` flow

merelcht:
## Description
https://github.com/kedro-org/kedro/issues/1361

## Development notes
- Removed the prompts for `repo_name` and `python_package`. These variables will be generated automatically from the `project_name` provided by the user. 
- Added validation for the `project_name` input and changed it to only accept alphanumeric symbols, hyphens and underscores and check it's at least 2 characters long. This is to prevent the user from adding in special characters that would then also be used for the `repo_name` and `python_package`. See comments on this below. 
- Users can still provide all variables in a configuration file if they want custom repository and python package names. 
- Added tests to check that users can provide custom prompts as well and fixed some bugs that at first failed those tests. 

The new flow looks like:

Screen Recording 2022-06-23 at 14 30 10

**Left to do:**
Need to update the starters as well and remove the `repo_name` and `python_package` prompts there. 
## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1643"></a>

### DIFF
--- a/docs/source/contribution/developer_contributor_guidelines.md
+++ b/docs/source/contribution/developer_contributor_guidelines.md
@@ -124,7 +124,7 @@ There are two special considerations when contributing a dataset:
 
 ## Create a pull request
 
-Create your pull request with a descriptive title. Before you submit it, consider the following:
+Create your pull request with [a descriptive title](#pull-request-title-conventions). Before you submit it, consider the following:
 
 * You should aim for cross-platform compatibility on Windows, macOS and Linux
 * We use [SemVer](https://semver.org/) for versioning
@@ -150,6 +150,17 @@ If Spark/PySpark/Hive tests for datasets are failing it might be due to the lack
 ```{note}
 We place [conftest.py](https://docs.pytest.org/en/latest/reference/fixtures.html) files in some test directories to make fixtures reusable by any tests in that directory. If you need to see which test fixtures are available and where they come from, you can issue the following command `pytest --fixtures path/to/the/test/location.py`.
 ```
+
+### Pull request title conventions
+
+The Kedro repository requires that you [squash and merge your pull request commits](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/about-pull-request-merges#squash-and-merge-your-pull-request-commits), and, in most cases, the [merge message for a squash merge](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/about-pull-request-merges#merge-message-for-a-squash-merge) then defaults to the pull request title.
+
+For clarity, your pull request title should be descriptive, and we ask you to follow some guidelines suggested by [Chris Beams](https://github.com/cbeams) in his post [How to Write a Git Commit Message](https://chris.beams.io/posts/git-commit/#seven-rules). In particular, for your pull request title, we suggest that you:
+
+* [Limit the length to 50 characters](https://chris.beams.io/posts/git-commit/#limit-50)
+* [Capitalise the first letter of the first word](https://chris.beams.io/posts/git-commit/#capitalize)
+* [Omit the period at the end](https://chris.beams.io/posts/git-commit/#end)
+* [Use the imperative tense](https://chris.beams.io/posts/git-commit/#imperative)
 
 ### Hints on `pre-commit` usage
 [`pre-commit`](https://pre-commit.com) hooks run checks automatically on all the changed files on each commit but can be skipped with the `--no-verify` or `-n` flag:
--- a/docs/source/get_started/standalone_use_of_datacatalog.md
+++ b/docs/source/get_started/standalone_use_of_datacatalog.md
@@ -2,9 +2,9 @@
 
 ## Introduction
 
-To make it easier to share a Jupyter notebook with others you need to avoid hard-coded file paths used to load or save data. One way to explore data within a shareable Jupyter notebook is take advantage of Kedro's [`DataCatalog`](../data/data_catalog.md), but in the early phases of a project, you may not want to use any other Kedro features.
+To make it easier to share a Jupyter notebook with others, avoid the use of hard-coded file paths to load or save data. One way to explore data within a shareable Jupyter notebook is to use Kedro's [`DataCatalog`](../data/data_catalog.md), but in the early phases of a project, you might not want to use any other Kedro features.
 
-The Kedro starter with alias `standalone-datacatalog` (formerly known as `mini-kedro`) provides this minimal functionality. You can specify the sources required to load and save data using a YAML API. For example:
+The Kedro starter with alias `standalone-datacatalog` (formerly known as `mini-kedro`) provides this minimal functionality. You can use a YAML API to specify the sources required to load and save data, for example,
 
  ```yaml
 # conf/base/catalog.yml
@@ -30,7 +30,7 @@ df_2 = catalog.save("example_dataset_2")
 
 ## Usage
 
-Create a new project using the [`standalone-datacatalog` starter](https://github.com/kedro-org/kedro-starters/tree/main/standalone-datacatalog):
+Use the [`standalone-datacatalog` starter](https://github.com/kedro-org/kedro-starters/tree/main/standalone-datacatalog) to create a new project:
 
 ```bash
 $ kedro new --starter=standalone-datacatalog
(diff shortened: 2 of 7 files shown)


tests/framework/cli/test_starters.py
```diff
@@ -130,61 +118,68 @@ def test_custom_project_name_with_hyphen(self, fake_kedro_cli):
...
+        custom_input = "\n".join(["My Project", "my_custom_repo"])
         result = CliRunner().invoke(
             fake_kedro_cli,
-            ["new"],
-            input=_make_cli_prompt_input(
-                project_name="My Project",
-                repo_name="my-repo",
-                python_package="my_package",
-            ),
+            ["new", "--starter", "template"],
```
noklam, line 173: is the `--starter` argument necessary? The old test is just generating project without any nodes.

noklam (commented):
Looks great! I like this change. I found one small bug here. The folder is actually correct but the path in the console log is incorrect.

image

noklam:
I think this is a bug because the validation logic is actually tightly coupled with the user prompts. See the snippets below

```python
    for variable_name, prompt_dict in prompts.items():
        prompt = _Prompt(**prompt_dict)

        # render the variable on the command line
        cookiecutter_variable = render_variable(
            env=StrictEnvironment(context=cookiecutter_context),
            raw=cookiecutter_context.get(variable_name),
            cookiecutter_dict=config,
        )

        # read the user's input for the variable
        user_input = read_user_variable(str(prompt), cookiecutter_variable)
        if user_input:
            prompt.validate(user_input)
            config[variable_name] = user_input
```

image

tests/framework/cli/test_starters.py
```diff
@@ -200,33 +195,6 @@ def test_fail_if_dir_exists(self, fake_kedro_cli):
         assert result.exit_code != 0
         assert "directory already exists" in result.output
 
-    @pytest.mark.parametrize(
-        "repo_name", [".repo\nvalid", "re!po\nvalid", "-repo\nvalid", "repo-\nvalid"]
```
noklam, line 204: I think we still need these test cases, it would catch the bug that I found.

noklam:
That's strange, I am on this commit. Can you try something more extreme like `my!new$project` see if it works?

```
 Author: Merel Theisen <merel.theisen@quantumblack.com>
Date:   Thu Jun 23 14:26:10 2022 +0100
```

MerelTheisenQB:
> That's strange, I am on this commit. Can you try something more extreme like `my!new$project` see if it works?
> 
> ```
>  Author: Merel Theisen <merel.theisen@quantumblack.com>
> Date:   Thu Jun 23 14:26:10 2022 +0100
> ```

You're absolutely correct about the non-alphanumeric characters bug! It's a tricky issue because we're using cookiecutter to fill `repo_name` and `python_package`. And sadly we can't add any advanced regex into `cookiecutter.json`. [They have hooks to validate values](https://cookiecutter.readthedocs.io/en/2.1.1/advanced/hooks.html#example-validating-template-variables), but I'm not sure we want to use that. 

Another option is, which Joel alluded to (https://github.com/kedro-org/kedro/issues/1361#issuecomment-1160478878) is just not allowing users to provide a project name that contains special characters, e.g. !%$&..

tests/framework/cli/test_starters.py
MerelTheisenQB: I have created a new test `test_invalid_project_name` for this case

tests/framework/cli/test_starters.py
```diff
@@ -85,6 +85,42 @@ def test_starter_list(fake_kedro_cli):
...
+    fake_kedro_cli, entry_point, specs, expected
+):
+    """Check that `kedro starter list` prints out the plugin starters."""
+    entry_point.load.return_value = specs
+    entry_point.module = "invalid_starter"
+    result = CliRunner().invoke(fake_kedro_cli, ["starter", "list"])
+    assert result.exit_code == 0, result.output
+    assert expected in result.output
+
+
```
noklam, line 123: do you intend to test the user prompt?

MerelTheisenQB: The tests below in `TestNewFromUserPromptsValid` and `TestNewFromUserPromptsInvalid` check the user prompts, or did you mean something else?

noklam: @MerelTheisenQB  Sorry forget about it.  I was looking [here](https://github.com/kedro-org/kedro/pull/1643/files/0620c5ba06fe17bccdfa8e34a91fe9e6b030cc0e..8e85558b1187c7a85fdec479584758ae63e668e1#) and saw changes like `test_starter_list_with_starter_plugin`. I think I am just confused by the merging changes. It's not the changes made in this PR.

noklam:
One minor comment, when user input an invalid prompt. It will throw an error `ValueError: new project!!!!` at the bottom,  but a more informative message is printed above the stack trace. Maybe they can all be combined in the `ValueError`? I don't know if we did this before for the coloring, but with the current `Rich` coloring I think we don't need both.
```
'new project!!!!' is an invalid value.
It must contain only alphanumeric symbols and/or underscores and/or hyphens.
```


```python
    def validate(self, user_input: str) -> None:
        """Validate a given prompt value against the regex validator"""
        if self.regexp and not re.match(self.regexp, user_input):
            click.secho(f"'{user_input}' is an invalid value.", fg="red", err=True)
            click.secho(self.error_message, fg="red", err=True)
            raise ValueError(user_input)
```

If you scroll up, you see this.
image

At the bottom, it shows this
image

noklam (approved):
Nice work! This makes it easier for the beginners.

MerelTheisenQB:
> One minor comment, when user input an invalid prompt. It will throw an error `ValueError: new project!!!!` at the bottom, but a more informative message is printed above the stack trace. Maybe they can all be combined in the `ValueError`? I don't know if we did this before for the coloring, but with the current `Rich` coloring I think we don't need both.
> 
> ```
> 'new project!!!!' is an invalid value.
> It must contain only alphanumeric symbols and/or underscores and/or hyphens.
> ```
> 
> ```python
>     def validate(self, user_input: str) -> None:
>         """Validate a given prompt value against the regex validator"""
>         if self.regexp and not re.match(self.regexp, user_input):
>             click.secho(f"'{user_input}' is an invalid value.", fg="red", err=True)
>             click.secho(self.error_message, fg="red", err=True)
>             raise ValueError(user_input)
> ```
> 
> If you scroll up, you see this. image
> 
> At the bottom, it shows this image

Ah good catch! The stacktrace doesn't really add anything and makes it quite a bit harder to find our own message. I'll change it.

SajidAlamQB (approved):
This looks great, so much more seamless! 🎉

AhdraMeraliQB (approved):
Great work on this! 🌟

AntonyMilneQB (approved):
Looks good! Just one small question.

Also I think it might be nice just to tidy `cookiecutter.json` up a bit. It is a constant source of minor irritation to me that it's not quite consistent 😅  Here's what we currently have:
```
    "repo_name": "{{ cookiecutter.project_name.replace(' ', '-').lower().strip('-') }}",
    "python_package": "{{ cookiecutter.project_name.replace(' ', '_').replace('-', '_').lower() }}",
```

and what I would suggest:
```
    "repo_name": "{{ cookiecutter.project_name.strip().replace(' ', '-').replace('_', '-').lower() }}",
    "python_package": "{{ cookiecutter.project_name.strip().replace(' ', '_').replace('-', '_').lower() }}",
```

kedro/framework/cli/starters.py
```diff
@@ -355,16 +355,28 @@ def _create_project(template_path: str, cookiecutter_args: Dict[str, str]):
         ) from exc
 
     _clean_pycache(Path(result_path))
+    extra_context = cookiecutter_args.get("extra_context", {})
+    project_name = extra_context.get("project_name", "")
+    python_package = extra_context.get(
+        "python_package", project_name.lower().replace(" ", "_").replace("-", "_")
+    )
```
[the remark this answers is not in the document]
MerelTheisenQB: Fair point! I guess I was just being overly cautious but it might be better if this throws an error if there's no `extra_context` because then something would be really wrong.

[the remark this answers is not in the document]
MerelTheisenQB: Oohh and actually I forgot about this, but we added in backward compatibility for when there's no `prompts.yml` file. Then you can still run `kedro new` and it will create a new project with "New Kedro Project" as project name which is the default in `cookiecutter.json`. So `project_name = extra_context["project_name"]` would break that and we should instead do extra_context.get("project_name", "New Kedro Project")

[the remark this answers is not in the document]
AntonyMilneQB: Thanks for checking this! I think I understand. We should definitely see if we can get rid of the prompts.yml file though, it is pretty confusing 😬

PR #1644, opened 2022-06-22
Put framework logging into LOGGING and fix import resetting logging

### LINKED ISSUE #1558: Get parallel runner to recognise framework-side logging config
The parallel runner does not seem to use the framework-side logging config. By default, if you do a `kedro run --runner=ParallelRunner` then the project-side logging config is picked up correctly and you will see the logs correctly in the console and in the logs/info.log file. However, if you delete conf/base/logging.yml then most messages will no longer be logged:
image

See [this explanation](https://github.com/kedro-org/kedro/issues/1461#issuecomment-1116955368) for how logging config is setup for `kedro run`. However, with the parallel runner there's something else that happens too. Assuming that conf/base/logging.yml does exist:
1. https://github.com/kedro-org/kedro/blob/676600c4b63eec53c13fc4e2536d0a990dac77ce/kedro/framework/session/session.py#L190
calls this function:
https://github.com/kedro-org/kedro/blob/62cc27a1441ad49761d38c44e425a77871b6ba74/kedro/framework/project/__init__.py#L208
This sets the global `LOGGING` to be the project-side logging config.
2. `LOGGING` is used in https://github.com/kedro-org/kedro/blob/main/kedro/runner/parallel_runner.py to `_bootstrap_subprocess` which means that all subprocesses have the project-side logging config.

This seems to work (although it's not very nice to have the global variable). However, when there's no conf/base/logging.yml, `configure_logging` isn't called, so `LOGGING` will be left as `None`, so `_bootstrap_subprocess` won't set the logging configuration. This is fine because there's no project-side logging.yml to use _BUT_ the problem is that the framework-side logging doesn't make it into the subprocesses, so no log messages get shown at all.

What we need to do is:
* make sure that the framework-side logging config makes it to the subprocesses. e.g. maybe the default `LOGGING` should not be `None` but instead what comes out of `kedro.config.default_logger`
* ideally figure out a way to do all this without needing the global `LOGGING` so we can remove that

AntonyMilneQB:
## Description
Fixes #1558. Fixes Windows CI problem in #1592.

Also changed `conf_logging` to `logging_config` in a few places for consistency.

## Development notes
This has been quite a journey, but after many, many different attempts at this I think I've now got a good solution 😬

Before:
* framework-side logging was not being passed to the parallel runner in certain cases (when subprocess is spawned, which means on Windows (all Python versions) and macOS (Python 3.8+) - see #1558 for details)
* a very mysterious bug was breaking Windows e2e tests involving parallel runner and a plugin in #1592
* `LOGGING` global variable was used to pass logging to spawned subprocesses

Now:
* `LOGGING` global variable still exists (there's no good way to get rid of it) but is now set to load framework default logging config initially and then gets overwritten with project logging config. The mechanism for this is a new `_ProjectLogging` class, which makes logging configuration much closer to how we configure other things in the same file
* we no longer rely on the side-effect of importing `kedro.config.default_logger` to load setup framework logging config
* in fact, we now no longer need this file. What it's doing fits in better in `project.__init__.py` anyway, so I've removed it
* this is a _very slight_ breaking change, because there's a small chance someone currently does `import kedro.config.default_logger`, which will no longer work. I checked on #kedro-users and discord and no one is knowingly doing this. The only case I can think of someone doing this would be from copy and pasting our current ipython extension script into a script of their own. This itself is quite unlikely, but if it is the case then it's comparable to very small breaking changes to non-essential workflows that we've put into minor releases before, so both Ivan and I are ok with this

## Cause of Windows e2e test failures in #1592
This was very subtle and mysterious so let me document it here. The error was nothing to do with rich but actually a result of an import in a plugin!
* `test_plugin` did `from kedro.framework.cli.starters import ...`
* by Python's import mechanism, this imports in turn `kedro.framework.cli` ➡️ `kedro.framework.main` (due to `__init__.py` import) ➡️ `import kedro.config.default_logger`
* this reset the logging mechanism partway through a spawned subprocess, which messed up the logging
Now that there's no logging side-effects on file import this is not a problem.

## Testing

Tested manually all the following cases:
* `kedro run` with and without project logging.yml, with and without plugin with imports
* `kedro run --runner=ParallelRunner` with and without project logging.yml, with and without plugin with imports
* the above with a packaged project
* `kedro ipython` and `kedro jupyter lab`
* throwing exception inside a node

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1644"></a>


kedro/framework/cli/cli.py
```diff
@@ -13,9 +12,8 @@
 import click
 import importlib_metadata
 
-# pylint: disable=unused-import
-import kedro.config.default_logger  # noqa
 from kedro import __version__ as version
+from kedro.framework.project import LOGGING_CONFIG
```
AntonyMilneQB, line 16:
```suggestion
# pylint: disable=unused-import
from kedro.framework.project import LOGGING_CONFIG  # noqa
```

kedro/framework/project/__init__.py
```diff
@@ -179,8 +187,26 @@ def configure(self, pipelines_module: Optional[str] = None) -> None:
...
+        logging.captureWarnings(True)
+
+    def configure(self, logging_config: Dict[str, Any]) -> None:
+        logging.config.dictConfig(logging_config)
+        self.data = logging_config
+
+
 PACKAGE_NAME = None
-LOGGING = None
+LOGGING = _ProjectLogging()
```
AntonyMilneQB, line 209: The major success here is that this now acts like `pipelines` and `settings`, and `configure_logging` is much closer to `configure_project`.

noklam (approved):
Awesome! Amazing work that fixes the very subtle bug and I like the more consistent `_ProjectLogging`

SajidAlamQB (approved):
Awesome work, this looks perfect thank you! 🌟

kedro/framework/cli/micropkg.py
```diff
@@ -62,7 +62,7 @@ def micropkg_cli():  # pragma: no cover
 
 
 @micropkg_cli.group()
-def micropkg():
+def anmicropkg():
```
AntonyMilneQB, line 65:
```suggestion
def micropkg():
```

docs/conf.py
```diff
@@ -206,6 +206,7 @@
     # "anchor not found" but it's a valid selector for code examples
     "https://docs.delta.io/latest/delta-update.html#language-python",
     "https://github.com/kedro-org/kedro/blob/main/kedro_technical_charter.pdf",
+    "https://github.com/kedro-org/kedro/blob/main/kedro/framework/project/config/default_logging.yml",
```
AntonyMilneQB, line 209:
```suggestion
    "https://github.com/kedro-org/kedro/blob/main/kedro/framework/project/default_logging.yml",
```

docs/conf.py
```diff
@@ -206,6 +206,7 @@
     # "anchor not found" but it's a valid selector for code examples
     "https://docs.delta.io/latest/delta-update.html#language-python",
     "https://github.com/kedro-org/kedro/blob/main/kedro_technical_charter.pdf",
```
AntonyMilneQB, line 208:
```suggestion
```

PR #1584, opened 2022-06-01
Replace backticks with single quotes

AhdraMeraliQB:
## Description
This is part one in #1514 ; Putting these changes in a separate PR so that it doesn't bloat the actual implementation,

## Development notes
Replaced backticks with single quotes in log messages (and error messages, for consistency - previously both were used interchangeably). Comments and docstrings remain unchanged.

In some instances triple quotes were introduced due to nesting of single quotes; instances like `" ' [bt] [bt] ' "` become `""" " ' ' " """`

## Checklist

- [ ] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [ ] Added tests to cover my changes

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1584"></a>

Issue #1514, opened 2022-05-06
Change kedro log messages to be more rich


AntonyMilneQB:
**🥳  FUN TICKET 🥳 Prerequisites: #1512. Based off point 3 in https://github.com/kedro-org/kedro/issues/1464**

Let's take advantage of the nice formatting we can do with rich and make Kedro's log messages prettier.
* replace backticks (`) in all log messages with quote marks (can be single or double). This will be _a lot_ of very small changes!
* edit log messages to be clearer and prettier using [console markup](https://rich.readthedocs.io/en/stable/markup.html). Note:
  - let's not go too crazy here. Try and restrict to sensible styles so that we have a somewhat unified appearance that we know works on a variety of background colours. Take inspiration from the styles in the picture below (e.g. `[bold bright_green]`  is used but won't appear much by default so it's a good candidate to use)
  - any log messages you add markup to will need to have `extra={"markup": True}` in the logging call as explained [in the rich docs](https://rich.readthedocs.io/en/stable/logging.html). If we end up adding markup to loads of logging calls it might be easier to alter the kedro logger (not the root one!!) in framework logging.yml to set `markup: true`
  - start with the logging messages that are most important/frequently seen during a kedro run. Check point 3 in https://github.com/kedro-org/kedro/issues/1464 and the [associated code](https://github.com/datajoely/kedro-rich/blob/bc0313f0bd909a52abbcd730a4995d211082dded/kedro_rich/utilities/kedro_override_utils.py#L16-L133) by @datajoely for what the "highest impact" logging calls are and some suggested ways to style them. But feel free to add markup to any other logging calls throughout kedro too - the sky's the limit here! 🚀 
  - similarly feel free to go beyond just adding markup and change the wording/style of brackets/arrows/etc. in the log messages if you think they can be improved, as @datajoely did.

## Useful context: how rich formatting works
We are using the `RichHandler` for logging. By default this highlights messages automatically according to the `ReprHighlighter`. Here's how different things get highlighted (generated using `python -m rich.default_styles`):
<img width="596" alt="image" src="https://user-images.githubusercontent.com/49395058/167221570-2e5b6174-c322-4222-9e1f-15b512f4de60.png">

Here's an example of what this means in practice:
<img width="619" alt="image" src="https://user-images.githubusercontent.com/49395058/167221867-a9006e17-cde2-4fb6-abc5-bdc0ec9618a3.png">

[...]

PR #1663, opened 2022-07-01
Further richify logs

AhdraMeraliQB:
## Description

Following on from #1584 this PR addresses issue #1514 
## Development notes

[The first part of this issue](#1584) changed the back-ticks that highlighted sections of the log messages with single quotes so that they would be identifiable to the RichHandler that we now use for logging. They were then highlighted in bright green. An example kedro run can be seen below:

Screenshot 2022-07-01 at 17 28 42

Within the logs we see this highlighting being used for nodes, datasets, paths, classes, keys and more. Of these, I believe datasets and nodes to be the most pertinent to differentiate as they relate directly to the pipeline and its elements. These changes are introduced by this PR. An example kedro run now yields the following logs:

Screenshot 2022-07-01 at 17 29 09

Here we can see the results of the changes applied to the dataset style. Below, we explore all the changes made.

### Node Styling

Nodes in the log messages are not highlighted by quotes consistently. The log message seen in the screenshot above (defined in node.py, line 327) does not make use of the single quotes, and is automatically given the `bold magenta` style by rich. To remain consistent, the quotation marks have been stripped from other nodes in log messages.

### Dataset Styling

Datasets, however, were consistently highlighted by single quotes in log messages. To differentiate their styling from other similarly highlighted elements, their colour has been set to `dark_orange`. This colour was selected for several reasons:

1. This colour is easily readable on a variety of terminal backgrounds. Whilst rich will change the colour of the text from black to white depending on the terminal, it is not yet possible to designate different colour tags for different colour backgrounds, making it important that any colour selected works on a variety of backgrounds (most of all, black and white). We can see an example kedro run on different terminal backgrounds below:

image (6)

2. Like seen in the example run, datasets often appear in logs with nodes, which are denoted in magenta. By using dark orange, the two elements are easily differentiable, even when considering various forms of colour blindness. We can see how the logs from the example kedro run would look to individuals with different types of colour blindness:

image (7)

<sub>(Top Far Left: Green-Blind Deuteranopia, Bottom Far Left: Green-Weak Deuteranomaly, Top Middle Left: Blue-Blind Tritanopia, Bottom Middle Left: Blue-Weak Tritanomaly, Top Middle Right: Red-Blind Protanopia, Bottom Middle Right: Red-Weak Protanomaly, Top Far Right: Monochromacy/Achromatopsia, Bottom Far Right: Blue Cone Monochromacy)

These images were generated using [this colour blindness simulator](https://pilestone.com/pages/color-blindness-simulator-1#)</sub>

3. This colour is very distinct from the bright green that is used by default as the highlighting applied to quotation marks, serving the motivation of distinguishing the elements of the pipeline from the other aspects highlighted in log messages.

Screenshot 2022-07-02 at 19 39 18
<sub>See: The version numbers in the warnings are now styled differently from the datasets in the logs produced by the run</sub>

### Path Styling

Paths are not only printed to the terminal in log messages, and as such rich applies a specific style - `not italic magenta` - to them by default. In this PR that style is also applied to the paths seen in log messages.

### Dumb vs Rich terminal

Finally, to ensure rich stripped the styling tags when a dumb terminal was used, the example kedro project was run on a dumb terminal: 

dumb vs rich

Rich successfully strips the tags, ensuring that the styled logs do not display weirdly.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1663"></a>


tests/io/test_data_catalog.py
```diff
@@ -309,12 +309,13 @@ def test_mutating_datasets_not_allowed(self, data_catalog_from_config):
...
     def test_confirm(self, mocker, caplog):
         """Confirm the dataset"""
+        caplog.set_level(logging.INFO)
         mock_ds = mocker.Mock()
         data_catalog = DataCatalog(data_sets={"mocked": mock_ds})
         data_catalog.confirm("mocked")
         mock_ds.confirm.assert_called_once_with()
         assert caplog.record_tuples == [
-            ("kedro.io.data_catalog", logging.INFO, "Confirming DataSet 'mocked'")
+            ("kedro.io.data_catalog", logging.INFO, "Confirming DataSet [dark_orange]mocked[/dark_orange]")
```
AhdraMeraliQB, line 318: I don't think this is the best way to modify this test. When run, the captured stdout call reads without the rich style tags, but they are still present in the captured log call:

> tests\io\test_data_catalog.py:316: AssertionError (this error is from before the most recent commit)
---------------------------- Captured stdout call -----------------------------
[07/02/22 19:17:52] INFO     Confirming DataSet mocked      data_catalog.py:600
------------------------------ Captured log call ------------------------------
INFO     kedro.io.data_catalog:data_catalog.py:600 Confirming DataSet [dark_orange]mocked[/dark_orange]

Ideally the test should be modified to compare the captured stdout call as opposed to the captured log call to avoid specifying the style tags in the test. Potentially [capsys](https://doc.pytest.org/en/latest/how-to/capture-stdout-stderr.html#accessing-captured-output-from-a-test-function) could be used to do. The following replacement was attempted but would only yield an empty string:

```
def test_confirm(self, mocker, capsys):
        """Confirm the dataset"""
        mock_ds = mocker.Mock()
        data_catalog = DataCatalog(data_sets={"mocked": mock_ds})
        data_catalog.confirm("mocked")
        mock_ds.confirm.assert_called_once_with()
        assert capsys.readouterr().out == [
            ("kedro.io.data_catalog", logging.INFO, "Confirming DataSet 'mocked'")
        ]
```
The test would fail with the following output:
> E       assert '' == [('kedro.io.d...et 'mocked'")]
E         +''
E         -[('kedro.io.data_catalog', 20, "Confirming DataSet 'mocked'")]

Other ideas/approaches are very welcome 😅

kedro/config/common.py
```diff
@@ -82,10 +82,10 @@ def _get_config_from_patterns(
         if common_keys:
             sorted_keys = ", ".join(sorted(common_keys))
             msg = (
-                "Config from path '%s' will override the following "
-                "existing top-level config keys: %s"
+                "Config from path [not italic magenta]%s[/not italic magenta] will override the "
```
AntonyMilneQB, line 85:
```suggestion
                "Config from path [magenta]%s[/magenta] will override the "
```
I don't think you need to force it to be `not italic` here. Just comparing to the rich default styles, `repr.path` is not include `not italic` when others do.

tests/io/test_data_catalog.py
noklam: One quick question, does the test result depends on the terminal type (Dumb vs Emulated Terminal?)

AntonyMilneQB: I actually think is exactly the right way to modify the test. We need to think about what we're actually testing for here: this is a unit test to ensure the correct behaviour of the `confirm` method. This is covered by the `assert_called_once_with`, and the benefit of testing logging at all after that is debatable. But let's say we want to check that `confirm` also outputs the correct log message. Then what we're interested in is "are we making the right call to `logging.INFO`", not "how does rich render that call".

Hence by including the `dark_orange` styling in the test, you're testing (a bit like a mock) whether the call to `logging.INFO` is correct. We can trust rich will figure out to strip that styling in dumb terminals (as you tested manually), as that is a separate, well-tested, third party library. If we don't include `dark_orange` in the test then you would not be testing the actual call we're making. Put another way: the fact that this passes on CI and your local machine is actually a good demonstration that you're testing the right thing here, i.e. the logging call itself and not how it gets rendered.

I very much recommend[ this series of articles on TDD in Python](https://www.thedigitalcatonline.com/blog/2020/09/10/tdd-in-python-with-pytest-part-1/). We don't really follow it strictly on kedro, but it's definitely worth understanding. The case here would be one of an [outgoing command](https://www.thedigitalcatonline.com/blog/2020/09/15/tdd-in-python-with-pytest-part-3/#outgoing-queries-and-commands).

AntonyMilneQB (approved):
Awesome write-up! Thank you for your very careful, thorough work and testing on this ⭐ I just have one small suggestion.

At some point I think we need to figure out how our node `name`, `__str__` and `short_name` work better. I have a long-standing set of issues on this. And maybe we should add a `__rich__` method in the future also to automatically render it nicely. But that's not to worry about now :)

tests/io/test_data_catalog.py
noklam: image

I believe these styling tags should only have an effect on `stdout`, and shouldn't be logged in files.

noklam (commented):
Thank you! Very nice write-up and detail testing. I think this could be merged once we figured out how to removed the styling tags in logs.

kedro/config/common.py
AhdraMeraliQB: Thanks for the catch - I was following the convenition on  `repr.attrib_value` instead

tests/io/test_data_catalog.py
AhdraMeraliQB: > One quick question, does the test result depends on the terminal type (Dumb vs Emulated Terminal?)

Nope, it passes on both

AntonyMilneQB:
@noklam This is a very good point, well spotted! 🙏 Unfortunately there's not much we can do to fix that without adding quite a bit of complexity. I do have some ideas on how to improve that and the file-based logging in general which might work though; will make a separate issue for it.

I also realised a second problem with this that I didn't realise before: it's actually a slightly breaking change 😬 🤦 If I start my project with kedro 0.18.1, my project `logging.yml` uses the `console` handler which doesn't know how to deal with rich markup. Hence when I upgrade to kedro 0.18.2 without changing my project template, I will see the raw `[dark_orange]` etc. It's easy to fix this (just delete your project `logging.yml` file, since it's no longer required) but I think this is sufficiently annoying that we should not allow it in a non-breaking release. Sorry I didn't think of this before!!

So let's merge this to `develop` instead. This way:
* we can make these changes, since people moving from 0.18.x to 0.19.0 should be expected to make changes to their project template. If we want to rich-ify more messages these changes should also go to `develop` (but ones we have already merged to `main` are fine to stay there)
* we have time to figure out a solution to the file logging problem

noklam (approved):
I agree we can merge this into `develop` now and handle the styling tags problem with another ticket.

PR #1672, opened 2022-07-04
Enable rich in the REPL

AntonyMilneQB:
## Description
One final small piece to get some more benefits of rich: https://rich.readthedocs.io/en/stable/introduction.html#rich-in-the-repl.

This is a pretty minor change. In reality the only effect it will have is that when IPython/Juptyer shows something in the REPL (without explicitly writing `print`) it will highlight it nicely.

Before:
image

After (don't worry about text being cut off here; there's a scroll bar I didn't show):
image

## Future

We can make this look even nicer without much effort, e.g. to achieve indentation. This would be a Part 2 to integrating rich into kedro. I'll make tickets for it in due course.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1672"></a>


noklam (approved):
Nice! I like this.

PR #1685, opened 2022-07-04
Update kedro_io.md

CarolineMLynch:
## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1685"></a>


docs/source/data/kedro_io.md
```diff
@@ -388,16 +388,16 @@ def concat_partitions(partitioned_input: Dict[str, Callable[[], Any]]) -> pd.Dat
     return result
 ```
 
-As you can see from the example above, on load `PartitionedDataSet` _does not_ automatically load the data from the located partitions. Instead, `PartitionedDataSet` returns a dictionary with partition IDs as keys and the corresponding load functions as values. It allows the node that consumes the `PartitionedDataSet` to implement the logic that defines what partitions need to be loaded and how this data is going to be processed.
+As you can see from the above ecample, on load `PartitionedDataSet` _does not_ automatically load the data from the located partitions. Instead, `PartitionedDataSet` returns a dictionary with partition IDs as keys and the corresponding load functions as values. It allows the node that consumes the `PartitionedDataSet` to implement the logic that defines what partitions need to be loaded, and how this data is going to be processed.
```
MerelTheisenQB, line 391:
```suggestion
As you can see from the above example, on load `PartitionedDataSet` _does not_ automatically load the data from the located partitions. Instead, `PartitionedDataSet` returns a dictionary with partition IDs as keys and the corresponding load functions as values. It allows the node that consumes the `PartitionedDataSet` to implement the logic that defines what partitions need to be loaded, and how this data is going to be processed.
```

docs/source/data/kedro_io.md
```diff
@@ -443,10 +443,10 @@ def create_partitions() -> Dict[str, Any]:
 ```
 
 ```{note}
-Writing to an existing partition may result in its data being overwritten, if this case is not specifically handled by the underlying dataset implementation. You should implement your own checks to ensure that no existing data is lost when writing to a `PartitionedDataSet`. The simplest safety mechanism could be to use partition IDs that have a high chance of uniqueness: for example, the current timestamp.
+Writing to an existing partition may result in its data being overwritten, if this case is not specifically handled by the underlying dataset implementation. You should implement your own checks to ensure that no existing data is lost when writing to a `PartitionedDataSet`. The simplest safety mechanism could be to use partition IDs with a high chance of uniqueness: for example, the current timestamp.
 ```
 
-`PartitionedDataSet` also supports lazy saving, where the partition's data is not materialized until it's time to write.
+`PartitionedDataSet` also supports lazy saving, where the partition's data is not materialized until it is time to write.
```
MerelTheisenQB, line 449:
```suggestion
`PartitionedDataSet` also supports lazy saving, where the partition's data is not materialised until it is time to write.
```

MerelTheisenQB (approved):
👍 
Added two additional fixes for typos.

SajidAlamQB (approved):

docs/source/data/kedro_io.md
```diff
@@ -132,9 +132,9 @@ The last row in the example above would attempt to load a CSV file from `data/01
 The `DataCatalog` does not re-generate save versions between instantiations. Therefore, if you call `catalog.save('cars', some_data)` twice, then the second call will fail, since it tries to overwrite a versioned dataset using the same save version. To mitigate this, reload your data catalog by calling `%reload_kedro` line magic. This limitation does not apply to `load` operation.
 ```
 
-### Versioning using the Code API
+### Version using the Code API
```
MerelTheisenQB, line 135:
```suggestion
### Versioning using the Code API
```

PR #1706, opened 2022-07-13
Implement autodiscovery of project pipelines :mag:

### LINKED ISSUE #1664: Implement pipeline autodiscovery
Following https://github.com/kedro-org/kedro/issues/1284 and a discussion with Ivan, here's what we're going to do:

### 1. Define a function `find_pipelines`

In `kedro.framework.project.__init__.py`, something like this:
```
def find_pipelines():
    registered_pipelines = {}
    pipeline in Path(f"{PACKAGE_NAME}.pipelines").iterdir():
        importlib.import_module(pipeline)
        if hasattr(pipeline, "create_pipeline"):
            registered_pipelines[pipeline] = pipeline.create_pipeline()
    registered_pipelines["__default__"] = pipeline([])
    return registered_pipelines
```

Name is TBD (note I changed from previously-suggested `autoregister` because it doesn't actually register any more, just finds them). It's also like setup.py's `find_packages` here.

### 2. Call this function from project template and all starters

In pipeline_registry.py:
```

    """Register the project's pipelines.

    MAKE THIS DOCSTRING MORE INFORMATIVE

    Returns:
        A mapping from a pipeline name to a ``Pipeline`` object.

    """
    pipelines = find_pipelines()
    pipelines["__default__"] = sum(pipelines.values())
    return pipelines
```

### 3. Add documentation, tests, etc.

deepyaman:
## Description

Closes #1664 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1706"></a>


deepyaman:
@AntonyMilneQB Would love your opinion on something in particular. Currently, the Behave tests are failing, because one checks for a literal empty pipeline in text:

```
    And the pipeline should contain no nodes                          # features/steps/cli_steps.py:431
        File "C:\tools\miniconda3\envs\kedro_builder\lib\site-packages\behave\model.py", line 1329, in run
          match.run(runner.context)
        File "C:\tools\miniconda3\envs\kedro_builder\lib\site-packages\behave\matchers.py", line 98, in run
          self.func(context, *args, **kwargs)
        File "features\steps\cli_steps.py", line 442, in check_empty_pipeline_exists
          assert '"__default__": pipeline([])' in pipeline_file.read_text("utf-8")
      AssertionError
```

I think there's two options here:

1. Actually try running the pipeline, and get the expected error message associated with running a pipeline with no nodes.
2. Instead of testing that there's an empty pipeline defined, just look for the pipeline summing code in the template.

I'm inclined to think 1 is better, but want to make sure I'm understanding the purpose of this test properly.

---

For my reference and for anybody else who's interested, I've managed to write this in a way that only works(-ish?) on Python 3.10. In order to resolve this:

1. Add `__init__.py` file under `pipelines` directory in test fixture. `importlib.resources.contents` seems to pick up the subfolders without this only in Python 3.10+.
2. Adding the `__init__.py` as is fails, because `__pycache__` files get found and loaded. I think we need to use `files` instead of `contents`. However, since `files` is only introduced and recommended since Python 3.9, we need to install `importlib-resources>=1.3`.

Need to take care of some other things, but I'll get back to this on the weekend if I can.

kedro/framework/project/__init__.py
```diff
@@ -254,3 +256,53 @@ def validate_settings():
...
+            ``Pipeline`` object, or if the module import fails up front.
+    """
+    pipelines_dict = {"__default__": pipeline([])}
+    for pipeline_name in importlib.resources.contents(f"{PACKAGE_NAME}.pipelines"):
+        try:
+            pipeline_module = importlib.import_module(
+                f"{PACKAGE_NAME}.pipelines.{pipeline_name}"
+            )
+        except ModuleNotFoundError:
+            continue
```
AntonyMilneQB, line 281: When would this be raised and why don't we want to emit a warning in this case?

kedro/framework/project/__init__.py
```diff
@@ -254,3 +256,53 @@ def validate_settings():
...
+        except:  # pylint: disable=bare-except  # noqa: E722
+            warnings.warn(
+                f"An error occurred while importing the "
+                f"'{PACKAGE_NAME}.pipelines.{pipeline_name}' module. "
```
AntonyMilneQB, line 285: Is it intentional that this warning message uses `{PACKAGE_NAME}.pipelines.{pipeline_name}` while the others use  `{pipeline_module.__name__}`?

AntonyMilneQB: Related: is `pipeline_module.__name__` the same as `pipeline_name`?

kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/pipeline_registry.py
```diff
@@ -1,13 +1,20 @@
...
+from kedro.framework.project import find_pipelines
 from kedro.pipeline import Pipeline, pipeline
 
 
 def register_pipelines() -> Dict[str, Pipeline]:
     """Register the project's pipelines.
 
+    Since Kedro 0.18.3, projects can use the ``find_pipelines`` function
+    to autodiscover pipelines. However, projects that require more fine-
+    grained control can still construct the pipeline mapping without it.
```
AntonyMilneQB, line 13: I think this will probably need a bit more explanation (e.g. what structure is required for a pipeline to be auto-discovered, which is that created by `kedro pipeline create`). Or maybe just a link to the docs once it's in there.

tests/framework/project/test_pipeline_discovery.py
```diff
@@ -0,0 +1,165 @@
...
+    [(x, x) for x in [set(), {"my_pipeline"}]],
+    indirect=True,
+)
+def test_find_pipelines_skips_modules_that_cause_exceptions_upon_import(
+    mock_package_name_with_pipelines,  # pylint: disable=unused-argument
+    pipeline_names,
+):
+    # Create a module that will result in errors when we try to load it.
+    pipelines_dir = Path(sys.path[0]) / mock_package_name_with_pipelines / "pipelines"
+    pipeline_dir = pipelines_dir / "boulevard_of_broken_pipelines"
```
AntonyMilneQB, line 154: Beautiful name.

AntonyMilneQB (commented):
This looks perfect so far ⭐ I'll respond to your questions in a separate post.

kedro/framework/project/__init__.py
deepyaman: This `except` block occurs when the `pipeline_module` variable can't be set, because the import failed.

kedro/framework/project/__init__.py
```diff
@@ -254,3 +256,53 @@ def validate_settings():
...
+    Returns:
+        A generated mapping from pipeline names to ``Pipeline`` objects.
+
+    Warns:
+        UserWarning: When a module does not expose a ``create_pipeline``
+            function, the ``create_pipeline`` function does not return a
+            ``Pipeline`` object, or if the module import fails up front.
+    """
+    pipelines_dict = {"__default__": pipeline([])}
+    for pipeline_name in importlib.resources.contents(f"{PACKAGE_NAME}.pipelines"):
```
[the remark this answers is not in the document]
deepyaman: With the change to `files`,  it's actually an `iterdir`; I'll try to push that as soon as I've got it working a bit better!

AntonyMilneQB:
> I think there's two options here:
> 
> 1. Actually try running the pipeline, and get the expected error message associated with running a pipeline with no nodes.
> 2. Instead of testing that there's an empty pipeline defined, just look for the pipeline summing code in the template.
> 
> I'm inclined to think 1 is better, but want to make sure I'm understanding the purpose of this test properly.

Very good question. Normally I would agree with you and am also not a fan of explicitly trying to match the string in the pipeline_registry.py file. Here I'm not so sure though.

We already have tests in run.feature that do as you describe here:
```
  Scenario: Run default python entry point without example code
    Given I have prepared a config file
    And I have run a non-interactive kedro new without starter
    When I execute the kedro command "run"
    Then I should get an error exit code
    And I should get an error message including "Pipeline contains no nodes"
```

Hence I think the tests in `new_project.feature` should be interpreted as just checking the project template that was created, and so the string matching is correct. But you could reasonably argue that this is not necessary given the above `run.feature` tests. Note that the `test_plugin_starter` in that file is new (though again maybe we don't need to do the string matching).

The general pattern currently seems to be that every `kedro xyz` command has its own `xyz.feature` file that tests specifically the behaviour of that command. According to that, it makes sense to have `run.feature` tests that actually do the `kedro run` and look at the output _and_ `new_project.feature` tests that do `kedro new` and look at the template created.

But I don't mind changing how `new_project.feature` works or even removing those tests altogether if they don't seem to add anything - up to you really. @noklam looked at some of these e2e tests recently so might have an opinion on it.

noklam:
Generally agree with @AntonyMilneQB about the pattern - one file per `kedro xxx` command, so `run.feature` should do a `kedro run` and catch the expected output, and `new.feature` should be checking on the files that are created. 

In that spirit, if we are not changing this structure, I tend to favor just checking the `pipeline` object as it feels more consistent. Similar to unit test that should really just test 1 thing at a time, I would prefer we are just testing the behavior of one `kedro xxx` at a time here.

AntonyMilneQB:
Also, if we do maintain (or adopt) the pattern that `kedro xyz` corresponds to `xyz.feature` then let's rename the file `new.feature` (and similarly for any other misnamed files).

deepyaman:
> Also, if we do maintain (or adopt) the pattern that `kedro xyz` corresponds to `xyz.feature` then let's rename the file `new.feature` (and similarly for any other misnamed files).

Renamed that one. ✅

kedro/framework/project/__init__.py
deepyaman: I killed it, with the move to `iterdir`. :)

kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/pipeline_registry.py
deepyaman: @AntonyMilneQB I've made a new ticket to add pipeline registry documentation (starting with the basic documentation that doesn't currently exist), so I'll resolve this for now.

kedro/framework/project/__init__.py
```diff
@@ -259,3 +261,58 @@ def validate_settings():
...
+    """
+    pipelines_dict = {"__default__": pipeline([])}
+    for pipeline_dir in importlib_resources.files(
+        f"{PACKAGE_NAME}.pipelines"
+    ).iterdir():
+        if not pipeline_dir.is_dir():
+            continue
+        pipeline_name = pipeline_dir.name
+        if pipeline_name == "__pycache__":
+            continue
```
AntonyMilneQB, line 287: This is fine, but just wondering whether there are there any other special cases we should just skip here? Or should we make it more general, e.g. skip anything that starts with `_` or `.`? Maybe there's even something build into pathlib or `importlib_resources` to skip any files that don't look promising?

kedro/framework/project/__init__.py
```diff
@@ -259,3 +261,58 @@ def validate_settings():
...
+        try:
+            pipeline_module = importlib.import_module(
+                f"{PACKAGE_NAME}.pipelines.{pipeline_name}"
```
AntonyMilneQB, line 290: Just to check: you need to reconstruct the dotted module path here and can't just do something like `import_module(pipeline_dir)`?

kedro/framework/project/__init__.py
```diff
@@ -259,3 +261,58 @@ def validate_settings():
...
+            )
+        except:  # pylint: disable=bare-except  # noqa: E722
+            warnings.warn(
+                f"An error occurred while importing the "
+                f"'{PACKAGE_NAME}.pipelines.{pipeline_name}' module. "
+                f"Nothing defined therein will be returned by "
+                f"'find_pipelines'."
+            )
```
AntonyMilneQB, line 298: Is it worth putting the exception message in here too to aid further debugging?

AntonyMilneQB: e.g. I am expecting that a common problem will be that people have the correct pipeline structure but have not installed the dependencies required by the pipeline. The solution to this (`pip install ...`) might be more obvious if the `ImportError` exception were shown. I guess the same exception might be shown when you do `kedro run` anyway though?

tests/framework/project/test_pipeline_registry.py
```diff
@@ -25,7 +25,15 @@ def register_pipelines():
     sys.path.pop(0)
 
 
-def test_pipelines_without_configure_project_is_empty():
+def test_pipelines_without_configure_project_is_empty(
+    mock_package_name_with_pipelines_file,  # pylint: disable=unused-argument
+):
+    # Reimport `pipelines` from `kedro.framework.project` to ensure that
+    # it was not set by a pior call to the `configure_project` function.
+    del sys.modules["kedro.framework.project"]
```
AntonyMilneQB, line 33: I remember spending a LONG time pinning down a very mysterious bug with tests a while ago which resulted in this bit of code:
https://github.com/kedro-org/kedro/blob/59064403e2590c992dceb51942f7dd9f6c97898a/tests/framework/cli/conftest.py#L146-L154

I think you're doing the same thing here and on L37 of `test_pipeline_discovery`. Do you feel ok with this or do we need some sort of `autouse` fixture throughout the whole test suite to make sure this doesn't trip us up anywhere else in future?

tests/framework/project/test_pipeline_discovery.py
```diff
@@ -0,0 +1,170 @@
...
+):
+    configure_project(mock_package_name_with_pipelines)
+    pipelines = find_pipelines()
+    assert set(pipelines) == pipeline_names | {"__default__"}
+    assert sum(pipelines.values()).outputs() == pipeline_names
+
+
+@pytest.mark.parametrize(
+    "mock_package_name_with_pipelines,pipeline_names",
+    [(x, x) for x in [set(), {"good_pipeline"}]],
```
AntonyMilneQB, line 62: Stupid question: why do we need `(x, x)` here? Why can't we just use one argument for both `mock_package_name_with_pipelines` and `pipeline_names`?

tests/framework/project/test_pipeline_discovery.py
```diff
@@ -0,0 +1,170 @@
...
+    indirect=True,
+)
+def test_find_pipelines(
+    mock_package_name_with_pipelines,
+    pipeline_names,
+):
+    configure_project(mock_package_name_with_pipelines)
+    pipelines = find_pipelines()
+    assert set(pipelines) == pipeline_names | {"__default__"}
+    assert sum(pipelines.values()).outputs() == pipeline_names
```
AntonyMilneQB, line 57: I feel like this check isn't _quite_ strict enough since it doesn't actually test that `pipelines[pipeline_name]` maps to the pipeline with output `pipeline_name`. Maybe we should also test that `pipelines["__default__"]` is as we expect? Something like:
```
assert pipelines["__default__"].outputs() == pipeline_names
```

tests/framework/project/test_pipeline_discovery.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+@pytest.fixture
+def pipeline_names(request):
+    return request.param
+
+
+@pytest.mark.parametrize(
+    "mock_package_name_with_pipelines,pipeline_names",
+    [(x, x) for x in [set(), {"my_pipeline"}]],
```
AntonyMilneQB, line 47: Stupid question: why do we need `(x, x)` here? Why can't we just use one argument for both `mock_package_name_with_pipelines` and `pipeline_names`?

AntonyMilneQB (approved):
This is awesome, very excited to have this added! Thank you for all your work on it 🙏 Don't forget to update RELEASE notes when you do the docs also.

I've left lots of minor comments and queries but nothing blocking at all. The only thing that's missing in my mind is some manual testing. Since this is quite a major new feature and a little bit tricky it would be great to have a few people try it out independently and check it all works as expected on a "real" (e.g. `spaceflights` starter) project. Stuff like:
* for a correctly structured project, are the relevant pipelines all registered correctly?
* for incorrectly structured project, are the right warnings raised and clear?
* what happens for a project with no pipelines?

This is all covered by automated tests but some manual testing here too would make me feel even more confident.

I will do this myself if and when I get a chance, but it might be a nice exercise for some of the new team members to familiarise themselves with things a bit more also - @ankatiyar @jmholzer?

AntonyMilneQB:
P.S. @deepyaman do you still need to fork kedro to make a PR?

kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/pipeline_registry.py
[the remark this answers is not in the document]
MerelTheisenQB: Yes agreed that we should give some more clarification in docs.

kedro/framework/project/__init__.py
```diff
@@ -259,3 +261,58 @@ def validate_settings():
         )
 
     importlib.import_module(f"{PACKAGE_NAME}.settings")
+
+
+def find_pipelines() -> Dict[str, Pipeline]:
```
MerelTheisenQB, line 266: Nit: could you add some white lines between some of the steps in the function, it's quite hard to read what parts belong to which if/else/try/catch blocks, e.g. between lines 306 and 307 https://github.com/kedro-org/kedro/blob/9fdd4e868444fd2f75dae19b4db1651862f88db2/kedro/framework/project/__init__.py#L306-L307

deepyaman:
> P.S. @deepyaman do you still need to fork kedro to make a PR?

Nope! And I also see I don't need to explicitly at the DCO comment, if it's on the Kedro repo.

features/steps/test_starter/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/pipeline_registry.py
```diff
@@ -11,7 +11,7 @@ def register_pipelines() -> Dict[str, Pipeline]:
     """Register the project's pipelines.
 
     Returns:
-        A mapping from a pipeline name to a ``Pipeline`` object.
+        A mapping from pipeline names to ``Pipeline`` objects.
```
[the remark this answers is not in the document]
deepyaman: @AntonyMilneQB What would the resulting registry look like?

```python
    pipelines = find_pipelines()
    pipelines["__default__"] = sum(pipelines.values())
    pipelines["data_processing"] = pipeline(
        pipelines["data_engineering"], namespace="data_processing"
    )
    return pipelines
```

And then update the references to use the full "data_engineering", "data_science", "data_processing" instead of "de", "ds", "dp", respectively?

kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/pipeline_registry.py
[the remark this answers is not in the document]
deepyaman: Resolving and adding to the docs ticket.

requirements.txt
```diff
@@ -7,6 +7,7 @@ dynaconf>=3.1.2, <4.0
 fsspec>=2021.4, <=2022.1
 gitpython~=3.0
 importlib_metadata>=3.6  # The "selectable" entry points were introduced in `importlib_metadata` 3.6 and Python 3.10.
+importlib_resources>=1.3  # The `files()` API was introduce in `importlib_resources` 1.3 and Python 3.9.
```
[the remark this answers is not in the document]
deepyaman: Good catch!

kedro/framework/project/__init__.py
deepyaman: Correct; `importlib.import_module` takes a dotted path, as in an import statement, not a directory.

kedro/framework/project/__init__.py
deepyaman: `__pycache__` is a weird case that gets generated automatically and is actually importable. I'm not sure what other scenarios would actually need to get skipped; technically, is there any reason we should stop somebody from defining a pipeline starting with underescore, for example?

tests/framework/project/test_pipeline_discovery.py
deepyaman: I think that's just not how parametrization works? The tuple should be of the same length as the number of variables you're parametrizing, unless I've misunderstood something.

tests/framework/project/test_pipeline_discovery.py
deepyaman: > `assert pipelines["__default__"].outputs() == pipeline_names`

This is essentially what `assert sum(pipelines.values()).outputs() == pipeline_names` is testing, but `pipelines["__default__"] = sum(pipelines.values())` is explicitly written in `register_pipelines()` by default (and not in `find_pipelines`), so wouldn't be part of the unit test.

features/steps/test_starter/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/pipeline_registry.py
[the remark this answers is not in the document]
AntonyMilneQB: If possible I'd like the spaceflights tutorial to work so that you don't have to touch `pipeline_registry.py` at all. I believe this should be possible until this point anyway: https://kedro.readthedocs.io/en/stable/tutorial/namespace_pipelines.html

Currently we do `kedro pipeline create data_processing` and `data_science` and then alias them as `dp` and `ds`. There's no namespacing involved in the spaceflights starter. I think the only thing that should change is that you need to specify the full pipeline name when using `kedro run -p` rather than the alias.

Hence the resulting registry used in the spaceflights starter and in this test starter should just be the same as the default `kedro new` template.

kedro/framework/project/__init__.py
AntonyMilneQB: No, no particular reason. I didn't realise that `__pycache__` was actually importable so now understand the need to skip it!

kedro/framework/project/__init__.py
```diff
@@ -259,3 +261,58 @@ def validate_settings():
...
+
+    Warns:
+        UserWarning: When a module does not expose a ``create_pipeline``
+            function, the ``create_pipeline`` function does not return a
+            ``Pipeline`` object, or if the module import fails up front.
+    """
+    pipelines_dict = {"__default__": pipeline([])}
+    for pipeline_dir in importlib_resources.files(
+        f"{PACKAGE_NAME}.pipelines"
+    ).iterdir():
```
noklam, line 282: I don't know `importlib_resource` well enough, but this seems actually iterating the filesystem.
```
importlib_resources.files(
        f"{PACKAGE_NAME}.pipelines"
    )
```

If this is the case we need to break it down to do error-handling when `importlib_resources.files(
        f"{PACKAGE_NAME}.pipelines"
    )` return `ModuleNotFoundError`

noklam (commented):
I have tested it with the current commit with GitPod by adding the `find_packages` fuction.

By default,  the `find_packages` looks for `package_name/pipelines` module, however these demo project only have a `package_name.pipeline.py` file but not `pipelines` module. As a result I get an error like this.

> 	ModuleNotFoundError: No module named 'project.pipelines'

This can be reproduced by changing the `pipeline_registry` file, note that the Kedro's project name is `project`.
```python
"""Project pipelines."""
from typing import Dict
 
from kedro.pipeline import Pipeline
from project.pipeline import create_pipeline
from kedro.framework.project import find_pipelines  # <-- New added  

	pipelines = find_pipelines()
	return pipelines
```

This is caused by the lines
```python
│ ❱ 280 │   for pipeline_dir in importlib_resources.files(                                         │
│   281 │   │   f"{PACKAGE_NAME}.pipelines"                                                        │
│   282 │   ).iterdir():     
```

AntonyMilneQB:
Very good point @noklam, thanks for the testing! I had thought of this case before but then forgotten all about it... Ideally  I think a very simple project like `pandas-iris` that doesn't have multiple pipelines but just a single pipeline.py file would also "just work" out of the box. i.e. correct behaviour here would be:
1. check if there's a pipeline.py file that exposes `create_pipeline`; if yes then use it. Just define `__default__`, no need for any other pipeline names
2. otherwise look for `pipelines` directory and register them as done now

However, since this single file is a bit of an unusual/special case I don't mind too much if we just don't support it. Or we can come back and add support for it later.

Sorry I forgot to put this in the ticket! 🤦

features/steps/test_starter/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/pipeline_registry.py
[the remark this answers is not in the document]
jmholzer: I managed to get some manual testing done:

1. Standard spaceflights worked without any problems, pipeline was registered correctly.
2. Having no pipelines gave an exception*, only default pipeline was registered.
3. Having no node.py in a pipeline's dir + empty pipeline.py gave an exception, pipeline still registered.
4. Having no pipeline.py in a pipeline's dir gave an exception, pipeline not registered.
5. Having empty node.py / pipeline.py gave an exception, pipeline registered.

* In every case, the exception I got was: `ValueError: Pipeline contains no nodes after applying all provided filters`, which is descriptive enough in my view.

requirements.txt
```diff
@@ -7,6 +7,7 @@ dynaconf>=3.1.2, <4.0
 fsspec>=2021.4, <=2022.1
 gitpython~=3.0
 importlib_metadata>=3.6  # The "selectable" entry points were introduced in `importlib_metadata` 3.6 and Python 3.10.
+importlib_resources>=1.3  # The `files()` API was introduced in `importlib_resources` 1.3 and Python 3.9.
```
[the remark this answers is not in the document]
MerelTheisenQB: Ohh yes I remember, and I remember I was a bit confused then as well 😅 Thanks for explaining again!

kedro/framework/project/__init__.py
deepyaman: Done!

features/steps/test_starter/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/pipeline_registry.py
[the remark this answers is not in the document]
deepyaman: @AntonyMilneQB I've updated this to use `find_pipelines`!

Also, thanks @jmholzer for testing all those scenarios. ❤️

kedro/framework/project/__init__.py
deepyaman: I've updated to print the traceback in the warning:

image

Does that work?

tests/framework/project/test_pipeline_registry.py
deepyaman: I don't think I'm familiar enough with the broader set of issues related to this to feel confident about adding it without more thinking. I'm not sure if this could happen, but I'm also worried in case our tests pass because we unload everything but some actual scenarios that shouldn't pass work (not sure this would actually be possible).

Should we perhaps create a new issue to look into this, or alternatively try to solve it when it happens a third time?

kedro/framework/project/__init__.py
deepyaman: @noklam Sorry, I don't think I follow what you're suggesting here...

deepyaman:
Thanks all for reviewing! I think this is more-or-less ready to go, with 3 open comments/confirmations. Now off to writing the associated docs PR!

tests/framework/project/test_pipeline_registry.py
MerelTheisenQB: I think creating a new issue to look at this separately would be the best idea.

deepyaman: #1787 created

kedro/framework/project/__init__.py
MerelTheisenQB: This looks clear enough to me 👍

MerelTheisenQB (approved):
This is a great addition to Kedro! Thanks for implementing it @deepyaman 😄 ⭐

kedro/framework/project/__init__.py
noklam: @deepyaman Sorry for the late response, GH notifications don't work very well sometimes.

I think this needs to be fixed before this PR is merged. The code block below is outside of the `try/catch` block and it will stop the program running if the `pipelines` folder isn't exist due to `ModuleNotFoundError`.

```python
pipeline_dir in importlib_resources.files(
        f"{PACKAGE_NAME}.pipelines"
    ).iterdir():
```

This assume `importlib_resources.files( f"{PACKAGE_NAME}.pipelines" )` always return a vaild `Path` object, thus the subsequent `iterdir()` method call. However, if the `f"{PACKAGE_NAME}.pipelines"` doesn't exist, you will get a `ModuleNotFoundError` instead. `ModuleNotFoundError.iterdir()` will be a bug and not handled correctly with the current logic.

Some of our starters only have a `package_name/pipeline.py` file but not a `package_name/pipelines/pipeline.py`

To fix this, we probably need something like this
```python
try: 
  module = importlib_resources.files(f"{PACKAGE_NAME}.pipelines")

except ModuleNotFoundError:  # Or maybe some more?
  logger.info(f"{PACKAGE_NAME}.pipelines does not exists")

else:
   # Continue with the iterations...
   for pipeline_dir in module.iterdir():
       ...
```

deepyaman: Based on https://github.com/kedro-org/kedro/pull/1706#issuecomment-1207937897, I assumed that it would be OK to not support for now:

> However, since this single file (compared to multiple directories) is a bit of an unusual/special case I don't mind too much if we just don't support it. Or we can come back and add support for it later. it does seem like a good case to support though since it's the "simplified kedro project structure" that I think we should still support.

I don't think we should swallow the `ModuleNotFoundError`, since `find_pipelines()` (currently) does expect a `pipelines` directory it can iterate over. That being said, it's also a choose-to-use feature, that's enabled by default when you do `kedro new`. Something with a simplified structure without a `pipelines` directory would simply not use this.

I personally am not even convinced that the simplified structure needs to be supported. If you just have a `pipeline.py` file, with the one `create_pipeline()` method in there, why do you need a magical `find_pipelines()`? The value of find pipelines is that you don't need to register each pipeline when you create it; in this case, just be explicit and define the one pipeline in the registry.

noklam: @deepyaman Thanks for the comment! I think you are right that the error shouldn't be swallowed.

My view is slightly different:
1. I have a stronger view that we should support the simple `pipeline.py` case.
2. The error should still be handled and throw a more explicit error, since the module is not imported by the user, it's not obvious why it is failing. It's also better if we mentioned this expected structure in the docstring 

I mainly see this feature as a **beginner-friendly** feature, something that "just work" even if you don't understand how pipeline registration works, as long as you are creating a project with `kedro new` and `kedro create pipelines` (no need to edit the `pipeline_registry.py`).

We also have [pandas-iris](https://github.com/kedro-org/kedro-starters/tree/main/pyspark-iris/%7B%7B%20cookiecutter.repo_name%20%7D%7D/src/%7B%7B%20cookiecutter.python_package%20%7D%7D) and other `iris` starters that actually follow this simplified structure. It is strange to me a beginner-friendly feature is not available for a simple project.

```python
    """Register the project's pipelines.
    Returns:
        A mapping from a pipeline name to a ``Pipeline`` object.
    """

    return {
        "__default__": create_pipeline(),
    }
```
If we support case 1, then 99% of the cases are covered and we won't see the error very often (Point 2) unless users are doing some custom thing (not using standard starters or they have custom structure/logic), which is okay to throw an error. 

Imagine a beginner that starts with the `pandas-iris` project, and he found `modular_pipeline` later, how likely is the user going to find out the `find_pipelines` function afterward? It would be a smoother experience if the pipeline is discoverable after he does `kedro pipeline create`. It's probably because I always view `starters` and Kedro's CLI as entry points for feature discovery.

I think a unified workflow is quite important, especially if we are focusing on expanding the user base with less software engineering background. We shouldn't try too hard to explain.

deepyaman: @noklam What's the user workflow if they start with the simplified structure, then `kedro pipeline create another_one`? Would the expected behavior be to include pipelines from both `pipeline.py` and `pipelines/another_one/pipeline.py`? Or would they refactor the `pipeline.py` into `pipelines/my_first/pipeline.py`?

noklam: @deepyaman Ideally they should be refactored into folders at the end, but I don't think it's unreasonable to combine both of them as `__default__`.

In that case, `find_pipelines` is always there and users don't have to update this block, and they can also do `kedro run --pipeline=modular_pipeline`.

```python
  pipelines = find_pipelines()
  pipelines["__default__"] = sum(pipelines.values())
  return pipelines
```

[Antony has a different view that](https://github.com/kedro-org/kedro/pull/1706#issuecomment-1207937897) it should stop as soon as `pipeline.py` is found.

AntonyMilneQB: Very good discussion here! I started 👍 the ones that I agreed with and that ended up being all the points made. I'll make a new issue so we can discuss this further. 

For the time being I'm happy to merge this without supporting the simple project structure (single `pipeline.py` file) case at all.

@deepyaman please could you update the starters which do have the multiple pipeline structure (probably just `spaceflights` from memory) to use the new `find_pipelines` functionality? 🙂

AntonyMilneQB: See https://github.com/kedro-org/kedro/issues/1812.

noklam:
Quick question, does `kedro new`  (with no starters) generate a project with `find_pipelines`?

AntonyMilneQB:
> Quick question, does `kedro new` (with no starters) generate a project with `find_pipelines`?

Yes.

kedro/framework/project/__init__.py
deepyaman: @AntonyMilneQB Thank you for chiming in. I'm in the process of updating to support single pipelines, but would also be more than happy to merge first and push it as a separate PR.

> @deepyaman please could you update the starters which do have the multiple pipeline structure (probably just `spaceflights` from memory) to use the new `find_pipelines` functionality? 🙂

Doesn't this need to be released first?

AntonyMilneQB: Yes indeed, this must be released first - just wanted to mention it while I remembered.

Issue #1583, opened 2022-06-01
Investigate `KedroSession.create: ModuleNotFoundError: No module named 'None'`


PetervanHeck:
## Description
Following the documentation on creating and running a session causes an error:
https://kedro.readthedocs.io/en/0.18.1/kedro.framework.session.session.KedroSession.html

When creating a KedroSession, the `validate_settings()` appears not to get the right 'PACKAGE_NAME', and thus fails.

However, following the following page does work:
https://kedro.readthedocs.io/en/stable/kedro_project_setup/session.html

## Context
Starting a Kedro session from within Python.

## Steps to Reproduce
```
from kedro.framework.session import KedroSession
with KedroSession.create('<package_name>') as session: session.run()
```

## Expected Result
Running the kedro project similarly to 'kedro run'

## Actual Result

```
>>> from kedro.framework.session import KedroSession
>>> session = KedroSession.create('my_package_name')
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "C:\Users\my_user\Environments\my_env\lib\site-packages\kedro\framework\session\session.py", line 138, in create
    validate_settings()
  File "C:\Users\my_user\Environments\my_env\lib\site-packages\kedro\framework\project\__init__.py", line 223, in validate_settings
    importlib.import_module(f"{PACKAGE_NAME}.settings")
  File "C:\Users\my_user\AppData\Local\Programs\Python\Python37\lib\importlib\__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 953, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 965, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'None'
```

## Your Environment
Include as many relevant details about the environment in which you experienced the bug:

* Kedro version used (`pip show kedro` or `kedro -V`): 0.18.0 and 0.18.1
* Python version used (`python -V`): 3.7.9
[...]

PR #1713, opened 2022-07-19
Update documentation to show how to use `KedroSession` outside of CLI mode

noklam:

## Description

#1583 introduce a behavioral change that was introduced in `0.18.0`, this PR updated the documentation to better reflect this.

## Development notes

- [x] update the documentation to reflect when does a user need to do `bootstrap_project`

bonus:
- [x] a more helpful message when you are not running kedro in the right directory
- [x] removed the duplicate code in `ipython` extension

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1713"></a>


kedro/extras/extensions/ipython.py
```diff
@@ -58,7 +58,6 @@ def reload_kedro(
 
     _remove_cached_modules(metadata.package_name)
 
-    configure_project(metadata.package_name)
```
noklam, line 61: The `bootstrap_project` method already included the `configure_project`, so I don't think it is needed

noklam:
Dev notes:
* Intended to add some helpful message denoted that user is running outside of a kedro project - found that `kedro-telemetry` is initializing the `KedroCLI` again so it causes duplicate message.
https://github.com/quantumblacklabs/kedro-telemetry/blob/2a01f3fd7c9ff33a849e8293edfe52475ed9aa18/kedro_telemetry/plugin.py#L46

noklam:
image

Mostly base on @datajoely's magic, I added some more so it will attempt to find the project if it finds a kedro project and show it in the suggestion.

The last commit could be done at a separate commit, but I find this is kinda related anyway since it's dealing with things break in "out of project" mode.

datajoely:
Really nice work @noklam  - would you mind providing some screenshots? I'll defer any points on implementation to the rest of the group, but would love to see the user journey here.

noklam:
If users are in the project directory:
Business as usual, any kedro command should work

If users are not in the project directory:
* 'kedro' - should a message suggesting this is running outside of a project? (Not sure so i didn't change it) - it's unclear whether user are aware of the project commands, there are no error but it will just show the global command. 

* 'kedro run' or any project command - it's pretty clear user are aware of these commands but just from wrong directory, so it will fail and show the screenshot above. If a project path is found, it will just show the path, if not it will revert to a simple help message suggesting user should look for the directory where the pyproject.toml


Typing this on my phone so sorry for badly formoatted.
@datajoely

datajoely:
I didn't realise that was a new screenshot, I thought it was mine from ages ago. Nice work!

kedro/extras/extensions/ipython.py
AntonyMilneQB: I've been meaning to do this for ages. I didn't do it before because at a glance I wasn't 100% sure whether it was intentional to have both `bootstrap_project` and `configure_project` here. Note there's a call to `_remote_cached_modules` in between - does that mean that actually we need to repeat the call to `configure_project`?

I'm very much in favour of removing this if we can, but please do double check to see that all still works correctly after removing it! Possibly there's a good reason it's there.

kedro/framework/session/session.py
```diff
@@ -78,14 +78,22 @@ class KedroSession:
...
         >>> from kedro.framework.session import KedroSession
-        >>>
+        >>> from kedro.framework.startup import bootstrap_project
+        >>> from pathlib import Path
+
+        >>> # If you are creating session outside of a kedro project, i.e. not using
+        >>> # `kedro run` or `kedro jupyter`. You need to run `bootstrap_project` to
+        >>> # let Kedro find your configuration.
+        >>> bootstrap_project(<your-pyproject.toml-directory>)
         >>> with KedroSession.create("<your-kedro-project-package-name>") as session:
```
AntonyMilneQB, line 94:
```suggestion
        >>> metadata = bootstrap_project(Path("<project_root>"))
        >>> with KedroSession.create(metadata.package_name) as session:
```
@deepyaman recently made all references to the project root in documentation look like this I think.

kedro/framework/project/__init__.py
```diff
@@ -253,4 +253,12 @@ def validate_settings():
     error message ``Expected an instance of `ConfigLoader`, got `NoneType` instead``.
     More info on the dynaconf issue: https://github.com/rochacbruno/dynaconf/issues/460
     """
+    if not PACKAGE_NAME:
```
AntonyMilneQB, line 256:
```suggestion
    if PACKAGE_NAME is None:
```
It's almost always slightly preferable to check against `None` like this.

kedro/framework/project/__init__.py
```diff
@@ -253,4 +253,12 @@ def validate_settings():
...
+        raise ValueError(
+            """Package name is not found, make sure you have set up the
+        project via ``bootstrap_project``, this should happens automatically if you
+        are using `kedro` command lines interface.
+        """
```
AntonyMilneQB, line 261:
```suggestion
            """Package name not found. Make sure you have configured the
        project using 'bootstrap_project'. This should happen automatically if you
        are using Kedro command line interface.
        """
```

kedro/framework/cli/cli.py
```diff
@@ -107,15 +115,50 @@ class KedroCLI(CommandCollection):
...
+            if _is_project(guessed_project_root):
+                self._guessed_project_root = guessed_project_root
         self._cli_hook_manager = get_cli_hook_manager()
 
         super().__init__(
             ("Global commands", self.global_groups),
             ("Project specific commands", self.project_groups),
         )
 
+    def get_command(self, ctx: Context, cmd_name: str) -> Optional[Command]:
```
AntonyMilneQB, line 133: I'm a huge fan of this change in principle, thank you for adding it! However, please could you split it off into a separate PR? There's a few things I think need discussing:

* not at all sure about using `black` for `find_project_root`. Note we have something similar `_find_kedro_project` in the ipython extension; if that seems suitable to reuse, let's move it to some common utils file and use it here
* more importantly, if we are inside a kedro project, just not at the top level, is there actually any reason these project commands shouldn't work? 🤔 Instead of an error message telling the user to `cd` can we actually just somehow make kedro work inside a project subfolder (like `kedro ipython` does)? But without `cd`ing to the kedro project top-level
* if we're totally outside a kedro project, a helpful error message is great but:
  * is there some better way to check if the command is a project one without overriding `get_command`, e.g. check if it's in `self.project_groups`?
  * I don't think this formatting is "on brand" yet since we haven't moved the rest of the CLI to rich yet. Let's just do it as `click.secho` for now. Doing rich styling for CLI messages is a separate issue for another time (e.g. we might move to https://github.com/ewels/rich-click)

noklam: #1720 Separate PR created here, I put it in the inbox so we can discuss it a bit more.

tests/extras/extensions/test_ipython.py
```diff
@@ -25,6 +25,11 @@ def pipeline_cleanup():
     pipelines.configure()
 
 
+PACKAGE_NAME = "fake_page_name"
+PROJECT_NAME = "fake_project_name"
+PROJECT_VERSION = "0.1"
+
+
```
noklam, line 32: just minor refactoring since these values appears a few time.

tests/extras/extensions/test_ipython.py
```diff
@@ -53,9 +58,7 @@ def my_register_pipeline():
             "kedro.framework.project._ProjectPipelines._get_pipelines_registry_callable",
             return_value=my_register_pipeline,
         )
-        mocker.patch("kedro.framework.project.settings.configure")
-        mocker.patch("kedro.framework.session.session.validate_settings")
-        mocker.patch("kedro.framework.session.KedroSession._setup_logging")
```
noklam, line 58: @AntonyMilneQB Are these actually needed? I found that they are not in the `test_load_kedro_objects_extra_args` test and removing them seems to have no effect, I am not sure why it is needed.

tests/extras/extensions/test_ipython.py
```diff
@@ -67,16 +70,21 @@ def my_register_pipeline():
         mock_register_line_magic = mocker.patch(
             "kedro.extras.extensions.ipython.register_line_magic"
         )
-        mock_context = mocker.patch("kedro.framework.session.KedroSession.load_context")
```
noklam, line 70: We should use session instead

kedro/extras/extensions/ipython.py
AntonyMilneQB: Sounds like you are confident we can remove this, so please go ahead as you were before 👍

tests/extras/extensions/test_ipython.py
AntonyMilneQB: Probably not needed 👍

AntonyMilneQB (approved):

AhdraMeraliQB (approved):
Great work! Good catch on the test refactoring too!

Issue #1632, opened 2022-06-18
Add a feature to allow user customize CLOUD_PROTOCOLS when developing new fsspec protocols


Felipeasg:
## Description
Is a little bit boring stay editing kedro [source code](https://github.com/kedro-org/kedro/blob/main/kedro/io/core.py#L31) when developing new protocols to different storage media. This list of `CLOUD_PROTOCOLS` could be defined in a declarative format or using some API to extend the list of available protocols.

## Context
When you are developing new protocols extending **fsspec** **AbstractFileSystem** to another storage media, is interesting add a simple way to add the protocol in kedro framework. 
This will avoid the change in the list of `CLOUD_PROTOCOLS` be overridden  always when you reinstall kedro to update version or create a new virtual environment.

## Possible Implementation
- Create a API function to be called that add new protocol in `CLOUD_PROTOCOLS` list (e.g. `add_new_cloud_protocol('some_protocol')`

AntonyMilneQB:
To do:
- [ ] work out whether we can go back to using fsspec's own `infer_storage_options` or not
- [ ] if yes: go back to using it. Remove `CLOUD_PROTOCOLS` (technically breaking and should only be done in `develop` if we think anyone might be using it)
- [ ] if no: figure out which protocols we're missing from fsspec and add them to `CLOUD_PROTOCOLS`

Felipeasg:
Hi Antony,

Thanks by your answer, really `fsspec` already have `infer_storage_options` and seems that kedro is duplicating this functionality, but if I understood well the `CLOUD_PROTOCOLS` and parser was inserted to avoid problems due strange behavior with old implementations of `fsspec` parser.

Really I agree add a python entry point to add new protocols could be a overkill.

Today we avoid problems with new installations of kedro using something like bellow, but we are not comfortable with it:
```shell
# Add a manual entry on Kedro IO protocols
# Look for Gdrive in Protocols
j=`grep gdrive -c $kedro_io`
if [ $j -eq 0 ]; then
    echo "Missing gdrive cloud protocol. Inserting..."
    # Substitute
    sed -i '/CLOUD_PROTOCOLS/s/"abfs"/"abfs", "gdrive"/' $kedro_io
    echo "Done"
else
    echo "Cloud protocol gdrive already inserted. Skiping!"
fi
```

[...]

PR #1714, opened 2022-07-19
Use fsspec's infer_storage_options

AntonyMilneQB:
## Description
Closes: #1632.

Thanks to https://github.com/fsspec/filesystem_spec/pull/988 we no longer need to write our own version of fsspec's `infer_storage_options`. However, we do need to wait until they do a release before we can use that code.

The tests in this PR will only pass once fsspec do that release and we edit our requirements.txt to include that new version of fsspec. So we need to keep an eye on fsspec to see when that release happens. When it does:
* edit requirements.txt as in the `TODO` to get the new fsspec version in as a dependency
* re-run tests
* assuming they pass, remove `test_parse_filepath` completely (it's no longer needed since it will all be tested on the fsspec side)
* merge this PR

Note that removing `CLOUD_PROTOCOLS` was only ever used in `_parse_filepath` and never advertised anywhere, so even though it's not got `_` in front of it I think it's fine to remove. If anyone wanted to use it to modify the list then they would have raised an issue to modify it on the kedro side (which has happened in the past). So I think we can safely remove this variable.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1714"></a>

### DIFF
--- a/tests/io/test_core.py
+++ b/tests/io/test_core.py
@@ -94,6 +94,7 @@ class TestCoreFunctions:
         ],
     )
     def test_parse_filepath(self, filepath, expected_result):
+        # TODO: delete this test completely once it passes
         from fsspec.utils import infer_storage_options
 
         assert infer_storage_options(filepath).items() >= expected_result.items()
--- a/tests/io/test_core.py
+++ b/tests/io/test_core.py
@@ -1,11 +1,12 @@
 from decimal import Decimal
 from fractions import Fraction
 from pathlib import PurePosixPath
-from typing import Any, List
+from typing import Any, List, re
+from urllib.parse import urlsplit
 
 import pytest
 
-from kedro.io.core import AbstractDataSet, _parse_filepath, get_filepath_str
+from kedro.io.core import AbstractDataSet, get_filepath_str
 
 # List sourced from https://docs.python.org/3/library/stdtypes.html#truth-value-testing.
 # Excludes None, as None values are not shown in the str representation.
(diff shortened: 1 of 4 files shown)


tests/io/test_core.py
```diff
@@ -93,4 +94,6 @@ def test_get_filepath_str(self):
         ],
     )
     def test_parse_filepath(self, filepath, expected_result):
-        assert _parse_filepath(filepath) == expected_result
+        from fsspec.utils import infer_storage_options
```
AntonyMilneQB, line 97:
```suggestion
        # TODO: delete this test completely once it passes
        from fsspec.utils import infer_storage_options
```

MerelTheisenQB:
Looks like[ fsspec has done a release ](https://pypi.org/project/fsspec/#history)a couple of days ago 😄

AntonyMilneQB:
@MerelTheisenQB indeed, but unfortunately it got reverted a couple of days later since apparently it broke other stuff 😬 I'm trying to figure out whether there's a solution on their side: https://github.com/fsspec/filesystem_spec/issues/1002.

MerelTheisenQB:
> @MerelTheisenQB indeed, but unfortunately it got reverted a couple of days later since apparently it broke other stuff 😬 I'm trying to figure out whether there's a solution on their side: [fsspec/filesystem_spec#1002](https://github.com/fsspec/filesystem_spec/issues/1002). If not then I'm afraid we'll have to keep our own `infer_storage_options` function :(

Ah that's a pity.. 😞

AntonyMilneQB:
Unfortunately I don't think this is going to be fixed on the fsspec side soon, so we will continue to maintain our own `infer_storage_options` function. tbh I don't understand why that apparently works when the fix on their side doesn't :/

noklam:
The original issue of replacing `infer_storage_option`
https://github.com/quantumblacklabs/private-kedro/pull/545


Quote 
> We need to split filepath without using `infer_storage_options()` from fsspec since it returns different results for different protocols that doesn't satisfy our needs (we are already making exceptions for HTTP protocol, we should avoid piling on exceptions e.g. for Azure).

Issue #1091, opened 2021-12-03
Running `kedro install` , `kedro ipython` or `kedro jupyter notebook` from my project root, I get `No such command` errors


AndrewLaganaro:
## Description
When running `kedro install` , `kedro ipython` or `kedro jupyter notebook` from my project root, I get `No such command` errors.

## Context
I've just intalled kedro on a new env and I'm trying to use it on a notebook.` jupyter notebook` command works normally, opening a fully functional notebook instance. Same doesn't happens with  `kedro jupyter notebook`

## Steps to Reproduce
1. Created new conda env, installed jupyter notebook and kedro
2. use `kedro jupyter notebook` or `kedro install`

## Expected Result
Kedro with notebook should open
## Actual Result
An error occurs:
```
Usage: kedro [OPTIONS] COMMAND [ARGS]...
Try 'kedro -h' for help
No such command 'jupyter'
```

```
Usage: kedro [OPTIONS] COMMAND [ARGS]...
Try 'kedro -h' for help
No such command 'install'
```
## Your Environment
* Kedro: 0.17.5
* Python: 3.6.12
* OS: Windows 10

datajoely:
Hi @AndrewLaganaro if you type `which python` does it point to the conda environment you created? Also if you do `pip freeze` can you see Kedro in the active python environment? 

This error usually means that something is up with the active environment.

shawnboltz:
I'm having the same issue in a new conda environment (Installed about a month ago, just barely got around to using it today). I was able to use `kedro run` to run a pipeline just fine, but now I'm not able to launch `kedro jupyter lab`, saying that it doesn't recognize `jupyter` (jupyter is installed in the environment and I can run it separately).

Kedro: 0.17.0
Python: 3.8.12
OS: Ubuntu 16.04

Relevant lines from `pip freeze`:
jupyter @ file:///home/conda/feedstock_root/build_artifacts/jupyter_1637233111631/work
jupyter-client @ file:///tmp/build/80754af9/jupyter_client_1640335223713/work
jupyter-console @ file:///home/conda/feedstock_root/build_artifacts/jupyter_console_1616560109969/work
jupyter-core @ file:///home/conda/feedstock_root/build_artifacts/jupyter_core_1636814262349/work
jupyter-server @ file:///home/conda/feedstock_root/build_artifacts/jupyter_server_1635811281209/work
jupyterlab==0.31.1
jupyterlab-launcher==0.10.5
[...]

PR #1720, opened 2022-07-21
Provide better error message when running project command like `kedro run` not from the root directory

noklam:

## Description

Spinning off from #1713 

The goal is to provide a more helpful message instead of `run command is not found` when user is running `Kedro's CLI` not from the project directory.

# Notes
* This PR already added a better error message 

## Points to discuss for Backlog grooming
* Should we take one more step to enable user run `kedro xxx` not from the top-directory instead of just showing the help message? (

Antony raised some good questions in #1713 

> * not at all sure about using black for find_project_root. Note we have something similar _find_kedro_project in the ipython extension; if that seems suitable to reuse, let's move it to some common utils file and use it here
> * more importantly, if we are inside a kedro project, just not at the top level, is there actually any reason these project commands shouldn't work? 🤔 Instead of an error message telling the user to cd can we actually just somehow make kedro work inside a project subfolder (like kedro ipython does)? But without cding to the kedro project top-level
> * if we're totally outside a kedro project, a helpful error message is great but:
> * is there some better way to check if the command is a project one without overriding get_command, e.g. check if it's in self.project_groups?
> * I don't think this formatting is "on brand" yet since we haven't moved the rest of the CLI to rich yet. Let's just do it as click.secho for now. Doing rich styling for CLI messages is a separate issue for another time (e.g. we might move to https://github.com/ewels/rich-click)

# Example
The current implementation will try to find the project directory and shows a more helpful error message when user is running `kedro run` not from the top directory.
image

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1720"></a>

### DIFF
--- a/kedro/framework/cli/cli.py
+++ b/kedro/framework/cli/cli.py
@@ -2,14 +2,20 @@
 
 This module implements commands available from the kedro CLI.
 """
+
 import importlib
+import logging
 import sys
 import webbrowser
 from collections import defaultdict
 from pathlib import Path
-from typing import Sequence
+from typing import Optional, Sequence
 
 import click
+from black import find_project_root, find_pyproject_toml, find_user_pyproject_toml
+from click.core import Command, Context
+from rich.console import Console
+from rich.panel import Panel
 
 from kedro import __version__ as version
 from kedro.framework.cli.catalog import catalog_cli
@@ -30,6 +36,8 @@ from kedro.framework.cli.utils import (
 )
 from kedro.framework.project import LOGGING  # noqa # pylint:disable=unused-import
 from kedro.framework.startup import _is_project, bootstrap_project
+
+logger = logging.getLogger(__name__)
 
 LOGO = rf"""
  _            _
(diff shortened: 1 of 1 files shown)


AntonyMilneQB:
More evidence that this would be handy: https://github.com/kedro-org/kedro/issues/1091.

My current feeling is that we should just allow `kedro` project commands to be run anywhere within the project directory, like we can already do with `kedro ipython`.

noklam:
> My current feeling is that we should just allow kedro project commands to be run anywhere within the project directory (not just root), like we can already do with kedro ipython. The error message here is still useful if you're not inside the project directory at all, but in the case that you are somewhere inside the project it would be nice if things just worked rather than giving an error.

When you say "we can already do with kedro ipython" do you mean we can run a `KedroSession` not from the root directory? Since I don't think any Kedro's project commands are available.
https://github.com/kedro-org/kedro/blob/e882862f0eeaa16656ad9d01abbae083fc47bd4c/kedro/extras/extensions/ipython.py#L51-L67

AntonyMilneQB:
@noklam I mean you can do `kedro ipython` from any directory that's inside a project, e.g. from inside `<project_root>/notebooks` folder. All we do is iterate up through parent directories and supply `KedroSession` with the right path according to that (see `_find_kedro_project`). Why shouldn't all `kedro` project commands work that way? i.e. why shouldn't I be able to do `kedro run` from inside a subdirectory of `<project_root>`.

AntonyMilneQB:
To do: check how other tools (e.g. git) handle operations inside a project that are not in the project root. If they allow this then we should too for all project commands (e.g. `kedro run`).

noklam:
Closed for now, for future implementation for #1820 we can take this PR as reference.

Issue #1712, opened 2022-07-19
logs print credentials if error in catalog with Rich Logging show_locals=True


Debbby57:
## Description
I run a pipeline but I made a mistake with a credentials key for fsspec (I used `HOST` instead of `host` ). Kedro crashes (I agree) but in the logs, all my credentials have been printed. Those credentials for this dataset are printed multiple times but all my credentials were printed too.

## Steps to Reproduce
1. create a credentials dictionnary to connect with a sftp connection
2. make a mistake in a key (for example : `HOST `instead of `host)
3. create a dataset in the catalog with needs thhose credentials
4. run the pipeline

## Expected Result
the log should hide credentials.

## Actual Result
the log prints all credentials.

```
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│  │
│ /site-packages/kedro/io/core.py:155 in from_config                                               │
│                                                                                                  │
│   152 │   │   │   ) from exc                                                                     │
│   153 │   │                                                                                      │
│   154 │   │   try:                                                                               │
│ ❱ 155 │   │   │   data_set = class_obj(**config)  # type: ignore                                 │
│   156 │   │   except TypeError as err:                                                           │
│   157 │   │   │   raise DataSetError(                                                            │
│   158 │   │   │   │   f"\n{err}.\nDataSet '{name}' must only contain arguments valid for the "   │
│                                                                                                  │
│ ╭─────────────────────────────────────────── locals ───────────────────────────────────────────╮ │
│ │    class_obj = <class 'kedro.extras.datasets.pandas.csv_dataset.CSVDataSet'>                 │ │
│ │          cls = <class 'kedro.io.core.AbstractDataSet'>                                       │ │
│ │       config = {                                                                             │ │
│ │                │   'filepath':                                                               │ │
│ │                'sftp://my_path/s… │ │
│ │                │   'save_args': {'sep': ';'},                                                │ │
│ │                │   'credentials': {                                                          │ │
│ │                │   │   'username': '',                                                │ │
│ │                │   │   'password': '',                                              │ │
│ │                │   │   'HOST': '',                                             │ │
│ │                │   │   'port': ''                                                          │ │
│ │                │   }                                                                         │ │
│ │                }                                                                             │ │
│ │ load_version = None                                                                          │ │
│ │         name = 'my_data'                                                              │ │
│ │ save_version = '2022-07-19T08.14.05.166Z'                                                    │ │
[...]

PR #1726, opened 2022-07-26
Use `show_locals` False default for rich logging

MerelTheisenQB:

## Description
When we added `rich` as our default logger we set `show_locals=True` to enable display of local variables. However, this can lead to credentials being shown in logs: https://github.com/kedro-org/kedro/issues/1712

## Development notes
Instead just use the default value for `show_locals` which is False. 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1726"></a>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -14,6 +14,7 @@
 ## Major features and improvements
 
 ## Bug fixes and other changes
+* Use default `False` value for rich logging `set_locals`, to make sure credentials and other sensitive data isn't shown in logs.
 
 ## Upcoming deprecations for Kedro 0.19.0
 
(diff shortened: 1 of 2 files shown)


noklam (approved):
LGTM, I think a conservative default make sense here. We should also think about how to let user disable or configure these rich hooks that is installed by default.

datajoely (approved):
Is this the right time to put in some things in `pyproject.toml` like `[tool.kedro.rich]` etc?

MerelTheisenQB:
> Is this the right time to put in some things in `pyproject.toml` like `[tool.kedro.rich]` etc?

You mean to configure `rich`? I think that would make sense, but let's create an issue for it.

PR #1734, opened 2022-07-28
Make Kedro instantiate datasets from `kedro_dataset`with higher priority than `kedro.extras.datasets`

### LINKED ISSUE #1494: Make Kedro instantiate datasets from `kedro-datasets` with higher priority than `kedro.extras.datasets`
https://github.com/kedro-org/kedro/blob/1b1558952c059eea5636d9ccf9a883f9cf4ef643/kedro/io/core.py#L346

noklam:

## Description

Close #1494

We are going to introduce `kedro-datasets` as a new Kedro's plugins for datasets. It will be initiated with a higher priority in case of conflicting datasets. The `kedro.extras.datasets` will be removed in 0.19.0 completely, but for 0.18.x we will keep it compatible thus we need to make sure the `kedro-datasets` package has a higher priority.

## Development notes

Tested manually. Adding an automatic test is a tricky since we don't want to introduce `kedro-datasets` into the test dependency.

```
from kedro.io import DataCatalog


config = {
        "catalog": {
            "boats": {"type": "pandas.CSVDataSet", "filepath": "123"},
        },
    }

catalog = DataCatalog.from_config(**config)


print(catalog._data_sets)
# {'boats': <kedro_datasets.pandas.csv_dataset.CSVDataSet at 0x7f96b1493d60>}
```

### More notes on test
I try to avoid touching the `sys.modules` as it seems to be problematic and creates some unknown side effects in our tests. We did it for a few tests to force reload. I can't find a better way to test a non-existing module, mocking a module is not trivial.

In the end, I go with spying, which basically swaps the method with a custom one (It's a bit annoying that it has to be a class, you can't spy a function), but it works. I would be very pleased if there is a better way to do this.
## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1734"></a>


noklam:
This was tested manually with local testing, maybe that's enough already. Optionally we could do add a test to make sure the `kedro-dataset` takes higher priority when it get instantiated, but that need to be done when the `kedro-dataset` package is published properly.

RELEASE.md
```diff
@@ -11,8 +11,9 @@
 # Upcoming Release 0.18.4
 
 ## Major features and improvements
-* The config loader objects now implement `UserDict` and the configuration is accessed through `conf_loader['catalog']`
-* You can configure config file patterns through `settings.py` without creating a custom config loader
+* Make Kedro instantiate datasets from `kedro.datasets` with higher priority than `kedro.extras.datasets`. `kedro.datasets` is the namespace for the new `kedro-datasets` python package.
```
merelcht, line 14: This should be `kedro_datasets` instead of `kedro.datasets` right?

merelcht (approved):
I like that you managed to create a test for this!! 👏 🏆

tests/io/test_data_catalog.py
```diff
@@ -373,6 +380,20 @@ def test_config_relative_import(self, sane_config):
         with pytest.raises(DataSetError, match=re.escape(pattern)):
             DataCatalog.from_config(**sane_config)
 
+    def test_config_import_kedro_datasets(self, sane_config, mocker):
+        """Test kedro.extras.datasets default path to the dataset class"""
+        # Spy _load_obj because kedro_datasets is installed
+        class DummyMock:
```
jmholzer, line 386: I think `DummyMock` and `dummy_mock` could have more meaningful names, would `DummyLoader` work here?

jmholzer, line 386: Could this be a function instead of a class? That may also make the test clearer. I checked the `pytest-mock` docs, it's not obvious if it can be in this case.

jmholzer (changes requested):
Nice work!

Really clever test, but I wonder if it could be a bit clearer. I left two comments but otherwise happy to approve.

tests/io/test_data_catalog.py
noklam: This is a good question. This isn't possible, the spy need to work with a class and method. I will add comments on it.

noklam: @jmholzer I manage to remove the dummy object, turns out I can spy on the `kedro.io.core` module object directly, so the `patch` is not needed too.

jmholzer (approved):
Good work 🎉.

I like the test, the comments and structure make things clear.

Issue #1736, opened 2022-07-30
Support for defining a pyarrow schema for dask.ParquetDataSet in catalog.yml


avsolatorio:
## Description
I am working on a pipeline that outputs a dask ParquetDataSet. The inferred schema is not accurate, so I need to specify the schema manually. I don't think the current implementation supports the definition of complex schema, e.g., nested list of int64.

## Context
This change will allow users to be able to have more expressivity in the kind of dataset they can output. This is significantly important when the automatic inference of the schema fails.

## Possible Implementation
I have implemented a local change that worked. It checks for the presence of a schema key in the save_args, then assumes that the items are pyarrow definitions of the expected field types for the columns.

I can do a PR to add this.

## Possible Alternatives
I haven't seen any other alternatives or workaround. I would be interested to know if this can be done in any other way.

datajoely:
We'd very much appreciate a PR @avsolatorio ! This sounds like a useful addition.

avsolatorio:
> We'd very much appreciate a PR @avsolatorio ! This sounds like a useful addition.

PR opened for this @datajoely : https://github.com/kedro-org/kedro/pull/1746 :)

PR #1746, opened 2022-08-01
Implement support for schema in saving dask parquet (#1736)

avsolatorio:
Signed-off-by: Aivin V. Solatorio <avsolatorio@gmail.com>

## Description
Resolves: https://github.com/kedro-org/kedro/issues/1736

## Development notes
This change allows the use of the schema argument in the `dask.to_parquet` API from kedro's `dask.ParquetDataSet`.

However, the solution requires a couple of non-pythonic implementations. The solution requires the use of `eval` which I'm also hesitant about, but I can't find an alternative solution at this point. Suggestions are totally welcome!

No explicit update on the documentation was done since this feature is a functionality of the underlying `dask.to_parquet` API.

The change requires the parsing of the `_save_args` for the `schema` key and handles the transformation of the fields to a `pyarrow.DataType or `pyarrow.Schema` accordingly.

Tests have also been written for this change.

## Checklist


kedro/extras/datasets/dask/parquet_dataset.py
```diff
@@ -105,8 +107,36 @@ def _load(self) -> dd.DataFrame:
...
+                    assert field.startswith("pa.") or field.startswith("pyarrow.")
+                    self._save_args["schema"][
+                        col
+                    ] = eval(  # nosec pylint: disable=W0123,I0021
+                        field
+                    )
+        elif isinstance(schema, str) and (
+            schema.startswith("pa.schema") or schema.startswith("pyarrow.schema")
+        ):
+            schema = eval(schema)  # nosec pylint: disable=W0123,I0021
```
datajoely, line 137: So I completely get why you want to do this, but I think we're unlikely to accept a raw `eval` statement into the core project due to the security risks. Additionally, it would reduce the project's score in the eyes of vendors like [Snyk](https://snyk.io/advisor/python/kedro).

datajoely:
Thanks for raising this @avsolatorio it's looking real good!

kedro/extras/datasets/dask/parquet_dataset.py
avsolatorio: @datajoely yeah, I totally understand. I'm using this patch locally for my use case, but I'm also hesitant about this because of the same reasons you pointed out. Just opening this in case someone has better ideas to make this work. Feel free to close anytime as well! Thank you! :D

avsolatorio:
@datajoely I have an idea, but this might take some time.

I'm looking at using `ast` to explicitly parse the input expressions. Then I will implement something that rebuilds the field definitions based on the parsed data. Does that sound acceptable? :)

I'll make this PR a draft until I finish that.

deepyaman:
I don't think PyArrow provides a native way to define complex schema as text (as mentioned in your issue), although simple stuff like `pa.schema({'hi': 'u1'}) == pa.schema({'hi': pa.uint8()})` works. However, [Fugue](https://fugue-tutorials.readthedocs.io/tutorials/advanced/schema_dataframes.html#schema) provides a way to do so. Maybe we should just use their logic, since it's non-trivial? The requirements for https://github.com/fugue-project/triad also look pretty lightweight, so I don't see an issue with depending on it.

avsolatorio:
> I don't think PyArrow provides a native way to define complex schema as text (as mentioned in your issue), although simple stuff like `pa.schema({'hi': 'u1'}) == pa.schema({'hi': pa.uint8()})` works. However, [Fugue](https://fugue-tutorials.readthedocs.io/tutorials/advanced/schema_dataframes.html#schema) provides a way to do so. Maybe we should just use their logic, since it's non-trivial? The requirements for https://github.com/fugue-project/triad also look pretty lightweight, so I don't see an issue with depending on it.

@deepyaman thanks for looking into this! I checked Fugue, and I don't think it supports complex schema structures such as nested list types. For example, I don't see any way to represent a field that has this representation `list_(list_(int64))`.

I have implemented a parser instead to supplant the use of `eval`, and I think it works relatively well.

kedro/extras/datasets/dask/parquet_dataset.py
```diff
@@ -105,9 +126,89 @@ def _load(self) -> dd.DataFrame:
...
+            expr = schema.body[0]
+            self._save_args["schema"] = ParquetDataSet.process_schema_node(expr.value)
+
     def _exists(self) -> bool:
         protocol = get_protocol_and_path(self._filepath)[0]
         file_system = fsspec.filesystem(protocol=protocol, **self.fs_args)
         return file_system.exists(self._filepath)
+
+    @staticmethod
+    def process_schema_node(node):
```
avsolatorio, line 163: By the way, I am unsure if this should be here as a static method, a module-level function, or should be elsewhere.

kedro/extras/datasets/dask/parquet_dataset.py
avsolatorio: I am resolving this @datajoely since I have implemented something that no longer requires the use of `eval`. Thanks!

deepyaman:
> > I don't think PyArrow provides a native way to define complex schema as text (as mentioned in your issue), although simple stuff like `pa.schema({'hi': 'u1'}) == pa.schema({'hi': pa.uint8()})` works. However, [Fugue](https://fugue-tutorials.readthedocs.io/tutorials/advanced/schema_dataframes.html#schema) provides a way to do so. Maybe we should just use their logic, since it's non-trivial? The requirements for https://github.com/fugue-project/triad also look pretty lightweight, so I don't see an issue with depending on it.
> 
> @deepyaman thanks for looking into this! I checked Fugue, and I don't think it supports complex schema structures such as nested list types. For example, I don't see any way to represent a field that has this representation `list_(list_(int64))`.

Ugh, you're right. [I wonder if they'd be open to a simple change to support this?](https://github.com/fugue-project/triad/pull/89)

> I have implemented a parser instead to supplant the use of `eval`, and I think it works relatively well.

I think something to this extent would work, but I'm personally in favor of a solution that doesn't involve maintaining a schema parser, if we can. I assume the subset of people using complex schemas in `dask.ParquetDataSet` via Kedro is very limited, so, while I think we should solve this need, I'd love for it to be as straightforward as possible.

kedro/extras/datasets/dask/parquet_dataset.py
```diff
@@ -105,9 +126,89 @@ def _load(self) -> dd.DataFrame:
...
+
+        if isinstance(schema, dict):
+            for col, field in self._save_args["schema"].items():
+                if isinstance(field, str):
+                    field = ast.parse(field)
+                    expr = field.body[0]
+                    self._save_args["schema"][col] = ParquetDataSet.process_schema_node(
+                        expr.value
+                    )
+        elif isinstance(schema, str) and schema.startswith("schema("):
```
MerelTheisenQB, line 152: I'm guessing `schema.startswith("schema("):` should be `schema.startswith("schema"):` ?

MerelTheisenQB (commented):
Hi @avsolatorio, thank you so much for taking this on. 

I don't know much about dask and or parquet, so I'd like to ask some questions:
1. I see you've added `pyarrow` as a dependency to the dataset. If a user just wants to use the existing functionality they will now to also need to install this dependency. Would it perhaps make sense to have a separate dataset that makes use of `pyarrow`, or do you think in general any user that uses dask/parquet would probably use `pyarrow` as well?
2.  You mention: "Alternatively, a full pyarrow.Schema type may also be provided if desired." How exactly would that work?
3. In the schema a user provides, is it necessary to provide all columns with datatypes? 
4. How would more complex datatypes work? Perhaps a more advanced example would be useful as well. 
I guess I'm mainly trying to understand how many use cases this covers or if it's rather specific.

deepyaman:
> * I see you've added `pyarrow` as a dependency to the dataset. If a user just wants to use the existing functionality they will now to also need to install this dependency. Would it perhaps make sense to have a separate dataset that makes use of `pyarrow`, or do you think in general any user that uses dask/parquet would probably use `pyarrow` as well?

That's a good catch. I don't think it's necessary to add `pyarrow` as a dependency for the Kedro dataset, because the underlying `read_parquet` uses `pyarrow` if installed, else `fastparquet`. `schema` wouldn't be used with `fastparquet`, anyway, so the choice to install `pyarrow` would be up to the user.

avsolatorio:
> 1. I see you've added `pyarrow` as a dependency to the dataset. If a user just wants to use the existing functionality they will now to also need to install this dependency. Would it perhaps make sense to have a separate dataset that makes use of `pyarrow`, or do you think in general any user that uses dask/parquet would probably use `pyarrow` as well?

@MerelTheisenQB you're very much correct. While I think pyarrow is widely used, this is an optional dependency to dask as well. And as @deepyaman pointed out, dask's `to_parquet` API falls back to `fastparquet` when `pyarrow` engine is not installed. In that case, the schema is not used by dask.

> 2. You mention: "Alternatively, a full pyarrow.Schema type may also be provided if desired." How exactly would that work?

So dask's `to_parquet` API accepts the following signature for its schema argument: (i) `None`, (ii) "infer" where the schema is infered from a sample of the data, (iii) a dictionary defining the columns and the intended field types, and (iv) a `pyarrow.Schema` instance.

In the parser that I implemented, it allows for all of these specifications in the `catalog.yml`. Specifically, if the user would like specify the full schema it is possible by doing something like:

```
save_args:
  schema: schema([("col1", int64()), ("col2", list_(list_(int64)))])
```

> 3. In the schema a user provides, is it necessary to provide all columns with datatypes?

Not necessarily. If the passed value is a dictionary of `col:field type` then it will only cast the types for the explicitly defined column data type.

> 4. How would more complex datatypes work? Perhaps a more advanced example would be useful as well.
I guess I'm mainly trying to understand how many use cases this covers or if it's rather specific.

I discovered this issue when I was trying to create a dataset where I had to store in a column a list of tokenized segments of text, e.g., sentences in a paragraph.

I need to use this data in a transformer model that requires the ordering of a sequence of segmented text data. Therefore the data structure that my pipeline output has this form `list_(list_(int64()))`. All is good while doing the processing since these are all represented as `objects`, but it fails when I actually store the data.

Indeed, I'm not sure how many would have this use case. I just thought that it would be good to support the functionalities of the underlying backends since kedro has the `save_args` argument exposed anyway.

I think @deepyaman's PR in Fugue is a very good alternative to decouple this dependency. Nonetheless, Fugue strictly depends on pyarrow. So it is inevitable that pyarrow will become an implicit dependency on kedro.

kedro/extras/datasets/dask/parquet_dataset.py
avsolatorio: Ah, I explicitly did `"schema("` to make sure that the text input is a callable format, something like `"schema(...)"`.

MerelTheisenQB: ah okay that makes sense!

MerelTheisenQB:
Responding again to your answers on my questions: 
1. Okay perfect, let's remove the dependency in that case and rely on the fall back mechanism from Dask.
2. That makes sense, thanks for explaining. I think it would be useful to add some additional documentation with examples of the schema signatures you can add.
3. Great, thanks for clarifying 🙂 

On question 4:
>> 4. How would more complex datatypes work? Perhaps a more advanced example would be useful as well.
>>    I guess I'm mainly trying to understand how many use cases this covers or if it's rather specific.

> I discovered this issue when I was trying to create a dataset where I had to store in a column a list of tokenized segments of text, e.g., sentences in a paragraph.
> 
> I need to use this data in a transformer model that requires the ordering of a sequence of segmented text data. Therefore the data structure that my pipeline output has this form `list_(list_(int64()))`. All is good while doing the processing since these are all represented as `objects`, but it fails when I actually store the data.

> Indeed, I'm not sure how many would have this use case. I just thought that it would be good to support the functionalities of the underlying backends since kedro has the `save_args` argument exposed anyway.

Thanks for explaining your need for this functionality in more detail, your use case definitely makes sense. And I appreciate your thoughts on adding the support in Kedro, it would be good to have this functionality for other users who run into this issue. I think my main concern with this implementation is explainability and maintainability. The first point is mainly around if this solution makes sense to others and is intuitive enough so they would make use of it. I think that can be solved by adding detailed docs and examples. The second point about maintainability echos @deepyaman's thoughts above for maintaining a schema parser, which wouldn't be ideal for the team. 
 
> I think @deepyaman's PR in Fugue is a very good alternative to decouple this dependency. Nonetheless, Fugue strictly depends on pyarrow. So it is inevitable that pyarrow will become an implicit dependency on kedro.

The above sounds okay to me, because `pyarrow` will only be a dependency for users that need this dataset, but not for every kedro framework user.

deepyaman:
@avsolatorio We've discussed this among the maintainer team, and while we think it's important to support this functionality, we'd like to do it in a way that would result in the least maintenance on our end--especially given our current expectation that the use case of complex schema is pretty niche. To this extent, our preference are, in order:

1. Use `fugue.triad` for parsing schema, including complex schema. I've reached out to one of the maintainers to verify if it will be feasible to add the nested schema parsing logic.
2. Use `fugue.triad` for parsing schema, even if it doesn't support nested schema. Maybe we can use it with a little patch applied in the Kedro dataset code, if and until my PR there is merged. :D
3. Implement a custom schema parser.

While we appreciate the work you've done, and think it's really cool, we probably won't go the third route. For now, this would also mean using a custom local dataset (would be true regardless, until next release). That being said, would you be open to updating your PR (or opening a new one) to leverage `fugue.triad` for this?

deepyaman:
> 1. Use `fugue.triad` for parsing schema, including complex schema. I've reached out to one of the maintainers to verify if it will be feasible to add the nested schema parsing logic.

Confirmed with the maintainers that this is just a pending bug, so will try and get it resolved this weekend!

avsolatorio:
> That being said, would you be open to updating your PR (or opening a new one) to leverage `fugue.triad` for this?

@deepyaman I'm very much open to updating the PR to use `fugue.triad` for this functionality. 👌🏽

deepyaman:
> > That being said, would you be open to updating your PR (or opening a new one) to leverage `fugue.triad` for this?
> 
> @deepyaman I'm very much open to updating the PR to use `fugue.triad` for this functionality. 👌🏽

@avsolatorio Great! Triad 0.6.7 was released a few hours ago, and it has the bug fix to support nested schema. :)

avsolatorio:
> @avsolatorio Great! Triad 0.6.7 was released a few hours ago, and it has the bug fix to support nested schema. :)

Hello, @deepyaman. I have already updated the PR to use `triad` for this functionality. Kindly let me know if there are any specific implementation details that I should consider further. Thank you! 😃

datajoely:
This is excellent work @avsolatorio really appreciate the effort here 💪

kedro/extras/datasets/dask/parquet_dataset.py
```diff
@@ -105,8 +129,40 @@ def _load(self) -> dd.DataFrame:
...
+
+        This method converts the Triad-parsed schema into a pyarrow schema.
+        The output directly supports Dask's specifications for providing a schema
+        when saving to a parquet file.
+        """
+        schema = self._save_args.get("schema")
+
+        if isinstance(schema, dict):
+            pa_schema = triad.Schema(
+                ",".join([":".join(i) for i in schema.items()])
```
deepyaman, line 155: You can construct `triad.Schema` directly from a dict; see https://github.com/fugue-project/triad/blob/master/triad/collections/schema.py#L59.

An example with string values:

```python
>>> Schema({"a": "int", "b": "float"}).pyarrow_schema
a: int32
b: float
```

I don't know that you can do more complex stuff, like nested lists, but I think it's a bit unnatural that you define the top level as a dict but the nested lists as strings (as in the test cases).

I think if you've got a power-user, they can just construct the schema string for more complex cases.

kedro/extras/datasets/dask/parquet_dataset.py
```diff
@@ -105,8 +129,40 @@ def _load(self) -> dd.DataFrame:
...
+        columns that are specified will be strictly mapped to a field type. The other
+        unspecified columns, if present, will be inferred from the data.
+
+        This method converts the Triad-parsed schema into a pyarrow schema.
+        The output directly supports Dask's specifications for providing a schema
+        when saving to a parquet file.
+        """
+        schema = self._save_args.get("schema")
+
+        if isinstance(schema, dict):
```
deepyaman, line 153: Do we need to also support the case in which somebody passes a `pa.Schema`? I assume so, since that's the native Dask expectation.

deepyaman (commented):
I think this looks great overall! Took a quick pass and left 2 comments; I'll try and go through it more properly later.

kedro/extras/datasets/dask/parquet_dataset.py
avsolatorio: Oh, I chose to implement it this way because of how `dask.to_parquet` distinguishes between a `dict`-like vs. a `pa.Schema`-like value for its `schema` argument.

If we directly cast a `dict` input into a schema as your snippet shows, `triad` would return a schema containing the `key:value` explicitly declared in the `dict` which would prevent dask from inferring the types for the other fields (if any). It could cause unexpected behaviors since there could be a case where a user simply wants to explicitly override the inferred field types for some of the columns and let dask do the inference for the rest.

In some sense, the `dict` input is a more general case since we can choose to specify the field types only for certain fields or for all the fields.

> it's a bit unnatural that you define the top level as a dict but the nested lists as strings (as in the test cases).
Ah, I was simulating how the parsed arguments from the `catalog.yml` would be passed into the dataset.

Since in the `catalog.yml`, the specification looks like:

```
save_args:
  schema:
    col1: int64
    col2: "[[int64]]"
```

then, I suppose that would be passed as a dict with this structure `{"col1": "int64", "col2": "[[int64]]"}`.

kedro/extras/datasets/dask/parquet_dataset.py
avsolatorio: Yep, that's a good point!

deepyaman:
I'm so sorry about the delay on this. I told myself I'd take a look this weekend, but I got busy with other things. I will try and make sure to take a look at it later this week and push it through!

deepyaman (changes requested):
Sorry for the very overdue response. I've finally had the chance to give this a much more thorough look, and made a suggestion. Let me know if you agree, or if I'm missing something again.

merelcht:
Hi @avsolatorio do you still want to complete this PR, or should someone from the team take-over? We'd like to get all PRs related to datasets to be merged soon now we're moving our datasets code to a different package (see our [medium blog post](https://medium.com/quantumblack/keeping-up-with-kedro-the-latest-developments-in-our-development-workflow-framework-cbcc415eea9c) for more details)

avsolatorio:
> Hi @avsolatorio do you still want to complete this PR, or should someone from the team take-over? We'd like to get all PRs related to datasets to be merged soon now we're moving our datasets code to a different package (see our [medium blog post](https://medium.com/quantumblack/keeping-up-with-kedro-the-latest-developments-in-our-development-workflow-framework-cbcc415eea9c) for more details)

Hello @merelcht , apologies for not being able to get back and close this PR earlier. I'll finalize this and update with @deepyaman 's comments.

merelcht:
Hi @avsolatorio, do you think you'll have time to complete this in the next few days? We are going to do a release early next week and need any dataset changes to be merged by then. Otherwise, we'll close this PR and ask you to re-open it on the new datasets repo when it's ready.

avsolatorio:
> Hi @avsolatorio, do you think you'll have time to complete this in the next few days? We are going to do a release early next week and need any dataset changes to be merged by then. Otherwise, we'll close this PR and ask you to re-open it on the new datasets repo when it's ready.

Hello @merelcht PR updated!

@deepyaman kindly check because I decided not to implement the `infer_schema` argument. How it is implemented now is exactly how Dask expects to accept a schema. That is, passing a `pa.Schema` instance will be passed directly to Dask, and it should specify a full schema for all columns. The implementation here basically forms an interface to handle the arbitrary field typing supported by passing a dictionary to Dask's `to_parquet` method.

The `infer_schema` argument seems redundant since if I don't want the schema to be inferred, I can just pass the full `pa.Schema` instance with all the necessary field type declarations. If I want inference, I can just choose not to specify a schema, or override inference for a few fields by specifying the fields' type declarations in a dictionary.

kedro/extras/datasets/dask/parquet_dataset.py
avsolatorio: Using `pa.Schema` is already supported since whatever value that is passed in `_save_args["schema"]` will also be passed to Dask's `to_parquet`.

kedro/extras/datasets/dask/parquet_dataset.py
```diff
@@ -6,6 +6,7 @@
 
 import dask.dataframe as dd
 import fsspec
+import triad
```
merelcht, line 9: Would it be possible to import `triad` only if the user has provided a schema argument? It would be nice if the library wasn't mandatory if you're not using the schema. But no big deal if that makes things too complicated!

avsolatorio: I tried moving the import inside the `_process_schema` method, but I ended up getting linting errors from the pre-commit suite. 😢

```
************* Module kedro.extras.datasets.dask.parquet_dataset
kedro/extras/datasets/dask/parquet_dataset.py:178:8: C0415: Import outside toplevel (triad) (import-outside-toplevel)
```

merelcht: No worries. Let's leave it as is. I realised it might be confusing for a user if they've already installed the dataset, but then when they add `schema` they need to install another dependency.

merelcht (approved):
Thanks for this contribution and your patience with our reviews @avsolatorio 🙏 ⭐  
I especially like the clear doc-strings you've added to explain the schema functionality.

kedro/extras/datasets/dask/parquet_dataset.py
```diff
@@ -128,8 +152,63 @@ def _load(self) -> dd.DataFrame:
...
+                triad.Schema(
+                    ",".join(
+                        [
+                            ":".join([k, v])
+                            for k, v in schema.items()
+                            if isinstance(v, str)
+                        ]
+                    )
+                )
+            )
```
jmholzer, line 202: It's already a nice implementation, but we could make this a bit cleaner and shorter by using a generator expression instead of a list and using an f-string instead of `":".join(...)`.

See my other comment though, I'm not sure this way of building the schema is the most optimal one.

```suggestion
            triad_schema.update(
                triad.Schema(
                    ",".join(f"{k}:{v}" for k, v in schema.items()
                    if isinstance(v, str))
                )
            )
```

kedro/extras/datasets/dask/parquet_dataset.py
```diff
@@ -128,8 +152,63 @@ def _load(self) -> dd.DataFrame:
...
+            triad_schema = triad.Schema(
+                {k: v for k, v in schema.items() if not isinstance(v, str)}
+            )
+
+            # Handle the schema keys that are represented as string and add them to the triad schema
+            triad_schema.update(
+                triad.Schema(
+                    ",".join(
+                        [
+                            ":".join([k, v])
```
deepyaman, line 196:
```suggestion
                            f"{k}:{v}"
```

Nit: I think this should be faster (not terribly important), but it also feels more readable to me. `k + ":" + v` is also fine, I guess.

deepyaman (approved):
@avsolatorio Thank you for the updates! I think this makes sense, and the way you've documented when inference happens (i.e. if you specify a dict) is great.

Look forward to seeing some more people use this and see how they like the added power! :)

kedro/extras/datasets/dask/parquet_dataset.py
jmholzer, line 202: It would be cleaner to build the schema in one loop, would something like the following work?

```suggestion
            triad_schema_dict = {}
            for k, v in schema.items():
                if isinstance(v, str):
                    ...
                else:
                    ...
            triad_schema = triad.Schema(triad_schema_dict)
```

jmholzer (approved):
Great work! Thanks for the contribution 😃

Two minor code optimisation suggestions from me, nothing blocking.

avsolatorio:
@merelcht @deepyaman @jmholzer, thanks for all the suggestions! Apologies that I was not able to work on the final set of comments. I was traveling, and I wasn't able to find time to implement the suggestions. Nonetheless, great to see this PR merged! 🥳

avsolatorio:
@merelcht btw, would this PR qualify for this: https://github.com/kedro-org/kedro/issues/2050#issuecomment-1326513789. Would be super to be added on this list https://github.com/kedro-org/kedro/pull/2064#issuecomment-1328806327! 🔋

merelcht:
> @merelcht btw, would this PR qualify for this: [#2050 (comment)](https://github.com/kedro-org/kedro/issues/2050#issuecomment-1326513789). Would be super to be added on this list [#2064 (comment)](https://github.com/kedro-org/kedro/pull/2064#issuecomment-1328806327)! 🔋

Absolutely! Let me create a new PR to add you 🙂

avsolatorio:
> Oh actually you're already there 😄 https://github.com/kedro-org/kedro/blob/main/RELEASE.md#contributions-from-the-kedroid-community

Yay! Thank you so much! 🥳

PR #1751, opened 2022-08-03
Update key documentations to show richly formatted logs

### LINKED ISSUE #1683: Update key docs to show richly formatted logs
Now we have beautiful richly formatted logs, we should update our docs to show it off!

The key places to update are https://kedro.readthedocs.io/en/stable/get_started/example_project.html and 
https://kedro.readthedocs.io/en/stable/tutorial/create_pipelines.html.

The best way to do this is probably to capture the logs, save them to HTML and then insert into sphinx somehow. This way we don't need to embed pictures in our docs and it remains easily maintainable. You can do this with hooks using the snippet given below. However, maybe it would result in overly long/difficult to navigate .md files (can the HTML be injected from a separate file?).

Alternatively, you could just capture the plaintext logs. It would still be nice to update the docs since even with no formatting they are more nicely arranged into columns than the current ones.

```
from rich import reconfigure
from rich import get_console

class Hooks:
    @hook_impl
    def after_context_created(self):
        reconfigure(record=True)

    @hook_impl
    def after_pipeline_run(self):
        get_console().save_html("logs.html")
```

noklam:

## Description

Fix #1683 

The example highlights changes with `rich`
1. `kedro run` logging
2.  The nicer formatted traceback with failed pipeline

## Development notes

https://kedro.readthedocs.io/en/stable/get_started/example_project.html and
https://kedro.readthedocs.io/en/stable/tutorial/create_pipelines.html.
* The color log could be added as HTML but it's quite difficult to maintain the HTML tag directly.
  *  The logs are slightly optimized by removing some warning etc.

This hook would save the `kedro run` output to a separate file.
* For HTML - you get a self-contained HTML page with the right CSS style
* For text - you get a plain text log without any of the colorings

```python
from rich import reconfigure
from rich import get_console

class Hooks:
    @hook_impl
    def after_context_created(self):
        reconfigure(record=True)

    @hook_impl
    def after_pipeline_run(self):
        # get_console().save_html("logs.html")  # Would save the HTML instead
        get_console().save_text("logs.text")  # Save the text version 
```

Equivalently you can just copy from the terminal and it works perfectly fine too.

## Side-note:
It's a bit manual to update the doc at the moment, partly due to the `spaceflights` tutorial being a long tutorial, the `starter` only give you the final stages of the project. We don't have any checkpoints for individual pages, so I actually have to start from an empty project and followed a few pages, and copy-paste stuff to run and get the logs.

I also looked into the possibility of keeping them as a separate HTML file, but that isn't enough, and have to mess with the CSS.

Putting the HTML directly into the Markdown file would render properly, it's just a bit painful to maintain that in the file. Particularly because I have to remove some of the logs manually to remove the noise from the Warning, hyperlink that link to local files, etc, and I don't want to repeat that 6 times.

### Example of HTML file - you can save this as a standalone .html file and open it
```html
<!DOCTYPE html>
<head>
<meta charset="UTF-8">
<style>
.r1 {color: #7fbfbf; text-decoration-color: #7fbfbf}
.r2 {color: #800000; text-decoration-color: #800000}
.r3 {color: #800080; text-decoration-color: #800080}
.r4 {color: #ff00ff; text-decoration-color: #ff00ff}
.r5 {color: #7f7f7f; text-decoration-color: #7f7f7f}
.r6 {color: #008080; text-decoration-color: #008080; font-weight: bold}
.r7 {font-weight: bold}
.r8 {color: #800080; text-decoration-color: #800080; font-weight: bold}
.r9 {color: #008000; text-decoration-color: #008000}
.r10 {color: #000080; text-decoration-color: #000080}
.r11 {color: #808000; text-decoration-color: #808000}
body {
    color: #000000;
    background-color: #ffffff;
}
</style>
</head>
<html>
<body>
    <code>
        <pre style="font-family:Menlo,'DejaVu Sans Mono',consolas,'Courier New',monospace"><span class="r1">[08/09/22 10:54:42] </span><span class="r10">INFO    </span> Model has accuracy of <span class="r6">0.933</span> on test data.                                        <a class="r5" href="file:///Users/Nok_Lam_Chan/GitHub/kedro/demo-project/src/demo_project/nodes.py">nodes.py</a><span class="r5">:</span><a class="r5" href="file:///Users/Nok_Lam_Chan/GitHub/kedro/demo-project/src/demo_project/nodes.py#74">74</a>
<span class="r1">                    </span><span class="r10">INFO    </span> Saving data to <span class="r9">&#x27;metrics&#x27;</span> <span class="r7">(</span>MetricsDataSet<span class="r7">)</span><span class="r11">...</span>                             <a class="r5" href="file:///Users/Nok_Lam_Chan/miniconda3/lib/python3.9/site-packages/kedro/io/data_catalog.py">data_catalog.py</a><span class="r5">:</span><a class="r5" href="file:///Users/Nok_Lam_Chan/miniconda3/lib/python3.9/site-packages/kedro/io/data_catalog.py#382">382</a>
<span class="r1">                    </span><span class="r10">INFO    </span> Completed <span class="r6">3</span> out of <span class="r6">3</span> tasks                                           <a class="r5" href="file:///Users/Nok_Lam_Chan/miniconda3/lib/python3.9/site-packages/kedro/runner/sequential_runner.py">sequential_runner.py</a><span class="r5">:</span><a class="r5" href="file:///Users/Nok_Lam_Chan/miniconda3/lib/python3.9/site-packages/kedro/runner/sequential_runner.py#85">85</a>
<span class="r1">                    </span><span class="r10">INFO    </span> Pipeline execution completed successfully.                                      <a class="r5" href="file:///Users/Nok_Lam_Chan/miniconda3/lib/python3.9/site-packages/kedro/runner/runner.py">runner.py</a><span class="r5">:</span><a class="r5" href="file:///Users/Nok_Lam_Chan/miniconda3/lib/python3.9/site-packages/kedro/runner/runner.py#89">89</a>
</pre>
    </code>
</body>
</html>
```

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1751"></a>


SajidAlamQB (approved):
Awesome, thank you for this! 🌟

AntonyMilneQB:
@noklam would you mind writing a bit more about how you did this, the other things you tried and what was good/bad about them? 

e.g. did the HTML logs work ok? Did you just do this by copying and pasting from terminal or capturing the output using hooks? (I'm curious because at some point in the future we might want to change the way we do file logging to something more like that)

noklam:
@AntonyMilneQB Sure, update this in the description.

I actually also experimented with `nbdev` version 2 a bit, which now uses the Quarto thing RStudio released. Particular because I like the idea that documentation should be runnable, instead of a separate generated file that requires copy & paste (which is inevitably fragile & painful to keep in sync). In the nbdev world, docs are just code & output from the notebook.

I found this concept is quite powerful and should improve documentation quality & maybe even tests. However, this is not something we can adopt straight away for sure, but I think there are good lessons to learn. I can share more or do a quick demo of how this work if this is some interest about this.

I put up a sample here
https://noklam.github.io/nbdev_kedro/card.html

AntonyMilneQB (approved):
Perfect, thank you for the write up @noklam, super helpful and completely makes sense. That HTML log would indeed be a nightmare to maintain - much better to have the plain text one.

I'd definitely be interested in learning more about nbdev and runnable documentation. Tangentially related, I've always been a fan of the idea of [docstrings forming tests](https://docs.python.org/3/library/doctest.html) (though in practice I've found it can be quite fiddly so not necessarily worth the effort).

PR #1759, opened 2022-08-04
Update description for `pipeline` arg to `run` CLI

deepyaman:
## Description

The pipeline name could refer to a combination of modular pipelines, or a pipeline that isn't a "modular pipeline". It really pulls from the registry instead.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1759"></a>


deepyaman:
Actually, why don't see use `default="__default__"` when constructing the CLI option, rather than defaulting to `None` and then `name = pipeline_name or "__default__"`? Seems circuitous, but maybe there's a good reason.

AntonyMilneQB (approved):

AntonyMilneQB:
> Actually, why don't see use `default="__default__"` when constructing the CLI option, rather than defaulting to `None` and then `name = pipeline_name or "__default__"`? Seems circuitous, but maybe there's a good reason.

I think there's no good reason for this. We follow this pattern of just using `None` for a default quite a bit in the kedro CLI, mistakenly I think.

MerelTheisenQB (approved):

noklam (approved):

Issue #1742, opened 2022-08-01
Rationalise data catalog examples in documentation


merelcht:
### Description
In the section on using the `DataCatalog` with the yaml api, we've listed a lot of examples: https://kedro.readthedocs.io/en/stable/data/data_catalog.html#use-the-data-catalog-with-the-yaml-api. However, there are no anchors to any of these examples, so it's very hard to link users to a specific solution.

### Implementation

- [x] https://github.com/kedro-org/kedro/pull/1770
- [ ] Create a separate page for these examples altogether or alternatively, only keep some of the example in Data Catalog to showcase what Kedro can do, but move the advanced examples to the corresponding API Docs.

noklam:
I think "Data Catalog" should show case what Kedro can do, showing examples for saving things in remote storage, but we don't need all the combination of DataSet + Remote Storage type.

This should make the page itself shorter, the more repeated example can be moved to the specific DataSet API docs instead.

stichbury:
This issue, plus #1743 and #1762 all seem to point to improving the API docs and streamlining the DataCatalog docs. I agree with @noklam that it would make sense to extract some of the examples to put them in API docs (or elsewhere).

stichbury:
Also to add that the page needs a significant rework to simplify it and attempt to make it more discoverable through search as per this discussion https://kedro-org.slack.com/archives/C03RKP2LW64/p1673459017999699

PR #1760, opened 2022-08-04
Add example for using multi-sheet Excel file

### LINKED ISSUE #1700: Document how to read and write a multi-sheet excel
## Introduction

>A high-level, short overview of the problem(s) you are designing a solution for.

We have the implementation, but it is not easy to point to a documented example:

image

## Background

> Provide the reader with the context surrounding the problem(s) you are trying to solve.

This is a common feature and something Kedro can already do.

ankatiyar:
## Description
Resolves #1700 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1760"></a>

### DIFF
--- a/docs/source/data/data_catalog.md
+++ b/docs/source/data/data_catalog.md
@@ -164,7 +164,17 @@ rockets:
     sheet_name: Sheet1
 ```
 
-Example 7: Saves an image created with Matplotlib on Google Cloud Storage
+Example 7: Loads a multisheet excel file from a local file system
+
+```yaml
+trains:
+  type: pandas.ExcelDataSet
+  filepath: data/02_intermediate/company/trains.xlsx
+  load_args:
+    sheet_name: [Sheet1, Sheet2, Sheet3]
+```
+
+Example 8: Saves an image created with Matplotlib on Google Cloud Storage
 
 ```yaml
 results_plot:
@@ -175,7 +185,7 @@ results_plot:
   credentials: my_gcp_credentials
 ```
 
-Example 8: Loads / saves an HDF file on local file system storage, using specified load and save arguments
+Example 9: Loads / saves an HDF file on local file system storage, using specified load and save arguments
 
 ```yaml
 skateboards:
@@ -189,7 +199,7 @@ skateboards:
     dropna: True
 ```
 
-Example 9: Loads / saves a parquet file on local file system storage, using specified load and save arguments
+Example 10: Loads / saves a parquet file on local file system storage, using specified load and save arguments
 
 ```yaml
 trucks:
@@ -206,7 +216,7 @@ trucks:
     partition_on: [name]
 ```
 
-Example 10: Loads / saves a Spark table on S3, using specified load and save arguments
+Example 11: Loads / saves a Spark table on S3, using specified load and save arguments
 
 ```yaml
 weather:
@@ -222,7 +232,7 @@ weather:
     header: True
 ```
 
-Example 11: Loads / saves a SQL table using credentials, a database connection, using specified load and save arguments
+Example 12: Loads / saves a SQL table using credentials, a database connection, using specified load and save arguments
 
 ```yaml
 scooters:
@@ -236,7 +246,7 @@ scooters:
     if_exists: replace
 ```
 
-Example 12: Loads an SQL table with credentials, a database connection, and applies a SQL query to the table
+Example 13: Loads an SQL table with credentials, a database connection, and applies a SQL query to the table
 
 ```yaml
 scooters_query:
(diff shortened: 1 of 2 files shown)


docs/source/data/data_catalog.md
```diff
@@ -164,7 +164,17 @@ rockets:
     sheet_name: Sheet1
 ```
 
-Example 7: Saves an image created with Matplotlib on Google Cloud Storage
+Example 7: Loads a multisheet excel file from a local file system
```
deepyaman, line 167:
```suggestion
Example 7: Loads a multi-sheet Excel file from a local file system
```
:)

ankatiyar: Made the change in 3f493a1921cf72bd3f81bd2f49f4ec3bedecc304, thanks!

noklam (approved):
Congrats on your first PR⭐️

Side notes: 
Follow up on previous discussion about DataSet API Docs should be the 1st entry point for people looking up examples, I think we should do the followings:

1. Remove these number, i.e. Example X, make them a sub-header like `## Save-an-image-xxxxx`, so we can link these example directly in the doc, and also better SEO I think?
2. Only keep some of the example in `Data Catalog` to showcase what Kedro can do, but leave the advance example in the Corresponding API Doc.
  * We have examples saving CSV on S3, matplotlib on GCP, excel on GCP, which I don't think add much value here, it's enough to show 1 example of GCP/S3.
  * Examples like saving HDF file doesn't worth the highlight in this page in my opinion, there could be more.

datajoely:
Can I ask that we include the example how to save a multi-sheet as well? The instructions to do so are buried in in the DataSet docstring: https://github.com/kedro-org/kedro/blob/59064403e2590c992dceb51942f7dd9f6c97898a/kedro/extras/datasets/pandas/excel_dataset.py#L104

MerelTheisenQB:
> Side notes: Follow up on previous discussion about DataSet API Docs should be the 1st entry point for people looking up examples, I think we should do the followings:
> 
> 1. Remove these number, i.e. Example X, make them a sub-header like `## Save-an-image-xxxxx`, so we can link these example directly in the doc, and also better SEO I think?
> 2. Only keep some of the example in `Data Catalog` to showcase what Kedro can do, but leave the advance example in the Corresponding API Doc.

Yes we should, but let's do that as separate tasks to keep this PR focussed on one thing. We've got this issue for task 1: https://github.com/kedro-org/kedro/issues/1742. I'll update it so it covers point 2 as well.

kedro/extras/datasets/pandas/excel_dataset.py
```diff
@@ -65,6 +65,32 @@ class ExcelDataSet(
...
+
+    .. code-block:: yaml
+
+        >>> trains:
+        >>>   type: pandas.ExcelDataSet
+        >>>   filepath: data/02_intermediate/company/trains.xlsx
+        >>>   load_args:
+        >>>     sheet_name: [Sheet1, Sheet2, Sheet3]
+
+    Example multi-sheet excel file using Python API:
```
datajoely, line 78: So (as we discovered working this out together) the important piece of info that user needs to know is that to save a multi-sheet excel the user **doesn't** need to do anything fancy in `save_args`. Would you mind putting a comment explaining that.

Once again - really good work and thanks for the contribution!

ankatiyar: I've added the comment in f723f0de9e68e788f33f230ca4acc7121ad39219. Thank you!

kedro/extras/datasets/pandas/excel_dataset.py
```diff
@@ -65,6 +65,36 @@ class ExcelDataSet(
         >>> reloaded = data_set.load()
         >>> assert data.equals(reloaded)
 
+    Note that using ``save_args`` is not required when saving multi-sheet excel files.
```
datajoely, line 68:
```suggestion
    Note: To save a multi-sheet excel files, no specials `save_args` are required.
```

kedro/extras/datasets/pandas/excel_dataset.py
```diff
@@ -65,6 +65,36 @@ class ExcelDataSet(
...
+    A dict of ``pd.DataFrame`` is saved as a multi-sheet excel file by default.
```
datajoely, line 69:
```suggestion
    Instead, return a dictionary of ``Dict[str, pd.DataFrame]`` where the string keys are your sheet names.
```

datajoely (changes requested):
Super close, slightly different wording suggested. Then I think we're good to go!

kedro/extras/datasets/pandas/excel_dataset.py
ankatiyar: Done!

datajoely (approved):
🔥 awesome work @ankatiyar

yetudada (approved):
Congratulations on your first PR @ankatiyar! 🥳 And thank you for helping our users with this! We get this question a lot. 

I'm approving this but I just note that we use `excel` (lowercase) when it should be `Excel` (uppercase). This actually correct based on what we have in our documentation at the moment. 

Can you create a separate GitHub issue to address this as another quick-fix PR because I just checked the rest of the documentation and it looks like we've made that mistake in other places too?

PR #1761, opened 2022-08-04
Improve IPython extension error handling

### LINKED ISSUE #1590: Improve Jupyter workflow when configuration is broken
Using the spaceflights demo on 0.18.1 if I open a `kedro jupyter notebook` session and write `catalog.list()` I will get the following:

image

If I break the `catalog.yml` file so that it is unparsable and run `kedro catalog list` (or any other command like `kedro run` I will get a helpful parse error in the terminal: 

image

However, if I start a `kedro jupyter notebook` session in this state, I get no warning or messaging that Kedro is unable to create the `catalog`, `session` and other global variables. As a user I get this confusing state where the `catalog` is no longer available to me:

image

I think novice users can get stuck here with no direction as to why the catalog is not available.

## Possible Implementation

Surface warning to the jupyter context, perhaps even make a dummy `catalog`, `context`, `session` objects available that expose the log message when called...

AntonyMilneQB:
## Description
Resolves #1590.

> **Note**. I still need to double check my tests cover everything they should, but this is ready for review.

There were two problems with how the IPython extension was handling errors:
1. The catching of `ImportError` and `ModuleNotFound` error was meant to notify the user about when they didn't have Kedro installed but was too broad. The same error (`Kedro appears not to be installed in your current environment`) would also be raised when other dependencies (e.g. pandas) weren't installed, which was very confusing
2. All other exceptions were being swallowed up and replaced with `WARNING  Kedro extension was registered but couldn't find a Kedro project`. This is also very confusing because e.g. if you have a syntax error in your catalog.yml file you would have no idea what the problem was and instead think you were just in the wrong directory

> **Warning**. Important lesson. This kind of behaviour is exactly what the pylint `broad-except` is trying to guard against. We should not swallow up exceptions unless we really mean to. In this case, it makes it much harder for users to debug what's going on.

## What happens now?
If the Kedro project can't be found, a warning is emitted saying so:
image

Any other exception will be shown as is.
e.g. dataset dependency not installed:
image

e.g. catalog.yml syntax error:
image

## Slight catch
When using `kedro ipython/jupyter`, these error messages appear on the terminal rather than in the Jupyter notebook itself. There's no easy way to push them to the Jupyter notebook. However, they can be reproduced there by executing `%reload_kedro`. I've added a note to the documentation to explain this. In the case of managed Jupyter instances this won't be a problem at all because you'll need to execute `%load_ext` in your notebook anyway, so the error will be surfaced there automatically.

This is perfectly satisfactory for now I think. I also considered:
* making a new `startup_error` variable which captures the exception and is available in Jupyter notebooks. This is actually what we used to do before the IPython extension was created and we could put it back. But all the cases it solves are already covered by running `%reload_kedro` in your notebook and it doesn't seem any more discoverable than that
* populate the `catalog`, `context`, etc. variables with `Exception`s by default, so that if they're not available then displaying them will explain why. This seems like overkill for now but could be done in future if people find the current system not clear enough. From observing what current user problems are, I think what we have here solves 99% of the problems.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1761"></a>

### DIFF
--- a/docs/source/development/commands_reference.md
+++ b/docs/source/development/commands_reference.md
@@ -491,18 +491,14 @@ To start an IPython shell:
 kedro ipython
 ```
 
-Every time you start or restart a notebook kernel, a startup script (`<project-root>/.ipython/profile_default/startup/00-kedro-init.py`) will add the following variables in scope:
+The [Kedro IPython extension](../tools_integration/ipython.md) will make the following variables available in your IPython or Jupyter session:
 
-- `context`: An instance of `kedro.framework.context.KedroContext` class or custom context class extending `KedroContext` if one was set to `CONTEXT_CLASS` in `settings.py` file (further details of how to use `context` can be found [in the IPython documentation](../tools_integration/ipython.md))
-- `startup_error` (`Exception`)
-- `catalog`
+* `catalog` (type `DataCatalog`): [Data Catalog](../data/data_catalog.md) instance that contains all defined datasets; this is a shortcut for `context.catalog`
+* `context` (type `KedroContext`): Kedro project context that provides access to Kedro's library components
+* `pipelines` (type `Dict[str, Pipeline]`): Pipelines defined in your [pipeline registry](../nodes_and_pipelines/run_a_pipeline.md#run-a-pipeline-by-name)
+* `session` (type `KedroSession`): [Kedro session](../kedro_project_setup/session.md) that orchestrates a pipeline run
 
-To reload these variables at any point in your notebook (e.g. if you updated `catalog.yml`) use the [line magic](https://ipython.readthedocs.io/en/stable/interactive/magics.html#line-magics) `%reload_kedro`, which can be also used to see the error message if any of the variables above are undefined.
-
-If you get an error message `Module ``<module_name>`` not found. Make sure to install required project dependencies by running ``pip install -r requirements.txt`` first.` when running any of those commands, it indicates that some Jupyter or IPython dependencies are not installed in your environment. To resolve this you will need to do the following:
-
-1. Make sure the corresponding dependency is present in `src/requirements.txt`
-2. Run [`pip install -r src/requirements.txt`](#install-all-package-dependencies) command from your terminal
+To reload these variables (e.g. if you updated `catalog.yml`) use the `%reload_kedro` line magic, which can be also used to see the error message if any of the variables above are undefined. 
 
 ##### Copy tagged cells
 To copy the code from [cells tagged](https://jupyter-notebook.readthedocs.io/en/stable/changelog.html#cell-tags) with a `node` tag into Python files under `src/<package_name>/nodes/` in a Kedro project:
--- a/docs/source/development/commands_reference.md
+++ b/docs/source/development/commands_reference.md
@@ -498,7 +498,7 @@ The [Kedro IPython extension](../tools_integration/ipython.md) will make the fol
 * `pipelines` (type `Dict[str, Pipeline]`): Pipelines defined in your [pipeline registry](../nodes_and_pipelines/run_a_pipeline.md#run-a-pipeline-by-name)
 * `session` (type `KedroSession`): [Kedro session](../kedro_project_setup/session.md) that orchestrates a pipeline run
 
-To reload these variables (e.g. if you updated `catalog.yml`) use the `%reload_kedro` line magic, which can be also used to see the error message if any of the variables above are undefined. 
+To reload these variables (e.g. if you updated `catalog.yml`) use the `%reload_kedro` line magic, which can also be used to see the error message if any of the variables above are undefined. 
 
 ##### Copy tagged cells
 To copy the code from [cells tagged](https://jupyter-notebook.readthedocs.io/en/stable/changelog.html#cell-tags) with a `node` tag into Python files under `src/<package_name>/nodes/` in a Kedro project:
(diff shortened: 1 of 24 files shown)


tests/extras/extensions/test_ipython.py
```diff
@@ -62,18 +56,18 @@ def my_register_pipeline():
         mocker.patch(
             "kedro.framework.startup.bootstrap_project", return_value=fake_metadata
         )
-        mock_line_magic = mocker.MagicMock()
+        mock_line_magic = mocker.Mock()
```
noklam, line 59: Just curious is this necessary? Or we just prefer `Mock` in general. I think we use `MagicMock` quite a lot but without actually needing it in some tests.

docs/source/development/commands_reference.md
```diff
@@ -491,18 +491,14 @@ To start an IPython shell:
...
+* `session` (type `KedroSession`): [Kedro session](../kedro_project_setup/session.md) that orchestrates a pipeline run
 
-To reload these variables at any point in your notebook (e.g. if you updated `catalog.yml`) use the [line magic](https://ipython.readthedocs.io/en/stable/interactive/magics.html#line-magics) `%reload_kedro`, which can be also used to see the error message if any of the variables above are undefined.
-
-If you get an error message `Module ``<module_name>`` not found. Make sure to install required project dependencies by running ``pip install -r requirements.txt`` first.` when running any of those commands, it indicates that some Jupyter or IPython dependencies are not installed in your environment. To resolve this you will need to do the following:
-
-1. Make sure the corresponding dependency is present in `src/requirements.txt`
-2. Run [`pip install -r src/requirements.txt`](#install-all-package-dependencies) command from your terminal
+To reload these variables (e.g. if you updated `catalog.yml`) use the `%reload_kedro` line magic, which can be also used to see the error message if any of the variables above are undefined.
```
MerelTheisenQB, line 501: "be also" reads a bit clunky imo, but I trust your final judgement 🙂 
```suggestion
To reload these variables (e.g. if you updated `catalog.yml`) use the `%reload_kedro` line magic, which can also be used to see the error message if any of the variables above are undefined. 
```

noklam (approved):
Great catch! I think this is a small but significant change. I just fall into this trap this morning😅

> When using kedro ipython/jupyter, these error messages appear on the terminal rather than in the Jupyter notebook itself. There's no easy way to push them to the Jupyter notebook. However, they can be reproduced there by executing %reload_kedro. I've added a note to the documentation to explain this. In the case of managed Jupyter instances this won't be a problem at all because you'll need to execute %load_ext in your notebook anyway, so the error will be surfaced there automatically.

I also checked this recently and couldn't find a way to do it, which is a bit annoying as almost 99% of people won't look at the terminal when they are using notebook.

requirements.txt
```diff
@@ -10,7 +10,6 @@ importlib_metadata>=3.6  # The "selectable" entry points were introduced in `imp
 jmespath>=0.9.5, <1.0
 pip-tools~=6.5
 pluggy~=1.0.0
-python-json-logger~=2.0
```
MerelTheisenQB, line 13: What was this requirement used for?

kedro/extras/extensions/ipython.py
```diff
@@ -95,12 +92,11 @@ def load_ipython_extension(ipython):
 
     default_project_path = _find_kedro_project(Path.cwd())
 
-    try:
-        reload_kedro(default_project_path)
-    except (ImportError, ModuleNotFoundError):
-        logger.error("Kedro appears not to be installed in your current environment.")
```
MerelTheisenQB, line 101: It makes sense that these have been removed, but I was just wondering if the user still gets a useful message if kedro actually isn't installed?

MerelTheisenQB (commented):
Great PR description as always! ⭐ I've got some questions, but the overall changes look good to me. It's slightly annoying the error messages don't show in the notebook, but I think your comment in the docs is very clear and I agree we can always populate the variables with `Exception`s if users don't find the error handling clear.

tests/extras/extensions/test_ipython.py
AntonyMilneQB: Good question - I did a quick survey of our use of `MagicMock` vs. `Mock` and found that nearly always it's `Mock` apart from in this file. There's no need for the extra capability of `MagicMock` here so I changed them all.

docs/source/development/commands_reference.md
AntonyMilneQB: Definitely agree, great point!

requirements.txt
AntonyMilneQB: It was for an old style of logging format that existed before we switched to rich logging. It should have been removed before, just we missed it.

kedro/extras/extensions/ipython.py
MerelTheisenQB: Ah yes that's great!

MerelTheisenQB (approved):
Really nice improvement! 👍

kedro/extras/extensions/ipython.py
```diff
@@ -39,16 +36,17 @@ def reload_kedro(
 ):
     """Line magic which reloads all Kedro default variables.
     Setting the path will also make it default for subsequent calls.
-
-
     """
+    # If a path is provided, set it as default for subsequent calls
```
AntonyMilneQB, line 40:
```suggestion
```

kedro/extras/extensions/ipython.py
```diff
@@ -39,16 +36,17 @@ def reload_kedro(
...
+    from IPython.core.magic import needs_local_scope, register_line_magic
+
     from kedro.framework.cli import load_entry_points
     from kedro.framework.project import LOGGING  # noqa # pylint:disable=unused-import
     from kedro.framework.project import configure_project, pipelines
     from kedro.framework.session import KedroSession
     from kedro.framework.startup import bootstrap_project
 
-    # If a path is provided, set it as default for subsequent calls
     global default_project_path
```
AntonyMilneQB, line 50:
```suggestion
    # If a path is provided, set it as default for subsequent calls
    global default_project_path
```

Issue #399, opened 2020-06-02
[KED-1740] Name-spaced Parameters


benhorsburgh:
## Description

First of all - I love Kedro 0.16.x namespacing! 

When I create a new modular pipeline, I can very easily re-use this across different name spaces, which is intuitive to use thanks to the new conf sub-directories, and pipeline structure. However, the name-spacing of parameters can only be achieved using messy work arounds.

Let me explain using an example.

Let's say I create `kedro pipeline create clean_timestamps` as a modular pipeline. In this, I create the following pipeline:

```python
def create_pipeline(**kwargs):
    return Pipeline([
        node(
            func=round_timestamps,
            inputs=dict(params="params:clean_timestamps_params", data="data"),
            outputs=dict(rounded="rounded"),
            name="round_timestamps"
        )
    ])
```

In my `conf/base/pipelines/clean_timestamps/catalog.yml` file I can make the datasets very obvious, along with name-spaced versions of my datasets.

```yaml
# no namespace
#data:
#  type: MemoryDataSet
#
#rounded:
#  type: MemoryDataSet

# hour namespace
hour.data:
  type: pandas.CSVDataSet
  filepath: data/01_raw/hour.csv

# minute namespace
minute.data:
  type: pandas.CSVDataSet
  filepath: data/01_raw/minute.csv

minute.rounded:
  type: pandas.CSVDataSet
  filepath: data/05_model_input/minute_rounded.csv
```

This gives me very clean and clear flexibility to keep the data that I need, and set up new namespaces. Using this, I can create the default and namespaced pipelines very easily.

```python
clean_timestamps_pipeline = clean_timestamps.create_pipeline()

hour_clean_timestamps_pipeline = pipeline(
    clean_timestamps_pipeline,
    namespace="hour"
)

minute_clean_timestamps_pipeline = pipeline(
    clean_timestamps_pipeline,
    namespace="minute"
)
```

and everything magically works because of how my config has been set up.

-----------

Now if we try to do the same with parameters! Intuitively, I should set up my `conf/base/pipelines/clean_timestamps/parameters.yml` in exactly the same way, which might look like:

```yaml
# no namespace
clean_timestamps_params:
  round_frequency: "1min"
  round_col: "dteday"

# hour namespace
hour.clean_timestamps_params:
  round_frequency: "1h"
  round_col: "dteday"

# minute namespace
minute.clean_timestamps_params:
    round_frequency: "1min"
    round_col: "dteday"
```

[...]

PR #1765, opened 2022-08-08
Enable dotted keys in params passed to `kedro run`

deepyaman:
## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1765"></a>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -14,7 +14,8 @@
 ## Major features and improvements
 
 ## Bug fixes and other changes
-* Use default `False` value for rich logging `set_locals`, to make sure credentials and other sensitive data isn't shown in logs.
+* Set default `False` value for rich logging `set_locals`, to make sure credentials and other sensitive data isn't shown in logs.
+* Enabled dotted keys in params passed to `kedro run`.
 
 ## Upcoming deprecations for Kedro 0.19.0
 
--- a/kedro/framework/cli/utils.py
+++ b/kedro/framework/cli/utils.py
@@ -1,4 +1,5 @@
 """Utilities for use with click."""
+import csv
 import difflib
 import logging
 import re
@@ -428,9 +429,8 @@ def _split_params(ctx, param, value):
                 f"cannot be an empty string."
             )
         value = item[1].strip()
-        result = _update_value_nested_dict(
-            result, _try_convert_to_numeric(value), key.split(".")
-        )
+        [path] = list(csv.reader([key], delimiter="."))
+        result = _update_value_nested_dict(result, _try_convert_to_numeric(value), path)
     return result
 
 
--- a/tests/framework/cli/test_cli.py
+++ b/tests/framework/cli/test_cli.py
@@ -647,8 +647,17 @@ class TestRunCommand:
             ("foo.nested:bar", {"foo": {"nested": "bar"}}),
             ("foo.nested:123.45", {"foo": {"nested": 123.45}}),
             (
-                "foo.nested_1.double_nest:123.45,foo.nested_2:1a",
-                {"foo": {"nested_1": {"double_nest": 123.45}, "nested_2": "1a"}},
+                "foo.nested_1.double_nested:123.45,foo.nested_2:1a",
+                {"foo": {"nested_1": {"double_nested": 123.45}, "nested_2": "1a"}},
+            ),
+            (
+                'foo.nested_1.double_nested:123.45,foo."nested_2.not_double_nested":1a',
+                {
+                    "foo": {
+                        "nested_1": {"double_nested": 123.45},
+                        "nested_2.not_double_nested": "1a",
+                    }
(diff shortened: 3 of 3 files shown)


datajoely (commented):
Yes! This inconsistency has annoyed me forever

AntonyMilneQB:
Hmmm, interesting idea. Given that we're planning to switch to `OmegaConf`, I would like to understand how this works for them? See https://omegaconf.readthedocs.io/en/2.2_branch/usage.html#from-a-dot-list and https://hydra.cc/docs/advanced/override_grammar/basic/ (won't be possible immediately, but might help understand possible future capabilities better)

If it's something that `OmegaConf` doesn't support then we should be very hesitant about adding this feature ourselves. Ideally we just want to delegate all the params overriding logic to them and not modify it ourselves.

Let me just note this down here since I always need to remind myself of the discussion and hunt down the links - https://github.com/quantumblacklabs/private-kedro/issues/965 (beginning of _Alternatives & Tradeoffs_ section) and https://github.com/kedro-org/kedro/issues/399#issuecomment-638087120 for original discussion of `.` being used for namespacing params.

AntonyMilneQB:
After a quick search, it looks like this is not possible in OmegaConf: https://stackoverflow.com/questions/64864518/how-to-escape-the-when-using-dot-notation-commandline-arguments

Closely related, they had a proposal for "compact key support" in https://github.com/omry/omegaconf/issues/152, which was then aborted in https://github.com/omry/omegaconf/issues/332. We should try to align the way we handle parameters with this as closely as possible I think. Possibly this would mean that we no longer encourage a parameter with name `a.b` to be accessed through `params:a.b` at all? Not sure.

deepyaman:
> Closely related, they had a proposal for "compact key support" in [omry/omegaconf#152](https://github.com/omry/omegaconf/issues/152), which was then aborted in [omry/omegaconf#332](https://github.com/omry/omegaconf/issues/332). We should try to align the way we handle parameters with this as closely as possible I think. Possibly this would mean that we no longer encourage a parameter with name `a.b` to be accessed through `params:a.b` at all? Not sure.

Compact keys are a bit different; a compact key has the same semantic meaning as the fully-expanded, nested version, whereas here we're talking about enabling a key name to have dots in it.

https://github.com/omry/omegaconf/issues/152#issuecomment-654332429 is more relevant to the use case that inspired this.

AntonyMilneQB:
Yeah, understood. I think the problem is that in Kedro we currently _sort of_ enable compact keys.

```
# nested dictionary
a:
  b: value

# compact key
c.d: value2
```

Here it doesn't actually form the nested dictionary in the compact key case:
```
In [19]: catalog.load("parameters")
Out[19]: {'a': {'b': 'value'}, 'c.d': 'value2'}
```

The problem (or benefit, depending on your view) is that both of these will work the same way as a node input, i.e. both `params:a.b` and `params:c.d` are possible. No need for any quote marks.

(In fact, as per https://github.com/quantumblacklabs/private-kedro/issues/965, there's a possible clash here since who knows which takes priority, i.e. if I change `c.d` to `a.b` above and do `catalog.load("params:a.b")`, would you end up with `value` or `value2`? It's `value2` but no particularly good reason for this.)

Hence, it's not obvious to me a priori what `kedro run --params:a.b:value3` should actually do. As you have obviously noticed, it sets things as a nested dictionary, and if we allow your syntax then it does indeed make things less ambiguous so that you can also set the value as a compact key. But should we then allow `params:"a.b"` as well as `params:a.b` to resolve the above ambiguity?

Overall, it seems like kind of an edge case which we haven't solved for the `params:...` loading syntax. Given we're planning to move to OmegaConf, I would tend to just let them take this lead on this and not do anything custom on the Kedro side.

MerelTheisenQB:
I agree with @AntonyMilneQB that we should probably hold off on this change and follow the `OmegaConf` way of doing this.

deepyaman:
@AntonyMilneQB @MerelTheisenQB Sorry for not responding sooner, but I've been thinking about it--isn't this more aligned with OmegaConf (as in, there's a way to avoid fully unrolling parameters, rather than making compact keys, for which there is no OmegaConf equivalent, the default)? I feel like this is a minor change that makes things less ambiguous (as pointed out by Antony) until we move to OmegaConf.

The need basically arises from a use case where was passing Kubernetes metadata as parameters.

deepyaman:
@merelcht Just to confirm, this is no longer relevant because of moving to OmegaConf, right? If so, I'll close it.

merelcht:
Sorry for the late reply @deepyaman ! Yes I think we should just leverage `OmegaConf` for this kind of stuff and follow their way of doing things. Additionally, I think this problem fits well with the parameters research https://github.com/kedro-org/kedro/issues/2240 Maybe you can add some comments there on what you use parameters for and how?

Issue #1301, opened 2022-02-28
[KED-2422] Provide useful error message when trying to run specific nodes within namespaces


MerelTheisenQB:
(transfer from Jira, created by @datajoely)

### Description
Responded to wizard thread on slack.

The user was trying to run `kedro run -n [node_name]` this was not working for the user since they were trying to run a node that actually lived within a namespace so he actually needed to address it via `modular_pipleine.node_name` 

The error message that the user gets when the node is not found is this one:

```
lib/python3.8/site-packages/kedro/pipeline/pipeline.py:415 

ValueError: Pipeline does not contain nodes named [xxx]
```

Whilst not incorrect, there is no prompt to get the names of the nodes that are in scope to identify if there is (a) A typo (b) what the accepted options will be

As we do more modular pipelines I expect this issue to come up more. Even offering an equivalent of catalog list for nodes would be useful.

If you do the following the namespaces of the specific nodes are not present:

kedro pipeline describe _default_
Nodes:

```
load data (<lambda>) Set Timestep (set_timestep) 
get_batches_time_series (get_batches_time_series) 
load_regressor (load_regressor) 
add_transformers (add_transformers) <Lambda> (<lambda>) plot_all_batches_profiles (plot_all_batches_profiles) 
<Lambda> (<lambda>) 
generate_features (generate_features) 
plot_all_features_overview (plot_all_features_overview) 
Set Batch Type (set_batch_type) 
generate performance report (generate_performance_report) 
```

But to run a particular node you need to specify a namespace:

```
ValueError: Pipeline does not contain nodes named ['generate_performance_report'].
```

Instead you have to specify:

```
kedro run --node train.train_model.generate_performance_report
```

### Proposed solution
So can we include the namespaces in the ValueError?

datajoely:
We could also say the node name `'generate_performance_report' is available under the following namespaces ['a', 'b', 'c'], did you mean to specify any of those?`?

PR #1773, opened 2022-08-09
Provide useful error message when trying to run specific nodes within namespaces

AhdraMeraliQB:
## Description

This PR addresses [this issue](https://github.com/kedro-org/kedro/issues/1301); currently omitting a namespace when trying to run individual nodes results in an uninformative Value Error:

> ValueError: Pipeline does not contain nodes named ['preprocess_companies_node'].

The changes made in this PR add suggestion to the error description, so that when a user tries to run a node without its namespace the error is more helpful:

> ValueError: Pipeline does not contain nodes named ['preprocess_companies_node']. Did you mean: 
['data_processing.preprocess_companies_node']?

## Development notes

This was tested on the spaceflights starter:
`kedro new --starter=spaceflights`

Expected behaviour:
From the project directory, running `kedro run -n preprocess_companies_node` should yield the following error message:
 ` "ValueError: Pipeline does not contain nodes named ['preprocess_companies_node']. Did you mean: 
['data_processing.preprocess_companies_node']?" `

**Important to note:**
The complexity of the namespace search is _O(n*m)_ with _n_ being the number of nodes in a project and _m_ being the number of arguments provided to `kedro run -n`; for larger pipelines this will slow down. However, this is not the only iteration of the nodes in the pipeline implementation - is the potential slowdown a concern?

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1773"></a>

### DIFF
--- a/kedro/pipeline/pipeline.py
+++ b/kedro/pipeline/pipeline.py
@@ -382,12 +382,13 @@ class Pipeline:  # pylint: disable=too-many-public-methods
         unregistered_nodes = set(node_names) - set(self._nodes_by_name.keys())
         if unregistered_nodes:
             # check if unregistered nodes are available under namespace
+            namespaces = []
             for name in node_names:
-                namespaces = [
-                    k
-                    for k, v in self._nodes_by_name.items()
-                    if re.match(rf"^.*\.{name}$", k)
-                ]
+                namespaces.extend([
+                    node_name
+                    for node_name in self._nodes_by_name.keys()
+                    if re.match(rf"^.*\.{name}$", node_name)
+                ])
             if namespaces:
                 raise ValueError(
                     f"Pipeline does not contain nodes named {list(unregistered_nodes)}. "
--- a/kedro/pipeline/pipeline.py
+++ b/kedro/pipeline/pipeline.py
@@ -381,11 +381,12 @@ class Pipeline:  # pylint: disable=too-many-public-methods
         unregistered_nodes = set(node_names) - set(self._nodes_by_name.keys())
         if unregistered_nodes:
             # check if unregistered nodes are available under namespace
-            for n in node_names:
-                namespaces = [k for k,v in self._nodes_by_name.items() if n in k]
+            for name in node_names:
+                namespaces = [k for k, v in self._nodes_by_name.items() if name in k]
             if namespaces:
                 raise ValueError(
-                    f"Pipeline does not contain nodes named {list(unregistered_nodes)}. Did you mean: {namespaces}?"
+                    f"Pipeline does not contain nodes named {list(unregistered_nodes)}. "
+                    f"Did you mean: {namespaces}?"
                 )
             raise ValueError(
                 f"Pipeline does not contain nodes named {list(unregistered_nodes)}."
--- a/kedro/pipeline/pipeline.py
+++ b/kedro/pipeline/pipeline.py
@@ -380,6 +380,13 @@ class Pipeline:  # pylint: disable=too-many-public-methods
         """
         unregistered_nodes = set(node_names) - set(self._nodes_by_name.keys())
         if unregistered_nodes:
+            # check if unregistered nodes are available under namespace
+            for n in node_names:
+                namespaces = [k for k,v in self._nodes_by_name.items() if n in k]
+            if namespaces:
+                raise ValueError(
+                    f"Pipeline does not contain nodes named {list(unregistered_nodes)}. Did you mean: {namespaces}?"
+                )
             raise ValueError(
                 f"Pipeline does not contain nodes named {list(unregistered_nodes)}."
             )
--- a/kedro/pipeline/pipeline.py
+++ b/kedro/pipeline/pipeline.py
@@ -384,11 +384,13 @@ class Pipeline:  # pylint: disable=too-many-public-methods
             # check if unregistered nodes are available under namespace
             namespaces = []
             for name in node_names:
-                namespaces.extend([
-                    node_name
-                    for node_name in self._nodes_by_name.keys()
-                    if re.match(rf"^.*\.{name}$", node_name)
-                ])
+                namespaces.extend(
+                    [
+                        node_name
+                        for node_name in self._nodes_by_name.keys()
+                        if re.match(rf"^.*\.{name}$", node_name)
+                    ]
+                )
             if namespaces:
                 raise ValueError(
                     f"Pipeline does not contain nodes named {list(unregistered_nodes)}. "
--- a/kedro/pipeline/pipeline.py
+++ b/kedro/pipeline/pipeline.py
@@ -382,12 +382,12 @@ class Pipeline:  # pylint: disable=too-many-public-methods
         if unregistered_nodes:
             # check if unregistered nodes are available under namespace
             namespaces = []
-            for name in node_names:
+            for unregistered_node in unregistered_nodes:
                 namespaces.extend(
                     [
                         node_name
                         for node_name in self._nodes_by_name.keys()
-                        if node_name.endswith(f".{name}")
+                        if node_name.endswith(f".{unregistered_node}")
                     ]
                 )
             if namespaces:
--- a/kedro/pipeline/pipeline.py
+++ b/kedro/pipeline/pipeline.py
@@ -5,7 +5,6 @@ produced outputs and execution order.
 """
 import copy
 import json
-import re
 from collections import Counter, defaultdict
 from itertools import chain
 from typing import Dict, Iterable, List, Set, Tuple, Union
@@ -388,7 +387,7 @@ class Pipeline:  # pylint: disable=too-many-public-methods
                     [
                         node_name
                         for node_name in self._nodes_by_name.keys()
-                        if re.match(rf"^.*\.{name}$", node_name)
+                        if node_name.endswith(f".{name}")
                     ]
                 )
             if namespaces:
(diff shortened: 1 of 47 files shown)


datajoely:
This is great! I've wanted this for so so long 💪

kedro/pipeline/pipeline.py
```diff
@@ -380,6 +381,18 @@ def only_nodes(self, *node_names: str) -> "Pipeline":
         """
         unregistered_nodes = set(node_names) - set(self._nodes_by_name.keys())
         if unregistered_nodes:
+            # check if unregistered nodes are available under namespace
+            for name in node_names:
+                namespaces = [
+                    k
+                    for k, v in self._nodes_by_name.items()
```
MerelTheisenQB, line 388: It would be helpful to give more meaningful names to these variables, to make it easier to read the code.

MerelTheisenQB (commented):
It looks like the error isn't showing correctly if you provide multiple node arguments in the CLI, e.g.

`kedro run -n evaluate_model_node -n split_data_node`
Will result in: 
```
ValueError: Pipeline does not contain nodes named ['evaluate_model_node', 'split_data_node']. Did you mean: ['data_science.active_modelling_pipeline.split_data_node', 'data_science.candidate_modelling_pipeline.split_data_node']?
```

And `kedro run -n split_data_node -n evaluate_model_node`, results in:
```
ValueError: Pipeline does not contain nodes named ['split_data_node', 'evaluate_model_node']. Did you mean: ['data_science.active_modelling_pipeline.evaluate_model_node', 'data_science.candidate_modelling_pipeline.evaluate_model_node']?
```
So it only shows the suggestion for the last argument. 

Another find, when you provide multiple arguments, where one is correct and the other isn't it will not mention the namespace suggestion anymore:
`kedro run -n split_data_node -n data_science.active_modelling_pipeline.evaluate_model_node`
results in:
```
ValueError: Pipeline does not contain nodes named ['split_data_node'].
```

whereas `kedro run -n split_data_node`
results in:
```
ValueError: Pipeline does not contain nodes named ['split_data_node']. Did you mean: ['data_science.candidate_modelling_pipeline.split_data_node', 'data_science.active_modelling_pipeline.split_data_node']?
```

kedro/pipeline/pipeline.py
AhdraMeraliQB: Done!

datajoely:
One last thought - perhaps a NameError is more appropriate here

AhdraMeraliQB:
> One last thought - perhaps a NameError is more appropriate here

@datajoely Would this impact consistency? Where `kedro run -n <incorrect_node> -n <missing_namespace>` would yield a NameError but `kedro run -n <incorrect_node>` yields a ValueError?

tests/pipeline/test_pipeline.py
```diff
@@ -791,6 +791,19 @@ def test_only_nodes_unknown(self, pipeline_list_with_lists, target_node_names):
         with pytest.raises(ValueError, match=pattern):
             full.only_nodes(*target_node_names)
 
+    @pytest.mark.parametrize(
+        "dummy_namespaced_nodes", ["node1", "node2", "node3", "node4", "node5", "node6"]
+    )
+    def test_only_nodes_with_namespacing(
```
MerelTheisenQB, line 797: I appreciate this test is written in quite a clever way, but it's also quite hard to understand what exactly is going on, as in it's not clear what the error message will actually contain. I'm also wondering if this should've caught the issue I found by manually testing and if not, it would be nice to have a test for the multiple node use cases I tried.

AhdraMeraliQB: Good point! I'll add some tests to check for multiple correct arguments, and multiple arguments where one is correct and one is not, to make sure all bases are covered.

AntonyMilneQB:
> 

To add to this, we use `ValueError` across Kedro for exactly this sort of error (e.g. invalid pipeline name). It's very common practice in Python to reserve `NameError` for actual missing variables and to use `ValueError` to include the case here that "the value you supplied refers to something that does not exist". So `ValueError` is definitely more appropriate here.

datajoely:
Gotcha it was more of a hairbrained thought than anything more :)

AntonyMilneQB (commented):
This looks good - will be a really nice feature to have. I'm not concerned about the complexity of it being a performance hit because this will only be triggered if `unregistered_nodes` is non-empty, which raises an error. So it's not like this code will run multiple times as part of a functioning pipeline or anything like that. Also, given the likely size of `n` and `m` the actual time added by this code be negligible anyway.

I have two general questions:
1. Elsewhere in kedro we use `difflib`, e.g. to [suggest catalog entry names](https://github.com/kedro-org/kedro/blob/main/kedro/io/data_catalog.py#L292) or CLI commands that are mistyped. Did you consider using that here rather than coding our own very simple matching algorithm? I don't know what the capabilities of `difflib` are so maybe it doesn't have anything suitable for this case, but worth checking.
2. Beyond the scope of this PR, so don't add it here. But why should we limit this to the case of only making the suggestion for namespaced nodes? Should we do a nearest match to any node names that the user has mis-specified? The namespaced case that this solves is definitely the most important case, so I don't think it's high priority. But maybe we get it for free with `difflib`.

tests/pipeline/test_pipeline.py
```diff
@@ -791,6 +791,56 @@ def test_only_nodes_unknown(self, pipeline_list_with_lists, target_node_names):
         with pytest.raises(ValueError, match=pattern):
             full.only_nodes(*target_node_names)
 
+    @pytest.mark.parametrize(
+        "namespaced_node", ["node1", "node2", "node3", "node4", "node5", "node6"]
```
noklam, line 795: I think this would be better named as ` non_namespaced_node` instead of `namespaced_node`. As the test is trying to pass incorrect node names into a namespaced pipeline and catch the error.

AhdraMeraliQB: Makes sense!

tests/pipeline/test_pipeline.py
```diff
@@ -791,6 +791,56 @@ def test_only_nodes_unknown(self, pipeline_list_with_lists, target_node_names):
...
+        "namespaced_nodes", [("node1", "node2"), ("node3", "node4"), ("node5", "node6")]
+    )
+    def test_only_nodes_with_namespacing_multiple_args(
+        self, pipeline_with_namespaces, namespaced_nodes
+    ):
+        # Tests that error message will contain suggestions for all provided arguments.
+        # Example of expected error message:
+        # "Pipeline does not contain nodes named ['node1', 'node2'].
+        # Did you mean: ['katie.node1', 'lisa.node2']?"
+        pattern = rf"('.*\.{namespaced_nodes[0]}')+.*('.*\.{namespaced_nodes[1]}')+"
```
noklam, line 821: Are `Pipeline does not contain nodes named ...` missing in the `pattern?`

AhdraMeraliQB: The test here is just to make sure that suggestions for both arguments appear in the string. This is because the `test_only_nodes_with_pipeline` tests for the full error and the suggestions cannot appear without the surrounding text.

kedro/pipeline/pipeline.py
```diff
@@ -380,6 +381,21 @@ def only_nodes(self, *node_names: str) -> "Pipeline":
         """
         unregistered_nodes = set(node_names) - set(self._nodes_by_name.keys())
         if unregistered_nodes:
+            # check if unregistered nodes are available under namespace
+            namespaces = []
+            for name in node_names:
```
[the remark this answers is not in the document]
AhdraMeraliQB: > 1. Elsewhere in kedro we use `difflib`, e.g. to [suggest catalog entry names](https://github.com/kedro-org/kedro/blob/main/kedro/io/data_catalog.py#L292) or CLI commands that are mistyped. Did you consider using that here rather than coding our own very simple matching algorithm? I don't know what the capabilities of `difflib` are so maybe it doesn't have anything suitable for this case, but worth checking.

`difflib` does offer some interesting capabilities, especially if we go for the route suggested in the second point, but I don't think the nature of its implementation works in this context. A quick test reveals the `get_close_matches` will omit the "correct" suggestions in favour for unrelated nodes that it computes as being more similar to the inputs 😬.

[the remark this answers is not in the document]
AhdraMeraliQB: > Surely we only need to check through unregistered_nodes?

Thanks for catching that! As a consequence the error message will no longer display the incorrect nodes in the order they were provided by the user - is this behaviour undesirable or should it be ignored?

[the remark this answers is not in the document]
AntonyMilneQB: I think it's fine - the order isn't important here.

tests/pipeline/test_pipeline.py
```diff
@@ -791,6 +791,65 @@ def test_only_nodes_unknown(self, pipeline_list_with_lists, target_node_names):
...
+        "non_namespaced_node_names",
+        [("node1", "node2"), ("node3", "node4"), ("node5", "node6")],
+    )
+    def test_only_nodes_with_namespacing_multiple_args(
+        self, pipeline_with_namespaces, non_namespaced_node_names
+    ):
+        # Tests that error message will contain suggestions for all provided arguments.
+        # Example of expected error message:
+        # "Pipeline does not contain nodes named ['node1', 'node2'].
+        # Did you mean: ['katie.node1', 'lisa.node2']?"
```
MerelTheisenQB, line 822: These comments help a lot in understanding the tests, but I'd be in favour of refactoring some of this to make it more readable in a separate task. Can you create an issue for that?

MerelTheisenQB (approved):
This final solution is very clean and easy to follow. Great work! ⭐ ⭐

Issue #1388, opened 2022-03-29
Versioned datasets don't work with Prefect


ardoi:
## Description
Versioned datasets used as outputs causes node to fail when scheduling runs with Prefect.

## Context
I'm trying to run a Kedro pipeline in Prefect. Because some of the output datasets used by nodes in the pipeline are versioned, running the node fails. 

## Steps to Reproduce
1. Follow the [Prefect deployment guide](https://kedro.readthedocs.io/en/stable/10_deployment/05_prefect.html) to set up Prefect and register the pipeline with Prefect. As a side note, the script on the current page doesn't work for me and I had to [update](https://github.com/kedro-org/kedro/pull/1379) one of the imports.
2. Running the pipeline in Prefect works once. Running it a second time or scheduling runs fails with a DataSetError.

## Expected Result
Pipeline should be able to run when triggered after the first time.

## Actual Result
Execution fails with the error. 
```
kedro.io.core.DataSetError: Save path `/a/b/c/2022-03-29T14.20.46.529Z/xyz` 
for ParquetDataSet(filepath=/a/b/c/xyz, load_args={}, protocol=file, save_args={}, 
version=Version(load=None, save='2022-03-29T14.20.46.529Z')) 
must not exist if versioning is enabled. 
```
It seems the Dataset is reusing the timestamp from when the `register_prefect_flow.py` is executed instead of the actual run time of the pipeline. In this case I ran the registration script at 14:20 and triggered the pipeline to be run at 14:23 but the timestamp in the error message above corresponds to the script run time and not trigger time. 

I looked around in Kedro code a bit and it seems the [function](https://github.com/kedro-org/kedro/blob/f11d29d44fee234e0bd45df86356c3aabca3f2fa/kedro/io/core.py#L546) generating the timestamp is cached, but not sure if that's all there is to it. 

## Your Environment
Include as many relevant details about the environment in which you experienced the bug:

* Kedro version used (`pip show kedro` or `kedro -V`): 0.17.7
* Python version used (`python -V`): 3.8.12
* Operating system and version: MacOS 11.6.4
* Prefect version: 1.1.0

avan-sh:
Hi @ardoi, 

I remember the same issue from one of the discord discussions. One working solution that we got to is [here](https://github.com/avan-sh/spaceflights-prefect) in the `prefect_flow.py`. 
This approach requires writing all datasets that are reused by another nodes back to disk. 

[...]

PR #1775, opened 2022-08-10
Fix Prefect deployment code

wim-qb:
## Description
Current Prefect deployment docs specify code which breaks when using versioned datasets and trying multiple prefect runs. 
This issue was picked up and described [here](https://github.com/kedro-org/kedro/issues/1388) in issue 1388

*The old code functioned as follows*:

- Start a kedro session
- Wrap all kedro node functions in a prefect Task object
- Construct a pipeline out of the Task objects
- Register pipeline on Prefect Server

Note that the kedro session is initialisation is outside of the scope of the prefect flow.

*The issue*:
Prefect allows for Pipeline re-runs through the web interface. As the kedro session is instantiated outside of the scope of the Prefect flow when a pipeline - which saves versioned datasets to file - is re-run collisions occur as the timestamps for the versioned files do not change. 

*Solution*:

- Move session instantiation to a custom kedro task object
- Wrap all node functions in a Prefect task object
- Construct a pipeline where the first task is the session instantiation task.
-  Pass the session to all subsequent tasks
- register flow

*Notes*:

Prefect allows for in memory transfer of data from task to task. This distinguishes it from airflow. Hence the code was made in such a way to allow for Kedro Memory dataset to be used (by keeping one session open per Flow run). This means that we can leverage shared memory when deploying to Prefect. 

## Development notes

In docs/source/deployment/prefect.md:

- Added custom KedroInitTask to wrap session initialisation
- Moved main function to the top of the file
- Parcelled out some logic to helper function to make code more readable
- in `generate_flow` moved to using `node.bind()` to allow for in memory transfer of session between prefect Tasks.

## Checklist

Have tested all changes on my local by following the documentation exactly and deploying the space flights tutorial kedro pipeline to prefect. Happy to jump online if reviewers don't want to go through the same ordeal and demonstrate the code. 

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1775"></a>

### DIFF
--- a/docs/source/deployment/prefect.md
+++ b/docs/source/deployment/prefect.md
@@ -71,7 +71,7 @@ def prefect_deploy(pipeline_name, env, package_name):
 
         tasks[node._unique_key]["parent_tasks"] = parent_tasks
 
-    # Below task is used to instantiate a kedro session within the scope of a
+    # Below task is used to instantiate a KedroSession within the scope of a
     # Prefect flow
     init_task = KedroInitTask(
         pipeline_name=pipeline_name,
@@ -93,12 +93,12 @@ def prefect_deploy(pipeline_name, env, package_name):
 
 
 class KedroInitTask(Task):
-    """Task to initialize kedro session"""
+    """Task to initialize KedroSession"""
 
     def __init__(
         self,
         pipeline_name: str,
-        package_name: str = None,
+        package_name: str,
         project_path: Union[Path, str] = None,
         env: str = None,
         extra_params: Dict[str, Any] = None,
@@ -152,11 +152,11 @@ def instantiate_task(
     way we avoid duplicate instantiations of the same node task.
 
     Args:
-        node: kedro node for which a prefect task is being created.
+        node: Kedro node for which a Prefect task is being created.
         tasks: dictionary mapping node names to a dictionary containing
         node tasks and parent node tasks.
 
-    Returns: prefect task for the passed node and task dictionary.
+    Returns: Prefect task for the passed node and task dictionary.
 
     """
     if tasks.get(node._unique_key) is not None:
@@ -174,14 +174,14 @@ def generate_flow(
     tasks: Dict[str, Dict[str, Union[KedroTask, List[KedroTask]]]],
 ):
     """
-    Constructs a prefect flow given a task dictionary. Task dictionary
-    maps kedro node names to a dictionary containing a node task and its
+    Constructs a Prefect flow given a task dictionary. Task dictionary
+    maps Kedro node names to a dictionary containing a node task and its
     parents.
 
     Args:
-        init_task: Prefect initialisation tasks. Used to instantiate a kedro
-        session within the scope of a prefect flow.
-        tasks: dictionary mapping kedro node names to a dictionary
+        init_task: Prefect initialisation tasks. Used to instantiate a Kedro
+        session within the scope of a Prefect flow.
+        tasks: dictionary mapping Kedro node names to a dictionary
         containing a corresponding node task and its parents.
 
     Returns: None
@@ -199,7 +199,7 @@ def generate_flow(
 
 
 def instantiate_client(project_name: str):
-    """Initiates prefect client"""
+    """Initiates Prefect client"""
     client = Client()
     try:
         client.create_project(project_name=project_name)
--- a/docs/source/deployment/prefect.md
+++ b/docs/source/deployment/prefect.md
@@ -7,7 +7,8 @@ In scope of this deployment, we are interested in [Prefect Server](https://docs.
 ```{note}
 Prefect Server ships out-of-the-box with a fully featured user interface.
 ```
-
+Please note that this deployment has been tested using of kedro 0.17.6, 0.17.7 and 0.18.2 and prefect version 1.1.0.
+The current implementation will not work for prefect 2.0.0.
 ## Prerequisites
 
 To use Prefect Core and Prefect Server, ensure you have the following prerequisites in place:
@@ -111,7 +112,7 @@ class KedroInitTask(Task):
         self.env = env
         super().__init__(name=f"{package_name}_initialization", *args, **kwargs)
 
-    def run(self) -> Tuple[DataCatalog, KedroSession]:
+    def run(self) -> Tuple[DataCatalog, str]:
         """Initializes a Kedro session and returns the DataCatalog and KedroSession"""
         # bootstrap project within task / flow scope
         bootstrap_project(self.project_path)
@@ -128,7 +129,7 @@ class KedroInitTask(Task):
         unregistered_ds = pipeline.data_sets() - set(catalog.list())  # type: ignore[union-attr]
         for ds_name in unregistered_ds:
             catalog.add(ds_name, MemoryDataSet())
-        return catalog, session
+        return catalog, session.session_id
 
 
 class KedroTask(Task):
@@ -138,8 +139,11 @@ class KedroTask(Task):
         self._node = node
         super().__init__(name=node.name, tags=node.tags)
 
-    def run(self, catalog, session):
-        run_node(self._node, catalog, _create_hook_manager(), session.session_id)
+    def run(self, argument_tuple: Tuple[DataCatalog, str]):
+        # argument_tuple definition: (data_catalog, session_id)
+        run_node(
+            self._node, argument_tuple[0], _create_hook_manager(), argument_tuple[1]
+        )
 
 
 def instantiate_task(
@@ -186,7 +190,7 @@ def generate_flow(
 
     Returns: None
     """
-    catalog, session = init_task
+    child_task_args = init_task
     for task in tasks.values():
         node_task = task["task"]
         if len(task["parent_tasks"]) == 0:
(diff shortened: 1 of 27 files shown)


datajoely:
Thanks @wim-qb  for doing this - this looks great!

docs/source/deployment/prefect.md
```diff
@@ -113,8 +92,124 @@ def build_and_register_flow(pipeline_name, env):
     flow.run_agent()
 
 
+class KedroInitTask(Task):
+    """Task to initialize kedro session"""
+
+    def __init__(
+        self,
+        pipeline_name: str,
+        package_name: str = None,
```
MerelTheisenQB, line 101: This shouldn't really be `None` though right? It wouldn't fail, but you'd get an odd name of "None_initialization" in the snippet below `f"{package_name}_initialization"`. I think you can just add it as a required argument.

docs/source/deployment/prefect.md
```diff
@@ -113,8 +92,124 @@ def build_and_register_flow(pipeline_name, env):
...
+        **kwargs,
+    ):
+        self.project_path = Path(project_path or Path.cwd()).resolve()
+        self.extra_params = extra_params
+        self.pipeline_name = pipeline_name
+        self.env = env
+        super().__init__(name=f"{package_name}_initialization", *args, **kwargs)
+
+    def run(self) -> Tuple[DataCatalog, KedroSession]:
+        """Initializes a kedro session and returns the data catalog and session"""
```
MerelTheisenQB, line 115:
```suggestion
        """Initializes a Kedro session and returns the DataCatalog and KedroSession"""
```

docs/source/deployment/prefect.md
```diff
@@ -113,8 +92,124 @@ def build_and_register_flow(pipeline_name, env):
     flow.run_agent()
 
 
+class KedroInitTask(Task):
+    """Task to initialize kedro session"""
```
MerelTheisenQB, line 96:
```suggestion
    """Task to initialize KedroSession"""
```

docs/source/deployment/prefect.md
```diff
@@ -113,8 +92,124 @@ def build_and_register_flow(pipeline_name, env):
...
+    # return tasks as it is mutated. We want to make this obvious to the user.
+    return node_task, tasks  # type: ignore[return-value]
+
+
+def generate_flow(
+    init_task: KedroInitTask,
+    tasks: Dict[str, Dict[str, Union[KedroTask, List[KedroTask]]]],
+):
+    """
+    Constructs a prefect flow given a task dictionary. Task dictionary
```
MerelTheisenQB, line 177:
```suggestion
    Constructs a Prefect flow given a task dictionary. Task dictionary
```

docs/source/deployment/prefect.md
```diff
@@ -113,8 +92,124 @@ def build_and_register_flow(pipeline_name, env):
...
+    maps kedro node names to a dictionary containing a node task and its
```
MerelTheisenQB, line 178:
```suggestion
    maps Kedro node names to a dictionary containing a node task and its
```

docs/source/deployment/prefect.md
```diff
@@ -113,8 +92,124 @@ def build_and_register_flow(pipeline_name, env):
...
+            # When a task has no parent only the session init task should precede it.
+            parent_tasks = [init_task]
+        else:
+            parent_tasks = task["parent_tasks"]
+        # Set upstream tasks and bind required kwargs
+        node_task.bind(upstream_tasks=parent_tasks, catalog=catalog, session=session)  # type: ignore[union-attr]
+
+
+def instantiate_client(project_name: str):
+    """Initiates prefect client"""
```
MerelTheisenQB, line 202:
```suggestion
    """Initiates Prefect client"""
```

MerelTheisenQB (approved):
This is awesome! 😄 Thank you so much for working on this @wim-qb ⭐ ⭐ 
I've left some minor non-blocking comments, mainly around capitalisation of `Prefect` and `Kedro` in the comments.  

It would be interesting to hear what @avan-sh and/or @ardoi think of this, as they originally reported [the issue ](https://github.com/kedro-org/kedro/issues/1388).

docs/source/deployment/prefect.md
wim-qb: Noticed that I haven't set up github to sign off on suggest changes I commit through the web interface. Will solve that problem some other time. For now I am implementing all your suggestions locally to allow for sign off.

docs/source/deployment/prefect.md
wim-qb: :+1:

noklam (commented):
> /Users/Nok_Lam_Chan/miniconda3/lib/python3.9/site-packages/prefect/__init__.py:68: UserWarning: Attempted import of 'prefect.Client', which is part of Prefect 1.x, while Prefect 2.0.4 is installed. If you're upgrading you'll need to update your code, see the Prefect 2.x migration guide: `https://orion-docs.prefect.io/migration_guide/`. Otherwise ensure that your code is pinned to the expected version.

While we can't keep up with the latest version all the time,  I think we should document which version of `kedro` and `perfect` would work together with this guide. I haven't get it working on my end yet, will drop more comments after that.

wim-qb:
> > /Users/Nok_Lam_Chan/miniconda3/lib/python3.9/site-packages/prefect/**init**.py:68: UserWarning: Attempted import of 'prefect.Client', which is part of Prefect 1.x, while Prefect 2.0.4 is installed. If you're upgrading you'll need to update your code, see the Prefect 2.x migration guide: `https://orion-docs.prefect.io/migration_guide/`. Otherwise ensure that your code is pinned to the expected version.
> 
> While we can't keep up with the latest version all the time, I think we should document which version of `kedro` and `perfect` would work together with this guide. I haven't get it working on my end yet, will drop more comments after that.
@noklam 
Very good point. I tested the code across two kedro versions 0.18.2 and 0.17.7. For both kedro versions I used one prefect version 1.1.0. The prefect could easily be giving you issues. Let me know if I should hop on a call and help you out. No one should have to struggle to make someone else's code work. Once you have it up and running and made any further comments I will address those and amend the documentation in a single commit.

avan-sh:
@wim-qb Thanks for these changes, particularly the ability to share data between tasks is great. 

I could verify the changes already & they work. I also have a couple of questions about the way tasks show up. I'm not a power user of prefect so these might be invalid.
1. I see there are multiple initialisation tasks , one for each task without any parent dependencies. The logs for index suffixed tasks doesn't have any run id but the original `kedro_prefect_initialization` task has a log stating `Session created with ID 2022-08-15T11.30.56.648Z`. 
image

2. [Unrelated to current changes] Do you know which terminal to enable kedro logger in the prefect flow run? I tried a couple of ways(including stated) to enable this but none of my approaches are showing logs from kedro/pipeline nodes in the flow tasks.

wim-qb:
> avan-sh

Hi, 

Thanks so much for the input. I have already observed issue 1. I'll link some of my findings tomorrow once I have a little more time.

noklam:
@MerelTheisenQB My mistake, I was trying to re-request myself but clicked on your name instead.

docs/source/deployment/prefect.md
```diff
@@ -113,8 +92,124 @@ def build_and_register_flow(pipeline_name, env):
...
+
+    Args:
+        init_task: Prefect initialisation tasks. Used to instantiate a Kedro
+        session within the scope of a Prefect flow.
+        tasks: dictionary mapping Kedro node names to a dictionary
+        containing a corresponding node task and its parents.
+
+    Returns: None
+    """
+    catalog, session = init_task
```
noklam, line 189: This is the weird bit that generated kedro_prefect_initialization[0] and kedro_prefect_intialization[1].

Not sure what is the correct thing, but it should be a `KedroInitTask` object instead and a session

noklam (commented):
@wim-qb Sorry for the delay. I have tested and it works if I run multiple times. The main issue is with the duplicate initialisation task, I have drop a comment on that.

Thanks for making this PR!

wim-qb:
@noklam @avan-sh It took me a little while to test all permutations of fixing the above requests concerning initialisation tasks. Let me step back a bit an explain exactly what was happening before. 

Prefect adds tasks to a flow when collections are queried under the scope of the prefect flow. For more see (these docs)[https://docs-v1.prefect.io/core/concepts/tasks.html#collections]. In our case we saw some tasks which were dedicated merely to accessing elements in a tuple. Which we don't want to see. 

An easy fix would be to return the output from the `init` task as a tuple. We can then unpack the tuple in the subsequent tasks.  I have done this in the final implementation. This does, however, result in an ugly task graph. Which can be seen under `schematic` on the prefect interface. 

According to the prefect documentation this graph pollution should be avoidable by casting the `init` return tuple a a prefect `Constant`. This definitely has the effect of cleaning up the graph. However, casting a data catalog to `Constant` prevents the use of in memory data sets. I can only assume that some kind of strange copying processes in occurring somewhere in the backend but I am not sure.  

Hence I am left with a best effort where we get non-indexed tasks but a polluted task graph. 

I have updated the code accordingly. Passing tuples around is never ideal when you can rather be explicit but at least there are not unexplained tasks with the code set up as is.  The task graph for space-flights can be seen below.

Screenshot 2022-08-16 at 14 13 29

noklam (approved):
Congrats for your 1st PR ⭐️ , not an easy one for the 1st PR!

MerelTheisenQB (approved):
Thanks for this awesome contribution @wim-qb 🎉

PR #1789, opened 2022-08-16
Enable passing argument to line magic `%reload_kedro` 

### LINKED ISSUE #1748: Add parameters to `%reload_kedro` line magic
## Description

Currently you cannot pass things like `env` or `extra_params` via the line magic, but you can by importing the function.

https://github.com/kedro-org/kedro/blob/5ae97cfb70e5b0d4490132847977d482f13c840f/kedro/extras/extensions/ipython.py#L38

Why don't we introduce feature parity here?

noklam:
## Description

Close #1748 

It's not trivial to enable positional & keywords arguments with line magics, this PR aims to support:
* Accepting keywords arguments with the line magic, so user don't have to import the function `reload_magic` and have a more consistent UX.
* Keeping the existing behavior - passing `path` is optional.

## How to test
1. Open GitPod 
2. Check if you are in a kedro project (It should by default)
3. Try the following with `kedro ipython` / `kedro jupyter notebook`
```python
%reload_kedro    # Default path=None
%reload_kedro .  # With positional argument only
%reload_kedro . --env=base # Mix use of argument type
%reload_kedro --env=base # Keyword argument only, path default=None
```

## Development notes

* To make `%reload_kedro` and `reload_kedro` possible, `@register_line_magic("reload_kedro")` is needed to register a different function as the line magic with name `reload_kedro`
* `path` is made positional and `env` and `extra_params` are keywords arguments, partly due to compatibility issues. I thought about making it accepts both positional or keyword arguments, but it seems that this is not aligned with how CLI works in general, so I did not do it in the end.
  * There are some hints that how you can do that in this link.
https://stackoverflow.com/questions/14585722/can-you-pass-keyword-arguments-instead-of-positional-arguments-to-argparse

Todo:
- [x] Update docstring - `reload_kedro` + Shift-Tab in Jupyter should show up some examples. 
image

- [ ] Check `reload_kedro` without parameter still works correctly 
- [ ] Check `reload_kedro` with `path` as positional argument still works
- [ ] 

### Side-notes (Only read if you have extra time)
https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.interactiveshell.html?highlight=register_magic_function#IPython.core.interactiveshell.InteractiveShell.register_magic_function suggest that the magic should have a signature `f(line)` but this is not the case with `reload_kedro`. See an example here:
```python
def csv(line, cell):
    sio = StringIO(cell)
    return pd.read_csv(sio)

def load_ipython_extension(ipython):
    """This function is called when the extension is
    loaded. It accepts an IPython InteractiveShell
    instance. We can register the magic with the
    `register_magic_function` method of the shell
    instance."""
    ipython.register_magic_function(csv, 'cell')
```
Currently, we have `def reload_kedro(
    path: str = None, env: str = None, extra_params: Dict[str, Any] = None
)`, it is working only because it takes the first argument as `line` 

This PR use `reload_kedro_magic(line)` which call the `reload_kedro` method, so we keep the backward compatible that people can still import the same function. but `%reload_kedro` will call the line magic instead.

#### Note 3 
The `IPython` use `argparse` for the magic, it's less consistent with most of the `kedro` codebase which use `click`, but I didn't want to go all the way to re-invent the logic. In theory, it is an exercise of passing a long string into arguments and making sure it has the correct type.

## What's more to do?
* `extra_params` now use a `json.loads` method to construct the dictionary, I also consider if we should make it the same style as `kedro run --params` which involves `_split_params` and `_split_string`, but it is probably a rare case, so I would rather wait for a user request to actually implement this.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1789"></a>

# tests
`kedro ipython`, the log should be the same with this PR.
```
[08/17/22 08:37:40] INFO     Updated path to Kedro project: /workspace/project                      ipython.py:53
[08/17/22 08:37:40] WARNING  /home/gitpod/.pyenv/versions/3.8.12/lib/python3.8/site-packages/requ warnings.py:109
                             ests/__init__.py:102: RequestsDependencyWarning: urllib3 (1.26.6) or                
                             chardet (5.0.0)/charset_normalizer (2.0.4) doesn't match a supported                
                             version!                                                                            
                               warnings.warn("urllib3 ({}) or chardet ({})/charset_normalizer                    
                             ({}) doesn't match a supported "                                                    
                                                                                                                 
[08/17/22 08:37:41] INFO     Kedro project project                                                  ipython.py:76
                    INFO     Defined global variable 'context', 'session', 'catalog' and            ipython.py:77
```


noklam:
@datajoely Are these the expected behaviors in mind? I tested on GitPod and it's already worked. If so I will move on to clean up the code a little bit, and docs and modify the tests if needed. 

The IPython magic isn't documented clearly so I am not 100% confident, especially since some part of the code looks obsoleted to me, I will need to confirm with Antony next week.

datajoely:
Does this work?

`%reload_kedro --env=base`

Otherwise, this is amazing - just need to make sure it's in the docs so we can link to it when people ask

kedro/extras/extensions/ipython.py
```diff
@@ -83,12 +83,50 @@ def reload_kedro(
...
+    """
+    import json
 
-    ipython.register_magic_function(reload_kedro, "line", "reload_kedro")
+    from IPython.core.magic import register_line_magic
+    from IPython.core.magic_arguments import argument, magic_arguments, parse_argstring
+
+    @magic_arguments()
+    @argument(
+        "path", help=("Path to the project root directory."), nargs="?", default=None
```
MerelTheisenQB, line 103: What does `nargs="?"` do?

kedro/extras/extensions/ipython.py
```diff
@@ -83,12 +83,50 @@ def reload_kedro(
...
+    )
+    @argument(
+        "--env",
+        dest="env",
```
MerelTheisenQB, line 107: What does `dest` mean?

tests/extras/extensions/test_ipython.py
```diff
@@ -218,3 +224,12 @@ def test_load_extension_not_in_kedro_project(self, mocker, caplog):
...
         assert expected_message in log_messages
+
+    def test_load_extension_register_line_magic(self, mocker):
+        from IPython.testing.globalipapp import get_ipython
+
+        mocker.patch("kedro.extras.extensions.ipython.reload_kedro")
+
+        ipython = get_ipython()
+        load_ipython_extension(ipython)
+        ipython.magic("reload_kedro")
```
MerelTheisenQB, line 235: Is it possible to test passing the arguments as well or would that require more of an end to end test setup?

docs/source/tools_integration/ipython.md
```diff
@@ -48,7 +48,8 @@ In [3]: %reload_kedro
 ```
 
 ```{note}
-If you want to pass arguments to the `reload_kedro` line magic, e.g. to specify a configuration environment, you should call it as a normal Python function (`reload_kedro(env=env, extra_params=extra_params)`) rather than using `%reload_kedro` in a Notebook cell (`%reload_kedro(extra_params=extra_params)` wouldn't work). You might have to call `%automagic False` beforehand to make this work.
+If you want to pass arguments to the `reload_kedro` line magic, e.g. to specify a configuration environment. You can provide extra keyword arguments like this:
+`%reload_kedro <path_to_project_root> --env=<env>`
```
MerelTheisenQB, line 52: I think it would be good to also include an example of how to pass `extra_params`, because users might otherwise not be able to find out that's a possible argument as well.

MerelTheisenQB (commented):
Great work @noklam! I've tested with providing some and all of the arguments and it works really well 😄 I've left some minor questions/comments. 

On your point about `extra_params` using json.loads method to construct the dictionary and not `_split_params` and `_split_string`, I agree we should leave it as is for now and only implement if users ask for it.

kedro/extras/extensions/ipython.py
datajoely: If you `?` on ipython it calls help - super useful

docs/source/tools_integration/ipython.md
datajoely: I also wonder if we should print this to user when the extension initialises

MerelTheisenQB (approved):
Great job! ⭐ 😄

AntonyMilneQB (commented):
I will test this out and give it a proper review tomorrow, but it looks great! Nice work figuring it all out - I thought it would indeed not be as trivial as it sounds! 😬 

> To make %reload_kedro and reload_kedro possible, @register_line_magic("reload_kedro") is needed to register a different function as the line magic with name reload_kedro

Why do we want both of these to be possible rather than just `%reload_kedro`? Just for backwards compatibility or some other reason?

Two other questions:
1. I was not expecting kwargs to have `--` before them, i.e. I thought we should do  `%reload_kedro env=base` and not `--env`. We don't need to match the CLI command exactly here (`extra_params` is called `params` on CLI for a start, so we can't make it perfectly consistent). Or is it normal practice to use `--` for kwargs on line magics?
2. I'm pretty sure I know the answer to this, but is the general format for calling a line magic with arguments `%line_magic positional_arg_1 positional_arg2 kwarg1=value1 kwarg2=value2`? Instead of, for example, `%line_magic(positional_arg_1, positional_arg2, kwarg1=value1, kwarg2=value)`. This was always my understanding but would be good if you could confirm that we're not doing something weird and making up our own syntax here!

noklam:
@AntonyMilneQB 
> I will test this out and give it a proper review tomorrow, but it looks great! Nice work figuring it all out - I thought it would indeed not be as trivial as it sounds! 😬
> 
> > To make %reload_kedro and reload_kedro possible, @register_line_magic("reload_kedro") is needed to register a different function as the line magic with name reload_kedro

I think this is the correct way to define line/cell magic, from the IPython docs
This will create an IPython magic (line, cell or both) from a standalone function. The functions should have the following signatures:

> For line magics: def f(line)
> For cell magics: def f(line, cell)
> For a function that does both: def f(line, cell=None)

https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.interactiveshell.html?highlight=register%20magic%20function#IPython.core.interactiveshell.InteractiveShell.register_magic_function.

Backward compatible is a side benefit here but not the main goal.


> Why do we want both of these to be possible rather than just `%reload_kedro`? Just for backwards compatibility or some other reason?




> 
> Two other questions:
> 
> 1. I was not expecting kwargs to have `--` before them, i.e. I thought we should do  `%reload_kedro env=base` and not `--env`. We don't need to match the CLI command exactly here (`extra_params` is called `params` on CLI for a start, so we can't make it perfectly consistent). Or is it normal practice to use `--` for kwargs on line magics?

I mainly reference the built-in IPython magic, it follows a CLI style of option, and it actually uses `argparse` for the decoration, so it feels natural to stick with CLI style. 
```bash
%autoreload 2
%timeit -r 3 do_something()
%env COLUMN=3  # This is an assignment, not an argument 
```

See example here: https://ipython.readthedocs.io/en/stable/interactive/magics.html#magic-timeit


> 2. I'm pretty sure I know the answer to this, but is the general format for calling a line magic with arguments `%line_magic positional_arg_1 positional_arg2 kwarg1=value1 kwarg2=value2`? Instead of, for example, `%line_magic(positional_arg_1, positional_arg2, kwarg1=value1, kwarg2=value)`. This was always my understanding but would be good if you could confirm that we're not doing something weird and making up our own syntax here!

I haven't seen the syntax `%line_magic(arg)`, again I try to stick with the standard IPython magic, so I end up with CLI style of command.

AntonyMilneQB (commented):
This looks great ⭐ Just a few comments while I understand it all a bit better, and then it would be good to talk about this after I've read a bit more about `argparse`.

* having `path` positional and the others keyword makes sense 👍 
* using `argparse` rather than `click` makes sense 👍 
* the CLI style of arguments makes sense 👍 
* not sure about using `json.loads` for `extra_params` 🤔 Especially once we've switched to using omegaconf for this, it would be nice to be consistent. Or maybe we should just provide a Python dictionary directly here given you can do that in IPython? I think best to be consistent with the CLI here though, given it's not hard to do (I think)
* not sure whether defining the function inside `load_ipython_extension` and having a separate `reload_kedro_line_magic` and `reload_kedro` function is the right thing to do 🤔 Let me play around with it a bit

kedro/extras/extensions/ipython.py
```diff
@@ -83,12 +83,56 @@ def reload_kedro(
...
+        help=(
+            "Optional dictionary containing extra project parameters for underlying "
+            "KedroContext. If specified, will update (and therefore take precedence "
+            "over) the parameters retrieved from the project configuration."
+        ),
+    )
+    @register_line_magic(
+        "reload_kedro"
+    )  # Register the line magic's name as `reload_kedro`
+    def reload_kedro_line_magic(line):
```
AntonyMilneQB, line 123:
```suggestion
    @register_line_magic
    def magic_reload_kedro(line):
```
Would that work? Following the example of `magic_cool` in https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.magic_arguments.html. It would be tidier if that does work, but maybe it's just sloppy documentation 😅 Maybe you don't even need `@register_line_magic` at all? 🤔

kedro/extras/extensions/ipython.py
```diff
@@ -83,12 +83,56 @@ def reload_kedro(
...
-    ipython.register_magic_function(reload_kedro, "line", "reload_kedro")
+    from IPython.core.magic import register_line_magic
+    from IPython.core.magic_arguments import argument, magic_arguments, parse_argstring
+
+    @magic_arguments()
+    @argument(
+        "path", help=("Path to the project root directory."), nargs="?", default=None
+    )
+    @argument(
+        "--env",
```
AntonyMilneQB, line 106:
```suggestion
        "--env",
        type=str,
```

kedro/extras/extensions/ipython.py
```diff
@@ -83,12 +83,56 @@ def reload_kedro(
...
+    """
+    import json
 
-    ipython.register_magic_function(reload_kedro, "line", "reload_kedro")
+    from IPython.core.magic import register_line_magic
+    from IPython.core.magic_arguments import argument, magic_arguments, parse_argstring
+
+    @magic_arguments()
+    @argument(
+        "path", help=("Path to the project root directory."), nargs="?", default=None
```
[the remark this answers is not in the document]
AntonyMilneQB: And maybe also `type=str`.

kedro/extras/extensions/ipython.py
AntonyMilneQB: I'm just skimming through the IPython source code now and saw this:
```
def real_name(magic_func):
    """ Find the real name of the magic.
    """
    magic_name = magic_func.__name__
    if magic_name.startswith('magic_'):
        magic_name = magic_name[len('magic_'):]
    return getattr(magic_func, 'argcmd_name', magic_name)
```
so I think indeed calling it `magic_reload_kedro` will magically work 🎉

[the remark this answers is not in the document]
AntonyMilneQB: Getting increasingly sure about this 😀

AntonyMilneQB: I've got this figured out now - will explain!

RELEASE.md
```diff
@@ -12,9 +12,12 @@
 # Upcoming Release 0.18.3
 
 ## Major features and improvements
+* The line magic `%reload_kedro` now accepts keywords arguments. i.e. `%reload_kedro . --env=base`.
```
AntonyMilneQB, line 15:
```suggestion
* The line magic `%reload_kedro` now accepts keywords arguments, e.g. `%reload_kedro --env prod`.
```

tests/extras/extensions/test_ipython.py
```diff
@@ -218,3 +233,40 @@ def test_load_extension_not_in_kedro_project(self, mocker, caplog):
...
+        # Calling the line magic to check if the line magic is available
+        ipython.magic("reload_kedro")
+        assert mock_reload_kedro.call_count == 2
+
+    @pytest.mark.parametrize(
+        "args",
+        [
+            "",
+            ".",
+            ". --env=base",
```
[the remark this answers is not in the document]
AntonyMilneQB: Just tried this, and It does indeed!

docs/source/tools_integration/ipython.md
```diff
@@ -48,7 +48,8 @@ In [3]: %reload_kedro
 ```
 
 ```{note}
-If you want to pass arguments to the `reload_kedro` line magic, e.g. to specify a configuration environment, you should call it as a normal Python function (`reload_kedro(env=env, extra_params=extra_params)`) rather than using `%reload_kedro` in a Notebook cell (`%reload_kedro(extra_params=extra_params)` wouldn't work). You might have to call `%automagic False` beforehand to make this work.
+If you want to pass arguments to the `reload_kedro` line magic, e.g. to specify a configuration environment. You can provide extra keyword arguments like this:
+`%reload_kedro <path_to_project_root> --env=<env> --extra_params=<extra_params_dict>`
```
AntonyMilneQB, line 52:
```suggestion
`%reload_kedro` accepts optional keyword arguments `env` and `extra_params`. For example, to use configuration environment `prod` you should run `%reload_kedro --env=prod`. For more details, run `%reload_kedro?`.
```

kedro/extras/extensions/ipython.py
AntonyMilneQB: Long story short, we now have a nice help message:
```
In [23]: %reload_kedro?
Docstring:
::

  %reload_kedro [-e ENV] [--extra_params EXTRA_PARAMS] [path]

The `%reload_kedro` IPython line magic. See
 https://kedro.readthedocs.io/en/stable/tools_integration/ipython.html for more.

positional arguments:
  path                  Path to the project root directory. If not given, use the previously setproject root.

optional arguments:
  -e ENV, --env ENV     Kedro configuration environment name. Defaults to `local`.
  --extra_params EXTRA_PARAMS
                        Optional dictionary containing extra project parameters for underlying KedroContext. If specified, will update (and therefore take precedence over) the parameters retrieved
                        from the project configuration.
File:      ~/kedro_stuff/kedro/kedro/extras/extensions/ipython.py
```

AntonyMilneQB (approved):

PR #1794, opened 2022-08-18
Make alias `%load_ext kedro` to replace `%load_ext kedro.extras.extensions.ipython`

### LINKED ISSUE #1763: Alias IPython extension
I'm fed up of (mis)typing `%load_ext kedro.extras.extensions.ipython`. Let's make everyone's lives a bit easier and alias it.

Previously we had said that `%load_ext kedro.ipython` would be an improvement, but I'm going to make a drastic suggestion and go even further to suggest `%load_ext kedro`. If you're doing `%load_ext` anyway then you know it's an IPython extension so the `ipython` part is just unnecessary extra characters.

### How?
Really just one line of code in `kedro/__init__.py` that says:
```
from .extras.extensions.ipython import load_ipython_extension
```
This should turn the module `kedro` into an IPython extension without affecting any of the existing kedro functionality.

### Is that it?
No! This is the important part. As per discussion we had as part of https://github.com/kedro-org/kedro/issues/712 we need to make sure that this `import` doesn't have undesired side effects, since the above line of code will run _every time anyone imports anything from `kedro`_. This means that any imports in the `extras -> extensions -> ipython` import chain will also be executed.

There are three possible risks here:
* the `import` tries to import something that is not a core dependency of kedro and the user doesn't have (_which includes IPython!_)
* the `import` imports loads of things that exist but makes things very slow
* the `import` has side effects that we don't want to be executed (e.g. importing `from kedro.framework.project import LOGGING` actually configures logging)

All three risks here should be mitigated against because in https://github.com/kedro-org/kedro/pull/1761 I already hid all the non-standard library imports inside functions as a guard. So, unless I missed something, what you need to do is write some tests to ensure that this does not get accidentally broken in the future. Something like these (will need some cunning mocking):
* `import kedro` does not try to import `IPython`
* `import kedro` does not do imports like `from kedro.framework.project import LOGGING`

Obviously manually test all this too, but the important thing here is writing the automated tests so it doesn't break in future if someone moves an import to be unguarded at the top level of the ipython.py file.

### Anything else to do?
Yes! Update all mentions in the codebase of `kedro.extras.extensions.ipython` to `kedro`. This is probably just docs and the `kedro ipython/jupyter` commands and should be easy. The hard thing here is the writing the tests.

noklam:

## Description

Close #1763 

Enable `%load_ext kedro` as a shortcut to replace `%load_ext kedro.extras.extensions.ipython`.

Write tests to ensure this

- [ ] import kedro does not try to import IPython
- [ ] import kedro does not do imports like from kedro.framework.project import LOGGING

## Development notes

Tests to make sure future changes doesn't break `kedro ` accidentally.

- [ ]  import kedro does not try to import IPython (added test)
- [ ] import kedro would work (didn't implement, as I think this is covered with `test_load_extension_missing_dependency` already)

### What doesn't work well
```python
import builtins  
spy = mocker.spy(builtins, "__import__")
import kedro
```

I tried a less intrusive approach with `mocker.spy`, but it doesn't seem to capture the right thing, when I run the code block above the `call_arg_list`'s length is 128, which I expect only `kedro` and `logging` should be there. At the end I go with delete the sys.modules (similar to the test for databricks).

`importlib.reload` would not reload the 2nd degree dependencies and `IPython.lib.deepreload.reload` causes `NotImplementedError` with some of the builtin libraries which is C code and cannot be reloaded easily.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1794"></a>


kedro/__init__.py
```diff
@@ -8,4 +8,10 @@
 
 import logging
 
+try:
+    from IPython import get_ipython
+    from .extras.extensions.ipython import load_ipython_extension
+except ModuleNotFoundError:
+    pass  # Skip it when IPython is not instsalled
+
```
noklam, line 16: This also works fine, but as a consequence, these configurations will be broken. Since everything is under `kedro` namespace and effectively they all depend on each other.


```
[[tool.importlinter.contracts]]
name = "CLI > Context > Library, Runner > Extras > IO & Pipeline"
type = "layers"
```

noklam:
@AntonyMilneQB The original issue wrote
*  import kedro does not try to import IPython
* import kedro does not do imports like from kedro.framework.project import LOGGING

Are we just doing this when Ipython is not available? As none of this is True and Ipython is installed and it will import `Ipython` and will import `LOGGING` due to the `kedro.framework.cli` namespace.

AntonyMilneQB:
Closed in favour of #1837

Issue #1477, opened 2022-04-25
[KED-2857] Suggestion for re-run command on failure, i.e. kedro run --from-nodes, should take into account memory dataset


yetudada:
(transfer from Jira, created by @limdauto)

Currently, when a pipeline fails, there is a suggested command to re-run the pipeline from a particular node. However, if the inputs to this node are MemoryDataSet, the pipeline won't be run correctly. A user suggested that Kedro would back track to the last succesful node with a persisted dataset and continue from there.

This ticket is to update the error message.

Issue #1778, opened 2022-08-10
Re-design io.core and io.data_catalog


antonymilne:
Spun out of https://github.com/kedro-org/kedro/issues/1691#issuecomment-1176679924... Let's collect ideas here on what current problems are with `io`. To me it feels like we've neglected it and it's ripe for a re-design.

> **Note**. Like the configuration overhaul, some of this would be non-breaking (behind the scenes implementation) and some would be breaking (changes to public API). _In theory_ we're free to do as we please  with any function starting with `_`, i.e. we can remove, rename, changes arguments at will. _In practice_, however, some of these might be very commonly used (e.g. `self._get_load_path()` used in lots of datasets) so should not be regarded as non-breaking.
--- 
#1691 and #1580 are actually just symptoms of a more fundamental underlying issue: the API and underlying workings of `io.core` and `io.data_catalog` are very confusing and should be rethought in general. These are very old components in kedro and maybe some of the decisions that were originally made about their design should be revised. I think there's also very likely to be old bits of code there that could now be removed or renamed  (e.g. who would guess that something named `add_feed_dict` is used to add parameters to the catalog?). It feels like tech debt rather than intentional design currently.

I don't think they're massively wrong as it stands, but I think it would be a good exercise to go through them and work out exactly what functionality we should expose in the API and how we might like to rework them. e.g. in the case raised here there is quite a bit of confusion about how to get the filepath:
* `catalog.datasets` is presumably the "official" route to get a dataset rather than `_get_dataset`, but `catalog.datasets` wouldn't allow a namespaced datasets to be accessed without doing `getattr`. There's some very subtle and non-obvious differences between `datasets` and `_get_dataset`, and then there's also `catalog._data_sets` (which I _think_ might just be a historical leftover... but not sure). In https://github.com/kedro-org/kedro/pull/1795 @jmholzer used `vars(catalog.datasets)[dataset_name]`.
* it also seems at a glance that  `_filepath` is only defined for versioned datasets (? seems weird)
* to actually get the correct versioned filepath it's even harder - in our datasets we do `get_filepath_str(self._get_load_path(), self._protocol)` which is pretty obscure. Similar to #1654

So I think we should look holistically at the structures involved here and work out what the API _should_ look like so there's one, clear way to access the things that people need to access. I actually don't think this is such a huge task. Then we can tell much more easily whether we need any new functionality in these structures (like a `catalog.dumps`) or whether it's just a case of making what we already have better organised, documented and clearer.

[...]

PR #1795, opened 2022-08-18
Improve resume pipeline suggestion

jmholzer:
## Description

Resolves [#1477](https://github.com/kedro-org/kedro/issues/1477)

## Development notes

After a failed run, Kedro suggests a command to the user:

`You can resume the pipeline run by adding the following argument to your previous command:
                               --from-nodes "node4_B"`

Before this PR, the suggested command will run from the last nodes to be executed, regardless of whether their input was persisted or not. If any of the inputs to the listed nodes is not persisted, the run immediately fails again.

After this PR, the suggested command will run from the closest successfully executed nodes with persisted inputs:

`You can resume the pipeline run from the nearest nodes with persisted inputs by adding the following argument to your previous command:
                               --from-nodes "node1_B,node1_A"`

This is achieved by performing a breadth-first search, starting at the last successfully executed nodes. This backward search yields a set of the nearest nodes that have persisted inputs.

Six tests are added to the `test_sequential_runner` test suite to test different cases on an X-shaped pipeline.

## Limitations
This change is a significant improvement, but there are still two important limitations:

- Persisted inputs are defined to be any that are not `MemoryDataSet`s. This definition has limitations; it does not account for custom datasets that are not persisted.
- Neither the approach in this PR nor the previous one handle the case where nodes append to datasets. Running these nodes repeatedly could have unintended consequences.

In the future, I think it would be a good idea to add a method to the API of AbstractDataSet that checks for persistence. I would love to hear thoughts on this.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1795"></a>


kedro/runner/runner.py
```diff
@@ -157,31 +158,121 @@ def create_default_data_set(self, ds_name: str) -> AbstractDataSet:
...
-        self._logger.warning(
-            "There are %d nodes that have not run.\n"
-            "You can resume the pipeline run by adding the following "
-            "argument to your previous command:\n%s",
-            len(remaining_nodes),
-            postfix,
-        )
+        if not postfix:
+            self._logger.warning(
+                "No nodes ran. Repeat the previous command to attempt a new run."
```
MerelTheisenQB, line 199: I'm not convinced this is a useful message.. In what scenario would `postfix` be empty?

kedro/runner/runner.py
```diff
@@ -157,31 +158,121 @@ def create_default_data_set(self, ds_name: str) -> AbstractDataSet:
...
+        node: the ``Node`` to check the inputs of.
+        catalog: the ``DataCatalog`` of the run.
+
+    Returns:
+        True if the ``Node`` being checked exclusively has inputs that
+        are not ``MemoryDataSet``, else False.
+
+    """
+    for node_input in node.inputs:
+        if isinstance(vars(catalog.datasets)[node_input], MemoryDataSet):
```
MerelTheisenQB, line 273: Why do you need to do `vars(catalog.datasets)`?

kedro/runner/runner.py
```diff
@@ -157,31 +158,121 @@ def create_default_data_set(self, ds_name: str) -> AbstractDataSet:
...
+    ancestor_nodes_to_run = set()
+    queue, visited = deque(children), set()
+    while queue:
+        current_node = queue.popleft()
+        visited.add(current_node)
+        if _has_persistent_inputs(current_node, catalog):
+            ancestor_nodes_to_run.add(current_node)
+            continue
+        for parent in _enumerate_parents(pipeline, current_node):
+            if parent not in visited:
```
MerelTheisenQB, line 238: Could there be a case where the parent node hasn't been visited but is already in the queue?

MerelTheisenQB (commented):
I've left some comments already, but need some more time to think about this.

This implementation works, but I'm wondering if it's the most efficient. You're building the queue in a way where you first check the remaining nodes (that haven't run yet), but I feel like you'd actually only want to check the parents because those are more likely to have persistent data. 

I'm also thinking that for really large datasets, we definitely don't want to check the whole pipeline and we should add in some mechanism that restricts how far up/down the tree we go.

kedro/runner/runner.py
jmholzer: If the first nodes in the pipeline fail, the postfix is empty. Before, this would result in the following message, which specifies no argument to add:
`You can resume the pipeline run by adding the following argument to your previous command:`
this should be fixed in my view.

kedro/runner/runner.py
jmholzer: catalog.datasets is an object of FrozenDatasets, which has no __getitem__ defined for it. It stores datasets in the __dict__ attribute, which can be accessed in a readable way by using vars().

The alternative would be catalog.datasets.__dict__, or to use the protected catalog._data_sets (which is already a dict) but linters won't allow these.

jmholzer:
Thanks for the review!
> This implementation works, but I'm wondering if it's the most efficient. You're building the queue in a way where you first check the remaining nodes (that haven't run yet), but I feel like you'd actually only want to check the parents because those are more likely to have persistent data.

Hmm, if I'm understanding this correctly, I don't think this is right. The queue starts with the furthest possible nodes that can be run given the failure plus the failing node. The remaining nodes (for which no run was attempted) are not checked at all.

> I'm also thinking that for really large datasets, we definitely don't want to check the whole pipeline and we should add in some mechanism that restricts how far up/down the tree we go.

This is a good idea, what would you think of me doing some time complexity analysis and estimating a rough order of magnitude of nodes for which the search would run in reasonable time?

MerelTheisenQB:
> Hmm, if I'm understanding this correctly, I don't think this is right. The queue starts with the furthest possible nodes that can be run given the failure plus the failing node. The remaining nodes (for which no run was attempted) are not checked at all.

Righttt.. this code is so complex! 😅 but I think you're right. So ignore my comment! 

> This is a good idea, what would you think of me doing some time complexity analysis and estimating a rough order of magnitude of nodes for which the search would run in reasonable time?

I'll have to think a bit more about this. Maybe if the pipeline is larger than X nodes we take a different strategy.

kedro/runner/runner.py
jmholzer: This is a very good question. The only scenario I can see is that the queue is initialised with a group of nodes where one is the parent of another, but I don't think the calling code allows for this.

Initialising `visited` with the same nodes as the queue would avoid this ever becoming a problem.

kedro/runner/runner.py
AntonyMilneQB: TIL! I've literally never used `vars` before so this is really interesting to hear! 👏 

That said, the way I would recommend doing this would actually be `._get_dataset(dataset_name)` and ignore any lint error it might give. This is a bit more obvious and would be consistent with[ how it's done in kedro-viz](https://github.com/kedro-org/kedro-viz/blob/1d7f37eab55fb12f596f5581997b9fbffefd8735/package/kedro_viz/data_access/repositories/catalog.py#L55).

AntonyMilneQB: P.S. yes, it is kind of ridiculous we don't have a nicer way of doing this. It's one thing I would love to improve: see the first bullet point in https://github.com/kedro-org/kedro/issues/1778.

tests/runner/test_sequential_runner.py
```diff
@@ -278,3 +279,77 @@ def test_confirms(self, mocker, pipeline, is_async):
         catalog = DataCatalog(data_sets={"ds1": fake_dataset_instance})
         SequentialRunner(is_async=is_async).run(pipeline, catalog)
         fake_dataset_instance.confirm.assert_called_once_with()
+
+
+@pytest.fixture
+def resume_scenario_pipeline():
```
AntonyMilneQB, line 285: Should we define this in conftest where the other pipeline fixtures are? Would be good to give it a more descriptive name also like `X_shaped_pipeline` (c.f. `branchless_pipeline`) and maybe a short docstring describing or picturing the structure (like `branchless_no_input_pipeline` has).

tests/runner/test_sequential_runner.py
```diff
@@ -278,3 +279,77 @@ def test_confirms(self, mocker, pipeline, is_async):
...
+    return Pipeline(
```
AntonyMilneQB, line 286:
```suggestion
    return pipeline(
```
very minor, but ideally all tests would use `pipeline` rather than `Pipeline` now (there's lots of other instances of this which seem to have been missed outside your code here - feel free to change them all).

tests/runner/test_sequential_runner.py
```diff
@@ -278,3 +279,77 @@ def test_confirms(self, mocker, pipeline, is_async):
...
+            "ds2_A": persistent_dataset,
+            "ds2_B": persistent_dataset,
+        }
+    )
+
+
+@pytest.mark.parametrize(
+    "failing_node_indexes,expected_pattern",
+    [
+        ([0], r"No nodes ran."),
```
AntonyMilneQB, line 326: Would it be possible to label these by node name rather than index? e.g. `["node1_A"]` here and then maybe `failing_nodes = pipeline.only_nodes(*failing_node_names)`. This way you don't have to understand how node index maps onto the pipeline topology.

tests/runner/test_sequential_runner.py
```diff
@@ -278,3 +279,77 @@ def test_confirms(self, mocker, pipeline, is_async):
...
+        ([2], r"(node1_A,node1_B|node1_B,node1_A)"),
+        ([3], r"(node3_A,node3_B|node3_B,node3_A)"),
+        ([5], r"(node3_A,node3_B|node3_B,node3_A)"),
+        ([3, 5], r"(node3_A,node3_B|node3_B,node3_A)"),
+        ([2, 5], r"(node1_A,node1_B|node1_B,node1_A)"),
+    ],
+)
+class TestSuggestResumeScenario:
```
AntonyMilneQB, line 334: Really clever test! This is top notch kedro-ing. ⭐

AntonyMilneQB (commented):
Wow, this looks awesome 💯 I'm going to need to take a look through the algorithm another time with a clearer head before I can comment on that, so just some general thoughts for now. Thank you for all the docstrings explaining it though - super helpful!

* limitation of defining a persisted input to be anything that's not `MemoryDataSet`: this is fine. Interesting idea to add something to `AbstractDataSet` that checks for persistence. How would this work? You manually set some class variable `persisted = False` for `MemoryDataSet` and `persisted = True` for others?
* limitation of not working for appending to datasets also seems fine. We probably need to rethink how we handle such cases and other SQL-type operations in datasets anyway, e.g. [delta tables](https://kedro.readthedocs.io/en/stable/tools_integration/pyspark.html#spark-and-delta-lake-interaction). But that's a big, difficult topic for another time... Possibly `persisted` would need to be more than `True/False` and also handle this 3rd sort of dataset?
* does this work for parallel runner? There's some funky stuff going on there with `SharedMemoryDataSet` so would be good to check it still works.

AntonyMilneQB:
Also, I don't know whether this would simplify things a lot or would be much worse. In https://github.com/kedro-org/kedro/issues/1802 I realised that we effectively define persisted datasets as those not defined in the `catalog.list()`. This is actually not quite as correct as checking `isinstance(..., MemoryDataSet)` as you do here, but might be worth considering.

kedro/runner/runner.py
MerelTheisenQB: Ah right, in that case this is definitely an improvement 👍

kedro/runner/runner.py
MerelTheisenQB: yeah that's probably very unlikely. I'm happy to leave it as is.

kedro/runner/runner.py
MerelTheisenQB: @AntonyMilneQB but isn't the problem here that you don't have the actual dataset names?

MerelTheisenQB: oh! yeah probably! I keep getting so confused by this part of the code base 😅

kedro/runner/runner.py
AhdraMeraliQB, line 238: As I understand it, this condition is to prevent nodes from being added to the queue several times over. Could this be extended to exclude parents that may already be in the queue already? For example, two nodes A and B that both share parents C and D, and a queue [A|B]; the current implementation would lead to a queue [C|D|C|D] with the set `visited` being {A,B}.

MerelTheisenQB:
@AntonyMilneQB raised a fair point:

> i'm not too fussed if we check back all the way through the pipeline in the case of a large pipeline. Would be interesting to do a quick stress test to get an idea of how long that takes for e.g. 10, 100, 1000 nodes, but I don't imagine it's prohibitive. This code only executes when an exception has been raised so it's not performance-critical, and kedro pipelines usually take a while to run so the extra time taken is likely to be very small relatively speaking. So if it's too much effort to do that time complexity analysis then I wouldn't worry about it and let's just wait to see if anyone complains that their error message takes too long to generate, in which case we can add some bound to look back only a certain depth. Unless the algorithm is very slow I highly doubt anyone will complain though.

And I guess this is also related to the task on getting data about how large Kedro projects are in reality. It would be great to see some complexity analysis, but I agree with Antony that if it's too much effort we should just leave it unless any users complain. It's hard to decide without any real numbers how we should bound look back, so I don't want to block having this merged just for those huge projects (which might turn out to just be a small percentage of all Kedro projects anyway).

noklam (commented):
Sorry for the delayed review as I have to spend some time to play with it. It works pretty well! 

image

Few issues that I notice (may not related to this change particularly)
1. When we don't have the `name` define in node, the suggestion doesn't work since the name are not the real node name (I think I hit this problem in notebook before, but it's out of the scope of this PR and should be addressed as a separate issue.)
2. The suggestion is hidden in the log before the `traceback` get printed, may be we could make it nicer so users don't need to scroll to see that?

AntonyMilneQB:
FYI @noklam point 1 that you raise is very long-standing issue. The topic of how to name nodes that don't have names explicitly given to them is something that a lot of research was done on before but not sure what the outcome was. We used to have a ticket for it in Jira but I think it didn't make it onto Github for some reason. Let me try and re-recreate it because it would be nice to get this resolved once and for all.

noklam:
@AntonyMilneQB Just want to point out that it's likely that there are some inconsistency buried in `pipeline` / `nodes` logic, which causes the `--from-node` argument in the CLI not have the same `node_name`. Even if the node name is ugly, i.e. `func(None) -> XYZ`, it should be a consistent name and it's not obvious to me why the CLI couldn't use it.

kedro/runner/runner.py
jmholzer: Thanks for the review Ahdra! You are right about this, nicely spotted. The improvement is in [836d583](https://github.com/kedro-org/kedro/pull/1795/commits/836d58364712e900dc794823e1eb8875beaf0ac5).

kedro/runner/runner.py
jmholzer: > That said, the way I would recommend doing this would actually be `catalog._get_dataset(dataset_name)` and ignore any lint error it might give.

Thanks for the feedback! Yeah ``vars()`` is quite uncommon I feel 😁. I made this change in [7c7e795](https://github.com/kedro-org/kedro/pull/1795/commits/7c7e79546c2739d034d03444a00dc0fdb6304414).

tests/runner/test_sequential_runner.py
jmholzer: I think you're right about this. I modified the names and moved my fixtures to conftest. I also moved the remaining fixtures in that file to conftest, in the spirit of leaving the codebase cleaner than when I found it. The remaining fixtures were from 2019 and came from a single contributor, while the more modern fixtures were all defined in conftest and were from multiple contributors, which is a good sign to me that this is the better approach.

These changes are made in [069a6fe](https://github.com/kedro-org/kedro/pull/1795/commits/069a6feede3bf53d7688aeb1a1511d58b9c3d3e9).

AntonyMilneQB: Perfect, thank you! Definitely sounds like an improvement to the codebase.

tests/runner/test_sequential_runner.py
jmholzer: Sure, I made this change for all instances of Pipeline in the file in [e4a6525](https://github.com/kedro-org/kedro/pull/1795/commits/e4a65257d878d8aa9776923c2c6d58ad25417610).

Where this created naming conflicts with variables / parameters, I renamed these `test_pipeline`.

tests/runner/test_sequential_runner.py
```diff
@@ -278,3 +279,77 @@ def test_confirms(self, mocker, pipeline, is_async):
...
+    )
+
+
+@pytest.fixture
+def resume_scenario_catalog():
+    def _load():
+        return 0
+
+    def _save(arg):
+        assert arg == 0
```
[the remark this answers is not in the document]
jmholzer: This is a holdover from the test I originally yanked the code from, my bad! Fixed in [e92a756](https://github.com/kedro-org/kedro/pull/1795/commits/e92a756fde94bbcac085c816859c449e694e518b).

jmholzer:
> 1. When we don't have the `name` define in node, the suggestion doesn't work since the name are not the real node name (I think I hit this problem in notebook before, but it's out of the scope of this PR and should be addressed as a separate issue.)

I'll open a separate issue for this.

> 2. The suggestion is hidden in the log before the `traceback` get printed, may be we could make it nicer so users don't need to scroll to see that?

I tried implementing this with a few approaches, using `sys.exc_info()` and various methods from `traceback` to print the traceback before logging the resume message and exiting the program. Unfortunately, this resulted in losing significant information from the traceback that is otherwise available, so I am reluctant to use these implementations. I would be happy to discuss this further, but for the moment I am out of ideas of how to implement this properly.

tests/runner/test_sequential_runner.py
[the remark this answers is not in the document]
noklam: Ah! I see.

noklam:
> @noklam
> 
> > 1. When we don't have the `name` define in node, the suggestion doesn't work since the name are not the real node name (I think I hit this problem in notebook before, but it's out of the scope of this PR and should be addressed as a separate issue.)
> 
> I'll open a separate issue for this.
> 
> > 2. The suggestion is hidden in the log before the `traceback` get printed, may be we could make it nicer so users don't need to scroll to see that?
> 
> I tried implementing this with a few approaches, using `sys.exc_info()` and various methods from `traceback` to print the traceback before logging the resume message and exiting the program. Unfortunately, this resulted in losing significant information from the traceback that is otherwise available, so I am reluctant to use these implementations. I would be happy to discuss this further, but for the moment I am out of ideas of how to implement this properly.

Thank you for trying! I think that's ok to leave it for now, as this is the same behavior we used to have, just that with `rich` logs now it's buried deeper and it's not as easy to spot the help message.

tests/runner/test_sequential_runner.py
jmholzer: Very good point, I have implemented this in [1bf5f1d](https://github.com/kedro-org/kedro/pull/1795/commits/1bf5f1de48ff9634b2c207212db0afb23f537b06).

AntonyMilneQB (approved):
Really great PR! ⭐  Thank you for doing that performance analysis. Completely agree with you that the performance hit here is perfectly acceptable for any pipeline of a realistic size. Having more than 1000 nodes between persisted datasets is not going to happen. In hindsight, given that people tend to over- rather than under-persist datasets, probably it's not even common to have pipelines with more than 10 such nodes. So all good here 👍 

Just one question from before that's left over 🙂 
> does this work for parallel runner? There's some funky stuff going on there with `SharedMemoryDataSet` so would be good to check it still works.

noklam (approved):
Great work tackling a tough issue! I don't have much concern since this feature wasn't doing much before, even if it doesn't work for all cases it will still be an improvement. But like Antony said it would be good to check if it works for other `Runner`.

jmholzer:
> Just one question from before that's left over but very happy to approve here! 🙂
> 
> > does this work for parallel runner? There's some funky stuff going on there with `SharedMemoryDataSet` so would be good to check it still works.

Thanks for the re-review! You're right I dropped this one somewhere, I'm sorry. I've been looking into this, I'll give an update when I've finished. `_SharedMemoryDataset` could be a problem.

AntonyMilneQB:
> Thanks for the re-review! You're right I dropped this one somewhere, I'm sorry. I've been looking into this, I'll give an update when I've finished. `_SharedMemoryDataset` could be a problem.

Cool, no worries. As @noklam says, if it doesn't work then it's not a showstopper. I'm happy to merge with it just working on sequential runner and we can fall back on using the previous inferior `_suggest_resume_scenario` for the parallel runner case if it's not easy to fix.

jmholzer:
Alright, I finished my investigation into `ParallelRunner`.

It is possible to implement the new scheme proposed in this PR for `ParallelRunner`, though it involves some workarounds due to `_SharedMemoryDataSet`.

Unfortunately, it isn't of much use, since the sequence in which nodes are run (and the resulting exception is reached) is not deterministic for ParallelRunner. This causes problems for both the new and the existing logic for generating suggestions. For example, with the existing logic a run with `ParallelRunner` will produce the message:
```
You can resume the pipeline run by adding the following argument to your previous command:
--from-nodes "node3_B,node4_A"
```
Another identical run will (stochastically) produce the message:
```
You can resume the pipeline run by adding the following argument to your previous command:
--from-nodes "node4_A"
```

Similar results are seen for the new logic implemented in this PR. One message is correct while the other isn't. Since these conflicting messages occur with roughly the same frequency, I don't think we should be suggesting a resume command at all at the moment for `ParallelRunner`. I think implementing this will first require that the order of execution of nodes is made deterministic, which is a large enough task to be a separate PR.

@noklam @AntonyMilneQB it would be good to hear your thoughts on this. If you agree with me, I will turn off this feature for `ParallelRunner` for the time being in a new commit, merge this PR and then write up an issue.

AntonyMilneQB:
That sounds like a perfect plan, thanks very much @jmholzer. Note that until recently the sequential runner was also not deterministic in the order of running nodes (something @noklam fixed). I don't know if the same sort of fix would be relevant for the parallel runner.

AntonyMilneQB (approved):

noklam:
I am happy that this is added just for `SequentialRunner`.

Note that there may be 2 sources of non-deterministic behavior:
1. Kedro itself order the nodes in a non-deterministic (Used to be the case with `SequentialRunner` due to some `set` operation) -> Distributed these nodes into subprocesses.
2. The nature of parallelism, the order of execution depends on if the computation is finished or not, so I think is non-deterministic by nature.

It's impossible to have deterministic nodes execution order for `ParallelRunner`, but there may be things that can be more deterministic for 1.

noklam (approved):

jmholzer:
Thanks for the feedback @noklam and @AntonyMilneQB! It's much appreciated.

@noklam thanks for the hint in 1. Regarding 2, you're right about this, the execution order is inherently indeterminate. Nonetheless I think we can at least reach a deterministic 'solution' (in this case, the correct warning) using join(s). I will open an issue and explain my thinking.

PR #1809, opened 2022-08-24
Catch PermissionError for versioned dataset

### LINKED ISSUE #1768: Catch error for issues with reading/writing data to object stores (e.g. S3)
The original Discord discussion https://discord.com/channels/778216384475693066/989568455512555540

# Context
> williamc — 06/23/2022
Let's say I just cloned my kedro project repo to another machine, and its datasets are versioned and configured to use S3 for storage. If I try to run a pipeline that depends on those datasets I get the infamous kedro.io.core.VersionNotFoundError . Bucket has versions all the way up to 2022-06-07T22.04.39.460Z/ and the error says 2022-06-23T16.20.52.945Z. Is this the intended behavior? Thanks

# Solution
User report that this is resolved after he changes the permission to `s3:*` from `s3:*Object`. I suspect we need  `ListObjects` or `ListObjectsV2` only.

# What's the problem?
I expect it to throw an error like `Insufficient permissions to list objects`, these logic is likely handled by `fsspec` and the `glob_function` we pass into `VersionedDataSet`. Some more investigation is needed

# Action:
1. Show a better error like `Insufficient permissions to list objects` instead of `VersionedNotFound`? This is confusing because the user can see the files but `kedro` would complain the file isn't there. It's not clear currently if the error doesn't come from Kedro directly but from e.g. fsspec
2. Check what kind of Policy are needed for `kedro` to work? (p.s. `*` is definitely not a good suggestion!) and document this clearly

ankatiyar:
Signed-off-by: ankatiyar <ankitakatiyar2401@gmail.com>

## Description
Resolves #1768 

## Development notes
This PR catches permission issues when trying to read/write data to object stores. This issue occurs when the `load_version` is not specified and needs to be resolved. `AbstractVersionedDataSet`'s `_fetch_latest_load_version` gets an empty list of files incase 1) No versions exist or 2) The user has entered wrong credentials for the object store. This led to `VersionNotoundError` being thrown in both cases. 

### Proposed solution 
`fsspec` has an inbuilt `PermissionError`. 
- For loading a versioned dataset without proper credentials :
In `AbstractVersionedDataSet`'s `load` function, perform `exists` check before `resolve_load_version` which will throw `PermissionError` if the credentials are not proper.
- For saving a versioned dataset without proper_credentials:
Catch `PermissionError` and display relevant error message.

### Changes to test files 
Added mock calls to patch `_filesystem_exists` to avoid actually calling the function on dummy datasets and failing.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1809"></a>


kedro/io/core.py
```diff
@@ -617,6 +623,10 @@ def save(self, data: _DI) -> None:
                 f"'{self._filepath.as_posix()}/{_default_version}/{self._filepath.name}"
                 f"')."
             ) from err
+        except PermissionError as err:
```
datajoely, line 626: TIL this was a built-in exception!

kedro/io/core.py
```diff
@@ -595,6 +595,13 @@ def _get_versioned_path(self, version: str) -> PurePosixPath:
         return self._filepath / version / self._filepath.name
 
     def load(self) -> _DO:
+        try:
+            if hasattr(self, "_fs"):
+                self._fs.exists(str(self._filepath))
+        except PermissionError as err:
+            raise DataSetError(
+                f"Cannot load versioned dataset '{self._filepath}' due to insufficient permission."
+            ) from err
```
noklam, line 604: Could you explain how this works? The original error `VersionNotFound` when it does not has sufficient permission. Would this error `PermissionError` be triggered?

noklam (commented):
Thanks for the PR. How did you setup the S3 and test with different policy? Could you also update the documentation to include what policies are needed for `kedro` to work properly?

kedro/io/core.py
```diff
@@ -594,7 +594,20 @@ def _get_save_path(self) -> PurePosixPath:
...
+                return True
+        return False
+
     def load(self) -> _DO:
+        try:
+            self._filesystem_exists(str(self._filepath))
+        except PermissionError as err:
+            raise DataSetError(
+                f"Cannot load versioned dataset '{self._filepath}' due to insufficient permission."
+            ) from err
```
noklam, line 610: Does this handle the case when `load_version=None`?

tests/extras/datasets/pandas/test_excel_dataset.py
```diff
@@ -151,6 +151,7 @@ def test_protocol_usage(self, filepath, instance_type, load_path, mocker):
         assert isinstance(data_set._filepath, PurePosixPath)
 
         mock_pandas_call = mocker.patch("pandas.read_excel")
+        mocker.patch("kedro.io.core.AbstractVersionedDataSet._filesystem_exists")
```
noklam, line 154: The test is for testing the DataSet implementation so it doesn't feel right to mock the implementation. What's the error if you didn't mock it?

ankatiyar: The datasets used for these tests are dummies and in the original version of the tests, the actual `pandas.read_excel` (and the respective read calls for the different pandas datasets) calls are mock calls. These tests were failing due to the changes in this PR because the `_filesystem_exists()` call tries to check if these datasets actually exist.

tests/framework/cli/test_jupyter.py
```diff
@@ -141,7 +141,12 @@ def test_create_new_kernel(self):
         assert kernel_spec.language == "python"
         assert kernel_spec.argv[-2:] == ["--ext", "kedro.ipython"]
         kernel_files = {file.name for file in Path(kernel_spec.resource_dir).iterdir()}
-        assert kernel_files == {"logo-32x32.png", "logo-64x64.png", "kernel.json"}
+        assert kernel_files == {
```
AhdraMeraliQB, line 144: Changed this test to reflect one of the updated dependencies

tests/framework/cli/test_jupyter.py
```diff
@@ -141,7 +141,11 @@ def test_create_new_kernel(self):
         assert kernel_spec.language == "python"
         assert kernel_spec.argv[-2:] == ["--ext", "kedro.ipython"]
         kernel_files = {file.name for file in Path(kernel_spec.resource_dir).iterdir()}
-        assert kernel_files == {"logo-32x32.png", "logo-64x64.png", "kernel.json"}
+        assert kernel_files >= {
```
AhdraMeraliQB, line 144: Change of plans - the version bump that caused this test to fail wasn't affecting all the test setups, just the Linux and Windows 3.10. Relaxing the constraints for now to get all tests passing.

MerelTheisenQB (approved):
This looks good to me and thank you for adding the tests @AhdraMeraliQB 😄 ⭐ 

Just one small thing to double check: Nok asked earlier if "this handle[s] the case when load_version=None?", is this the case? And should we perhaps add a test for it?

AhdraMeraliQB:
@MerelTheisenQB 
> Just one small thing to double check: Nok asked earlier if "this handle[s] the case when load_version=None?", is this the case? And should we perhaps add a test for it?

The current test setup has the dataset version as None. In both load and save the dataset's version only affects `self.resolve_load_version()` which is only called after the PermissionError is thrown - if the execution gets to that point then permissions should be sufficient and it should either load/save normally or throw the correct VersionNotFoundError

noklam:
This is my current understanding, my intuition is this doesn't solve the original issue and 

1. When permission policy is incorrect and `load_version=None`, it will resolved in `VersionedNotFound` error instead of `PermissionsError` - so none of the changes will address this.
3. Why do we need to catch this `PermissionError`? How does previous logic handle this? Would be great to see a before/after comparison, I think the `DataSetError` should have captured the `PermissionError`, but not 100% sure.

I am happy to pair on this or do some manual testing locally, @AhdraMeraliQB do you know how to set up locally?

AhdraMeraliQB:
> I am happy to pair on this or do some manual testing locally, @AhdraMeraliQB do you know how to set up locally?

Let's set up a real-time session

noklam:
Just notice that this branch was created under `ankatiyar`, I this should be created under `kedro` as a feature branch.

noklam:
Document some more as I tried to dive into `fsspec` to understand how these protocols work exactly.

`fsspec` has a mapping for all the protocols stored in a dictionary `known_implementations` see https://github.com/fsspec/filesystem_spec/blob/bb9989ce5bf0ed0c0a5f7d3540c3a59581d259ce/fsspec/registry.py#L88-L95

So there are actually 2 places to look for the protocols.
1. `known_implementations`
2. `entrypoints` - see `ablfs` add some new protocols via `setup.py` here https://github.com/fsspec/adlfs/blob/591485b9d77448cd6e791b49bda8942ef03507bf/setup.py#L44-L48

ankatiyar:
Closing this in favour of https://github.com/kedro-org/kedro/pull/1881 
Taking a different approach.

PR #1814, opened 2022-08-26
Add pip freeze stage to docs build CI

AntonyMilneQB:
## Description
Following the recent Sphinx CI saga, I've added a `pip freeze` stage to the process so that it's easier to troubleshoot in future.

I've slightly rejigged things so that `make build-docs` is a bit faster (no longer tries to re-install `test_requirements.txt`; removed redundant ipython thing).

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1814"></a>


AhdraMeraliQB (approved):
Awesome! This was definitely much needed 😬

noklam (approved):
Thanks, it would be nice if `pip freeze` happens before the doc gets built but after the libraries are installed. If something breaks during the make-docs, we won't see which `sphinx-xxx` libraries are installed as this `pip freeze` won't be executed.

I agree it is cleaner to move the `pip install` to `Makefile` and remove the `test_requirements.txt`

```
	pip install -e ".[docs]"
        # pip freeze
	./docs/build-docs.sh "docs"
```

noklam:
It would be nice if `pip freeze` happens before the doc gets built but after the libraries are installed. If something breaks during the make-docs, we won't see which `sphinx-xxx` libraries are installed.

```
	pip install -e ".[docs]"
        # pip freeze
	./docs/build-docs.sh "docs"
```

AntonyMilneQB:
@noklam Ahdra had exactly the same thought when we were discussing this earlier. This is why there's `when: always` in the `pip freeze` step. This means that it will always be run, even if `make build-docs` crashes 🙂

noklam (approved):

noklam:
@AntonyMilneQB Got it! Thanks for explaining.

PR #1821, opened 2022-08-30
Support `find_pipelines()` for simplified projects

### LINKED ISSUE #1812: Consider whether `find_pipelines` should handle a simplified project structure
Following a conversation in https://github.com/kedro-org/kedro/pull/1706. Just creating the issue while I remember. I'll write more details when I get the chance.

1. Should we support the "simplified" project structure? I think probably yes. If no, we should change the starters that use this.
2. If yes, should `find_pipelines` support this simplified project structure? I think probably yes. 
3. Should `find_pipelines` support both the simplified and extended project structure in a single project? Probably not very important either way.

deepyaman:
## Description

Close #1812 

## Development notes

Still need to update documentation; will do so after the broader pipeline registry documentation is merged.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1821"></a>


noklam (commented):
Thank you! It looks great overall! I have dropped some minor comments.

kedro/framework/project/__init__.py
```diff
@@ -283,7 +315,28 @@ def find_pipelines() -> Dict[str, Pipeline]:
...
     """
-    pipelines_dict = {"__default__": pipeline([])}
+    pipeline_module = None
+
+    # Handle the simplified project structure found in several starters.
+    pipeline_module_name = f"{PACKAGE_NAME}.pipeline"
+    try:
+        pipeline_module = importlib.import_module(pipeline_module_name)
+    except ModuleNotFoundError as exc:
+        if str(exc) != f"No module named '{pipeline_module_name}'":
```
AntonyMilneQB, line 325: I'm not a big fan of string matching exception messages tbh - it feels kind of hacky. Or is this a well-accepted pattern? Can't we just see if the file `{PACKAGE_NAME}/pipeline.py` exists (or maybe there's an `importlib` way to check without actually doing the import? like how you do `importlib_resources.files`) and then only try to `import_module` on it if it does?

AntonyMilneQB: Possible alternative: maybe put this `if` into the general `except` clause and remove this `except` block? It just doesn't feel quite right to me as it stands.

kedro/framework/project/__init__.py
```diff
@@ -283,7 +315,28 @@ def find_pipelines() -> Dict[str, Pipeline]:
...
+            )
+    except:  # pylint: disable=bare-except  # noqa: E722
+        warnings.warn(
+            IMPORT_ERROR_MESSAGE.format(
+                module=pipeline_module_name, tb_exc=traceback.format_exc()
+            )
+        )
+
+    pipelines_dict = {"__default__": _create_pipeline(pipeline_module) or pipeline([])}
+
```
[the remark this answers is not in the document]
AntonyMilneQB: I think this makes sense 👍 Let's think of the (probably unusual) case where we have both `pipeline.py` and a `pipelines` directory. In this case `pipeline.py` basically defines your `__default__` pipeline, and `pipelines` defines all the others. I think the behaviour in these cases should be the same:
* non-existent `pipeline.py` or `pipelines` directory should just be ignored and not flag any warnings
* any exceptions, including an `ImportError` caused by a missing dependency, inside `pipeline.py` or `pipelines` should flag warning just be skipped

Is that how things work now?

kedro/framework/project/__init__.py
```diff
@@ -283,7 +315,28 @@ def find_pipelines() -> Dict[str, Pipeline]:
...
     for pipeline_dir in importlib_resources.files(
         f"{PACKAGE_NAME}.pipelines"
     ).iterdir():
```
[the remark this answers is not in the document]
AntonyMilneQB: Also, just to check: what happens when both `pipeline.py` and `pipelines` directory don't exist? I think we should just have `__default__: pipeline([])` with no warnings, right?

kedro/framework/project/__init__.py
```diff
@@ -283,7 +315,28 @@ def find_pipelines() -> Dict[str, Pipeline]:
...
+                )
+            )
+    except:  # pylint: disable=bare-except  # noqa: E722
+        warnings.warn(
+            IMPORT_ERROR_MESSAGE.format(
+                module=pipeline_module_name, tb_exc=traceback.format_exc()
+            )
+        )
+
+    pipelines_dict = {"__default__": _create_pipeline(pipeline_module) or pipeline([])}
```
AntonyMilneQB, line 338: Very minor, but it might be nice to make this a bit more consistent with the `pipelines` directory logic on L361 etc.

```suggestion
    except:  # pylint: disable=bare-except  # noqa: E722
        warnings.warn(
            IMPORT_ERROR_MESSAGE.format(
                module=pipeline_module_name, tb_exc=traceback.format_exc()
            )
        )
    else:
         pipeline_obj = _create_pipeline(pipeline_module)
    
    pipelines_dict = {"__default__": pipeline_obj or pipeline([])}
```
and set `pipeline_obj = None` on L318 instead of `pipeline_module`.

AntonyMilneQB (commented):
This looks great, thanks so much for adding it! Just a few small comments.

kedro/framework/project/__init__.py
deepyaman: > like how you do `importlib_resources.files`) and then only try to `import_module` on it if it does?

`importlib_resources.files` actually does `importlib.import_module` under the hood to make sure that the package passed to `files` is actually importable; I was actually about to go add another instance of string matching there! Let me see...

kedro/framework/project/__init__.py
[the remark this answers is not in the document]
deepyaman: That should be how they work now, after I pushed a fix to allow a non-existent `pipelines` directory.

kedro/framework/project/__init__.py
deepyaman: Went ahead and merged the `except` clauses, but going to keep the string matching, as discussed. Trying to load based on folders seems overly complicated, because it would involve calling `bootstrap_project`.

kedro/framework/project/__init__.py
deepyaman: Nice suggestion; done!

kedro/framework/project/__init__.py
```diff
@@ -270,6 +276,32 @@ def validate_settings():
     importlib.import_module(f"{PACKAGE_NAME}.settings")
 
 
+def _create_pipeline(pipeline_module: Optional[types.ModuleType]) -> Optional[Pipeline]:
+    if pipeline_module is None:
+        return None
```
AntonyMilneQB, line 281: Is this case still relevant? Does `_create_pipeline` ever get called with `pipeline_module = None` after the changes you made yesterday?

AntonyMilneQB (approved):
This looks great, thank you for all your work on it! 🙏

kedro/framework/project/__init__.py
deepyaman: Shouldn't happen! Let me remove this and update the function signature.

deepyaman:
Tested on GitPod with the default project included therein (simplified structure) and Spaceflights starter (with `pipelines` directory).

tests/framework/project/test_pipeline_discovery.py
```diff
@@ -162,9 +161,100 @@ def test_find_pipelines_skips_modules_that_cause_exceptions_upon_import(
...
+                    return pipeline([node(lambda: 1, None, "simple_pipeline")])
+                """
+            )
+        )
+
+    configure_project(mock_package_name_with_pipelines)
+    pipelines = find_pipelines()
+    assert set(pipelines) == {"__default__"}
+    assert sum(pipelines.values()).outputs() == (
+        {"simple_pipeline"} if simplified else {}
```
deepyaman, line 259:
```suggestion
        {"simple_pipeline"} if simplified else set()
```
I'm so bad at this...

noklam (approved):
Looks good! Test on Gitpod

1. With simplified structure
2. Without any recognized structure -> generated empty `Pipeline`

One minor issue with `warnings.warn`, it generates a weird `warnings.warn(` suffix in the log message, but it's super minor so let's deal with it separately.

Issue #712, opened 2021-03-05
Simplify api heirarchy


WaylonWalker:
## Description

When using kedro as a user it can sometimes be hard to remember where even the most common kedro features are located.  

## Context

This will lower the barrier to entry to using kedro, and make it easier to use without needing to use the docs for even the most simple task such as "How do I make a node".

## Possible Implementation

To simplify the barrier to entry into kedro import the most commonly used features right into main.

**WARNING** - for the person browsing through, this is not the real api, this is simply a suggestion.

``` python
import kedro as kd

my_pipeline = kd.Pipeline([kd.node(create_cars, None, "cars")])
```

Alternatively users can also use the same api as so.

``` python
from kedro import Pipeline, node

my_pipeline = Pipeline([node(create_cars, None, "cars")])
```

At a minimum I feel that the very core features that every kedro user needs should be at the top of the api.

* Pipeline
* node

The next priority would be those used by hooks authors.

* KedroSession
* ConfigLoader

Finally the full scope of what should be considered to be made easy to reach should be every occurance of `from kedro import x` in the project template.

limdauto:
I think this is a great idea. 

Just thinking out loud here: one thing will land in 0.17.2 is the ability to import user-defined `pipelines` as follows:

```
from kedro.framework.project import pipelines
```

And that will be bounded with the pipelines defined by users in their project. Hopefully this will be the way forward to access other user-defined values such as `catalog` and `config_loader` as well (we are still looking for use cases to support making this move). We do have a big use case that needs `pipelines` to be importable though. 

So essentially there will 2 things you can to import from `kedro`:

* Things you define in your project, e.g. `settings`, `pipelines`
* Library components define by `kedro`, e.g. `Pipeline`, `node()`, `pipeline()`, etc.

For the first one, importing from `kedro.framework.project` makes sense because it reads as "importing pipelines from my project". I wonder if we can come up with a similar API for the second one?

WaylonWalker:
Do other frameworks do this?  I feel that its a bit confusing to get project defined objects from kedro rather than `from my_project import settings, pipelines`  What is the benefit of importing `from kedro` rather than `from my_project`.  I do not have any familiarity with a framwork that behaves this way, so it might be more common than I realize.

mzjp2:
> Do other frameworks do this?

Django is a (pretty big) example that I've seen work really well. You define your settings in `project/setting.py` and then use them in the rest of your app by doing:

```python
# project/settings.py
[...]

PR #1837, opened 2022-09-07
Alias Kedro IPython extension to kedro.ipython

### LINKED ISSUE #1763: Alias IPython extension

AntonyMilneQB:
## Description
Closes #1763.

With many apologies to @noklam for the original work done in #1794 trying to alias `kedro.extras.extensions.ipython` to just `kedro`, it turns out to be much better to alias as `kedro.ipython` as we originally thought. The pattern `%load_ext name.ipython` seems to be about as common as `%load_ext name`. Its only disadvantage is a few extra characters. The main advantage is that it feels cleaner, there's no breaking import linter contracts, and there's no danger that importing kedro could possibly import IPython or have other undesired side-effects like setting up logging.

> **Note**. We should delete `kedro/extras/extensions/` from `develop` after this has merged.

## Note to reviewers
I recommend reviewing this commit by commit since I moved a lot of code from one file to another. This will make it clear when it's just code that's moved or where I've actually changed something. The only real change to code itself is in https://github.com/kedro-org/kedro/pull/1837/commits/637053e2b7af505f17e0c5154875c4a20e450d94, where the definition of the line magic and imports and moved to top level.

## Development notes

Tested:
- [x] the new alias works
- [x] backwards compatibility of `kedro.extras.extensions.ipython`
- [x] packaging kedro still includes the logo png files thanks to MANIFEST.in
- [x] deprecation warning shows: image

Note that doing the imports this way round (`kedro.extras.extensions.ipython` imports `kedro.ipython` and not vice versa) is much neater because it will allow us to remove the file from `develop` straight away. This way, as we modify the IPython extension going forward, the file we'll be modifying is the new correct one rather than the old deprecated one.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1837"></a>


noklam (approved):
Thanks! I think `%load_ext kedro.ipython` is a better solution.

After hustling with the original proposal `%load_ext kedro`, I think we should be very careful about any top-level import alias i.e. `import kedro`. I am now more cautious about the idea in #712, due to the way that Python Import system works, it's very dangerous to put things in the top-level. It now make more sense to me why Django do `django.shortcuts` instead of just `django`.

MerelTheisenQB (approved):
Great work and thanks for explaining so clearly why just `kedro` as alias isn't such a good solution after all @noklam @AntonyMilneQB 👍 

Should we keep one test to check `%load_ext kedro.extras.extensions.ipython` still works until it's deprecated?

AntonyMilneQB:
@MerelTheisenQB good thinking, done in [645ce1f](https://github.com/kedro-org/kedro/pull/1837/commits/645ce1faa73587c207c076c98f2dbf5531e5b33c)!

Actually I've just found out that most of the tests in this file haven't really doing anything useful for a long time. The mocking is all wrong and so not testing much... I'm going to make a new ticket to fix this up properly.

noklam:
I found `test_logging.py` and `test_pipeline_registry.py` interfere with `test_ipython.py`, I am slightly worried that this is due to the manual deletion of `sys.modules` and it has some unexpected effect.

`pytest tests/ipython/test_ipython.py tests/framework/project/test_logging.py ` will pass
`pytest tests/framework/project/test_logging.py  tests/ipython/test_ipython.py ` will fail

tests/ipython/test_ipython.py
```diff
@@ -62,21 +63,13 @@ def my_register_pipeline():
             return_value=my_register_pipeline,
         )
         mocker.patch("kedro.framework.startup.configure_project")
```
noklam, line 65:
```suggestion
        mocker.patch("kedro.ipython.configure_project")
```

This is the wrong patch that causing most of the confusion.

AntonyMilneQB: Ahh, if I change this then can I remove the `no pragma` and the tests pass?

noklam: Let's address this altogether at #1839

PR #1848, opened 2022-09-12
Fix micro package pull from PyPI

### LINKED ISSUE #1811: kedro micropkg pull from pypi is not working in 0.18.2
## Description

A micro-packaged pipeline which is uploaded to a PyPi repository can not be successfully pulled.

We have been using kedro very actively, so thanks for your support with this issue!

## Context
I was trying to push and pull a micro package via a private PyPi repository to share and reuse pipelines.

## Steps to Reproduce
1. kedro pipeline create <pipeline_name>
2. kedro micropkg package pipeline.<pipeline_name>
3. twine upload -dist/* 
4. kedro micropkg pull <pipeline_name>

## Expected Result
The micro-package should be downloaded via pip and successfully pulled via `kedro micropkg pull`

## Actual Result
The micro-package is downloaded via pip but there seems to be an isse here:
https://github.com/kedro-org/kedro/blob/4843fadfc887c410bb03ba3e54f6737019eeef3b/kedro/framework/cli/micropkg.py#L147

`Path(package_path).name` is expected to be a tar.gz file, but in my case it is actually the name of the pip package that was downloaded. As a result `egg_info_file` will be an empty list.

This is the error:

```
kedro.framework.cli.utils.KedroCliError: More than 1 or no egg-info files found from <pipeline_name>. There has to be exactly one egg-info directory.
```

I can bypass the issue by running this:
1. `python -m pip download --no-deps --dest <destination>  <pipeline_name>`
2. Copying the tar.gz path of the downloaded package `<micro-package-path>`
3. ` kedro micropkg pull <micro-package-path>`

## Your Environment

* Kedro version used (`pip show kedro` or `kedro -V`): 0.18.2
* Python version used (`python -V`): 3.9.12
* Operating system and version: Ubuntu 20.04.3 and MacOS 12.5

FlorianGD:
## Description
This PR fixes a bug when installing micro packages from PyPI

## Development notes
The way the package name was computed relied on the assumption that the package path could compute the name of the folder where the egg info file is. This is not true if the package is downloaded from PyPI

## Checklist

Fix #1811 

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1848"></a>


FlorianGD:
I believe the failing tests came after the merge of `main`. They seem Windows related, which I cannot easily test locally. The weird thing is that it passed on windows for python 3.9.

I am not sure what I should do

SajidAlamQB (commented):
Thanks for picking this up, your solution looks good! I am thinking, do we need to update some of the other PyPI tests as well?

FlorianGD:
I do not think other tests need to be updated. The tests regarding PyPI are:
- `test_invalid_pull_from_pypi`
- `test_pull_from_pypi_more_than_one_sdist_file`
- `test_pull_from_pypi`

The first 2 should fail before reaching the changed part of the code (in `_unpack_sdist`). The last one is updated.

MerelTheisenQB (approved):
Thanks for investigating and fixing this @FlorianGD ! 👍 

I wonder if we could have caught this earlier.. I guess our assumption that package name on PyPi has the version was wrong looking at the test change?

FlorianGD:
Yes, I think that this assumption was indeed incorrect. Maybe having an integration test on https://test.pypi.org would have caught it also?

SajidAlamQB (approved):

Issue #1528, opened 2022-05-13
Unable to specify save format for SparkHIveDataSet


jstammers:
## Description
The implementation for `SparkHiveDataSet` allows the user to specify additional save arguments. This should enable a delta table to be saved which is done using the following pyspark code 

```python
table = spark.table(...)
table.write.saveAsTable("db.table", format='delta')
```

Which should be replicated using the following configuration

```yaml
table:
  type: spark.SparkHiveDataSet
  database: db
  table: table
  write_mode: overwrite
  save_args:
    format: delta
```

However, this raises a `DataSetError` because the `SparkHiveDataSet` constructor gets the format from the `save_args`
https://github.com/kedro-org/kedro/blob/805da3279e4ff2228094aabe9adfb05d9deebd85/kedro/extras/datasets/spark/spark_hive_dataset.py#L117

but if it exists, does not remove it from `self._save_args`. The error is raised when creating a hive table because there are two arguments named `format`
https://github.com/kedro-org/kedro/blob/805da3279e4ff2228094aabe9adfb05d9deebd85/kedro/extras/datasets/spark/spark_hive_dataset.py#L144-L151

## Context
I am trying to save a table using the delta format which is possible using the pyspark API, but currently not supported using `SparkHiveDataSet`. With the current implementation, the only supported format is the `'hive'` default.

A possible solution would be to `pop` the 'format' value if it exists in save_args, e.g.

```python
 self._format = self._save_args.pop("format", "hive")  #returns "hive" if "format" not in self._save_args.keys()
```

## Your Environment
Include as many relevant details about the environment in which you experienced the bug:

* Kedro version used (`pip show kedro` or `kedro -V`): 0.18.0
* Python version used (`python -V`): 3.9
* Operating system and version: ubuntu 18.04

MerelTheisenQB:
Hi @jstammers, have you tried using the [DeltaTAbleDataSet](https://kedro.readthedocs.io/en/stable/kedro.extras.datasets.spark.DeltaTableDataSet.html)?

MerelTheisenQB:
Hi @jstammers do you still need help with this?

jstammers:
Hi @MerelTheisenQB, I've been able to save my dataset as a delta table using the DeltaTableDataSet.

If you think this functionality should be available using the HiveDataSet class, I'd be happy to submit a PR that implements the change I proposed above. 

Otherwise, please feel free to close this issue and thanks for the help

[...]

PR #1857, opened 2022-09-20
Fix issue with specifying format for SparkHiveDataSet

jstammers:
## Description
The current implementation of `SparkHiveDataSet` contains a bug that prevents a user from specifying a save format. This issue is discussed in #1528 .

## Development notes
The proposed fix for this is to `pop` the format argument if it exists.
A unit test has also been added to assert that the dataset has the required format and that the table can be saved.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1857"></a>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -39,6 +39,7 @@
 * The Kedro IPython extension now surfaces errors when it cannot load a Kedro project.
 * Relaxed `delta-spark` upper bound to allow compatibility with Spark 3.1.x and 3.2.x.
 * Added `gdrive` to list of cloud protocols, enabling Google Drive paths for datasets.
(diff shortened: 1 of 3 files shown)


noklam (approved):
Thanks! I think this is the correct fix. It's also more consistent how we handle `eager_checkpoint`.

The original issue is `format` argument could appear twice in this function.

https://github.com/kedro-org/kedro/blob/d6bdfc11391c3d9e772f44996c5910649555ae38/kedro/extras/datasets/spark/spark_hive_dataset.py#L144-L151

MerelTheisenQB (approved):
Thanks for fixing this @jstammers 👍 ⭐

Issue #1810, opened 2022-08-25
Use dependabot to bump dependencies


antonymilne:
In https://github.com/kedro-org/kedro/pull/1805 it was generally agreed that we should use dependabot to automatically bump our dependencies. To be honest I don't know why we didn't just do this ages ago - it will hopefully make our lives way easier. We have a good enough test suite that we can rely on it to tell us whether it's safe to bump versions or not.

My original plan was to do it on kedro-viz first, which is a bit simpler in terms of requirements: https://github.com/kedro-org/kedro-viz/issues/814. And then, if it seems to work well for us there, introduce it on kedro also.

However, I'd be happy to try it out on kedro first. I'm not sure exactly how dependabot works so don't know which dependencies exactly it would be able to track since we have them spread in a few places (requirements.txt, test_requirements.txt [mainly needed for `kedro-datasets`], setup.py). Highest priority is to get it working on requirements.txt since those are the core requirements.

> **Warning**. Make sure that dependabot does not bump the major version (breaking changes) automatically for dataset dependences.

datajoely:
I would like to push that we also get into the habit of doing smaller, more frequent patch releases for this sort of thing.

SajidAlamQB:
Dependabot monitors these files in kedro: 

- pyproject.toml
- requirements.txt
- test_requirements.txt
- features/windows_reqs.txt
- setup.py

I've been testing dependabot on a forked kedro repo you can view the dependency graph here: https://github.com/SajidAlamQB/kedro/network/dependencies

Currently dependabot opens 46 PRs for kedro which you can view here: https://github.com/SajidAlamQB/kedro/network/dependencies

SajidAlamQB:
Dependabot remarks:

Dependabot version updates for Kedro will help us bump dependencies automatically but has some limitations. With the work done on kedro-datasets and the https://github.com/kedro-org/kedro/issues/1498, we decided to simplify dependabot to just requirments.txt. Due to the https://github.com/dependabot/dependabot-core/issues/4364, we can't tell it what files it can access explicitly so we get around this by limiting which directory dependabot can scan. With this in mind, we have moved requirments.txt from root into dependency/requirments.txt forcing dependabot to only scan in this directory.

[...]

Issue #1498, opened 2022-05-04
Check if improvements are needed for `test_requirements.txt` setup


merelcht:
Check if any of the below is still needed now we've implemented the `pyproject.toml` setup + datasets are removed:

- [x] Try and figure out a nicer way of installing all the dependencies listed in the `setup.py` file rather than needing to maintain a separate `test_requirements.txt` file there -> `test_requirements.txt` doesn't exist anymore. Instead of this do:
   - [x] Remove `setup.py`.
 - [x] Clean up kedro's test requirements to remove requirements for datasets that aren't needed any more

sbrugman:
`test` could be another `extras_require`. The `tests_require` now used in `setup.py` seem not to be part of the `pyproject.toml` metadata ([ref]), so this move would be a step in that direction (https://github.com/kedro-org/kedro/issues/1724).

deepyaman:
> `test` could be another `extras_require`. The `tests_require` now used in `setup.py` seem not to be part of the `pyproject.toml` metadata ([ref]), so this move would be a step in that direction (#1724).

+1 on this approach; some packaging frameworks (like `flit`) have this special `test` extra (which I guess lends credibility to the approach?), and I've also used it on some other projects personally.

astrojuanlu:
Note that in #2475 `tests_requires` was removed, there's only `test_requirements.txt`. #2152 was partially addressed, so at least now even if there's a `setup.py` for some dynamic stuff, most of the metadata is in `pyproject.toml`. We could definitely add another `optional-dependencies` group (former "extras").

astrojuanlu:
Reopening, still some work to be done

merelcht:
Completed in https://github.com/kedro-org/kedro/pull/3144 and https://github.com/kedro-org/kedro/pull/3143

PR #1862, opened 2022-09-21
Create dependabot.yml configuration file for version updates

SajidAlamQB:
## Description

Related Issue: https://github.com/kedro-org/kedro/issues/1810

This PR aims to add Dependabot Version updates for Kedro. 
Dependabot will monitor only these files in kedro:

- requirements.txt

As this is the initial PR for dependabot it has identified 10 different dependencies that need updating. As discussed with @rashidakanchwala the configuration for dependabot will be initially set to daily checks with a limit of 50 PRs that the bot can open. Follow-up PRs will be made to change the configuration into a weekly check for updates and remove the maximum PR limit.

> **Note**
> When this PR is merged 10 new PRs will be opened by dependabot.

## Development notes

To keep things similar between framework and viz a branch, `dependency-update`, will be created and used by dependabot. We will also keep the `dependency` label consistent.

I have tested this out on a forked kedro repo see the result here: https://github.com/SajidAlamQB/kedro-3/pulls

After some testing on starters, I was unable to get dependabot working properly. Dependabot can only run version updates on manifest files if it can access all of the dependencies in the file, even if you add inaccessible dependencies to the ignore option of your configuration file. In our starters requirments.txt we have kedro[pandas.CSVDataSet]~={{ cookiecutter.kedro_version }} dependency which dependabot couldn't parse hence it is failing.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1862"></a>


AntonyMilneQB:
Thanks for the explanation @SajidAlamQB! At a glance it looks like the vast majority of these updates are for dataset requirements - is that right? 

Given that all these requirements would move to `kedro-datasets` I wonder whether we should just make those changes on that repo instead. I'm not sure exactly when we're going to remove the datasets tests (and hence the requirements) from this repo though.

There's also a problem here that the bumps to dataset requirements are just altering test_requirements.txt and not the ones in setup.py which are the ones that actually matter to users e.g. https://github.com/SajidAlamQB/kedro/pull/2. We have a ticket to clean up this setup but again I'm not sure when we'll be doing it: https://github.com/kedro-org/kedro/issues/1498.

Ideally I'd say that here we just limit dependabot to looking at requirements.txt and not test_requirements.txt for now, and deal with the dataset dependencies separately. Problem is that from what @rashidakanchwala said there didn't seem to be a good way to do that... Maybe there's something clever we could do with `ignore` though? https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file I don't even see documented anywhere how dependabot works out which files to try and modify - can you? Does looking at their source code make this clearer?

@MerelTheisenQB thoughts?

MerelTheisenQB:
You raise a lot of great points here @AntonyMilneQB !

> Given that all these requirements would move to `kedro-datasets` I wonder whether we should just make those changes on that repo instead. I'm not sure exactly when we're going to remove the datasets tests (and hence the requirements) from this repo though.

I agree it would be better to make the changes on `kedro-datasets`, however, we're only fully removing the datasets in `0.19.0` so in the meantime we'll need to bump dataset requirements on Kedro as well. 

> There's also a problem here that the bumps to dataset requirements are just altering test_requirements.txt and not the ones in setup.py which are the ones that actually matter to users e.g. [SajidAlamQB#2](https://github.com/SajidAlamQB/kedro/pull/2). We have a ticket to clean up this setup but again I'm not sure when we'll be doing it: #1498.

This is a very crucial point. If dependabot only bumps `test_requirements.txt` users won't benefit and we'll still have to manually go and change the `extra_requires`.. It would be great if we can find a way that it bumps both.

> Ideally I'd say that here we just limit dependabot to looking at requirements.txt and not test_requirements.txt for now, and deal with the dataset dependencies separately. Problem is that from what @rashidakanchwala said there didn't seem to be a good way to do that... Maybe there's something clever we could do with `ignore` though? https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file I don't even see documented anywhere how dependabot works out which files to try and modify - can you? Does looking at their source code make this clearer?

☝️ This would be a good solution in the meantime. But I definitely feel that we should look at how `extra_requires` could be bumped automatically, because that's what really affects/benefits our users.

SajidAlamQB:
Thanks @AntonyMilneQB that makes a lot of sense.

> Ideally I'd say that here we just limit dependabot to looking at requirements.txt and not test_requirements.txt for now, and deal with the dataset dependencies separately. Problem is that from what @rashidakanchwala said there didn't seem to be a good way to do that... Maybe there's something clever we could do with `ignore` though? https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file I don't even see documented anywhere how dependabot works out which files to try and modify - can you? Does looking at their source code make this clearer?

Looking through this thread https://github.com/dependabot/dependabot-core/issues/4364 there isn't a direct way through the `dependabot.yml` to exclude files/directories. I had a brief look through their [source code](https://github.com/dependabot/dependabot-core/tree/main/python/spec/dependabot/python/file_parser) but hasn't helped me identify how exactly they choose what files to modify. I will continue to look for any workarounds we can do in the meantime.

rashidakanchwala:
> Thanks for the explanation @SajidAlamQB! At a glance it looks like the vast majority of these updates are for dataset requirements - is that right?
> 
> Given that all these requirements would move to `kedro-datasets` I wonder whether we should just make those changes on that repo instead. I'm not sure exactly when we're going to remove the datasets tests (and hence the requirements) from this repo though.
> 
> There's also a problem here that the bumps to dataset requirements are just altering test_requirements.txt and not the ones in setup.py which are the ones that actually matter to users e.g. [SajidAlamQB#2](https://github.com/SajidAlamQB/kedro/pull/2). We have a ticket to clean up this setup but again I'm not sure when we'll be doing it: #1498.
> 
> Ideally I'd say that here we just limit dependabot to looking at requirements.txt and not test_requirements.txt for now, and deal with the dataset dependencies separately. Problem is that from what @rashidakanchwala said there didn't seem to be a good way to do that... Maybe there's something clever we could do with `ignore` though? https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file I don't even see documented anywhere how dependabot works out which files to try and modify - can you? Does looking at their source code make this clearer?
> 
> @MerelTheisenQB thoughts?

you could also ignore test_requirements.txt if you change directory to the specific directory which only has requirements.txt (if that's the case)

SajidAlamQB:
A user on the GitHub Issue I posted earlier recommended using renovate GitHub app: https://github.com/renovatebot/renovate which is an alternative to Dependabot. It does allow for more customizability but wanted to know everyone's thoughts on using this. @AntonyMilneQB @MerelTheisenQB

SajidAlamQB:
Below I have highlighted three options we can take for this issue (Please let me know if you think of any others):

Option 1 (Simple solution): We use dependabot as is and manually go through its PRs selecting the ones we want to be merged.

Option 2 (Breaking change?): The only real way we can tell dependabot which manifest files to use is by specifying the directory. Therefore if we place the requirements.txt or any other files in a specific directory rather than root we can limit which files dependabot uses.

Option 3 (Complex solution): We can alternatively forgo dependabot in favour of [renovate](https://github.com/renovatebot/renovate), which is a GitHub app that is more configurable than dependabot. But this solution may mean we also have to change kedro-viz.

> **Note**
>After discussion we have decided to go with Option 2 for the framework side, this change will only benefit development, we 
>have also decided to add dependabot to kedro-starters repo, so each new starter will have an updated template we can >use the PRs opened their as a guide for the ones to add on framework. Moreover, once work on the datasets is done we will >implement dependabot there as well.

MerelTheisenQB:
I wonder if this needs to be changed as well:
https://github.com/kedro-org/kedro/blob/main/.circleci/continue_config.yml#L149

I'm not actually sure how that checksum stuff works. The file is mentioned in a string so I don't know if it matters that it's now in a different directory.. 🤔

noklam:
Looks good!

The fork branch is not accessible to me, I think it helps if I can see the result of the dependent bot PR.


A couple of questions:
1. Is it possible to do a batch update since you mentioned it will raise 46 PRs, it may take a lot of time to just manually merge it one by one.

2. Does it addresses the problem Antony raised now? I.e. the setup.py - I want to see an example PR.

3.I suggest do a global search of requirements.txt and other dependencies file to make sure we didn't miss any doc.
I think we need to update the **MAKEFILE** as well as this affect the contributor guide.

SajidAlamQB:
> Looks good!
> 
> The fork branch is not accessible to me, I think it helps if I can see the result of the dependent bot PR.
> 
> A couple of questions:
> 
> 1. Is it possible to do a batch update since you mentioned it will raise 46 PRs, it may take a lot of time to just manually merge it one by one.
> 2. Does it addresses the problem Antony raised now? I.e. the setup.py - I want to see an example PR.
> 
> 3.I suggest do a global search of requirements.txt and other dependencies file to make sure we didn't miss any doc. I think we need to update the **MAKEFILE** as well as this affect the contributor guide.

Thanks for the reply @noklam, after some discussion with the team we decided to limit dependabot just to update requirements.txt hence why requirements.txt has been moved to its own directory so dependabot can only access that.

With this change, there are no longer 46 PRs only 10 you can view the updated fork here: https://github.com/SajidAlamQB/kedro-3/pulls

noklam:
Thank you @SajidAlamQB, I have had a quick look at the PRs and I have a follow-up question.

What PR are we going to merge? Do we accept all PRs as long as it passes the tests? @MerelTheisenQB 

> Update fsspec requirement from <=2022.7.1,>=2021.4 to >=2021.4,<2022.8.3 in /dependency
https://github.com/SajidAlamQB/kedro-3/pull/10

For example, I think this is a useful PR - as it updates the upper bound and user may need it. But why is it also changing the lower bound to `2021.4`?


> Update rich requirement from ~=12.0 to ~=12.6 in /dependency
https://github.com/SajidAlamQB/kedro-3/pull/4

For this PR, does it matters since we use `~=`, I don't think it affects the users unless it is changing to `~=13.0`.

MerelTheisenQB:
I think in practice we will merge most PRs, but we should be critical and have a proper look at what dependabot is actually changing. It shouldn't change lower bounds, but if it's a change that doesn't do much but isn't harmful either, we'll just get it in.

SajidAlamQB:
> Thank you @SajidAlamQB, I have had a quick look at the PRs and I have a follow-up question.
> 
> What PR are we going to merge? Do we accept all PRs as long as it passes the tests? @MerelTheisenQB
> 
> > Update fsspec requirement from <=2022.7.1,>=2021.4 to >=2021.4,<2022.8.3 in /dependency
> > [SajidAlamQB#10](https://github.com/SajidAlamQB/kedro-3/pull/10)
> 
> For example, I think this is a useful PR - as it updates the upper bound and user may need it. But why is it also changing the lower bound to `2021.4`?
> 
> > Update rich requirement from ~=12.0 to ~=12.6 in /dependency
> > [SajidAlamQB#4](https://github.com/SajidAlamQB/kedro-3/pull/4)
> 
> For this PR, does it matters since we use `~=`, I don't think it affects the users unless it is changing to `~=13.0`.

We can add versioning strategies to avoid updating lower bounds and constrain it just upper bounds.

MerelTheisenQB:
> We can add versioning strategies to avoid updating lower bounds and constrain it just upper bounds.

I think the only problem with this, is that sometimes you have to bound the lower version, because of some vulnerability that was exposed and you should only use newer versions going forward.

SajidAlamQB:
Dependabot remarks:

Dependabot version updates for Kedro will help us bump dependencies automatically but has some limitations. With the work being done on kedro-datasets and the [improvements to test_requirments.txt setup](https://github.com/kedro-org/kedro/issues/1498), we decided to simplify dependabot to just requirments.txt. Due to the [limitation of dependabot configuration](https://github.com/dependabot/dependabot-core/issues/4364), we can't tell it what files it can access explicitly so we get around this by limiting which directory dependabot can scan. With this in mind, we have moved `requirments.txt` from root into `dependency/requirments.txt` forcing dependabot to only scan in this directory. 

Have a look at a forked Kedro repo with the above-mentioned changes: https://github.com/SajidAlamQB/kedro-3/pulls

Furthermore, after some testing on starters, dependabot was not able to work properly. Dependabot can only run version updates on manifest files if it can access all of the dependencies in the file, even if you add inaccessible dependencies to the ignore option of your configuration file. In our starters `requirments.txt` we have kedro[pandas.CSVDataSet]~={{ cookiecutter.kedro_version }} dependency which dependabot couldn't parse hence it is failing. Following this, we have decided to forgo implementing dependabot on starters for now.

MerelTheisenQB (approved):
Thanks for investigating all options with dependabot @SajidAlamQB ! This is already a step closer in improving our dependency management 👍

noklam (approved):
Thanks for explaining and nice write up! I think it is a good first step even if it's not automating everything.

PR #1865, opened 2022-09-22
Add execution_options to SQLQueryDataSet

### LINKED ISSUE #752: Adding execution_options to kedro.extras.datasets.pandas.SQLQueryDataSet
## Description
Is it possible to add execution_options to kedro.extras.datasets.pandas.SQLQueryDataSet?

For example, I would like to add stream_results=True to the connection string.
```python
  engine = create_engine(
      "postgresql://postgres:pass@localhost/example"
  )
  conn = engine.connect().execution_options(stream_results=True)
```
## Context
This article https://pythonspeed.com/articles/pandas-sql-chunking/ displays that its helpful to read large amounts of data using pandas with stream_results. I would like to use this method with SQLQueryDataSet.

clotildeguinard:
Signed-off-by: Clotilde Guinard <clotilde_guinard@hotmail.fr>

## Description

Resolves #752 

## Development notes

Now a SQLQueryDataSet can be customized with execution options (see https://docs.sqlalchemy.org/en/latest/core/connections.html#sqlalchemy.engine.Connection.execution_options).

Internally, the SQLQueryDataSet class used to "cache" the different engines in order to reuse the same object when 2 datasets where sharing the same connection string. So we had to extend the behaviour of this cache in order to take the execution options into account in the cache key.

## Checklist

Resolves #752 

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1865"></a>


clotildeguinard:
Sorry I realize a bit late that I may simplify my implementation by injecting the execution_options at load time. This would avoid to complexify the caching behaviour... I will submit an additional commit on top of those, and un-draft the PR after that.

clotildeguinard:
Ok, I have pushed the new implementation with updated tests. 

Question for later: should I apply the same approach to the SQLTableDataSet? (which is quite close to the SQLQueryDataSet)

noklam:
@MerelTheisenQB 

I am slightly concerned about this change. `SQLQueryDataSet` is essentially a wrapper on `pandas` API, with this change we are adding new behaviors on top of `pandas`. It's a question why are we doing this and it starts to blur the API, is this still a `pandas/SQLQueryDataSet` then?

Note that at the moment `pandas` doesn't really support this natively. See this recently [closed PR](https://github.com/pandas-dev/pandas/pull/46166/files)

clotildeguinard:
@noklam I will let you decide with @MerelTheisenQB  about the relevance of the issue itself.

Just to answer about the following question that you asked: "Do we have a generic to handle connection instead of injecting parameters into each dataset?". Actually, this is what I did in my first implementation, which injected the execution options at engine level.

Then I added the following commit (the last one of the branch at the moment) https://github.com/kedro-org/kedro/pull/1865/commits/cfa45af5a89b68f19f3cfed74f6eec1121392506 in order to modify this and inject params into each dataset, because I thought it was adding less complexity on the "caching" mechanism (hint: 16 line changes instead of 53). But we can easily go back to the previous implementation, this is just 1 commit to revert.

MerelTheisenQB:
I'm not too concerned about this change adding extra functionality which doesn't exist in the pandas API. The way I see it is that we leverage existing APIs where we can, but also add in features if they benefit users. If a user wants to stick closer to the original pandas API they can always decide not to use this functionality. 

I see @deepyaman was tagged in the original issue, do you have any thoughts on this?

kedro/extras/datasets/pandas/sql_dataset.py
```diff
@@ -346,6 +350,10 @@ def __init__(  # pylint: disable=too-many-arguments
                 https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.spec.AbstractFileSystem.open
                 All defaults are preserved, except `mode`, which is set to `r` when loading.
             filepath: A path to a file with a sql query statement.
+            execution_options: A dictionary with non-SQL options for the connection to
+                be applied to the underlying engine. To find all supported execution
+                options, see here:
+                https://docs.sqlalchemy.org/en/12/core/connections.html#sqlalchemy.engine.Connection.execution_options
```
noklam, line 356: I think this deserves more documentation as this is an advance feature and not documented in the pandas API. 

i.e. What options are common and when do we use it?

The article discusses client-side & server-side computation, are there any tradeoffs here, or server-side is always preferred? I think we should at least mention that if users are running into memory issues, it is worths trying this option.
https://pythonspeed.com/articles/pandas-sql-chunking/

noklam (approved):
Thank you for your contribution, this is something I don't know and happy to learn more about this batching option :) I drop some comments on the documentation, I am happy with the implementation otherwise.

merelcht (commented):
I left some small comments, but I think when those are resolved this is good to be merged 🙂

kedro/extras/datasets/pandas/sql_dataset.py
```diff
@@ -286,7 +286,19 @@ class SQLQueryDataSet(AbstractDataSet[None, pd.DataFrame]):
         >>>   type: pandas.SQLQueryDataSet
         >>>   sql: "select shuttle, shuttle_id from spaceflights.shuttles;"
         >>>   credentials: db_credentials
-        >>>   layer: raw
+
+    Advance example using the `stream_results` and `chunk_size` option to reduce memorgy usage
```
merelcht, line 290:
```suggestion
    Advanced example using the `stream_results` and `chunksize` option to reduce memory usage
```

tests/extras/datasets/pandas/test_sql_dataset.py
```diff
@@ -11,6 +12,7 @@
 TABLE_NAME = "table_a"
 CONNECTION = "sqlite:///kedro.db"
 SQL_QUERY = "SELECT * FROM table_a"
+EXECUTION_OPTIONS = {"streaming": True}
```
merelcht, line 15: Should this be `stream_results` as well?

clotildeguinard:
Hi all, thanks a lot for the reviews and comments! I am really sorry but I am not able to handle comments currently. I will probably be back starting from next week. I will have a look if the PR is not merged by then.

kedro/extras/datasets/pandas/sql_dataset.py
noklam: Thanks a tonsss for picking this up, I am really bad at this😅.

tests/extras/datasets/pandas/test_sql_dataset.py
noklam: Yes, it should be

noklam:
@clotildeguinard Thank you for creating the PR! No worries I will take over the PR for the minor changes, they are mostly just comments and documentation. It's good to go. 👯

Issue #1863, opened 2022-09-21
Make core config accessible in dict get way (e.g. config_loader["catalog"])


MerelTheisenQB:
## Description
Make it easier and clearer to fetch core configuration: catalog, parameters, logging and credentials from the `ConfigLoader`.

## Context
Currently, you can get configuration by calling `.get()` on the `ConfigLoader` and passing a regex: `self.config_loader.get("catalog*", "catalog*/**", "**/catalog*")`. This is hardcoded in the `KedroContext` which is problematic for several reasons:
1. It should be part of `ConfigLoader` and not `KedroContext`
2. It's hard for users to fetch their config, because they need to know the exact regex patterns. 

## Proposed Implementation
Move these patterns out of the `KedroContext` and make it possible to access those core configs like fetching values from a dict: `config_loader["catalog"]`

## Related issues
https://github.com/kedro-org/kedro/issues/1462
https://github.com/kedro-org/kedro/issues/1363

Issue #1864, opened 2022-09-21
Make the config patterns configurable through `settings.py` 


merelcht:
To be done after: 
- [x] https://github.com/kedro-org/kedro/issues/1863

## Description
Users can't currently change what regex the `ConfigLoader` uses to fetch configuration. 

## Context

- As written up in https://github.com/kedro-org/kedro/issues/1363 this feels restrictive 
- On top of that users might want to load other config as well e.g. for Spark or MlFlow

## Proposed Implementation
Make the `ConfigLoader` configurable through `settings.py` for the mapping of catalog -> catalog glob pattern, parameters -> parameters glob pattern, etc (spark -> glob patterns)

Issue #1527, opened 2022-05-12
How to pass extra params to config loader since registration moved to settings.py in 0.18.x


Mercurrent:
## Description

In 0.17.X it was possible to pass params from cli into config loader:
```
class ProjectHooks:
    @hook_impl
    def register_config_loader(
        self,
        conf_paths: Iterable[str],
        env: str,
        extra_params: Dict[str, Any],
    ) -> TemplatedConfigLoader:
        self.extra_params = extra_params
        return TemplatedConfigLoader(conf_paths=conf_paths, globals_pattern="*globals.yml", globals_dict=extra_params)
```
With 0.18.x it doesn't seem to be possible anymore.

## Context
Seems close to:[ issue 1411](https://github.com/kedro-org/kedro/issues/1411)

How has this bug affected you? What were you trying to accomplish?
Needed to be able to pass/override params via CLI

noklam:
@Mercurrent Did you mean `kedro run - - params` is not working for you?

Mercurrent:
Yes, @noklam seems I cannot pass it anymore when I use custom Config Loader. Or I'm missing the way to do this.

Any suggestions for migration of the ProjectHook described?

noklam:
I think ProjectHook is no longer needed. The official way to do it is define TemplateConfigLoader as your loader class in 'settings.py'.

You can find these changes from the RELEASE.md and migration guides.

Mercurrent:
Indeed, it's not needed anymore. How would you migrate the mentioned piece to keep access to params from cli?
We're moving from 0.17.7 onto 0.18.1
TemplateConfigLoader within settings doesn't have access to cli args.

noklam:
@Mercurrent The short answer is you just need to put `TemplatedConfigLoader` in `settings.py` and it should just work.

Before it was moved to `settings.py`, you will need to define it in `ProjectHook`. These parameters are passed from the CLI to the corresponding kedro's class. Under the hood, this is what happened.

https://github.com/kedro-org/kedro/blob/686dba54775ddc7a1ddaef5481e3b82717d33b0f/kedro/framework/session/session.py#L257-L262

Mercurrent:
On 0.17.7 I could use the hook described in the first message above and call a pipeline this way: 
`kedro run --pipeline=data_engineering_pipeline  --params:mode=train` This way mode would override any config params via globals_dict

On 0.18.1
when I use TemplateConfigLoader from settings, there's no way to use CLI params for config params override.

[...]

Issue #1462, opened 2022-04-21
Provide better documentation on how the config glob pattern works


idanov:
Whilst issue #1363 raises more fundamental issues about the glob paths used by the configuration loader classes being tightly coupled with the low level `context`  object, as an interim solution we should better document what actually happens because it's currently not very clear in the docs.

https://github.com/kedro-org/kedro/blob/e78990c6b606a27830f0d502afa0f639c0830950/kedro/framework/context/context.py#L356

Looking for this information in the docs I checked the following places:

1. **configuration.md**
It is sort of present [here](https://kedro.readthedocs.io/en/stable/kedro_project_setup/configuration.html), but not really comprehensive. We could also do a better job describing the thinking and relationship of `AbstractConfigLoader`

2. **`kedro.config.ConfigLoader`**
It's sort of in the docstring here, but not really comprehensive. We could make it very clear for readers.

We should also make sure we use word 'glob' in the header it's easily retrievable / search indexed / linkable

noklam:
The `ConfigLoader` docstring maybe a good place for mentioning the default.

PR #1870, opened 2022-09-23
Make core config accessible in dict get way 

MerelTheisenQB:
## Description
https://github.com/kedro-org/kedro/issues/1863

## Development notes

- Changed `AbstractConfigLoader` to inherit from `UserDict` 
- Moved regex patterns for core config locations from `KedroContext` to `ConfigLoader` and `TemplatedConfigLoader`
- Core config patterns are stored in a dictionary so they can be accessed with dict syntax
- `config_patterns` is added as an argument to the `ConfigLoader` and `TemplatedConfigLoader`

## To discuss
Is allowing users to access config in this way something we want for the future `0.19.0` version as well? How does this allow for using non file based config, e.g stored in a database?

**Discussion summary**

- This change opens up to ability for users to change the config file patterns.
- The pattern paths are added to the `ConfigLoader` and `TemplatedConfigLoader` and not to the `AbstractConfigLoader` to not make add assumptions about file based config in the base abstract class.
- This also means that (in the future) users will be able to point to a database. The implementation of reading from databases would not happen on the Kedro side: it's not possible for us to build functionality for all possible databases that users have. However, the improvements to `ConfigLoader`will enable users to implement their own DB access so they can load their config. 
- In `0.19.0` we will remove the custom `.get()` method and rely on `OmegaConf` for loading the actual config.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1870"></a>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -11,7 +11,8 @@
 # Upcoming Release 0.18.4
 
 ## Major features and improvements
-* Made configuration patterns accessible with dictionary get syntax, by making `AbstractConfigLoader` inherit from `UserDict`. The hardcoded patterns are moved from the `KedroContext` to the config loader classes instead.
+* The config loader objects now implement `UserDict` and the configuration is accessed through `conf_loader['catalog']`
+* You can configure config file patterns through `settings.py` without creating a custom config loader
 
 ## Bug fixes and other changes
 * Fixed `kedro micropkg pull` for packages on PyPI.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -11,6 +11,7 @@
 # Upcoming Release 0.18.4
 
 ## Major features and improvements
+* Made configuration patterns accessible with dictionary get syntax, by making `AbstractConfigLoader` inherit from `UserDict`. The hardcoded patterns are moved from the `KedroContext` to the config loader classes instead.
 
 ## Bug fixes and other changes
 * Fixed `kedro micropkg pull` for packages on PyPI.
--- a/docs/source/kedro_project_setup/configuration.md
+++ b/docs/source/kedro_project_setup/configuration.md
@@ -19,7 +19,7 @@ from kedro.framework.project import settings
 
 conf_path = str(project_path / settings.CONF_SOURCE)
 conf_loader = ConfigLoader(conf_source=conf_path, env="local")
-conf_catalog = conf_loader.get("catalog*", "catalog*/**")
+conf_catalog = conf_loader["catalog"]
 ```
 
 This recursively scans for configuration files firstly in the `conf/base/` (`base` being the default environment) and then in the `conf/local/` (`local` being the designated overriding environment) directory according to the following rules:
@@ -180,7 +180,7 @@ from kedro.framework.project import settings
 
 conf_path = str(project_path / settings.CONF_SOURCE)
 conf_loader = ConfigLoader(conf_source=conf_path, env="local")
-parameters = conf_loader.get("parameters*", "parameters*/**")
+parameters = conf_loader["parameters"]
 ```
 
 This will load configuration files from any subdirectories in `conf` that have a filename starting with `parameters`, or are located inside a folder with name starting with `parameters`.
@@ -189,7 +189,7 @@ This will load configuration files from any subdirectories in `conf` that have a
 Since `local` is set as the environment, the configuration path `conf/local` takes precedence in the example above. Hence any overlapping top-level keys from `conf/base` will be overwritten by the ones from `conf/local`.
 ```
 
-Calling `conf_loader.get()` in the example above will throw a `MissingConfigException` error if no configuration files match the given patterns in any of the specified paths. If this is a valid workflow for your application, you can handle it as follows:
+Calling `conf_loader[]` in the example above will throw a `MissingConfigException` error if no configuration files match the given key. If this is a valid workflow for your application, you can handle it as follows:
 
 ```python
 from kedro.config import ConfigLoader, MissingConfigException
@@ -199,7 +199,7 @@ conf_path = str(project_path / settings.CONF_SOURCE)
 conf_loader = ConfigLoader(conf_source=conf_path, env="local")
 
 try:
-    parameters = conf_loader.get("parameters*", "parameters*/**", "**/parameters*")
+    parameters = conf_loader["parameters"]
 except MissingConfigException:
     parameters = {}
 ```
@@ -315,7 +315,7 @@ from kedro.framework.project import settings
 
 conf_path = str(project_path / settings.CONF_SOURCE)
 conf_loader = ConfigLoader(conf_source=conf_path, env="local")
-credentials = conf_loader.get("credentials*", "credentials*/**")
+credentials = conf_loader["credentials"]
 ```
 
 This will load configuration files from `conf/base` and `conf/local` whose filenames start with `credentials`, or that are located inside a folder with a name that starts with `credentials`.
@@ -324,7 +324,7 @@ This will load configuration files from `conf/base` and `conf/local` whose filen
 Since `local` is set as the environment, the configuration path `conf/local` takes precedence in the example above. Hence, any overlapping top-level keys from `conf/base` will be overwritten by the ones from `conf/local`.
 ```
 
-Calling `conf_loader.get()` in the example above throws a `MissingConfigException` error if no configuration files match the given patterns in any of the specified paths. If this is a valid workflow for your application, you can handle it as follows:
+Calling `conf_loader[]` in the example above throws a `MissingConfigException` error if no configuration files match the given key. If this is a valid workflow for your application, you can handle it as follows:
 
 ```python
 from kedro.config import ConfigLoader, MissingConfigException
@@ -334,7 +334,7 @@ conf_path = str(project_path / settings.CONF_SOURCE)
 conf_loader = ConfigLoader(conf_source=conf_path, env="local")
 
 try:
-    credentials = conf_loader.get("credentials*", "credentials*/**")
+    credentials = conf_loader["credentials*"]
 except MissingConfigException:
     credentials = {}
 ```
--- a/docs/source/kedro_project_setup/configuration.md
+++ b/docs/source/kedro_project_setup/configuration.md
@@ -189,7 +189,7 @@ This will load configuration files from any subdirectories in `conf` that have a
 Since `local` is set as the environment, the configuration path `conf/local` takes precedence in the example above. Hence any overlapping top-level keys from `conf/base` will be overwritten by the ones from `conf/local`.
 ```
 
-Calling `conf_loader[]` in the example above will throw a `MissingConfigException` error if no configuration files match the given key. If this is a valid workflow for your application, you can handle it as follows:
+Calling `conf_loader[key]` in the example above will throw a `MissingConfigException` error if no configuration files match the given key. If this is a valid workflow for your application, you can handle it as follows:
 
 ```python
 from kedro.config import ConfigLoader, MissingConfigException
@@ -324,7 +324,7 @@ This will load configuration files from `conf/base` and `conf/local` whose filen
 Since `local` is set as the environment, the configuration path `conf/local` takes precedence in the example above. Hence, any overlapping top-level keys from `conf/base` will be overwritten by the ones from `conf/local`.
 ```
 
-Calling `conf_loader[]` in the example above throws a `MissingConfigException` error if no configuration files match the given key. If this is a valid workflow for your application, you can handle it as follows:
+Calling `conf_loader[key]` in the example above throws a `MissingConfigException` error if no configuration files match the given key. If this is a valid workflow for your application, you can handle it as follows:
 
 ```python
 from kedro.config import ConfigLoader, MissingConfigException
--- a/docs/source/kedro_project_setup/configuration.md
+++ b/docs/source/kedro_project_setup/configuration.md
@@ -334,7 +334,7 @@ conf_path = str(project_path / settings.CONF_SOURCE)
 conf_loader = ConfigLoader(conf_source=conf_path, env="local")
 
 try:
-    credentials = conf_loader["credentials*"]
+    credentials = conf_loader["credentials"]
 except MissingConfigException:
     credentials = {}
 ```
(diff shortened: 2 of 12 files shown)


kedro/config/abstract_config.py
```diff
@@ -1,11 +1,11 @@
 """This module provides ``kedro.abstract_config`` with the baseline
 class model for a `ConfigLoader` implementation.
 """
-from abc import ABC, abstractmethod
+from abc import abstractmethod
 from typing import Any, Dict
 
 
-class AbstractConfigLoader(ABC):
+class AbstractConfigLoader(dict):
```
noklam, line 8: I think it would be better to inherit from `UserDict` instead of the `dict`, there are some traps with inheriting from the Python standard `dict` class.

It would also be more consistent, see `_ProjectLogging` that we are already using `UserDict`.
https://github.com/kedro-org/kedro/blob/4db0ae3be611f60f3c98da2501364d5549c784f2/kedro/framework/project/__init__.py#L199-L208

Fluent Python has a detail explanation for it.

See these two relevant chapter
* https://learning.oreilly.com/library/view/fluent-python-2nd/9781492056348/ch14.html#subclass_builtin_woes
* https://learning.oreilly.com/library/view/fluent-python-2nd/9781492056348/ch03.html#missing_method

kedro/config/config.py
```diff
@@ -86,18 +86,32 @@ def __init__(
...
         self.base_env = base_env
         self.default_run_env = default_run_env
 
+        core_config_patterns = {
+            "catalog": ["catalog*", "catalog*/**", "**/catalog*"],
+            "parameters": ["parameters*", "parameters*/**", "**/parameters*"],
+            "credentials": ["credentials*", "credentials*/**", "**/credentials*"],
+            "logging": ["logging*", "logging*/**", "**/logging*"],
+        }
+        self.config_patterns = core_config_patterns
```
noklam, line 98: ⭐️  I like this

noklam (commented):
I like the direction of this change and it should make it easier towards customisable path for all configurations ⭐️. 

This is just something always in my mind since this question almost come up every week and I always has to point people to make a copy & paste `CustomTemplateConfigLoader`. Would be great if we can do something about it while we are making changes to these classes. This is definitely out of scope of this PR, but I just want bring it out so we can see if there are any opportunity to improve it.
https://github.com/kedro-org/kedro/issues/1527

AntonyMilneQB (commented):
tl;dr: I think this is a good start, definitely an improvement or where things are now and probably fine to merge after thinking about the inheritance a bit more, but I'm not sure exactly how the next step will work. I'd feel more confident about it if @idanov could weight in with his thoughts!

I think there were a few major requirements here:
* enable access through getitem syntax:`config_loader["params"]` etc. ✅ 
* change all our `.get` calls to `__getitem__` ✅ 
* maintain backwards compatibility so that 3rd party `.get` calls don't need to change ✅ 

So I think that in basic terms we've achieved what we wanted to here.

But I'm left with several questions about exactly how it should be done:
1. what should `AbstractConfigLoader` inherit from? IMO at the moment it should just be `ABC` still. I would personally leave the dictionary inheritance for the future if and when it becomes a purely dict-like object
2. where should the `__getitem__` definition go - abstract config loader or the implementations? At the moment it's just copied and pasted between the implementations which seems a bit unDRY.
3. likewise do `core_config_patterns` belong in the abstract class or the implementations?
4. is the intention to remove `.get` entirely in the future?

tbh these points are all kind of moot though given that it's not at all clear what the abstract config loader is actually for currently...

More practically, what is actually the expected user journey following this change if I want to (a) add spark config and (b) add `*parameters` as a glob pattern? This is  https://github.com/kedro-org/kedro/issues/1864 but it feels a bit hard to judge the proposed change here without understanding how that would work. Currently I think it would be something like this:
```
class MyConfigLoader(ConfigLoader):
    def __init__(...):
         super().__init__(...)
         self.core_config_patterns.append("*parameters")
         self.core_config_patterns["spark"] = ["spark*", ...]
         # really we should have another object that's not called "core" to add spark to
```

I believe @idanov's plan was instead of doing class inheritance to make this patterns dictionary directly accessible in settings.py. But I don't quite see how that works. Would we then want to pass the patterns into the constructor of `ConfigLoader`? And how would that eventually enable different sort of config, e.g. database vs. file? We don't need to have it all worked out now, but I'd personally feel more comfortable with this change if I could see better what the ultimate consequence will look like.

kedro/config/abstract_config.py
MerelTheisenQB: I don't remember this 100% anymore, but I think initially the plan was to have [`AbstractConfigLoader` inherit from `MutableMapping`](https://treyhunner.com/2019/04/why-you-shouldnt-inherit-from-list-and-dict-in-python/#Abstract_base_classes:_they%E2%80%99ll_help_you_quack_like_a_duck) but then it ended up being a [lot of overwriting of stuff we don't need.](https://github.com/kedro-org/kedro/commit/6dc6eb151bc9112752d6fe669912c689a1921732) And then the decision was made to inherit from `dict` directly. This was all over Slack and not very coherent. 

Thinking about this again now, I feel like you might be right and we don't actually need to inherit from `dict` at all since we're overwriting `get` and `__getitem__` anyway.

kedro/framework/context/context.py
```diff
@@ -240,10 +240,7 @@ def params(self) -> Dict[str, Any]:
                 extra parameters passed at initialization.
         """
         try:
-            # '**/parameters*' reads modular pipeline configs
-            params = self.config_loader.get(
-                "parameters*", "parameters*/**", "**/parameters*"
-            )
+            params = self.config_loader["parameters"]
```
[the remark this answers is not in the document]
MerelTheisenQB: Yes exactly! This is introduces as a non-breaking change and so we can't completely change/remove `get` yet.

MerelTheisenQB:
As always, thank you for your thorough comments and feedback @AntonyMilneQB ⭐ 

> But I'm left with several questions about exactly how it should be done:
> 
> 1. what should `AbstractConfigLoader` inherit from? IMO at the moment it should just be `ABC` still. I would personally leave the dictionary inheritance for the future if and when it becomes a purely dict-like object

Yes I think that would make sense, see my earlier comment: https://github.com/kedro-org/kedro/pull/1870#discussion_r982258144 


> 2. where should the `__getitem__` definition go - abstract config loader or the implementations? At the moment it's just copied and pasted between the implementations which seems a bit unDRY.
> 3. likewise do `core_config_patterns` belong in the abstract class or the implementations?

These are good questions, my feeling is it should be in the loader implementations because it's too specific to be in the abstract class. As in the mapping from "params" to the parameter regexes isn't something the abstract implementation should care about or anyone inheriting from it. But I also agree that it's not clear what the role of the `AbstractConfigLoader` is anymore.. 

> 4. is the intention to remove `.get` entirely in the future?

Yes as far as I understand it, and instead just leverage the `dict` get methods. 

> More practically, what is actually the expected user journey following this change if I want to (a) add spark config and (b) add `*parameters` as a glob pattern? This is #1864 but it feels a bit hard to judge the proposed change here without understanding how that would work. Currently I think it would be something like this:
> 
> ```
> class MyConfigLoader(ConfigLoader):
>     def __init__(...):
>          super().__init__(...)
>          self.core_config_patterns.append("*parameters")
>          self.core_config_patterns["spark"] = ["spark*", ...]
>          # really we should have another object that's not called "core" to add spark to
> ```
> 
> I believe @idanov's plan was instead of doing class inheritance to make this patterns dictionary directly accessible in settings.py. But I don't quite see how that works. Would we then want to pass the patterns into the constructor of `ConfigLoader`? And how would that eventually enable different sort of config, e.g. database vs. file? We don't need to have it all worked out now, but I'd personally feel more comfortable with this change if I could see better what the ultimate consequence will look like.

I coded this up in an older (less clean) branch: https://github.com/kedro-org/kedro/compare/abstract-config-loader, but essentially you would have something like this in `settings.py`:

```
CONFIG_LOADER_CLASS = ConfigLoader
CONFIG_LOADER_ARGS = {
   "custom_patterns":{"spark" : ["spark*/"]}
}
```

and inside the `ConfigLoader`
```
 if custom_patterns:
    core_patterns.update(custom_patterns)
 self.config_patterns = core_patterns
```
So you can add whatever custom config you have as well as overwrite any of the "core" config locations.

noklam:
> 1. what should AbstractConfigLoader inherit from? IMO at the moment it should just be ABC still. I would personally leave the dictionary inheritance for the future if and when it becomes a purely dict-like object

I just want to point out that with this PR, we already assume `config_loader` is a dict-like object. If we are going with `ABC`, then we should also add `__getitem__` as the class method.

@MerelTheisenQB @AntonyMilneQB 
Example:
```
# context.py
 params = self.config_loader["parameters"]
```

docs/source/kedro_project_setup/configuration.md
```diff
@@ -334,7 +334,7 @@ conf_path = str(project_path / settings.CONF_SOURCE)
 conf_loader = ConfigLoader(conf_source=conf_path, env="local")
 
 try:
-    credentials = conf_loader.get("credentials*", "credentials*/**")
+    credentials = conf_loader["credentials*"]
```
idanov, line 337:
```suggestion
    credentials = conf_loader["credentials"]
```

kedro/config/abstract_config.py
```diff
@@ -19,12 +19,13 @@ def __init__(
...
         **kwargs  # pylint: disable=unused-argument
     ):
+        super().__init__()
         self.conf_source = conf_source
         self.env = env
         self.runtime_params = runtime_params
 
     @abstractmethod  # pragma: no cover
-    def get(self) -> Dict[str, Any]:
+    def get(self) -> Dict[str, Any]:  # type: ignore
```
idanov, line 28: Is this still needed if we are inheriting from `dict` or `UserDict`? I suppose we can completely remove it from here, right?

kedro/framework/context/context.py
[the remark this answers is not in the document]
idanov: I wonder whether it's worth opening a follow up PR straight to develop with the breaking change so we don't forget to do this when we release 0.19?

kedro/config/config.py
idanov, line 98: Can we include a `config_patterns` argument in the `__init__`, so those patterns are configurable through `settings.py`? Then the code here could look like:

```
self.config_patterns = {
    "catalog": ["catalog*", "catalog*/**", "**/catalog*"],
    "logging": ["logging*", "logging*/**", "**/logging*"],
}

self.config_patterns.update(config_patterns or {})
```

kedro/config/templated_config.py
```diff
@@ -114,6 +114,14 @@ def __init__(
                 obtained from the globals_pattern. In case of duplicate keys, the
                 ``globals_dict`` keys take precedence.
         """
+        core_config_patterns = {
+            "catalog": ["catalog*", "catalog*/**", "**/catalog*"],
+            "parameters": ["parameters*", "parameters*/**", "**/parameters*"],
+            "credentials": ["credentials*", "credentials*/**", "**/credentials*"],
+            "logging": ["logging*", "logging*/**", "**/logging*"],
+        }
+        self.config_patterns = core_config_patterns
```
idanov, line 123: Same here, see the comment above for `ConfigLoader`.

idanov (commented):
Awesome, such a small code change enabling so much extra functionality!

Could you add a very short description of this to the RELEASE.md?

kedro/framework/context/context.py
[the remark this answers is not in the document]
MerelTheisenQB: We decided to wait with changes for `0.19.0` for other issues to make sure the `main` and `develop` branch don't diverge so much as with `0.18.0`. I'll make sure to add GH issues for any follow up actions that come out of this work.

kedro/config/abstract_config.py
MerelTheisenQB: I think the bigger question raised by @noklam and @AntonyMilneQB was if we should actually be inheriting from `dict` or `UserDict` or just keep it to `ABC`: https://github.com/kedro-org/kedro/pull/1870#discussion_r981264452 
https://github.com/kedro-org/kedro/pull/1870#issuecomment-1261131796

I don't have a strong opinion on this, but feel that the intention to make the `ConfigLoader` classes more dict like is clearer if we inherit from `dict`/`UserDict`.

What do you think @idanov ?

kedro/config/config.py
MerelTheisenQB: Yes can do. Originally I intended to have a separation between `core` and `custom` config patterns so it's clear which patterns have been changed by the user:
```
core_patterns = {
            "catalog": ["catalog*", "catalog*/**", "**/catalog*"],
            "logging": ["logging*", "logging*/**", "**/logging*"],
        }

if custom_patterns:
    core_patterns.update(custom_patterns)
self.config_patterns = core_patterns
```

But perhaps this way it's also clear that you can actually update the core patterns as well as add your own, e.g. spark config.

kedro/config/abstract_config.py
idanov: Agree, let's do it with `UserDict` then.

MerelTheisenQB: Done!

noklam (approved):
Nice PR! I think the development notes needed to be update a bit. I would add that the `config_pattern` is now an argument, although it's not configurable in `settings.py` yet. Maybe we should update the `RELEASE.md` too?

RELEASE.md
```diff
@@ -11,6 +11,7 @@
 # Upcoming Release 0.18.4
 
 ## Major features and improvements
+* Made configuration patterns accessible with dictionary get syntax, by making `AbstractConfigLoader` inherit from `UserDict`. The hardcoded patterns are moved from the `KedroContext` to the config loader classes instead.
```
idanov, line 14: I would rephrase it to something more user-friendly, e.g. along the lines of:
```suggestion
* The config loader objects now implement `UserDict` and the configuration is accessed through `conf_loader['catalog']`
* You can configure config file patterns through `settings.py` without creating a custom config loader
```

idanov (approved):
Awesome! Should we add a small section in the docs to show how to change file patterns through `settings.py`?

MerelTheisenQB:
> Awesome! Should we add a small section in the docs to show how to change file patterns through `settings.py`?

I was planning on doing this in the follow up tasks:
https://github.com/kedro-org/kedro/issues/1864
https://github.com/kedro-org/kedro/issues/1462

Issue #1849, opened 2022-09-14
Should testing and linting dependencies and configuration be dropped from starters and / or templates?


jmholzer:
## Introduction

In #1766, it was decided to deprecate the commands `kedro test` and `kedro lint`.

It should now be decided whether to keep the folder structure, dependencies and configuration files (referred to from here on as the 'items') that these commands rely upon in Kedro's project starters and project template.

## Background

As a starting point for this exploration, the key arguments for / against removing the items are summarised.

The key advantages to removing them are:
1. A simpler folder structure, which @yetudada identified as being a key issue for new users.
2. Fewer dependencies to manage.

The key disadvantages to removing them are:
1. Testing and linting the template and starters currently uses the items, removing them will mean finding an alternative.
2. These items (e.g. the src/tests folder and linting dependencies) are viewed as a nudge to the user, reminding them of the importance of testing and linting.

## Outcome

The outcome of this ticket should be an actionable plan describing which of the items to keep and which to remove in:
1. The project template
2. The starter projects

An alternative means of testing / linting the starters and template is a key outcome of this issue in the case that the items should be removed.

MerelTheisenQB:
`test`
Testing requires a lot less setup, so in my opinion it's fine to remove the setup and dependencies for it from both the template and starters. 

`lint`
I've checked how we ensure the template and starters are linted currently and found the following:
- The template is not linted at the moment. `pre-commit` skips the template because of the cookiecutter `{{ ... }}` format in several files. In the e2e tests we run `kedro lint` on a test starter, which is not the actual template. 
- When you run `kedro lint` on a new project it will actually reformat some things.
- Starters are linted in the e2e tests, and so we can just replace the `kedro lint` call with the underlying calls to black, flake8 and isort. 
   - To make these checks all pass we would need to add test setup for flake8, because the default max line length is 79 and we've always worked with a max lint length of 88.

Considering the template isn't actually linted currently, I'd suggest to remove all lint setup from that.
I'd still be in favour of having one starter (probably spaceflights) with full linting config in it so we can point to it as an example for users how to configure their linting.

MerelTheisenQB:
Closing in favour of the follow up tickets:

- [ ] https://github.com/kedro-org/kedro/issues/1947
- [ ] https://github.com/kedro-org/kedro-starters/issues/106
- [ ] https://github.com/kedro-org/kedro/issues/1948

MerelTheisenQB:
Closing in favour of the follow up tickets:

[...]

Issue #1620, opened 2022-06-14
Document how users can use linting tools instead of `kedro lint`


MerelTheisenQB:
In https://github.com/kedro-org/kedro/issues/1293 we decided to remove the `kedro lint` command. 

We need to update our documentation to remove any suggestions of `kedro lint` and instead write instructions for our users on how to use the linting tools that `kedro lint` combined: `black`, `flake8` and `isort`. We can also recommend them to setup pre-commit hooks to run linting automatically, as some users reported using already. 

Also , explain how users should setup configuration for the linting tools.

PR #1873, opened 2022-09-26
Deprecate `kedro test` and `kedro lint`

### LINKED ISSUE #1872: Deprecate `kedro lint` and `kedro test`
## Description
https://github.com/kedro-org/kedro/issues/1766

## Implementation
Deprecate `kedro lint` and `kedro test` in the same way the other development related CLI commands were deprecated in https://github.com/kedro-org/kedro/issues/1616

noklam:

## Description

Close #1872 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1873"></a>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -14,7 +14,9 @@
 
 ## Bug fixes and other changes
 
-## Breaking changes to the API
+## Upcoming deprecations for Kedro 0.19.0
+
+* `kedro test`, `kedro lint` will be deprecated.
 
 # Release 0.18.3
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -16,7 +16,7 @@
 
 ## Upcoming deprecations for Kedro 0.19.0
 
-* `kedro test`, `kedro lint` will be deprecated.
+* `kedro test` and `kedro lint` will be deprecated.
 
 # Release 0.18.3
 
--- a/kedro/framework/cli/project.py
+++ b/kedro/framework/cli/project.py
@@ -76,7 +76,6 @@ def test(metadata: ProjectMetadata, args, **kwargs):  # pylint: disable=unused-a
         "DeprecationWarning: Command 'kedro test' is deprecated and "
         "will not be available from Kedro 0.19.0. "
         "Use the command 'pytest' instead. "
-
     )
     click.secho(deprecation_message, fg="red")
 
--- a/kedro/framework/cli/project.py
+++ b/kedro/framework/cli/project.py
@@ -71,7 +71,15 @@ def project_group():  # pragma: no cover
 @forward_command(project_group, forward_help=True)
 @click.pass_obj  # this will pass the metadata as first argument
 def test(metadata: ProjectMetadata, args, **kwargs):  # pylint: disable=unused-argument
-    """Run the test suite."""
+    """Run the test suite. (DEPRECATED)"""
+
+    deprecation_message = (
+        "DeprecationWarning: Command 'kedro test' is deprecated and "
+        "will not be available from Kedro 0.19.0. "
+        "Use the command 'pytest' instead. "
+
+    )
+    click.secho(deprecation_message, fg="red")
     try:
         _check_module_importable("pytest")
     except KedroCliError as exc:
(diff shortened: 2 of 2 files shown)


kedro/framework/cli/project.py
```diff
@@ -90,7 +98,13 @@ def test(metadata: ProjectMetadata, args, **kwargs):  # pylint: disable=unused-a
 def lint(
     metadata: ProjectMetadata, files, check_only, **kwargs
 ):  # pylint: disable=unused-argument
-    """Run flake8, isort and black."""
+    """Run flake8, isort and black. (DEPRECATED)"""
+    deprecation_message = (
+        "DeprecationWarning: Command 'kedro lint' is deprecated and "
+        "will not be available from Kedro 0.19.0."
```
deepyaman, line 104: Do we need to point out that users should call linters directly? Do we need to specify which ones?

kedro/framework/cli/project.py
```diff
@@ -71,7 +71,15 @@ def project_group():  # pragma: no cover
 @forward_command(project_group, forward_help=True)
 @click.pass_obj  # this will pass the metadata as first argument
 def test(metadata: ProjectMetadata, args, **kwargs):  # pylint: disable=unused-argument
-    """Run the test suite."""
+    """Run the test suite. (DEPRECATED)"""
+
+    deprecation_message = (
+        "DeprecationWarning: Command 'kedro test' is deprecated and "
+        "will not be available from Kedro 0.19.0. "
+        "Use the command 'pytest' instead. "
```
deepyaman, line 79: Just to be clear, this means we are not wrapping `pytest` in a Kedro command, but we're still supporting `pytest`, right? Since, if we were just going to say "testing is up to you now," we shouldn't call out `pytest` in the message.

I think it's fine as-is, if I've understood correctly.

kedro/framework/cli/project.py
```diff
@@ -71,7 +71,15 @@ def project_group():  # pragma: no cover
 @forward_command(project_group, forward_help=True)
 @click.pass_obj  # this will pass the metadata as first argument
 def test(metadata: ProjectMetadata, args, **kwargs):  # pylint: disable=unused-argument
-    """Run the test suite."""
+    """Run the test suite. (DEPRECATED)"""
+
+    deprecation_message = (
```
deepyaman, line 76: Nit: I'm overly picky, but it would be nice to be consistent on space between docstring and `deprecation_message = ...` and between `click.secho` and the remaining logic below, IMO. :) I checked the previously-deprecated stuff, and it's not consistent, so I don't have a recommendation one way or another--just generally prefer if it's all the same.

RELEASE.md
```diff
@@ -14,7 +14,9 @@
 
 ## Bug fixes and other changes
 
-## Breaking changes to the API
+## Upcoming deprecations for Kedro 0.19.0
+
+* `kedro test`, `kedro lint` will be deprecated.
```
deepyaman, line 19:
```suggestion
* `kedro test` and `kedro lint` will be deprecated.
```

kedro/framework/cli/project.py
noklam: I think we will still have some tests in starters? This isn't too important though, as this deprecating message is for 0.18.x, we will still support `pytest` until 0.19.0, so this message is just asking people to use `pytest` instead of `kedro test`.
#1879 is a more detailed docs for how to use `pytest`.

kedro/framework/cli/project.py
noklam: I am less sure about this one, maybe it will become more clear when #1620 is done. I am fine we either wait until #1620 is done or just merge this one and add some links when the docs PRs are done.

kedro/framework/cli/project.py
noklam: I wish the linter would do this for me, but it didn't. I will make a change to the styling.

noklam: I move all the deprecated message at the top of the function, and a newline between the `click.secho` and the actual logic.

kedro/framework/cli/project.py
jmholzer: Determining whether or not tests will still be in starters is part of an open issue (#1849).

jmholzer: I personally wouldn't mention `pytest` as an alternative in this message, I think this also would keep things consistent with the linting message.

jmholzer (approved):
Happy to approve this PR as it is, though I left a comment on the deprecation of `kedro test`.

kedro/framework/cli/project.py
noklam: I am gonna leave it as is. I don't think it matters too much and I am fine with either option.

I checked in 0.17.x we did similar thing for `kedro install`
https://github.com/kedro-org/kedro/blob/80c0d3a47166c54f2bff4bed42cf63e06e65273c/kedro/framework/cli/project.py#L175-L180

PR #1877, opened 2022-09-26
[TEST] - remove cloud protocol

noklam:

## Checklist

### DIFF
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -681,10 +681,9 @@ def _parse_filepath(filepath: str) -> Dict[str, str]:
     options = {"protocol": protocol, "path": path}
 
     if parsed_path.netloc:
-        if protocol in CLOUD_PROTOCOLS:
-            host_with_port = parsed_path.netloc.rsplit("@", 1)[-1]
-            host = host_with_port.rsplit(":", 1)[0]
-            options["path"] = host + options["path"]
+        host_with_port = parsed_path.netloc.rsplit("@", 1)[-1]
+        host = host_with_port.rsplit(":", 1)[0]
+        options["path"] = host + options["path"]
 
     return options
 


noklam:
@AntonyMilneQB

noklam:
@datajoely may have some comments on this? I don't have the full context of this issue but basically the description of this PR is what I found after looking into JIRA

noklam:
Please ignore the `print`, I obviously have to do some clean up for `CLOUD_PROTOCOLS` too.

AntonyMilneQB:
I'm very unsure about this, but that's just because I don't understand why (or even if) it currently works. Maybe our assumption that it currently works actually doesn't hold? The tests passing is definitely a good sign, but really some manual testing is probably needed here. If we can resolve this once and for all then I would LOVE it 😅 ❤️  So thank you for trying! 

Ideally we should remove our `infer_storage_options` entirely. If that's not possible, then removing our custom `CLOUD_PROTOCOLS` like you do here _in theory_ sounds like a good step towards that, but it's now more divergent from what fsspec have and I don't know which is correct.

For reference:
```
CLOUD_PROTOCOLS = ["s3", "s3a", "gcs", "gs"]  # fsspec 
CLOUD_PROTOCOLS = ["s3", "gcs", "gs", "adl", "abfs"]  # kedro now
CLOUD_PROTOCOLS = ["hdfs"]  # your proposal
```
(fsspec `CLOUD_PROTOCOLS` comes from https://github.com/fsspec/filesystem_spec/blob/master/fsspec/utils.py#L75. Apart from that I think our `infer_storage_options` is functionally identical to theirs so comparing `CLOUD_PROTOCOLS` is the only thing we need to do?) 

So my questions would be: 
1. If our tests pass _without_ s3, s3a, gcs, gs in `CLOUD_PROTOCOLS`, as they have already in fsspec, then which is correct (i.e. actually works in practice)? Our tests or fsspec?
2. Related: why are things (apparently) working for us with `adl` and `abfs` while when we tried adding it to fsspec it didn't work? Apparently it needed updates to `ablfs` to work properly... But then why does it work for us? Does it actually work for us in practice or is it just that the tests pass?
3. If `hdfs` should go in `CLOUD_PROTOCOLS`, should that in theory be done on the fsspec side too (if they can coordinate it, which I know probably they won't easily)? 
4. Why do things currently work for us without `hdfs`? Why do things work for fsspec without `hdfs`?

Basically I'm very keen for us to get to the bottom of this, but I have limited faith in our tests here unless we can really understand the above.

PR #1881, opened 2022-09-27
Update Error message for `VersionNotFoundError` to handle Permission related issues better

### LINKED ISSUE #1768: Catch error for issues with reading/writing data to object stores (e.g. S3)

ankatiyar:
Signed-off-by: Ankita Katiyar [110245118+ankatiyar@users.noreply.github.com](mailto:110245118+ankatiyar@users.noreply.github.com)

## Description
Resolves #1768

## Development notes
Fresh PR to resolve the issue.
Closing https://github.com/kedro-org/kedro/pull/1809 in favour of this solution - `PermissionError` was being handled outside of Kedro in some cases. This PR is a workaround as the error message is insufficient and it's hard to pinpoint the source of error precisely in case of cloud storage.

### Proposed Solution
As a result, this PR modifies the error message for `VersionNotFoundError` to indicate that this error could be happening due to insufficient permission in case of cloud storage i.e. S3, GCS etc.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1881"></a>

### DIFF
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -541,7 +541,8 @@ class AbstractVersionedDataSet(AbstractDataSet[_DI, _DO], abc.ABC):
         protocol = getattr(self, "_protocol", None)
         if not most_recent:
             if protocol in CLOUD_PROTOCOLS:
-                message = f"Did not find any versions for {self}. This could be due to insufficient permission."
+                message = f"Did not find any versions for {self} This could be " \
+                          f"due to insufficient permission."
             else:
                 message = f"Did not find any versions for {self}"
             raise VersionNotFoundError(message)
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -542,7 +542,7 @@ class AbstractVersionedDataSet(AbstractDataSet[_DI, _DO], abc.ABC):
         if not most_recent:
             if protocol in CLOUD_PROTOCOLS:
                 message = (
-                    f"Did not find any versions for {self} This could be "
+                    f"Did not find any versions for {self}. This could be "
                     f"due to insufficient permission."
                 )
             else:
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -541,8 +541,10 @@ class AbstractVersionedDataSet(AbstractDataSet[_DI, _DO], abc.ABC):
         protocol = getattr(self, "_protocol", None)
         if not most_recent:
             if protocol in CLOUD_PROTOCOLS:
-                message = f"Did not find any versions for {self} This could be " \
-                          f"due to insufficient permission."
+                message = (
+                    f"Did not find any versions for {self} This could be "
+                    f"due to insufficient permission."
+                )
             else:
                 message = f"Did not find any versions for {self}"
             raise VersionNotFoundError(message)
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -538,9 +538,13 @@ class AbstractVersionedDataSet(AbstractDataSet[_DI, _DO], abc.ABC):
         most_recent = next(
             (path for path in version_paths if self._exists_function(path)), None
         )
-
+        protocol = getattr(self, "_protocol", None)
         if not most_recent:
-            raise VersionNotFoundError(f"Did not find any versions for {self}")
+            if protocol in CLOUD_PROTOCOLS:
+                message = f"Did not find any versions for {self}. This could be due to insufficient permission."
+            else:
+                message = f"Did not find any versions for {self}"
+            raise VersionNotFoundError(message)
 
         return PurePath(most_recent).parent.name
 
--- a/tests/io/test_data_catalog.py
+++ b/tests/io/test_data_catalog.py
@@ -653,13 +653,13 @@ class TestDataCatalogVersioned:
         assert "ds3__csv" in catalog.datasets.__dict__
         assert "jalapeño" in catalog.datasets.__dict__
 
-    def test_no_versions_cloud_protocol(self):
+    def test_no_version_cloud(self):
         """Check the error if no versions are available for load from cloud storage"""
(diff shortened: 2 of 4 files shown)


noklam:
@ankatiyar Don't forget to document how this decision was made!

tests/io/test_data_catalog.py
```diff
@@ -652,3 +652,14 @@ def test_replacing_nonword_characters(self):
         assert "ds2_spark" in catalog.datasets.__dict__
         assert "ds3__csv" in catalog.datasets.__dict__
         assert "jalapeño" in catalog.datasets.__dict__
+
+    def test_no_version_cloud(self):
```
noklam, line 656:
```suggestion
    def test_no_versions_with_cloud_protocol(self):
```

tests/io/test_data_catalog.py
```diff
@@ -652,3 +652,14 @@ def test_replacing_nonword_characters(self):
...
+        """Check the error if no versions are available for load from cloud storage"""
+        version = Version(load=None, save=None)
+        ds = CSVDataSet("s3://bucket/file.csv", version=version)
```
noklam, line 659:
```suggestion
        versioned_dataset = CSVDataSet("s3://bucket/file.csv", version=version)
```

tests/io/test_data_catalog.py
```diff
@@ -652,3 +652,14 @@ def test_replacing_nonword_characters(self):
...
+        pattern = re.escape(
+            f"Did not find any versions for {ds} "
+            f"This could be due to insufficient permission."
+        )
+        with pytest.raises(DataSetError, match=pattern):
+            ds.load()
```
noklam, line 665:
```python
    def test_no_versions(self, versioned_plot_writer):
        """Check the error if no versions are available for load."""
        pattern = r"Did not find any versions for MatplotlibWriter\(.+\)"
            versioned_plot_writer.load()
```

I found this test in one of our datasets test, can we follow the same pattern here? I think "rf" should be the same as escape.

noklam (commented):
Thank you for sticking with it, I think this is a tough one as the first few issues. It's in a good shape already, most of my comments are around the styling.

A good tip is always check the existing codebase and see if there are something similar exist. I

MerelTheisenQB (commented):
Just to double check that I understand this correctly: the error returned by `fsspec` isn't a `PermissionError` and that's why we cannot know for sure that there is something wrong with permission?

kedro/io/core.py
```diff
@@ -538,9 +538,16 @@ def _fetch_latest_load_version(self) -> str:
         most_recent = next(
             (path for path in version_paths if self._exists_function(path)), None
         )
-
+        protocol = getattr(self, "_protocol", None)
         if not most_recent:
-            raise VersionNotFoundError(f"Did not find any versions for {self}")
+            if protocol in CLOUD_PROTOCOLS:
+                message = (
+                    f"Did not find any versions for {self} This could be "
```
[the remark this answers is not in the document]
noklam: @MerelTheisenQB Correct, it's possible to get an error with some other functions , but all datasets rely on `self._fs.glob` and it won't give you any error with permission issues.

It seems that this is the expected behavior of `glob`. For example, if you actually `ls` the system, you will get a permission error. However, this is a very expensive operation especially with object stores, thus I didn't like it.

tests/io/test_data_catalog.py
ankatiyar: Made the change in 211c449b68a52d3204dcb884b9af671d24753b1f

tests/io/test_data_catalog.py
ankatiyar: Made the change in 211c449b68a52d3204dcb884b9af671d24753b1f

tests/io/test_data_catalog.py
ankatiyar: I tried using "rf" but it doesn't seem to work.

kedro/io/core.py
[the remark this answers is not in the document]
ankatiyar: Made the change in 211c449b68a52d3204dcb884b9af671d24753b1f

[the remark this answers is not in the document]
MerelTheisenQB: Thanks for the explanation @noklam 🙂

MerelTheisenQB (approved):
This was definitely a lot trickier than we thought, but this solution makes sense to me 👍 

Don't forget to add the change in the release notes.

AhdraMeraliQB (approved):
Nice work on this!! Thanks for sticking with it 😄

Issue #1691, opened 2022-07-05
Improving the I/O transparency with `kedro run`


noklam:
## Introduction

With a highly parameterized configuration (`Jinja`, `hydra` or `OmegaConf`), it is not easy to troubleshoot data easily. Often, it is useful to get the full path so users can inspect the data manually. Currently, user need to hack into `context` and do `yaml.dump` to get this information.

i.e. `s3://{base_path}//{special_parameter}` -> Should be compile to `s3://prefix/filename`

Ultimately, the goal is to provide full transparency about the I/O within a `kedro run`, user should be able to get these information for logging or reproducing a particular experiment.

## Background

Related Issues:
1. [kedro compile](https://github.com/kedro-org/kedro/issues/891#issuecomment-930228305) - basically a feature to generate the compiled version of configuration at run-time, potentially also with `catalog.dumps` which is more suitable for Jupyter workflow.
2. https://github.com/kedro-org/kedro/issues/1580 - the run-time `load_version` isn't available to users with `VersionedDataSet` and it's something that we need to fix.
3. Enriching the logging message within `kedro run` - potentially with some DEBUG level message 

## Rollout strategy
There should be no breaking changes, 1 & 2 can be done in parallel. For 3 we can default with no changes and optionally expose more verbose logging.

datajoely:
I would like to say that `dbt compile` is a nice workflow that we could take inspiration from. They have two folders `model` and `build` this is essentially like `src` and `dist` for us in some ways. The `build` part is a neat way of showing what SQL will actually get passed to the database.

https://docs.getdbt.com/reference/commands/compile

AntonyMilneQB:
The compiled filepath is actually already available in the data catalog, just it's quite hidden away: `catalog._get_dataset("dataset_name")._filepath` or `catalog.datasets.dataset_name._filepath`.

1. `kedro compile`: I understand the need for this and I think there's room for something here, but not sure that a new `kedro compile` command is the right way to handle it. While it's not breaking, adding a new CLI command is quite a big thing that's not so easily reversed. The correct solution here should become clearer as we solve the configuration question. e.g. `OmegaConf.resolve` does pretty much what we want here already, so if we use OmegaConf then the question would be how we want to expose that resolved config - maybe it would be more naturally done through a hook writing to a file rather than a separate command.

2. Agreed, but I think this and the heart of the issue here are just symptoms of a more general issue - see below.

[...]

Issue #1654, opened 2022-06-29
Refactor load version logic


merelcht:
## Background
Pre-requisite for https://github.com/kedro-org/kedro/issues/1580 

The code for loading the latest data/fetching the load version needs further refactoring. `resolve_load_version` is called multiple times, where it probably shouldn't. Moreover, inside `kedro-viz` we have replicated some of the logic to load the latest data for experiment tracking, which should be improved as well.

AntonyMilneQB:
Here's one part of kedro-viz where I'm confident a refactoring would make things cleaner. There may be others as well.
https://github.com/kedro-org/kedro-viz/blob/b2011d5b34fdcc7e02974a8c0834c2d70a7dc27f/package/kedro_viz/models/graph.py#L639-L691

AntonyMilneQB:
After this is completed, we should consider what to do with https://github.com/kedro-org/kedro/issues/1691. e.g. should we show `load=None` or the actual load version?

merelcht:
I suggest we close this ticket. I don't think refactoring makes sense now we've decided that we're basically not going to invest much more in our own versioning. Thoughts @astrojuanlu ?

astrojuanlu:
💯 agreed

PR #1911, opened 2022-10-06
Remove redundant `resolve_load_version` call

noklam:

## Description

Related to #1691
Partially related to #1654 but I want to split this out since this is a separate bug.

## Development notes

* Remove a redundant `resolve_load_version` call
* Fix the incorrect test for `MatplotlibWriter`
* Similar fix for `HoloviewsWriter`

Note: 
`MatplotlibWriter` doesn't support load at all. The test is likely copy&paste error, it should be a more specific pattern.
```python
    def _load(self) -> NoReturn:
        raise DataSetError(f"Loading not supported for '{self.__class__.__name__}'")
```

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1911"></a>


SajidAlamQB (approved):
Nice catch, LGTM! 👍

kedro/io/core.py
```diff
@@ -601,8 +600,7 @@ def _get_save_path(self) -> PurePosixPath:
     def _get_versioned_path(self, version: str) -> PurePosixPath:
         return self._filepath / version / self._filepath.name
 
-    def load(self) -> _DO:
-        self.resolve_load_version()  # Make sure last load version is set
```
noklam, line 605: This is actually quite interesting - pylint complains this is a useless parent delegation since the `load` method is exactly the same as `AbstractDataSet`. But let's address this in the refactoring PR instead. It feels like this is the initial design - thus the `load` method was necessary but later other changes come in, it become obsolete for a long time without anyone noticing.

noklam:
More context in this [Refactor PR of versioning](https://github.com/quantumblacklabs/private-kedro/pull/404) in 2+ years ago

kedro/io/core.py
noklam: More context of this [Refactor PR of versioning](https://github.com/quantumblacklabs/private-kedro/pull/404) in 2+ years ago

noklam: I am not sure about the point about extra protection - the function doesn't store any state in the dataset, so calling the function itself won't have any effect.

As the point of this function is hidden in a few levels of indirection - I think this should be more related to #1778

```python
    def resolve_load_version(self) -> Optional[str]:
        """Compute the version the dataset should be loaded with."""
        if not self._version:
            return None
        if self._version.load:
            return self._version.load
        return self._fetch_latest_load_version()
```

tests/extras/datasets/holoviews/test_holoviews_writer.py
```diff
@@ -176,7 +176,9 @@ def test_http_filesystem_no_versioning(self):
 
     def test_no_versions(self, versioned_hv_writer):
```
MerelTheisenQB, line 177: Maybe we should rename this and the other test to `test_load_not_supported` instead.

MerelTheisenQB (approved):
Thanks for investigating and explaining what was going on here @noklam ! ⭐  It's a great improvement that the holoviews and matplotlib datasets now return the correct error message.

tests/extras/datasets/holoviews/test_holoviews_writer.py
noklam: Agree, this will be more accurate - will do

Issue #696, opened 2021-02-23
TensorFlowModelDataset doesn't overides existing local model when trying to save new model at the same local path


aquerkent:
## Description
TensorFlowModelDataset doesn't overides existing local model when saving new model

## Context
The bug occurs when experimenting with different tensorflow models architectures trained and saved by running a kedro pipeline. The versioning of those experiments is done with Kedro-mlflow (@Galileo-Galilei @takikadiri) which manages the upload of the different local training artifacts (including the saved model) in to a remote mlflow server)

## Steps to Reproduce
1. Create a toy tensorflow model and train it by running a dedicated a kedro pipeline 
2. Persist the trained model as a TensorFlowModelDataset declared in the data catalog
3. Change the architecture of the model (by modyfing the node used to create the model archictecture), retrain the model by exectuting the training pipeline again and check to see if your new model has replaced the old model in your local path

## Expected Result
The new trained model should overide the older model that exists at the declared local path

## Actual Result
The new model is silently NOT saved : the local path still contains the old model (no warning or error is shown...)

```

## Your Environment
Include as many relevant details about the environment in which you experienced the bug:

* Kedro version used (`pip show kedro` or `kedro -V`): 0.16.5
* Python version used (`python -V`): Python 3.6.8
* Operating system and version: CentOS Linux 7 (Core)

**Comment by MerelTheisenQB** [comment]
Hi @aquerkent, thanks for reporting this issue. Could you provide the catalog entry you're using for the model?

**Comment by MerelTheisenQB** [comment]
Hi @aquerkent , do you still need help resolving this issue?

**Comment by MerelTheisenQB** [comment]
Hi @aquerkent I'm closing this issue, but feel free to re-open if you still need help with this.

**Comment by datajoely** [comment]
Discord user experiencing [similar issue](https://discord.com/channels/778216384475693066/778998585454755870/958443528931909682)

**Comment by williamcaicedo** [comment]
What I found is that, the old model still exists after saving the new one, and the new one is in a subdirectory. It's like the `fsspec.put` operation copies the root tmp directory that `TensorFlowModelDataset` creates into the dataset's filepath.

Catalog entry:

``` 
tf_model_deploy: 
  type: kedro.extras.datasets.tensorflow.TensorFlowModelDataset
  filepath: data/09_deployment/tf_model 
```

The `_save` method that creates a temporal folder and then makes a call to `fsspec.put`:
```
def _save(self, data: tf.keras.Model) -> None:
[...]

PR #1915, opened 2022-10-12
Enable `TensorFlowModelDataset` to overwrite existing model, and add support for `tf.device`

williamcaicedo:
## Description
- [ ] #696 
 prevents TF models from being saved correctly using an unversioned dataset.

## Development notes
For models using the SavedModel format, a check was added to find out if the save path already exists. In case it does, path is deleted prior to saving the new model.

Removed the code that manually created all save path intermediate directories if they didn't exist, as `_fs_args.setdefault("auto_mkdir", True)` is already being set.

Model uses gpu by default unless `tf_device: cpu` load arg is supplied.

## Checklist

- [*] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [*] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [*] Updated the documentation to reflect the code changes
- [*] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [*] Added tests to cover my changes

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1915"></a>

### DIFF
--- a/kedro/extras/datasets/tensorflow/tensorflow_model_dataset.py
+++ b/kedro/extras/datasets/tensorflow/tensorflow_model_dataset.py
@@ -3,7 +3,7 @@ TensorFlow models.
 """
 import copy
 import tempfile
-from pathlib import Path, PurePath, PurePosixPath
+from pathlib import PurePath, PurePosixPath
 from typing import Any, Dict
 
 import fsspec
@@ -118,14 +118,16 @@ class TensorFlowModelDataset(AbstractVersionedDataSet[tf.keras.Model, tf.keras.M
                 self._fs.get(load_path, path, recursive=True)
 
             # Pass the local temporary directory/file path to keras.load_model
-            return tf.keras.models.load_model(path, **self._load_args)
+            device = self._load_args.pop("tf_device", "gpu")
+            if device == "cpu":
+                with tf.device("/CPU:0"):
+                    model = tf.keras.models.load_model(path, **self._load_args)
+            else:
+                model = tf.keras.models.load_model(path, **self._load_args)
+            return model
 
     def _save(self, data: tf.keras.Model) -> None:
         save_path = get_filepath_str(self._get_save_path(), self._protocol)
-
-        # Make sure all intermediate directories are created.
-        save_dir = Path(save_path).parent
-        save_dir.mkdir(parents=True, exist_ok=True)
 
         with tempfile.TemporaryDirectory(prefix=self._tmp_prefix) as path:
             if self._is_h5:
@@ -138,7 +140,9 @@ class TensorFlowModelDataset(AbstractVersionedDataSet[tf.keras.Model, tf.keras.M
             if self._is_h5:
                 self._fs.copy(path, save_path)
             else:
-                self._fs.put(path, save_path, recursive=True)
+                if self._fs.exists(save_path):
+                    self._fs.rm(save_path, recursive=True)
+                self._fs.put(path, save_path + "//", recursive=True, overwrite=True)
 
     def _exists(self) -> bool:
         try:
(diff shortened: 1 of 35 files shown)


kedro/extras/datasets/tensorflow/tensorflow_model_dataset.py
```diff
@@ -118,15 +118,17 @@ def _load(self) -> tf.keras.Model:
                 self._fs.get(load_path, path, recursive=True)
 
             # Pass the local temporary directory/file path to keras.load_model
-            return tf.keras.models.load_model(path, **self._load_args)
+            device = self._load_args.pop("tf_device", "gpu")
```
deepyaman, line 121: Nit: I'm inclined to call the argument `d
```suggestion
            device_name = self._load_args.pop("device_name", None)
```

kedro/extras/datasets/tensorflow/tensorflow_model_dataset.py
```diff
@@ -118,15 +118,17 @@ def _load(self) -> tf.keras.Model:
...
+            if device == "cpu":
+                with tf.device("/CPU:0"):
```
deepyaman, line 123: I personally think it would be nicer if this more broadly supported `tf.device`, rather than adding a special case for CPU. Maybe, if `device_name` is passed, It can use the `tf.device` context manager with specified `device_name` (that way it's not just restricted to `"/CPU:0"`); else, don't.

deepyaman (changes requested):
Took a quick pass; shouldn't have started at 1:30a, so going to leave it at that for now. 😅 Happy to take a look at any responses later.

kedro/extras/datasets/tensorflow/tensorflow_model_dataset.py
```diff
@@ -138,7 +140,9 @@ def _save(self, data: tf.keras.Model) -> None:
             if self._is_h5:
                 self._fs.copy(path, save_path)
             else:
-                self._fs.put(path, save_path, recursive=True)
+                if self._fs.exists(save_path):
+                    self._fs.rm(save_path, recursive=True)
+                self._fs.put(path, save_path + "//", recursive=True, overwrite=True)
```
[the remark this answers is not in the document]
williamcaicedo: No reason at all 😅 I was just trying to figure out fsspec semantics. I have removed it.

kedro/extras/datasets/tensorflow/tensorflow_model_dataset.py
```diff
@@ -138,7 +140,9 @@ def _save(self, data: tf.keras.Model) -> None:
             if self._is_h5:
                 self._fs.copy(path, save_path)
             else:
-                self._fs.put(path, save_path, recursive=True)
+                if self._fs.exists(save_path):
+                    self._fs.rm(save_path, recursive=True)
```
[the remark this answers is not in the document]
williamcaicedo: According to my tests, `fsspec.put`'s semantics are a bit different than what we expect and the `overwrite` option wasn't really overwriting the old model (as documented in the associated Github [issue](https://github.com/kedro-org/kedro/issues/696)). I have removed the `overwrite` flag from the `fsspec.put` invocation to rely instead on the explicit removal of the old model directory.

kedro/extras/datasets/tensorflow/tensorflow_model_dataset.py
williamcaicedo: I have updated the variable name

kedro/extras/datasets/tensorflow/tensorflow_model_dataset.py
williamcaicedo: Makes sense. I have updated the code to use an explicit device name from `load_args`

deepyaman (approved):
I haven't tested this myself yet, but LGTM from a code review perspective!

williamcaicedo:
> I haven't tested this myself yet, but LGTM from a code review perspective!
> 
> Edit: @williamcaicedo I added a note to PR title that added support for `tf.device`; can you actually include that in the release notes, too?

@deepyaman done 👍

deepyaman:
> > I haven't tested this myself yet, but LGTM from a code review perspective!
> > Edit: @williamcaicedo I added a note to PR title that added support for `tf.device`; can you actually include that in the release notes, too?
> 
> @deepyaman done 👍

Amazing, thanks!

PR #1938, opened 2022-10-17
Update dependabot.yml config 

SajidAlamQB:

## Description

Dependabot has been opening PRs on a target branch, dependency-update, we should change this and make dependabot open PRs only on main to avoid the branch going stale and us forgetting to merge it.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1938"></a>


MerelTheisenQB (approved):
👍

SajidAlamQB:
Also adding `jupyterlab_server` pin as [with their latest release 2.16.0](https://github.com/jupyterlab/jupyterlab_server/pull/316/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711) requires `importlib_metadata >= 4.8.3` this conflicts with `flake8` which has `importlib_metadata <4.3`. `jupyterlab` has `jupyterlab_server` as a [downstream dependency](https://github.com/jupyterlab/jupyterlab/blob/26999c5df85dd8a088580b5dafbc5404ef9b4a44/pyproject.toml#L47) with this requirement, `jupyterlab_server>=2.11.1,<3`.

SajidAlamQB:
> > Also adding `jupyterlab_server` pin as [with their latest release 2.16.0](https://github.com/jupyterlab/jupyterlab_server/pull/316/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711) requires `importlib_metadata >= 4.8.3` this conflicts with `flake8` which requires `importlib_metadata <4.3` creating an unresolvable dependency conflict. `jupyterlab` has `jupyterlab_server` as a [downstream dependency](https://github.com/jupyterlab/jupyterlab/blob/26999c5df85dd8a088580b5dafbc5404ef9b4a44/pyproject.toml#L47) with this requirement, `jupyterlab_server>=2.11.1,<3`.
> 
> This is interesting, so `jupyterlab_server` is a downstream dependency coming through `jupyterlab` and even though they have pinned it we need to pin it as well? I would've thought that the downstream dependencies would be resolved with the pins of the other libraries.. 🤔
> 
> Would this actually be solved when we remove the `flake8` requirement from the template in #1947?

Their pin is slightly too open as they allow `jupyterlab_server>=2.11.1,<3` but we don't want to include 2.16.0 so our pin will be `jupyterlab_server>=2.11.1,<2.16.0`. 

And yes removing flake8 requirements will also resolve this issue.

MerelTheisenQB:
> Their pin is slightly too open as they allow `jupyterlab_server>=2.11.1,<3` but we don't want to include 2.16.0 so our pin will be `jupyterlab_server>=2.11.1,<2.16.0`.

Aaahh I see the difference now! Okay that makes more sense. 
 
> And yes removing `flake8` requirements will also resolve this issue. Would it be better to wait for #1947?

I think this is blocking all builds at the moment, so let's get this temporary fix out. I'm not too keen on adding another dependency in the requirements, but since it will be removed again when we get rid of `flake8` I'm okay with having it in there. Just leave a comment on #1947 so we don't forget to unpin `juptyerlab_server`.

MerelTheisenQB (approved):

noklam (approved):
Thank you, just in time! Just found out the e2etest is failing for py3.7!

Issue #1580, opened 2022-05-30
Exposing load/save version in hooks or included these information in logs?


noklam:
(Created by Nok, converted from Discord Discussion)

# Desciption
A user want to have dataset load/save version logged, potentially like this
```
2022-05-30 10:01:09,039 - kedro.io.data_catalog - INFO - Loading data from `batch` (PartitionedDataSet) (version: 2022-05-15T05.24.31.017Z)
2022-05-30 10:01:18,467 - kedro.io.data_catalog - INFO - Saving data to `processed_batch` (ParquetDataSet)... (version: 2022-05-30T08.23.54.197Z)
```

This is not possible currently as kedro does not track this information, it should belongs to either 
1. Extra hook spec (after_dataset_loaded hook)
2. Session store (experiment tracking)

# How current load version is determined when version=None?
Currently, this `load_version` information is buried deep down in the framework, and it is determined only when a dataset is loaded at runtime.
The details of how "latest" version is in a method called `resolve_load_version`, which further calls `_fetch_latest_load_version`

## Further Studies
For some reason, `resolve_load_version` is being called twice, L594 seems to be a leftover from historical refactoring, need further confirmation.
https://github.com/kedro-org/kedro/blob/b2e59facaa5f97f693287be590b4b4d297db344e/kedro/io/core.py#L558-L564

https://github.com/kedro-org/kedro/blob/b2e59facaa5f97f693287be590b4b4d297db344e/kedro/io/core.py#L594

The refactor PR is here:
https://github.com/kedro-org/kedro/commit/f03226e29b8a018a0f6edab6d3f1a0d37c1b1812

AntonyMilneQB:
FYI I think it's already possible to achieve this since `save_version` and `load_versions` are available in `after_catalog_created`. So you can shift that information into the right `after_dataset_loaded` etc. hooks abd recreate the `load_version` logic by doing something like `load_version = load_versions["dataset_name"] if dataset_name in load_versions else save_version`.

That's obviously not a good solution though, and it would be nice to have a better way of doing it. Currently we expose `dataset_name` and `data` in the dataset hooks. If we want to add `load/save_version` then I wonder whether we should just expose something like the whole `catalog.datasets[dataset_name]`. I've seen people asking for `filepath` in these hooks before, so it might be best just to make the whole thing available rather than adding more and more arguments piecemeal. (Like should we expose `context` vs. `config_loader` in a hook discussion)

AntonyMilneQB:
[...]

PR #1951, opened 2022-10-18
Expose the loaded_version and saved_version for `AbstractVersionedDataSet`

noklam:

## Description

Related #1580 

### How current load version is determined when version=None?
Currently, this `load_version` information is buried deep down in the framework, and it is determined only when a dataset is loaded at runtime.
The details of how "latest" version is in a method called `resolve_load_version`, which further calls `_fetch_latest_load_version`

https://github.com/kedro-org/kedro/blob/b2e59facaa5f97f693287be590b4b4d297db344e/kedro/io/core.py#L558-L564

###  How version is determined?
```python
  def resolve_load_version(self) -> Optional[str]:
      """Compute the version the dataset should be loaded with."""
      if not self._version:  # Not version at all - not sure when will this path executed?
          return None 
      if self._version.load:  # when load-version is specified in CLI or config
          return self._version.load
      return self._fetch_latest_load_version().  # When version=None
```

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1951"></a>

# Alternatives
### 1st attempt - Add new attributes in `AbstractVersionedDataSet`
It may not be a bad idea to add new attributes at all - we can update the `load_version` `save_version` or the `version_cache`, but that way we also lost the initial input which makes things complicated.

[8fe0b1a](https://github.com/kedro-org/kedro/pull/1951/commits/8fe0b1a855e4870d01874cefbcb60b0cdbc7ac85)
image

# More Context
- [ ] https://github.com/quantumblacklabs/private-kedro/pull/404
> Currently every call to `AbstractVersionedDataSet.load()` resolves the load version once again. This may result in inconsistent versioned being passed to different nodes (if new version is created mid-run).

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -13,6 +13,7 @@
 ## Major features and improvements
 * The config loader objects now implement `UserDict` and the configuration is accessed through `conf_loader['catalog']`
 * You can configure config file patterns through `settings.py` without creating a custom config loader
+* Expose latest load version when version is `None`.
 
 ## Bug fixes and other changes
 * Fixed `kedro micropkg pull` for packages on PyPI.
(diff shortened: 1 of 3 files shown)


noklam:
Closed in favor of #1654 as the pre-requisite and we should come back to think about how to expose these information.

Issue #1840, opened 2022-09-07
Destroy the default_project_path variable in IPython extension


AntonyMilneQB:
> **Note**. Related: #1839. Some of the tests that currently exist might not be necessary or would at least be much simpler following this work. 

[From the docs](https://kedro.readthedocs.io/en/stable/tools_integration/ipython.html#managed-jupyter-instances):
<img width="999" alt="image" src="https://user-images.githubusercontent.com/49395058/188971468-e56da5ea-df4a-48ea-a721-db31f9dc6abc.png">

The current implementation uses a global variable `default_project_path` to remember the Kedro project path. This is annoying:
* it's confusing to follow, especially since I want to split the file into multiple modules, which would mean sharing the variable between modules
* it makes unit tests tricky
* plus all the usual reasons people tend to dislike global variables

So let's do it a different way instead. Possible methods include:
1. Turn `%reload_kedro` into a [line magic with state](https://ipython.readthedocs.io/en/stable/config/custommagics.html)
2. Define `reload_kedro` function as a closure inside `magic_reload_kedro` and use `nonlocal`
3. Use `context.project_path` to get the project path once it's been set. This is what we now do in `%run_viz` (`context` is available in `local_ns` there)

> **Note**. Even though we're changing a "public" variable (no `_`) this is definitely a non-breaking change. This variable is not meant to be used externally anywhere and I'm sure it's not (including kedro-viz). The name of it was changed in the past in 0.17.x as a non-breaking change and no one complained, so we can be confident removing it in a non-breaking release.

---

The above all assumes that we want to keep this "memory" behaviour. It is possible that the extra code required to do this doesn't seem to merit this small convenience feature. If this seems to be the case then _in principle_ I'm happy to drop it as a feature, i.e. `%reload_kedro` would need to repeat the project path (but still only when it's needed i.e. kedro project path can't be found automatically from your cwd). This would need a quick bit of user research to check that no one is very attached to the feature. My instinct is that it's a very minor convenience that is rarely used.

MerelTheisenQB:
Do quick user research into whether people are using `%reload_kedro`, e.g. a poll.

noklam:
A side note, some people did use the function `reload_kedro` directly instead of the libe magic `%reload_kedro`, partly because this is the only workaround and `%reload_kedro` arguments only work since `0.18.3`. I think it's nice to keep it working but not a must.

PR #1955, opened 2022-10-18
Destroy default project path variable

jmholzer:
## Description

Resolves [#1840](https://github.com/kedro-org/kedro/issues/1840)

A summary of the issue is as follows:

The global variable `default_project_path` of `kedro.ipython` allows the line magic `%reload_kedro` to remember the project path. Global variables such as these decrease code quality and maintainability for a number of reasons and should be removed. The project path can either be specified by the user, or an attempt can be made to find it based on the current working directory if it is not specified (using `_find_kedro_project`). The `default_project_path` variable is 100% coupled to the 'in-memory' feature of `%kedro_reload` and is decoupled from the entire remaining code base. To remove it we could do one of two things:

1. Decouple the 'in-memory' feature from this variable by changing its implementation.
2. Remove the 'in-memory' feature from %kedro_reload

@AntonyMilneQB gave 3 suggestions of ways to accomplish option 1. @MerelTheisenQB and @AntonyMilneQB asked for user research as a prerequisite to option 2.

Option 1 does not change anything from the user's perspective, it only changes the implementation of the feature in the code base.

This PR implements option 1, because:
- The implementation change is relatively straightforward and simplifies maintenance going forward.
- Removing the 'in-memory' feature would have an impact on some users.

## Development notes

This PR:

1. Refactors `kedro/ipython/__init__.py`:
1.1. By delegating the responsibility for resolving a project path to a dedicated function (single responsibility / separation of concerns).
1.2. By sorting the functions in the file in order of importance.
2. Removes the `default_project_path` global variable while retaining the 'in-memory' feature of `%reload_kedro`, which was re-implemented using the ['local scope'](https://ipython.readthedocs.io/en/stable/config/custommagics.html#accessing-user-namespace-and-local-scope) functionality provided by IPython.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1955"></a>


kedro/ipython/__init__.py
```diff
@@ -21,46 +20,75 @@
...
-):  # pragma: no cover
+    path: str = None,
+    env: str = None,
+    extra_params: Dict[str, Any] = None,
+    local_ns: Optional[Dict[str, Any]] = None,
+) -> None:  # pragma: no cover
     """Function that underlies the %reload_kedro Line magic. This should not be imported
     or run directly but instead invoked through %reload_kedro."""
 
     # If a path is provided, set it as default for subsequent calls
```
noklam, line 83: This is not always true now - if a user is using `reload_kedro` instead of `%reload_kedro`. It only remembers the path when it's using the line magic

tests/ipython/test_ipython.py
```diff
@@ -266,3 +245,23 @@ def test_line_magic_with_invalid_arguments(self, mocker, ipython):
...
+        assert result == expected
+
+    def test_only_local_namespace_specified(self):
+        result = _resolve_project_path(
+            local_ns={"project_path": Path("/test").resolve()}
+        )
+        expected = Path("/test").resolve()
+        assert result == expected
+
+    def no_path_no_local_namespace_specified(self, mocker):
```
noklam, line 263: should it be a test?
```suggestion
    def test_no_path_no_local_namespace_specified(self, mocker):
```

noklam (commented):
Looks quite good already as a 1st draft! I left some minor comments there, I think it's nice to keep the memory behaviour. 

Removing the global should make the ipython testing easier.  Have you manually test the `test_ipython.py` ? As we skipped these tests it would be great to manually test it.

tests/ipython/test_ipython.py
jmholzer: Nice, thanks! Fixed in [211706c](https://github.com/kedro-org/kedro/pull/1955/commits/211706ced9422f0a56e77e7fc0cee8551b067a6c)

kedro/ipython/__init__.py
jmholzer: You're right, fixed this in [72509ec](https://github.com/kedro-org/kedro/pull/1955/commits/72509ec930513d8992818df69b25233d1f48bfeb)

jmholzer:
Thanks for the review @noklam! 😃 I addressed the comments you left individually.

> Removing the global should make the ipython testing easier. Have you manually test the `test_ipython.py` ? As we skipped these tests it would be great to manually test it.

Yes these tests work, do you know why they are being skipped?

MerelTheisenQB:
Any reason why you changed the order of the methods in `ipython/__init__.py`? In the diff it now looks you almost everything was changed/removed 😅

jmholzer:
> Any reason why you changed the order of the methods in `ipython/__init__.py`? In the diff it now looks like almost everything was changed/removed 😅

Yes, this was to improve the structure of the file by making the precedence of functions clearer. Before, the ordering of functions was more or less random. I think these changes will make it easier for newbies (like me) to follow how this module works in the future. This is the advice of Robert Martin in [Clean Code](https://www.amazon.co.uk/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882), which I try to follow.

It makes the diff messy I know :/ but I feel it's warranted in this case.

MerelTheisenQB:
> > Any reason why you changed the order of the methods in `ipython/__init__.py`? In the diff it now looks like almost everything was changed/removed 😅
> 
> Yes, this was to improve the structure of the file by making the precedence of functions clearer. Before, the ordering of functions was more or less random. I think these changes will make it easier for newbies (like me) to follow how this module works in the future. This is the advice of Robert Martin in [Clean Code](https://www.amazon.co.uk/Clean-Code-Handbook-Software-Craftsmanship/dp/0132350882), which I try to follow.
> 
> It makes the diff messy I know :/ but I feel it's warranted in this case.

That makes sense! I think in future it would be helpful if there was a specific commit that only contains the changes to make reviewing easier, because now it's not really possible to tell what the actual change is.

jmholzer:
> That makes sense! I think in future it would be helpful if there was a specific commit that only contains the changes to make reviewing easier, because now it's not really possible to tell what the actual change is.

You're right about this, I'm sorry. The changes are to `_resolve_project_path`, `magic_reload_kedro` and `load_ipython_extension`.

MerelTheisenQB:
> You're right about this, I'm sorry. The changes are to `_resolve_project_path`, `magic_reload_kedro` and `load_ipython_extension`.

Awesome, I'm going to focus on those then!

kedro/ipython/__init__.py
```diff
@@ -21,46 +20,71 @@
...
+    reload_kedro(args.path, args.env, args.params, local_ns)
 
 
 def reload_kedro(
-    path: str = None, env: str = None, extra_params: Dict[str, Any] = None
-):  # pragma: no cover
+    path: str = None,
+    env: str = None,
+    extra_params: Dict[str, Any] = None,
+    local_ns: Optional[Dict[str, Any]] = None,
```
MerelTheisenQB, line 75: What is `local_ns`? I'd be in favour of having longer argument names if that means it's more descriptive.

kedro/ipython/__init__.py
```diff
@@ -84,51 +108,51 @@ def reload_kedro(
...
+        project_path = Path(path).expanduser().resolve()
+    else:
+        if local_ns and "project_path" in local_ns:
+            project_path = local_ns["project_path"]
+        else:
+            project_path = _find_kedro_project(Path.cwd())
+        if project_path:
+            logger.info(
+                "Resolved project path as: %s.\nTo set a different path, run "
+                "'%%reload_kedro <project_root>'",
```
MerelTheisenQB, line 133: Shouldn't this be
```suggestion
                "'%reload_kedro <project_root>'",
```

kedro/ipython/__init__.py
```diff
@@ -84,51 +108,51 @@ def reload_kedro(
...
+        path: the path to use as a string object
+        local_ns: Namespace with local variables of the scope where the line
+            magic is invoked in a dict.
     """
-    args = parse_argstring(magic_reload_kedro, line)
-    reload_kedro(args.path, args.env, args.params)
+    if path:
+        project_path = Path(path).expanduser().resolve()
+    else:
+        if local_ns and "project_path" in local_ns:
```
MerelTheisenQB, line 126: When and where is this `local_ns` thing set?

MerelTheisenQB (commented):
Generally this looks good 👍 I've got some questions specifically about the `local_ns` stuff

kedro/ipython/__init__.py
jmholzer: `local_ns` is the argument that IPython passes to magics that are decorated with `needs_local_scope`, which is the feature that allows this implementation to work. `local_ns` is a dict containing all declarations in the scope of the IPython session and is persisted between calls to magics / extensions being reloaded. Unfortunately, the name must be `local_ns` in the function signature of the magic for IPython to use it correctly. This feature of IPython seems poorly documented, or I would share some docs.

That being said, we should use a more descriptive name elsewhere -- `local_namespace`. I will also change the `project_path` key of `local_ns` to  `_project_path`, to avoid users accidentally overwriting this.

These changes are in [838aaf1](https://github.com/kedro-org/kedro/pull/1955/commits/838aaf1bb7e8251e45e0794991151fc34ac5fb91)

kedro/ipython/__init__.py
jmholzer: Without the second `%`, this line will result in a `TypeError`:
`TypeError: not enough arguments for format string`
The leading `%` is being used to escape the second. It renders / passes tests correctly. Alternative would be to use f-strings, though IIRC the logging library is not optimised for them and the `%` style of format string is currently being used everywhere else in the code base so it may be inconsistent.

kedro/ipython/__init__.py
jmholzer: `local_ns` is passed in by automatically by IPython when the line magic is called. It is a `dict` containing the contents of the local scope. I've tried to explain it a bit more deeply in my comment above.

jmholzer:
> @jmholzer I think we need to manually test out the IPython stuff in GitPod -> try out the `kedro ipython` command and check if the variables are instantiated properly.

Hey Nok, the PR is in a state where it's ready for some manual testing. I have manually tested the changes and am happy that the functionality is the same as before for the cases I could think of. It would be great if you could test some of the cases you think of too.

SajidAlamQB (commented):
This looks good, just had some lingering questions about local_ns and how it exactly works.

kedro/ipython/__init__.py
MerelTheisenQB: Thanks for explaining @jmholzer, this makes sense now.

kedro/ipython/__init__.py
```diff
@@ -21,46 +20,71 @@
...
+    reload_kedro(args.path, args.env, args.params, local_ns)
 
 
 def reload_kedro(
-    path: str = None, env: str = None, extra_params: Dict[str, Any] = None
-):  # pragma: no cover
+    path: str = None,
+    env: str = None,
+    extra_params: Dict[str, Any] = None,
+    local_namespace: Optional[Dict[str, Any]] = None,
```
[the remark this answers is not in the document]
MerelTheisenQB: As far as I understood it, it must be `local_ns` in the `magic_reload_kedro` method above, but it's okay if it's a different name in the other methods.

MerelTheisenQB (approved):
Nicely done! ⭐ 
Don't forget to add this change to the release notes 📝

SajidAlamQB (approved):
Awesome work on this! Tested it manually all looks good. 🌟

noklam (approved):
Thanks! I manually tested the `kedro ipython` and it works fine.

* `kedro ipython`
* `%reload_kedro   # with/without arguments
* Check if the `context`, `pipeline`, `session` object are created correctly
* `session.run` works

tests/ipython/test_ipython.py
```diff
@@ -266,3 +247,56 @@ def test_line_magic_with_invalid_arguments(self, mocker, ipython):
...
+class TestProjectPathResolution:
+    def test_only_path_specified(self):
+        result = _resolve_project_path(path="/test")
+        expected = Path("/test").resolve()
+        assert result == expected
+
+    def test_only_local_namespace_specified(self):
+        # pylint: disable=too-few-public-methods
+        class Context:
+            _project_path = Path("/test").resolve()
```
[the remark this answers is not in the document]
jmholzer: The context for this change is that I moved to storing the project path in `context` (an instance of `KedroContext`), which is a variable in the IPython local namespace. This is the way that Viz does it and it is much safer than storing it in the 'top-level' of the IPython namespace, where a user can accidentally overwrite it.

This means that I had to change the testing implementation, I am using a dummy `Context` class rather than importing `KedroContext` and instantiating this, since its `__init__` takes a lot of arguments that we do not need here.  Do you think I should do the latter instead?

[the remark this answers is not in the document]
merelcht: Thanks for explaining! I'm happy to leave it like this, but perhaps add a small comment in the tests with this `Context` class explaining why it's there. I didn't immediately understand it was the `KedroContext` and not some completely different context used.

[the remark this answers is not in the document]
jmholzer: Great, thanks for taking another look :). Made those changes and will merge

Issue #1743, opened 2022-08-01
Make sure all example code blocks for datasets are runnable


MerelTheisenQB:
### Description

Some of the code examples we provide in the API docs for datasets (https://kedro.readthedocs.io/en/stable/kedro.extras.datasets.html#module-kedro.extras.datasets) aren't actually runnable. Some datasets have easy and straightforward examples that can be copy-pasted and run straight away, others reference setup including S3, but it's not clear these snippets won't be runnable.

### Implementation

Update all code snippets on in the dataset API docs to basic examples that can be run. Move more complex examples to https://kedro.readthedocs.io/en/stable/data/data_catalog.html#use-the-data-catalog-with-the-yaml-api. And in case a simpler example doesn't make sense, clarify that this snippet can't be run as is and what additional setup would be needed.

merelcht:
Part of this task was completed in https://github.com/kedro-org/kedro/pull/1962

Remaining datasets are:

- dask.ParquetDataSet
- holoviews.HoloviewsWriter
- pandas.GBQQueryDataSet
- pandas.GBQTableDataSet
- pandas.SQLQueryDataSet
- pandas.SQLTableDataSet
- redis.PickleDataSet
- all Spark datasets
- tensorflow.TensorFlowModelDataset

PR #881, opened 2021-09-08
Add GML and GraphML NetworkX datasets

sbrugman:
## Description
Introduces feature described in https://github.com/quantumblacklabs/kedro/issues/828

## Development notes
Follow up from https://github.com/quantumblacklabs/kedro/pull/830 with comments processed, based on the lastest development branch. Thanks @ignacioparicio for reviewing!
Omitted GPickleDataSet as GPickle is deprecated in NetworkX and will be removed in the third version ([ref](https://networkx.org/documentation/stable/_modules/networkx/readwrite/gpickle.html#write_gpickle)).
Included GraphML in addition to GML. Some tools prefer GraphML as it's able to perfectly preserve internal property maps ([ref](https://graph-tool.skewed.de/static/doc/graph_tool.html#graph_tool.load_graph)).

## Checklist

- [X] Read the [contributing](https://github.com/quantumblacklabs/kedro/blob/master/CONTRIBUTING.md) guidelines
- [X] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [X] Updated the documentation to reflect the code changes
- [X] Added a description of this change and added my name to the list of supporting contributions in the [`RELEASE.md`](https://github.com/quantumblacklabs/kedro/blob/master/RELEASE.md) file
- [X] Added tests to cover my changes

## Notice

- [X] I acknowledge and agree that, by checking this box and clicking "Submit Pull Request":

- I submit this contribution under the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0.txt) and represent that I am entitled to do so on behalf of myself, my employer, or relevant third parties, as applicable.
- I certify that (a) this contribution is my original creation and / or (b) to the extent it is not my original creation, I am authorised to submit this contribution on behalf of the original creator(s) or their licensees.
- I certify that the use of this contribution as authorised by the Apache 2.0 license does not violate the intellectual property rights of anyone else.

PR #1962, opened 2022-10-21
Validate code snippets in datasets documentation - Part 1

AhdraMeraliQB:
## Description
This follows on from #1743. In this PR I have tested _all the datasets that can be instantiated locally_. The following datasets remain to be tested in part 2:

- dask.ParquetDataSet
- holoviews.HoloviewsWriter
- pandas.GBQQueryDataSet
- pandas.GBQTableDataSet
- pandas.SQLQueryDataSet
- pandas.SQLTableDataSet
- redis.PickleDataSet
- all Spark datasets
- tensorflow.TensorFlowModelDataset

The main changes in this PR are as follows:

- Some code blocks had duplicate lines for providing filepath arguments. I believe this was to illustrate both local and external (S3/GCS) filesytems, but after discussion with @MerelTheisenQB we have agreed this is a bit redundant.
- Some comments rendered with weird line breaks. This was fixed.
- There was a comment for a dependency to be installed within an example code block. This has been moved out to `extras_require` in `setup.py`
- Added missing imports

Also warranting a flag 🚩: **GraphMLDataSet and GMLDataSet appear to be identical in purpose and in implementation**

#### Both the YAML and Python examples have been tested under the following setup:

- New conda environment
- `pip install kedro`
- `pip install kedro[all]`

##### For Python:

`kedro jupyter lab`
Code snippets were copied and pasted over
An additional cell was added to check `<dataset name>.exists()` returned True

##### For YAML:

`kedro new -s spaceflights`
The dataset was added to the data catalog in `conf/base/catalog.yml` and added to the pipeline arbitrarily.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1962"></a>


kedro/extras/datasets/json/json_dataset.py
```diff
@@ -28,17 +28,13 @@ class JSONDataSet(AbstractVersionedDataSet[Any, Any]):
         >>> json_dataset:
         >>>   type: json.JSONDataSet
         >>>   filepath: data/01_raw/location.json
-        >>>   load_args:
```
AhdraMeraliQB, line 31: JSON Dataset does not support the `load_args` parameter: https://kedro.readthedocs.io/en/stable/kedro.extras.datasets.json.JSONDataSet.html#kedro.extras.datasets.json.JSONDataSet.__init__

MerelTheisenQB:
I found the PR where the GraphMLDataSet and GMLDataSet were added: https://github.com/kedro-org/kedro/pull/881 It looks like both of them were added because "Included GraphML in addition to GML. Some tools prefer GraphML as it's able to perfectly preserve internal property maps ([ref](https://graph-tool.skewed.de/static/doc/graph_tool.html#graph_tool.load_graph))." 

So let's just keep them both.

kedro/extras/datasets/json/json_dataset.py
MerelTheisenQB: Great catch!

setup.py
```diff
@@ -78,6 +78,7 @@ def _collect_requirements(requires):
     "pandas.XMLDataSet": [PANDAS, "lxml~=4.6"],
     "pandas.GenericDataSet": [PANDAS],
 }
+pickle_require = {"pickle.PickleDataSet": ["compress-pickle~=2.1.0"]}
```
MerelTheisenQB, line 81: The original comment said: "# Add "compress_pickle[lz4]" to requirements.txt" Do we need to specify [lz4] for this requirement?

MerelTheisenQB (approved):
Great work on going through all these examples and testing them! 👍 ⭐

AhdraMeraliQB:
@MerelTheisenQB 
> I found the PR where the GraphMLDataSet and GMLDataSet were added: #881 It looks like both of them were added because "Included GraphML in addition to GML. Some tools prefer GraphML as it's able to perfectly preserve internal property maps ([ref](https://graph-tool.skewed.de/static/doc/graph_tool.html#graph_tool.load_graph))."
> 
> So let's just keep them both.

I took a look at the ref but couldn't see the preference; is it the name that they prefer? As the implementation is carbon copies 😬

ankatiyar (approved):
Looks good to me!

MerelTheisenQB:
> I took a look at the ref but couldn't see the preference; is it the name that they prefer? As the implementation is carbon copies 😬

The difference between `GraphML` and `GML` (apparently) is that `GraphML` preserves internal property maps, but `GML` doesn't. See this comment in the docs: "The only file formats which are capable of perfectly preserving the internal property maps are “gt” and “graphml”. Because of this, they should be preferred over the other formats whenever possible." https://graph-tool.skewed.de/static/doc/graph_tool.html#graph_tool.load_graph

Perhaps the question is whether our users make use of both the datasets or not. It's not an extreme maintenance burden so I'd just keep them both.

setup.py
AhdraMeraliQB: Good catch - this is what the `pip list` showed after manually installing `compress_pickle[lz4]`, I'll make the change to follow the dask example further up

PR #1975, opened 2022-10-26
Add dot nested credentials support

nickolasrm:
## Description
Added dot nested credentials support. With this feature, it is possible to replicate the same functionality seen in nested entries with the globals file. Check out the example below:

```yml
# catalog
dataset:
  type: pandas.CSVDataSet
  filepath: abfs://....csv
  credentials: azure.storage.credentials
```

```yml
# credentials
azure:
  storage:
    credentials:
      account_name: test
      anon: false
```

## Development notes
* Modified the `get_credentials` function to reduce the credentials dict with `split(".")` parts, instead of accessing directly with `dict[key]`
* Added two test cases, one for sane dot nested credentials, and another for dot nested but missing the key

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1975"></a>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -23,6 +23,7 @@
 * Updated Experiment Tracking docs with working examples.
 * Modified implementation of the Kedro IPython extension to use `local_ns` rather than a global variable.
 * Refactored `ShelveStore` to it's own module to ensure multiprocessing works with it.
+* Added dot nested credentials support.
 
 ## Minor breaking changes to the API
 
--- a/docs/source/data/data_catalog.md
+++ b/docs/source/data/data_catalog.md
@@ -403,6 +403,11 @@ CSVDataSet(
 )
 ```
 
+```{note}
+You can use the same dot notation as the one used with `globals.yml` to reference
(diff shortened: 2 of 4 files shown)


nickolasrm:
@noklam is this a valid pr for kedro team?

noklam:
@nickolasrm Thank you for drafting this PR.

I think using `reduce` is quite neat, but this is inconsistent with how `credentials` work. I am quite confident this is by design you pass in the name of the `credentials` instead of the exact key, @idanov may have a better idea why this is implemented this way. 

Secondly, overloading `.` isn't a good idea as this immediately raises the question of why this `.` notation isn't used for the other attributes. For `globals`, it's enabled by the Jinja syntax, so you explicitly use a key like `${key}`

nickolasrm:
@noklam thanks for the feedback, it does make more sense now! I think I'm gonna figure out a solution using prefixed keys like a flattened dict, because the only purpose of nesting credentials, in my case, is for organizing things.

I'm closing the pr, and thanks again for the reply!

PR #2005, opened 2022-11-07
Draft Pull Request : Add incremental run method

pedro-sarpen:
## Description
Add `AbstractRunner.run_incremental` method. 
Allows to automatic skip re-running already existing intermediate node outputs.

## Development notes
This is a RFC PR.
The feature seems quite useful to avoid re-running expensive nodes. Yet there was no similar feature in the API.
Explicity specifying the nodes is tedious and may be prone to errors, requiring multiples runs to figure which nodes one needs to run.

Any comments on if it is the best place to add the feature or if there is a better solution are welcome.

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/2005"></a>

### DIFF
--- a/kedro/runner/runner.py
+++ b/kedro/runner/runner.py
@@ -128,6 +128,43 @@ class AbstractRunner(ABC):
 
         return self.run(to_rerun, catalog, hook_manager)
 
+    def run_incremental(
+        self,
+        pipeline: Pipeline,
+        catalog: DataCatalog,
+        hook_manager: PluginManager = None,
+        session_id: str = None,
+    ) -> Dict[str, Any]:
+        """Run the ``Pipeline`` using the datasets provided by ``catalog``
+        and save results back to the same objects.
+        Existing intermediate outputs are not processed.
+
+        Args:
+            pipeline: The ``Pipeline`` to run.
+            catalog: The ``DataCatalog`` from which to fetch data.
+            hook_manager: The ``PluginManager`` to activate hooks.
+            session_id: The id of the session.
+
+        Raises:
+            ValueError: Raised when ``Pipeline`` inputs cannot be satisfied.
+
+        Returns:
+            Any node outputs that cannot be processed by the ``DataCatalog``.
+            These are returned in a dictionary, where the keys are defined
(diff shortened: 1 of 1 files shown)


noklam:
Thank you for your PR, this looks quite similar to this.

What is not supported by this already?

https://github.com/kedro-org/kedro/blob/f52249b262b8716d40cbfac80de0198980390bb2/kedro/runner/runner.py#L94-L121

pedro-sarpen:
> Thank you for your PR, this looks quite similar to this.
> 
> What is not supported by this already?
> 
> https://github.com/kedro-org/kedro/blob/f52249b262b8716d40cbfac80de0198980390bb2/kedro/runner/runner.py#L94-L121

`run_only_missing` looks only for pipeline inputs / outputs, intermediate nodes are not accounted.
For example, if a pipeline output is missing it still run intermediate nodes even if a CachedDataSet for them already exists.

merelcht:
Hi @pedro-sarpen, thanks for raising this PR! This is quite a big feature, so the maintainer team would like to discuss it in a Technical Design meeting to decide on best implementation. It might take a while for us to get to it, but know that we haven't forgotten your PR 🙂

jmholzer:
Tech design 01/02/23:

- We decided to open a new issue to gather thoughts and discuss this in more depth in a future tech design session.

jmholzer:
Hey @pedro-sarpen thanks for opening this PR 🙂.

As an update for you, the next steps for this feature are:

1. A new parent ticket which states and frames the problem (nodes with persisted outputs being re-run unnecessarily).
2. A technical design session (or investigation ticket) to define the possible solutions.
3. An implementation ticket.

Generally the way we consider new features in Kedro is by opening an issue with a high-level discussion of the problem before we go to problem-solving and implementation (starting at step 1 above).

This is a great suggestion and it's good to have this issue on our radar, though this PR goes straight to step 3. We may find a better solution to the underlying problem by starting at step 1.

I will leave this PR open, as it may be we decide this is the right way to go. In the mean time, I will open a ticket to kickstart the process above 🙂.

PR #2030, opened 2022-11-16
Make Kedro compatible with Airflow 2.4.2

rxm7706:
## Description
To Make Kedro Compatible with Airflow 2.4.2
https://github.com/kedro-org/kedro-plugins/issues/73

## Development notes
Updates the requirements on [attrs](https://github.com/python-attrs/attrs) to permit the latest version.
Changed dependency/requirements.txt
From attrs~=21.3 to attrs >=22.1.0,<23.0

## Checklist

- [] Updated the documentation to reflect the code changes
- [] Added tests to cover my changes

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/2030"></a>


noklam:
Thanks for the PR.

Since airflow isn't a core dependency of Kedro, I think it is ok to lift the upper bound to make it compatible with airflow, but at the same time we should keep the lower bound. There could be some other libraries pinned a lower version and kedro should allow that.

What do you think?

rxm7706:
> Thanks for the PR.
> 
> Since airflow isn't a core dependency of Kedro, I think it is ok to lift the upper bound to make it compatible with airflow, but at the same time we should keep the lower bound. There could be some other libraries pinned a lower version and kedro should allow that.
> 
> What do you think?

@noklam I though about using attrs >=21.3,<23.0 ; but looking at the limited Kedro code dependency and usage of attrs ; and assuming Airflow compatibility was more important - I went with >=22.1.0,<23.0. 

But thinking through it a bit more, you have a good point - Pinning the lower bound of attrs so high, will actually break compatibility with older versions of airflow (e.g. airflow 2.2)

I think a better pin will be attrs >=20.0,<23.0 - which will give us backward compatibility to Airflow 2.0 & Airflow 1.X ; from the time attrs was introduced as an airflow dependency. Let me give that a shot - attrs >=20.0,<23.0

noklam:
@rxm7706 `attrs` was always a 2nd-level of dependencies until couple of months ago I added it into our core dependency. I think we are using some features only introduced in `21.3`, thus the pinned version. Feel free to try it out tho!

See the release note of `attrs` here:
https://github.com/python-attrs/attrs/releases

rxm7706:
@noklam Is it odd that only one set to tests failed - Just E2E Windows and Python ; and it seems to be a connection time out while launching a notebook, Attrs bounds itself don't seem to have caused any failures

https://app.circleci.com/pipelines/github/kedro-org/kedro/15100/workflows/6d48b1bb-790c-4d42-ab2c-1256c5472630/jobs/192465

Is there a way to rerun the CI suite ? Without a resorting to a PR just to trigger it.

rxm7706:
@noklam changes made as suggested attrs>=21.3, <23.0 
The failure in CI is again jupyter launch timeout
https://app.circleci.com/pipelines/github/kedro-org/kedro/15106/workflows/bcb867ba-061d-4d47-b4c1-863ec8e9fd4b/jobs/192591

RELEASE.md
```diff
@@ -34,6 +34,7 @@
 * Modified implementation of the Kedro IPython extension to use `local_ns` rather than a global variable.
 * Refactored `ShelveStore` to it's own module to ensure multiprocessing works with it.
 * `kedro.extras.datasets.pandas.SQLQueryDataSet` now takes optional argument `execution_options`.
+* Updated Attrs minimum required version to >=21.3, and increased max to 23.0 to make Kedro compatible with Airflow >2.4.2 that requires 22.1.0
```
deepyaman, line 37:
```suggestion
* Updated `attrs` upper bound to 23.0 to make Kedro compatible with Airflow >2.4.2 that requires 22.1.0.
```
Since the lower bound wasn't changed in the end (and also some minor formatting).

dependency/requirements.txt
```diff
@@ -1,5 +1,5 @@
 anyconfig~=0.10.0
-attrs~=21.3
+attrs>=21.3, <23.0  # The minimum version required for Airflow 2 is >=20.0 and Airflow 2.4.2 is >=22.1
```
deepyaman, line 2: Maybe we shouldn't have an upper bound? https://www.attrs.org/en/22.1.0/changelog.html?highlight=upgrade#changelog They have a very strong backward-compatibility policy.

deepyaman: Also, I'm not sure if we need to explain Airflow bounds here; the 21.3 is a Kedro-based lower bound, and we should just strive to be generally loose beyond that.

deepyaman (approved):
Thanks for the PR! Looks good overall, but left a couple small comments and thoughts.

Also, can you shorten the PR title a bit, please? :) Maybe just drop the part after the hyphen.

dependency/requirements.txt
noklam: I think it's ok to remove the upper bound since I don't remember `attrs` have ever created a problem for us. Though technically it wouldn't matter too much since it's a CalVer release.

noklam:
@rxm7706 The DCO is failing, would you be able to fix it? I removed the upper bound, it should be fine unless the tests tell otherwise.

noklam (approved):

rxm7706 (commented):
Signed-off-by: rxm7706 <95496360+rxm7706@users.noreply.github.com>

rxm7706 (commented):
Signed-off-by: rxm7706

rxm7706:
> @rxm7706 The DCO is failing, would you be able to fix it? I removed the upper bound, it should be fine unless the tests tell otherwise.

@noklam I can't seem to fix the DCO - which is a strange fail, because I used the github UI to approve & make the change - If you cant fix it, just let me know - might just close this PR and do a new one

PR #2041, opened 2022-11-20
Add missing part for the dict-like interface of `_ProjectPipelines`

noklam:

## Background
The `pipelines` object is designed as a lazy dictionary. It only loads the pipeline code when it is being accessed.

## Description

A slight improvement mainly for interactive workflow. Modify the keys/values/items method to return the inner dict, which has a more meaningful __repr__ when printed. The current `pipelines` is a dict-like object but it behaves a bit different from dictionary and could confuse users when interacting with the object.

We had these code to make the `pipelines` appears like a `dict`.
```python
    # Presentation methods
    __repr__ = _load_data_wrapper(repr)
    __str__ = _load_data_wrapper(str)
```
image

## Development notes

* Release Notes
* Update the `_ProjectPipeline` docstring
* Implement the missing interface of dict-like behavior

Arguably the `.items()` method doesn't really need the change, but it feels more consistent to have all 3 methods refer to the `dict`. Previously it returns an `abc.XXXView` which isn't particular helpful when you print out the object.

## Before changes
image

## After Changes
image

## Potential Alternative
* Implement the `KeyView`, `ValueView` and `ItemView` 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/2041"></a>


noklam:
Another question comes up when I read the source code it said it is a read-only lazy dictionary, but it never was. 
https://github.com/kedro-org/kedro/blob/45a79d0fb8dc8fc97ab14091f23d83fe69bee8c8/kedro/framework/project/__init__.py
```python
class _ProjectPipelines(MutableMapping):
    """A read-only lazy dictionary-like object to hold the project pipelines.
    On configure it will store the pipelines module.
    On first data access, e.g. through __getitem__, it will load the registered pipelines and merge
    them with pipelines defined from hooks.
    """
```

noklam:
The PR introduced Lazy Pipeline
* https://github.com/quantumblacklabs/private-kedro/pull/996/
* [JIRA](https://jira.quantumblack.com/browse/KED-2393)

> [idanov](https://github.com/idanov) [on Feb 22, 2021](https://github.com/quantumblacklabs/private-kedro/pull/982#discussion_r580170426)
These objects will be created upon import of the module in Python, which is something we should avoid. I believe in the Jira ticket it was stated to check how we can make the dictionary here lazy, i.e. you only instantiate the pipeline object if you are about to use it.

> The reason for that is that when people work, they often have broken pipelines and by making the instantiations of the objects lazy, you allow them to run the pipelines that are not broken without needing to fix the broken ones.

> Long time ago, we had user complains about this and that's why the convention of defining your pipeline in a function remained, thus the infamous xxxx.create_pipeline() call. So making the dictionary here lazy is a hard requirement unfortunately. Could you think of a way to that easily?

I am quite unsure if this lazy loading is working at all. Since any of the breaking code would result in error immediately. I am not sure if I understand this correctly.


This [attempt PR](https://github.com/quantumblacklabs/private-kedro/pull/991/files) seems to be doing the lazy loading in a more meaningful way, however hooks for register pipelines isn't used anymore?

kedro/framework/project/__init__.py
```diff
@@ -184,6 +184,16 @@ def configure(self, pipelines_module: Optional[str] = None) -> None:
...
 
+    def keys(self):
+        self._load_data()
+        return self._content.keys()
+    def values(self):
+        self._load_data()
+        return self._content.values()
+    def items(self):
+        self._load_data()
+        return self._content.items()
```
deepyaman, line 195:
```suggestion
    keys = _load_data_wrapper(operator.methodcaller("keys"))
    values = _load_data_wrapper(operator.methodcaller("values"))
    items = _load_data_wrapper(operator.methodcaller("items"))
```
and move it below to the `# Dict-like interface` section?

I think it's functionally equivalent to what you have (haven't tested it, just in my mind 🙈), just a bit more consistent with the existing code, and doesn't duplicate the `self._load_data()` and `self._content` access.

But yeah, makes sense to me overall to do this, and to provide a fuller dict-like interface. To be honest, it just took me a while to understand in my mind that the purpose of this PR is to essentially add missing parts of that dict-like interface.

deepyaman (approved):
Approved with suggestion

Also, I think this is worth adding to release notes, etc.

kedro/framework/project/__init__.py
noklam: That's a better summary and I put that as the PR description instead.

merelcht (approved):
LGTM! 👍 

Just one thing I was wondering about is whether we should add this functionality somewhere in the docs so that users know it exists?

noklam:
@merelcht I think it's ok to leave it, as we already introduce this as a "dictionary" to the user. As long as it behaves consistently like a dictionary I think it should be fine. The Laziness is also not important as a user-facing feature, it's more an internal thing.

image

PR #2046, opened 2022-11-21
Rewriting the IPython test

### LINKED ISSUE #1839: Rewrite IPython extension tests
> **Note**. Somewhat related and probably best done in parallel or by the same person (@noklam? 😀): #1840.

test_ipython.py is a bit of a mess. There's way too much mocking going on and it causes quite a few problems. I strongly suspect that it's not testing what we think it tests either, e.g. in `test_load_kedro_objects` if you alter the paths from `kedro_path / ...` to something else then the test will still pass. 

I think we should rewrite the whole file. This isn't necessarily a huge task in terms of writing code, because I suspect that some of the tests there are basically pointless even if they worked. But it will take some careful thought.

First stage of this is to ask _what do we actually want to test here?_ 
* If it's the behaviour of underlying functions (like `reload_kedro` _function_) then which parts of that are actually valuable to test? Are those best tested through unit tests or e2e `behave` test instead?
* If it's ipython-specific behaviour (like line magics) then there should be ipython-specific ways to test that without mocking so much. e.g. https://medium.com/@davide.sarra/how-to-test-magics-in-ipython-extensions-86d99e5d6802 and Google how to test ipython line magics, extensions, etc. @noklam recently introduced `IPython.testing.globalipapp.get_ipython` to the tests, which feels like a much better solution than all the mocking

Second stage is to write the new tests! Don't be afraid to delete old tests if they are no longer valuable. A good example of tests that will no longer be valuable is those that test behaviour of `default_project_path`. The flow for this should be way simpler and easier to test (maybe not even need tests at all depending on how it works) after #1840.

> **Warning**. Much of this file is currently marked as `pragma: no cover` so that test coverage doesn't break CI due to the skipped tests. Fixing tests should increase coverage and then these `no cover`s should be removed.

noklam:

## Description

Fix #1839

## Development notes

The plan is to break down `TestLoadKedroObjects` into proper unit tests. the original tests are testing many things at the same time and arguably duplicated. The e2e behaviour is tested in `behave` already, so this test should really just focus on one thing at a time. This can reduce the mocking a little bit (a few mocking are still needed), but they are now easier to reason.

Main Changes:
* Remove all the mocking of `IPython` when possible, replace with the `ipython` fixture instead
* Remove the duplicate tests - tests are now more atomic with single focus - there are more but shorter tests.
* Bug fixes - previously the assertion about `pipeline` are wrong
* Mock the `pipelines` object instead of `_ProjectPipelines`, previously the tests will pass/fail sometimes because of the order of instantiation of the `global` or `mocking` sometimes flipped.

Notes:
https://github.com/kedro-org/kedro/blob/82d69f43b3d8ce93605c272557e22547bca5bf11/kedro/framework/session/session.py#L258-L260

`_ProjectPipelines` is supposed to be lazy-loaded, but I found out that this line of code triggers the `_load_data` twice for some reason.  Need more investigation.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/2046"></a>


noklam:
I cannot I passed the linting in one go.

merelcht (approved):
The tests are so much cleaner and easier to follow now! Great work on refactoring them ⭐ 

> _ProjectPipelines is supposed to be lazy-loaded, but I found out that this line of code triggers the _load_data twice for some reason. Need more investigation.

Did you already found out what's happening with the above or do we need a follow up issue for that?

PR #2073, opened 2022-11-28
Remove outdated credentials.yml warning

### LINKED ISSUE #1952: Remove/change log level for "Credentials not found in your Kedro project config"
## Description
When getting the credentials in the KedroContext, we specifically look for the `credentials` file. If a project doesn't make use of this file a warning log will be thrown saying "Credentials not found in your Kedro project config". Not all users use the credentials file and so this warning isn't applicable. 

## Context
We're adding functionality to allow users to have credentials in environment variables (which is currently also possible with the `TemplatedConfigLoader`), which makes this error irrelevant.

AhdraMeraliQB:

## Description

Fixes #1952 

**Note:** Tests for credentials handling live in `test_data_catalog.py`

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/2073"></a>


kedro/io/data_catalog.py
```diff
@@ -42,10 +43,16 @@ def _get_credentials(
...
+        MissingConfigError: When no credentials file is found in the
+            project.
 
     """
     try:
         return credentials[credentials_name]
+    except TypeError as exc:
+        raise MissingConfigException(
+            f"Credentials not found in your Kedro project config."
+        )
```
noklam, line 55: Would it be better if this moved before the `try` block? When this function is called but `credentials` is None then it indicates it doesn't have the configuration.

merelcht:
I think these tests don't exist because they're not fully valid cases:

> Credentials file exists but does not have the credentials specified in dataset

This should at maximum just emit a logging message, but definitely not an error. It's completely fine for a `credentials.yml` file to exist without it being used. It shouldn't stop the pipeline from running. 

> No credentials file provided but credentials are specified in dataset

The issue here is that credentials aren't necessarily coming from a file. It could be that users have their own implementation where they make use of a vault or environment variables. That's also the reason why the original logging message should be removed.

noklam:
> > No credentials file provided but credentials are specified in dataset
> 
> The issue here is that credentials aren't necessarily coming from a file. It could be that users have their own implementation where they make use of a vault or environment variables. That's also the reason why the original logging message should be removed.

@merelcht 
I don't think this necessarily assumes file-based credentials. Even if the user somehow injected Vault secret to the `ConfigLoader`, it should still respect the `credentials` key in the `DataCatalog`. In that case, if the `DataCatalog` is looking for credentials but there are no credentials given `None`(files or some other way), it is reasonable to raise a `MissingConfiguration` error.
However, it would be great to see an implementation that injects credentials, as I am not sure if we have a preferred way to do so, as many of these methods are protected method

AhdraMeraliQB:
@merelcht 

> I think these tests don't exist because they're not fully valid cases:
> 
> > Credentials file exists but does not have the credentials specified in dataset
> 
> This should at maximum just emit a logging message, but definitely not an error. It's completely fine for a `credentials.yml` file to exist without it being used. It shouldn't stop the pipeline from running.

Ah, I think I may have been a bit ambiguous in my description - this case doesn't describe the existence of credentials that are unused, but rather datasets that specify credentials that cannot be found in the file/location provided when resolving credentials.

Currently this behaviour would already produce a KeyError: https://github.com/kedro-org/kedro/blob/main/kedro/io/data_catalog.py#L50 . The test `test_empty_credentials` was just added to confirm that this behaviour plays out as it should.


> 
> > No credentials file provided but credentials are specified in dataset
> 
> The issue here is that credentials aren't necessarily coming from a file. It could be that users have their own implementation where they make use of a vault or environment variables. That's also the reason why the original logging message should be removed.

The current assumption would be that the datacatalog credentials would no longer be None if credentials are otherwise provided, but let's have a chat to walk through the possible options - I might be missing something.

noklam (dismissed):
Thanks for looking into it, I'm ok with this fix since we are looking into the refactoring with `OmegaConf` later.

Don't forget to summarise the decision!

In ideal case, I still prefer `MissingConfigException` over `KeyError` when possible. However, this is tricky as we don't offer a clear interface for people to inject credentials from Credential Managers like Key Vault. This feels like a responsibility of `ConfigLoader`, but the easiest way to override it currently is to inject the credentials via `KedroContext`.

kedro/framework/context/context.py
```diff
@@ -335,8 +335,7 @@ def _get_config_credentials(self) -> Dict[str, Any]:
         """Getter for credentials specified in credentials directory."""
         try:
             conf_creds = self.config_loader["credentials"]
-        except MissingConfigException as exc:
-            warn(f"Credentials not found in your Kedro project config.\n{str(exc)}")
```
idanov, line 339: Could we simply do this at debug level instead of fully removing it?

AhdraMeraliQB: Wouldn't that still be misleading to users who are parsing through their credentials without using the credentials.yml file?

idanov: I don't think so, we don't mention anything about `credentials.yml` file, we just say there's no config for the credentials, which could mean that there's no file, or fetching the credentials failed in any other way.

AhdraMeraliQB:
Following a chat with @idanov the following decisions have been made:

1. Changing the warning to debug level instead of removing it entirely.

When considering alternative ways users may provide credentials, @idanov explained that this would be done with the `after_context_created` hook, _before_ the first call to create the catalog. This means by the point the execution reaches `context._get_config_credentials`, credentials should have already been provided. 

For this reason, it has been decided that the exception should not be swallowed silently, but should raise the warning at debug level.


2. Handling no credentials and missing credentials separately.

These cases can be ambiguous, so I will define them further:
-  Missing credentials: A credentials configuration has been provided but is missing some or all of the specific credentials listed under datasets in the data catalog.
- No credentials: No credentials configuration has been provided.

@noklam suggested handling these two cases separately, with the implementation seen in [this commit](https://github.com/kedro-org/kedro/pull/2073/commits/b5f00d5a5c3bc5f4e46d00f8e1ec149d8ca6e345). However, after discussion with @merelcht and @idanov this was deemed unnecessary, and that the error raised [here](https://github.com/kedro-org/kedro/blob/main/kedro/io/data_catalog.py#L49) is more than sufficient. @idanov also highlighted that Kedro may also be used as a library (as opposed to through the cli) in which case the `MissingConfigException` may be misleading.

For this reason, it's been decided that the current implementation is fine as it is; there's no reason to handle the cases differently.

merelcht (approved):
LGTM! 👍 Don't forget to add this to the release notes.

idanov (approved):
👌

Issue #1909, opened 2022-10-06
Allow users to pass credentials through environment variables


merelcht:
## Description
In the new config loader (https://github.com/kedro-org/kedro/issues/1868), allow users to pass credentials through environment variables using the `OmegaConf` resolver `oc.env`: https://omegaconf.readthedocs.io/en/latest/custom_resolvers.html#oc-env

## Context
The "default" way for Kedro to deal with credentials is through the `credentials.yml` file. @AntonyMilneQB has explained very thoroughly how this works and why it's problematic here: https://github.com/kedro-org/kedro/issues/1646

Over time many users have flagged that they want to be able to pass credentials through environment variables. Currently, this is possible if users use the `TemplatedConfigLoader`, but ideally this functionality would be available by default. 

Insight 11 of the configuration research also showed that users are currently storing credentials in environment variables: https://github.com/kedro-org/kedro/issues/1847 

## Implementation

- The new configloader will use `OmegaConf` instead of `anyconf` https://github.com/kedro-org/kedro/issues/1868
- The built-in resolvers will be turned off
- In this task add logic to load credentials using `OmegaConf` with the oc.env resolver turned on. This might mean separating the credentials loading logic so that the resolver for that loading process hasn't been turned off. If this is impossible or really difficult, we'll allow env variables for all config, but that should be discussed. 

- Document how should users pass credentials - as current way to do it is via `TemplatedConfigLoader

noklam:
One question I have it, why `TemplatedConfigLoader` isn't the default already? If this ability to use resolver (using environment variable like a templated variable) is enabled in the basic `ConfigLoader`, it feels weird that a user had to switch to another class to use a template variable that isn't an environment variable.

Effectively, people can then just use the environment variable as a template variable and bypass the `TemplatedConfigLoader` most of the time. 

Would this still work with `Jinja2`?

MerelTheisenQB:
That's a great question @noklam and my preference would be to make the `TemplatedConfigLoader` the default (with some adjustments) for the 0.19 series. I think the main reason for it not being the default is that it was a contribution from outside of the core team. See the history of config loader classes:

![Screenshot 2022-08-02 at 12 18 07](https://user-images.githubusercontent.com/49397448/194553969-a061b2fc-74d6-4fcf-b5b9-843a2c5ced37.png)

[...]

Issue #1868, opened 2022-09-23
Add a new configloader with `OmegaConf` as replacement for `anyconf` 


merelcht:
## Introduction

As part of the work on improving configuration in Kedro we assessed alternatives for `anyconfig`. `anyconfig` was initially chosen because it supports reading different configuration formats. In practice, most users use `yml` files and so we want to use an alternative library that offers better functionality for `yml` configuration.

In https://github.com/kedro-org/kedro/issues/1657 it was decided to use `OmegaConf` as this replacement. More info about that conclusion can be found here: https://github.com/kedro-org/kedro/issues/1657#issuecomment-1181524403

We also did research to verify what config file formats users use in https://github.com/kedro-org/kedro/issues/1703 

## Implementation
In order to introduce this change in a non-breaking way we have to do the following:
1. Add a new configloader
2. Add `OmegaConf` as replacement for `anyconf` in this new configloader.
3. Turn off all built-in resolvers in `OmegaConf` in `ConfigLoader`. In the first iteration we want to keep these advanced features turned off https://omegaconf.readthedocs.io/en/latest/custom_resolvers.html#clear-resolver. It's easier to introduce them step by step, but removing them would be a breaking change. In https://github.com/kedro-org/kedro/issues/1909 we will introduce environment variables for credentials.

MerelTheisenQB:
> @MerelTheisenQB Do you know if `OmegaConf` implementation will keep the current configs merging behaviour? The challenge I've seen in the implementations is that top-level keys are overridden.

`OmegaConf` has a different merge strategy and will only replace the specified fields. So in your example if you merge the two files you'd end up with: 

```
tracking:
  disable_tracking:
    pipelines:
    - on_exit_notification
  experiment:
    name: name-of-prod-experiment
  params:
    long_params_strategy: tag
```

See the OmegaConf docs for an example snippet you can run: https://omegaconf.readthedocs.io/en/2.2_branch/usage.html#merging-configurations

szczeles:
@MerelTheisenQB Fantastic! Thanks a lot for the confirmation

Issue #1657, opened 2022-06-30
Assess `OmegaConf` as replacement for `anyconfig`


MerelTheisenQB:
## Introduction

As part of the work on improving configuration in Kedro we should assess alternatives for `anyconfig`. `anyconfig` was initially chosen because it supports reading configuration of lots of different types. In practice, most users use `yml` files and so we might be able to use an alternative library that offers better functionality for `yml` configuration. 

## Task

Assess the following alternatives ordered by preference:

1. [OmegaConf](https://github.com/omry/omegaconf)
2. [Hydra](https://hydra.cc/) (specifically the compose API)
3. [DynaConf](https://github.com/rochacbruno/dynaconf) this is the least preferred, but already used in Kedro.

datajoely:
@DavidRoschewitzQB it would be great to get your thoughts on Hydra here

noklam:
`OmegaConf` is part of the `Hydra` dependencies, is the compose feature coming from `OmegaConf` itself or hydra actually add more on it?

Found a [OmegaConf Deck](https://docs.google.com/presentation/d/e/2PACX-1vT_UIV7hCnquIbLUm4NnkUpXvPEh33IKiUEvPRF850WKA8opOlZOszjKdZ3tPmf8u7hGNP6HpqS-NT5/pub?start=false&loop=false&delayms=3000&slide=id.g84632f636b_3_7) here.

datajoely:
If you look at the Hydra compose API docs `OmegaConf` is an import:
https://hydra.cc/docs/1.0/experimental/compose_api/#internaldocs-banner

noklam:
@datajoely Yes they are created by the same author, I think there may be a chance that we only need `OmegaConf` for config loader.

If we want to enable hierarchical config then [overriding nested config in CLI](https://github.com/kedro-org/kedro/issues/750) becomes relevant again.

DavidRoschewitzQB:
Thanks for the tag @datajoely. Happy to share some of our thoughts and considerations. 

There are a few reasons why we decided to use hydra (specifically compose API) in our prototype (some of these might be pros or cons for configuration in kedro):

- The entry point is is a single `.yaml` file, which serves as the "root" of all other configuration. Therefore there is no need to search / loop over various files with a pattern. This combination of files is done explicitly in config. 
[...]

PR #2085, opened 2022-11-30
Add `OmegaConfLoader`

merelcht:
## Description
Added a new config loader class based on `AbstractConfigLoader` that uses `OmegaConf` to load and merge configuration. 

## Background 🔍
- https://github.com/kedro-org/kedro/issues/1657
- https://github.com/kedro-org/kedro/issues/1868

## Development notes 🏗
Some differences between the `OmegaConfLoader` and the other config loader classes:
- It's not recommended to use YAML anchors with `OmegaConf`, because they offer [variable interpolation](https://omegaconf.readthedocs.io/en/2.2_branch/usage.html#variable-interpolation). Hence I didn't copy over the test [`test_ignore_hidden_keys`](https://github.com/kedro-org/kedro/blob/main/tests/config/test_config.py#L179-L195)
- YAML parsing is done a bit differently. I added `ScannerError` with `ParserError` to catch this. 

Other notes:
- While writing a test for checking whether a file has the `.yml` or `.yaml` extension, I found out that `OmegaConf` can also handle `.json` files. So I've added `.json` as an acceptable extension for this loader as well. 
- After refactoring my initial implementation, I made sure not to use any of the helper methods from `config/common.py`. The methods in there are deeply nested and hard to follow. Usage of these methods makes it pretty much impossible for users to customise any configloader implementations. 

## Review notes ✍️
- I added changes incrementally, so you can review this PR by checking out each individual commit. 
- All commits starting from https://github.com/kedro-org/kedro/pull/2085/commits/1c12a87259aca2b72be7277c46116da7c0e0d3de are a refactor to not use any methods from `config/common.py`

### To Discuss 🗣
❓**Which built-in resolvers should be turned off?**

We decided to initially turn off all build-in resolvers. It's easier to introduce them step by step, but removing them would be a breaking change. In https://github.com/kedro-org/kedro/issues/1909 we will introduce environment variables for credentials. You can find a description of all built-in resolvers here: https://omegaconf.readthedocs.io/en/2.2_branch/custom_resolvers.html#built-in-resolvers
I feel like maybe it's okay to keep some resolvers turned on. For example `oc.dict` seems useful and not something we wouldn't want to add. I would like to hear what others think.

❓ **Is `OmegaConfLoader` the best name?**

I first wanted to call this config loader `YAMLConfigLoader` but then I found out that `OmegaConf` can also load `.json` files. Any better name suggestions?

❓ **Will `OmegaConfLoader` replace `ConfigLoader`?**

Follow up on the above question. I'm thinking that in `0.19.0` we should remove the `ConfigLoader` and `TemplatedConfigLoader` and rename this new `OmegaConfLoader` to `ConfigLoader`. If we do that, maybe the name of this new class doesn't matter so much for the `0.18` series.

## Checklist ✅

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/2085"></a>


kedro/config/omegaconf_config.py
```diff
@@ -0,0 +1,262 @@
...
+                f"Invalid YAML file {config_filepath}, unable to read line {line}, "
+                f"position {cursor}."
+            ) from exc
+
+        _check_duplicate_keys(seen_file_to_keys, config_filepath, dict(single_config))
+        seen_file_to_keys[config_filepath] = single_config.keys()
+        aggregate_config.append(single_config)
+
+    if len(aggregate_config) > 1:
+        return dict(OmegaConf.merge(*aggregate_config))
```
noklam, line 249: Note for myself - need to revisit this later.

tests/config/test_omegaconf_config.py
```diff
@@ -0,0 +1,381 @@
...
+        assert params["param1"] == 1
+        assert catalog["trains"]["type"] == "MemoryDataSet"
+        assert catalog["cars"]["type"] == "pandas.CSVDataSet"
+        assert catalog["boats"]["type"] == "MemoryDataSet"
+        assert not catalog["cars"]["save_args"]["index"]
+
+    @use_config_dir
+    def test_load_local_config(self, tmp_path):
+        """Make sure that configs from `local/` override the ones
+        from `base/`"""
```
noklam, line 109: Could you explain how this is testing `local` override from `base`? It almost looks the same as the previous test.

noklam (commented):
Leave 1st round of comments - haven't fully understood the code but since this is a relatively large PR, it's nice to get some early comments.

kedro/config/__init__.py
```diff
@@ -8,6 +8,7 @@
     MissingConfigException,
 )
 from .config import ConfigLoader
+from .omegaconf_config import OmegaConfLoader
```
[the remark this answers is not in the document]
merelcht: You were too fast! 🏎️  I was still working on the description and have just added a comment about exactly this 😄

kedro/config/omegaconf_config.py
```diff
@@ -0,0 +1,262 @@
...
+            str(Path(self.conf_source) / run_env),
+        ]
+
+
+def _get_config_from_patterns_omegaconf(
+    conf_paths: Iterable[str],
+    patterns: Iterable[str] = None,
+) -> Dict[str, Any]:
+    """Recursively scan for configuration files, load and merge them using OmegaConf, and
+    return them in the form of a config dictionary.
```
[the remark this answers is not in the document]
merelcht: I'm not sure if a deprecation warning would make the most sense for that purpose? It's a new method so maybe a bit odd to immediately add a deprecation warning to it. I can just move this method into `config/common.py`.

kedro/config/omegaconf_config.py
```diff
@@ -0,0 +1,262 @@
+"""This module provides ``kedro.config`` with the functionality to load one
+or more configuration files of yaml or json type from specified paths through OmegaConf.
+"""
+import logging
+from pathlib import Path
+from typing import AbstractSet, Any, Dict, Iterable, List, Set  # noqa
+
+from omegaconf import OmegaConf
+from yaml.parser import ParserError
+from yaml.scanner import ScannerError
```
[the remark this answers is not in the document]
merelcht: They each catch slightly different syntax problems in `yaml` files. For example, this throws a `ScannerError`:
```yaml
bad:
config
```

And this throws a `ParserError`:
```yaml
example_iris_data:
              type: pandas.CSVDataSet
          filepath: data/01_raw/iris.csv
```

kedro/config/omegaconf_config.py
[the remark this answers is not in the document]
noklam: You are right, I am thinking of something more lightweight. Something like this, or maybe just one sentence in the docstring.

https://www.attrs.org/en/21.1.0/backward-compatibility.html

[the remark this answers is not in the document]
merelcht: That's a nice example! Let me explore some good options. I think in general we might want to clarify to users that any "private" methods aren't meant to be used directly.

kedro/config/omegaconf_config.py
```diff
@@ -0,0 +1,262 @@
...
+            ) from exc
+
+        _check_duplicate_keys(seen_file_to_keys, config_filepath, dict(single_config))
+        seen_file_to_keys[config_filepath] = single_config.keys()
+        aggregate_config.append(single_config)
+
+    if len(aggregate_config) > 1:
+        return dict(OmegaConf.merge(*aggregate_config))
+
+    return config
```
noklam, line 251: This should be able to  handle case when len=1 ? Then the line 251 `return config` can be removed

noklam: More importantly, is `OmegaConf.merge` a syntatic suger of merging a list of dictionary? Is it possible to remove `aggregated_config` altogether and just merging the dictionary at each iteration?

merelcht: `OmegaConf.merge` throws an assertion error if the list is empty: https://github.com/omry/omegaconf/blob/master/omegaconf/omegaconf.py#L267. As far as I understand it's not just syntactic sugar and does the merging of configs in a considerate way.

merelcht: `OmegaConf.merge` throws an assertion error if the list is empty: https://github.com/omry/omegaconf/blob/master/omegaconf/omegaconf.py#L267. As far as I understand it's not just syntactic sugar and does the merging of configs in a considerate way.

kedro/config/omegaconf_config.py
```diff
@@ -0,0 +1,262 @@
...
+        new_conf = _load_configs_omegaconf(config_filepaths=config_filepaths)
+
+        aggregate_config.append(new_conf)
+        processed_files |= set(config_filepaths)
+
+    if not processed_files:
+        raise MissingConfigException(
+            f"No files of YAML or JSON format found in {conf_paths} matching the glob "
+            f"pattern(s): {patterns}"
+        )
```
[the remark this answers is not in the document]
merelcht: Almost, but we use `processed_files` to check if stuff has been processed before so that we don't load it again and then we emit a log message: https://github.com/kedro-org/kedro/blob/main/kedro/config/common.py#L182-L187

tests/config/test_omegaconf_config.py
merelcht: I've changed this so the first two tests focus on verifying the `.get()` and `[]` access and the third test actually verifies that `local` config overrides `base`.

kedro/config/omegaconf_config.py
```diff
@@ -0,0 +1,252 @@
...
+                f"No files of YAML or JSON format found in {base_path} or {env_path} matching"
+                f" the glob pattern(s): {[*self.config_patterns[key]]}"
+            )
+        return config
+
+    def __repr__(self):  # pragma: no cover
+        return (
+            f"OmegaConfLoader(conf_source={self.conf_source}, env={self.env}, "
+            f"config_patterns={self.config_patterns})"
+        )
```
SajidAlamQB, line 174: Slightly confused about what these strings are for.

kedro/config/omegaconf_config.py
```diff
@@ -0,0 +1,252 @@
...
+        except (ParserError, ScannerError) as exc:
+            assert exc.problem_mark is not None
+            line = exc.problem_mark.line
+            cursor = exc.problem_mark.column
+            raise ParserError(
+                f"Invalid YAML or JSON file {config_filepath}, unable to read line {line}, "
+                f"position {cursor}."
+            ) from exc
+
+        _check_duplicate_keys(seen_file_to_keys, config_filepath, dict(single_config))
```
SajidAlamQB, line 227: Do we still want to use this method from `common.py` if we will be reworking it?

SajidAlamQB (commented):
This looks great thank you for this awesome work! Just had a few comments.

kedro/config/omegaconf_config.py
merelcht: The `__repr__` method in Python is used to represent a Class object as a string (https://www.educative.io/answers/what-is-the-repr-method-in-python). It's useful for debugging and in the interactive ipython mode. 

I added the `__repr__` method for `ConfigLoader` and `TemplatedConfigLoader` in https://github.com/kedro-org/kedro/pull/1966 after @noklam pointed out they're not printed properly when using in interactive mode (see https://github.com/kedro-org/kedro/pull/1966#discussion_r1004477060). The one here in `OmegaConfLoader` is basically the same as for the other config loader classes.

kedro/config/omegaconf_config.py
merelcht: Good question! The reason I used this method from `common.py` is because we do want this behaviour in the `OmegaConfLoader` and it's exactly the same as for the other config loaders. The other methods in `common.py` that I didn't end up using were very indirect and nested calls, so I refactored them for this config loader. In `0.19.0` I envision we'll get rid of `common.py` so we could move `_check_duplicate_keys` inside `omegaconf_config.py`. 

I could've refactored all methods in `common.py` but that felt like unnecessary work since ideally we wouldn't have this file full of helper methods. 

Do you think this makes sense or would you prefer there's no calls to `common.py` methods at all in this class?

SajidAlamQB: This makes sense, thank you. I agree we should move the `_check_duplicate_keys` method in `0.19.0` to `omegaconf_config.py`.

kedro/config/omegaconf_config.py
```diff
@@ -0,0 +1,258 @@
...
+            _check_duplicate_keys(
+                seen_file_to_keys, config_filepath, dict(single_config)
+            )
+            seen_file_to_keys[config_filepath] = single_config.keys()
+            aggregate_config.append(single_config)
+
+        if len(aggregate_config) > 1:
+            return dict(OmegaConf.merge(*aggregate_config))
+
+        return config
```
noklam, line 242: Can we remove the special handling of `single_config` when `len(aggregate_config) == 1`? It should handle with len == 1 and thus no need to have an extra variable `config` and
```suggestion
        single_config = {}
        aggregate_config = []
        seen_file_to_keys = {}  # type: Dict[Path, AbstractSet[str]]

        for config_filepath in config_files_filtered:
            try:
                single_config = OmegaConf.load(config_filepath)
            except (ParserError, ScannerError) as exc:
                assert exc.problem_mark is not None
                line = exc.problem_mark.line
                cursor = exc.problem_mark.column
                raise ParserError(
                    f"Invalid YAML or JSON file {config_filepath}, unable to read line {line}, "
                    f"position {cursor}."
                ) from exc

            _check_duplicate_keys(
                seen_file_to_keys, config_filepath, dict(single_config)
            )
            seen_file_to_keys[config_filepath] = single_config.keys()
            aggregate_config.append(single_config)


        return dict(OmegaConf.merge(*aggregate_config))

```

tests/config/test_omegaconf_config.py
```diff
@@ -0,0 +1,388 @@
...
+
+    @use_config_dir
+    def test_load_core_config_get_syntax(self, tmp_path):
+        """Make sure core config can be fetched with .get()"""
+        conf = OmegaConfLoader(str(tmp_path))
+        params = conf.get("parameters")
+        catalog = conf.get("catalog")
+
+        assert params["param1"] == 1
+        assert catalog["trains"]["type"] == "MemoryDataSet"
```
noklam, line 112: 👍🏼

noklam (commented):
It is cleaner now with some of the methods moving into the class. I just re-iterate on one of the comment, generally look great!

kedro/config/omegaconf_config.py
```diff
@@ -0,0 +1,258 @@
...
+
+        Returns:
+            Dict[str, Any]:  A Python dictionary with the combined
+               configuration from all configuration files.
+        """
+
+        # Allow bypassing of loading config from patterns if a key and value have been set
+        # explicitly on the ``OmegaConfLoader`` instance.
+        if key in self:
+            return super().__getitem__(key)
```
AhdraMeraliQB, line 135: Why aren't we doing this check on the other config loader implementations?

merelcht: The only reason is that I just added it today and didn't think about the other implementations yet 😉 I'll add it.

noklam (approved):
Thank you for explaining! Nice work🌟

kedro/config/omegaconf_config.py
```diff
@@ -0,0 +1,258 @@
...
+        config = {}
+        aggregate_config = []
+        seen_file_to_keys = {}  # type: Dict[Path, AbstractSet[str]]
+
+        for config_filepath in config_files_filtered:
+            try:
+                single_config = OmegaConf.load(config_filepath)
+                config = single_config
+            except (ParserError, ScannerError) as exc:
+                assert exc.problem_mark is not None
```
idanov, line 225: What is this supposed to do? Could we get rid of the assert here?

kedro/config/omegaconf_config.py
```diff
@@ -0,0 +1,258 @@
...
+                line = exc.problem_mark.line
+                cursor = exc.problem_mark.column
+                raise ParserError(
+                    f"Invalid YAML or JSON file {config_filepath}, unable to read line {line}, "
+                    f"position {cursor}."
+                ) from exc
+
+            _check_duplicate_keys(
+                seen_file_to_keys, config_filepath, dict(single_config)
+            )
```
idanov, line 235: Could we do this check only at the end of the loop, so we can give error message for all duplicates at once?

In fact, maybe we could keep only one variable, say `config_per_file`, mapping `filepath -> config`, and then check the duplicates in a separate loop, which will produce all errors at once.

After all `aggregate_config` is just `config_per_file.values()` and `seen_file_to_keys = {file: set(config.keys()) for file, config in config_per_file.items()}`. Updating two data structures in the same loop is harder to read.

kedro/config/omegaconf_config.py
```diff
@@ -0,0 +1,258 @@
...
+        """
+
+        # Allow bypassing of loading config from patterns if a key and value have been set
+        # explicitly on the ``OmegaConfLoader`` instance.
+        if key in self:
+            return super().__getitem__(key)
+
+        # In the first iteration of the OmegaConfLoader we'll keep the resolver turned-off.
+        # It's easier to introduce them step by step, but removing them would be a breaking change.
+        self._clear_omegaconf_resolvers()
```
idanov, line 139: I wonder whether we could call this on `OmegaConfLoader.__init__` once, instead on each request for config. WDYT the implications of this could be?

kedro/config/omegaconf_config.py
```diff
@@ -0,0 +1,258 @@
...
+
+    """
+
+    def __init__(
+        self,
+        conf_source: str,
+        env: str = None,
+        runtime_params: Dict[str, Any] = None,
+        config_patterns: Dict[str, List[str]] = None,
+        *,
```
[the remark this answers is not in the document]
merelcht: No they're not, but we did add it to `ConfigLoader` and `TemplatedConfigLoader`. Do you think it shouldn't be an argument at all in `__init__`?

kedro/config/omegaconf_config.py
merelcht: As far as I understand this is to verify whether we can get `exc.problem_mark.line` and `exc.problem_mark.column` for nicely formatting the error message. Perhaps the assert isn't necessary here.

kedro/config/omegaconf_config.py
merelcht: So instead you would load all the files and then compare all of them against each other to find duplicates? Wouldn't that me a lot more inefficient? I guess the benefit is that you would immediately know all duplicates, but the downside is that if you have a lot of config everything needs to be loaded first and then checked before you know something is wrong.

kedro/config/omegaconf_config.py
merelcht: Yes actually I think that makes more sense. Otherwise you could get a situation where if the `__getitem__` function has been called users could technically use the resolvers. Seems unlikely, but I think it's better to move it up to `__init__`.

kedro/config/omegaconf_config.py
merelcht: Ahh I think this might have been put there for the `mypy` linter:

```
kedro/config/omegaconf_config.py:224: error: Item "None" of "Optional[Mark]" has no attribute "line"
kedro/config/omegaconf_config.py:225: error: Item "None" of "Optional[Mark]" has no attribute "column"
Found 2 errors in 1 file (checked 265 source files)
```

I'll just add an ignore.

kedro/config/omegaconf_config.py
merelcht: I've rewritten this to now check duplicates between all config files in a directory + subdirectories.

idanov (approved):

PR #2087, opened 2022-11-30
Replace comma in default node names with semicolon

### LINKED ISSUE #2013: Change autogeneration of node names to not contain commas
## Description
Change how node names are generated to not contain commas. Currently we stringify a node's definition and set that as the node name if no name has been set explicitly: https://github.com/kedro-org/kedro/blob/main/kedro/pipeline/node.py#L239

Change is so that the generated name will contain the function name and output, where the outputs are separated by `_` or `-` instead of `,`.

So `two_inputs([A0,B0]) -> [C1,C2]` would become `two_inputs -> [C1-C2]`

## Context
https://github.com/kedro-org/kedro/issues/1828

jmholzer:
Resolves #2013
## Checklist


ankatiyar (approved):
Minor comment - The error that is shown when you try to name a node with a string containing commas is `ValueError: 'node, name' is not a valid node name. It must contain only letters, digits, hyphens, underscores and/or fullstops.`. I think this could also possibly be modified to add semicolons for consistency. 
Otherwise, looks good to me! 🎉

jmholzer:
> Minor comment - The error that is shown when you try to name a node with a string containing commas is `ValueError: 'node, name' is not a valid node name. It must contain only letters, digits, hyphens, underscores and/or fullstops.`. I think this could also possibly be modified to add semicolons for consistency. Otherwise, looks good to me! 🎉

Thanks for the review! I think we might want to leave semicolons from this list, since if Kedro creates default node names with semicolons, that is an implicit statement that these names are correct. Currently, we do it differently; Kedro says that node names with commas are disallowed even though it also creates default node names with commas in them but I feel this is a bug (/ inconsistency) and not a feature.

jmholzer:
> This is a breaking change and should go on the `develop` branch. The ticket is tagged with the `0.19.0` version 🙂

Done 🙂

merelcht (approved):
LGTM! Don't forget to update the release notes to include this change.

Issue #1908, opened 2022-10-06
Add the ability to change `conf_source` in `kedro run` CLI command


merelcht:
## Description
Add the ability to change `conf_source` from CLI as part of the `kedro run` command

* `session.run()`
* `kedro run --conf-source=xxxx` (temp name)

## Context
Currently, it's not possible to change the configuration location when running `kedro run` on the CLI. You can only change it in [`settings.py`](https://github.com/kedro-org/kedro/blob/main/kedro/templates/project/%7B%7B%20cookiecutter.repo_name%20%7D%7D/src/%7B%7B%20cookiecutter.python_package%20%7D%7D/settings.py#L25). This is problematic especially for packaged Kedro project. So we should allow users to specify where conf is from the CLI.

noklam:
Agree with this in principle, but I have a couple of questions.

- What should be configurable at runtime in `settings.py` other than `CONF_SOURCE`? Do we just want to add this one configuration in the CLI?
  - Should this works for `KedroSession.run` or `__main__` too?

takikadiri:
I confirm that we have currently this problem. When using a packaged kedro project, the conf folder can be deployed in a location different from the CWD. _for example : in a kubernetes volume so the conf can be accessible by all the replicate Pods._  
Overriding the CLI by adding a conf-source entry does not work because the `settings.py` [is called before using the CLI ](https://github.com/kedro-org/kedro/blob/main/kedro/templates/project/%7B%7B%20cookiecutter.repo_name%20%7D%7D/src/%7B%7B%20cookiecutter.python_package%20%7D%7D/__main__.py#L41)
The current workaround that we did is using an environment variable in `settings.py`. That way it's resoved at runtime.  

Moreover in a packaged and deployed kedro project pattern, we have another observation regarding conf locations.  
The [third factor of the twelve-factor app](https://12factor.net/config) advocate that the conf belong to the environnement, not the application. So when we deploy a kedro app in a given environnement, the kedro app is no longer meant to resolve the conf environnemnt with `--env env`,  because there is just one conf anyway in that environnement.  We'll love having the possibility to give just a `conf-path` to the kedro app intead of `conf-source` when deploying kedro projet in a cloud native (and 12 factor) way.  
To sum up we want our environnements to run the kedro app this way `python -m package_name --conf-path /path/to/kedro/conf`  the conf folder contains directly globals.yml, catalog.yml,... without env folder.
[...]

PR #2117, opened 2022-12-12
Add configuration source option to `kedro run` CLI command

jmholzer:
Resolves [#1908](https://github.com/kedro-org/kedro/issues/1908)

## Development notes
I added the `--conf-source` option to the `run` endpoint of the Kedro CLI.

This is achieved by adding `_conf_source` as an attribute to `KedroSession`, which can optionally be set when an object of this class is instantiated. If it is not passed at instantiation, its value is determined is before (it defaults to `.conf/`).

I believe using an attribute for `_conf_source` is the right thing to do, since `_get_config_loader` is called in two separate methods in `KedroSession`. Passing `_conf_source` in as an argument to `_get_config_loader` would complicate the class.

I would particularly welcome comments on my approach to testing, I have purposefully kept these as simple as possible.

## Checklist


kedro/framework/cli/project.py
```diff
@@ -60,6 +60,7 @@
 param_group.param1:value1."""
 INPUT_FILE_HELP = """Name of the requirements file to compile."""
 OUTPUT_FILE_HELP = """Name of the file where compiled requirements should be stored."""
+CONF_SOURCE_HELP = """Path of a directory where project configuration is stored"""
```
idanov, line 63:
```suggestion
CONF_SOURCE_HELP = """Path of a directory where project configuration is stored."""
```

merelcht (commented):
I'll have a more thorough review in a bit, but it looks good on first glance 😃 One more thing you might want to try if you haven't yet, is whether this works with a packaged project. https://kedro.readthedocs.io/en/stable/tutorial/package_a_project.html#package-your-project

kedro/framework/session/session.py
```diff
@@ -112,12 +116,17 @@ def __init__(
...
+            self._project_path / settings.CONF_SOURCE
+        )
+
     @classmethod
     def create(  # pylint: disable=too-many-arguments
         cls,
         package_name: str = None,
         project_path: Union[Path, str] = None,
         save_on_close: bool = True,
+        conf_source: Optional[str] = None,
```
idanov, line 129:
```suggestion
        conf_source: Optional[str] = None,
        save_on_close: bool = True,
```

Also I think we might need to add it last in order to be strictly non-breaking change, because someone might actually rely on the ordering of the arguments (very unlikely, but still)... WDYT?

kedro/framework/session/session.py
```diff
@@ -93,16 +94,19 @@ class KedroSession:
...
         session_id: str,
         package_name: str = None,
         project_path: Union[Path, str] = None,
         save_on_close: bool = False,
+        conf_source: Optional[str] = None,
     ):
         self._project_path = Path(project_path or Path.cwd()).resolve()
         self.session_id = session_id
         self.save_on_close = save_on_close
+        self._conf_source = conf_source
```
idanov, line 109:
```suggestion
```
No point of this if you are reassigning it again later.

idanov (commented):
Nice!

kedro/framework/cli/project.py
jmholzer: Thanks for catching this, added in [80ee277](https://github.com/kedro-org/kedro/pull/2117/commits/80ee27733552924408544c57e676e94e9d4fd1f5).

kedro/framework/session/session.py
jmholzer: I think you're right, I added it at the end of the function signature in [7f5281c](https://github.com/kedro-org/kedro/pull/2117/commits/7f5281c6fd411ac48f76c8537db7c0a947cfe43a).

kedro/framework/session/session.py
jmholzer: Yes! Thanks for the catch, this was a leftover from an earlier implementation that I missed. Fixed in [fde701c](https://github.com/kedro-org/kedro/pull/2117/commits/fde701c1ed0cffa5ecf5f340291ecb1267470caa).

jmholzer:
> I'll have a more thorough review in a bit, but it looks good on first glance 😃 One more thing you might want to try if you haven't yet, is whether this works with a packaged project. https://kedro.readthedocs.io/en/stable/tutorial/package_a_project.html#package-your-project

Awesome 😃 thanks for the quick feedback. I have done manual testing for a packaged project and can confirm that it works, though please feel free to verify.

merelcht (approved):
I tried it in packaged mode and it works well! Great work ⭐ 

I don't think we need a new issue for the docs, but if you prefer to do them in a separate PR that's fine 🙂

jmholzer:
> I tried it in packaged mode and it works well! Great work ⭐
> 
> I don't think we need a new issue for the docs, but if you prefer to do them in a separate PR that's fine 🙂

Awesome, thanks Merel. I'll open a separate PR for the docs, as I think it will make it easier to align with @stichbury, who I think is a stakeholder here.

RELEASE.md
```diff
@@ -11,6 +11,7 @@
 # Upcoming Release 0.18.5
 
 ## Major features and improvements
+* Add the `--conf-source` option to `kedro run`; allowing users to specify a source for project configuration for the run.
```
AhdraMeraliQB, line 14:
```suggestion
* Add the `--conf-source` option to `kedro run`, allowing users to specify a source for project configuration for the run.
```
More a nitpick than anything else - iirc it'd be more appropriate to use a comma instead of a semicolon here; a good rule of thumb is seeing if replacing the semicolon with a comma and conjunction (and, but, or, e.t.c) would still hold.

AhdraMeraliQB (approved):
Nice work!

RELEASE.md
jmholzer: Changed in [bccb13b](https://github.com/kedro-org/kedro/pull/2117/commits/bccb13baaa0994fcf863dc698dbb98a268756262).

PR #2120, opened 2022-12-13
Allow key, values in `--params` to be separated by equal sign

### LINKED ISSUE #1907: Provide OmegaConf syntax as option for `--params`
## Description
Provide OmegaConf syntax as second option for CLI `--params`. Still have Kedro syntax as default + deprecation warning and then replace in 0.19.0. 

## Context
Currently, Kedro has its own custom syntax and logic for passing parameters in the CLI. We can simplify this by leveraging `OmegaConf`, which we're also going to use in the configloaders instead of `anyconf`: https://github.com/kedro-org/kedro/issues/1868

## Possible Implementation
Option 1:
image (17)

Option 2:
Add a new boolean flag to support OmegaConf syntax: `--new-style`
`kedro run --new-style --params ...`

ankatiyar:

## Description

Resolves #1907. Allows OmegaConf syntax `key=value` to be used with `--params` while doing `kedro run`. Original Kedro syntax `key:value` also still works.

## Development notes

* In `kedro/framework/cli/utils.py` -
    - Modified `_split_params()` : still has a bit of original parsing to catch errors but essentially splits the parameter string on commas and adds the key-value pairs to `dot_list` to pass to `OmegaConf.from_dotlist()`
    - Removed `_update_value_nested_dict()` : `OmegaConf.from_dotlist()` can handle nested parameters so this is no longer needed.
* In `tests/framework/cli/test_cli.py` -
    - Removed test for `_update_value_nested_dict()`
    - Modified `test_run_extra_params` to account for Omegaconf's handling of parameters + added test cases and `test_bad_extra_params` 
* Manually tested that the following commands would work, (colon, equal sign or a combination) -
    - `kedro run --params 'test_size=0.3, random_state=3'`
    - `kedro run --params 'test_size:0.3, random_state:3'`
    - `kedro run --params 'test_size=0.3, random_state:3'`

## Checklist


kedro/framework/cli/project.py
```diff
@@ -53,8 +53,8 @@
 PIPELINE_ARG_HELP = """Name of the registered pipeline to run.
 If not set, the '__default__' pipeline is run."""
 PARAMS_ARG_HELP = """Specify extra parameters that you want to pass
-to the context initializer. Items must be separated by comma, keys - by colon,
-example: param1:value1,param2:value2. Each parameter is split by the first comma,
+to the context initializer. Items must be separated by comma, keys - by colon or equal sign,
```
AhdraMeraliQB, line 56: An incredibly minor note here - we want to use UK English for this one: https://kedro.readthedocs.io/en/stable/contribution/documentation_contributor_guidelines.html
```suggestion
to the context initialiser. Items must be separated by comma, keys - by colon or equals sign,
```

RELEASE.md
```diff
@@ -13,7 +13,7 @@
 ## Major features and improvements
 
 ## Bug fixes and other changes
-
+* Added `omegaconf` syntax as option for `--params`. Keys and values can now be separated by colons or equal signs.
```
AhdraMeraliQB, line 16:
```suggestion
* Added `omegaconf` syntax as option for `--params`. Keys and values can now be separated by colons or equals signs.
```
Same note 😅

merelcht (commented):
Where does this use `omegaconf`? It now looks like it's still using the Kedro parsing of parameters. The idea is that we can remove that parsing completely in `0.19.0` and just keep the `omegaconf` way of reading CLI parameters.

tests/framework/cli/test_cli.py
```diff
@@ -643,13 +629,13 @@ def test_run_with_params_in_config(
                 {"foo": "bar", "baz": "https://example.com"},
             ),
             ("foo:bar,baz:fizz buzz", {"foo": "bar", "baz": "fizz buzz"}),
-            ("foo:bar, foo : fizz buzz  ", {"foo": "fizz buzz"}),
```
ankatiyar, line 646: OmegaConf handles this case a bit differently. Instead of overwriting the value for the key, it keeps both. 
`{foo:"bar", foo:"fizz buzz"}`

merelcht: It would be breaking functionality if that behaviour suddenly changes. Can we make it such that the second provided argument overrides the first?

tests/framework/cli/test_cli.py
```diff
@@ -643,13 +629,13 @@ def test_run_with_params_in_config(
...
-            ("foo.nested:123.45", {"foo": {"nested": 123.45}}),
+            ("foo.nested=123.45", {"foo": {"nested": 123.45}}),
             (
                 "foo.nested_1.double_nest:123.45,foo.nested_2:1a",
                 {"foo": {"nested_1": {"double_nest": 123.45}, "nested_2": "1a"}},
             ),
+            ("foo=bar", {"foo": "bar"}),
         ],
     )
     def test_run_extra_params(
```
merelcht, line 641: I think it would be clearer if we just have this test twice, once with `:` and once with `=` so we can clearly see both syntaxes work for all cases.

merelcht (commented):
I think we're getting closer to the solution, but we need to be very careful not to make any breaking changes to the existing functionality. If it's easier to maintain the current behaviour following Antony's description in the screenshot here: https://github.com/kedro-org/kedro/issues/1907 I would just implement it that way.

tests/framework/cli/test_cli.py
ankatiyar: The test with `:` is already there. I've moved the test parameters together to make it more readable.

tests/framework/cli/test_cli.py
ankatiyar: Done in bdf35cf746e74741bf1706db3136f261181549a6

merelcht (approved):
This is looking so good now! 🎉 Much simpler code than before!

It would be good to update the docs as well so users are aware they can use the `=` syntax. I'm happy for that to go in a separate PR.

AhdraMeraliQB (approved):
Great work! 🌟 Especially well done on the detailed development notes ❤️

kedro/framework/cli/utils.py
```diff
@@ -401,68 +402,33 @@ def _reformat_load_versions(  # pylint: disable=unused-argument
...
     if isinstance(value, dict):
         return value
-    result = {}
+    dot_list = []
     for item in split_string(ctx, param, value):
-        item = item.split(":", 1)
-        if len(item) != 2:
+        equals_idx = item.find("=")
+        colon_idx = item.find(":")
+        if equals_idx != -1 and colon_idx != -1 and equals_idx < colon_idx:
```
ankatiyar, line 412: While updating the docs I realised that the original implementation would not work for cases where both colons and equal signs are used. For example : `--params url=http://example.com`. This if statement checks for this specific case. I've also added these examples to the tests.

merelcht: It might be useful to add a comment in the code as well to explain what this check is for.

PR #2139, opened 2022-12-19
Add 'safe splitting' for node names in options for `kedro run`

### LINKED ISSUE #2012: Fix comma splitting for node re-run suggestion
## Description
Running kedro with the command `kedro run --from-nodes "two_inputs([A0,B0]) -> [C1]"` will cause an error: `Pipeline does not contain nodes named ['B0]) -> [C1]', 'two_inputs([A0'].`

Kedro has incorrectly split the name of the intended Node into the names of two Nodes that do not exist.

This bug prevents the user from re-running kedro from any Node with a comma in its name. This is the case for any Node that is not explicitly given a name and has more than one input or output.

### Steps to Reproduce
1. Create a kedro project
2. Attempt to run the pipeline from a node with a comma in its name (`kedro run --from-nodes 'My node, a good node'`)
3. Receive the error `Pipeline does not contain nodes named ['a good node', 'My node']`.

## Context
https://github.com/kedro-org/kedro/issues/1828

## Implementation
@jmholzer idea of a fine-state machine csv-parsing type solution to do the string splitting rather than just blindly splitting on `,`. This would basically mean that `,` inside `()` would not split the string.

jmholzer:
## Description
Resolves #2012

Commas surrounded by an even number of `"` and `'` characters will no longer split the arguments to `kedro run` options which take a list of nodes as inputs:

- `--from-nodes`
- `--to-nodes`

For instance example, `kedro run --from-nodes "two_inputs([A0,B0]) -> [C1]"` will now run Kedro from the `node with name `two_inputs([A0,B0]) -> [C1]`.

## Development notes
I used regex to implement this, rather than manually coding a finite-state machine.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/2139"></a>


kedro/framework/cli/utils.py
```diff
@@ -294,6 +294,26 @@ def split_string(ctx, param, value):  # pylint: disable=unused-argument
...
+    """Split string by comma, ignoring commas enclosed by an even number
+    of \" and \' characters.
+
+    Args:
+        to_split: the string to split safely
+
+    Returns:
+        A list containing the result of safe-splitting the string.
+    """
+    # Only split on commas with an even number of ' and " ahead
```
merelcht, line 308: In the PR description you said: "Commas surrounded by an even number of " and ' characters will no longer split the arguments to kedro run options", and that's also what the regex does, so shouldn't this comment read "Don't split on commas with an even number of ' and " ahead"?

merelcht (approved):
LGTM! Nice one on using regex for this 👍 

Just left two minor comments about the release notes and doc string.

ankatiyar:
I checked this PR out and tested it with a Kedro project. What I noticed is that it still doesn't work with `kedro run --from-nodes "do_nothing[X,Y] -> [Z]"`. The node name/s have to be in another set of quotes for it to work -> `kedro run --from-nodes "'do_nothing[X,Y] -> [Z]'"`. I think the issue is that the outer sets of quotes don't get passed on to `split_node_names` so it still splits the string on the comma and throws the same error.

jmholzer:
> I checked this PR out and tested it with a Kedro project. What I noticed is that it still doesn't work with `kedro run --from-nodes "do_nothing[X,Y] -> [Z]"`. The node name/s have to be in another set of quotes for it to work -> `kedro run --from-nodes "'do_nothing[X,Y] -> [Z]'"`. I think the issue is that the outer sets of quotes don't get passed on to `split_node_names` so it still splits the string on the comma and throws the same error.

Thanks for trying out this PR with some manual tests, I really appreciate it. I'm aware of this, unfortunately this is because on Unix systems, shells will consume any enclosing quotations before passing them as arguments for a process (https://stackoverflow.com/questions/667540/full-command-line-as-it-was-typed). There is nothing really we can do to fix this, that information is simply lost by the time it reaches the `kedro` process. My argument would be that at least this gives our users some way to stop the splitting, though I agree with you that it is a problem.

I also hadn't considered that it actually still makes our re-run message inaccurate until now. On a node failure, the suggestion would still be:
`You can resume the pipeline run by adding the following argument to your previous command: --from-nodes "do_nothing[X,Y] -> [Z]"`
which is still inaccurate.

I see two options:
1. Go ahead with this approach, and fix the re-run message for the edge-case that a default-name node is used.
2. Change the criteria for splitting, so that instead of using enclosure by `'` or `"` as our 'no-split' criterion, we could use the pattern `<function_name>([<node_name>,<node_name>])` instead. On release of Kedro 0.19.0 we would then remove this logic, as from then we won't have any node names with commas in them any more.

I prefer option 2, what do you think @merelcht and @ankatiyar?

RELEASE.md
```diff
@@ -17,6 +17,8 @@
 
 ## Bug fixes and other changes
 * Fix bug where `micropkg` manifest section in `pyproject.toml` isn't recognised as allowed configuration.
+* Commas surrounded by an even number of `"` and `'` characters will no longer split the arguments to `kedro run` options which take a list of nodes as inputs (`--from-nodes` and `--to-nodes`).
+* Added `omegaconf` syntax as option for `--params`. Keys and values can now be separated by colons or equals signs.
```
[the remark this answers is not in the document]
jmholzer: Changed in [1d034be](https://github.com/kedro-org/kedro/pull/2139/commits/1d034be78d2c8a384bc3267bdc18d2af88a2895a), thanks for the heads up, I didn't see this 🙂.

ankatiyar:
> I see two solutions:
> 
> 1. Go ahead with this approach, and fix the re-run message for the edge-case that a node with a default name is used.
> 2. Change the criteria for splitting, so that instead of using enclosure by `'` or `"` as our 'no-split' criterion, we could use the pattern `<function_name>([<node_name>,<node_name>])` instead. On release of Kedro 0.19.0 we would then remove this logic, as from then we won't have any node names with commas in them any more.
> 
> I prefer option 2, what do you think @merelcht and @ankatiyar?

Option 2 sounds good but would that be easily doable? Since nodes would only follow that pattern if they were auto-generated names and even then the number of inputs and outputs vary. 
I agree that this still gives the users the option to run from/to nodes with auto-generated names which would be difficult otherwise - option 1 also sounds good, with the updated re-run message and maybe a note in the docs?

jmholzer:
I discussed this matter more with @ankatiyar, who pointed out that square bracket chars are not allowed in user-defined node names. Therefore, the right approach in this case is to ignore commas enclosed in square brackets, the logic is:

1. Square brackets are present in the option argument if and only if a default node name is used
2. therefore any occurrence of commas in square brackets will always be well formed
3. therefore there is never a risk that we ignore a comma incorrectly.

tests/framework/cli/test_cli.py
```diff
@@ -719,6 +719,51 @@ def test_fail_reformat_load_versions(self, fake_project_cli, fake_metadata):
         )
         assert expected_output in result.output
 
+    @mark.parametrize(
```
idanov, line 722: If `split_node_names` is an importable function, does it make more sense to not have a CLI test, but unit test the function instead?

RELEASE.md
```diff
@@ -17,6 +17,7 @@
 
 ## Bug fixes and other changes
 * Fix bug where `micropkg` manifest section in `pyproject.toml` isn't recognised as allowed configuration.
+* Commas surrounded by an even number of `"` and `'` characters will no longer split the arguments to `kedro run` options which take a list of nodes as inputs (`--from-nodes` and `--to-nodes`).
```
idanov, line 20:
```suggestion
* Commas surrounded by an even number of `[` and `]` characters will no longer split the arguments to `kedro run` options which take a list of nodes as inputs (`--from-nodes` and `--to-nodes`).
```

merelcht (approved):
This approach makes sense to me! 👍 I also manually tested it and it is working well 🙂

tests/framework/cli/test_cli.py
merelcht: This would be nice, but we don't really have separate tests for any of the other "util" cli functions. All of that functionality is tested through CLI tests. So I think for consistency it makes sense to test this functionality in a CLI test as well.

RELEASE.md
jmholzer: Thanks for the reminder! Fixed in [a74a105](https://github.com/kedro-org/kedro/pull/2139/commits/a74a1056abcd22ae9501a12e888ee561c018f1d8).

tests/framework/cli/test_cli.py
jmholzer: I think this is a really good question. As @merelcht says, I think the current set up fits our strategy for testing `utils`, and I actually personally think the current strategy is the right way to go:

1. It tests two parts of the code base at once, which reduces development and testing burden.
2. It can can pick up broken integrations.

Since this is a (maybe rare) instance where writing integration tests is easier / faster than writing unit tests, I think we should stick with it.

idanov (approved):

PR #2159, opened 2022-12-23
Fix `after_dataset_saved` hook so that it saves multiple outputs when run in async mode

### LINKED ISSUE #2157: Our `after_dataset_saved` hook does not work with nodes saving more than one datasets
## Description

The code here has a bug:
https://github.com/kedro-org/kedro/blob/08c45b4ada1ec2160d3c2670d8207b4c98330a95/kedro/runner/runner.py#L445-L455
Our `after_dataset_saved` hook has been misbehaving and is always called with the same dataset name and data, regardless of how many other datasets the node will output.

## Context

This bug makes it impossible for users to attach working hooks for `after_dataset_saved` if their nodes have more than one outputs.

jmholzer:
## Description
Resolves #2157

## Development notes
Added a mapping between futures and dataset info in order to access `dataset_name` and `data` when looping over the futures and calling the `after_dataset_saved` hook.

I manually tested these changes with the Iris starter, run in async mode.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/2159"></a>


kedro/runner/runner.py
```diff
@@ -406,6 +406,7 @@ def _run_node_sequential(
     return node
 
 
+# pylint: disable=too-many-locals
```
idanov, line 409: Does it make sense to make use of the fact that `save_futures` is a set, containing exactly the same elements as the keys of `future_dataset_mapping` and only use one local variable, thus having not too many local variables in this function?

idanov (commented):
Awesome, thanks for the quick turnaround on this! Could you add a unit test for this and a line in the `RELEASE.md` about this?

kedro/runner/runner.py
jmholzer: You're right about this, good catch, fixed in [0482020](https://github.com/kedro-org/kedro/pull/2159/commits/0482020abd9e3e01d205ab08cd75ea32bcbff2be).

jmholzer:
Thanks for the quick review @idanov 🙂.

> Could you add a unit test for this
Added in [fb7563b](https://github.com/kedro-org/kedro/pull/2159/commits/fb7563b0b8e73e83043c48b5be4c1cbf0ae1ecbd).

> and a line in the `RELEASE.md`
Added in [a2944e6](https://github.com/kedro-org/kedro/pull/2159/commits/a2944e6b0e1e330a2121d017120ead1fd0e50fef).

tests/framework/session/test_session_extension_hooks.py
```diff
@@ -526,6 +526,20 @@ def wait_and_identity(x: Any):
     return node(wait_and_identity, inputs="ds1", outputs="ds2", name="test-node")
 
 
+@pytest.fixture
+def sample_node_multiple_outputs():
+    def wait_and_identity(x: Any, y: Any):
+        time.sleep(0.1)
+        return (x, y)
```
SajidAlamQB, line 533: Why do we need the `wait_and_identity` function?

SajidAlamQB (approved):
Thanks for this fix it looks great just left one question. 👍

RELEASE.md
```diff
@@ -19,6 +19,7 @@
 * Fix bug where `micropkg` manifest section in `pyproject.toml` isn't recognised as allowed configuration.
```
SajidAlamQB, line 19:
```suggestion
* Fixed bug where `micropkg` manifest section in `pyproject.toml` isn't recognised as allowed configuration.
```

RELEASE.md
```diff
@@ -19,6 +19,7 @@
...
 * Fix bug causing `load_ipython_extension` not to register the `%reload_kedro` line magic when called in a directory that does not contain a Kedro project.
```
SajidAlamQB, line 20:
```suggestion
* Fixed bug causing `load_ipython_extension` not to register the `%reload_kedro` line magic when called in a directory that does not contain a Kedro project.
```

RELEASE.md
```diff
@@ -19,6 +19,7 @@
...
 * Added anyconfig's `ac_context` parameter to `kedro.config.commons` module functions for more flexible `ConfigLoader` customizations.
+* Fix bug causing the `after_dataset_saved` hook only to be called for one output dataset when multiple are saved in a single node and async saving is in use.
```
SajidAlamQB, line 22:
```suggestion
* Fixed bug causing the `after_dataset_saved` hook only to be called for one output dataset when multiple are saved in a single node and async saving is in use.
```

tests/framework/session/test_session_extension_hooks.py
idanov: Do we really need the `time.sleep()` part?

tests/framework/session/test_session_extension_hooks.py
```diff
@@ -562,6 +578,34 @@ def test_after_dataset_load_hook_async(
...
+        after_dataset_saved_mock = mocker.patch.object(
+            mock_session._hook_manager.hook, "after_dataset_saved"
+        )
+        _run_node_async(
+            node=sample_node_multiple_outputs,
+            catalog=memory_catalog,
+            hook_manager=mock_session._hook_manager,
+        )
+
+        after_dataset_saved_mock.assert_has_calls(
```
idanov, line 600: This looks like a nice test! Could you confirm that this test will fail if your change in `runner.py` is not implemented?

tests/framework/session/test_session_extension_hooks.py
```diff
@@ -562,6 +578,34 @@ def test_after_dataset_load_hook_async(
...
 
+    def test_after_dataset_load_hook_async_multiple_outputs(
+        self,
+        mocker,
+        memory_catalog,
+        mock_session,
+        sample_node_multiple_outputs,
+    ):
+        # load mock context to instantiate Hooks
+        mock_session.load_context()
```
idanov, line 589: Could we minimise the dependency tree of the test here by just mocking the `hook_manger` and not mock the whole session? `_run_node_async` doesn't need anything else from the session.

RELEASE.md
jmholzer: Fixed in [78f2f18](https://github.com/kedro-org/kedro/pull/2159/commits/78f2f18df0e1b817af1399a87ed0665a5e6d2ed5).

tests/framework/session/test_session_extension_hooks.py
jmholzer: Yes, I wrote this test on main (where it failed) before adding it to this branch 🙂.

tests/framework/session/test_session_extension_hooks.py
jmholzer: I thought about this. The test will clearly work without this call, I thought to keep it because:

1. `_run_node_async` is designed to work specifically for nodes which block the thread, it makes sense to test it with such a node, even though it is not immediately clear what we gain from doing so.
2. It is a signal to people who want to modify the test / function in the future what the purpose of `_run_node_async` is.
3. The maximum gain from removing it is a decrease of 0.2 seconds in running our test suite.

I don't see a large advantage / disadvantage to either keeping or removing it, so I would be OK with either, though I don't think too much thought should be invested.

tests/framework/session/test_session_extension_hooks.py
jmholzer: Good point, I hadn't thought of this. Fixed in [08c0c9e](https://github.com/kedro-org/kedro/pull/2159/commits/08c0c9e7d50373961ca4be9e9c620a1406c991ee).

The existing test (`test_after_dataset_load_hook_async`) still requires a `mock_session` and can't be modified in the same way.

idanov (approved):
Awesome 🎉

PR #2174, opened 2023-01-04
Make sure `OmegaConfLoader` returns primitive dict type

merelcht:
## Description
When writing docs about the new `OmegaConfLoader` I discovered there were some issues with loading the `logging.yml` file for the spaceflights starter. It turned out that the `OmegaConfLoader` wasn't returning a primitive dict but a omegaconf `DictConfig`. 

## Development notes
Just casting with `dict` like I'd done here: https://github.com/kedro-org/kedro/blob/main/kedro/config/omegaconf_config.py#L236 doesn't work, so I changed it to use the `OmegaConf.to_container()` method (https://omegaconf.readthedocs.io/en/2.3_branch/usage.html#omegaconf-to-container). 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/2174"></a>


jmholzer (approved):
Interesting, I wonder what the `__iter__` on whatever object OmegaConf.merge returns (a `DictConfig` object?) looks like that passing it to `dict` would behave this way, maybe it should at least throw a warning that casting with `dict()` isn't the right way of doing things 🤔.

merelcht:
> Interesting, I wonder what the `__iter__` on whatever object `OmegaConf.merge` returns (a `DictConfig` object?) 

Yes! And if it's a dict with dicts then all dicts will be of type `omegaconf.dictconfig.DictConfig`. 

> looks like that passing it to `dict` would behave this way, maybe `DictConfig` should at least warn the user that `to_container` should be used instead. I might investigate this / open an issue on the OmegaConf repo.

So passing it to `dict` does convert the "outer" `omegaconf.dictconfig.DictConfig` to `dict`, but not any of the nested "inner" `omegaconf.dictconfig.DictConfig`.

PR #2178, opened 2023-01-06
Allow users to pass credentials through environment variables

jmholzer:
## Description

Resolves [#1909](https://github.com/kedro-org/kedro/issues/1909).

## Development notes

`OmegaConf` resolves `DictConfig` objects on read. In the following example (where credentials.yml contains environment variable grammar such as `"key": "${oc.env:TEST_KEY}"`), the `oc.env` resolver would have to be registered, or line 3 will raise an `UnsupportedInterpolationType` exception.

```
conf = OmegaConfLoader(...)
credentials = conf["credentials"]
credentials["test"]
```

However, the scope of the issue is to keep the `oc.env` resolver switched off. The way around this is to use `OmegaConf.resolve`, which resolves `DictConfig` objects in-place when these objects are created, using the `oc.env` resolver. This is done in the `_resolve_environment_variables` method.

Includes four automated tests:

1. Check that environment variables are resolved when the `credentials` is used as a key.
2. Check that environment variables are not resolved when `credentials` is not used as a key.
3. Check that the `oc.env` resolver is not registered after environment variables are read if it was not registered beforehand.
4. Check that the `oc.env` resolver remains registered after environment variables are read if it was registered beforehand.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/2178"></a>


AntonyMilneQB:
Stupid question... why can't we just leave `oc.env` switched on? Something to do with not wanting to allow environment variables except in credentials?

Just a quick comment on naming also: I _think_ @idanov previously suggested that we might drop the `oc.` namespace for this and just have `env` instead.

jmholzer:
> Stupid question... why can't we just leave `oc.env` switched on? Something to do with not wanting to allow environment variables except in credentials?
Exactly, ideally we'd only load credentials.

> Just a quick comment on naming also: I _think_ @idanov previously suggested that we might drop the `oc.` namespace for this and just have `env` instead.
Thanks for passing it on, I'll have a chat with him as I would personally be for keeping it.

datajoely:
so happy this is coming!

kedro/config/omegaconf_config.py
```diff
@@ -266,6 +281,22 @@ def _check_duplicates(seen_files_to_keys: Dict[Path, Set[Any]]):
...
+    def _resolve_environment_variables(config: Dict[str, Any]) -> None:
+        """Use the ``oc.env`` resolver to read environment variables and replace
+        them in-place, clearing the resolver after the operation is complete if
+        it was not registered beforehand.
+
+        Arguments:
+            config {Dict[str, Any]} -- The configuration dictionary to resolve.
+        """
+        if not OmegaConf.has_resolver("oc.env"):
+            OmegaConf.register_new_resolver("oc.env", oc.env)
```
idanov, line 294: Maybe we can drop the `oc.` or simply support both `env` and `oc.env`?

kedro/config/omegaconf_config.py
```diff
@@ -266,6 +281,22 @@ def _check_duplicates(seen_files_to_keys: Dict[Path, Set[Any]]):
...
+            OmegaConf.resolve(config)
+            OmegaConf.clear_resolver("oc.env")
```
idanov, line 296: It would have been nice if `OmegaConf` did not rely on global state...

idanov (approved):

kedro/config/omegaconf_config.py
jmholzer: Do you mean drop the "oc." in:
1.  name of the resolver (first argument to `register_new_resolver`)
or in:
2. the name assigned to `oc.env` (second argument to `register_new_resolver`)?

I personally wouldn't drop the `oc.` from 1., since this is the name assigned to the resolver by `OmegaConf` when it is instantiated. For 2. I would be ok with losing the namespace from the import: `from omegaconf.resolvers.oc import env`.

PR #2187, opened 2023-01-10
Document use of `--conf-source` option

### LINKED ISSUE #2138: Add documentation for the upcoming `--conf-source` option for `kedro run`
## Description
In https://github.com/kedro-org/kedro/pull/2117, the `--conf-source` option was added to `kedro run`. This lets users specify a location for their configuration at run time.

## Context
We need to clearly communicate to our users the benefits of the new option and how to use it.

## Possible Implementation
I think we should add new documentation in three places:
https://kedro.readthedocs.io/en/stable/kedro_project_setup/configuration.html
https://kedro.readthedocs.io/en/stable/development/commands_reference.html
https://kedro.readthedocs.io/en/stable/nodes_and_pipelines/run_a_pipeline.html

ankatiyar:

## Description

Resolves #2138 

## Development notes

* Document the use of `--conf-source` flag with `kedro run` command in the following places:
    * [`docs/source/development/commands_reference.md`](docs/source/development/commands_reference.md)
    * [`docs/source/kedro_project_setup/configuration.md`](docs/source/kedro_project_setup/configuration.md)
* Also updated the "Customise or Override Project-specific Kedro commands" section in [`docs/source/development/commands_reference.md`](docs/source/development/commands_reference.md) with recent changes 

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/2187"></a>


docs/source/kedro_project_setup/configuration.md
```diff
@@ -8,7 +8,10 @@ We recommend that you keep all configuration files in the `conf` directory of a
 ```python
 CONF_SOURCE = "new_conf"
 ```
-
+You can also specify a source directory for the configuration files at run time using the `--conf-source` flag with the `kedro run` command as follows:
+```bash
+kedro run --conf-source = "new_conf"
```
yetudada, line 13: I think we might need to be a little bit clearer here - is `new_conf` a path or just folder location? Maybe just indicate what users could put there.

yetudada (commented):
Just left a minor comment. This is great!

docs/source/kedro_project_setup/configuration.md
ankatiyar: Thanks Yetu! Updated in f6994bd76c335577c5cbd1dbfd87b3860cf5829a!

datajoely:
Is it worth documenting the minimum required `conf-source` directory structure?

docs/source/development/commands_reference.md
```diff
@@ -312,19 +322,20 @@ kedro run
...
-| `kedro run --config config.yml`                                     | Specify all command line options in a configuration file called config.yml                                                                                                                                                                              | No                          |
-| `kedro run --params param_key1:value1,param_key2:2.0`               | Does a parametrised kedro run with `{"param_key1": "value1", "param_key2": 2}`. These will take precedence over parameters defined in the `conf` directory. Additionally, dot (`.`) syntax can be used to address nested keys like `parent.child:value` | Yes                         |
+| CLI command                                                        | Description                                                                                                                                                                                                                                             | Multiple instances allowed? |
```
stichbury, line 325: I would be tempted to go a bit further with this table since we are editing it. I know you didn't write it @ankatiyar so you may not want to venture into the changes now, which is fine, but I wonder if it wouldn't make more sense to include a bit more standard layout.

So, rather than use examples (like `kedro run --pipeline de`) we should say (`kedro run --pipeline <pipeline_name>` and before the table starts, we make clear that anything within angle brackets is a placeholder for what the user would actually type.

I'm also confused about why there are 2 node names for a starting point (doesn't it start at a single place?) and some are ambiguous (`kedro run --config config.yml` -- can you call it something else or does it have to be `config.yml`). I wonder if we should add another column with examples or further descriptive text for each.

WDYT? Do you want to do this, or shall I put on the backlog for a future revision?

ankatiyar: Thanks @stichbury! I've updated the table. 

> I'm also confused about why there are 2 node names for a starting point (doesn't it start at a single place?) and some are ambiguous (`kedro run --config config.yml` -- can you call it something else or does it have to be `config.yml`). I wonder if we should add another column with examples or further descriptive text for each.

The workflow can start with multiple nodes - for example in the spaceflights tutorial `preprocess_companies_node` and `preprocess_shuttles_node` which are independent of each other. Running the project from one or the other will exclude other nodes on the same level and continue down to nodes on the next level. 
For `kedro run --config config.yml` the config file can have any name, I've updated the table to reflect this.

docs/source/development/commands_reference.md
```diff
@@ -310,21 +320,23 @@ kedro run
...
+| `kedro run --load-version="<dataset_name>:YYYY-MM-DDThh.mm.ss.sssZ"` | Specify a particular dataset version (timestamp) for loading                                                                                                                                                                                            | Yes                         |
+| `kedro run --pipeline <pipeline_name>`                               | Run the whole pipeline by its name                                                                                                                                                                                                                      | No                          |
+| `kedro run --config <config_file_name>.yml`                          | Specify all command line options in a configuration file called config.yml                                                                                                                                                                              | No                          |
```
stichbury, line 337:
```suggestion
| `kedro run --config <config_file_name>.yml`                          | Specify all command line options in a named YAML configuration file                                                                                                                                                                              | No                          |
```

stichbury: One minor change since you've changed the name of the config file. I'd either modify the text from `config.yml` -> `config_file_name.yml` or just avoid mentioning the name as in my suggestion, since the user can choose what they want.

docs/source/kedro_project_setup/configuration.md
```diff
@@ -8,7 +8,10 @@ We recommend that you keep all configuration files in the `conf` directory of a
 ```python
 CONF_SOURCE = "new_conf"
 ```
-
+You can also specify a source directory for the configuration files at run time using the `--conf-source` flag with the `kedro run` command as follows:
```
stichbury, line 11: Maybe include a link here to the `commands_reference.md` table?

stichbury (approved):
Nice work, thanks for the additional changes 🤩 🏆

docs/source/kedro_project_setup/configuration.md
ankatiyar: Done! c1b8e2c46974495cce4b0134ab3beb482efdd160

ankatiyar:
> Is it worth documenting the minimum required `conf-source` directory structure?

@datajoely The section following this in the docs does talk about the `conf/base` and the `conf/local` directory structure. Happy to add it in this section as well but it might be a little redundant?

datajoely:
> > Is it worth documenting the minimum required `conf-source` directory structure?
> 
> @datajoely The section following this in the docs does talk about the `conf/base` and the `conf/local` directory structure. Happy to add it in this section as well but it might be a little redundant?

Oh I didn't realise that - we could link to that anchor possibly, but maybe not needed.

idanov (approved):

PR #1708, opened 2022-07-13
Add gdrive to CLOUD_PROTOCOLS list

SajidAlamQB:
Signed-off-by: SajidAlamQB <90610031+SajidAlamQB@users.noreply.github.com>

## Description

related issue: https://github.com/kedro-org/kedro/issues/1632

Raised a [PR](https://github.com/fsspec/filesystem_spec/pull/988) on fsspec repo to introduce support for some of the missing cloud protocols. In the meantime, this PR will add the missing protocols for Kedro.

## Checklist

- [x] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [x] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [ ] Added tests to cover my changes

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/1708"></a>

PR #2190, opened 2023-01-10
Add support for Azure Data Lake Storage Gen2 URIs

### LINKED ISSUE #2110: Support full URIs in `abfss://` paths
## Description
I found a breaking change in Azure Data Lake Storage Gen 2 paths (`abfss://` protocol) when upgrading from kedro 0.17.7 to 0.18.4 that was undocumented, I was wondering if that was intentional and by design or if it's a bug that should be fixed. 

In Kedro 0.17.7 I had a `JSONDataSet` (it would apply for any `AbstractVersionedDataSet`) that had the path set in the yml catalog configuration using the [abfss URI syntax](https://learn.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-introduction-abfs-uri#uri-syntax). It looked like: `abfss://mycontainer@mystorageaccount.dfs.core.windows.net/mypath`, and worked fine

After upgrading to Kedro 0.18.4, it started failing, because it wouldn't find any of the versions of my file. I believe this is cause by the addition of `abfss` to the list of `CLOUD_PROTOCOLS` in fa02d9b6aed350251e2c136fd593978ceca2b9c3:

https://github.com/kedro-org/kedro/blob/adc7e9f83b3938492c7fa1e897b1e8ff56d737d0/kedro/io/core.py#L31

This causes the filepath I introduced to be altered in this part of the `AbstractVersionedDataSet` source code: 

https://github.com/kedro-org/kedro/blob/adc7e9f83b3938492c7fa1e897b1e8ff56d737d0/kedro/io/core.py#L688-L692

From what I see in the following kedro test, `abfss` implementation was designed to just use the path and not the full URI that includes the container and storage account, but I think supporting full URIs would also be great: 

https://github.com/kedro-org/kedro/blob/adc7e9f83b3938492c7fa1e897b1e8ff56d737d0/tests/io/test_core.py#L67-L70

I don't understand well why the `CLOUD_PROTOCOLS` filepaths prostprocessing is there, and how should we fix this without affecting other protocols and not breaking the "simple path" approach like in the test above. 
How should we update the `CLOUD_PROTOCOLS` implementation for full URIs to work without breaking other protocols?

jmholzer:
## Description

Resolves #2110

## Development notes
There are two ways to fix #2110:

1. Add some custom logic in `kedro.io.core::_parse_filepath` for the `protocol == "abfss"` case.
2. Refactor `kedro.io.core::_parse_filepath` and `kedro.io.core::get_protocol_and_path` in a way that breaks an interface that datasets currently rely on and delegate the `abfss` case to the dataset itself.

Both @merelcht and I do not think 2. is a viable option, since such a refactor is expensive and breaking.

In the past, we have discussed possible refactors of `kedro.io.core` and delegation to `fsspec` (#1708, #1348). I won't be opening an issue concerning this, as @merelcht and I agreed that such work is likely to be low-priority and expensive.

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/2190"></a>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -17,6 +17,7 @@
 * Added support for generator functions as nodes, i.e. using `yield` instead of return.
   * Enable chunk-wise processing in nodes with generator functions.
   * Save node outputs after every `yield` before proceeding with next chunk.
+* Added support for Azure Data Lake Storage Gen2 URIs in datasets.
 
 ## Bug fixes and other changes
 * Commas surrounded by square brackets (only possible for nodes with default names) will no longer split the arguments to `kedro run` options which take a list of nodes as inputs (`--from-nodes` and `--to-nodes`).
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -18,6 +18,7 @@
   * Enable chunk-wise processing in nodes with generator functions.
   * Save node outputs after every `yield` before proceeding with next chunk.
 * Fixed incorrect parsing of Azure Data Lake Storage Gen2 URIs used in datasets.
+* Added support for loading credentials from environment variables using OmegaConfLoader.
 
 ## Bug fixes and other changes
 * Commas surrounded by square brackets (only possible for nodes with default names) will no longer split the arguments to `kedro run` options which take a list of nodes as inputs (`--from-nodes` and `--to-nodes`).
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -17,7 +17,7 @@
 * Added support for generator functions as nodes, i.e. using `yield` instead of return.
   * Enable chunk-wise processing in nodes with generator functions.
   * Save node outputs after every `yield` before proceeding with next chunk.
-* Added support for Azure Data Lake Storage Gen2 URIs in datasets.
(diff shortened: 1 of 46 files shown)


RELEASE.md
```diff
@@ -17,6 +17,7 @@
 * Added support for generator functions as nodes, i.e. using `yield` instead of return.
   * Enable chunk-wise processing in nodes with generator functions.
   * Save node outputs after every `yield` before proceeding with next chunk.
+* Added support for Azure Data Lake Storage Gen2 URIs in datasets.
```
merelcht, line 20: Isn't this more of a bug fix?

merelcht (approved):
LGTM! 👍

RELEASE.md
jmholzer: Yeah that's fair, I think it's a bit of both. Changed the release note in [a33664c](https://github.com/kedro-org/kedro/pull/2190/commits/a33664cd54630b8c7adcdd28a0a3c412fab7a672).

SajidAlamQB (approved):
Awesome! 🌟

MigQ2:
That's great! Thanks a lot @merelcht  and @jmholzer for the fix !

Issue #2038, opened 2022-11-18
More consistent kedro version information in pyproject.toml


sinha1:
## Introduction

[tool.kedro] in pyproject.toml mandates putting kedro version in project_version, which is inconsistent with remaining config defined in pyproject.toml

## Problem
In pyproject.toml mandatory attributes for [tool.kedro] are `package name`, `project_name` and `project_version`. At the moment version in project_version does not represent version of project defined in `project_name` but represents version on kedro used to create the project. This creates conflicting information presented in final pyproject.toml. Example below.

![image](https://user-images.githubusercontent.com/35728909/202738189-28468c97-09bd-4204-97db-585c7b026662.png)

The check is implemented [here](https://github.com/kedro-org/kedro/blob/82d69f43b3d8ce93605c272557e22547bca5bf11/kedro/framework/startup.py#L95) 

As a side effect, current nature of check also allows user to enter patch version in their pyproject.toml which does not exist. For example if you are running kedro 0.17.7 `project_version = "0.17.99"` will be a valid configuration for pyproject.toml even though the kedro version don't exist. 

### Alternatives considered

It will create a consistent interface if we call `project_version` as  `max_kedro_version` or a similar name, due to following reasons:
1. It will represent actual information defined in this attribute i.e. max compatible kedro version for this project.
2. In the context of a kedro project, `project_version` signifies version of the project, rather than version of kedro, so it will make interface more intuitive.
3. It would remove conflicting information from pyproject.toml illustrated in screenshot above.

## Rollout strategy

It will be great to have this resolved for kedro 0.17.

merelcht:
Hi @sinha1, we will be renaming the `project_version` in https://github.com/kedro-org/kedro/issues/2118. However, we won't be able to change this for older versions of Kedro.

PR #2219, opened 2023-01-17
Rename `project_version` in `pyproject.toml`

SajidAlamQB:

## Description
In `pyproject.toml` we set the following fields for kedro:
```
[tool.kedro]
package_name = "{{ cookiecutter.python_package }}"
project_name = "{{ cookiecutter.project_name }}"
project_version = "{{ cookiecutter.kedro_version }}"
```

However, when you've packaged your project it's not intuitive for the `project_version` to point to the `kedro_version`. We should rename `project_version`. 

related to: https://github.com/kedro-org/kedro/issues/2038

- Rename `project_version` to `kedro_init_version`
- If `project_version` is still in `pyproject.toml` the project shouldn't fail, but emit a deprecation warning saying the user should use `kedro_init_version` instead. 

## Development notes

- Renamed `project_version` to `kedro_init_version`
- If `project_version` is still in `pyproject.toml` its treated as `kedro_init_version` and emits a deprecation warning saying the user should use `kedro_init_version` instead. 
- Modified tests for `kedro_init_version` and added two new tests to cover when `project_version` is used (these need to be removed in 0.19.0)
## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/2219"></a>


merelcht (commented):
I'll do a more thorough review later, but the `pyproject.toml` in the template should also be updated to have `kedro_init_version` instead of `project_version`. 

One thought I also had is what would happen if the file contains both `kedro_init_version` and `project_version`? It's not a super likely scenario, but maybe it could happen if a user updates their `pyproject.toml` incorrectly.

SajidAlamQB:
> I'll do a more thorough review later, but the `pyproject.toml` in the template should also be updated to have `kedro_init_version` instead of `project_version`.
> 
> One thought I also had is what would happen if the file contains both `kedro_init_version` and `project_version`? It's not a super likely scenario, but maybe it could happen if a user updates their `pyproject.toml` incorrectly.

If both `kedro_init_version` and `project_version` exist in `metadata_dict` then it will go through the first if condition only which will remove `project_version` and assign its value to `kedro_init_version` replacing whatever value `kedro_init_version` had.

kedro/framework/startup.py
```diff
@@ -19,13 +20,13 @@ class ProjectMetadata(NamedTuple):
     package_name: str
     project_name: str
     project_path: Path
-    project_version: str
+    kedro_init_version: str
```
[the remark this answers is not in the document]
SajidAlamQB: Since this is a factory method, `NamedTuple`, there doesn't seem to be a way to create a direct alias for its properties. Instead, I've kept both the `project_version` and `kedro_init_version` properties in `ProjectMetadata` as a temporary solution until 0.19. Both properties are treated the same way and given the same value with the only difference being that if `project_version` is found a deprecation warning is thrown. 

Do let me know your thoughts on this proposed solution.

kedro/framework/startup.py
```diff
@@ -86,14 +88,38 @@ def _get_project_metadata(project_path: Union[str, Path]) -> ProjectMetadata:
...
+            DeprecationWarning,
+        )
+        kedro_init_version = metadata_dict["project_version"]
+        metadata_dict["kedro_init_version"] = kedro_init_version
+        mandatory_keys.append("kedro_init_version")
+        # check the match for major and minor version (skip patch version)
+        if kedro_init_version.split(".")[:2] != kedro_version.split(".")[:2]:
+            raise ValueError(
+                _version_mismatch_error(metadata_dict["kedro_init_version"])
+            )
```
merelcht, line 109: This is quite a big chunk of repeated code. Can we re-order the if/else so this can happen outside of it and doesn't need to be repeated?

tests/framework/cli/conftest.py
```diff
@@ -87,6 +87,7 @@ def fake_metadata(fake_root_dir):
         "CLI Testing Project",
         fake_root_dir / REPO_NAME,
         kedro_version,
+        kedro_version,
```
merelcht, line 90: `kedro_version` seems to be duplicated here now, is that necessary?

merelcht (commented):
This is definitely going into the right direction! I've left some comments & questions

tests/framework/cli/conftest.py
SajidAlamQB: As @idanov's comments recommend we should keep `project_version` in `ProjectMetadata` as it would be a breaking change otherwise. Since the `ProjectMetadata` class is a factory method, `NamedTuple`, there didn't seem to be a way to create a direct alias for `project_version` to `kedro_init_version`. Instead, I've kept both the `project_version` and `kedro_init_version` properties in `ProjectMetadata` as a temporary solution until 0.19. They both will always have the same value.

Not sure if this is the best approach?

As a result, when doing the unit tests we now have to take into account both `project_version` and `kedro_init_version` hence the repeated `kedro_version`.

kedro/framework/startup.py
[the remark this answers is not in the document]
idanov: Could we then use `@dataclass(frozen=True)` instead then? Would that potentially break anything?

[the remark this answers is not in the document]
idanov: Also `kedro_init_version` should go last, otherwise this is still a breaking change due to the order of arguments in the constructor.

merelcht (commented):
I think this is nearly done. The only remaining changes should be to move `kedro_init_version` to be the last argument in the `ProjectMetadata` class and to add the change to the release notes.

merelcht (approved):
Nice work! Don't forget to update the release notes ✍️

idanov (approved):

kedro/framework/startup.py
```diff
@@ -86,14 +88,31 @@ def _get_project_metadata(project_path: Union[str, Path]) -> ProjectMetadata:
...
+    else:
+        raise RuntimeError(
+            f"Missing required key kedro_init_version from '{_PYPROJECT}'."
+        )
+
+    mandatory_keys.append("kedro_init_version")
     # check the match for major and minor version (skip patch version)
-    if metadata_dict["project_version"].split(".")[:2] != kedro_version.split(".")[:2]:
-        raise ValueError(_version_mismatch_error(metadata_dict["project_version"]))
+    if kedro_init_version.split(".")[:2] != kedro_version.split(".")[:2]:
```
noklam, line 114: I was a bit confused when reading this logic, as I understand that `metadata_dict` will be patched based on whether certain keys exist. If this is the case, I think we can remove the variable `kedro_init_version` and use `metadata_dict["kedro_init_version"]` as this will be the final truth.

noklam (approved):
Looks good! Left a small comment.

PR #2252, opened 2023-01-24
Comment out a CircleCI workflow that isn't run now

deepyaman:
## Description

There's an error on `main`/`develop` nightly pipelines:

image

Error message is:

```
Value does not match schema: {:workflows {:nightly_build_develop {:jobs (not (map? nil))}}}
```

I figured this is because it's registering an empty matrix (whenever `develop` and `main` support the same set of Python versions, which is probably most of the time).

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/2252"></a>


merelcht:
Do we need to keep this workflow at all? I'm not quite sure what it's supposed to do.. 🤔

deepyaman:
> Do we need to keep this workflow at all? I'm not quite sure what it's supposed to do.. 🤔

@merelcht Sorry I missed your response. I believe it exists in case we support different sets of Python versions on `main` and `develop` (e.g. if we add 3.12 support, but it requires breaking changes). It's not symmetric, so it doesn't help if we deprecate 3.7 on `develop` ahead of EOL in order to take advantage of some new functionality.

I'm OK with removing or commenting it; the main benefit for commenting it is, if we come across this situation again where `develop` supports additional Python versions, it's more obvious where the supporting code is.

merelcht:
Ahh now I get the use of this. In that case let's comment it out but leave it in!

merelcht (approved):

noklam (approved):

PR #2180, opened 2023-01-06
Patch CVE-2007-4559

TrellixVulnTeam:
# Patching CVE-2007-4559

Hi, we are security researchers from the Advanced Research Center at [Trellix](https://www.trellix.com). We have began a campaign to patch a widespread bug named CVE-2007-4559. CVE-2007-4559 is a 15 year old bug in the Python tarfile package. By using extract() or extractall() on a tarfile object without sanitizing input, a maliciously crafted .tar file could perform a directory path traversal attack. We found at least one unsantized extractall() in your codebase and are providing a patch for you via pull request. The patch essentially checks to see if all tarfile members will be extracted safely and throws an exception otherwise. We encourage you to use this patch or your own solution to secure against CVE-2007-4559. Further technical information about the vulnerability can be found in this [blog](https://www.trellix.com/en-us/about/newsroom/stories/research/tarfile-exploiting-the-world.html).

If you have further questions you may contact us through this projects lead researcher [Kasimir Schulz](mailto:kasimir.schulz@trellix.com).

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/2180"></a>

PR #2268, opened 2023-01-31
Add support for reading config from tar.gz/zip files

### LINKED ISSUE #2146: Add support to read Kedro config from a zip/tar.gz file
## Description

After adding `--conf-source` to our `kedro run` command with https://github.com/kedro-org/kedro/pull/2117, we should extend this functionality by making it possible to point to a zip/tar.gz archive in order to simplify further Kedro project deployments. 

## Context

Deploying Kedro projects will be simplified if the configuration can be read from an archive, instead of pointing to a full directory. Many big projects rely on multiple configuration files and shipping all of them to a production environment is a bit more difficult than just pointing to a zip/tar.gz archive. 

## Possible Implementation

We can take advantage of `fsspec` and provide support to any endpoints (including remote endpoints) to read the configuration from. We can also package the existing configuration as a `conf-<project_package>-<version>.tar.gz` file (example naming) on running `kedro package` for convenience in deployment.

Relevant resources to explore (and maybe even document on our docs): https://filesystem-spec.readthedocs.io/en/latest/features.html#url-chaining

Also add docs for this functionality. Can be done in a separate PR, but should be released in the same release.

ankatiyar:

## Description

Resolves #2146 This PR allows users to read their config from a tar.gz/zip file

### Documentation changes -
* Release notes - `RELEASE.md`
* Docs -
    * CLI command table in `docs/source/development/commands_reference.md`
    * First section in `docs/source/kedro_project_setup/configuration.md`
### Micro-packaging related changes - 
* Moved `safe_extract` and `_is_within_directory` function introduced in #2180 to `kedro/utils.py` so it can be used for extracting config files as well
* Updated `safe_extract` function to also extract zip files
* Subsequently updated `test_path_traversal` also added in #2180 in `tests/framework/cli/micropkg/test_micropkg_pull.py` to work with the above changes
### Config Changes
* In `kedro/framework/cli/project.py` - allow files as well as directory to `--conf-source`
* In `kedro/framework/session/session.py` added function `resolve_conf_source` which returns the `conf_source` path which can be - 
    * path to config provided by the user 
    * path to the unarchived tar.gz/zip config directory
* Tests for both tar.gz and zip config sources in `tests/framework/cli/test_cli.py`

NOTE : I originally implemented the `resolve_conf_source` function as a callback function to be called with `--conf-source` flag but I moved it to `kedro/framework/session/session.py` so that it also works when `CONF_SOURCE` is set in `settings.py`

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/2268"></a>


merelcht (commented):
I think we need to add some more concrete examples of how this works. So adding an explicit example of how to call the command with an archive file and some guidelines as well around what structure is expected of the config that's compressed. 

I would expect to be able to zip up a default Kedro config structure: 
```
conf
   /base
       /catalog.yml
       /parameters.yml
       /logging.yml
   /local
       /catalog.yml
   /prod
      /catalog.yml
      /logging.yml
```

Would something like that work?

ankatiyar:
Exploring a different approach - using `fsspec` so closing this in favour of #2298

Issue #2271, opened 2023-02-01
Add node to dataset hooks


AntonyMilneQB:
Following https://github.com/kedro-org/kedro/issues/2258 and tech design discussion, we're going ahead with adding `node` to the dataset hooks.

This should be straightforward:
* all four dataset hook specifications should be updated to include `node: Node` as an argument
* all hook calls (in both `_run_node_sequential` and `_run_node_async`) need to be modified to include `node` 
* update tests

Issue #2258, opened 2023-01-25
Add node to the dataset hooks


AntonyMilneQB:
## Description

Here's the current behaviour of hooks when a node is run:
1. For each input dataset:
    1.  run `before_dataset_loaded` hook
    2. load dataset
    3. run `after_dataset_loaded` hook
4. Run `before_node_run` hook
5. Run node
6. Run `after_node_run` hook
7.  For each output dataset:
    1.  run `before_dataset_saved` hook
    2. load dataset
    3. run `after_dataset_saved` hook

The dataset hooks are quite sparse in terms of arguments: they only have `dataset_name` and `data` (apart from `before_dataset_loaded`, which obviously doesn't have `data` available yet). We should add `node` to these hook specifications.

This change has been discussed before in the context of W&B. It's a trivial change in terms of code, but does increase the API surface so should be given due consideration.

> **Note**. Both `_run_node_sequential` and `_run_node_async` need to be modified similarly here.

## Context
I've now seen about five independent cases in which dataset hooks require the context of the `node`, most recently @DavidRoschewitzQB. The current workaround is something like this:
```
class Hooks:
    @hook_impl
    def after_node_run(self, node: Node) -> None:
        self.node = node

    @hook_impl
    def before_dataset_saved(self, dataset_name: str) -> None:
        # use self.node
```

It would be much cleaner to just make `node` an argument to the dataset hooks instead.

Here's some rough pseudocode of where it would be useful for W&B. Other examples I've seen are along similar lines (mlflow artifactory, profiling hooks, rich dataset loading/saving hooks).
```
class WandBHooks:
    def __init__(self):
        self.wandb_nodes = {} # dict mapping node name to set of node outputs

    @hook_spec
    def before_dataset_loaded(node):
        if node.name not in wandb_nodes:
            wandb.init(name=node.name)
            self.wandb_nodes[node.name] = set(node.outputs)

    def after_dataset_saved(dataset_name, data, node):
        self.wandb_nodes[node.name].remove(dataset_name)
        wandb.log_artififact(data)
        if not self.wandb_nodes[node.name]: # all outputs have been written
            wandb.finish(name=node.name)
```

## Possible Alternatives
[...]

PR #2296, opened 2023-02-07
Add node as an argument to all datasets' hook

noklam:

## Description

See #2271 for more details. To support [wider use cases](https://github.com/kedro-org/kedro/issues/2258) with dataset hook.

* all four dataset hook specifications should be updated to include node: Node as an argument
* all hook calls (in both `_run_node_sequential` and `_run_node_async`) need to be modified to include node
* update tests

## Development notes

* Update all 4 hooks' spec to include `node` argument.
* Update the hook call in `runner.py` to include `node` argument
* Update the tests, and move some fixtures around to make sure it's serialisable.
* 
## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/2296"></a>


noklam:
Notes:
When working on the tests I find that the Catalog was created multiple times, and the data has been saved more than once. Thus I have to hard-coded the `call_args[1]` when I expect it to be `call_args[0]`, the root cause is not clear but I suspect it's the setup we have in `conftest.py`. I ignore it to make the test passed will come back later if I have time. Note that this is not introduced by the changes, this issue exists before the change.

image

tests/framework/session/test_session_extension_hooks.py
```diff
@@ -523,21 +523,20 @@ def test_broken_input_update_parallel(
             mock_session_with_broken_before_node_run_hooks.run(runner=ParallelRunner())
 
 
+def wait_and_identity(*args: Any):
```
jmholzer, line 526: This is a nicer implementation of this fixture ✅.

Out of curiosity, could you explain what you meant in the commit message 'move the inner function out to make the node serializable'?

jmholzer (approved):
Looks like a clean implementation.

Would you like me to manually test these changes?

tests/framework/session/test_session_extension_hooks.py
noklam: @jmholzer Sure. The original fixture defined an inner function, such object is not picklable and does not work well with multiprocessing. This wasn't an issue until node is added as an argument for hooks. You can simply move the function inside and run the test.

IIRC it's the TestAsyncDataSet.

noklam: You are more than welcome to test it manually 😁

merelcht (commented):
LGTM! 👍 

Only left some small comments on doc strings. This should also be included in the release notes, and perhaps also update this page for consistency: https://kedro.readthedocs.io/en/stable/hooks/common_use_cases.html#use-hooks-to-customise-the-dataset-load-and-save-methods

tests/framework/session/test_session_extension_hooks.py
jmholzer: > @jmholzer Sure. The original fixture defined an inner function, such object is not picklable and does not work well with multiprocessing. This wasn't an issue until node is added as an argument for hooks. You can simply move the function inside and run the test.
> 
> IIRC it's the TestAsyncDataSet.

TIL, thanks 😃

> You are more than welcome to test it manually 😁

I tried it out, looks good ✅.

kedro/framework/hooks/specs.py
```diff
@@ -240,41 +240,45 @@ class DatasetSpecs:
...
 
     @hook_spec
-    def after_dataset_loaded(self, dataset_name: str, data: Any) -> None:
+    def after_dataset_loaded(self, dataset_name: str, data: Any, node: Node) -> None:
         """Hook to be invoked after a dataset is loaded from the catalog.
 
         Args:
             dataset_name: name of the dataset that was loaded from the catalog.
             data: the actual data that was loaded from the catalog.
+            node: The ``Node`` to ran.
```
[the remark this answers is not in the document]
noklam: @merelcht I think both `before_dataset_loaded` and `after_dataset_loaded` executed before the node is ran, we should keep this one as it is.

kedro/framework/hooks/specs.py
```diff
@@ -240,41 +240,45 @@ class DatasetSpecs:
     """Namespace that defines all specifications for a dataset's lifecycle hooks."""
 
     @hook_spec
-    def before_dataset_loaded(self, dataset_name: str) -> None:
+    def before_dataset_loaded(self, dataset_name: str, node: Node) -> None:
         """Hook to be invoked before a dataset is loaded from the catalog.
 
         Args:
             dataset_name: name of the dataset to be loaded from the catalog.
+            node: node: The ``Node`` to ran.
```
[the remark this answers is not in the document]
noklam: I made some more changes to the documentation, the previous one is buggy and won't even print the log. I also considered updating the example for `MemoryProfilingHooks`, but decided not as that example is purely about dataset profiling thus the node argument isn't useful.

Couple changes
* `logging.getLogger(self._class.__name__)` -> `logging.getLogger(__name__)` - this makes sure the log will be printed properly with the default `logging.yml`
* Instead of printing the absolute timestamp, the hook print how many time it takes to load a dataset instead.

kedro/framework/hooks/specs.py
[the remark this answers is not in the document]
merelcht: In that case it should be "The ``Node`` to run" instead of "to ran", to make it grammatically correct.

Issue #2246, opened 2023-01-23
Change docs to use plural flags in `kedro run` suggestions and examples 


AhdraMeraliQB:
## Description
Prerequisites: #2243 and #2244 

We are switching over to using plural flags to replace `--node`, `--tag`, and `--load-version`. These changes should also be reflected in the docs wherever we give examples of using these flags.

Additionally, the use of these plural flags affects the docs on using `config.yml` for kedro runs: nodes and tags will no longer take lists, instead they should take comma separated arguments. See https://github.com/kedro-org/kedro/issues/1792#issuecomment-1397090812 for further context. As a result, https://kedro.readthedocs.io/en/stable/kedro_project_setup/configuration.html#configure-kedro-run-arguments should also be updated under this ticket.

AhdraMeraliQB:
Note: The table of commands on the docs page https://kedro.readthedocs.io/en/stable/development/commands_reference.html#modifying-a-kedro-run has a column titled "Multiple instances allowed?". This should be removed as no commands will have multiple instances allowed anymore, making this column redundant.

Issue #2244, opened 2023-01-23
Add plural flags `--nodes`, `--tags`, and `--load-versions` to kedro run


AhdraMeraliQB:
## Description
As part of #2014 it was agreed that we ought to be using plural flags for arguments that accept one or many values, which means `--node`, `--tag` and `--load-version` need to be replaced. This implementation **should not** make use of `multiple=True`.

Note: For the flags `--node` and `--tag` it is important to strip whitespaces from the values passed in as node names do not allow for spaces, and tags will be changed to follow suit. See the callback functions for parsing `--from-nodes` and `--from-inputs` as examples.

To test that whitespaces are stripped properly, ensure `kedro run --nodes="node1, node2"` runs as expected (and equivalents for `--tags` and `--load-versions`).

Issue #2247, opened 2023-01-23
Investigate further inconsistencies between the CLI and Kedro API


AhdraMeraliQB:
## Description
#2014 Explored inconsistencies within the CLI itself, but it stands that the CLI is tightly coupled with the API. This ticket looks at investigating any inconsistencies in this wider context with the aim of streamlining our user experience. For example:

- Parameters in [KedroSession.run()](https://kedro.readthedocs.io/en/stable/kedro.framework.session.session.KedroSession.html) vs CLI `kedro run`
- Parameter names in `config.yml` supplied to `kedro run` vs CLI `kedro run` flags

This list is far from exhaustive, but serves to highlight areas of consideration when conducting this investigation.

astrojuanlu:
Potentially useful: https://clig.dev/

AntonyMilneQB:
Copied from https://github.com/kedro-org/kedro/pull/2301#discussion_r1104266236:

> Fine to keep `node_names` for now in #2245 but as part of https://github.com/kedro-org/kedro/issues/2247 I think we should consider just calling this `nodes`. (Just while I remember, Ivan's argument for why we might not want to do this: it's deliberately written as `node_names` to deter people from trying to pass node objects themselves to the underlying `session.run`. My argument for it is consistency and simplicity, particularly when we eventually return to https://github.com/kedro-org/kedro/pull/1423)

AntonyMilneQB:
From discussion with @jmholzer and @noklam in https://github.com/kedro-org/kedro/pull/2306: there's inconsistent ordering between our click definition of `run` and the Python `KedroSession.run` API. This isn't a huge problem but would be nice to have them the same.

Proposed solution:
1. Decide what the correct order of arguments is in click definition of `run` (which determines order flags appear in `kedro run --help`). Maybe this is already correct.
2. Re-order arguments in `KedroSession.run`.
3. Make `KedroSession.run` arguments keyword-only (i.e. `def run (self,  *, ...)` so we can alter the order or insert new arguments wherever we see fit in the future  

I doubt anyone calls `KedroSession.run` without using keywords at the moment, but this is a breaking change so should only be done in 0.19. Plus it would make sense to do any renaming of arguments at the same time as this anyway.

noklam:
I like the **keyword-only** idea and it should make things more flexible, if we shuffle the order in the future it won't be a breaking change anymore, is it correct?

[...]

Issue #2245, opened 2023-01-23
Remove kedro run flags `--node`, `--tag`, and `--load-version`


AhdraMeraliQB:
## Description
**IMPORTANT: This is a breaking change and should go into develop. This should only be implemented after https://github.com/kedro-org/kedro/issues/2243 and https://github.com/kedro-org/kedro/issues/2244.**

Once we have implemented the replacement flags, these singular flags should be removed. In this change, ensure the short-form flags `-n`, `-t` and `-lv` are modified to alias the plural form flags `--nodes`, `--tags`, and `--load-versions` instead.

AhdraMeraliQB:
Migrating over some comments from #2244:

- Currently `--node` results in the parameter `node_names` and `--nodes` in `nodes_names`. When `--node` is removed make sure to rename the respective parameter to `node_names`.
- Once the old flags are deleted `_get_vaules_as_tuple` should be able to be deleted as well in favour of using `_split_string`
- `_reformat_load_version` will no longer need to be separate from `_split_load_versions` and can be combined into one function under `_split_load_versions`
- Try to have as much argument wrangling/formatting as possibke handled in the callback functions instead of the call to `session.run`

AhdraMeraliQB:
Don't forget to remove instances of the singular flags from the docs. These pages are documented in #2246.

AhdraMeraliQB:
Addressed in #3279

PR #2301, opened 2023-02-08
Add plural flags --nodes, --tags, and --load-versions to `kedro run`

AhdraMeraliQB:

## Description

This PR completes #2244 by adding pluralised flags for the node, tag, and load-version options on `kedro run`

### For Reviewers:

- `_split_load_versions` was added to handle input strings, as previously --load-version would process the input as tuples due to `multiple=True`.
- `.strip()` is used to trim the whitespaces from user input
- I couldn't find any unit tests for the individual tags; I've added tests for the new flags but have not duplicated them for the old singular flags as they will just be deleted in 0.19.0. These tests check inputs with a singular value, multiple comma-separated values, and multiple comma-and-space-separated values.
- The test for `--load-versions` is subsumed under the test for `_split_load_versions` just as testing `--load-version` was by `test_reformat_load_versions`.
- For the period between 0.18.5 and 0.19.0 users will be able to use both versions of the flags in their `kedro run` commands. Whilst I don't think they will, is it worth considering limiting this behaviour?

## Development notes
The helper function `_split_load_versions` was added as `--load-version` and `--load-versions` return different types, the singular returns a tuple and the plural a string. This is as a result of using `multiple=True` for the singular flag. 

The both new and old tags have been tested manually to ensure they (still) behave as expected.

- [x] Add plural flags
- [x] Test they work for single inputs
- [x] Test they work for multiple, comma-separated inputs
- [x] Test behaviour against whitespaces
N/B: Tags are robust against whitespaces
- [x] Trim whitespaces for load-versions
- [x] Trim whitespaces for nodes
- [x] Add tests for plural flags (single input, multiple input)
- [x] Make sure tests check whitespaces handling in input
- [x] Test behaviour when using both singular and plural flags in one command - it works!
- [x] Add test for _split_load_versions
- [x] Add additional tests (if necessary)

## Checklist

- [ ] Updated the documentation to reflect the code changes - to be handled in #2246 
- [x] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file - to be handled in #2303 

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/2301"></a>


AhdraMeraliQB:
Question for reviewers: Do we want to restrict users from using both `--node` and `--nodes` while both are available?

kedro/framework/cli/project.py
```diff
@@ -374,18 +398,21 @@ def activate_nbstripout(
...
-# pylint: disable=too-many-arguments,unused-argument
+# pylint: disable=too-many-arguments,unused-argument, too-many-locals
 def run(
     tag,
+    tags,
     env,
     runner,
     is_async,
     node_names,
+    nodes_names,
```
AhdraMeraliQB, line 409: This is named this way for differentiation from the already existing `node_names` - in #2245 I'd suggest keeping `node_names` as the parameter name

merelcht (commented):
This looks good! I think the one thing missing is a test (or tests) to check that providing both `--node` and `--nodes` and tags, and load versions, works properly as well. Because now we're not really checking that these lines:

```
tag = tag + tags
node_names = node_names + nodes_names
load_version = {**load_version, **load_versions}
```
work fine

kedro/framework/cli/project.py
```diff
@@ -360,6 +369,13 @@ def activate_nbstripout(
     help=TAG_ARG_HELP,
     callback=_deprecate_options,
 )
+@click.option(
+    "--tags",
+    type=str,
+    default="",
```
AntonyMilneQB, line 375: Is this `default` needed? I see it's in many other options as well so fine to leave it. I'm just curious.

kedro/framework/cli/project.py
AntonyMilneQB: Fine to keep `node_names` for now in #2245 but as part of https://github.com/kedro-org/kedro/issues/2247 I think we should consider just calling this `nodes`. (Just while I remember, Ivan's argument for why we might not want to do this: it's deliberately written as `node_names` to deter people from trying to pass node objects themselves to the underlying `session.run`. My argument for it is consistency and simplicity, particularly when we eventually return to https://github.com/kedro-org/kedro/pull/1423)

kedro/framework/cli/project.py
```diff
@@ -412,6 +438,16 @@ def run(
     tag = _get_values_as_tuple(tag) if tag else tag
     node_names = _get_values_as_tuple(node_names) if node_names else node_names
 
+    # temporary duplicates for the plural flags
+    tags = _get_values_as_tuple(tags) if tags else tuple(tags)
```
AntonyMilneQB, line 442: Minor, but do we need these `if/else` clauses? Since `_get_values_as_tuple("") == tuple("") == tuple()` already?

kedro/framework/cli/utils.py
```diff
@@ -422,9 +423,12 @@ def _reformat_load_versions(  # pylint: disable=unused-argument
     """Reformat data structure from tuple to dictionary for `load-version`, e.g.:
     ('dataset1:time1', 'dataset2:time2') -> {"dataset1": "time1", "dataset2": "time2"}.
     """
-    _deprecate_options(ctx, param, value)
+    if param.name != "load_versions":
```
AntonyMilneQB, line 426:
```suggestion
    if param.name == "load_version":
```
Feels a bit easier to understand if we condition based on what we _are_ deprecating here rather than what we're not.

kedro/framework/cli/utils.py
```diff
@@ -468,6 +472,11 @@ def _split_params(ctx, param, value):
     return conf
 
 
+def _split_load_versions(ctx, param, value):
+    lv_tuple = _get_values_as_tuple([value])
```
AntonyMilneQB, line 476: Do we need the `[value]` here rather than `value`? Elsewhere we call `_get_values_as_tuple` just with a single string. 

Also do we need the `if/else` here? Wouldn't `_reformat_load_versions` work the same just passing the empty `value` into it directly?

After we've removed the old `load_version` we should definitely just move `_reformat_load_versions` code into this instead of having two separate functions.

AntonyMilneQB (approved):
Really great PR 👍 Super clear explanations and excellent testing ⭐  I'm a little bit confused about `_get_values_as_tuple`, but that's not really a result of your changes. I wonder if that could be simplified after (or as part of) the ticket to remove the old flags, since it will no longer need to handle anything more complicated than just a single string (?). Happy to chat more about this anyway so we can try to figure it out!

> Question for reviewers: Do we want to restrict users from using both `--node` and `--nodes` while both are available?

I think it's fine as it is. I don't think anyone will really try to do that, it's only going to be possible for a limited period of time, plus they will already get the deprecation warning and things will work ok when you use both. So not worth the extra code required to handle it - already combining the arguments together is more than enough I think.

kedro/framework/cli/project.py
AhdraMeraliQB: `default` provides a value in the case that the flag isn't used - here it needs to be an empty string otherwise we'll get an error in the callback function when a string is expected but `None` is passed through. It eliminates the need to have a `if tags:` statement in the callback function

kedro/framework/cli/project.py
AhdraMeraliQB: Good catch - was following convention from the earlier implementation but looks like they weren't needed there either. My suspicion is that `multiple=True` implies a default value of an empty tuple 🤔

AhdraMeraliQB:
@AntonyMilneQB 
> I'm a little bit confused about _get_values_as_tuple, but that's not really a result of your changes. I wonder if that could be simplified after (or as part of) the ticket to remove the old flags, since it will no longer need to handle anything more complicated than just a single string (?)

It should be able to be removed 🤔 it's functionality then wouldn't extend beyond `split_string`, if the return type is changed from list to tuple. Definitely worth evaluating in #2245

kedro/framework/cli/project.py
AntonyMilneQB: This makes sense, thank you.

kedro/framework/cli/utils.py
AntonyMilneQB: Cool, that's what I suspected about `multiple=True` 👍  

> The `if/else` isn't needed as `_reformat_load_versions` would still work - the idea was to prevent calling another method if unnecessary. Which approach would be considered more pythonic?

Overall I think without the `if/else`. The performance overhead of calling another function here is very small compared to the (also small) additional complexity of branching in the code. It's a small thing, but I'd only skip calling the function here if it was particularly expensive.

kedro/framework/cli/project.py
AntonyMilneQB: Cool, that's what I suspected about `multiple=True` 👍  Just to double check though, it's ok to remove the `if/else` for both the `multiple=True` and the `multiple=False` versions of these arguments?

AntonyMilneQB:
@AhdraMeraliQB 

> It should be able to be removed 🤔 it's functionality then wouldn't extend beyond `split_string`, if the return type is changed from list to tuple. Definitely worth evaluating in #2245

This would be a nice simplification. It would also be nice if we could put all this argument-munging into the callbacks rather than needing to do them inside the `run` function itself. Please could you add any of these proposed simplifications to #2245 (or wherever you think is appropriate)? I think we should try and do them sooner rather than later while they're fresh in our minds.

merelcht (approved):
Great work! 👍 Don't forget to update the release notes.

merelcht:
> The changes were combined in the release notes of #2303: "Deprecate (singular flags) in favour of (plural flags)" - should this be elaborated on further?

Yes, I would separately add that these flags have been added. That way it's also easier to trace which PR the changes were added in.

kedro/framework/cli/project.py
AhdraMeraliQB: Yep! For the other flags their input is immediately handled by the callback function which goes on to return a list, and it's this list that get's passed to `_get_values_as_tuple`, which splits them safely

PR #2306, opened 2023-02-10
Add new function to specify namespace for `KedroSession` and `kedro run`

noklam:
## Checklist


datajoely:
Super cool!

kedro/framework/session/session.py
```diff
@@ -336,6 +337,7 @@ def run(  # pylint: disable=too-many-arguments,too-many-locals
                 used as an end point of the new ``Pipeline``.
             load_versions: An optional flag to specify a particular dataset
                 version timestamp to load.
+            namespace: The namespace of the nodes that is being run.
```
jmholzer, line 340:
```suggestion
            namespace: An optional string specifying a namespace to run. If specified, only the nodes in this namespace will be run.
```

kedro/framework/cli/project.py
```diff
@@ -353,6 +355,7 @@ def activate_nbstripout(
     callback=_reformat_load_versions,
 )
 @click.option("--pipeline", "-p", type=str, default=None, help=PIPELINE_ARG_HELP)
+@click.option("--namespace", "-ns", type=str, default=None, help=NAMESPACE_ARG_HELP)
```
jmholzer, line 358: To keep the ordering of `click.option` decorators consistent with the arguments in the function signature, I think this should be moved to the bottom.

jmholzer (approved):
Nice work ✅

> Do we need a specific test for the namespace option? I don't think it's needed as it is more or less equal to other arguments and the filtering function is tested in the test_pipeline.py already.

I think the updated `test_run_with_pipeline_filters` makes an extra test unnecessary.

> The ordering for these cli options and KedroSession.run is kind of arbitrary, for KedroSession it is the code API so it makes sense to add the new option as the last argument to keep it backward compatible.

I think the ordering of `click.option` decorators is consistent with the order in which their passed value appears in the function signature of `run`, so personally I think it should be added to the end if possible.

> Noted some arguments have inconsistent names. For example kedro run --pipeline vs KedroSession.run(pipline_name=xxx)

I'll take a look if this is on our radar, if not I'll open a ticket.

kedro/framework/cli/project.py
noklam: Do we actually follow the order? The first `click` option is `from_input` but the first argument would be `tag`. I think it only affects when you do `kedro run -h`, it will follow the order that how you decorate your function. In general, I think it should follow the order of importance. I would expect `pipeline` to be the first argument in that case.

I don't have a strong opinion about moving it to the back, the only reason I keep it next to `pipeline` is that I feel these two arguments are in the same category. Similarly, the `nodes` arguments are together, and the config-related arguments are grouped as well. @merelcht or @AntonyMilneQB Is this something we've discussed in the past?

RELEASE.md
```diff
@@ -19,6 +19,7 @@
   * Save node outputs after every `yield` before proceeding with next chunk.
 * Fixed incorrect parsing of Azure Data Lake Storage Gen2 URIs used in datasets.
 * Added support for loading credentials from environment variables using `OmegaConfigLoader`.
+* Added new ability to filter by namespace for `KedroSession.run` and `kedro run` with `--namespace`.
```
AntonyMilneQB, line 22:
```suggestion
* Added new `--namespace` flag to `kedro run` to enable filtering by node namespace.
```
I don't think `KedroSession.run` is worth mentioning here just to keep release notes concise, since CLI `kedro run` is how 99% of kedro runs are implemented rather than the Python API.

kedro/framework/cli/project.py
```diff
@@ -53,6 +53,8 @@
 override the loaded ones."""
 PIPELINE_ARG_HELP = """Name of the registered pipeline to run.
 If not set, the '__default__' pipeline is run."""
+NAMESPACE_ARG_HELP = """Name of the namespace to run.
+If not set, no namespace is used."""
```
AntonyMilneQB, line 57:
```suggestion
NAMESPACE_ARG_HELP = """Name of the node namespace to run."""
```
IMO the "no namespace is used" just makes this less clear since it can't be really explained fully in just a line.

kedro/framework/cli/project.py
AntonyMilneQB: Fully agree with you here @noklam 👍  We should go for what makes sense to the user when you do `kedro run --help`; no need to be consistent with the order of `KedroSession.run` (which as you've said is somewhat arbitrary).

AntonyMilneQB (commented):
LGTM ⭐ 

On the questions you ask:
1. Agreed with @jmholzer, tests look sufficient 👍
2. Agreed with @noklam. We should aim for a good CLI UX first and then consistency with Python API second. Adding the argument to the end of the `KedroSession.run` is correct to maintain backwards compatibility but doesn't need to be reflected in the click `run` command - as you notice, we already don't have consistency there. (Backwards compatibility is kind of an edge case here because I doubt anyone calls `KedroSession.run` with positional arguments, but technically what you've done is right.)

Going beyond this PR, it would be nice to have eventual consistency between the Python API and click `run` command. More importantly than the order is the naming of the arguments. This is something I've noticed before and it's on our radar here: https://github.com/kedro-org/kedro/issues/2247. Note the `pipeline_name` vs. `pipeline` question follows the same pattern as the `node_name` vs. `node` one noted there.

In fact, I now have another thought on the ordering of arguments - will add it to that issue to keep it in one place.

AntonyMilneQB (approved):

AntonyMilneQB:
@noklam One thing I just thought of. We should add this to the documentation [here](https://kedro.readthedocs.io/en/stable/development/commands_reference.html#modifying-a-kedro-run). Maybe other places too, have a look. But that's the one that springs to mind.

noklam:
Thank you @AntonyMilneQB, I will update the docs🙂

tests/framework/cli/test_cli.py
```diff
@@ -562,9 +565,13 @@ def test_run_with_pipeline_filters(
     ):
         from_nodes = ["--from-nodes", "splitting_data"]
         to_nodes = ["--to-nodes", "training_model"]
+        tags = ["--tag", "de"]
+        namespace = ["--namespace", "fake_namespace"]
         tags = ["--tags", "de"]
```
AhdraMeraliQB, line 570: Looks like `tags` is being overwritten here 🤔

kedro/framework/session/session.py
AhdraMeraliQB: Nice!

AhdraMeraliQB (approved):

Issue #2168, opened 2023-01-03
Create a `kedro jupyter-init` command to setup a Jupyter kernel for Kedro


idanov:
Design a `kedro jupyter-init` command (name is up for debate) which will create a Jupyter Kernel to automatically load the Kedro extension
- Use [the jupyter client kernelspec API](https://jupyter-client.readthedocs.io/en/stable/api/kernelspec.html#jupyter_client.kernelspec.KernelSpecManager.install_kernel_spec) to do that from Python
- Make the command available only if `jupyter` is installed, else show a warning Jupyter is not installed
- Add the Kedro icon to the KernelSpec
- Use the work done here: https://github.com/kedro-org/kedro/pull/1355/files#r830876411

AntonyMilneQB:
This should be very easy to do since all the relevant code is in `_create_kernel` so we just need to create an appropriate CLI command that wraps it.

I mentioned the idea of this to @BenjaminLevyQB a while ago and he seemed to like it. Note that it seems to be possible e.g. on databricks to create (and then select from the notebook) a Jupyter kernel even though you wouldn't execute `kedro jupyter notebook/lab` itself there, so I think having a separate command to just do the setup here makes sense.

So long as we're keeping the `kedro jupyter lab/notebook` commands let's make sure the command is part of that group, e.g. `kedro jupyter init` or `kedro jupyter setup`.

jmholzer:
Discussed in backlog grooming 30/01/23:

1. Potentially modify name of command to e.g. `setup` or `kernel`.
2. Command to be made part of `jupyter` group, not a 'root' command.

PR #2353, opened 2023-02-22
Create a new Kedro CLI command `kedro jupyter setup` to setup Jupyter kernel for Kedro

noklam:
## Description

> Design a `kedro jupyter-init` command (name is up for debate) which will create a Jupyter Kernel to automatically load the Kedro extension
- Use [the jupyter client kernelspec API](https://jupyter-client.readthedocs.io/en/stable/api/kernelspec.html#jupyter_client.kernelspec.KernelSpecManager.install_kernel_spec) to do that from Python
- Make the command available only if `jupyter` is installed, else show a warning Jupyter is not installed
- Add the Kedro icon to the KernelSpec
- Use the work done here: https://github.com/kedro-org/kedro/pull/1355/files#r830876411
## Development notes

* New command `kedro jupyter init`

More changes:
* By default `click` sort the commands alphabetically, the PR change it so deprecated command will be moved to the bottom of the command group.

# Screenshot
image

## Checklist

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/2353"></a>


kedro/framework/cli/jupyter.py
```diff
@@ -31,20 +32,54 @@
...
+
+    def __init__(self, name=None, commands=None, **attrs):
+        super().__init__(name, commands, **attrs)
+        #: the registered subcommands by their exported names.
+        self.commands = commands or {}
+
+    def list_commands(self, ctx):
+        """List commands according to a custom order"""
+        return sorted(self.commands, key=lambda k: jupyter_command_group_order[k])
+
```
noklam, line 50: This allows us to inject a custom order. Initially I think about adding something like a `@deprecated` flag and which will automatically lower the rank of the commands.

However, I found that this is not worth the effort. Looking at the `kedro` command itself we may need to customize the order (It's alphabetically sorted now). 

Instead of having Customised Order + deprecated command move to the end, a simple list of order can be enough.

features/jupyter.feature
```diff
@@ -4,6 +4,10 @@ Feature: Jupyter targets in new project
     Given I have prepared a config file
     And I have run a non-interactive kedro new with starter "default"
 
+  Scenario: Execute jupyter-init target
+    When I execute the kedro jupyter command "init"
+    Then I should get a message including "The kernel has been created successfully at"
+
   Scenario: Execute jupyter-notebook target
     When I execute the kedro jupyter command "notebook --no-browser"
```
noklam, line 12: `kedro jupyter` is not an accurate descrption anymore, since `kedro init` isn't using the `jupyter` command. There is another ticket so I put it here as a sidenote.

kedro/framework/cli/jupyter.py
merelcht: I really like this!!

merelcht (approved):
This looks great! I've tried it and it works perfectly 👍  I also like the re-ordering of the commands.

RELEASE.md
```diff
@@ -11,6 +11,7 @@
 # Upcoming Release 0.18.6
 
 ## Major features and improvements
+* Added new Kedro CLI `kedro jupyter-init` to setup Jupyter Kernel for Kedro.
```
AntonyMilneQB, line 14:
```suggestion
* Added new CLI command `kedro jupyter init` to setup a Jupyter kernel for Kedro.
```

features/jupyter.feature
```diff
@@ -4,6 +4,10 @@ Feature: Jupyter targets in new project
     Given I have prepared a config file
     And I have run a non-interactive kedro new with starter "default"
 
+  Scenario: Execute jupyter-init target
```
AntonyMilneQB, line 7:
```suggestion
  Scenario: Execute jupyter init target
```

AntonyMilneQB: Nitpick, but the other tests in this file are written in a similar weird way and should probably be changed `jupyter-notebook` ➡️ `jupyter notebook` etc.

kedro/framework/cli/jupyter.py
```diff
@@ -31,20 +31,56 @@
...
-@jupyter_cli.group()
+@jupyter_cli.group(cls=JupyterCommandGroup)
 def jupyter():
     """Open Jupyter Notebook / Lab with project specific variables loaded, or
     convert notebooks into Kedro code.
     """
 
 
+@forward_command(jupyter, "init", forward_help=True)
+@env_option
```
AntonyMilneQB, line 67: Do we actually want `@env_option` here? I don't think so. `env` doesn't actually do anything here?

kedro/framework/cli/jupyter.py
```diff
@@ -31,20 +31,56 @@
...
+@click.pass_obj  # this will pass the metadata as first argument
+def init(
+    metadata: ProjectMetadata,
+    env,
+    args,
+    **kwargs,
```
AntonyMilneQB, line 73: Are these used anywhere? I think they could probably be removed.

kedro/framework/cli/jupyter.py
```diff
@@ -31,20 +31,56 @@
...
+):  # pylint: disable=unused-argument
+    """Initialise the Jupyter Kernel for a kedro project."""
+    _check_module_importable("notebook")
```
AntonyMilneQB, line 76: It probably won't make any difference in practice, since anyone executing this command will probably have it installed... but I don't think `notebook` is actually necessary for making a Jupyter kernel. Looking at `_create_kernel` we should probably be checking `ipykernel` instead.

kedro/framework/cli/jupyter.py
```diff
@@ -31,20 +31,56 @@
...
+    validate_settings()
+
+    kernel_name = f"kedro_{metadata.package_name}"
+    kernel_path = _create_kernel(kernel_name, f"Kedro ({metadata.package_name})")
+    click.secho(f"\nThe kernel has been created successfully at {kernel_path}")
```
AntonyMilneQB, line 81: I haven't tried running this, but from memory doesn't the ipykernel code in `_create_kernel` already emit a message saying this? But if your new message makes it clearer then great 👍

kedro/framework/cli/jupyter.py
```diff
@@ -31,20 +31,56 @@
 OVERWRITE_HELP = """If Python file already exists for the equivalent notebook,
 overwrite its contents."""
 
+jupyter_command_group_order = ["init", "notebook", "lab", "convert"]
+jupyter_command_group_order = {
+    command: order for order, command in enumerate(jupyter_command_group_order)
+}
+
+
+class JupyterCommandGroup(click.Group):
```
AntonyMilneQB, line 40: Love the idea of doing this 💯 But why doesn't `list_commands` just do `return ["init", "notebook", "lab", "convert"]`? I get that `self.commands` is unordered, but then `list_commands` is surely just returning the original ordered list here? 🤔 

Not a big thing, but I feel like there's a more direct way of getting the correctly ordered commands here.

AntonyMilneQB (commented):
Lots of small comments but basically LGTM 👍 

Just two additional things that might need doing:
* do we prefer `jupyter init` compared to `jupyter setup` or `jupyter kernel` mentioned [here](https://github.com/kedro-org/kedro/issues/2168#issuecomment-1408748512)? No strong feelings from me really but seems like people might have preferred the alternatives when this was last discussed
* would be nice to make a note of this in the docs, maybe [here](https://kedro.readthedocs.io/en/stable/notebooks_and_ipython/kedro_and_notebooks.html#useful-to-know). e.g. on a managed service like Databricks it makes more sense to run `kedro jupyter init` and then select the kernel from the databricks notebook rather than running `kedro jupyter notebook` and ignoring the output like we currently recommend

kedro/framework/cli/jupyter.py
noklam: Updated the description, it doesn't seem to printing anything from `_create_kernel`

kedro/framework/cli/jupyter.py
noklam: (Y) This actually works. As what I get from `self.commands` is `Dict[str|Commands]` . I didn't notice it actually works with a list of string and it would resolve the commands. Brilliant.

RELEASE.md
noklam: I totally missed the comments about the command names, I might slightly prefer `setup`.

noklam: > would be nice to make a note of this in the docs, maybe [here](https://kedro.readthedocs.io/en/stable/notebooks_and_ipython/kedro_and_notebooks.html#useful-to-know). e.g. on a managed service like Databricks you can run kedro jupyter init and then select the kernel from the databricks notebook rather than needing to explicitly run %load_ext kedro.ipython

@AntonyMilneQB Can we actually do this? Can you select a custom Jupyter kernel on Databricks? It could still work for managed JupyterHub though.

astrojuanlu:
After I `pip install .` on this branch (and nothing else) I did a quick test, and noticed that

- `kedro jupyter --help` doesn't show `setup`:

```
juan_cano@M-PH9T4K3P3C /t/jupyter-test> kedro jupyter --help                                                                (kedro310-dev-test-kernel) 
Usage: kedro jupyter [OPTIONS] COMMAND [ARGS]...

  Open Jupyter Notebook / Lab with project specific variables loaded, or
  convert notebooks into Kedro code.

Options:
  -h, --help  Show this message and exit.

Commands:
  notebook  Open Jupyter Notebook with project specific variables loaded.
  lab       Open Jupyter Lab with project specific variables loaded.
  convert   Convert selected or all notebooks found in a Kedro project to...
```

- `kedro jupyter setup` gave me an error:

```
juan_cano@M-PH9T4K3P3C /t/jupyter-test> kedro jupyter setup                                                                 (kedro310-dev-test-kernel) 
│ /Users/juan_cano/.micromamba/envs/kedro310-dev-test-kernel/bin/kedro:8 in <module>               │
│ /Users/juan_cano/.micromamba/envs/kedro310-dev-test-kernel/lib/python3.10/site-packages/kedro/fr │
│ amework/cli/cli.py:211 in main                                                                   │
│   208 │   """                                                                                    │
│   209 │   _init_plugins()                                                                        │
│   210 │   cli_collection = KedroCLI(project_path=Path.cwd())                                     │
│ ❱ 211 │   cli_collection()                                                                       │
│   212                                                                                            │
│ /Users/juan_cano/.micromamba/envs/kedro310-dev-test-kernel/lib/python3.10/site-packages/click/co │
│ re.py:1130 in __call__                                                                           │
│ /Users/juan_cano/.micromamba/envs/kedro310-dev-test-kernel/lib/python3.10/site-packages/kedro/fr │
│ amework/cli/cli.py:139 in main                                                                   │
│   136 │   │   )                                                                                  │
│   137 │   │                                                                                      │
│   138 │   │   try:                                                                               │
│ ❱ 139 │   │   │   super().main(                                                                  │
│   140 │   │   │   │   args=args,                                                                 │
│   141 │   │   │   │   prog_name=prog_name,                                                       │
│   142 │   │   │   │   complete_var=complete_var,                                                 │
│ /Users/juan_cano/.micromamba/envs/kedro310-dev-test-kernel/lib/python3.10/site-packages/click/co │
│ re.py:1055 in main                                                                               │
│ /Users/juan_cano/.micromamba/envs/kedro310-dev-test-kernel/lib/python3.10/site-packages/click/co │
│ re.py:1657 in invoke                                                                             │
│ /Users/juan_cano/.micromamba/envs/kedro310-dev-test-kernel/lib/python3.10/site-packages/click/co │
│ re.py:1657 in invoke                                                                             │
│ /Users/juan_cano/.micromamba/envs/kedro310-dev-test-kernel/lib/python3.10/site-packages/click/co │
│ re.py:1404 in invoke                                                                             │
│ /Users/juan_cano/.micromamba/envs/kedro310-dev-test-kernel/lib/python3.10/site-packages/click/co │
│ re.py:760 in invoke                                                                              │
│ /Users/juan_cano/.micromamba/envs/kedro310-dev-test-kernel/lib/python3.10/site-packages/click/de │
│ corators.py:38 in new_func                                                                       │
TypeError: setup() got an unexpected keyword argument 'args'
```

Am I doing something wrong?

noklam:
@astrojuanlu I am checking, I think it may be a bug when I rename the CLI, as the tests are correctly catching this error too.

kedro/framework/cli/jupyter.py
noklam: It's argument forward from kedro's cli and it has to be there even when not used.

astrojuanlu:
It now works 👍🏽

RELEASE.md
AntonyMilneQB: Hmm, good question. I thought you could select a custom Jupyter kernel on databricks but I might just be remembering wrongly. So yeah, we should check before recommending that... Maybe I was thinking of Sagemaker instead? I'm sure you can select custom kernel there.

kedro/framework/cli/jupyter.py
AntonyMilneQB: Excellent! Do we even need to override `__init__` now or just `list_commands`?

kedro/framework/cli/jupyter.py
```diff
@@ -32,19 +32,44 @@
...
 
 
+class JupyterCommandGroup(click.Group):
+    """A custom class for ordering the `kedro jupyter` command groups"""
+
+    def __init__(self, name=None, commands=None, **attrs):
+        super().__init__(name, commands, **attrs)
+        #: the registered subcommands by their exported names.
+        self.commands = commands or {}
+
```
AntonyMilneQB, line 42:
```suggestion
```

Maybe?

noklam:
@AntonyMilneQB 

On SageMaker Studio there is an option to select Kernel as it is build on top of JupyterLab.
image

However, it doesn't work as expected. The kernel is just hanging.

image

It's possible to have a custom Kernel but it uses LifeCycle Configuration Script. I have a quick glance it should work similarly as it used this line to set up the kernel.

`python -m ipykernel install --user --name "$BASENAME" --display-name "Custom ($BASENAME)"`

https://docs.aws.amazon.com/sagemaker/latest/dg/nbi-add-external.html.

I am opening this as a separate issue as I think it would be quite challenging to solve it across different platforms (i.e. Databricks/SageMaker/AzureML). But until then the command may not be too useful.

Issue #2351, opened 2023-02-21
Mermaid diagram for deployment guide isn't rendering


stichbury:
## Description
In 0.18.1 we had a graphic at [the bottom of the deployment guide](https://kedro.readthedocs.io/en/0.18.1/deployment/deployment_guide.html#deployment-choices)

In 0.18.2 it looks like this was replaced by a mermaid graphic ([Rendered the deployment guide flowchart as a Mermaid diagram, and added Dask.](https://github.com/kedro-org/kedro/releases/tag/0.18.2))

I don't know if this has ever worked on RTD though, since the same page for 0.18.2 isn't showing a rendered graphic (to me, at least, on latest Chrome, latest M1 MacOS) https://kedro.readthedocs.io/en/0.18.2/deployment/deployment_guide.html#deployment-choices

@deepyaman I think this was your initiative originally (https://github.com/kedro-org/kedro/pull/1392) but I'm not heaping it on you to solve! I just wanted to get your perspective on what you saw on kedro.readthedocs.io when that release went out (did it work back then?) -- if you can recall :)

stichbury:
Thanks to @ankatiyar for the spot -- it's been out for 6 months without anyone reporting AFAICT, so thanks for bringing it to my attention!

deepyaman:
It definitely was working for the most part; I remember somebody--maybe @merelcht?--pointing out that it didn't render when there was highlight on the page, but it worked if you removed that.

I see an error in the console now saying:

> Refused to execute https://unpkg.com/mermaid/dist/mermaid.min.js as script because "X-Content-Type-Options: nosniff" was given and its Content-Type is not a script MIME type.

I haven't looked into it, but seems like somebody pointed out a similar issue (https://github.com/just-the-docs/just-the-docs/pull/857#issuecomment-1198404402) that should be addressable.

As a side note, may also be worth trying out https://github.com/mgaitan/sphinxcontrib-mermaid/pull/101, as pointed out by @AntonyMilneQB, although I don't expect it'll fix anything.

stichbury:
Thanks @deepyaman that's very helpful. I wonder what changed to break this and whether we want to consider whether to fix it or remove for now, if this is something that changes beyond our control.

deepyaman:
> Thanks @deepyaman that's very helpful. I wonder what changed to break this and whether we want to consider whether to fix it or remove for now, if this is something that changes beyond our control.

[...]

Issue #1772, opened 2022-08-09
Mermaid Diagram disappeared in Doc when search with keywords


noklam:
## Description
Short description of the problem here.

When searching keywords in Kedro's ReadTheDoc, It will highlight the keywords and generated URL like this:
https://kedro.readthedocs.io/en/stable/deployment/deployment_guide.html?highlight=deployment

The Mermaid Diagram is not rendered properly, sometimes it is rendered as plain text or an empty diagram.
https://kedro.readthedocs.io/en/stable/deployment/deployment_guide.html

Removing the `highlight` will make it work again.

CC @deepyaman 

# Diagnosis
Initial investigation suggests that this is due to an additional `<span class='highlighted'> </span> HTML tag and probably causing it fail to render.

![image](https://user-images.githubusercontent.com/18221871/183655192-07741d09-3426-4c2b-95a8-558d7109c3d9.png)

## Context
How has this bug affected you? What were you trying to accomplish?

## Steps to Reproduce
1. [First Step]
2. [Second Step]
3. [And so on...]

## Expected Result
Tell us what should happen.

## Actual Result
Tell us what happens instead.

```
-- If you received an error, place it here.
```

```
-- Separate them if you have more than one.
```

## Your Environment
Include as many relevant details about the environment in which you experienced the bug:

* Kedro version used (`pip show kedro` or `kedro -V`):
* Python version used (`python -V`):
* Operating system and version:

AntonyMilneQB:
Just to be clear, I believe this is a problem only when the term to be highlighted is in the diagram. 

e.g. https://kedro.readthedocs.io/en/stable/deployment/deployment_guide.html?highlight=asdfsadf works fine 
https://kedro.readthedocs.io/en/stable/deployment/deployment_guide.html?highlight=container does not

Unfortunately I think it might not be easy to fix...

MerelTheisenQB:
Timebox this to max 2 days, if not fixable, report to read the docs.

astrojuanlu:
Don't think this is an issue anymore?

![image](https://user-images.githubusercontent.com/316517/221842333-0486cc83-10fe-4a41-ab05-23380f70540c.png)

(https://kedro.readthedocs.io/en/latest/deployment/deployment_guide.html?highlight=container)

merelcht:
I'm seeing this.. 

[...]

PR #2362, opened 2023-02-24
Fix mermaid diagrams in documentation

astrojuanlu:
## Description
gh-2351

## Development notes
Fixed the Mermaid JS URL.

Also, I removed the PDF documentation, because it's making build longers at the cost of producing a 730-page document that I bet almost nobody is reading... but I could be wrong here, happy to rollback.

## Checklist

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -24,10 +24,6 @@ sphinx:
 # mkdocs:
 # configuration: mkdocs.yml
 
-# Optionally build your docs in additional formats such as PDF and ePub
-formats:
-  - pdf
-
 # Optionally set the version of Python and requirements required to build your docs
 python:
   install:
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -29,7 +29,7 @@ from sphinxcontrib.mermaid import mermaid
 
 from kedro import __version__ as release
 
-MERMAID_JS_URL = "https://unpkg.com/mermaid/dist/mermaid.min.js"
+MERMAID_JS_URL = "https://unpkg.com/mermaid@9.4.0/dist/mermaid.min.js"
 
 # -- Project information -----------------------------------------------------
 


astrojuanlu:
Welp, now the build finished without warnings :+1: and the `mmdc` error message is not shown :+1: but apparently the diagram still does not show :-1:

https://kedro--2362.org.readthedocs.build/en/2362/deployment/deployment_guide.html#deployment-choices

Wanted to see if giving this a quick stab would do the trick, but it didn't. Turning this PR to a draft, will revisit later.

stichbury:
Thanks for trying @astrojuanlu -- I wonder if the formatting of the diagram is still valid -- I've seen some examples where there's a blank line between the `{mermaid}` specifier and the start of the text. I doubt it's anything that simple, but sometimes with Sphinx, details like an extra single space seem to be enough to break things. Anyway, it's appreciated that you've tried to resolve it 🙏

astrojuanlu:
I tried locally and was still unable to produce the diagrams, but it turns out fixing the Mermaid JS URL did the trick. Let's see if that works on RtD, and if so I'll unroll the commits that are not necessary so we have a minimal set of changes.

astrojuanlu:
🎉

AntonyMilneQB (approved):
Amazing work @astrojuanlu, thank you for the fix.

I don't know much about how unpkg.com works, but is this the right approach to pin to version `9.4.0`? I had assumed before that pointing to the generic (unversioned) address would be good since then we automatically get the latest version, but that's obviously not worked well here...

If you're in the mood for battling sphinx/mermaid, you might like to take a look at https://github.com/kedro-org/kedro/issues/1772 also, which I think might be very tricky to fix 😬

stichbury (approved):

stichbury:
You absolute ⭐  Thank you. It's not wonderful that our docs were effectively broken by something that changed in a newer version, but thankfully the pin is working.

astrojuanlu:
Thanks for the review folks! Mermaid 10 does not ship a minified JS anymore and instead requires a "module". I'm not versed enough in frontend to give it a quick stab but it shouldn't be impossible either. See: https://github.com/mgaitan/sphinxcontrib-mermaid/issues/110

I say, let's merge this one to fix the docs, and I'll give this another pass if I tackle gh-1772

PR #2378, opened 2023-03-01
Fix overwriting of nested parameters in config by runtime parameters

### LINKED ISSUE #2357: [BUG] Passing runtime nested params to `kedro run --params` overwrites params from config files in latest release
Hello team - nice to be back in these discussions :)

## Description
There was a regression in the latest release (0.18.5), which introduced `OmegaConf` for parsing parameters provided to the `kedro run --params`.

More specifically, in previous versions

```kedro run --params foo.bar=a```

was properly merged with an existing parameter ```{"foo": {"baz": 1}}```.

Now, `kedro.framework.cli.hooks.utils._split_params` returns an `omegaconf.DictConfig` object (rather than a plain dictionary), and the second part of the if clause [here](https://github.com/kedro-org/kedro/blob/45c345d5a8109938e2735c1056a62182eb85d2f3/kedro/framework/context/context.py#L157) returns false:

```python
if isinstance(old_dict[key], dict) and isinstance(value, dict):
    _update_nested_dict(old_dict[key], value)
```

Which means that after merging we get 
```
{"foo": {"bar": a}}
```

instead of the expected

```
{"foo": {"bar": a, "baz": 1}}
```

I could open a PR to fix but it ll take way longer that letting you fix it :)
In the meantime I 've got a local workaround by subclassing `KedroContext` and casting params to a dict!

ankatiyar:
## Description
Resolves #2357 

#### Nested params
If the config files already contained a nested parameter -
```yaml
param1 : 5
foo : 
  bar : 1
```
And you do `kedro run --params foo.baz=2`, this would completely overwrite the `foo` parameter at runtime. 
The parameters at runtime are then - 
```yaml
param1 : 5
foo: 
   baz : 2
```
Instead of 
```yaml
param1 : 5
foo: 
   bar : 1
   baz : 2
```
#### Kedro viz bug
The runtime params being passed as `DictConfig` instead of a python dictionary is also causing problems with Kedro Viz because it's not json-serialisable. This is breaking experiment tracking.

## Development notes

* This is because `omegaconf` loads the runtime parameters as a `DictConfig` object instead of a `dict`. In case of nested dicts, it also loads the nested parameter as a `DictConfig` object which was causing the conditions in the changed lines to return `False`.
* Manually tested this solution works with both `ConfigLoader` as well as the `OmegaConfigLoader`.
* when the `OmegaConfigLoader` is used instead of `ConfigLoader`, it also loads the nested params as `DictConfig`. `DictConfig` key-value pairs can be accessed in a similar way to `dict` so I think this solution should suffice. 

#### Update :
* We decided to convert runtime params from `DictConfig` to `dict` right when they're read. -> This solves the Kedro-viz issue
* Didn't change the way `OmegaConfigLoader` reads parameters from `conf` for now, while merging allow both `DictConfig` and `dict` types. -> Solves the nested params issue

## Checklist

### DIFF
--- a/kedro/config/omegaconf_config.py
+++ b/kedro/config/omegaconf_config.py
@@ -250,7 +250,7 @@ class OmegaConfigLoader(AbstractConfigLoader):
             return {}
         if len(aggregate_config) == 1:
             return list(aggregate_config)[0]
-        return dict(OmegaConf.merge(*aggregate_config))
+        return OmegaConf.to_container(OmegaConf.merge(*aggregate_config))
 
     @staticmethod
     def _is_valid_config_path(path):
--- a/kedro/config/omegaconf_config.py
+++ b/kedro/config/omegaconf_config.py
@@ -250,7 +250,7 @@ class OmegaConfigLoader(AbstractConfigLoader):
             return {}
         if len(aggregate_config) == 1:
             return list(aggregate_config)[0]
-        return OmegaConf.to_container(OmegaConf.merge(*aggregate_config))
+        return dict(OmegaConf.merge(*aggregate_config))
 
     @staticmethod
     def _is_valid_config_path(path):
--- a/kedro/framework/context/context.py
+++ b/kedro/framework/context/context.py
@@ -7,6 +7,7 @@ from typing import Any, Dict, Optional, Union
 from urllib.parse import urlparse
 from warnings import warn
 
+from omegaconf import DictConfig
 from pluggy import PluginManager
 
 from kedro.config import ConfigLoader, MissingConfigException
@@ -154,7 +155,9 @@ def _update_nested_dict(old_dict: Dict[Any, Any], new_dict: Dict[Any, Any]) -> N
         if key not in old_dict:
             old_dict[key] = value
         else:
-            if isinstance(old_dict[key], dict) and isinstance(value, dict):
+            if isinstance(old_dict[key], (dict, DictConfig)) and isinstance(
+                value, (dict, DictConfig)
+            ):
                 _update_nested_dict(old_dict[key], value)
             else:
                 old_dict[key] = value
@@ -322,8 +325,7 @@ class KedroContext:
             """
             key = f"params:{param_name}"
             feed_dict[key] = param_value
-
-            if isinstance(param_value, dict):
+            if isinstance(param_value, (dict, DictConfig)):
                 for key, val in param_value.items():
                     _add_param_to_feed_dict(f"{param_name}.{key}", val)
 
--- a/kedro/framework/context/context.py
+++ b/kedro/framework/context/context.py
@@ -7,7 +7,6 @@ from typing import Any, Dict, Optional, Union
 from urllib.parse import urlparse
 from warnings import warn
 
-from omegaconf import DictConfig
 from pluggy import PluginManager
 
 from kedro.config import ConfigLoader, MissingConfigException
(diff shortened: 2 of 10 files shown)


kedro/framework/context/context.py
```diff
@@ -154,7 +155,9 @@ def _update_nested_dict(old_dict: Dict[Any, Any], new_dict: Dict[Any, Any]) -> N
         if key not in old_dict:
             old_dict[key] = value
         else:
-            if isinstance(old_dict[key], dict) and isinstance(value, dict):
+            if isinstance(old_dict[key], (dict, DictConfig)) and isinstance(
```
tsanikgr, line 158: Maybe a better way would be to check against the superclass of all dict-like objects to remove the strict dependency on `omegaconf` and make this future-proof.

```python
from collections.abc import Mapping
from omegaconf import DictConfig

assert isinstance({}, Mapping) 
assert isinstance(DictConfig({1:1}), Mapping)
```

I would also double check for other occurrences of `isinstance(x, dict)` in the code

tsanikgr (commented):
Thank you for the PR :) A suggestion below!

kedro/config/omegaconf_config.py
```diff
@@ -250,7 +250,7 @@ def load_and_merge_dir_config(
             return {}
         if len(aggregate_config) == 1:
             return list(aggregate_config)[0]
-        return dict(OmegaConf.merge(*aggregate_config))
+        return OmegaConf.to_container(OmegaConf.merge(*aggregate_config))
```
ankatiyar, line 253: going to revert this.

rashidakanchwala (approved):
I tested it and it works on viz. thanks

tests/framework/session/test_session.py
```diff
@@ -920,3 +923,25 @@ def test_setup_logging_using_omega_config_loader_class(
     ).as_posix()
     actual_log_filepath = call_args["handlers"]["info_file_handler"]["filename"]
     assert actual_log_filepath == expected_log_filepath
+
+
+@pytest.mark.parametrize("params", ["a=1,b.c=2", "a=1,b=2,c=3"])
```
AntonyMilneQB, line 928:
```suggestion
@pytest.mark.parametrize("params", ["a=1,b.c=2", "a=1,b=2,c=3", ""])
```
Would be good to test the case that no params are passed here also - does adding it as a test case work?

ankatiyar: Yeah this would work, added it to the test!

tests/framework/session/test_session.py
```diff
@@ -920,3 +923,25 @@ def test_setup_logging_using_omega_config_loader_class(
...
+    params,
+    mock_package_name,
+    fake_project,
+):
+    extra_params = _split_params(None, None, params)
+    session = KedroSession.create(
+        mock_package_name, fake_project, extra_params=extra_params
+    )
+
+    def get_all_values(mapping: Mapping):
```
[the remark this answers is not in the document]
ankatiyar: Done!

kedro/framework/context/context.py
ankatiyar: This is creating trouble with `mypy` in `_update_nested_dict`. 😅

ankatiyar:
> Just to check that the tests are working as they should: do both `test_no_DictConfig_in_store` and the new cases for `test_update_nested_dict` fail on `main` currently?

Thanks Antony! Yeah, I checked and both the tests are currently failing on `main` currently without this fix.

Issue #2323, opened 2023-02-15
Add documentation for Datasets stored on Databricks DBFS


jmholzer:
## Description
As part of the milestone to improve the Databricks user experience, we should add documentation regarding how to use datasets stored via DBFS.

## Context
Currently, there is no documentation informing users how they should handle Datasets stored and accessed via DBFS. Incorrectly specifying file paths for these datasets causes Kedro to fail to load them (as in https://github.com/kedro-org/kedro/issues/1801 and  https://github.com/kedro-org/kedro/issues/2272).

PR #2385, opened 2023-03-02
DBFS documentation

SajidAlamQB:
## Description

Resolves: https://github.com/kedro-org/kedro/issues/2323

As part of the milestone to improve the Databricks user experience, we should add documentation regarding how datasets are stored via DBFS and why it's useful.

## Checklist


docs/source/deployment/databricks.md
```diff
@@ -259,3 +259,7 @@ Kedro-Viz can then be launched in a new browser tab with the `%run_viz` line mag
 ```ipython
 In [2]: %run_viz
 ```
+
+### How to use Datasets stored on Databricks DBFS
+
+DBFS is a distributed file system mounted into a DataBricks workspace and accessible on a DataBricks cluster. It maps cloud object storage URIs to relative paths simplifying the process of persisting files. This is useful for libraries to read/write distributed storage as if it's a local file. Some of our datasets like `SparkDataSet` expect the filepath in DBFS to be prefixed with `/dbfs` this is because DataBricks will resolve any filepath given to PySpark in DBFS. However, when loading datasets using Pandas, DataBricks will resolve the filepath in the block storage volume attached to the driver node. This will result in a `DataSetError`, as the code will try to load files from the DBFS save location which doesn't exist in the local filesystem.
```
stichbury, line 265:
```suggestion
DBFS is a distributed file system mounted into a DataBricks workspace and accessible on a DataBricks cluster. It maps cloud object storage URIs to relative paths so as to simplify the process of persisting files. With DBFS, libraries can read from or write to distributed storage as if it's a local file. 

Some of Kedro's datasets like `SparkDataSet` expect the file path in DBFS to be prefixed with `/dbfs` because DataBricks resolves any path given to PySpark in DBFS. **However, Kedro's Pandas datasets do not work in the same way** and will result in a `DataSetError`, as the code will try to load files from a DBFS save location which doesn't exist in the local filesystem.
```

stichbury: @SajidAlamQB I've made a few edits to simplify the text a bit as I don't think the reader needs all the information about the driver node/block storage volume for use with a pandas dataset. But feel free to revert/ignore my suggestion if you think they do need that info.

stichbury: Also, I didn't really feel like this was sufficient because you've titled the section "How to use..." but all it contains is a warning about an error. Don't you need to show how to use the datasets stored on DBFS so you don't get the error? That is, either present the code needed or a warning to say that you just cannot use Pandas because of the fact it's local.

SajidAlamQB: Good point maybe making this a `note` would be better? The main point we wanted to get across was that `SparkDataSet` expects file paths prefixed with `/dbfs`.

stichbury: Oh, that's interesting because that didn't come across at all to me, sorry. I would definitely revise then...Let me suggest again...

stichbury (approved):
Nice work!

docs/source/deployment/databricks.md
```diff
@@ -259,3 +259,9 @@ Kedro-Viz can then be launched in a new browser tab with the `%run_viz` line mag
 ```ipython
 In [2]: %run_viz
 ```
+
+### How to use Datasets stored on Databricks DBFS
+
+DBFS is a distributed file system mounted into a DataBricks workspace and accessible on a DataBricks cluster. It maps cloud object storage URIs to relative paths so as to simplify the process of persisting files. With DBFS, libraries can read from or write to distributed storage as if it's a local file.
```
merelcht, line 265:
```suggestion
DBFS is a distributed file system mounted into a Databricks workspace and accessible on a Databricks cluster. It maps cloud object storage URIs to relative paths so as to simplify the process of persisting files. With DBFS, libraries can read from or write to distributed storage as if it's a local file.
```

docs/source/deployment/databricks.md
```diff
@@ -259,3 +259,9 @@ Kedro-Viz can then be launched in a new browser tab with the `%run_viz` line mag
...
+To use datasets like `SparkDataSet` with DBFS, the file path must be prefixed with `/dbfs`.
+> **Note**: Datasets that use the local filesystem, such as Kedro's Pandas datasets, will return a `DataSetError` if the code tries to load files from a remote DBFS save location. Do not use the `/dbfs` prefix for these datasets.
```
merelcht, line 267: Does this also apply if the dataset is stored in the cloud like S3? I'm just a bit confused about how you can know which datasets need `dbfs` and which don't?

jmholzer: Only `SparkDataSet` requires the prefix, because of the unique way it resolves its file path. No other data sets work in this way. I have opened a [ticket](https://github.com/kedro-org/kedro-plugins/issues/117) to fix this.

docs/source/deployment/databricks.md
```diff
@@ -263,3 +263,9 @@ Kedro-Viz can then be launched in a new browser tab with the `%run_viz` line mag
...
```
jmholzer, line 271: At the moment, I don't think this is accurate. All Python code (apart from PySpark) will try to resolve a file path in driver node storage by default, but it can resolve file paths on DBFS if the file path points to it. Therefore, adding the `/dbfs` prefix is not only valid but sometimes what the user needs to do.

The rules for whether code will look in DBFS or in driver node storage by default are described [here](https://docs.databricks.com/files/index.html). We should link to this documentation in this section also.

docs/source/deployment/databricks.md
```diff
@@ -263,3 +263,9 @@ Kedro-Viz can then be launched in a new browser tab with the `%run_viz` line mag
 ```ipython
 In [2]: %run_viz
 ```
+
+### How to use Datasets stored on Databricks DBFS
+
+DBFS is a distributed file system mounted into a DataBricks workspace and accessible on a DataBricks cluster. It maps cloud object storage URIs to relative paths so as to simplify the process of persisting files. With DBFS, libraries can read from or write to distributed storage as if it's a local file.
+To use datasets like `SparkDataSet` with DBFS, the file path must be prefixed with `/dbfs`.
```
jmholzer, line 270: See my comment below, if users want to use DBFS, they should explicitly add the prefix to every file path.

There is a subtlety here, which is that `PySpark` code will natively look in DBFS by default, though as we established in https://github.com/kedro-org/kedro-plugins/pull/114, even if the user wants to use `PySpark` via `SparkDataSet` (which *should* work natively, but doesn't because of the way we coded this data set), they should still specify the `/dbfs` prefix.

```suggestion
To use datasets with DBFS, the file path passed to the dataset **must** be prefixed with `/dbfs/`. This applies to all datasets, including `SparkDataSet`.
```

jmholzer, line 270: Minor, but could you also give an example of a correct DBFS file path here?

jmholzer (changes requested):
Good start! I have left some comments 🙂.

docs/source/deployment/databricks.md
```diff
@@ -263,3 +263,9 @@ Kedro-Viz can then be launched in a new browser tab with the `%run_viz` line mag
 ```ipython
 In [2]: %run_viz
 ```
+
+### How to use Datasets stored on Databricks DBFS
```
astrojuanlu, line 267: A small thing, but this header is on the same level as the 1..10 steps that come before, and I understand this is not necessarily tied to those. Should we make it a second level heading? cc @stichbury

stichbury: Yes, let's do that. I was in two minds, but it makes sense to make it H2.

docs/source/deployment/databricks.md
```diff
@@ -263,3 +263,9 @@ Kedro-Viz can then be launched in a new browser tab with the `%run_viz` line mag
 ```ipython
 In [2]: %run_viz
 ```
+
+## How to use datasets stored on Databricks DBFS
+
+DBFS is a distributed file system mounted into a DataBricks workspace and accessible on a DataBricks cluster. It maps cloud object storage URIs to relative paths so as to simplify the process of persisting files. With DBFS, libraries can read from or write to distributed storage as if it's a local file.
+To use datasets with DBFS, the file path passed to the dataset **must** be prefixed with `/dbfs/` and look something like, `/dbfs/example_project/data/02_intermediate/processed_data`. This applies to all datasets, including `SparkDataSet`.
+> **Note**: Most Python code, except PySpark, will try to resolve a file path in the driver node storage by default, this in turn could result in a `DataSetError` if the code tries to load files from a remote DBFS save location. To avoid this make sure to point the file path to `/dbfs`. For more rules on what is saved in DBFS versus driver node please refer to this [guide](https://docs.databricks.com/files/index.html#what-is-the-root-path-for-databricks).
```
jmholzer, line 271:
```suggestion
> **Note**: Most Python code, except PySpark, will try to resolve a file path in the driver node storage by default, this will result in an `DataSetError` if the code is using a file path that is actually a DBFS save location. To avoid this, always make sure to point the file path to `/dbfs` when storing or loading data on DBFS. For more rules on what is saved in DBFS versus driver node storage by default, please refer to the [Databricks documentation](https://docs.databricks.com/files/index.html#what-is-the-root-path-for-databricks).
```

jmholzer (approved):
Excellent, thanks for picking this up ⭐️.

I have some minor changes to suggest on the note, but otherwise looks good 🙂.

PR #2421, opened 2023-03-14
Docs/re structure config docs

merelcht:
## Development notes

- Added an introduction to configuration in Kedro: what do we mean by project configuration
- Added more headers. Don't know it this made it easier to follow or actually too nested 🤔 
- Moved the sections on `parameters` and `credentials` up to be before the config loader sections. The main reason is because those sections just describe how parameters and credentials can be used and that's pretty much the same for all config loaders.
- Moved the section about the `TemplatedConfigLoader` and `OmegaConfigLoader` under "Specify the configuration loader class"

I still feel like we can improve this more, but thought this was an okay first step. 

The way I see configuration is basically this:
You've got the configuration loader classes (default `ConfigLoader`, `TemplatedConfigLoader` and `OmegaConfigLoader`) and then you've got the things that these loaders load: `parameters`, `credentials`, `catalog`, `logging`. So I tried to divide it in a similar way where the first sections are about the default config functionality, then some specifics about parameters and credentials (logging and the data catalog are pretty complex and already have full chapters of their own) and then more details about the other config loaders. 

## Checklist


stichbury:
* I think I'd be tempted to split the page and have the Configuration basics on one, and the "How to" information about parameters, Credentials and the loader on another page. 

* Also, to that effect, I'd make the sections in that section page much more focussed on "How to..." so you'd have "How to load credentials", "How to use AWS credentials",  "How to specify a loader" kind of thing.

* I'd like a set of FAQs on the basics page too -- "Why use this?" "Can I load X?" with answers, and then an additional set of questions which you point through to the relevant information on the other page (point to the anchors of the relevant subsections)

* Also, I think putting some of the code snippets behind expandables would help

astrojuanlu:
Discussed synchronously between @merelcht, @stichbury and I:

- Let's try to avoid mixing explanations with how-to guides in the Diátaxis spirit https://diataxis.fr/
- Here "config" refers to "whatever is under `conf/`", therefore not `settings.py` and not `pyproject.toml` (they can be treated separately)
- From an Information Architecture perspective this can be tricky because "config" includes the catalog, but we have a separate set of pages for the catalog because it's central to using Kedro
- On the other hand, for the content here, the baseline is that "regular users don't need any of this", so let's make it clear at the top and maybe let's rename this whole page to "Advanced Configuration" (or try to separate the really typical use cases from the advanced ones)
- The examples are good, but we should reframe them in terms of use cases, e.g. "how do I pass Python objects to a catalog entry" rather than "how do I instantiate a new `TemplatedConfigLoader`"

docs/source/configuration/configuration_basics.md
```diff
@@ -0,0 +1,147 @@
...
+
+### How to change the default overriding environment?
+By default, `local` is the overriding environment for `base`. If you want to change this, you must customise the configuration loader argument settings you're using to set `default_run_env` in `src/<package_name>/settings.py` under the `CONFIG_LOADER_ARGS` key.
+For example, if you want to always override `base` configuration with configuration in a custom environment called `prod`, you change the configuration loader arguments like this:
+
+```python
+CONFIG_LOADER_ARGS = {"default_run_env": "prod"}
+```
+
+#### How to use only one configuration environment?
```
astrojuanlu, line 142:
```suggestion
### How to use only one configuration environment?
```

?

astrojuanlu (commented):
I really like the direction this is taking! 💯

docs/source/configuration/configuration_basics.md
```diff
@@ -0,0 +1,190 @@
...
+```text
+<conf_dir>
+├── base               <-- the files inside may be different, but this is an example of a standard Kedro structure.
+│   └── parameters.yml
+│   └── catalog.yml
+└── local              <-- the top level local folder is required, but no files should be inside when distributed.
+└── README.md          <-- optional but included with the default Kedro conf structure.
+```
+
+### How to directly access configuration
```
astrojuanlu, line 142:
```suggestion
### How to access configuration from your code
```

? Or "How to access configuration in code"

docs/source/configuration/parameters.md
```diff
@@ -0,0 +1,146 @@
...
+Since key-value pairs are split on the first colon or equals sign, values can contain colons/equals signs, but keys cannot. These are valid CLI commands:
+
+```bash
+kedro run --params=endpoint_url:https://endpoint.example.com
+```
+```bash
+kedro run --params=endpoint_url=https://endpoint.example.com
+```
+
+## How to use parameters
```
astrojuanlu, line 77: I started to read this page wondering "what are parameters, or what are there _for_?", and then this last how-to solved it for me. Maybe we could move it to the very beginning, or use part of this text as an explanation?

docs/source/configuration/credentials.md
```diff
@@ -0,0 +1,52 @@
+# Credentials
+
+For security reasons, we strongly recommend that you *do not* commit any credentials or other secrets to version control.
+
+Kedro is set up so that, by default, if a file inside the `conf` folder (and its subfolders) contains `credentials` in its name, it will be ignored by git.
+
+## How to load credentials
```
astrojuanlu, line 7: These examples are like the "how to directly access configuration in code" from the Basic page, however if I understand correctly the user doesn't need to do that unless they have a reason to? Or is that different for credentials?

Maybe related to my other comment in the parameters file, I'm missing a simple example of how would I use such a credential at the top of this page.

docs/source/configuration/advanced_configuration.md
```diff
@@ -0,0 +1,312 @@
...
+```yaml
+raw_boat_data:
+    type: "${datasets.spark}"  # nested paths into global dict are allowed
+    filepath: "s3a://${bucket_name}/${key_prefix}/${folders.raw}/boats.csv"
+    file_format: parquet
+
+raw_car_data:
+    type: "${datasets.csv}"
+    filepath: "s3://${bucket_name}/data/${key_prefix}/${folders.raw}/${filename|cars.csv}"  # default to 'cars.csv' if the 'filename' key is not found in the global dict
+```
```
astrojuanlu, line 92: Maybe I'd move this example upwards, right after `CONFIG_LOADER_ARGS = {"globals_pattern": "*globals.yml"}`, so the user sees how to use them, and after that, "Alternatively, you can declare which values to fill..."

docs/source/configuration/advanced_configuration.md
```diff
@@ -0,0 +1,312 @@
...
+    }
+}
+```
+
+### How to bypass the configuration loading rules
+You can bypass the configuration patterns and set configuration directly on the instance of a config loader class. You can bypass the default configuration (catalog, parameters, credentials, and logging) as well as additional configuration.
+
+```python
+from kedro.config import ConfigLoader
+from kedro.framework.project import settings
```
astrojuanlu, line 160: It was not entirely clear to me which line was different, this is a way to highlight it:

````suggestion
```{code-block} python
:lineno-start: 10
:emphasize-lines: 8

from kedro.framework.project import settings
````

astrojuanlu (commented):
Read the whole thing, fantastic job @merelcht 👏🏽 I did learn a lot! Left a few style comments but other than that this is a great addition

docs/source/configuration/advanced_configuration.md
merelcht: Done!

docs/source/configuration/parameters.md
merelcht: I've added an intro about parameters.

ankatiyar (approved):
+1 on @stichbury's comment about providing links in the basic docs to related sections in the advanced config page that might be of interest. 
Read through the docs, it looks great! 💯

merelcht:
I fixed all the links and linked the individual config pages to each other where it made sense. I guess over to you one more time @stichbury / @astrojuanlu for a final review.

stichbury (approved):
I'm really happy with how this turned out -- thanks @merelcht! 🏆

Issue #2341, opened 2023-02-17
Rewrite e2e jupyter tests


AhdraMeraliQB:
## Description

The jupyter e2e tests are flaky. Recently [the tests started failing on the Windows CI](https://github.com/kedro-org/kedro/issues/2320), despite no changes being made to any of the components tested. Though resolved now, checking the test output shows that even when e2e tests pass, it's often the case that an error was thrown but resolved itself before the test timed out. This has been the case for some time, not just with the new fix.

Suspicions are that the step "Then jupyter <notebook/lab> should run on port 8888" in `jupyter.feature` is implemented poorly/incorrectly in `cli_steps.py`. This check should be rewritten such that the test becomes dependable and is less prone to failing randomly.

SajidAlamQB:
We have temporarily fixed the issue in https://github.com/kedro-org/kedro/pull/2331 but the solution wasn't optimal. In this issue, we should also investigate this fix further.

PR #2331, opened 2023-02-16
Investigate win e2e tests

SajidAlamQB:
> **NOTE:** Kedro datasets are moving from `kedro.extras.datasets` to a separate `kedro-datasets` package in
> [`kedro-plugins` repository](https://github.com/kedro-org/kedro-plugins). Any changes to the dataset implementations
> should be done by opening a pull request in that repository.
## Description

Attempted to fix windows e2e tests.

Not sure why this fixed it but after discussing it with the team we will go forward with these changes to avoid blockers we should revisit this to investigate further.

## Checklist

- [x] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [x] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [ ] Added tests to cover my changes

PR #2446, opened 2023-03-21
Jupyter e2e test

SajidAlamQB:
## Description

Related to: https://github.com/kedro-org/kedro/issues/2341

The jupyter e2e tests are flaky and need to be rewritten.

## Development notes

Jupyter notebook forks a child process from a parent process and only kills the parent process when it is terminated, therefore we need to keep the extended `ChildTerminatingPopen()` in `sh_run.py` from `subprocess.Popen`. This extension overrides the `terminate()` method from `subprocess.Popen` so that orphan processes are terminated after the parent process is killed.

Our Jupyter e2e tests use this subclass but cause it to fail intermittently. After some digging the reason for this flaky behaviour is down to how long it takes the `kedro jupyter notebooks` command to set up, we flood the port with requests before it has time to fully initialise causing it to spit out failures. Another issue is caused by how we collect the file objects from `stdout` and `stderr`.

### Problem 1 request floods:

- One solution I have been looking at is to poll when the jupyter web server is up and only then proceed with requests. This made use of the `poll` library. I chose to forgo this solution as it requires another dependency on our test requirements and doesn't seem maintained.

- A second solution I looked into involved using some `Popen` objects such as `wait` and `communicate` to poll the Jupyter process but this relied on processes running to completion which wasn't suitable for ours as it needed to continue running in the background, moreover these objects could result in deadlocks as we output `stdout` and `stderr` to `PIPE` the child process could generate enough output that it blocks waiting for the OS pipe buffer to accept more data.

- The final solution was to customise our own `wait_for_notebook_to_run` method and add a new `behave` `Scenario` that would read the output from the process and wait for a given keyword to let it know the webserver was running. This allows us to wait before sending requests to the web server without sending them prematurely. 

### Problem 2 stdout and stderr:
For Windows machines sending the `stdout` and `stderr` to `PIPE` was causing the process to hang and timeout, looking at their outputs it seemed that `warnings` were going to `stdout` while the jupyter information was going to `stderr` the warning on `stdout` would be the cause of the hang. The temporary fix I made back in #2331 worked as it printed everything to the console this isn't optimal as it cluttered the information on CI. To fix this we need to redirect the `stderr` output to `stdout` so that all outputs are in one place.

Other notes:

- `universal_newlines=True` argument is used in `Popen` as it allows the output to be in text rather than binary streams.

## Checklist


features/jupyter.feature
```diff
@@ -10,10 +10,12 @@ Feature: Jupyter targets in new project
 
   Scenario: Execute jupyter notebook target
     When I execute the kedro jupyter command "notebook --no-browser"
+    Then I wait for the jupyter webserver to run for up to "360" seconds
```
merelcht, line 13: How did you decide on 360 seconds? Does it generally take that long for the jupyter webserver to startup? 😮

features/steps/cli_steps.py
```diff
@@ -565,14 +576,7 @@ def check_jupyter_nb_proc_on_port(context: behave.runner.Context, port: int):
     """
     url = f"http://localhost:{port}"
     try:
-        util.wait_for(
-            func=_check_service_up,
-            context=context,
-            url=url,
-            string="Jupyter Notebook",
-            timeout_=15,
```
merelcht, line 573: I'm guessing we don't need this timeout here anymore because the test now waits for the notebook to be up before checking the port?

merelcht (commented):
Thank you @SajidAlamQB for investigating what was going on with these tests and describing it so clearly! ⭐ 

The approach makes total sense to me. I was wondering how you checked that this solves the problem? 

Also don't forget to update the release notes, I think it's worth to have this change there for our own reference.

features/jupyter.feature
SajidAlamQB: It's arbitrary my reasoning was just to give it 5 minutes in case CI performs slow one day, in actuality, it runs the server pretty quickly, within 30 seconds.

features/steps/cli_steps.py
SajidAlamQB: Exactly, we don't need to be constantly flooding the server with requests, rather just wait for the server to tell us when it is ready.

features/jupyter.feature
merelcht: My suggestion would be to lower the timeout in that case. If it usually runs within 30 secs, 2 mins timeout is already quite generous. We can always up it again in case it does take longer on CI generally.

AhdraMeraliQB (approved):
Excellent write-up and great work as a whole! 🌟

merelcht:
I think this is almost ready to be merged 👍  Just one remaining question I left earlier: I was wondering how you checked that this solves the problem?

SajidAlamQB:
> I think this is almost ready to be merged 👍 Just one remaining question I left earlier: I was wondering how you checked that this solves the problem?

I manually tested this on my windows and linux machines as both had slightly different problems due to the way the file handles worked. I observed the outputs of `stdout` and `stderr` before and after to confirm the correct output and that no errors occurred during the requests.

merelcht:
> I manually tested this on my windows and linux machines as both had slightly different problems due to the way the file handles worked. I observed the outputs of `stdout` and `stderr` before and after to confirm the correct output and that no errors occurred during the requests.

That sounds good! I guess it's hard to tell if the flakiness won't happen again, but 🤞

merelcht (approved):

Issue #2310, opened 2023-02-13
Explore adding the `DryRunner` into the Kedro package


merelcht:
### Description

Users ask about the `DryRunner` functionality frequently, so it might be a good idea to move it from the docs and add it as a real class into the Kedro Framework.

### Approach

Search slack for questions/mentions about the `DryRunner` + talk to users directly to ask how they use it.

astrojuanlu:
Last mention of `DryRunner` in Slack according to Linen was in November 2022 https://linen-slack.kedro.org/t/9703736/hello-i-would-like-to-use-something-like-this-dry-runner-htt#14e2145f-ceaa-4f94-9e96-546a21a25158

I'm voting to put it in a separate repo, like @noklam did with https://github.com/noklam/kedro-softfail-runner, and close this issue.

merelcht:
Since there hasn't been a lot of ask recently, I'll close this issue now. We can always add the runner to a repo when demand spikes again.

PR #2460, opened 2023-03-23
Update the DryRunner doc to make it more useful

### LINKED ISSUE #1997: Dry run option that checks DAG and existence of pipeline inputs
## Description
Hello, I would like to use something like this dry runner https://kedro.readthedocs.io/en/stable/nodes_and_pipelines/run_a_pipeline.html#custom-runners but in a slightly different context - I would like to also check whether the data exists.
I am using multiple kedro environments https://kedro.readthedocs.io/en/stable/kedro_project_setup/configuration.html for different model experiments
When I do a partial run with `--from-nodes` and `--to-nodes` (to save time over a full pipeline run), I often discover that some data does not exist in my environment - but it takes a while to discover this, as the code needs to run first
Then "checking whether the data exists" is a bit complex:
- Either check whether it is an intermediate output of the provided pipeline
- Or check whether it can be read from the catalog using the _exists method of the abstract dataset class

## Context
This would save considerable time in debugging pipeline runs.

## Possible Implementation
I shared a rough, unpolished implementation in this slack thread: https://kedro-org.slack.com/archives/C03RKP2LW64/p1667314369586859
Improvements:
- Using public `exists` method
- Raising the appropriate data not existing error

noklam:
## Description

Close #1997 
I worked with this user previously and created a more useful DryRunner.

We can decide later whether we want to include this officially in `kedro` but we can change our doc first before that happens.
See 
* #2310 

## Development notes

The `DryRunner` is more powerful now as it checks whether the underlying data exists and surface the error early. Previously if you have a long pipeline the error will not show up until the node is executed and you can have `FileNotFoundError`. The DryRunner wouldn't load the data so it's fast, it only use the `exists` method to make sure such data exist in the path.

## Checklist


stichbury (approved):
LGTM, thanks! I take it that we don't have this as code anywhere, so the docstring you've fixed doesn't also need to be changed within some Python somewhere?

noklam:
@stichbury Thanks! Correct, at the moment it's just an example in our doc, #2310 is the ticket to decide whether we should include it in the Python package.

astrojuanlu:
Rendered version: https://kedro--2460.org.readthedocs.build/en/2460/nodes_and_pipelines/run_a_pipeline.html#custom-runners

merelcht (approved):
Thanks @noklam! This looks good 👍

noklam:
Thank you @jwaton for suggesting this initially. :)

Issue #2466, opened 2023-03-24
OmegaConfigLoader returns a DictConfig object instead of dictionary


noklam:
## Context
See parent issue #2458 

## Description
Short description of the problem here.

1. This looks a bit weird when inspected since the printed object looks like `Dict[str]`, and force users to know more about `OmegaConf` since this parameters is returning some Omega's object instead of a real dictionary 
2. Technically if the function is expecting a dict `def some_fun(arg1: dict)`, is it the right thing to do to pass a `Omega` object? It may confuse static code analyzers like `mypy`, but I doubt it works with `kedro` anyway. 
The more problematic case will be when the function is doing some kind of type-checking `isinstance(arg1, dict)`, it will get it wrong. Arguably it makes the code more coupled with Kedro.
Few possible outcome:
1. Improve the display of parameters - so it prints out the interpolated value instead of weird string that could confuse user
2. Could we return an interpolated dictionary instead? This needs to work well with resolvers.

## Steps to Reproduce
`git clone git@github.com:noklam/omega-nest-params-bug-demo.git`

```yaml
# parameters.yml
demo: ${global}

demo_nested:
  param: ${global}
```

```yaml
# parameters_global.yml
global: test_1
```

The function will simply print out the `parameters`.
```python
def demo_node(parameters, nested_parameters):
    pprint(parameters)
    print(parameters["demo_nested"]["param"])
    pprint(nested_parameters)
```

### Problem 1 - Only top-level parameters are interpolated
`kedro run`

#### Expected Result
```python
{'demo': 'test_1', 'demo_nested': {'param': 'test_1'}, 'global': 'test_1'}
test_1
{'param': 'test_1'}
```
#### Actual Result
```python
{'demo': 'test_1', 'demo_nested': {'param': '${global}'}, 'global': 'test_1'}
test_1
{'param': '${global}'}
```

```python
    print(type(parameters))
    print(type(parameters["demo_nested"]))
    print(type(parameters["demo_nested"]["param"]))

<class 'dict'>
<class 'omegaconf.dictconfig.DictConfig'>
<class 'str'>
```

```
-- If you received an error, place it here.
```

```
-- Separate them if you have more than one.
```

## Your Environment
Include as many relevant details about the environment in which you experienced the bug:

* Kedro version used (`pip show kedro` or `kedro -V`):
* Python version used (`python -V`):
* Operating system and version:

astrojuanlu:
Was this closed in gh-2467?

[...]

PR #2176, opened 2023-01-05
Only convert omegaconf DictConfig to primitive type for logging

merelcht:
Signed-off-by: Merel Theisen <merel.theisen@quantumblack.com>

## Description
After merging https://github.com/kedro-org/kedro/pull/2174 where I made `OmegaConfLoader` return the primitive dict type I realised that in order to leverage other `omegaconf` functionality such as variable interpolation (https://omegaconf.readthedocs.io/en/2.3_branch/usage.html#variable-interpolation) it's necessary to have it return the `omegaconf.dictconfig.DictConfig` type. 

## Development notes
The problems I discovered with logging were related to this line 
[`LOGGING.configure(logging_config)`](https://github.com/kedro-org/kedro/blob/main/kedro/framework/project/__init__.py#L258)

and so I decided to revert my previous change and only convert the `omegaconf.dictconfig.DictConfig` type to the primitive dict for the logging files. 

## Checklist

- [ ] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [ ] Added tests to cover my changes

<a href="https://gitpod.io/#https://github.com/kedro-org/kedro/pull/2176"></a>

Issue #1965, opened 2022-10-24
Remove custom Kedro syntax for --params


merelcht:
## Description
Follow up to https://github.com/kedro-org/kedro/issues/1907

To-do:
* Remove the support of Kedro custom syntax
  * `kedro run`
  * `reload_kedro` magic

`kedro run --params=key:value` (The Kedro syntax for params)

## Context
In `0.19.0` we will remove the Kedro syntax for providing parameters and only use `OmegaConf` syntax.

ankatiyar:
Following #2378 -
* The output for `_split_params` function in `kedro/framework/cli/utils.py` has changed from `omegaconf.DictConfig` type to `dict`. This should ideally be changed back to `DictConfig` so `omegaconf` can be used to merge runtime params with parameters from config files.
* We need to be careful with merging nested parameters (see #2357) - this is happening right now in `_update_nested_dict()` in `context.py` 
    * `ConfigLoader` loads the parameters and the nested params from config files as a `dict` type. 
    * `OmegaConfigLoader` loads config params as a `dict` right now but the nested parameters are of the type `DictConfig`
* Using `DictConfig` was causing issues with `Kedro Viz` because the `DictConfig` type is not json-serialisable. The runtime parameters still need to be converted to `dict` type before going in the `session._store` to not break experiment tracking.

AntonyMilneQB:
Just to add to the above, I think what we want is:
* `_split_params` should be extremely simple, just doing `OmegaConf.from_dotlist` really
* `_update_nested_dict` should disappear altogether and just use `OmegaConf` merging functionality, nothing custom

As part of a separate ticket, we should probably also be doing something about what the `OmegaConfigLoader.load_and_merge_dir_config` returns. This was changed in https://github.com/kedro-org/kedro/pull/2176 but I don't understand why we're using `dict` rather than just doing this:
```
OmegaConf.to_container(OmegaConf.merge(*aggregate_config), resolve=True)
```
If we changed this then we could also remove the special `to_container` behaviour we have on the logging config.

Ultimately, the config loader should probably be returning a pure Python dictionary and nothing `OmegaConf` specific, otherwise it's a leaky abstraction as it is now.

With that said, I don't know what the ultimate plan is about how we do the parameters merging in `context.params`. I assume we can use `OmegaConf` as a dependency here but don't actually want to rely on the fact that someone is using `OmegaConfigLoader`. So to keep it independent of config loader implementation I guess we'd do:
[...]

PR #2467, opened 2023-03-24
OmegaConfigLoader returns `Dict` instead of `DictConfig`, resolves runtime_params properly

### LINKED ISSUE #2458: OmegaConfigLoader parameters does not propogate properly
## Description
An user report this in Slack https://kedro-org.slack.com/archives/C03RKP2LW64/p1679502160140669

To summarise this, there are two potential issues here
1. Only top-level parameters are interpolated - any nested parameters will be an `OmegaConf` object instead - Functionally it will still works when accessing a group of parameters
2. CLI override `global` but the `global` isn't  propagate to the references

## Context
1. is potentially a bug
2. Definitely a bug

## Steps to Reproduce
I created an example so you can clone and run with this command:
`git clone git@github.com:noklam/omega-nest-params-bug-demo.git`

```yaml
# parameters.yml
demo: ${global}

demo_nested:
  param: ${global}
```

```yaml
# parameters_global.yml
global: test_1
```

The function will simply print out the `parameters`.
```python
def demo_node(parameters, nested_parameters):
    pprint(parameters)
    print(parameters["demo_nested"]["param"])
    pprint(nested_parameters)
```

### Problem 1 - Only top-level parameters are interpolated
`kedro run`

#### Expected Result
```python
{'demo': 'test_1', 'demo_nested': {'param': 'test_1'}, 'global': 'test_1'}
test_1
{'param': 'test_1'}
```
#### Actual Result
```python
{'demo': 'test_1', 'demo_nested': {'param': '${global}'}, 'global': 'test_1'}
test_1
{'param': '${global}'}
```

This may not be a real problem, since when access `parameters["demo_nested"]["param"] it will still interpolate correctly. 
Two minor issues here:
1. This looks a bit weird when inspected since the printed object looks like `Dict[str]`, and force users to know more about `OmegaConf` since this parameters is returning some Omega's object instead of a real dictionary 
2. Technically if the function is expecting a dict `def some_fun(arg1: dict)`, is it the right thing to do to pass a `Omega` object? It may confuse static code analyzers like `mypy`, but I doubt it works with `kedro` anyway. 
The more problematic case will be when the function is doing some kind of type-checking `isinstance(arg1, dict)`, it will get it wrong. Arguably it makes the code more coupled with Kedro.

```python
    print(type(parameters))
    print(type(parameters["demo_nested"]))
    print(type(parameters["demo_nested"]["param"]))

<class 'dict'>
<class 'omegaconf.dictconfig.DictConfig'>
<class 'str'>
```

### Problem 2 - CLI override `global` but the `global` isn't  propagate to the references
`kedro run --params=global:test_2 `

#### Expected Result
```python
{'demo': 'test_2', 'demo_nested': {'param': '${global}'}, 'global': 'test_2'}
test_2
{'param': '${global}'}
```
#### Actual Result
```python
{'demo': 'test_1', 'demo_nested': {'param': '${global}'}, 'global': 'test_2'}
test_1
{'param': '${global}'}
```

Here the issue is `global` is override by CLI, but all the `${global}` reference is not. This probably has to do with the order of execution.

@merelcht

noklam:
## Description

Close #2458, #2466 

This PR changes a lot how OmegaConfigLoader behavior although not many lines of code are changed.
https://github.com/kedro-org/kedro/blob/fd42a5a17523e645dce2df7488a204ff3662fb81/kedro/config/omegaconf_config.py#L278

The main changes are the resolving order and what's return by the config_loader
1. Lazily evalution -> Eager resolution as Dictionary
2. The merge with `runtime_params` is pushed to `OmegaConfigLoader` instead of `KedroContext`, this ensure that CLI params are updated properly

1. will ensure that the OmegaConfigLoader comply with `AbstractConfigLoader` and 2. will ensure that overriding parameters with CLI work as expected.

Note: This will only make `parameters` work, more work needs to be done to generalise it to all configurations. #2531 

## Checklist


noklam:
Need some more time to study. But for answering the 1st question. I don't think this is the correct way to do it, but currently cross-env interpolation is blocked because of the `dict` cast (see more in the description) @MatthiasRoels

image

noklam:
The root cause of the issue has to do with the resolution order, this is roughly what happen behind the scene:
1. `config_loader["parameter"]` - it needs to return a Dictionary
2. `load_and_merge_dir_config` (OmegaConfigLoader) - this is where the merging happen, ideally we need to use `OmegaConf.to_container(..., resolve=True) here
3. The result of 2 will merge `context.extra_params` recursively

When user do `kedro run --params global:val`, it will update the `context.extra_params`. In order to make the extra_params update all the reference, we need to keep it as a OmegaConf.DictConfig, because **resolver** has to be the last step. It also create the leaky abstraction problem because the behavior is not supposed to be ConfigLoader specific.

noklam:
[5b0bff8](https://github.com/kedro-org/kedro/pull/2467/commits/5b0bff8504fbe1dcb00b4132dcc08a01b059c7dd) Would solve both problem
* #2458 
* #2466 
* And potentially remove the need of #2176

merelcht:
> [5b0bff8](https://github.com/kedro-org/kedro/pull/2467/commits/5b0bff8504fbe1dcb00b4132dcc08a01b059c7dd) Would solve all these problem

Just to clarify something. In the earlier comment you mentioned:
> There are 2 options for step 2
A) Convert it to a Dictionary using OmegaConf.to_container(..., resolve=True)
B) Keep it as a OmegaConf.DictConfig

> Option A seem to be the right thing to do, as it is consistent with what AbstractConfigLoader requires. However, overriding with params= will not work anymore because all the reference to ${global} is gone once the resolve=True happen.

So I'm a bit confused as to [5b0bff8](https://github.com/kedro-org/kedro/pull/2467/commits/5b0bff8504fbe1dcb00b4132dcc08a01b059c7dd) solving all problems. Will it or will it not allow overriding with `params=` ?

AntonyMilneQB:
Great work @noklam! I had reached similar conclusions to you also. In my mind there are just two different problems here:
1. whether and how to resolve the omegaconf `DictConfig` on [this line](https://github.com/kedro-org/kedro/blob/3b5f595229a24f5129aef7d9915ce9872eb14857/kedro/config/omegaconf_config.py#L278)
2. how to handle `extra_params`

On 1: both me (https://github.com/kedro-org/kedro/issues/1965#issuecomment-1453489361) and @matthiasroels  (https://github.com/kedro-org/kedro/issues/2458) independently reached the same conclusion here: the current code looks wrong and the correct, non-leaky thing to do here is your option A above where you do `to_container(resolve=True)`. I imagine the reason this wasn't done originally in #2176 is just that @merelcht didn't spot that you could set `resolve=True` in the `to_container` call? I'm sure this would solve the 3rd problem you list as "potentially solved" in your above comment also.

Now the 2nd problem of `extra_params` is a bit trickier. Here I _previously_ thought the "correct" flow is the one that I suggested in https://github.com/kedro-org/kedro/issues/1965#issuecomment-1453489361. This is somewhat unsatisfactory for two reasons though:
1. as I said there, it feels a bit circular/redundant to convert these objects to dicts, then back into dictconfig and then back into dict again. But in order to achieve something that's non-leaky I think that's what we need to do
2. something I didn't realise at the time I wrote that but you noticed with your `global` example: if `extra_params` are injected after `to_container(resolve=True)` then it's too late. As you also observed with `TemplatedConfigLoader` this actually isn't a new problem, but it would be nice if it did work

So what is the ideal solution here? I think exactly what you propose: since we already have `runtime_params` available in the configloader, let's merge them in _before_ the call to `to_container(resolve=True)`. I think there are a few potential objections here, but IMO none should stop us from making this change:
1. what is the point of the update in `context.params`? The original intention for this would have been for the interactive workflow where `context.params` can be changed on the fly without needing to do `%reload_kedro`. But as we've discussed before, this functionality is basically unknown and pointless, and I'd be very happy to remove it and just resolve everything eagerly instead
2. it means that `--params` can be used to inject variables into _any_ config file, not just `parameters`. e.g. you could inject variables into your catalog through `--params`. This goes against the original intention of `--params` (that they are just for overriding parameters.yml) but is functionality that is sometimes called for (as in #1527) and so in my mind this is actually an improvement. If we really wanted to limit this functionality then we could put some conditional into the config loader itself so that the merging is only done when `parameters` are called for (rather than `catalog` etc.), but I don't like this
3. _if_ we move the merging behaviour out of `context.params` entirely then any other configloaders would also need to perform the merging behaviour inside the config loader. This would include the existing `ConfigLoader` and `TemplatedConfigLoader` if they remain. But I think in future `OmegaConfLoader` should be the only one shipped with kedro, so then there's no need to worry about this. Anyone writing their own custom config loader would need to add it into their implementation though

Overall I think there's possibly a current objection (maybe @idanov would say this?) that the config loader should not be responsible for merging in runtime parameters. Or maybe `--params` wouldn't be the right name for the CLI flag any more if it's more general than that. But personally I think this would overall be an improvement and I don't see any better solution anyway.

So here's what I think the ideal final state would look like, without worrying right now about how to implement it or what's a breaking change etc.
* `--params` are processed with `_split_params` that does `OmegaConf.from_dotlist` and then `to_container(resolve=False)` to turn into a `runtime_params` dictionary (_not dictconfig_)
* pass `runtime_params` into configloader as you do here
* do `to_container(resolve=True)` inside omegaconfig loader including `runtime_params` as you do here
* ship just `omegaconfigloader` with kedro
* remove the merge from `context.params` entirely
* consider changing the name of `--params` if it's going to be used for more than just `parameters`

kedro/config/omegaconf_config.py
```diff
@@ -275,7 +277,9 @@ def load_and_merge_dir_config(
             return {}
         if len(aggregate_config) == 1:
             return list(aggregate_config)[0]
-        return dict(OmegaConf.merge(*aggregate_config))
+        return OmegaConf.to_container(
+            OmegaConf.merge(*aggregate_config, self.runtime_params), resolve=True
```
AntonyMilneQB, line 281: Does this definitely work when `self.runtime_params` is a dictionary and not a `DictConfig`? i.e. is it ok to not do `OmegaConf.create(runtime_params)` here? Looking through omegaconf source I'm pretty sure this is fine (only the first argument of `OmegaConf.merge` needs to be `DictConfig`) but please do check this!

kedro/config/omegaconf_config.py
```diff
@@ -129,6 +129,8 @@ def __init__(
             env=env,
             runtime_params=runtime_params,
         )
+        if not self.runtime_params:
+            self.runtime_params = {}
```
AntonyMilneQB, line 133: Probably nicer to do `runtime_params=runtime_params or {}` above instead?

AntonyMilneQB: Or even better `AbstractConfigLoader` should actually be the class responsible for this defaulting behaviour, i.e. it should do `self.runtime_params = runtime_params or {}`.

noklam:
@AntonyMilneQB Maybe slightly off topic.  https://github.com/kedro-org/kedro/issues/2481 but in my mind this is the more correct way to do in long term, but should be discussed separately as I think we need a temporary solution.

AntonyMilneQB:
Related to all this, we also need to think about this bit of code:

```
        if len(aggregate_config) == 1:
            return list(aggregate_config)[0]
```
As it stands if we do [5b0bff8](https://github.com/kedro-org/kedro/pull/2467/commits/5b0bff8504fbe1dcb00b4132dcc08a01b059c7dd) then it will still return an unresolved `DictConfig` if there's only one file in `aggregate_config`. I don't actually know what this special `len == 1` case is for - the code should work even when `len(aggregate_config) == 1` without any special treatment? @merelcht @noklam 

```
# Works fine, so no need for `len(aggregate_config) == 1` case
OmegaConf.merge(*[OmegaConf.create({1: 2})])

# Doesn't work fine, hence need for `if not aggregate_config` case
OmegaConf.merge(*[])
```

merelcht:
I don't exactly remember right now why I did this

> if len(aggregate_config) == 1:
>             return list(aggregate_config)[0]

but there was definitely a good reason. There was something about just 1 config file not working the same as multiple.

noklam:
@AntonyMilneQB I had the [same comment](https://github.com/kedro-org/kedro/pull/2085#discussion_r1036996001) before, I haven't really tested it so I am not sure here.

kedro/config/omegaconf_config.py
noklam: I checked with the source code, `merge` can take any primitive type  I think it's fine here.

```python
    def merge(
        *configs: Union[
            DictConfig,
            ListConfig,
            Dict[DictKeyType, Any],
            List[Any],
            Tuple[Any, ...],
            Any,
        ]
```

kedro/config/omegaconf_config.py
noklam: 100% agree with this.

RELEASE.md
```diff
@@ -11,6 +11,9 @@
 # Upcoming Release 0.18.9
 
 ## Major features and improvements
+* `OmegaConfigLoader` will return a `dict` instead of `DictConfig`.
+* `kedro run --params` now update interpolated parameters correctly.
```
merelcht, line 15: That's also just for the `OmegaConfigLoader` right?

kedro/config/omegaconf_config.py
```diff
@@ -217,6 +218,7 @@ def load_and_merge_dir_config(
         Args:
             conf_path: Path to configuration directory.
             patterns: List of glob patterns to match the filenames against.
+            key: Str of the dictionary keys to access.
```
merelcht, line 221: This is quite a generic description. Maybe change it to: "Key of the configuration type to fetch."

tests/config/test_omegaconf_config.py
```diff
@@ -530,3 +536,30 @@ def zipdir(path, ziph):
...
+    @use_config_dir
+    def test_runtime_params_override_interpolated_value(self, tmp_path):
+        """Make sure interpolated value is updated correctly with runtime_params"""
+        conf = OmegaConfigLoader(str(tmp_path), runtime_params={"global": "global"})
+        params = conf["parameters"]
+        assert params["templated_param"] == "global"
+
+    @use_config_dir
+    def test_runtime_params_not_propogate_non_parameters_config(self, tmp_path):
+        """Make sure `catalog`, `credetials` won't get updated by runtime_params"""
```
merelcht, line 557:
```suggestion
        """Make sure `catalog`, `credentials` won't get updated by runtime_params"""
```

merelcht (approved):
LGTM! Thanks for getting to the bottom of this ⭐ 

One minor comment: in the tests it might be better to not refer to "global" but instead just a random parameter name, like test_split or whatever. Now it seems like this functionality is specifically for "globals" like what we have in `TemplatedConfigLoader`, but it's actually different.

RELEASE.md
noklam: That's correct, I will make changes to this

noklam: I change `templated_param` -> `interpolated_param` to align with the OmegaConf terminology.

I have changed the `global` name, but I think we need to come up with a better name since we are calling it global with `OmegaConfigLoader`, i.e.`parameters_global.yml` is what suggested in the doc.

RELEASE.md
antonymilne, line 1: Personally I'd put all this as bug fix rather than major improvement.

RELEASE.md
```diff
@@ -11,6 +11,10 @@
 # Upcoming Release 0.18.9
 
 ## Major features and improvements
+* `OmegaConfigLoader` will return a `dict` instead of `DictConfig`.
+* `kedro run --params` now update interpolated parameters correctly when using
+`OmegaConfigLoader`.
```
antonymilne, line 16:
```suggestion
* `kedro run --params` now updates interpolated parameters correctly when using `OmegaConfigLoader`.
```

kedro/config/omegaconf_config.py
```diff
@@ -275,7 +277,13 @@ def load_and_merge_dir_config(
             return {}
         if len(aggregate_config) == 1:
             return list(aggregate_config)[0]
-        return dict(OmegaConf.merge(*aggregate_config))
+
```
antonymilne, line 280: I think the flow here is a little unclear (too many different `return` cases) and, more importantly, unless I'm missing something there's actually a bug: if you have a single element in `aggregate_config` (i.e. one parameters file) then `runtime_params` don't get merged in.

This would be solved if we removed lines 276-279. I know @merelcht said there was some reason to pick out the `len == 1` case before but I don't see what this could be now that we always want to do `to_container(resolve=True)`.

antonymilne (commented):
This is nearly there but I think there's a bug with it unless I'm missing something: what happens if you have a single `parameters.yml` file and want to use `runtime_params`?

As part of a separate PR I think we should revert #2378 (but keep `test_no_DictConfig_in_store` as it's a good regression test) and #2176.

kedro/config/omegaconf_config.py
noklam: I would be keen on removing it, unless we have a specific failing case. I try remove line 278-279 and all test case passed.

Should we separate this into another issue? Removing Line 276-277 will cause a few tests fail, I suspect it's also link to https://github.com/kedro-org/kedro/pull/2556 and some refactoring may be needed.

noklam:
> As part of a separate PR I think we should revert https://github.com/kedro-org/kedro/pull/2378 (but keep test_no_DictConfig_in_store as it's a good regression test) and https://github.com/kedro-org/kedro/pull/2176.

Agree with this I can open an issue, it will be redundant.

kedro/config/omegaconf_config.py
merelcht: I'm happy for those lines to be removed if everything is still working fine. I don't remember why I added it, so really should have left a comment 😅

antonymilne: I think we should remove line 278-279 here since as I understand it at the moment they are not just unnecessary but following this change will actually be causing a bug (would be good if you could check this): if you just have a single parameters.yml file, the `runtime_params` don't get merged in.

Lines 276-277 don't actually cause a bug I think, so if removing them makes some tests fail and it seems related to #2556 then let's leave those here for now and try to remove them in that PR.

RELEASE.md
antonymilne: This is a fair comment I think for the "interpolated parameters" point, but I still think the `dict` vs. `DictConfig` point should be under bug fix.

antonymilne (approved):
Thank you for fixing it! Do take a look at my most recent comments but basically I think this is good to go 🙂 Would be nice to follow up with #2556 and the reversions I mentioned soon though just to get it all cleared up.

RELEASE.md
noklam: @antonymilne I agree with this and I will move it to bug fix.

PR #2469, opened 2023-03-27
Update kedro-viz version in docs

AntonyMilneQB:
[kedro-viz 6.0.0 has been released](https://github.com/kedro-org/kedro-viz/pull/1297) but is not actually a breaking release from a kedro user perspective.

We should bump the requirement in our docs so that users get the latest version.

See also https://github.com/kedro-org/kedro-starters/pull/121.

### DIFF
--- a/docs/source/tutorial/tutorial_template.md
+++ b/docs/source/tutorial/tutorial_template.md
@@ -52,7 +52,7 @@ pytest~=7.2
 kedro~=0.18.7
 kedro-datasets[pandas.CSVDataSet, pandas.ExcelDataSet, pandas.ParquetDataSet]~=1.1
 kedro-telemetry~=0.2.0
-kedro-viz~=5.0 # Visualise pipelines
+kedro-viz~=6.0 # Visualise pipelines
 
 # For modelling in the data science pipeline
 scikit-learn~=1.0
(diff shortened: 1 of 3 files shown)


stichbury (approved):
Thank you!

tynandebold (approved):

AntonyMilneQB:
@stichbury what is the status of the `docs_linkcheck` job? It's been a while since I've built docs so I'm not sure what to do about that 😅 Try and fix something? Rerun CI till it passes? Get someone to merge this anyway?

astrojuanlu:
`docs_linkcheck` is fixed by https://github.com/kedro-org/kedro/pull/2459 @AntonyMilneQB, not worth waiting I think. You can safely go ahead.

stichbury:
@AntonyMilneQB This looks like /en/stable/faq/faq.html#what-is-data-engineering-convention is causing problems. It was removed recently. I've added a redirect in the RTD console but anchor tags are always hit and miss. Would it be possible to remove it from the API docstring that is referencing it? Let me just see where it is and do it (if I can).

astrojuanlu:
You can cherry pick https://github.com/kedro-org/kedro/pull/2459/commits/e3bf47a997361410e4b272304368b5bda5aab889

stichbury:
I would, and could, but right now there's an issue with GitHub that's blocking that.

AntonyMilneQB:
Thank you @stichbury!

PR #2489, opened 2023-04-03
Replace command_path session data with full CLI command

MaximeSteinmetz:
## Description
It is not currently possible to easily reproduce a run that used CLI args (e.g. `--params=key:value`) because said args are not easily accessible.

## Development notes
- Recover the command string using `sys.argv` and persist it in session store by changing the content of `command_path` in `_jsonify_cli_context` (Basically followed @AntonyMilneQB 's [suggestion](https://github.com/kedro-org/kedro-viz/pull/1303#issuecomment-1488633175) from previous PR)

This allows for the command with arguments to be accessible in Kedro Viz as shown below

image

Limitation: this won't work for runs made without CLI (for instance using `session.run()`)

## Checklist


datajoely:
Love this ❤️ - thanks @MaximeSteinmetz

AntonyMilneQB:
@noklam This is not a breaking change since everything in the session store is just stored as a big blob string, so can be modified at will. Within that blob the structure has not changed either because the value of `command_path` used to be a string and still is. It's just a longer (and more correct) string now 😀 

Thank you very much for the contribution @MaximeSteinmetz! I am also curious exactly how this works before and after your change with `session.run` outside the CLI and will give it a quick test.

MaximeSteinmetz:
Sure, so:
1. The Session logic is set up so that in the absence of a `click` context, no cli-related info is stored in session store. This means that both before and after changes, `run_command` will be absent which results in an empty Run Command string in Kedro Viz -> no reproducibility. It looks like recovering a command that reproduces the run in this scenario would be trickier, one would have to convert the `extra_params` as well as arguments to `session.run()` like `pipeline_name` to reconstruct a command string. Unless there is an easier path I'm missing
2. Completely agree here, with this solution we can reproduce when there are no local changes. Also good point for the data versioning, I'm not familiar with this concept yet. If the data version used can be added to source code then we can reproduce (maybe using something like https://docs.kedro.org/en/stable/data/kedro_io.html#checkpoint-configuration)

AntonyMilneQB:
I just spent a while playing around and agree with @MaximeSteinmetz. Basically this line will mean that we don't look for a CLI command when a kedro run is invoked other than through the CLI:
https://github.com/kedro-org/kedro/blob/0497d45b4ddb8ba53118ced0e855cbb088d7323e/kedro/framework/session/session.py#L172

Indeed there's not currently a good way of saving the run command to the session store for runs done outside CLI `kedro run`. Possibly we should come up with something in future but there hasn't been any call for it yet really. Another (probably more important) issue here is that the session store isn't actually saved to until `session.close` is called. With the session context manager that's done automatically, but in `kedro ipython` you'd need to do it explicitly if you actually wanted to save to your session store. So really we're not currently setup to make reproducible experiment runs from a Jupyter notebook anyway.

The idea of making a run fully reproducible through kedro-viz has definitely been mentioned before but not much progress made on it yet. There's actually a flag `git["dirty"]` that's recorded in the session store that should say whether you have uncommitted code changes, but that doesn't get used anywhere on kedro-viz at the moment and won't actually record what the code changes are anyway.

If you've loaded a particular version of data using `kedro run --load_versions` then after this change that will also be saved in the run command and so is reproducible. But other mechanisms of versioning data won't be accounted for.

So yeah, if we want full reproducibility there's definitely some way to go. This PR is a very good step forward though 🚀

RELEASE.md
```diff
@@ -17,6 +17,7 @@
 * Improvements to documentation about configuration.
 * Improvements to Jupyter E2E tests.
 * Improvements to documentation on visualising Kedro projects on Databricks.
+* Added CLI arguments information to session store to improve run reproducibility using `Kedro-Viz` Run Command.
```
AntonyMilneQB, line 20:
```suggestion
* Added full `kedro run` CLI command to session store to improve run reproducibility using `Kedro-Viz` experiment tracking.
```

tests/framework/session/test_session.py
```diff
@@ -244,6 +244,7 @@ def test_create(
         fake_username,
     ):
         mock_click_ctx = mocker.patch("click.get_current_context").return_value
+        mock_sys_argv = mocker.patch("sys.argv").return_value
```
AntonyMilneQB, line 247: We can make this test a bit more explicit and powerful I think. Something like this (similar changes to the other test in this file):

```suggestion
        mock_sys_argv = mocker.patch("sys.argv", ["kedro", "run", "--params=x"])
```

tests/framework/session/test_session.py
```diff
@@ -253,7 +254,7 @@ def test_create(
             "args": mock_click_ctx.args,
             "params": mock_click_ctx.params,
             "command_name": mock_click_ctx.command.name,
-            "command_path": mock_click_ctx.command_path,
+            "command_path": " ".join(["kedro"] + mock_sys_argv[1:]),
```
AntonyMilneQB, line 257:
```suggestion
            "command_path": "kedro run --params=x",
```

AntonyMilneQB (approved):
Thank you very much for the contribution! ⭐

MaximeSteinmetz:
Thanks very much for the reviews all!
Kind of messed up the merging operations here, apologies! I'd appreciate some help with the build steps that aren't passing

MaximeSteinmetz:
Hi everyone, let me know if there is anything more needed on my side for merging

AntonyMilneQB:
Just one more ✅  approval basically. @noklam maybe?

noklam (approved):
Sorry for blocking this, I actually thought I have approved it already.

PR #2496, opened 2023-04-06
Fix/tensorflow macos test

PtrBld:
On Macs there is "tensorflow-macos" instead of "tensorflow" for installing Tensorflow. I added the dependency in setup.py and test-requirements since it was missing there

## Development notes
There are no tests required since it is just a dependency issue


noklam:
Thank you for making this PR. Noted that all datasets change are in kedro-datasets instead of kedro.

https://github.com/kedro-org/kedro-plugins/pull/142

This have been merged. You can do pip install kedro-dataset without any changes as long as you are using kedro > 0.18.4. You can find more in the Relalease.md

Issue #2456, opened 2023-03-22
IncrementalDataSet lazy loading


PtrBld:
## Description
We are using IncrementalDataSets since we get new data every week and want only process the new data. However, the data is very large and every week we get multiple partitions that are processed on a weekly basis. This data is then too large to be loaded into RAM at once. As far as we understand there is no way to use IncrementalDataSets but still load every partition on its own similar to the PartitionedDataSet.

## Possible Implementation
Possibly this could be solved by not calling the load function in https://github.com/kedro-org/kedro/blob/05915901d43351766669eca50c3a2fcd6246bffb/kedro/io/partitioned_dataset.py#L532-L534, but return the function instead. This would be the same functionality as provided in the PartitionedDataSet and would also simplify switching between PartitionedDataSet and IncrementalDataSet without changing the node code.

astrojuanlu:
Hi @PtrBld, thanks for opening this issue! Do you think https://github.com/kedro-org/kedro/issues/2374 is close enough to your use case?

PtrBld:
Hi @astrojuanlu, thanks for the very quick response. I think we could create something like a LazyLoadIncrementalDataSet that inherits from IncrementalDataSet and just overrides the load function to not call load, if this is what you are hinting at. I was just wondering if there was a specific design decision that lead to the loading of the data in the IncrementalDataSet?

As I understand kedro the different data sets should be as close as possible in their behavior especially since IncrementalDataSet directly inherits from PartitionedDataSet

astrojuanlu:
I don't have a good answer to that, will defer to the Kedro engineering team to address that question.

HOZHENWAI:
Hi, I also have the same issue. I also have the same question about the different behavior, there is probably a usecase but I fail to see it.

noklam:
@PtrBld Will you be happy to make a PR to fix this? From the top of my head I think it should have same behavior as Partitioned Dataset and I can't remember why it is not. Could be an oversight but I am not sure.

PtrBld:
I am happy to create a PR. However, this will be breaking changes to the current behavior. Maybe we could add this as an option to partitioned and incremental data sets?

jmholzer:
Tech Design 21/06/23:

**Intro to problem**
`PartitionedDataSet` and `IncrementalDataSet` are datasets used to handle data that is split across multiple files.

`PartitionedDataSet`:
[...]

PR #2497, opened 2023-04-06
[Draft] #2456 Make incremental data loading lazy

PtrBld:
## Description
The implementation of the load function for the `PartitionedDataSet` and `IncrementalDataSet` do not match for no apparent reason (one is lazy loaded the other is eager loaded). See #2456. 

## Development notes
This is a breaking change. I have not yet adjusted the release file because I do not know if the implemented behavior is really the wanted behavior. If everything looks good I can prepare the release notes. Another alternative for this issue would be to introduce a boolean ``eager`` parameter to the PartitionedDataSet and let the user decide if the dataset should be eager or lazy loaded.

## Checklist


PtrBld:
Does someone have an opinion on the breaking change problem?

noklam:
@PtrBld From what I understand the main difference is that you need to call `()` when you consume the dataset. I think that's good, I don't see why it should deviated from PartitionedDataSet's behavior. However this will be a breaking change and goes in 0.19.

I will make sure this gets discussed next week with the team, putting it into the backlog discussion.

Issue #2129, opened 2022-12-14
Rename all mentions of `DataSet` to `Dataset`


merelcht:
### Description
Rename `DataSet` to `Dataset` in `kedro`:
- [x] Remove aliases from `core.io` datasets: https://github.com/kedro-org/kedro/pull/3147
- [x] Replace any mentions of `DataSet` in `.py` files: https://github.com/kedro-org/kedro/pull/3147
- [x] Replace any mentions of `DataSet` in `.md` (docs) files: https://github.com/kedro-org/kedro/pull/3148
- [x] Replace any other mentions of `DataSet` in the project: https://github.com/kedro-org/kedro/pull/3147

Rename `DataSet` to `Dataset` in `kedro-datasets` -> This is for `kedro-datasets` `2.0.0` so moving it in a separate issue: https://github.com/kedro-org/kedro/issues/3159

Other:
- [x] Replace any mentions of `DataSet` in docs in all repos: `kedro-starters`: https://github.com/kedro-org/kedro-starters/pull/155, `kedro-plugins`: https://github.com/kedro-org/kedro-plugins/pull/384
- [x] Replace any mentions of `DataSet` in tests in all repos: `kedro-starters`: https://github.com/kedro-org/kedro-starters/pull/155, `kedro-plugins`: https://github.com/kedro-org/kedro-plugins/pull/384
- [x] [Graphic in the Data Catalog section on the website](https://kedro.org/#features): https://github.com/kedro-org/kedro-website/pull/167
- [x] On `main`: add messaging about move of datasets + rename: https://github.com/kedro-org/kedro/pull/3155
- [x] Any mentions in `kedro-viz`: https://github.com/kedro-org/kedro-viz/pull/1572

astrojuanlu:
I'm so happy this is within scope 😬 should we transfer this to kedro-plugins? How do we want to handle the breaking change?

merelcht:
@astrojuanlu I'd prefer to keep it here so it can be part of the milestone. Milestones unfortunately can't be shared between repos. This change will be part of the `0.19.0` release.

PR #2500, opened 2023-04-10
Replace and deprecate `DataSet` use in class names

deepyaman:
## Description

Deprecate core "DataSet" names and warn about the impending rename to "Dataset".

Once 0.19.0 is released, the deprecated aliases (and even the utils, if so desired) can be removed.

## Checklist


merelcht:
@deepyaman I'm not sure it actually makes sense to the rename on the Kedro framework side. It's a breaking change and we're removing all datasets from here when releasing `0.19.0` anyway.

deepyaman:
> @deepyaman I'm not sure it actually makes sense to the rename on the Kedro framework side. It's a breaking change and we're removing all datasets from here when releasing `0.19.0` anyway.

@merelcht Won't the datasets in `kedro.io.core` stay in core framework, even in 0.19.0? Looking at https://github.com/kedro-org/kedro/issues/2129, are you suggesting that we would keep `DataSet` in the core repo and `Dataset` in plugins?

merelcht:
Ignore me, I should've had my morning coffee before commenting 😂 

We should rename all of them, but it will be a breaking change so maybe it's better to wait until we start work on `0.19.0` to make sure `develop` and `main` don't diverge too much until then.

merelcht:
If we do a deprecation warning what will be users see and how often? It might be annoying if there's nothing they can do about it..

deepyaman:
> If we do a deprecation warning what will be users see and how often? It might be annoying if there's nothing they can do about it..

If this functionality were to be released, they would get a warning if they used `MemoryDataSet` (or `pandas.CSVDataSet` in an updated `kedro-datasets`); but they could change their code to `MemoryDataset` (or `pandas.CSVDataset`) and avoid the warning. So the idea would be, yes, they can do something about it, and they have a grace period until it's removed that their projects wouldn't be broken on a newer version.

kedro/utils.py
```diff
@@ -26,3 +27,64 @@ def load_obj(obj_path: str, default_obj_path: str = "") -> Any:
...
+
+        if alias is not None:
+
+            def new(cls, *args, **kwargs):
+                alias = getattr(cls, "_DeprecatedClassMeta__alias")
+
+                if alias is not None:
+                    warn(
+                        f"{cls.__name__} has been renamed to {alias.__name__}, and the "
+                        f"alias will be removed in the future",
```
merelcht, line 49:
```suggestion
                        f"alias will be removed in the Kedro 0.19.0",
```

kedro/utils.py
```diff
@@ -26,3 +27,64 @@ def load_obj(obj_path: str, default_obj_path: str = "") -> Any:
...
+
+        fixed_bases = []
+
+        for b in bases:
+            alias = getattr(b, "_DeprecatedClassMeta__alias", None)
+
+            if alias is not None:
+                warn(
+                    f"{b.__name__} has been renamed to {alias.__name__}, and the alias "
+                    f"will be removed in the future",
```
merelcht, line 67:
```suggestion
                    f"will be removed in Kedro 0.19.0",
```

kedro/utils.py
deepyaman: @merelcht Do we have another release (or few) before 0.19.0? With this change, we could also warn in 0.19.x at drop in 0.20.0 (or something), if there's any reason to give more time.

If there is no release before 0.19.0, then of course doesn't really help to have this.

merelcht: Yes we'll definitely have a couple more releases before `0.19.0`.

merelcht:
@deepyaman Yes, let's keep this simple and only do the non-abstract classes. When we're working on `0.19.0` we can properly rename everything here and in `kedro-datasets`.

RELEASE.md
```diff
@@ -162,7 +162,7 @@ Many thanks to the following Kedroids for contributing PRs to this release:
 * Added support for `tf.device` in `TensorFlowModelDataset`.
 * Updated error message for `VersionNotFoundError` to handle insufficient permission issues for cloud storage.
 * Updated Experiment Tracking docs with working examples.
-* Updated MatplotlibWriter Dataset, TextDataset, plotly.PlotlyDataSet and plotly.JSONDataSet docs with working examples.
+* Updated MatplotlibWriter Dataset, TextDataSet, plotly.PlotlyDataSet and plotly.JSONDataSet docs with working examples.
```
deepyaman, line 165:
```suggestion
* Updated `MatplotlibWriter`, `text.TextDataSet`, `plotly.PlotlyDataSet` and `plotly.JSONDataSet` docs with working examples.
```

stichbury:
You are truly the patron saint of consistency @deepyaman and I'm happy to approve this change because it seems wholesale and straightforward. However, I'm not able to speak for @merelcht and @idanov in terms of whether it's the right point to merge this change or the implications thereof. But if they're happy with the intent, I'm happy with the execution.

merelcht (approved):
Thanks for persisting with this @deepyaman ! I'm happy for this to get merged ⭐

stichbury (approved):
If it's OK with @merelcht it's a 👍 from me too.

kedro/utils.py
```diff
@@ -26,3 +27,65 @@ def load_obj(obj_path: str, default_obj_path: str = "") -> Any:
...
+
+        if alias is not None:
+
+            def new(mcs, *args, **kwargs):
+                alias = getattr(mcs, "_DeprecatedClassMeta__alias")
+
+                if alias is not None:
+                    warn(
+                        f"{mcs.__name__} has been renamed to {alias.__name__}, and the "
+                        f"alias will be removed Kedro 0.19.0",
```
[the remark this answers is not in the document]
deepyaman: Ah, thanks! I fixed it in a separate place but forgot to add it back in here. :)

astrojuanlu:
RTD failing with:

```
sphinx.errors.SphinxWarning: /home/docs/checkouts/readthedocs.org/user_builds/kedro/checkouts/2500/kedro/extras/datasets/api/api_dataset.py:docstring of kedro.io.core.AbstractDataSet.exists:1:py:exc reference target not found: DatasetError
```

PR #2514, opened 2023-04-13
Fix CI

astrojuanlu:
It seems like some of the expected Circle CI checks were dropped in https://github.com/kedro-org/kedro/pull/2459/files#diff-71e295330550f17d011d6354c7e2c231e9f4a43cdbf92e511ffa0f15cea7ae60

This PR fixes it.

### DIFF
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -516,7 +516,7 @@ jobs:
 workflows:
   version: 2.1
 
-  lint_docs_only:
+  lint_only:
     when:
       and:
         - <<pipeline.parameters.docs_change>>
@@ -533,7 +533,7 @@ workflows:
           requires:
             - lint
 
-  build_code_and_docs:
+  build_code:
     when:
       and:
         - <<pipeline.parameters.code_change>>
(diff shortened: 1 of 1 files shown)


.circleci/continue_config.yml
```diff
@@ -516,6 +516,23 @@ jobs:
 workflows:
   version: 2.1
 
+  lint_docs_only:
```
merelcht, line 519: This might be slightly misleading, because afaik the linting jobs run on both the code and docs. So this should probably just be `lint_only`

.circleci/continue_config.yml
```diff
@@ -516,6 +516,23 @@ jobs:
...
+    jobs:
+      - lint:
+          matrix:
+            parameters:
+              python_version: ["3.7", "3.8", "3.9", "3.10"]
+      - all_circleci_checks_succeeded:
+          requires:
+            - lint
+
   build_code_and_docs:
```
merelcht, line 536: This doesn't build docs anymore right? Again might be good to rename for clarity.

astrojuanlu: Oh yep, on it

merelcht (approved):
👍

jmholzer (approved):

PR #2515, opened 2023-04-13
Revert aggressive requirements bumps by Dependabot

### LINKED ISSUE #2511: Dependabot has been aggressively bumping the lower bound on dependencies
https://github.com/kedro-org/kedro/pulls?q=is%3Apr+is%3Aclosed+author%3Aapp%2Fdependabot

I've never looked into these until today, but the vast majority of the Dependabot PRs have been unnecessarily restricting the lower bound? Like, just glancing through it, we now require Rich 13.3; every time a minor version comes out, we make that our lower bound?

As an example, any Typer-based plugins may now be incompatible with Kedro, if they use the `typer[all]` requirement, because they also follow a questionable (Kedro-like :P) policy of restricting the upper bound, and maintainers haven't yet allowed Rich 13.x even though it works.

Just using Rich as an example, but I'm pretty sure the vast majority of these Dependabot PRs over the last 6 months should be undone/refactored...

I think:

* `toposort ~=1.10` -> `toposort ~=1.5` (even the comment in the requirements file indicates it should be 1.5)
* `pip-tools ~=6.12` -> `pip-tools ~=6.5` (at least)
* `rope ~=1.7.0` -> `rope >=0.21, <2.0` or something (less certain on this, since it's bumped such a wide range and with such a weird policy)
* `fsspec` bumps seem fine, although there are separate questions about
  * why do we need to separate the ranges for Python 3.7 and Python >=3.8 in this case?
  * I think it's a pretty bad idea to only allow up to latest `fsspec` since it releases every month, unless you want to release monthly just to enable users to use newer `fsspec`
* `rich ~=13.3` -> `rich >=12.0, <14.0` at least (I haven't looked into why it was bounded at `~=12.0` to start with, but this would be a huge benefit regardless)
* `more_itertools ~=9.1` -> `more-itertools ~=9.0`

Happy to make a PR to make these changes, but I think the Dependabot configuration should also be revisited?

deepyaman:
## Description

Resolves #2511 

## Checklist

### DIFF
--- a/dependency/requirements.txt
+++ b/dependency/requirements.txt
@@ -4,20 +4,19 @@ cachetools~=5.3
 click<9.0
 cookiecutter>=2.1.1, <3.0
 dynaconf>=3.1.2, <4.0
-fsspec>=2021.4, <=2023.1.0; python_version == '3.7'
-fsspec>=2023.1, <2023.5; python_version >= '3.8'
+fsspec>=2021.4, <2024.1
 gitpython~=3.0
 importlib-metadata>=3.6; python_version >= '3.8'
 importlib_metadata>=3.6, <5.0; python_version < '3.8'  # The "selectable" entry points were introduced in `importlib_metadata` 3.6 and Python 3.10. Bandit on Python 3.7 relies on a library with `importlib_metadata` < 5.0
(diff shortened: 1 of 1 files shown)


noklam (approved):
Looks good.

@SajidAlamQB If you know it's possible to change the settings so dependabot are not bumping lower bound aggressively?

Issue #2078, opened 2022-11-29
Follow up and investigate the Linting Error


noklam:
## Background
Short description of the problem here.
#2068 This PR fixed a dependencies issue that blocks CI. While I was working on the issue, another linting error pop up which is irrelevant to the issue I was trying to fix. I did a quick fix with `# noqa`, but it would be great to understand why this error shows up and if is it a valid error.

From my understanding, the original error is complaining the `Set` is imported but not used. It is only used as comments (Probably for flake8). More importantly, why are we using `flake8` for type checking? Would `mypy` actually be enough?

## Follow-ups
1. Check if the type-checking could be replaced by `mypy` 
2. Investigate why the error show up, is it a valid error?

Either 1/2 is enough to solve this issue, so the person who pick up this task should evaluate which option is enough.

Side note:
We used a few linters and they have overlapping features. It would be great to understand which features do we actually need for a specific linter/formatter. In this case, personally, I would prefer just having `mypy` for static type checking, unless `flake8` actually did something more.

## Steps to Reproduce
1. Run `pre-commit run flake8 -a --hook-stage manual` on GitPod (For some reason I don't get the error on my Macbook)

cc @merelcht @jmholzer

PR #2538, opened 2023-04-25
Remove type comment in favor of PEP 526 for variable annotation 

merelcht:
## Description
Addresses https://github.com/kedro-org/kedro/issues/2078

## Development notes
The problem found in #2078 was caused by the following: https://github.com/PyCQA/pyflakes/issues/747#issuecomment-1326939923 Essentially, `flake8` doesn't support type checking in comments anymore with the `# type: ...` syntax. Instead it now looks for the PEP 526 syntax for variable annotation: https://peps.python.org/pep-0526/

## Checklist


astrojuanlu (commented):
Thanks @merelcht! I was always a fan of `type: ` comments, I guess Python is solidly heading towards types as annotations...

Some questions:

- What's the role of `flake8` in typing? Does it replace MyPy, or it only checks that the annotations are correct?
- May I suggest using `import typing as t`? (Silly personal preference, but I think it's nice to avoid polluting the global namespace with typing stuff, and instead I prefer to do `t.Any`, `t.TypedDict` and such)
- May I suggest adding `from __future__ import annotations` to leverage [PEP 585](https://peps.python.org/pep-0585/) and hence be able to do `list[str]` instead of `t.List[str]`? (Available from Python 3.7)
- Where it makes sense for this PR (without having to overhaul the whole codebase) may I suggest using `str | None` instead of `Optional[str]` and `str | bytes` instead of `Union[str, bytes]`?

astrojuanlu:
I see, thanks! Okay with keeping `from typing import *`, I had to try 😸 

About PEP 604, I'm not sure, it's not clearly written in the PEP 🤔

merelcht:
I'll create a new PR to leverage [PEP 585](https://peps.python.org/pep-0585/).

astrojuanlu (approved):
LGTM!

noklam (approved):
Looks good to me!

PR #2540, opened 2023-04-26
Leverage PEP-585

merelcht:
## Description
Follow up on https://github.com/kedro-org/kedro/pull/2538

## Checklist


merelcht:
@astrojuanlu any clue on why read the docs is failing? 🤔

astrojuanlu:
Sphinx:

```
sphinx.errors.SphinxWarning: /home/docs/checkouts/readthedocs.org/user_builds/kedro/checkouts/2540/kedro/framework/cli/pipeline.py:docstring of kedro.framework.cli.pipeline.PipelineArtifacts:1:py:class reference target not found: Path
```

merelcht:
> Sphinx:
> 
> ```
> sphinx.errors.SphinxWarning: /home/docs/checkouts/readthedocs.org/user_builds/kedro/checkouts/2540/kedro/framework/cli/pipeline.py:docstring of kedro.framework.cli.pipeline.PipelineArtifacts:1:py:class reference target not found: Path
> ```

This [Try adding more classes to ignore in nitpick](https://github.com/kedro-org/kedro/pull/2540/commits/677ccb9b4b59313ad7fe29746fe0850a214077e5) seems to have fixed it. Is that the right thing to do though? @astrojuanlu

astrojuanlu:
Interesting - Initially I just copy pasted the error without looking, but here's the code:

https://github.com/kedro-org/kedro/blob/677ccb9b4b59313ad7fe29746fe0850a214077e5/kedro/framework/cli/pipeline.py#L37-L42

There is no docstring there, but I think I know what's happening - it's being autogenerated by https://github.com/tox-dev/sphinx-autodoc-typehints/.

This issue could take some effort to debug and we are not even on the stable version because of some other bug, so I'd say your solution is okay @merelcht 👍🏽

astrojuanlu (approved):
🔥

jmholzer (commented):
Seems comprehensive. I'll double-check that we picked everything up, but for now I just have a few comments about use of `Mapping`.

kedro/framework/cli/micropkg.py
```diff
@@ -7,7 +8,7 @@
 import tempfile
 from importlib import import_module
 from pathlib import Path
-from typing import Iterable, List, Optional, Set, Tuple, Union
+from typing import Iterable, List, Tuple, Union
```
jmholzer, line 11:
```suggestion
from typing import Iterable
```

kedro/framework/cli/micropkg.py
jmholzer, line 1: I can't add a suggestion as you did not modify it, but on line 711 I believe you can change `_SourcePathType = Union[Path, List[Tuple[Path, str]]]` to `_SourcePathType = Path | list[tuple[Path, str]]]`.

jmholzer (commented):
Nice work ⭐️.

I found one instance where you haven't leveraged PEP 585 and the PEP 604 `|` syntax. Was this on purpose?

kedro/framework/cli/utils.py
```diff
@@ -14,7 +16,7 @@
 from importlib import import_module
 from itertools import chain
 from pathlib import Path
-from typing import Dict, Iterable, List, Mapping, Sequence, Set, Tuple, Union
+from typing import Iterable, Mapping, Sequence
```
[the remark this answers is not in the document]
merelcht: Great catch! 🦅 I'll update it.

kedro/framework/cli/micropkg.py
merelcht: For some reason `mypy` doesn't like this. Changing the `Union` to `|` gives this errors locally when running `make lint`:
`kedro/framework/cli/micropkg.py:711: error: Unsupported left operand type for | ("Type[Path]")`

And as you can see the build fails for the `list` and `tuple` use: https://app.circleci.com/pipelines/github/kedro-org/kedro/21475/workflows/c4793a04-5d4e-402f-a486-e24a291b807d/jobs/250854

astrojuanlu: Unfortunately the compact syntax doesn't work for type assignments, only for type declarations... Or so I think

jmholzer (approved):
Great! Super comprehensive and complete work 🙂.

Issue #2508, opened 2023-04-12
Prototype parsing rules for dataset factories


merelcht:
## Description
Subtask of https://github.com/kedro-org/kedro/issues/2423

## Context
For the dataset factories solution we'll need a way to parse the syntax to match datasets against. 
Parse (https://github.com/r1chardj0n3s/parse) is a library that has a pattern-matching syntax which uses reverse Python f-strings format. 

Example of what our syntax could be:
```yaml
"{root_namespace}.{*}@{spark}":
  type: spark.SparkDataSet
  filepath: data/{chunks[0]}/{chunks[1]}.parquet
  file_format: parquet
```

And the function that would create the dataset entry:
```python
def create_spark_dataset(dataset_name: str, *chunks):
    # e.g. here chunks=["root_namespace", "something-instead-the-*", "spark"]
    return  SparkDataSet(filepath=f"data/{chunks[0]}/{chunks[1]}.parquet", file_format="parquet")
```

## Things to investigate
The goal of this ticket is to experiment with that library and come up with rules on how the matching should work. 

It's especially important to determine what needs to happen when two patterns would match a datasets. E.g. `france.companies@spark` against:

```yaml
"{root_namespace}.{dataset_name}@spark":
  type: spark.SparkDataSet
  filepath: data/{root_namespace}/{dataset_name}.parquet
  file_format: parquet

"{dataset_name}@spark":
  type: spark.SomethingElse
  filepath: data/{dataset_name}.parquet
  file_format: parquet
```

Pattern matching is a common problem in url matching, so take inspiration from web frameworks who solve that problem, e.g. Ruby/React. 

## Initial thoughts on rules
1. The most explicit pattern should match first
2. Then match alphabetically (?)

noklam:
I did some investigation about how Ruby/Django/React doing with their Router, they all differ in slightly way but it follows a clear declaration order. i.e. for Django you have a `view.py` which has a list of routes, the first match will win.

antonymilne:
One heuristic for "most specific pattern should match first" would be to count the number of `{}` in the expression (in fact, off the top of my head I can't think of any other way to do it).

[...]

PR #2559, opened 2023-05-05
[DRAFT] Dataset factory parsing rules demo

ankatiyar:
## Description

Related to https://github.com/kedro-org/kedro/issues/2508 
This PR is just to demonstrate/experiment with parsing rules - not the actual dataset factory prototype. 

## Development notes

I've added a `kedro catalog resolve` (Similar to what @merelcht did in https://github.com/kedro-org/kedro/pull/2560)
For each named dataset used by the pipelines ->
It loops over all the catalog entries, finds patterns that would be a match and then selects the best match pattern (`pick_best_match` function)

### Problem Statement 
The `matches` list contains all the patterns that match the dataset name - for example for `france.companies`, all the patterns listed below would be a match -
```yaml
# Catalog Entry 1
"{namespace}.{dataset}":
  type: pandas.CSVDataSet
  filepath: path1/{namespace}_{dataset}.csv

# Catalog Entry 2
"{dataset}":
  type: pandas.CSVDataSet
  filepath: path2/{dataset}.csv

# Catalog Entry 3
"{namespace}.{a}":
  type: pandas.CSVDataSet
  filepath: path3/{namespace}_{a}.csv

# Catalog Entry 4
"{namespace}.companies":
  type: pandas.CSVDataSet
  filepath: path4/{namespace}_companies.csv

# Catalog Entry 5
france.companies:
  type: pandas.CSVDataSet
  filepath: path5/france_companies.csv

# Catalog Entry 6
"{dataset}s":
  type: pandas.CSVDataSet
  filepath: path6/{dataset}s.csv

```
We need to decide the ranked order of these entries to find the best match.

### Proposed Solution 
* Sort according to "specificity" (for the lack of a better word?) - The number of characters outside of the brackets that match. The order then would be ->
`#5 -> #4 -> #6/ #3/ #1 -> #2`
*  When "specificity" is the same - 
    * sort them according to the number of brackets? Between  `#6`, `#3` & `#1` - pick the highest number of bracket pairs. Example: `{namespace}.{dataset}` would be chosen over `{dataset}s`. 
    * either after sorting them according to bracket pairs or instead we can also sort them alphabetically. Example: `f{*}` should rank above `{*}s`

## Notes / Challenges

* When the dataset entry is too generic, for example : 
``` yaml
{dataset}:
  type: pandas.CSVDataSet
  filepath: path/path
```
The datasets that are supposed to be default/`MemoryDataSet` and parameters (eg. `params:model_input`) will falsely match against the pattern. This will lead to runtime errors and we will need to document this.

* Catalog entry 1 & 3 are basically the same thing and we could check for duplicating patterns and disallow it.

## Checklist

### DIFF
--- a/kedro/framework/cli/catalog.py
+++ b/kedro/framework/cli/catalog.py
@@ -179,7 +179,7 @@ def _add_missing_datasets_to_catalog(missing_ds, catalog_path):
 
 
 def pick_best_match(matches):
-    matches = sorted(matches, key=lambda x: (specificity(x[0]), x[0]))
+    matches = sorted(matches, key=lambda x: (specificity(x[0]), -x[0].count("{"), x[0]))
     return matches[0]
 
 
--- a/kedro/framework/cli/catalog.py
+++ b/kedro/framework/cli/catalog.py
@@ -179,9 +179,22 @@ def _add_missing_datasets_to_catalog(missing_ds, catalog_path):
 
 
 def pick_best_match(matches):
-    # according to number of { then alphabetical
-    matches = sorted(matches, key=lambda x: (x[0].count("}"), x[0]))
+    matches = sorted(matches, key=lambda x: (specificity(x[0]), x[0]))
     return matches[0]
+
+
+def specificity(pattern):
+    """This function will check length of exactly matched characters not inside brackets
+    Example -
+    specificity("{namespace}.companies") = 10
+    specificity("{namespace}.{dataset}") = 1
+    specificity("france.companies") = 16
+    """
+    pattern_variables = parse(pattern, pattern).named
+    for k in pattern_variables:
+        pattern_variables[k] = ""
+    specific_characters = pattern.format(**pattern_variables)
+    return -len(specific_characters)
 
 
 @catalog.command("resolve")
--- a/kedro/framework/cli/catalog.py
+++ b/kedro/framework/cli/catalog.py
@@ -1,9 +1,11 @@
 """A collection of CLI commands for working with Kedro catalog."""
+import copy
 from collections import defaultdict
 
 import click
 import yaml
 from click import secho
+from parse import parse
 
 from kedro.framework.cli.utils import KedroCliError, env_option, split_string
 from kedro.framework.project import pipelines, settings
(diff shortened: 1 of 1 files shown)


kedro/framework/cli/catalog.py
```diff
@@ -174,3 +176,64 @@ def _add_missing_datasets_to_catalog(missing_ds, catalog_path):
...
     with catalog_path.open(mode="w") as catalog_file:
         yaml.safe_dump(catalog_config, catalog_file, default_flow_style=False)
+
+
+def pick_best_match(matches):
+    matches = sorted(matches, key=lambda x: (specificity(x[0]), -x[0].count("{"), x[0]))
+    return matches[0]
+
+
+def specificity(pattern):
```
merelcht, line 186: Is this method just counting the number of characters that are not in brackets? Wouldn't it be easier to just remove the text in {} and then just get the length of the string?

kedro/framework/cli/catalog.py
```diff
@@ -174,3 +176,64 @@ def _add_missing_datasets_to_catalog(missing_ds, catalog_path):
...
+        if len(matches) == 0:
+            # print(f"skipping {pipeline_dataset} -> maybe params or MemoryDataSet")
+            continue
+        best_match, result = pick_best_match(matches)
+        best_match_config = copy.deepcopy(catalog_conf[best_match])
+        # Match results to patterns in best matching catalog entry
+        for key, value in best_match_config.items():
+            string_value = str(value)
+            try:
+                formatted_string = string_value.format_map(result.named)
```
merelcht, line 233: This should only happen for dataset fields with `{}` otherwise you get weird things like: 
```
shuttles:
  load_args: data/01_raw/shuttles.xlsx
  type: pandas.ExcelDataSet
```

But that's not really part of the matching logic, so not super important here.

antonymilne:
I think these rules look generally good 👍  I also think we need all three of them, and the order you suggest them is right, i.e. count non-bracketed characters, then count number of brackets, then sort alphabetically.

We shouldn't need to consider the specificity of your example #5 (`france.companies`) at all because it doesn't have `{}` in it. This isn't a pattern and so if you `catalog.load` a dataset with this name you're not interested in looking at patterns that match.

Also I don't think it's a problem that #1 and #3 are the same or that we need to go through and remove duplicates. These mean exactly the same thing so we don't really care how they rank relative to each other. In practice they would be distinguished based on the final alphabetical order, and then it will always be the same one that gets matched against first and the other one will just get ignored.

noklam:
I also like the fact that the rule is simple. I suspect there will be edge cases but I am not too nervous about it. In general we should encourage people to write simple catalog, if it requires very obscure rules it's likely that they can just simplify it by changing the structure.


One missing example that I can think of is whether we should consider namespace as a priority.

Consider the name France.companies

Which pattern will win? 
* France.{something}
* {namespace}.companies

Previous conversation favor the former one as it's more specific about namespace. This however require some more sophisticated rule. It shouldn't be hard tho, we can simply split on the dot and derive the score from there.

antonymilne:
Hmm, that's an interesting example about the namespaces. I haven't thought it through fully, but I can't immediately think of a rule that would would order these the way you want without it being quite complex though? To me it's also not immediately obvious that `france.{something}` should necessarily beat `{namespace}.companies` in the first place. So I think it's probably fine to not handle namespaces using any special logic and just treat `.` like any other character. I suspect it will just get too complex otherwise for not that much benefit.

antonymilne:
Something else I just thought of. Let's say you have these patterns:
```
1. italy.{dataset}
2. france.{dataset}
3. switzerland.{dataset}
4. {namespace}.companies
```

Then the order of these would be 3 > 4 > 2 > 1 which is not really obvious, just because "switzerland" is a longer word than "companies".

Maybe it would be less arbitrary to count number of unbracketed "chunks" outside brackets rather than number of characters? This would rank all the equally so would then need another rule to order them. Or maybe we should consider the order of where unbracketed chunks appear relative to bracketed ones in the pattern? Or maybe this is just an edge case that won't come up in practice and we shouldn't care about it?

Note this is nothing specific to do with namespacing: if you replace `.` with `_` in all the above examples then the same arguments hold. So I still think we should probably not treat `.` specially in any way.

noklam:
I agree we don't need to treat the namespace specially, however for a good convention it's most likely that people should split it using some kind of delimeter for these pattern.


It's also important to note that the patterns aren't mean for refactoring, there are still value to create explicit entry, it's probably not worth it to use a pattern to just replace two or three similar entries. (same apply for Jinja)


Can we try to do this exercise with a realistic example? I think @merelcht example could be useful for documentation anyway, so it won't be a waste. It would be interesting to have at least two people to go through the same exercise and see what's the difference. 

I see the main use cases are two:
1. Shorten the catalog with long repetitive entries (maybe 5 or 10, it varies because you will most likely use it with variable interpolation ) 
2. Enumerating is not possible, where you have to loop through a dynamic list.

noklam:
Just checking with my understanding, for 2. with jinja you will loop through the list inside the YAML file. In contrast, with Dataset Factory you will loop through this directly inside the nodes file instead?

kedro/framework/cli/catalog.py
ankatiyar: This can be changed to only match with patterns containing `{}` but this logic will still work for dataset entries that are not patterns(exact matches) because of how `parse` works -
`parse("xyz", "abc")` returns `None` (not a match at all - ignore)
`parse("xyz", "xyz")` returns `<Result () {}>` (exact match, not a pattern - empty result)
`parse("hello {name}", "hello world")` returns `<Result () {'name': 'world'}>` (pattern matching)

So the exact match then gets chosen as the best match and since it wouldn't contain any brackets anywhere in the catalog entry, it would not be changed at all.

ankatiyar:
My assumption is that any user ideally would not have a lot of "dataset patterns" that would match a given dataset name. As long as we have a set of simple + deterministic rules to order the patterns, we could leave it up to the user to deal with cases where multiple matches exist.

kedro/framework/cli/catalog.py
ankatiyar: That is what is happening here but using `parse` instead of a regex.

ankatiyar:
Closing in favour of #2635

Issue #2423, opened 2023-03-15
Eliminate the need of having all datasets defined in the catalog


merelcht:
## Description
Users struggle with ever increasing catalog files:
https://github.com/kedro-org/kedro/issues/891
https://github.com/kedro-org/kedro/issues/1847

Current "solutions":
- Use yaml anchors: https://www.educative.io/blog/advanced-yaml-syntax-cheatsheet#anchors
- The `TemplatedConfigLoader`
- Use the `TemplatedConfigLoader` with Jinja2: https://docs.kedro.org/en/stable/kedro_project_setup/configuration.html#jinja2-support

Instead of allowing anchors or loops in config let's just tackle the need of having so many datasets defined in the catalog in the first place. 

## Context
The new `OmegaConfigLoader` doesn't currently allow the use of yaml anchors (the omegaconf team discourages people to use this with omegaconf: https://github.com/omry/omegaconf/issues/93) or Jinja2. While Jinja2 is very powerful, we've always had the standpoint as a team that it's better to not use it, because it makes files hard to read and reason about. Jinja2 is meant for HTML templating, not for configuration. 

In order to make `OmegaConfigLoader` a suitable replacement for the `ConfigLoader` and `TemplatedConfigLoader` we need to find another solution to reduce the size of catalog files. 

## Possible Implementation
This is an idea from @idanov (written up by me, so I hope I have interpreted it correctly 😅 ) :

In every runner we have a method to create a default dataset for datasets that are used in the pipeline but not defined in the catalog:

https://github.com/kedro-org/kedro/blob/8acfb2a67049cce8f8d9126fc2373cb40e26b0d8/kedro/runner/sequential_runner.py#L32-L43

We can allow users to define what a "default" dataset should be so that in the catalog they only need to specify these default settings and Kedro would then create those without needing them to be defined in the catalog. The default dataset definition could then look like something like this:

```yaml
default_dataset: 
  type: spark.SparkDataSet
  file_format: parquet
  load_args:
    header: true
  save_args:
    mode: overwrite
```

And any dataset used in the pipeline, but not mentioned in the catalog, would get the above specifications. 

### Extra info
I collected several catalog files from internal teams to see what patterns I could find in there. I think these will be useful when designing our solution. https://miro.com/app/board/uXjVMZM7IFY=/?share_link_id=288914952503

datajoely:
Would you only allow a singular `default_dataset`?

deepyaman:
[...]

PR #2560, opened 2023-05-05
[PROTOTYPE NOT TO BE MERGED] Dataset factories prototype

merelcht:
## Description
Prototype for the dataset factories proposal in https://github.com/kedro-org/kedro/issues/2423
This doesn't include any sophisticated parsing rules, but just uses the basic `parse()` method from https://github.com/r1chardj0n3s/parse#readme. 

## Development notes

- When the `DataCatalog` class instance gets instantiated from the config in: https://github.com/kedro-org/kedro/blob/807c4e64fb/kedro/framework/context/context.py#L287, it now checks for entries that are patterns (they contain "}" ) and skips those from dataset creation and instead adds those to a new "dataset_patterns" dictionary that's kept on the catalog. 
- Any datasets that match a pattern get resolved on load time and then added to the catalog as proper dataset instances. So resolution only has to happen once and not at every access.
- I'm not super happy yet with `remove_pattern_matches()` this is a helper method to exclude any datasets that match a pattern from the runner checks for missing inputs, free outputs and unregistered datasets that need a default dataset. 
- **Bonus:** Added two CLI methods (these are really ugly code, but just to show what could be possible): `kedro catalog show` prints the raw catalog config files to the CLI. `kedro catalog resolve` resolves any patterns in the catalog config file against datasets used in the pipeline and prints this to the CLI.

**Test project** that I've used to test the prototype: https://github.com/merelcht/dataset-factories

## Checklist

### DIFF
--- a/kedro/framework/cli/catalog.py
+++ b/kedro/framework/cli/catalog.py
@@ -184,31 +184,21 @@ def _add_missing_datasets_to_catalog(missing_ds, catalog_path):
 def show_catalog_datasets(metadata: ProjectMetadata, env):
     session = _create_session(metadata.package_name, env=env)
     context = session.load_context()
-    catalog = context.config_loader["catalog"]
-    secho(yaml.dump(catalog))
+    catalog_conf = context.config_loader["catalog"]
+    secho(yaml.dump(catalog_conf))
 
 
 @catalog.command("resolve")
 @env_option
-@click.option(
-    "--pipeline",
-    "-p",
-    type=str,
-    default="",
-    help="Name of the modular pipeline to run. If not set, "
-    "the project pipeline is run by default.",
-    callback=split_string,
-)
 @click.pass_obj
-def resolve_catalog_datasets(metadata: ProjectMetadata, pipeline, env):
+def resolve_catalog_datasets(metadata: ProjectMetadata, env):
     session = _create_session(metadata.package_name, env=env)
     context = session.load_context()
-    catalog = context.config_loader["catalog"]
+    catalog_conf = context.config_loader["catalog"]
 
-    target_pipelines = pipeline or pipelines.keys()
-
+    # Create a list of all datasets used in the project pipelines.
     pipeline_datasets = []
-    for pipe in target_pipelines:
+    for pipe in pipelines.keys():
         pl_obj = pipelines.get(pipe)
         if pl_obj:
             pipeline_ds = pl_obj.data_sets()
@@ -220,7 +210,11 @@ def resolve_catalog_datasets(metadata: ProjectMetadata, pipeline, env):
                 f"'{pipe}' pipeline not found! Existing pipelines: {existing_pls}"
             )
 
-    catalog_copy = copy.deepcopy(catalog)
+    # Create a copy of the catalog config to not modify the original.
+    catalog_copy = copy.deepcopy(catalog_conf)
+    # Loop over all entries in the catalog, find the ones that contain a pattern to be matched,
+    # loop over al datasets in the pipeline and match these against the patterns.
+    # Then expand the matches and add them to the catalog copy to display on the CLI.
     for ds_name, ds_config in catalog.items():
         if "}" in ds_name:
             for pipeline_dataset in set(pipeline_datasets):
--- a/kedro/framework/cli/catalog.py
+++ b/kedro/framework/cli/catalog.py
@@ -5,12 +5,13 @@ from collections import defaultdict
 import click
 import yaml
 from click import secho
+from parse import parse
 
 from kedro.framework.cli.utils import KedroCliError, env_option, split_string
 from kedro.framework.project import pipelines, settings
 from kedro.framework.session import KedroSession
 from kedro.framework.startup import ProjectMetadata
-from parse import parse
+
 
 def _create_session(package_name: str, **kwargs):
     kwargs.setdefault("save_on_close", False)
@@ -180,11 +181,12 @@ def _add_missing_datasets_to_catalog(missing_ds, catalog_path):
 @catalog.command("show")
 @env_option
 @click.pass_obj
-def list_datasets(metadata: ProjectMetadata, env):
+def show_catalog_datasets(metadata: ProjectMetadata, env):
     session = _create_session(metadata.package_name, env=env)
     context = session.load_context()
     catalog = context.config_loader["catalog"]
     secho(yaml.dump(catalog))
+
 
 @catalog.command("resolve")
 @env_option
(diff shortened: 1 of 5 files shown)


kedro/framework/cli/catalog.py
```diff
@@ -174,3 +176,56 @@ def _add_missing_datasets_to_catalog(missing_ds, catalog_path):
...
+                result = parse(ds_name, pipeline_dataset)
+                if result:
+                    config_copy = copy.deepcopy(ds_config)
+                    # Match results to patterns in catalog entry
+                    for key, value in config_copy.items():
+                        if isinstance(value, Iterable) and "}" in value:
+                            string_value = str(value)
+                            config_copy[key] = string_value.format_map(result.named)
+                    catalog_copy[pipeline_dataset] = config_copy
+
```
idanov, line 230: I find this extremely hard to read if not for the comments. I think it would be much better if the code actually does what the comment says, step by step, rather than nesting levels.

Moreover, this complexity seems to have hidden a logical incorrectness of the code: all datasets that appear exactly as they are in the catalog, shouldn't be matched against the patterns (which the original code doesn't do).

```suggestion
    skip = [ds_name for catalog_conf.keys() if "}" not in ds_name]
    datasets = set(pipeline_datasets) - set(skip)
    patterns = [ds_name for catalog_conf.keys() if "}" in ds_name]
    matches = [(ds, pattern, parse(pattern, ds)) for ds in datasets for pattern in patterns]
    matches = [(ds, pattern, result) for ds, pattern, result in matches if result]
    for ds, pattern, result in matches:
        cfg = copy.deepcopy(catalog_conf[pattern])
        # Match results to patterns in catalog entry
        for key, value in cfg.items():
            if isinstance(value, Iterable) and "}" in value:
                string_value = str(value)
                cfg[key] = string_value.format_map(result.named)
        catalog_copy[pipeline_dataset] = cfg

```

The code for populating the template should probably be a separate function for readability and also to ensure that we can call it recursively (as each value can be a dictionary with other values, etc).

The original code does all the matching and it doesn't stop at the first match, maybe that's intended so we can see all possible matches?

kedro/io/data_catalog.py
```diff
@@ -520,6 +543,34 @@ def add_feed_dict(self, feed_dict: Dict[str, Any], replace: bool = False) -> Non
 
             self.add(data_set_name, data_set, replace)
 
+    def match_name_against_dataset_factories(self, dataset_input_name: str) -> Optional[AbstractDataSet]:
```
idanov, line 546: Crazily long name 😅

kedro/io/data_catalog.py
```diff
@@ -520,6 +543,34 @@ def add_feed_dict(self, feed_dict: Dict[str, Any], replace: bool = False) -> Non
...
+        """
+        For a given dataset name, try to match it against the dataset patterns in the catalog.
+        If it's a match, return the dataset instance.
+        """
+        dataset = None
+        # Loop through all dataset patterns and check if the given dataset name has a match.
+        for dataset_name, dataset_config in self.dataset_patterns.items():
```
idanov, line 553: The `dataset_name` here is a `pattern` and the `dataset_config` is a `template`/`factory`, right? We could use that terminology for ease and readability.

kedro/io/data_catalog.py
```diff
@@ -520,6 +543,34 @@ def add_feed_dict(self, feed_dict: Dict[str, Any], replace: bool = False) -> Non
...
+                # Match results to patterns in catalog entry
+                for key, value in config_copy.items():
+                    # Find all dataset fields that need to be resolved with
+                    # the values that were matched.
+                    if isinstance(value, Iterable) and "}" in value:
+                        string_value = str(value)
+                        # result.named: {'root_namespace': 'germany', 'dataset_name': 'companies'}
+                        # format_map fills in dict values into a string with {...} placeholders
+                        # of the same key name.
+                        config_copy[key] = string_value.format_map(result.named)
```
idanov, line 569: As above, will make this a function for readability and call it recursively down the line, returning materialised config dictionary.

kedro/io/data_catalog.py
```diff
@@ -520,6 +543,34 @@ def add_feed_dict(self, feed_dict: Dict[str, Any], replace: bool = False) -> Non
...
+                # Create dataset from catalog config.
+                dataset = AbstractDataSet.from_config(dataset_name, config_copy)
+        return dataset
```
idanov, line 572: Probably worth caching the match here before returning and adding it to `self._data_sets`.

kedro/io/data_catalog.py
```diff
@@ -565,13 +616,33 @@ def list(self, regex_search: Optional[str] = None) -> List[str]:
...
+            for dataset in dataset_list:
+                # If dataset matches a pattern, remove it from the list.
+                for dataset_name in self.dataset_patterns.keys():
+                    result = parse(dataset_name, dataset)
+                    if result:
+                        break
+                    else:
+                        dataset_list_minus_matched.append(dataset)
+            return set(dataset_list_minus_matched)
+        return dataset_list
```
idanov, line 633: This is much nicer:
```suggestion
        if not self.dataset_patterns:
                    return dataset_list

        dataset_list_minus_matched = dataset_list.copy()
        for dataset in dataset_list:
            matches = (parse(pattern, dataset) for pattern in self.dataset_patterns.keys())
            # If dataset matches any pattern, remove it from the list.
            if any(matches):
                dataset_list_minus_matched -= dataset
        return set(dataset_list_minus_matched)
```

As a general rule:

1. The shorter case in an `if` statement should **always** go first
2. If there's return, there's no need of `else` (like in the original code)
3. There's usually a nicer solution than using a `break` in a loop (not always though)

kedro/framework/cli/catalog.py
merelcht: Oh yeah you can completely ignore this code. As I wrote in the description this is just bonus for the showcase, but in no way meant to be merged. I just went for a quick implementation, because this needs to proper parsing logic and will change anyway.

kedro/io/data_catalog.py
```diff
@@ -255,6 +261,7 @@ class to be loaded is specified with the key ``type`` and their
             >>> catalog.save("boats", df)
         """
         data_sets = {}
+        dataset_patterns = {}
```
noklam, line 264: What are the difference between these two patterns?

merelcht:
Closing this prototype PR, because the proper implementation is now open: https://github.com/kedro-org/kedro/pull/2635

PR #2582, opened 2023-05-15
Revert special logic handling DictConfig

### LINKED ISSUE #2561: Revert all special conversion logic in `OmegaConfigLoader` for `DictConfig`
## Description
Short description of the problem here.

> This is nearly there but I think there's a bug with it unless I'm missing something: what happens if you have a single `parameters.yml` file and want to use `runtime_params`?
> 
> As part of a separate PR I think we should revert #2378 (but keep `test_no_DictConfig_in_store` as it's a good regression test) and #2176.

As part of the conversation in #2467, we should revert the changes that we added in #2378 and #2176 as we no longer need to deal with the special `DictConfig` object. Historically we need that because OmegaConfigLoader return DictConfig instead of a primitive `Dict`.

We also need to keep the test to make sure there is no regression and #2357 will not happen again.

## Context

noklam:
## Description

Close #2561 

## Development notes

- 45f34beeac87c899809f628efdbde642a83df3ae - Partially revert #2378, to_container is kept because we need to make sure dict is return. @ankatiyar 
  - some tests are removed because we shouldn't have any OmegaConf.DictConfig return from config
- cbe7813bd26d2f199895168321d761a5045a17e7 -  Partially revert #2176 
## Checklist

### DIFF
--- a/tests/framework/context/test_context.py
+++ b/tests/framework/context/test_context.py
@@ -12,7 +12,6 @@ import pandas as pd
 import pytest
 import toml
 import yaml
-from omegaconf import OmegaConf
 from pandas.util.testing import assert_frame_equal
 
 from kedro import __version__ as kedro_version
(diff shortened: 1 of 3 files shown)


tests/framework/context/test_context.py
```diff
@@ -486,11 +485,6 @@ def test_validate_layers_error(layers, conflicting_datasets, mocker):
             {"a": {"a.c": {"a.c.b": 4}}},
             {"a": {"a.a": 1, "a.b": 2, "a.c": {"a.c.a": 3, "a.c.b": 4}}},
         ),
-        (
-            {"a": OmegaConf.create({"b": 1}), "x": 3},
-            {"a": {"c": 2}},
-            {"a": {"b": 1, "c": 2}, "x": 3},
-        ),
```
noklam, line 493: I remove this test since it's not meaningful as we don't expect any OmegaConf object. I keep the test that checking no OmegaConf is in the params.

ankatiyar (approved):
LGTM!

antonymilne (approved):
Perfect!

PR #2584, opened 2023-05-16
`OmegaConfigLoader` should not show `MissingConfigException` when the file is empty

### LINKED ISSUE #2566: `OmegaConfigLoader` should not throw `MissingConfigException` when the file is empty
## Description
Related #2556 

Currently, if you have an empty configuration file with `OmegaConfigLoader`, you will get this error which isn't very accurate. 

This draft PR is just a starting point, it's expected to be picked up by someone and worked on it further.
https://github.com/kedro-org/kedro/blob/653145771fa440a46b36d2c96c57a5997613da5e/kedro/config/omegaconf_config.py#L195-L200

This is due to the fact that we always return an empty dictionary when we load the configuration. In other words, we cannot distinguish between an empty file and a missing file.
https://github.com/kedro-org/kedro/blob/653145771fa440a46b36d2c96c57a5997613da5e/kedro/config/omegaconf_config.py#L274-L275
We should take this as an opportunity to refactor and simplify these conditions. Bottom line we should have a better error message to handle empty configuration properly.

## Context
I was confused with the error because I have the config file. Initially I thought my configuration is not in the right place or I didn't config the `config_patterns` properly. The config file is actually read but the error is confusing.

## Expected Result
Tell us what should happen.
It should return an empty dictionary `{}` instead of getting error.
## Actual Result
Tell us what happens instead.
`MissingConfigException` is thrown.

```
-- If you received an error, place it here.
```

```
-- Separate them if you have more than one.
```

ankatiyar:
Resolves #2566 

## Development notes

When config files exist but are empty, OmegaConfigLoader should not throw `MissingConfigException` but pass on an empty dict. 
Solved this in a similar way to how it's handled in `ConfigLoader` -> throw `MissingConfigException` when `processed_files` which is a set of files read for a given type of config is empty. In cases where the file exists but is empty, empty dict `{}` is returned without an error. 

As opposed to `ConfigLoader` where the config for base and runtime envs is loaded in the same function call (`ConfigLoader.get()`-> `_get_config_from_patterns()`), `OmegaConfigLoader` calls `load_and_merge_dir_config()` twice for base and runtime env, I've declared `processed_files` outside and passed it as a parameter to `load_and_merge_dir_config` and kept the logic where we raise the exception outside of the function. 

Also added a unit test to `test_omega_config.py`

NOTE : 
I checked the merging logic in `load_and_merge_dir_config()` and we still need the following lines because it handles the case when no relevant config files exist for a given pattern + env. `aggregate_config` is `dict_values([])` which `OmegaConf` complains about while merging.
https://github.com/kedro-org/kedro/blob/16d8a7559573f331680fef13ad27e484e51764f6/kedro/config/omegaconf_config.py#L278-L279
## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -15,6 +15,7 @@
 
 ## Bug fixes and other changes
 * `OmegaConfigLoader` will return a `dict` instead of `DictConfig`.
+* `OmegaConfigLoader` does not show a `MissingConfigError` when the config files exist but are empty.
 
 ## Breaking changes to the API
 * `kedro package` does not produce `.egg` files anymore, and now relies exclusively on `.whl` files.
--- a/kedro/config/omegaconf_config.py
+++ b/kedro/config/omegaconf_config.py
@@ -162,13 +162,14 @@ class OmegaConfigLoader(AbstractConfigLoader):
 
         read_environment_variables = key == "credentials"
 
+        processed_files: set[Path] = set()
         # Load base env config
         if self._protocol == "file":
             base_path = str(Path(self.conf_source) / self.base_env)
         else:
             base_path = str(Path(self._fs.ls("", detail=False)[-1]) / self.base_env)
         base_config = self.load_and_merge_dir_config(
-            base_path, patterns, key, read_environment_variables
+            base_path, patterns, key, processed_files, read_environment_variables
         )
         config = base_config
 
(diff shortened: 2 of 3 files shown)


noklam (approved):
LGTM!

merelcht (approved):
In general I would discourage modifying input in the way `processed_files` is here, because it makes debugging very difficult. But I can't think of an elegant way to not do it here, so I'm okay merging it as is 🙂

noklam:
Agree with @merelcht here. The alternative would be making the `load_and_merge_dir_config` return a `Dict | None`, then we need to do some checking. It doesn't require an extra variable but one more condition to check.

I thought we are doing this for other config loaders so I am not against this implementation. The `processed_file` here doesn't do anything here or am I crazy?

https://github.com/kedro-org/kedro/blob/16d8a7559573f331680fef13ad27e484e51764f6/kedro/config/common.py#L69-L80


https://github.com/kedro-org/kedro/blob/16d8a7559573f331680fef13ad27e484e51764f6/kedro/config/common.py#L190-L206

noklam:
Note: We should try to break down the function, it's a huge function that build path, load config, and merge

PR #2587, opened 2023-05-17
Proposal of updated settings.py

### LINKED ISSUE #1538: Update default suggestions in `settings.py` to ones that work
## Description
Update docs and default suggestions in `settings.py`, because currently some of those suggestions don't actually work. 

Currently, the `BaseSessionStore` is the default session store. The other possible stores a user can use are the `ShelveStore` and the `SQLiteStore` (currently part of viz).

The `ShelveStore` is the default suggestion to override the default in `settings.py`, but when users are using some sort of multiprocessing this store type will not work. See: https://github.com/kedro-org/kedro/issues/1442

Also look at the other default suggestions and verify that they make sense. 

(Later consideration, but not part of this work)
If we move the `SQLiteStore` from viz to kedro core, we could add that as the default suggestion instead.

noklam:
## Description

Fix #1538 

Since @astrojuanlu ask about the latest suggestion, I just go ahead and create this PR as the change is small. If there are controversial opinions we can take it to tech design. 
## Checklist

### DIFF
--- a/kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/settings.py
+++ b/kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/settings.py
@@ -3,20 +3,20 @@ from the Kedro defaults. For further information, including these default values
 https://kedro.readthedocs.io/en/stable/kedro_project_setup/settings.html."""
 
 # Instantiated project hooks.
-# HOOKS = (,)
+# For example, after creating a hooks.py and defining a ProjectHooks class there, do
+# from {{cookiecutter.python_package}}.hooks import ProjectHooks
+# HOOKS = (ProjectHooks(),)
 
 # Installed plugins for which to disable hook auto-registration.
 # DISABLE_HOOKS_FOR_PLUGINS = ("kedro-viz",)
 
 # Class that manages storing KedroSession data.
-# from kedro_viz.integrations.kedro.sqlite_store import SQLiteStore
-# from pathlib import Path
-# SESSION_STORE_CLASS = SQLiteStore
+# from kedro.framework.session.store import BaseSessionStore
+# SESSION_STORE_CLASS = BaseSessionStore
 # Keyword arguments to pass to the `SESSION_STORE_CLASS` constructor.
 # SESSION_STORE_ARGS = {
 #     "path": "./sessions"
 # }
-
 
 # Directory that holds configuration.
 # CONF_SOURCE = "conf"
--- a/kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/settings.py
+++ b/kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/settings.py
@@ -3,29 +3,27 @@ from the Kedro defaults. For further information, including these default values
 https://kedro.readthedocs.io/en/stable/kedro_project_setup/settings.html."""
 
 # Instantiated project hooks.
-# from {{cookiecutter.python_package}}.hooks import ProjectHooks
-# HOOKS = (ProjectHooks(),)
+# HOOKS = (,)
 
 # Installed plugins for which to disable hook auto-registration.
 # DISABLE_HOOKS_FOR_PLUGINS = ("kedro-viz",)
 
 # Class that manages storing KedroSession data.
-# from kedro.framework.session.shelvestore import ShelveStore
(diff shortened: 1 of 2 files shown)


kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/settings.py
```diff
@@ -34,6 +32,10 @@
 #       }
 # }
 
+# Class that manages Kedro's library components.
+# from kedro.framework.context import KedroContext
+# CONTEXT_CLASS = KedroContext
```
noklam, line 37: I move context down because I think `DataCatalog` and `KedroContext` are unlikely to be changed.

kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/settings.py
```diff
@@ -3,29 +3,27 @@
...
 
 # Installed plugins for which to disable hook auto-registration.
 # DISABLE_HOOKS_FOR_PLUGINS = ("kedro-viz",)
 
 # Class that manages storing KedroSession data.
-# from kedro.framework.session.shelvestore import ShelveStore
-# SESSION_STORE_CLASS = ShelveStore
+# from kedro_viz.integrations.kedro.sqlite_store import SQLiteStore
+# from pathlib import Path
+# SESSION_STORE_CLASS = SQLiteStore
```
noklam, line 14: The ShelveStore isn't really useful and create a lot of confusion before. SQLiteStore is the most common one, however user need to install kedro-viz first.

kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/settings.py
```diff
@@ -3,29 +3,27 @@
 https://kedro.readthedocs.io/en/stable/kedro_project_setup/settings.html."""
 
 # Instantiated project hooks.
-# from {{cookiecutter.python_package}}.hooks import ProjectHooks
-# HOOKS = (ProjectHooks(),)
+# HOOKS = (,)
```
noklam, line 6: I don't know what's the best example here, since we don't have a built-in hook available. The strongest candidate is SparkHook but we only put it in the `pyspark-iris`.

The important think here is that you need an instance, not a class inside `HOOKS`

noklam: @astrojuanlu

astrojuanlu: What about this?

```
# For example, after creating a hooks.py and defining a ProjectHooks class there, do
# from {{cookiecutter.python_package}}.hooks import ProjectHooks
# HOOKS = (ProjectHooks(),)
```

so we keep the recommendation, but we add a bit more context to it.

kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/settings.py
```diff
@@ -3,29 +3,27 @@
...
-# from kedro.framework.context import KedroContext
-# CONTEXT_CLASS = KedroContext
 
 # Directory that holds configuration.
 # CONF_SOURCE = "conf"
 
 # Class that manages how configuration is loaded.
-# CONFIG_LOADER_CLASS = ConfigLoader
+# from kedro.config import OmegaConfigLoader
+# CONFIG_LOADER_CLASS = OmegaConfigLoader
```
astrojuanlu, line 26: 💯

astrojuanlu (commented):
Left some suggestions!

kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/settings.py
noklam: I think it's a sensible choice

merelcht (approved):
Thanks for doing this @noklam ⭐

astrojuanlu (approved):
⭐ Thanks for the quick turnaround @noklam!

astrojuanlu:
Noooooo I updated with rebase and now some commits are not signed-off :rage1: Sorry folks, will amend this locally.

noklam:
Let's also make the changes in starters

Issue #1733, opened 2022-07-28
difficulty to read logs with Jupyter (VSCode Dark theme) and new logging


Debbby57:
## Description
I'm working on VSCode, and I custom my vscode with the Dark+ color theme. When I use interactive python to work it's really diffcult to read logs because of colors. I need to select them to be able to read them correctly. I notice that the colors are slightly different in interactive python and in terminal.

## Steps to Reproduce
1. use Vscode and  Dark+ color theme 
2. install kedro 0.18.2
3. some code like : 
 ``` python
from pathlib import Path
from kedro.framework.session import KedroSession
from kedro.framework.startup import bootstrap_project
my_path = Path(__file__).resolve().parent.parent 
metadata = bootstrap_project(my_path)
session = KedroSession.create(metadata.package_name, project_path=my_path)
local_context = session.load_context()
my_data= local_context.catalog.load("my_partitioned_data")
 ``` 
you will have : 
![image](https://user-images.githubusercontent.com/57007314/181473668-8ae45549-1b83-4190-8d77-112fe3006aec.png)

4.  
``` python
my_data.items()
```
you will have :
![image](https://user-images.githubusercontent.com/57007314/181473989-a8248b41-e505-47a9-a253-bd9289a83bc0.png)

## Expected Result
I would like to have readable logs.

## Actual Result
logs are difficulty to read.

## Your Environment

* Kedro version used (`pip show kedro` or `kedro -V`): 0.18.2
* Python version used (`python -V`): 3.9.13
* Operating system and version: linux and VSCode with  Dark+ theme

thank you for your help :)

datajoely:
Can you run the following command `python -m rich.logging` on your terminal?

![image](https://user-images.githubusercontent.com/35801847/181481724-dccf9447-5132-42a0-b502-65fbd6be680d.png)

I'm using the Dark+ them with Zsh on my end and the logs look like they do above?

Debbby57:
thank you for your reply.
I have no problem with my logs in terminal. When I run your command on terminal I have the same logs as you. I only have those color problems when I use the interactive python window (Jupyter) of VSCode. Sorry if my issue wasn't explicit :/

datajoely:
@noklam maybe let's group this under the similar databricks issue?

noklam:
@datajoely Do you think this is a `rich` problem?

![image](https://user-images.githubusercontent.com/18221871/181542928-668a525c-cffd-4cd5-b738-f0fe737b632e.png)

[...]

Issue #1719, opened 2022-07-21
Jupyter nb output in PyCharm IDE breaks with Kedro 0.18.2


CraftingLevi:
## Description
Jupyter notebook cell output related to the logging package and pandas package in the IDE Pycharm >= 2022.1.2 renders as a web file instead of an interactive 'Table' after upgrading from kedro 0.18.1 to 0.18.2. My hypothesis is that this is caused by the native integration of the Rich package as of kedro 0.18.2. Current resolution is to remain at kedro 0.18.1 for time being, hoping this issues is fixed in a future release. 

Code run prior to the outputs in the image below:
`import numpy as np`
`import pandas as pd`
`from kedro.framework.cli import catalog`
`%load_ext kedro.extras.extensions.ipython`
`%reload_kedro D:\Work\vz-ona-kedro`
`df = catalog.load("node_dataset").head(5)`
`df.columns = ['Example'+str(idx) for idx, _ in enumerate(df.columns) ]`

*Outputs as of 0.18.2
![output0182](https://user-images.githubusercontent.com/19408835/180187302-219ed84d-0997-4b73-b5a2-311b41dc8122.png)
*Output types types as of 0.18.2
![output0182_outputtypes](https://user-images.githubusercontent.com/19408835/180187303-6c79029b-7af9-4553-8364-da4c8191a8ea.png)
*Outputs in 0.18.1
![output01801](https://user-images.githubusercontent.com/19408835/180187299-75e46ac9-4279-4a6c-a266-9fde38afc8f1.png)
*Output Types in 0.18.1
![output01801_outputtypes](https://user-images.githubusercontent.com/19408835/180187305-a5d96ca3-a330-4450-b9ae-3765d262e056.png)

## Context
Pycharm version >= 2022.1.2
Upgrade from Kedro 0.18.1 to 0.18.2 caused issue. Downgrade resolves issue. 
Jupyter cell outputs as a result of the logging package and pandas package render as 'Web' instead of Text / Table. 
After outputting, cells might become unresponsive to editing for +/- 15 seconds. 

Works fine in the browser version of Jupyter, however this does not have the functionality of an IDE, impacting productivity.

## Steps to Reproduce
1. Upgrade to Kedro 0.18.2 from 0.18.1
2. Run notebook cell that outputs a pandas dataframe, stacktrace or logs

## Expected Result
Output of cell is of type 'Table' and renders as an interactive table

## Actual Result
Output of cell is of type Web, and is non-interactive

[...]

PR #2592, opened 2023-05-19
Add the Kedro RichHandler

### LINKED ISSUE #2277: Introduce our own rich console logging handler
> **Note**. Non-breaking, goes in 0.18.x. Easiest to complete after #2206.

We want people to be able to e.g. turn rich tracebacks on/off from logging.yml. However, we don't want to start adding our own custom keys to logging.yml (the format is [specified already by Python](https://docs.python.org/3/library/logging.config.html#logging-config-dictschema)).

The way to do this is to write our own rich console logging handler. It will be _very_ simple, just a thin wrapper for `rich.logging.RichHandler`. We need to move all the custom behaviour out of `_ProjectLogging` and into the init method of the new logging handler. Let's call it `kedro.logging.RichHandler` (tbc exactly where it should live - we don't have `extras/logging` in `develop` so don't just put it there). 

Since we're piggybacking off `RichHandler`'s `rich_tracebacks` argument, which defaults to `False`, we'll need to change the logging config everywhere (default_logging.yml + all starters) to set this to `True`:

```
handlers:
  rich:
    class: kedro.logging.RichHandler
    rich_tracebacks: True
```

Add docs to describe how you can set `rich_tracebacks: False` if you'd like to disable them.

The new handler should look something like this:
```
class RichHandler(rich.logging.RichHandler):
    """Identical to rich's logging handler but with a few extra behaviours:
        * warnings issued by the `warnings` module are redirected to logging
        * pretty printing is enabled on the Python REPL (including IPython and Jupyter)
        * all tracebacks are handled by rich when rich_tracebacks=True
    """
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        logging.captureWarnings(True)
        rich.pretty.install()
        if self.rich_tracebacks:
            # We suppress click here to hide tracebacks related to it conversely,
            # kedro is not suppressed to show its tracebacks for easier debugging.
            # sys.executable is used to get the kedro executable path to hide the
            # top level traceback.
            # Rich traceback handling does not work on databricks. Hopefully this will be
            # fixed on their side at some point, but until then we disable it.
            # See https://github.com/Textualize/rich/issues/2455

            if "DATABRICKS_RUNTIME_VERSION" not in os.environ:
                rich.traceback.install(
                    suppress=[click, str(Path(sys.executable).parent)])
            ###                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
            ### NOTE YOU MAY WELL NEED TO CHANGE THIS PATH DEPENDING ON WHERE
            ### THIS FILE ENDS UP SO THAT IT POINTS TO THE RIGHT PLACE
```

Following #2206, `_ProjectLogging` would be left looking something like this:
```
class _ProjectLogging(UserDict):
    # pylint: disable=super-init-not-called
    def __init__(self):
        """Initialise project logging. The path to logging configuration is given in
        environment variable KEDRO_LOGGING_CONFIG (defaults to default_logging.yml)."""
        path = os.environ.get("KEDRO_LOGGING_CONFIG", Path(__file__).parent / "default_logging.yml")
        logging_config = Path(path).read_text(encoding="utf-8")
        self.configure(yaml.safe_load(logging_config))
```

noklam:

## Description

Fix #2277 

- #2277

It should also open up the solutions for these logging issues
- https://github.com/kedro-org/kedro/issues/1733
- https://github.com/kedro-org/kedro/issues/1719

This will enable users to customise `rich` logging and optionally disable the rich's traceback if needed (Some platform doesn't play very well with it, e.g. Databricks and we have some logic to disable it by default)

Bonus: It now also supports more arguments that is compatible with RichHandler and `rich.traceback.install` 

## Questions for Reviewer
- Where should we put the logger? We have `kedro.extras.logging` but we plan to remove this namespace, it is already removed in `dev`. Do we need to put it in `kedro.extras.logging` and move it to `kedro.logging` in 0.19.0 or just do it now since it's not breaking?
- If some error happen before the log was set up it won't be rich-formatted properly, i.e. invalid RichHandler. I think it's fine but worth to mention.
- I did some extra work to map arguments between RichHandler to `rich.traceback.install` to maximize the flexibility (Tests are not written yet, but if we agree this is useful then I will go on to write them). For example `tracebacks_show_locals` is mapped as `show_locals` for `rich.pretty.install`

Note: I haven't created unit tests since there are many decisions about what should be enable/disable, I will add the test after the 1st round of review.

## Development notes

- [x] Add some unit tests (test rich is enable by default, test the option to disable rich, and test all known available arguments of RichHandler)
- [x] Check if we can simplify the test, we manually delete `sys.modules` for testing `_ProjectLogging`, it may be easier now because this is configurable we don't have to reload module.
- [x] Add changes to all logging.yml (template)
- [ ] Add changes to all starters logging.yml (starters)
- [x] Add documentation about the changes 
- [x] Implementation of RichLogger

## Checklist

# Manual Testing
1. Run `kedro new -s pandas-iris`
2. Edit `conf/base/logging.yml` 

```yaml
  rich:
    class: kedro.extras.logging.RichHandler
    # class: rich.logging.RichHandler
    rich_tracebacks: True
   # tracebacks_show_locals: True
```

3. kedro run

# Note:
Currently user need to set `KEDRO_LOGGING_CONFIG` to disable rich's traceback, because the default logging installed it and there is no way to uninstall it.

We need to solve this by either implementing it ourselves or raise a PR to `rich`


features/steps/test_starter/{{ cookiecutter.repo_name }}/conf/base/logging.yml
```diff
@@ -34,7 +34,9 @@ handlers:
     delay: True
 
   rich:
-    class: rich.logging.RichHandler
+    class: kedro.extras.logging.RichHandler
+    rich_tracebacks: True
+    # tracebacks_show_locals: False
```
noklam, line 39: Like `settings.py`, we could have some commented default here to help user discover possible options. I put `show_locals` because initially when we introduce `rich`, we want to enable it as `True`. It turns out to be a bit noisy for some users, and there was no good way to turn it off by the user.

kedro/extras/logging/rich_logger.py
```diff
@@ -0,0 +1,53 @@
...
+            prefix = "tracebacks_"
+            if key.startswith(prefix):
+                key_prefix_removed = key[len(prefix) :]
+                if key_prefix_removed == "suppress":
+                    mapped_kwargs[key_prefix_removed] += value
+                else:
+                    mapped_kwargs[key_prefix_removed] = value
+            elif key in ("locals_max_length", "locals_max_string"):
+                mapped_kwargs[key] = value
+
```
noklam, line 48: The original issue propose to support `self.rich_tracebacks` only, I take it further because it is possible to support most of the arguments as long as they both exists in `RichHandler` and `rich.traceback.install`, the signature are slightly different so we need to map it.

kedro/extras/logging/rich_logger.py
```diff
@@ -0,0 +1,53 @@
...
+        if self.rich_tracebacks:
+            if "DATABRICKS_RUNTIME_VERSION" not in os.environ:
+                # https://rich.readthedocs.io/en/stable/reference/logging.html?highlight=rich%20handler#rich.logging.RichHandler
+                # Support compatible arguments between RichHandler and rich.traceback.install
+                rich.traceback.install(**mapped_kwargs)
```
noklam, line 53: Alternatively, we can also remove the special logic about Databricks but just change the default to False in the starter because it is now customisable.

I didn't do it because it's still a breaking change, and there's no point to use Rich's traceback on Databricks.

kedro/framework/project/__init__.py
```diff
@@ -221,18 +218,6 @@ def __init__(self):
...
-        # We suppress click here to hide tracebacks related to it conversely,
-        # kedro is not suppressed to show its tracebacks for easier debugging.
-        # sys.executable is used to get the kedro executable path to hide the
-        # top level traceback.
-        # Rich traceback handling does not work on databricks. Hopefully this will be
-        # fixed on their side at some point, but until then we disable it.
-        # See https://github.com/Textualize/rich/issues/2455
-        if "DATABRICKS_RUNTIME_VERSION" not in os.environ:
-            rich.traceback.install(suppress=[click, str(Path(sys.executable).parent)])
-        rich.pretty.install()
```
noklam, line 235: The rich's traceback won't be installed when `kedro.framework.project` is imported, it's now just a logging handler. This is an improvement because importing a module no longer has side-effect.

docs/source/logging/logging.md
```diff
@@ -43,6 +43,23 @@ Alternatively, if you would like to keep other configuration in `conf/base/loggi
 +  handlers: [console]
 ```
 
+### Customize the `rich` Handler
+
+Kedro's `kedro.extras.logging.RichHandler` is a subclass of [`rich.logging.RichHandler`](https://rich.readthedocs.io/en/stable/reference/logging.html?highlight=rich%20handler#rich.logging.RichHandler) and supports the same set of arguments. By default, `rich_tracebacks` is set to `True` to use `rich` to render exceptions. However, you can disable it by setting `rich_tracebacks: False`.
```
stichbury, line 48:
```suggestion
The link to `rich.logging.RichHandler` points to https://rich.readthedocs.io/en/stable/reference/logging.html?highlight=rich%20handler#rich.logging.RichHandler -- do you need the highlight? 


Do you need the

Comment (stichbury) [reply docs/source/logging/logging.md:48]: And likewise for the other links

**Review by merelcht — COMMENTED** [review]
To answer your questions:

> Where should we put the logger? We have kedro.extras.logging but we plan to remove this namespace, it is already removed in dev. Do we need to put it in kedro.extras.logging and move it to kedro.logging in 0.19.0 or just do it now since it's not breaking?

I would just put it in `kedro.logging` straightaway, so we don't need to move it again for the `0.19.0` release. 

> If some error happen before the log was set up it won't be rich-formatted properly, i.e. invalid RichHandler. I think it's fine but worth to mention.

That makes sense to me and I don't think there's a way around that.

> I did some extra work to map arguments between RichHandler to rich.traceback.install to maximize the flexibility (Tests are not written yet, but if we agree this is useful then I will go on to write them). For example tracebacks_show_locals is mapped as show_locals for rich.pretty.install

I don't know that much about Rich so I'm not entirely sure yet what this means, but it sounds good 😄 But it would be good to get @antonymilne 's view on this.

**Comment by antonymilne** [comment]
Let me start by answering the questions and will then review the code.

> * Where should we put the logger? We have `kedro.extras.logging` but we plan to remove this namespace, it is already removed in `dev`. Do we need to put it in `kedro.extras.logging` and move it to `kedro.logging` in 0.19.0 or just do it now since it's not breaking?

Agree with @merelcht that we should just put it in `kedro.logging` to begin with. Note there's no need to create a subpackage for it - just a single file `logging.py` should be good.

> * If some error happen before the log was set up it won't be rich-formatted properly, i.e. invalid RichHandler. I think it's fine but worth to mention.

This is unavoidable really and should be fine. Users will be able to fix it using the `KEDRO_LOGGING_CONFIG` environment variable to point to a different configuration anyway.

> * I did some extra work to map arguments between RichHandler to `rich.traceback.install` to maximize the flexibility (Tests are not written yet, but if we agree this is useful then I will go on to write them). For example `tracebacks_show_locals` is mapped as `show_locals` for `rich.pretty.install`

I think this is a good idea 👍 

> * [ ]  Add some unit tests (test rich is enable by default, test the option to disable rich, and test all known available arguments of RichHandler)

Probably you don't need to test all the known available arguments - just one of them should be fine.

File: kedro/extras/logging/rich_logger.py [hunk 6]
```diff
@@ -0,0 +1,53 @@
...
+import rich.traceback
+
+logger = logging.getLogger(__file__)
+
+
+class RichHandler(rich.logging.RichHandler):
+    """Identical to rich's logging handler but with a few extra behaviours:
+    * warnings issued by the `warnings` module are redirected to logging
+    * pretty printing is enabled on the Python REPL (including IPython and Jupyter)
+    * all tracebacks are handled by rich when rich_tracebacks=True
```
antonymilne, line 19: We should have anoter point here to describe the extra `rich.install` behaviour you added.

kedro/extras/logging/rich_logger.py
antonymilne: I would rather keep it as you have it, so the default is `True` and we include the special logic that disables it on Databricks.

kedro/extras/logging/rich_logger.py
```diff
@@ -0,0 +1,53 @@
...
+                key_prefix_removed = key[len(prefix) :]
+                if key_prefix_removed == "suppress":
+                    mapped_kwargs[key_prefix_removed] += value
+                else:
+                    mapped_kwargs[key_prefix_removed] = value
+            elif key in ("locals_max_length", "locals_max_string"):
+                mapped_kwargs[key] = value
+
+        if self.rich_tracebacks:
+            if "DATABRICKS_RUNTIME_VERSION" not in os.environ:
```
antonymilne, line 50:
```suggestion
        if self.rich_tracebacks and "DATABRICKS_RUNTIME_VERSION" not in os.environ:
```

kedro/extras/logging/rich_logger.py
```diff
@@ -0,0 +1,53 @@
...
+        logging.captureWarnings(True)
+        rich.pretty.install()
+
+        # We suppress click here to hide tracebacks related to it conversely,
+        # kedro is not suppressed to show its tracebacks for easier debugging.
+        # sys.executable is used to get the kedro executable path to hide the
+        # top level traceback.
+        # Rich traceback handling does not work on databricks. Hopefully this will be
+        # fixed on their side at some point, but until then we disable it.
+        # See https://github.com/Textualize/rich/issues/2455
```
antonymilne, line 33: This comment should be moved down to where it belongs.

antonymilne (commented):
Looks good in general! I will try it out but basically I think you're on the right track here and should go ahead and write tests. The only big thing is to put the code in `kedro/logging.py` instead of in extras.

docs/source/logging/logging.md
```diff
@@ -43,6 +43,23 @@ Alternatively, if you would like to keep other configuration in `conf/base/loggi
 +  handlers: [console]
 ```
 
+### Customize the `rich` Handler
+
+Kedro's `kedro.extras.logging.RichHandler` is a subclass of [`rich.logging.RichHandler`](https://rich.readthedocs.io/en/stable/reference/logging.html#rich.logging.RichHandler) and supports the same set of arguments. By default, `rich_tracebacks` is set to `True` to use `rich` to render exceptions. However, you can disable it by setting `rich_tracebacks: False`.
+
+When `rich_tracebacks` is set to `True`, the configuration is propagated to [`rich.traceback.install`](https://rich.readthedocs.io/en/stable/reference/traceback.html#rich.traceback.install). If an argument is compatible with `rich.pretty.install`, it will be passed to the traceback's settings.
```
antonymilne, line 50:
```suggestion
When `rich_tracebacks` is set to `True`, the configuration is propagated to [`rich.traceback.install`](https://rich.readthedocs.io/en/stable/reference/traceback.html#rich.traceback.install). If an argument is compatible with `rich.traceback.install`, it will be passed to the traceback's settings.
```
Unless I misunderstood something?

docs/source/logging/logging.md
```diff
@@ -43,6 +43,25 @@ Alternatively, if you would like to keep other configuration in `conf/base/loggi
 +  handlers: [console]
 ```
 
+### Customize the `rich` Handler
```
merelcht, line 46: Nit, sticking with UK spelling 🇬🇧 

```suggestion
### Customise the `rich` Handler
```

kedro/logging.py
```diff
@@ -0,0 +1,63 @@
...
+        # We suppress click here to hide tracebacks related to it conversely,
+        # kedro is not suppressed to show its tracebacks for easier debugging.
+        # sys.executable is used to get the kedro executable path to hide the
+        # top level traceback.
+
+        traceback_install_kwargs = {
+            "suppress": [click, str(Path(sys.executable).parent)]
+        }
+
+        # Mapping Arguments from RichHandler's Constructor to rich.traceback.install
```
merelcht, line 46: Nit 

```suggestion
        # Mapping arguments from RichHandler's constructor to rich.traceback.install
```

merelcht (commented):
I haven't tried it out yet, but the code looks good. Are you adding docs in a separate PR? Also don't forget to add this change to the release notes.

antonymilne (approved):
Nice work! Just a few small comments, but happy to approve.

Did you decide to remove the logic for the arguments `"locals_max_length", "locals_max_string"`?

merelcht (approved):
I agree with Antony's comments but other than that this looks good to be merged! Great work 👍 ⭐

noklam:
@antonymilne I found out the local_max_length is new additions and my local rich version actually don't support the arguments.

I am thinking either we drop it for now, or we only parse tracebacks_xxx arguments and pass whatever it is to rich.traceback.install and let rich raise the error in case of invalid arguments.

docs/source/kedro.logging.RichHandler.rst
```diff
@@ -0,0 +1,9 @@
+kedro.logging.RichHandler
+=========================
+
+.. currentmodule:: kedro.logging
+
+.. autoclass:: RichHandler
+   :members:
+   :undoc-members:
+   :inherited-members:
```
astrojuanlu, line 9: Maybe we'll have to remove the `:undoc-members:` and `:inherited-members:`

noklam: Thank you! This seems to work. Sphinx is a puzzle, I cannot reproduce the same error in my local environment, but I am glad it's fixed now.

astrojuanlu: I can't believe it worked 😂 I guess I'm a Sphinx black belt now 🥷

Issue #2588, opened 2023-05-17
Install Kedro, plugin, and test requirements in a single `pip install` command


deepyaman:
I think this is a critical problem that needs resolving.

What this means is that, in reality, we don't have resolvable requirements, and we're only able to get to a resolvable state by overwriting some of the previously-installed requirements. Some of the stuff installed in the `pip install -r test_requirements.txt` are not actually going to be compatible with Kedro on `main`, it seems.

_Originally posted by @deepyaman in https://github.com/kedro-org/kedro-plugins/pull/155#discussion_r1164384301_

astrojuanlu:
I agree this needs fixing.

This can be achieved by moving the contents of `test_requirements.txt` to the `optional-dependencies` subtable of `pyproject.toml`, which requires moving the contents of `dependency/requirements.txt` to the same file. This will require slight adjustments on our CI, but it's for the greater good.

deepyaman:
> I agree this needs fixing.
> 
> This can be achieved by moving the contents of `test_requirements.txt` to the `optional-dependencies` subtable of `pyproject.toml`, which requires moving the contents of `dependency/requirements.txt` to the same file. This will require slight adjustments on our CI, but it's for the greater good.

Yeah, the CI adjustments and shuffling of requirements shouldn't be bad at all (and generally beneficial, as you say); the bigger problem is that our requirements aren't actually compatible if installed together, and this needs to be resolved. 😿

astrojuanlu:
In progress in gh-2597

astrojuanlu:
It will be easier to tackle this after gh-2569 is merged.

PR #2597, opened 2023-05-22
Consolidate all requirements

astrojuanlu:
## Description
Fix gh-2588.

## Development notes
- Moved `dependency/requirements.txt` to `pyproject.toml`, removed relevant code from `setup.py`
- Moved `test_requirements.txt` to `setup.py` (where all the optional dependencies live at the moment)
- Removed all references to `test_requirements.txt` and `dependency/requirements.txt` in CI
- Adjusted various dependencies

Notice that I had to partially revert https://github.com/quantumblacklabs/private-kedro/pull/365 (private link). I did a quick test and the change seemed innocuous, but we should make sure that there are no problems when making a release (see original rationale https://github.com/quantumblacklabs/private-kedro/pull/352#issuecomment-564948371 also private)

Tested on Gitpod (Linux) on 3.7, 3.8, 3.9, and 3.10 ✔️ `pip check` always successful after `pip install .[all,test]` (no conflicting requirements).

## Checklist


astrojuanlu:
The e2e failures look legitimate and I'll look at them after tomorrow.

astrojuanlu:
Note to self: the Dependabot config needs fixing (noted by @merelcht)

https://github.com/kedro-org/kedro/blob/41f03d903cda4d4b4f61b01091f494b7b0817e79/.github/dependabot.yml#L7-L9

sbrugman:
>  error in project_dummy setup command: 'install_requires' must be a string or list of strings containing valid project/version requirement specifiers; Expected package name at the start of dependency specifier

Might be solved by upgrading [setuptools](https://stackoverflow.com/a/59164147) (otherwise the requirements syntax is probably invalid).

Great to see progress on moving to `pyproject.toml`. Would be amazing if Kedro and the starter templates set the right standard (unfortunately not there yet)!

astrojuanlu:
Thanks for the encouragement @sbrugman! Indeed that's the idea, we want to move everything to modern packaging standards. This is the way 🚀

features/steps/cli_steps.py
```diff
@@ -407,18 +407,17 @@ def update_pyproject_toml(context: behave.runner.Context, new_source_dir):
...
+    """Remove kedro as a standalone requirement.
     """
     reqs_path = context.root_project_dir / "src" / "requirements.txt"
-    kedro_reqs = f"-r {context.requirements_path.as_posix()}"
 
     if reqs_path.is_file():
         old_reqs = reqs_path.read_text().splitlines()
         new_reqs = []
         for req in old_reqs:
             if req.startswith("kedro"):
```
astrojuanlu, line 418: I'm not even sure this was correct to start with, now that we have `kedro-datasets`, `kedro-airflow` etc. I should probably parse the requirement and only remove `kedro` proper.

sbrugman:
@astrojuanlu Have you seen this somewhat related PR? Might be good to take them together
https://github.com/kedro-org/kedro-plugins/pull/263

astrojuanlu:
I saw @sbrugman ! 🙌🏽 will have a look soon

astrojuanlu:
I don't understand our linting 😅 `pylint` and `isort` have different opinions on import order. If I do what pylint says, isort fails.

I'm marking this as ready for review, hopefully someone else from the team can help me get this over the line.

merelcht:
> I don't understand our linting 😅 `pylint` and `isort` have different opinions on import order. If I do what pylint says, isort fails.
> 
> I'm marking this as ready for review, hopefully someone else from the team can help me get this over the line.

When I run `make lint` on Gitpod it runs completely fine... 🤔

astrojuanlu:
I removed import sorting from pylint since we were using isort anyway. All tests passing 👍🏽

features/steps/cli_steps.py
```diff
@@ -407,18 +408,16 @@ def update_pyproject_toml(context: behave.runner.Context, new_source_dir):
...
     reqs_path = context.root_project_dir / "src" / "requirements.txt"
-    kedro_reqs = f"-r {context.requirements_path.as_posix()}"
 
     if reqs_path.is_file():
         old_reqs = reqs_path.read_text().splitlines()
         new_reqs = []
         for req in old_reqs:
-            if req.startswith("kedro"):
-                new_reqs.append(kedro_reqs)
+            if req.startswith("kedro") and Requirement(req).name.lower() == "kedro":
```
astrojuanlu, line 418: The logic before was to replace `kedro` with all `kedro` dependencies. In a way, something like `pip install kedro --dependencies-only` (which doesn't exist I believe).

Now I'm just removing kedro, because this is only for our tests and the `behave` environment has the development version of kedro preinstalled already.

astrojuanlu: Also I made one improvement: parsing the requirement name so we are only dropping `kedro` and not, say, `kedro-datasets` or `kedro-airflow`. This was incorrect before.

setup.py
```diff
@@ -139,10 +132,77 @@ def _collect_requirements(requires):
 }
 
 extras_require["all"] = _collect_requirements(extras_require)
+extras_require["test"] = [
```
astrojuanlu, line 135: These are more or less the same as `kedro-datasets[all]`, but since we're removing datasets from framework soon, I just copy pasted `test_requirements.txt` here and didn't aim to substantially change what we're doing. Most of these can go away in 0.19.0.

merelcht (approved):
Thank you for making this change @astrojuanlu ⭐ I'm glad the `requirements` are cleaned up like this. I would add this change to the release notes as well.

merelcht:
> Note to self: the Dependabot config needs fixing (noted by @merelcht)
> 
> https://github.com/kedro-org/kedro/blob/41f03d903cda4d4b4f61b01091f494b7b0817e79/.github/dependabot.yml#L7-L9

Just a reminder for this ☝️

SajidAlamQB (approved):
Thank you @astrojuanlu!

Issue #2567, opened 2023-05-08
Stop relying on `.egg-info` directories for `kedro micropkg`


astrojuanlu:
## Description
While working on gh-2273, I realized that `kedro micropkg` relies on finding `.egg-info` directories to locate packages:

https://github.com/kedro-org/kedro/blob/653145771fa440a46b36d2c96c57a5997613da5e/kedro/framework/cli/micropkg.py#L146-L154

However, `.egg-info` directories are a legacy setuptools artifact and they're in the process of going away, see https://github.com/pypa/setuptools/milestone/3

The code was implemented in dbe6d895c0, which is part of https://github.com/kedro-org/kedro/pull/1263 and was recently touched by https://github.com/kedro-org/kedro/pull/1848. I couldn't find any discussion as to why `dist-info` was dropped in favor of `egg-info`.

## Possible Implementation

Since `_unpack_sdist` is called right before, IIUC the location of the project metadata file (at the moment `setup.py`, in the future `pyproject.toml`, see gh-2414) should be known. But I am surely missing something here.

astrojuanlu:
Just in time 👀 https://github.com/pypa/setuptools/blob/49fec9fafb0e23e0dde52d3c4c410d23a2de9b0d/CHANGES.rst?plain=1#L8-L11

> * #3948: Remove autofixing of broken ``.egg-info`` directories containing the ``-``
>   character in their base name (without suffix).
>   They should no longer be produced by sufficiently new versions of ``setuptools``
>   (warning introduced in 2005).

Issue #2542, opened 2023-04-26
Document clearly that`kedro micropkg pull` relies on sdists


astrojuanlu:
## Description

[The documentation for `kedro micropkg pull`](https://docs.kedro.org/en/stable/nodes_and_pipelines/micro_packaging.html?highlight=kedro%20micropkg#pull-a-micro-package) lists three possible use cases:

- `kedro micropkg pull dist/<pipeline_name>-0.1-py3-none-any.tar.gz` (local directory)
- `kedro micropkg pull s3://my_bucket/<pipeline_name>-0.1-py3-none-any.tar.gz` (cloud storage)
- `kedro micropkg pull <pypi_package_name>` (PyPI-like endpoint)

The last one relies on `pip download --no-deps` to download the corresponding source distributions:

https://github.com/kedro-org/kedro/blob/ae2022840147d0c46a985fb3061beac452b2ab59/kedro/framework/cli/micropkg.py#L325-L327

However, if the micropackage in question was published as a wheel, then `pip download` grabs that wheel and this code fails:

https://github.com/kedro-org/kedro/blob/ae2022840147d0c46a985fb3061beac452b2ab59/kedro/framework/cli/micropkg.py#L328-L336

## Steps to Reproduce

```
kedro new --starter=spaceflights
cd spaceflights
pip install -r src/requirements.txt

kedro micropkg package pipelines.data_processing

pushd dist && pip wheel *.tar.gz && popd

twine upload --repository-url http://0.0.0.0:8080/ dist/*

kedro micropkg pull data_processing
```

## Expected Result
The wheel is properly used.

## Actual Result

```
> PIP_TRUSTED_HOST=0.0.0.0 PIP_INDEX_URL=http://0.0.0.0:8080 kedro micropkg pull data_processing
/Users/juan_cano/.micromamba/envs/_test38/bin/python3.8 -m pip download --no-deps --dest /private/var/folders/r7/ywj0_kvj0mxfkdkx0jrgh73r0000gn/T/tmpq275beek data_processing
Looking in indexes: http://0.0.0.0:8080
Collecting data_processing
  Downloading http://0.0.0.0:8080/packages/data_processing-0.1-py3-none-any.whl (2.7 kB)
Saved /private/var/folders/r7/ywj0_kvj0mxfkdkx0jrgh73r0000gn/T/tmpq275beek/data_processing-0.1-py3-none-any.whl
Successfully downloaded data_processing
[...]

PR #2614, opened 2023-05-29
micropkg packaging improvements

astrojuanlu:
## Description
- Fix gh-2567.
- Fix gh-2542.
- Beginning of gh-2414 (still generates a `setup.py` on the fly, but the code is ready to use a `pyproject.toml` instead).
- Obliterate any remaining usage of `pkg_resources`, which is deprecated and will go away any time.

## Development notes
All the packaging stuff is quite complex - in fact, after an initially good streak, this is now my second backtrack attempt, after opening gh-2569 and gh-2597 (hopefully those two can go next). Keeping the micropackaging workflow intact while removing the strong assumptions around `setup.py` was extra hard.

I even decided to have a runtime dependency on `setuptools` because I couldn't find any "blessed", standalone PyPA package to do some tasks.

It's been mentioned that this workflow is weird and that we should evaluate whether we want to keep it or not https://github.com/kedro-org/kedro/milestone/21 but this work blocks the rest of the progress in https://github.com/kedro-org/kedro/milestone/36, so I didn't want to block it on that decision (which could be taken after 0.19).

## Checklist


astrojuanlu:
A small number of corner cases is failing, will have a look at those later this week.

astrojuanlu:
The last set of test failures is due to a behavior change from `pkg_resources` to `packaging`, noted it here https://github.com/pypa/packaging/issues/644#issuecomment-1567982812

I am hoping that we get an official answer soon, but while we wait, I think this is ready for review.

stichbury (approved):
Happy with this from docs perspective, but obvs not the main point of the PR.

kedro/framework/cli/micropkg.py
```diff
@@ -854,19 +912,36 @@ def _append_package_reqs(
     )
 
 
+def _get_all_library_reqs(metadata):
+    # Inspired by https://discuss.python.org/t/\
```
merelcht, line 916: (nit) I personally find it more useful to have a short explanation of what the function does instead of just the link.

kedro/framework/cli/micropkg.py
```diff
@@ -854,19 +912,36 @@ def _append_package_reqs(
...
     """
     parseable_requirements = set()
-    for requirement in pkg_resources.yield_lines(requirements):
-        try:
-            parseable_requirements.add(pkg_resources.Requirement.parse(requirement))
-        except ValueError:
-            continue
+    if isinstance(requirements, str):
+        requirements = requirements.splitlines()
+    # TODO: Properly handle continuation lines,
```
merelcht, line 935: Do you still need to do this or is this blocked by something?

merelcht:
I've tried packaging and pulling with this new version and the old (current) one and it all seems to work fine and to be backwards compatible 👌 It seems like the micropackage that's generated is still exactly the same. I guess that's supposed to be the case right?

astrojuanlu:
Yes, absolutely! That's why this PR would not fully close gh-2414 - the generated micropkg pretty much still uses `setup.py`.

merelcht (approved):
This looks like to be working as expected, so I'm happy to have it merged in. Maybe add to the release notes which changes were made here for traceability? And thanks for helping to improve this part of messy code base @astrojuanlu 🙇 ⭐

kedro/framework/cli/micropkg.py
astrojuanlu: Good point, I added a proper docstring

astrojuanlu:
I have addressed all the outstanding comments and now I believe all tests should pass. If everything goes well and unless there are other opinions, I'll merge this.

astrojuanlu:
All tests passing, CI failures due to non-100 % coverage. Will need to add tests to cover the extra possible (but unlikely) error messages.

astrojuanlu:
Added tests, finally! This is ready to go in. It has 2 approvals already but waiting until EOD to merge.

astrojuanlu:
Meh, playing whack-a-mole. I can reproduce these new test failures locally, will tackle them later.

merelcht:
> Added tests, finally! This is ready to go in. It has 2 approvals already but waiting until EOD to merge.

Ideally another engineer would approve this since the majority of the changes are in code.

kedro/framework/cli/micropkg.py
```diff
@@ -47,6 +50,50 @@
 """
 
 
+class _EquivalentRequirement(Requirement):
```
SajidAlamQB, line 53: Maybe we should be adding some docstrings here to describe the purpose of the class and its methods.

SajidAlamQB (approved):
Overall, the changes you've made to the micro-packaging workflow seem comprehensive and well thought out. Just some minor comments about readability. 

Thank you @astrojuanlu

kedro/framework/cli/micropkg.py
astrojuanlu: I'll add a docstring to make it more clear that this essentially overrides `packaging.requirements.Requirement.__eq__` to be backwards compatible with `pkg_resources`

PR #2632, opened 2023-06-02
[Draft] Dataset factories - Eager resolving approach

### LINKED ISSUE #2423: Eliminate the need of having all datasets defined in the catalog

ankatiyar:
## Description
Resolve #2423 This PR is introduces the dataset factories feature. This implementation demonstrates the **eager** resolving of catalog approach.

### TO DO:

- [ ] Documentation
- [ ] Tests
- [ ] Update `kedro catalog list` function

## Checklist


ankatiyar:
Closing this in favour of the lazy resolving approach. #2635

Issue #2603, opened 2023-05-22
Add new catalog CLI commands for dataset factories


merelcht:
## Description
Part of https://github.com/kedro-org/kedro/issues/2423

We should decide what new catalog CLI commands to add to support the dataset factories syntax:
- a command that ranks all patterns so a user can understand the priority matching order
- a resolve command that shows the fully resolved catalog when all patterns are matched (this excludes the actual patterns for ease of reading)

## Context
These CLI commands are mainly to make it clear for users how exactly the dataset factories work and match the catalog entries.

ankatiyar:
Adding updating `kedro catalog list` command to be done with this.

merelcht:
This has now been completed with:
- https://github.com/kedro-org/kedro/issues/2789
- https://github.com/kedro-org/kedro/issues/2790
- https://github.com/kedro-org/kedro/issues/2791
- https://github.com/kedro-org/kedro/issues/2849

PR #2635, opened 2023-06-02
Dataset factories

### LINKED ISSUE #2423: Eliminate the need of having all datasets defined in the catalog

ankatiyar:
## Description
Resolve #2423 This PR is introduces the dataset factories feature. Look at #2670 for the documentation on how to use it.

#### Kedro run workflow 
-->`kedro run` in terminal
--> `Session` is created  and run
--> `Context` created which loads all the config from the `conf_source` 
--> `DataCatalog.fromconfig(dict:catalog)` which initialises the `DataCatalog` object
* At this point the DataCatalog is created with explicitly mentioned datasets in `DataCatalog._data_sets` 
* Sort the patterns and save them to `DataCatalog._dataset_patterns`
* Also save `load_version` and `save_version` to `self._load_version` and `self._save_version`
    
--> `runner.run(pipeline, catalog, .., ..)`
* loop through all datasets used by the pipeline and save all that exist (either as explicit datasets or patterns) to `registered_ds`
* ~~The check for existence will materialise and add the datasets to the catalog (in `DataCatalog.__contains__()`)~~

--> Pipeline is run and `catalog._get_dataset(dataset_name)-> AbstractDataSet` : ~No major change because datasets should already be materialised at the check for existence~
* Add datasets that exist as patterns as materialised datasets to the catalog

*NOTE* : The tests are not updated to reflect the latest changes (TODO)

Toy project to test with : https://github.com/ankatiyar/space-patterns
#### Follow up tasks to do 
- [ ] Update tests
- [ ] Documentation #2666 
- [ ] #2668 
- [ ] Update `kedro catalog list` + https://github.com/kedro-org/kedro/issues/2603

## Checklist


kedro/io/data_catalog.py
```diff
@@ -170,7 +190,14 @@ def __init__(
...
         self.layers = layers
-
+        # Keep a record of all patterns in the catalog.
+        # {dataset pattern name : dataset pattern body}
+        self.dataset_patterns = dict(dataset_patterns or {})
+        self._sorted_dataset_patterns = sorted(
+            self.dataset_patterns.keys(),
+            key=lambda x: (_specificity(x), -x.count("{"), x),
+        )
+        self._pattern_name_matches_cache: dict[str, str] = {}
```
merelcht, line 200: I don't quite understand what this is for. If the dataset has been matched it's added to the catalog and will be in `self._data_sets` so why exactly do you need this?

kedro/io/data_catalog.py
```diff
@@ -567,13 +636,42 @@ def list(self, regex_search: str | None = None) -> list[str]:
...
+        ):
+            return True
+        matched_pattern = self.match_name_against_pattern(dataset_name)
+        if self.dataset_patterns and matched_pattern:
+            # cache the "dataset_name -> pattern" match
+            self._pattern_name_matches_cache[dataset_name] = matched_pattern
+            return True
+        return False
+
+    def match_name_against_pattern(self, dataset_name: str) -> str | None:
```
merelcht, line 653: What's the purpose of this method?

kedro/io/data_catalog.py
```diff
@@ -311,6 +358,28 @@ def _get_dataset(
 
         return data_set
 
+    def _get_resolved_dataset(
+        self, dataset_name: str, matched_pattern: str
+    ) -> AbstractDataSet:
+        result = parse(matched_pattern, dataset_name)
```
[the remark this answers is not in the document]
ankatiyar: Not necessarily needed to do this twice, I could also save the `result` returned by parse earlier in the cache along with the name  in the `match_name_against_pattern()` method but I didn't want to use tuples in the cache dict. Doing this again here isn't really a performance bottleneck since this only happens once.

kedro/io/data_catalog.py
```diff
@@ -567,13 +638,42 @@ def list(self, regex_search: str | None = None) -> list[str]:
...
 
+    def exists_in_catalog(self, dataset_name: str) -> bool:
+        """Check if a dataset exists in the catalog as an exact match or if it matches a pattern."""
+        if (
+            dataset_name in self._data_sets
+            or dataset_name in self._pattern_name_matches_cache
+        ):
+            return True
+        matched_pattern = self.match_name_against_pattern(dataset_name)
+        if self.dataset_patterns and matched_pattern:
```
merelcht, line 649: Is it necessary to check `if self.dataset_patterns`? If that it `False` then it shouldn't be possible to have a `matched_pattern` in the line above right?

kedro/io/data_catalog.py
```diff
@@ -311,6 +360,28 @@ def _get_dataset(
 
         return data_set
 
+    def _get_resolved_dataset(
```
merelcht, line 363: I'd rename this to `_resolve_dataset` because that's what actually happens in this method. 
```suggestion
    def _resolve_dataset(
```

kedro/io/data_catalog.py
```diff
@@ -311,6 +360,28 @@ def _get_dataset(
...
+                # result.named: gives access to all dict items in the match result.
+                # format_map fills in dict values into a string with {...} placeholders
+                # of the same key name.
+                try:
+                    template_copy[key] = string_value.format_map(result.named)
+                except KeyError as exc:
+                    raise DataSetError(
+                        f"Unable to resolve '{key}' for the pattern '{matched_pattern}'"
+                    ) from exc
+            # Create dataset from catalog template.
```
merelcht, line 382: Nit: bit this comment isn't aligned correctly with the line it belongs to.

kedro/io/data_catalog.py
merelcht: Since it goes over the list I would rename it : `match_name_against_patterns` I know it's minor, but does make it easier to follow the code imo.

kedro/io/data_catalog.py
```diff
@@ -567,13 +638,42 @@ def list(self, regex_search: str | None = None) -> list[str]:
...
+            return True
+        return False
+
+    def match_name_against_pattern(self, dataset_name: str) -> str | None:
+        """Match a dataset name against existing patterns"""
+        # Loop through all dataset patterns and check if the given dataset name has a match.
+        for pattern in self._sorted_dataset_patterns:
+            result = parse(pattern, dataset_name)
+            # If there's a match resolve the rest of the pattern template to create
+            # a dataset instance. A result can be None or contain a dictionary of matched items:
```
merelcht, line 661: I think this comment is left over from the previous implementation and should be removed.

tests/io/test_data_catalog.py
```diff
@@ -683,3 +745,109 @@ def test_no_versions_with_cloud_protocol(self):
...
+            ("ford_cars", True),
+            ("sail_boats", True),
+        ],
+    )
+    def test_exists_in_catalog_cache_updated(
+        self, config_with_dataset_factories, dataset_name, expected
+    ):
+        """Check that the pattern match cache is updated when exists_in_catalog() is called"""
+        catalog = DataCatalog.from_config(**config_with_dataset_factories)
+        assert catalog.exists_in_catalog(dataset_name) == expected
```
merelcht, line 795: I don't quite understand the difference between this test and the previous one.

merelcht (commented):
Thanks for clarifying the use of the cache @ankatiyar, it makes more sense now! I think we can keep it, but perhaps give it a slightly shorter name e.g. `_pattern_matches_cache`.
In general, I struggled with following the code because some of the names of the methods. I tried giving suggestions on how to improve, but it's quite personal so it would be good to get some other opinions as well 🙂 

One thing every single user in the interviews mentioned is a warning about the catch-all pattern, so we must really implement that, including tests. 

Another thing that was mentioned several times was a warning about when multiple patterns match a dataset, it would be good to have a test for that to see how that would happen and if we can also warn about that.

kedro/io/data_catalog.py
ankatiyar: updated

kedro/io/data_catalog.py
ankatiyar: Done

tests/io/test_data_catalog.py
ankatiyar: The one before this was meant to check if the fn returns true or false correctly for datasets that exist either as explicit entries or patterns. This one is meant to check if the cache is updated correctly so all the examples here are just pattern matches. Updated it!

kedro/io/data_catalog.py
```diff
@@ -567,13 +636,42 @@ def list(self, regex_search: str | None = None) -> list[str]:
             ) from exc
         return [dset_name for dset_name in self._data_sets if pattern.search(dset_name)]
 
+    def exists_in_catalog(self, dataset_name: str) -> bool:
```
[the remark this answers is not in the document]
ankatiyar: Updated!

ankatiyar:
@merelcht I'll leave the warning about catch-all patterns and multiple matches out of this PR but do it as a part of https://github.com/kedro-org/kedro/issues/2668

kedro/io/data_catalog.py
```diff
@@ -170,7 +187,23 @@ def __init__(
         self._data_sets = dict(data_sets or {})
         self.datasets = _FrozenDatasets(self._data_sets)
         self.layers = layers
-
+        # Keep a record of all patterns in the catalog.
+        # {dataset pattern name : dataset pattern body}
+        self.dataset_patterns = dict(dataset_patterns or {})
```
noklam, line 192: What is the type for `dataset_patterns`? Why would we want `dataset_patterns` as a dict but `self.dataset_patterns` as dict of dict?

kedro/io/data_catalog.py
```diff
@@ -170,7 +187,23 @@ def __init__(
...
+        # 2. Decreasing number of placeholders (no of curly brackets)
+        # 3. Alphabetical
+        self._sorted_dataset_patterns = sorted(
+            self.dataset_patterns.keys(),
+            key=lambda pattern: (
+                -(_specificity(pattern)),
+                -pattern.count("{"),
+                pattern,
+            ),
+        )
```
noklam, line 204: This should be refactor as a method and have its own unit tests

kedro/io/data_catalog.py
```diff
@@ -271,35 +305,54 @@ class to be loaded is specified with the key ``type`` and their
...
-                layers[ds_layer].add(ds_name)
-
-            ds_config = _resolve_credentials(ds_config, credentials)
-            data_sets[ds_name] = AbstractDataSet.from_config(
-                ds_name, ds_config, load_versions.get(ds_name), save_version
-            )
+            # Assume that any name with } in it is a dataset factory to be matched.
+            if "}" in ds_name:
+                # Add each factory to the dataset_patterns dict.
+                dataset_patterns[ds_name] = ds_config
```
noklam, line 311: I have trouble to map the types, it looks like from the definition `dataset_patterns` is `dict[str, Any], but here it's actually a dict or dict?

kedro/io/data_catalog.py
```diff
@@ -311,6 +364,28 @@ def _get_dataset(
 
         return data_set
 
+    def _resolve_dataset(
```
noklam, line 367: I think this definitely conflicts with `_suggest_resume_scenario`, can you try to test with a fail pipeline and see if it breaks it?

kedro/io/data_catalog.py
```diff
@@ -567,16 +642,47 @@ def list(self, regex_search: str | None = None) -> list[str]:
             ) from exc
         return [dset_name for dset_name in self._data_sets if pattern.search(dset_name)]
 
+    def exists_in_catalog_config(self, dataset_name: str) -> bool:
```
noklam, line 645: Is this the cache that add matched pattern into catalog? If so why do we need to keep it separately instead of just creating the entry in catalog?

kedro/io/data_catalog.py
```diff
@@ -567,16 +642,47 @@ def list(self, regex_search: str | None = None) -> list[str]:
...
             Copy of the current object.
         """
-        return DataCatalog(data_sets=self._data_sets, layers=self.layers)
+        return DataCatalog(
+            data_sets=self._data_sets,
+            layers=self.layers,
+            dataset_patterns=self.dataset_patterns,
+        )
 
     def __eq__(self, other):
```
noklam, line 680: Maybe a good exercise to use dataclass or attr to remove these boilerplate later

noklam (commented):
Drop some comments for 1st round of review.

kedro/io/data_catalog.py
ankatiyar: It is meant to be a dict of dict. pattern name -> unresolved config stored as a dict

kedro/io/data_catalog.py
ankatiyar: fixed it

kedro/io/data_catalog.py
ankatiyar: This fn just adds the dataset name -> pattern in the `self._pattern_matches_cache` does not create an entry in the catalog. This is because this function is also called in the runner to check for existence of datasets from the pipeline in the catalog.  The actual adding of a dataset to the catalog happens in `self._get_dataset()`
This is to avoid doing the pattern matching multiple times when it's called inside a Session run but the actual resolving and adding is still at `_get_dataset` time to allow for DataCatalog standalone usage

kedro/io/data_catalog.py
ankatiyar: I checked, it doesn't break `_suggest_resume_scenario`

kedro/io/data_catalog.py
```diff
@@ -110,6 +110,22 @@ def _specificity(pattern: str) -> int:
...
+    return sorted(
+        dataset_patterns,
+        key=lambda pattern: (
+            -(_specificity(pattern)),
+            -pattern.count("{"),
+            pattern,
+        ),
+    )
+
+
```
noklam, line 128: 👍🏼

kedro/io/data_catalog.py
```diff
@@ -328,31 +336,30 @@ class to be loaded is specified with the key ``type`` and their
...
+            # slow down plugins like `kedro-viz`
+            if suggest:
+                matches = difflib.get_close_matches(
+                    data_set_name, self._data_sets.keys()
+                )
+                if matches:
+                    suggestions = ", ".join(matches)
+                    error_msg += f" - did you mean one of these instead: {suggestions}"
+
+            raise DatasetNotFoundError(error_msg)
```
noklam: It's small change but I think this is easier to understand now and the nested logic is less complex.

tests/io/test_data_catalog.py
```diff
@@ -825,18 +826,44 @@ def test_dataset_not_in_catalog_when_no_pattern_match(
...
+                ],
+                [
+                    "{brand}_scooters",
+                    "{brand}_planes",
+                    "{brand}_cars",
+                ],
+            ),
+            (
+                ["{brand}_{vehicle}", "{dataset}a"],
+                ["{brand}_{vehicle}", "{dataset}a"],
```
noklam, line 860: This feels a bit redundant, is this testing something new?

tests/io/test_data_catalog.py
```diff
@@ -825,18 +826,44 @@ def test_dataset_not_in_catalog_when_no_pattern_match(
...
+            (
+                [
+                    "{brand}_cars",
+                    "{brand}_scooters",
+                    "{brand}_planes",
+                ],
+                [
+                    "{brand}_scooters",
+                    "{brand}_planes",
+                    "{brand}_cars",
```
noklam, line 855: Nit, I will reorder this to be the first test case since this is more straightforward, the first test case is the more complicated one.

kedro/io/data_catalog.py
idanov: Is there a reason this one is a public property? If it is needed as a public one, it should be a frozen one so no one changes it. But in my opinion, we don't need it at all or we only need the pattern names to be public.

Also could this be made sorted by default? Dictionaries in python maintain their insertion order from Python 3.6/3.7 and if not, `OrderedDict` can be used.

We shouldn't keep the same data twice, once as `self._sorted_dataset_patterns`.

kedro/io/data_catalog.py
```diff
@@ -94,6 +96,36 @@ def _sub_nonword_chars(data_set_name: str) -> str:
...
+    specificity("france.companies") = 16
+    Args:
+        pattern: The factory pattern
+    """
+    # Remove all the placeholders from the pattern
+    result = re.sub(r"\{.*?\}", "", pattern)
+    return len(result)
+
+
+def _sort_dataset_factory_patterns(dataset_patterns: list) -> list:
```
idanov, line 113: Having helper functions outside of the class makes inheritance a bit harder. Can you add this one and the previous one as a class method to the `DataCatalog`?

kedro/io/data_catalog.py
```diff
@@ -141,6 +173,7 @@ def __init__(
         data_sets: dict[str, AbstractDataSet] = None,
```
idanov, line 173: Maybe we need to align the spelling here in 0.19. It's quite evident here:

- `data_sets`
- `dataset_patterns`

kedro/io/data_catalog.py
```diff
@@ -170,7 +203,15 @@ def __init__(
...
-
+        # Keep a record of all patterns in the catalog.
+        # {dataset pattern name : dataset pattern body}
+        self.dataset_patterns = dict(dataset_patterns or {})
+        # Sort all the patterns according to the parsing rules -
+        self._sorted_dataset_patterns = _sort_dataset_factory_patterns(
+            list(self.dataset_patterns.keys())
+        )
+        # Cache that stores {name : matched_pattern}
+        self._pattern_matches_cache: dict[str, str] = {}
```
idanov, line 214: Do we really need a separate variable for this cache? Couldn't this just be `self._data_sets` and just have a `set()` of all original (non-pattern) data set names?

kedro/io/data_catalog.py
```diff
@@ -271,22 +313,40 @@ class to be loaded is specified with the key ``type`` and their
...
         for ds_name, ds_config in catalog.items():
-            ds_layer = ds_config.pop("layer", None)
-            if ds_layer is not None:
-                layers[ds_layer].add(ds_name)
-
-            ds_config = _resolve_credentials(ds_config, credentials)
-            data_sets[ds_name] = AbstractDataSet.from_config(
-                ds_name, ds_config, load_versions.get(ds_name), save_version
-            )
+            # Assume that any name with "}" in it is a dataset factory to be matched.
```
idanov, line 316: I would've rewritten this for readability as follows:

1. Create a factory method for creating a dataset out of `def create_dataset(cls, ds_name, ds_config):`
2. Convert this single for loop with an if in it into two Python comprehensions for the two different cases

kedro/io/data_catalog.py
```diff
@@ -271,22 +313,40 @@ class to be loaded is specified with the key ``type`` and their
...
+            layers=dataset_layers,
+            dataset_patterns=dataset_patterns,
+        )
 
     def _get_dataset(
         self, data_set_name: str, version: Version = None, suggest: bool = True
     ) -> AbstractDataSet:
-        if data_set_name not in self._data_sets:
+        if self.exists_in_catalog_config(data_set_name):
+            if data_set_name not in self._data_sets:
```
idanov, line 340: Could we get rid of the extra indentation level here?

```suggestion
        if self.exists_in_catalog_config(data_set_name) and data_set_name not in self._data_sets:
```

kedro/io/data_catalog.py
idanov: This isn't a good name for a public method, it's way too descriptive. I would suggest that you change it to `def __contains__(self, dataset_name):`. This will allow us to do a test like this:
`if dataset_name in catalog:...`, which is way more intuitive and nicer than `catalog.exists_in_catalog_config(dataset_name)`.

kedro/io/data_catalog.py
```diff
@@ -567,16 +649,47 @@ def list(self, regex_search: str | None = None) -> list[str]:
             ) from exc
         return [dset_name for dset_name in self._data_sets if pattern.search(dset_name)]
 
+    def exists_in_catalog_config(self, dataset_name: str) -> bool:
+        """Check if a dataset exists in the catalog as an exact match or if it matches a pattern."""
+        if (
+            dataset_name in self._data_sets
+            or dataset_name in self._pattern_matches_cache
```
idanov, line 656: If `self._data_set` contains all "materialised" datasets, this would be just one test here.

kedro/io/data_catalog.py
```diff
@@ -567,16 +649,47 @@ def list(self, regex_search: str | None = None) -> list[str]:
...
+        ):
+            return True
+        matched_pattern = self.match_name_against_patterns(dataset_name)
+        if matched_pattern:
+            # cache the "dataset_name -> pattern" match
+            self._pattern_matches_cache[dataset_name] = matched_pattern
+            return True
+        return False
```
idanov, line 664: If we got rid of the overly verbose method name `self.match_name_against_patterns` and simply replace it with `resolve`, which could do all that's needed without leakage of abstraction, all this could be just `return self.resolve(dataset_name) is not None`.

kedro/io/data_catalog.py
```diff
@@ -567,16 +649,47 @@ def list(self, regex_search: str | None = None) -> list[str]:
...
+
+    def match_name_against_patterns(self, dataset_name: str) -> str | None:
+        """Match a dataset name against existing patterns"""
+        # Loop through all dataset patterns and check if the given dataset name has a match.
+        for pattern in self._sorted_dataset_patterns:
+            result = parse(pattern, dataset_name)
+            if result:
+                return pattern
+        return None
```
idanov, line 673: Is there a reason we need this separately from directly resolving upon check of existence? Could we just call this `resolve` and do all of the matching and calling the resolve helper here?

kedro/runner/runner.py
```diff
@@ -74,14 +74,35 @@ def run(
...
+            if not catalog.exists_in_catalog_config(ds)
+        ]
+
+        # Check if there are any input datasets that aren't in the catalog and
+        # don't match a pattern in the catalog.
+        unsatisfied = [
+            input_name
+            for input_name in pipeline.inputs()
+            if input_name in unregistered_ds
+        ]
```
idanov, line 91: Using comprehension here makes it look more complicated than it is. Could we keep the original code here where we are using simple set difference?
```suggestion
        unsatisfied = pipeline.inputs() - set(registered_ds)
```

This reads in English as "give me all inputs minus all the registered ones".

kedro/runner/runner.py
```diff
@@ -74,14 +74,35 @@ def run(
...
         catalog = catalog.shallow_copy()
 
-        unsatisfied = pipeline.inputs() - set(catalog.list())
+        # Check which datasets used in the pipeline aren't in the catalog and don't match
+        # a pattern in the catalog
+        unregistered_ds = [
+            ds
+            for ds in pipeline.data_sets()
+            if not catalog.exists_in_catalog_config(ds)
+        ]
```
idanov, line 83: Or code for the positive instead of the negative so to create your equivalent of `catalog.list()` (named as either `catalog_list` or `registered_ds`) and not needing to change almost anything in the runner code:
```suggestion
        registered_ds = [ds for ds in pipeline.data_sets() if ds in catalog]
```

kedro/runner/runner.py
```diff
@@ -74,14 +74,35 @@ def run(
...
-        unregistered_ds = pipeline.data_sets() - set(catalog.list())
+        # Check if there's any output datasets that aren't in the catalog and don't match a pattern
+        # in the catalog.
+        free_outputs = [
+            output_name
+            for output_name in pipeline.outputs()
+            if output_name in unregistered_ds
+        ]
+
+        # Create a default dataset for unregistered datasets
```
idanov, line 105:
```suggestion
        free_outputs = pipeline.outputs() - set(registered_ds)
        unregistered_ds = pipeline.data_sets() - set(registered_ds)
```

idanov (commented):
It's closing up to become a pretty nice PR. I've added a few comments, especially for the API surface with the public method names.

kedro/io/data_catalog.py
```diff
@@ -257,35 +266,113 @@ class to be loaded is specified with the key ``type`` and their
...
+    ) -> str | None:
+        """Match a dataset name against patterns in a dictionary containing patterns"""
+        for pattern, _ in data_set_patterns.items():
+            result = parse(pattern, data_set_name)
+            if result:
+                return pattern
+        return None
+
+    @staticmethod
+    def _create_dataset(
```
ankatiyar, line 329: Potentially unnecessary?

kedro/io/data_catalog.py
```diff
@@ -136,11 +138,14 @@ class DataCatalog:
...
 
-    def __init__(
+    def __init__(  # pylint: disable=too-many-arguments
         self,
         data_sets: dict[str, AbstractDataSet] = None,
         feed_dict: dict[str, Any] = None,
         layers: dict[str, set[str]] = None,
+        dataset_patterns: dict[str, dict[str, Any]] = None,
+        load_versions: dict[str, str] = None,
+        save_version: str = None,
```
merelcht, line 148: These new arguments should also be added to the doc string below.

kedro/io/data_catalog.py
```diff
@@ -257,36 +266,105 @@ class to be loaded is specified with the key ``type`` and their
...
+            dataset_patterns=sorted_patterns,
+            load_versions=load_versions,
+            save_version=save_version,
+        )
 
-            ds_config = _resolve_credentials(ds_config, credentials)
-            data_sets[ds_name] = AbstractDataSet.from_config(
-                ds_name, ds_config, load_versions.get(ds_name), save_version
-            )
+    @staticmethod
```
merelcht, line 310: What's the reason behind some of these new methods being static methods and some class methods? None of these actually return a class instance.

kedro/io/data_catalog.py
```diff
@@ -257,36 +266,105 @@ class to be loaded is specified with the key ``type`` and their
...
+            if result:
+                return pattern
+        return None
+
+    @classmethod
+    def _sort_patterns(
+        cls, data_set_patterns: dict[str, dict[str, Any]]
+    ) -> dict[str, dict[str, Any]]:
+        """Sort a dictionary of dataset patterns according to parsing rules -
+        1. Decreasing specificity (no of characters outside the curly brackets
```
merelcht, line 333: Nit
```suggestion
        1. Decreasing specificity (number of characters outside the curly brackets)
```

kedro/io/data_catalog.py
```diff
@@ -257,36 +266,105 @@ class to be loaded is specified with the key ``type`` and their
...
+        2. Decreasing number of placeholders (no of curly bracket pairs)
```
merelcht, line 334: Nit
```suggestion
        2. Decreasing number of placeholders (number of curly bracket pairs)
```

kedro/io/data_catalog.py
```diff
@@ -257,36 +266,105 @@ class to be loaded is specified with the key ``type`` and their
...
+
+    @staticmethod
+    def _specificity(pattern: str) -> int:
+        """Helper function to check the length of exactly matched characters not inside brackets
+        Example -
+        specificity("{namespace}.companies") = 10
+        specificity("{namespace}.{dataset}") = 1
+        specificity("france.companies") = 16
+        Args:
+            pattern: The factory pattern
```
merelcht, line 358: For completeness this doc string should also include the return type/value.

kedro/io/data_catalog.py
```diff
@@ -257,36 +266,105 @@ class to be loaded is specified with the key ``type`` and their
...
+        """
+        # Remove all the placeholders from the pattern and count the number of remaining chars
+        result = re.sub(r"\{.*?\}", "", pattern)
+        return len(result)
 
     def _get_dataset(
         self, data_set_name: str, version: Version = None, suggest: bool = True
     ) -> AbstractDataSet:
-        if data_set_name not in self._data_sets:
```
merelcht, line 289: Why did this check change from `self._data_sets` to `self` ?

kedro/io/data_catalog.py
```diff
@@ -311,6 +387,53 @@ def _get_dataset(
 
         return data_set
 
+    def __contains__(self, data_set_name):
```
merelcht, line 390: IMO `__contains__` should really only check if a dataset is in the catalog or not, but not add it. I would move that responsibility to the `_get_dataset` method instead. Mainly because `__contains__` is a magic function and is meant for implementing the membership operation, whereas now it does more than just that.

merelcht (commented):
I reviewed both this PR and https://github.com/kedro-org/kedro/pull/2743 and I definitely prefer the approach here with adding `load_versions` and `save_version` arguments instead of `raw_catalog`. I left some more comments, but in general I think this is nearly finished.

kedro/io/data_catalog.py
ankatiyar: Done

kedro/io/data_catalog.py
ankatiyar: `_is_pattern`, `_match_pattern` & `_sort_patterns` (also `_specificity` which is used by `_sort_patterns` need to be either class methods or static methods because they're used in `from_config` which is a class method. Class/static methods do not necessarily need to return class instances, they just can be used without having an instance of the class.

kedro/io/data_catalog.py
ankatiyar: Moved it to `_get_dataset`

kedro/io/data_catalog.py
ankatiyar: It was because the materialising of datasets was happening in `__contains__` which was materialising the patterned datasets when the runner was checking for existence of these datasets. The check in `self` would also do the same if the `DataCatalog` was being used as a stand alone. But now that `_get_dataset` has this responsibility I have changed it back to the original code.

kedro/io/data_catalog.py
ankatiyar: It's a private method, the `Args:` part is leftover from when it was a helper fn outside `DataCatalog` I've removed it

kedro/io/data_catalog.py
merelcht: Ah sorry I don't think my question was very clear. I understand why these aren't regular instance methods, but my question is why `_is_pattern` and `_match_pattern` are static methods and why `_sort_patterns` and `_specificity` are class methods? Why aren't they all the same type?

kedro/io/data_catalog.py
```diff
@@ -257,35 +275,119 @@ class to be loaded is specified with the key ``type`` and their
...
     ) -> AbstractDataSet:
+        matched_pattern = self._match_pattern(self._dataset_patterns, data_set_name)
+        if data_set_name not in self._data_sets and matched_pattern:
+            # If the dataset is a patterned dataset, materialise it and add it to
+            # the catalog
+            data_set_config = self._resolve_config(data_set_name, matched_pattern)
+            ds_layer = data_set_config.pop("layer", None)
+            if ds_layer:
+                if not self.layers:
+                    self.layers = {}
```
merelcht, line 382: This can be simplified to: 
```
self.layers = self.layers or {}
```

kedro/io/data_catalog.py
ankatiyar: `_specificity`, `_match_pattern` and `_is_pattern` now are all static methods but `_sort_patterns()` still needs to be a class method because it uses `_specificity` inside it which is only accessible if we pass `cls` as an argument.

merelcht: Okay cool, thanks for clarifying 🙂

merelcht (approved):
I left one more minor comment, but apart from that I'm very happy with how this looks! It would be good to get another review from @noklam or @idanov, but from my side this is good to go 👍

kedro/io/data_catalog.py
```diff
@@ -136,11 +138,14 @@ class DataCatalog:
     to the underlying data sets.
     """
 
-    def __init__(
+    def __init__(  # pylint: disable=too-many-arguments
         self,
         data_sets: dict[str, AbstractDataSet] = None,
         feed_dict: dict[str, Any] = None,
         layers: dict[str, set[str]] = None,
+        dataset_patterns: dict[str, dict[str, Any]] = None,
```
noklam, line 146: Maybe it is good to introduce keyword-only arguments here, similar to the proposal to get us more freedom to re-arrange the arguments later without breaking changes. The current argument lists does not make too much sense

kedro/io/data_catalog.py
```diff
@@ -170,8 +184,12 @@ def __init__(
         self._data_sets = dict(data_sets or {})
         self.datasets = _FrozenDatasets(self._data_sets)
         self.layers = layers
+        # Keep a record of all patterns in the catalog.
+        # {dataset pattern name : dataset pattern body}
+        self._dataset_patterns = dict(dataset_patterns or {})
```
noklam, line 189: This still confuse me a lot, is there any difference? Isn't dataset_patterns a dict already?
```suggestion
        self._dataset_patterns =dataset_patterns or {}
```

kedro/io/data_catalog.py
```diff
@@ -170,8 +184,12 @@ def __init__(
...
+        self._load_versions = dict(load_versions or {})
```
noklam, line 190: Same as `self._dataset_patterns`

kedro/io/data_catalog.py
```diff
@@ -257,35 +275,118 @@ class to be loaded is specified with the key ``type`` and their
             >>> catalog.save("boats", df)
         """
         data_sets = {}
+        dataset_patterns = {}
         catalog = copy.deepcopy(catalog) or {}
         credentials = copy.deepcopy(credentials) or {}
         save_version = save_version or generate_timestamp()
         load_versions = copy.deepcopy(load_versions) or {}
+        layers: dict[str, set[str]] = defaultdict(set)
```
noklam, line 283: Same argument, good to have keywords-only argument here even we are going to remove layers soon

noklam (commented):
I have some comments here, but I figured it is easier if I drop all my comments in one draft PR, it's not ready yet. I will ping you when it's ready.

 [Review PR ](https://github.com/kedro-org/kedro/pull/2770)since it is easier.

kedro/io/data_catalog.py
```diff
@@ -257,35 +275,118 @@ class to be loaded is specified with the key ``type`` and their
...
-            ds_config = _resolve_credentials(ds_config, credentials)
-            data_sets[ds_name] = AbstractDataSet.from_config(
-                ds_name, ds_config, load_versions.get(ds_name), save_version
-            )
+    @staticmethod
+    def _is_pattern(pattern: str):
+        """Check if a given string is a pattern. Assume that any name with '{' is a pattern."""
+        if "{" in pattern:
+            return True
+        return False
```
noklam, line 324:
```suggestion
       return True if "{" in pattern else False
```

noklam:
As discussed, let's separate the Type Alias proposed in #2770.
 I will close #2770 and approve this PR now. :)

noklam (approved):
LGTM now! Let's ship it and do the type alias separately.

kedro/io/data_catalog.py
```diff
@@ -257,35 +275,124 @@ class to be loaded is specified with the key ``type`` and their
...
+    @staticmethod
+    def _match_pattern(
+        data_set_patterns: dict[str, dict[str, Any]], data_set_name: str
+    ) -> str | None:
+        """Match a dataset name against patterns in a dictionary containing patterns"""
+        for pattern, _ in data_set_patterns.items():
+            result = parse(pattern, data_set_name)
+            if result:
+                return pattern
+        return None
```
idanov, line 333: This could be nicely rewritten to:
```
matches = (parse(pattern, data_set_name) for pattern in data_set_patterns.keys())
return next(filter(None, matches), None)
```

kedro/io/data_catalog.py
```diff
@@ -257,35 +275,124 @@ class to be loaded is specified with the key ``type`` and their
...
+                    layers[ds_layer].add(ds_name)
+                data_sets[ds_name] = AbstractDataSet.from_config(
+                    ds_name, ds_config, load_versions.get(ds_name), save_version
+                )
+        dataset_layers = layers or None
+        sorted_patterns = cls._sort_patterns(dataset_patterns)
+        missing_keys = [
+            key
+            for key in load_versions.keys()
+            if not (cls._match_pattern(sorted_patterns, key) or key in catalog)
```
idanov, line 303: This test should be reordered, since `key in catalog` is easier to test and probably will match more often. No need to go through `_match_pattern` for all the non-patterned datasets.

kedro/io/data_catalog.py
```diff
@@ -257,35 +275,124 @@ class to be loaded is specified with the key ``type`` and their
...
+            key=lambda pattern: (
+                -(cls._specificity(pattern)),
+                -pattern.count("{"),
+                pattern,
+            ),
+        )
+        sorted_patterns = {}
+        for key in sorted_keys:
+            sorted_patterns[key] = data_set_patterns[key]
+        return sorted_patterns
```
idanov, line 355: Shorter and neater.
```
return {key: data_set_patterns[key] for key in sorted_keys}
```

PR #2673, opened 2023-06-12
Deprecate dataset errors with module `__getattr__`

deepyaman:
## Description

Deprecating using `DeprecatedClassMeta` only deprecates the name when a deprecated object is instantiated. In the case of errors, they are often used without instantiation (e.g. in `except` blocks), so `DeprecatedClassMeta` doesn't work.

## Development notes

Use `__getattr__` approach for deprecation. This is a relatively-newer method (added in 3.7, see https://peps.python.org/pep-0562/).

## Checklist


tests/io/test_core.py
```diff
@@ -27,6 +32,13 @@
 ]
 
 
+@pytest.mark.parametrize("name", _DEPRECATED_ERROR_CLASSES)
+@pytest.mark.parametrize("package", ["kedro.io", "kedro.io.core"])
+def test_deprecation(name, package):
+    with pytest.warns(DeprecationWarning, match=f"{name} has been renamed"):
+        exec(f"from {package} import {name}")  # pylint: disable=exec-used
```
astrojuanlu, line 39: Does this work if we use https://docs.python.org/3/library/importlib.html#importlib.import_module instead?

kedro/io/core.py
```diff
@@ -71,10 +70,17 @@ class DatasetAlreadyExistsError(DatasetError):
...
+def __getattr__(name):
+    if name in _DEPRECATED_ERROR_CLASSES:
+        alias = name.replace("DataSet", "Dataset")
+        warnings.warn(
+            f"{name} has been renamed to {alias}, and the alias will be removed in "
+            f"Kedro 0.19.0",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return globals()[alias]
```
astrojuanlu, line 82: What about having a mapping between old names new objects so that we don't have to use the `globals()` mechanism? Something like

```python
_DEPRECATED_ERROR_CLASSES = {
    "DataSetError": DatasetError,
    "DataSetNotFoundError": DatasetNotFoundError,
    "DataSetAlreadyExistsError": DatasetAlreadyExistsError,
}
```

the `if name in _DEPRECATED_ERROR_CLASSES` would still work, and then we could do `return _DEPRECATED_ERROR_CLASSES[name]`.

kedro/io/__init__.py
```diff
@@ -24,6 +22,22 @@
...
+DataSetAlreadyExistsError: type[DatasetError]
+
+
+def __getattr__(name):
+    import kedro.io.core  # pylint: disable=import-outside-toplevel
+
+    if name in (
+        kedro.io.core._DEPRECATED_ERROR_CLASSES  # pylint: disable=protected-access
+    ):
+        return getattr(kedro.io.core, name)
```
astrojuanlu, line 37: While I oppose using `globals()` for the exceptions, this is fine by me 👍🏽

astrojuanlu (commented):
Thanks a lot @deepyaman! Left a few suggestions for improvement.

noklam (commented):
some minor comments on styling, I have done manual testing and it's working, but I have some questions.


image

Questions:
We will get different a error when we do `from xxx import` versus `importlib.import_module`, is this fine? I think it should be fine, as this behavior is consistent when I test a 3-rd party library like `pandas`, we didn't change anything, but just want to check to make sure.


ref: the code I used for manual test
```python
# Expected

from kedro.io import DataSetNotFoundError

# Expected
import importlib
io = importlib.import_module("kedro.io")
io.DataSetNotFoundError

# ImportError
from kedro.io import DUMMY

# AttributeError
io.DUMMY
```

tests/io/test_core.py
deepyaman: No, it doesn't. :( `exec` was not even on my list of things to try (I didn't even know about it?), but nothing else seemed to work.

kedro/io/core.py
deepyaman: The code with `globals()` pretty much comes from the PEP that proposed adding module `__getattr__`: https://peps.python.org/pep-0562/

But I don't know why it shouldn't work to use a dictionary here, so I can try if `globals()` is bad to use. My understanding was that this may be a good use of `globals()`.

I can also see how other libraries using `__getattr__` for depreciation do it, potentially.

tests/io/test_core.py
astrojuanlu: Weird, we did some manual tests and seemed to work https://github.com/kedro-org/kedro/pull/2673#issuecomment-1598916317 in any case, if it doesn't, it's fine

kedro/io/__init__.py
```diff
@@ -24,6 +22,22 @@
...
+    raise AttributeError(f"module {repr(__name__)} has no attribute {repr(name)}")
```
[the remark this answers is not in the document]
deepyaman: I created https://github.com/kedro-org/kedro/pull/2706 to fix it for most cases. I realized that I did indeed also write the same code again i this PR, so I've also updated it here (and updated the test).

kedro/io/core.py
deepyaman: `globals()` removed in https://github.com/kedro-org/kedro/pull/2673/commits/9b13345b52b9c23baea74b7f24aa96493ebfdd57 (following your suggestion).

About the `exec()`, I'm not sure what to do/haven't looked into it more. Open to suggestions; also OK to ignore it for now since it's only in test code that can be removed post-0.19.0-release.

deepyaman: Actually, I will try overwriting `__dir__` and see if it happens to help.

tests/io/test_core.py
deepyaman: Oh, wow, I'm dumb. I tried `importlib.import_module("DataSetError", "kedro.io.core")` and stuff like that, but obviously you can't import a class using `import_module`. Fixing...

astrojuanlu (approved):
💯

kedro/io/__init__.py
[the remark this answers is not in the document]
astrojuanlu: gh-2706 is merged!

astrojuanlu:
I don't understand why DCO was happy until now, but then when I updated the branch it started complaining about all the previous commits, and even after I removed my merge commit, it keeps complaining. @deepyaman do you want to do the honors and signoff everything?

tynandebold (approved):

noklam (approved):

astrojuanlu:
Submitted a linearized version of this to gh-2711.

PR #2684, opened 2023-06-14
Show project logs at info level with default logging config

### LINKED ISSUE #2520: Make sure project logs of INFO level are shown
> **Note**. Technically this is non-breaking, but best to do this following https://github.com/kedro-org/kedro/issues/2426, which is a breaking change. Hence this is for 0.19.0.

We are removing logging.yml from the project template (see https://github.com/kedro-org/kedro/issues/2281). However, there's a subtle piece of functionality that exists in project-side logging.yml that's not easily reproduced in the framework-side one: the project-side logging.yml has something like this:
```
loggers:
  kedro:
    level: INFO

  iris:
    level: INFO
```

It's not possible to put this in the framework-side logging.yml because you don't know what the project name is there. We can achieve this by adding it in the Python code in `framework/project/__init__.py` though by using `PACKAGE_NAME` (defined in `configure_project`), something like this:

```
logging_config["loggers"].append(PACKAGE_NAME: {"level": "INFO"})
```

_Originally posted by @AntonyMilneQB in https://github.com/kedro-org/kedro/issues/2281#issuecomment-1419040400_

AhdraMeraliQB:
## Description

Resolves #2520 

When the default logging setup is being used we need to make sure that any logs written in the project itself are logged as expected. 

As this will be run regardless of whether the default logging setup is being used, we ensure that it does not overwrite any specification that the user has supplied themselves.

## Development notes
This was tested locally with a clean version of spaceflights.
In the implementation we reconfigure the loggers as they are first configured before `configure_project()` is run, i.e. before the package name is available. Once this is defined we configure them again, with the project logger included.

## Checklist

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -1,0 +1,13 @@
+FROM gitpod/workspace-full:2023-05-08-21-16-55
+
+# Some datasets work on 3.8 only
+RUN pyenv install 3.8.15\
+    && pyenv global 3.8.15
+
+# VideoDataSet
+RUN sudo apt-get update && sudo apt-get install -y --no-install-recommends libgl1
+RUN sudo apt-get install make
+RUN npm install -g @mermaid-js/mermaid-cli
+# https://stackoverflow.com/questions/69564238/puppeteer-error-failed-to-launch-the-browser-process
+# https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md#chrome-doesnt-launch-on-linux
+RUN sudo apt-get install -y --no-install-recommends libatk-bridge2.0-0 libcups2 ca-certificates fonts-liberation libasound2 libatk-bridge2.0-0 libatk1.0-0 libc6 libcairo2 libcups2 libdbus-1-3 libexpat1 libfontconfig1 libgbm1 libgcc1 libglib2.0-0 libgtk-3-0 libnspr4 libnss3 libpango-1.0-0 libpangocairo-1.0-0 libstdc++6 libx11-6 libx11-xcb1 libxcb1 libxcomposite1 libxcursor1 libxdamage1 libxext6 libxfixes3 libxi6 libxrandr2 libxrender1 libxss1 libxtst6 lsb-release wget xdg-utils
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,32 +1,18 @@
-image: gitpod/workspace-full:legacy-dazzle-v1
-
+image:
+  file: .gitpod.Dockerfile
 tasks:
-  # We want packages installed during the pre-build init steps to go to /workspace
-  # rather than ~ so that they are persisted. Gitpod sets PIP_USER=yes to ensure this,
-  # but pre-commit requires PIP_USER=no. Hence we set PIP_USER=no and use
-  # pip install --user to install to /workspace.
   - name: kedro
-    before: |
-      echo PIP_USER=no >> ~/.bashrc && export PIP_USER=no
+
     init: |
       make sign-off
-      pip install -r test_requirements.txt --user
-      gp sync-done kedro
-    command: |
-      pre-commit install --install-hooks
-      clear
-  # Generate a test Kedro project at /workspace/project using the local Kedro version.
-  - name: test-project
-    init: |
-      gp sync-await kedro
-      export PIP_USER=no
+      pip install -e /workspace/kedro
       cd /workspace
-      pip install -e /workspace/kedro --no-deps
-      source ~/.bashrc
       yes project | kedro new -s pandas-iris --checkout main
+      pip install -r /workspace/kedro/test_requirements.txt
+      cd /workspace/kedro
+      pre-commit install --install-hooks
+
     command: |
-      pip install -e /workspace/kedro --no-deps
-      cd /workspace/project
       clear
       kedro info
 
--- a/features/jupyter.feature
+++ b/features/jupyter.feature
@@ -22,6 +22,6 @@ Feature: Jupyter targets in new project
     Given I have added a test jupyter notebook
     When I execute the test jupyter notebook and save changes
     And I execute the kedro jupyter command "convert --all"
-    And Wait until the process is finished
+    And Wait until the process is finished for up to "120" seconds
     Then I should get a successful exit code
     And Code cell with node tag should be converted into kedro node
--- a/features/steps/cli_steps.py
+++ b/features/steps/cli_steps.py
@@ -352,10 +352,10 @@ def wait_for_notebook_to_run(context, timeout):
         raise TimeoutError("Failed to run Jupyter server in time")
 
 
-@when("Wait until the process is finished")
-def wait(context):
+@when('Wait until the process is finished for up to "{timeout:d}" seconds')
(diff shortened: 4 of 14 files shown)


kedro/framework/project/__init__.py
```diff
@@ -252,6 +252,11 @@ def configure_project(package_name: str):
     global PACKAGE_NAME
     PACKAGE_NAME = package_name
 
+    # Don't overwite any defined project log level in KEDRO_LOGGING_CONFIG
+    if PACKAGE_NAME and PACKAGE_NAME not in LOGGING["loggers"].keys():
+        LOGGING["loggers"].update({PACKAGE_NAME: {"level": "INFO"}})
+        LOGGING.configure(LOGGING.data)
+
```
noklam, line 259: @antonymilne 
I guess this is the weirdest part in the code. I trace back to ParallelRunner, it passed the `LOGGING` as `logging_config` directly. So essentially it's doing something like `LOGGING.configure(LOGGING`. This isn't possible as it cause recursion error.

AhdraMeraliQB: I suspect the recursion error boils down to addressing because we are using the same object, but that's just a hunch. Using `LOGGING.data` successfully bypasses this.

tests/framework/project/test_logging.py
```diff
@@ -23,9 +23,22 @@ def default_logging_config():
...
+        },
+        "loggers": {"kedro": {"level": "INFO"}, "test_project": {"level": "INFO"}},
+        "root": {"handlers": ["rich"]},
+    }
+    return logging_config
+
+
 @pytest.fixture(autouse=True)
 def reset_logging(default_logging_config):
-    yield
```
AhdraMeraliQB, line 28: Unsure why yield statement was introduced as it prevents `LOGGING` from being reset between logging tests

kedro/framework/project/__init__.py
antonymilne: Can we do this as something like this instead?
```
# In ProjectLogging
    def set_project_logging(self, package_name):
        if package_name not in self.data["loggers"]:
            self.data["loggers"][package_name] = {"level": "INFO"}

# in configure_project
    LOGGING.set_project_logging(package_name)
```

antonymilne: P.S. I know the `configure_logging` function is super weird but I don't think there's any other way to do it unfortunately. The parallel runner thing is a huge pain. I recommend not touching `configure_logging` at all if possible 😅

tests/framework/project/test_logging.py
AhdraMeraliQB: Ah, I see -  I've changed it because `LOGGING.data` perpetuates from tests that are run on different mock projects before this particular suite of tests, and for each mock project a corresponding logger is now added to the default logger - which would cause `test_default_logging_config` to fail. These project loggers cannot be explicitly defined in `default_logging_config` as it is dependent on the order of the tests run. 

Are there any implications to the tests by having the logs reset before them as opposed to after? As I understand it the ordering of operations is the same for the most part.

merelcht: Ideally the order in which the tests are run shouldn't matter and they should pass. If that's not the case then data cleanup isn't done properly after each test. If `LOGGING.data` is causing issues, I'd assume we have to reset that after each test.

AhdraMeraliQB: I agree @merelcht , the only issue is that `LOGGING.data` isn't being reset after other tests modify it in test suites that are run before this file gets run - running this test file alone behaves as expected as there are no pre-existing modifications to `LOGGING.data` and it is reset after every test. The problem arises when running this test suite after the tests in `test_startup.py` (and some other files) - `LOGGING.data` is not reset after those tests, and so the first test in `test_logging.py` is run on a modified version of `LOGGING.data`. To resolve this I have changed the test cleanup from being performed after each test to before each test. 

The running of the file is generally unaffected; instead of having test 1, clean-up, test 2, clean-up, test 3, clean-up... we have clean-up, test 1, clean-up, test 2, clean-up, test 3... .

Alternative 1: We could have the clean-up be run after the tests automatically and manually enact it before the first test - the result would be virtually the same.

Alternative 2: Introducing data cleanup to the other test files that modify `LOGGING.data`.

I don't see any detriment to changing when the clean-up is introduced, be it before or after tests are called, other than feeling a bit counter-intuitive. However if we are against cleaning the data beforehand then I would vouch for alternative 2 as it allows for consistent implementation across the test suites and prevents this problem from happening in other tests in future. 

Would love to hear opinions from reviewers on which approach to go with.

ankatiyar (approved):
Manually tested, it works! (I don't have any strong opinions about cleanup before or after the test tbh)

tests/framework/project/test_logging.py
noklam: https://docs.pytest.org/en/6.2.x/fixture.html#teardown-cleanup-aka-fixture-finalization is a good read. I think it makes more sense to have a teardown process which clean up the test afterward.

Just to add more context, this is what happening now
`pytest tests/framework/project/test_logging.py  tests/framework/test_startup.py` - pass
 `pytest  tests/framework/test_startup.py  tests/framework/project/test_logging.py ` - fail

As test should be isolated, the problem here is then `test_startup.py` isn't clean up properly and we should update this file instead

antonymilne: @AhdraMeraliQB I 100% agree with @noklam here. The state after running a test should be identical to what is was before running the test. Hence responsibility for clear up belongs in the teardown process rather than in the setup process. So alternative 2 is definitely the right one to go for here. If it means editing tests that already exist that is fine - it just means that those tests were not correctly doing teardown which they should have been doing before.

AhdraMeraliQB:
The test directories that currently modify the logging config are: 
- `tests/framework/cli/...`
- `tests/framework/context/...`
- `tests/framework/project/...`
- `tests/framework/test_startup.py`

As a result, the teardown fixtures were moved to `test/framework/conftest.py`

tests/framework/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+    return logging_config
+
+
+@pytest.fixture(autouse=True)
+def reset_logging(request, default_logging_config):
+    yield
+    if "nologreset" in request.keywords:
+        return
+
+    configure_logging(default_logging_config)
```
noklam, line 26: Nice

noklam (commented):
Thank you for investigating this! I think this is now much cleaner, and possibly solves the problem once and for all! I believes there are few other tests need to be run in the order, hopefully we don't need that anymore.

> The test directories that currently modify the logging config are:
> 
> * `tests/framework/cli/...`
> * `tests/framework/context/...`
> * `tests/framework/project/...`
> * `tests/framework/test_startup.py`
> 
> As a result, the teardown fixtures were moved to `test/framework/conftest.py`
I am slightly concern that `kedro/framework` is too big for the scope, is it possible to import an `autouse` fixture to a file instead?
I see `test_session_hook_manager` is the only exception here, is it the only file that doesn't require resetting log? If this is the case I am fine with the current approach.

noklam (approved):
Approved :) thanks for taking care of this tricky PR

PR #2705, opened 2023-06-20
Remove layer validation from Kedro 

rashidakanchwala:

## Description

This PR is based on this issue - [Validate layers for transcoding datasets](https://github.com/kedro-org/kedro-viz/issues/1382)

This part of the ticket is where we remove `_validate_layers_for_transcoding` from kedro as we have a similar function in kedro-viz. The kedro-viz part of the ticket is [here](https://github.com/kedro-org/kedro-viz/pull/1406)

## Development notes

The `_validate_layers_for_transcoding` functions calls `_transcoding_split` function which checks if the transcoded dataset name is correct. This `_transcoding_split` function is also called in modular_pipelines.py. 

Earlier if we define transcoded dataset in the catalog.yml in the below way, a value error would raised.

```yaml
ingestion.int_typed_shuttles@pandas1@something:
  type: pandas.ParquetDataSet
  filepath: ${base_location}/02_intermediate/typed_shuttles.pq
  layer: intermediate

ingestion.int_typed_shuttles@pandas2:
  type: pandas.ParquetDataSet
  filepath: ${base_location}/02_intermediate/typed_shuttles.pq
  layer: intermediate
```

```bash
ValueError: Expected maximum 1 transcoding separator, found 2 instead: 
'ingestion.int_typed_shuttles@pandas1@something'.
```

However it is no longer raised once we have removed `_validate_layers_for_transcoding`  function. I think this is a bit confusing and any help appreciated

## Checklist


noklam (commented):
I think this makes sense, looking at the function definition.

```python
def _transcode_split(element: str) -> tuple[str, str]:
    split_name = element.split(TRANSCODING_SEPARATOR)
    if len(split_name) > 2:
        raise ValueError(
            f"Expected maximum 1 transcoding separator, found {len(split_name) - 1} "
            f"instead: '{element}'."
```

The error was raised inside `_transcode_split`, it is only used in the modular_pipeline `_strip_transcoding` which you also imported in the other PR.

I think you need to check what's happened when there is multiple `@`, which one should be the layer instead? I think the answer to this is unknown, this should simply be banned during the validation step.

rashidakanchwala:
> I think this makes sense, looking at the function definition.
> 
> ```python
> def _transcode_split(element: str) -> tuple[str, str]:
>     split_name = element.split(TRANSCODING_SEPARATOR)
>     if len(split_name) > 2:
>         raise ValueError(
>             f"Expected maximum 1 transcoding separator, found {len(split_name) - 1} "
>             f"instead: '{element}'."
> ```
> 
> The error was raised inside `_transcode_split`, it is only used in the modular_pipeline `_strip_transcoding` which you also imported in the other PR.
> 
> I think you need to check what's happened when there is multiple `@`, which one should be the layer instead? I think the answer to this is unknown, this should simply be banned during the validation step.

ok so here's the thing @something refers to a dataset not a layer (layer is always defined as a key-value pair) so I think it shouldn't really belong in the layer validation step. it should belong to someplace else where we check if the transcoded_name is correct and multiple '@'s is wrong syntax so definitely an error should be raised. Do you know where we can add this?

AhdraMeraliQB:
From what I can see it looks like within `_validate_layers_for_transcoding` there were two validation steps being performed, the first being a validation that the transcoding notation was declared accurately, and the second being that the transcoded datasets were defined with the same layer. Whilst the second validation step is being moved to Viz as it's to do with the layer attribute, I believe that the first should stay within Kedro framework.

To do this `_validate_layers_for_transcoding` should probably be replaced with something like `_validate_transcoded_datasets` which could look like the following:

```
def _validate_transcoded_datasets(catalog: DataCatalog): -> None
    for dataset_name in catalog.datasets.keys():
        _ , _ = _transcode_split(dataset_name)
    return
```

This would preserve using `_transcode_split` to check the validity of the transcoded dataset name without touching the layer attribute.

AhdraMeraliQB:
Also - I believe this should go into develop as until 0.19.0 we will still be supporting defining layer at the top level

rashidakanchwala:
> From what I can see it looks like within `_validate_layers_for_transcoding` there were two validation steps being performed, the first being a validation that the transcoding notation was declared accurately, and the second being that the transcoded datasets were defined with the same layer. Whilst the second validation step is being moved to Viz as it's to do with the layer attribute, I believe that the first should stay within Kedro framework.
> 
> To do this `_validate_layers_for_transcoding` should probably be replaced with something like `_validate_transcoded_datasets` which could look like the following:
> 
> ```
> def _validate_transcoded_datasets(catalog: DataCatalog): -> None
>     for dataset_name in catalog.datasets.keys():
>         _ , _ = _transcode_split(dataset_name)
>     return
> ```
> 
> This would preserve using `_transcode_split` to check the validity of the transcoded dataset name without touching the layer attribute.

Yup, i thought the same just wasn't sure whether it should be in context.py or someplace else. Cool i will make the change above. thanks Ahdra :)

rashidakanchwala:
> Also - I believe this should go into develop as until 0.19.0 we will still be supporting defining layer at the top level

So i think it will not affect anything. We are just moving this function from kedro to kedro-viz and on kedro-viz we are supporting both old and new way of defining layers. The major difference is if you make a mistake in the way you define layers now -- kedro run will still finish without throwing any errors but kedro -viz will fail.

AhdraMeraliQB:
Makes sense; will this affect the release ordering? 🤔

rashidakanchwala:
> Makes sense; will this affect the release ordering? 🤔

well, I think we need to release Kedro-viz first. because if we release it after then there's no handling done at all. 
and well if there's a gap then it would fail both kedro run and kedro-viz for the time being (which is ok) 
@AhdraMeraliQB -- would u mind having a look at the viz PR. i can merge it to main if that part is good ? thanks

rashidakanchwala:
> > > Makes sense; will this affect the release ordering? 🤔
> > 
> > 
> > well, I think we need to release Kedro-viz first. because if we release it after then there's no handling done at all. and well if there's a gap then it would fail both kedro run and kedro-viz for the time being (which is ok) @AhdraMeraliQB -- would u mind having a look at the viz PR. i can merge it to main if that part is good ? thanks
> 
> Hm... I think this is considered a breaking change? This is a case that `kedro` should raise an issue not until `kedro-viz` IMO.
> 
> `something@dummy1@dummy2` is defintely not a valid transcoding scheme and it's a valid error in `kedro`. Does the example that you show in description pass silently now?

Nope, it's not a breaking change. Kedro-viz is handling both the old and new way of defining layers and also validating_transcoding_datasets for both ways of defining layers. It's just now Kedro will not throw any error if there's an issue with transcoding datasets having different layers. It will still throw an error if you define in your catalog dataset@something1@something2.

And that error will be thrown during kedro viz

noklam:
I will run some quick manual tests later but otherwise I am fine with the changes.

noklam:
@rashidakanchwala Can you fix the test? I saw the unit tests are still failing

kedro/framework/context/context.py
```diff
@@ -118,28 +118,9 @@ def _convert_paths_to_absolute_posix(
...
-                else:
-                    base_names_to_layer[base_name] = current_layer
-
-    conflicting_datasets = sorted(_find_conflicts())
-    if conflicting_datasets:
-        error_str = ", ".join(conflicting_datasets)
-        raise ValueError(
-            f"Transcoded datasets should have the same layer. Mismatch found for: {error_str}"
-        )
+def _validate_transcoded_datasets(catalog: DataCatalog):
```
AhdraMeraliQB, line 121: Just a note that the linters will fail here as there is no docstring

AhdraMeraliQB (commented):
Looks good to me! Thank you for making these changes - happy to approve once tests are passing

noklam (approved):
Thanks for fixing this. It's cleaner now the logic move to `kedro-viz` and we don't have to handle metadata in a special way.

AhdraMeraliQB (approved):
👍

astrojuanlu:
Doc builds are failing on missing GitHub anchors. It's happening in several unrelated pull requests:

- https://readthedocs.org/projects/kedro/builds/21229765/
- https://readthedocs.org/projects/kedro/builds/21229865/

Here's the error:

```
sphinx.errors.SphinxWarning: /home/docs/checkouts/readthedocs.org/user_builds/kedro/checkouts/2705/docs/source/nodes_and_pipelines/nodes.md:292:broken link: https://github.com/kedro-org/kedro-starters/blob/dacdd56f1c1afde00a03a1e342fc0f44e1567a1e/pandas-iris/%7B%7B%20cookiecutter.repo_name%20%7D%7D/src/%7B%7B%20cookiecutter.python_package%20%7D%7D/nodes.py#L13 (Anchor 'L13' not found)
```

I suspect something has changed in the GitHub HTML structure. I recommend that, for now, we remove the faulty link:

https://github.com/kedro-org/kedro/blob/9500d60081dde839d060c316d3557ebe0ad41ddf/docs/source/nodes_and_pipelines/nodes.md?plain=1#L292

cc @stichbury (whenever you see this)

PR #2715, opened 2023-06-22
Check for the whole protocol including delimited (<scheme>://) when j…

juliushetzel:
Check for the whole protocol including delimited (<scheme>://) when joining path for partitions in partitioned data set

## Description
The PartitionedDataSet does not join the protocol properly to delegate it to its partitions. It does not work if the name of the path starts with the same letters as the protocol as it is only a startsWith comparison.

If a path starts with the prefix as same as the protocol (e.g. s3-kedro-pipeline-filesystem for a bucket) the protocol is not delegated to the partitions of the PartitionedDataSet

## Development notes
Changed from only scheme comparison to scheme plus delimited comparison.
Changed the test to use s3 as prefix for the bucket name.

## Checklist

### DIFF
--- a/tests/io/test_partitioned_dataset.py
+++ b/tests/io/test_partitioned_dataset.py
@@ -401,7 +401,7 @@ class TestPartitionedDatasetLocal:
         assert pds._credentials == global_creds
 
 
-BUCKET_NAME = "fake_bucket_name"
+BUCKET_NAME = "s3_fake_bucket_name"
 S3_DATASET_DEFINITION = [
     "pandas.CSVDataSet",
     "kedro.extras.datasets.pandas.CSVDataSet",
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -3,6 +3,7 @@
 ## Major features and improvements
 
 ## Bug fixes and other changes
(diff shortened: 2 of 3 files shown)


datajoely:
Great catch!

astrojuanlu (approved):
Thanks a lot @juliushetzel! Would you mind amending the commit with a signoff (`git commit -s` or `git commit --amend --no-edit --signoff`) to make the DCO bot happy?

juliushetzel:
I changed the commit messages

tests/io/test_partitioned_dataset.py
```diff
@@ -401,7 +401,7 @@ def test_dataset_creds(self, pds_config, expected_ds_creds, global_creds):
         assert pds._credentials == global_creds
 
 
-BUCKET_NAME = "fake_bucket_name"
+BUCKET_NAME = "s3_fake_bucket_name"
```
marrrcin, line 404: Wouldn't it be better to cover both cases (the one before the fix and the one after)? Just to make sure 🤞🏻

astrojuanlu: Wdyt @noklam ?

noklam: I will make a follow up PR on this.

PR #2718, opened 2023-06-23
[spike] Introduce `tui` to better navigate the Kedro CLI

datajoely:
## What is this?

This is experimental spike 🧪  to introduce a TUI which simply builds upon the existing Kedro CLI and associated help messages to provide an intuitive and clean interface for users to explore and navigate the Kedro CLI.

This is achieved by leveraging the Trogon library from the people behind [Rich](https://rich.readthedocs.io/en/stable/introduction.html) and [Textual](https://textual.textualize.io/).

📣 I'm raising this to gauge interest, do you like it? would you use it? please let me know!

https://github.com/kedro-org/kedro/assets/35801847/efe8df80-f719-4e15-8619-50a55f431f34

## Development notes

- I'm not wedded to my implementation, but would love to discuss with the team if we think there is significant user value here.
- This wasn't the two line change that I was hoping for, Trogon doesn't support the `click.MultiCommand` construct so I've had to be a bit funky with `click.Groups`.
- I have zero idea why the unit tests work on Python 3.9+ and don't on 3.8 below (if it was 3.7 only [it would be so easy to ignore](https://endoflife.date/python)). 

## Checklist


noklam:
Does this mainly provide a GUI to construct the CLI command or you can also have callbacks executing some code but not from the CLI?

datajoely:
We could absolutely extended it, but the main value is that we get something for free for existing work today

datajoely:
Really good push @notniknot

datajoely:
@notniknot the licence issue has been addressed on the Textualize side. I will note that from one of my other projects the start up cost of Trogon is quite considerable, if we end up integrating it into Kedro we do need to find a way of initialising it in a lazy way

astrojuanlu:
Any way to turn this into a `kedro-tui` plugin that can be installed separately?

datajoely:
So the answer is almost certainly yes - but I would defer to an real engineer on how to do it, the hack I've had to do to merge commands to the right level feels like something that needs to be overcome.

stichbury:
Do you think this is a PR we could close for now @datajoely as it's been open over 6 months?

Issue #2270, opened 2023-01-31
Support python 3.11


sbrugman:
Are there any changes that need to be made, or is it as simple as updating the package information?

depends on #2742

merelcht:
That's a good question @sbrugman. I don't know on top of my head how much needs to be changed, but we can definitely add this to our backlog and start the work on it.

merelcht:
The first step is to add a CI build with 3.11 and see if that just works, both on `kedro` and `kedro-plugins`. If it turns out to be more involved, we should re-estimate or create a follow up ticket.

noklam:
Summary
See details in [CircleCI Logs](https://app.circleci.com/pipelines/github/kedro-org/kedro/19891/workflows/005d826f-c5b9-42af-93fb-d915bc006744/jobs/238590)
* Tensorflow isn't ready for python3.11 - so we can't add Python3.11 until 0.19.
* Spark 3.11 requires Spark >3.4 https://stackoverflow.com/questions/75048688/picklingerror-could-not-serialize-object-indexerror-tuple-index-out-of-range
* some more random tests breaking

Going to support this will probably take some effort, I would guess an 8 or more. 

Further questions
1. Do we ever drop DataSet supports?
Every dataset added is making CI more complex. In reality, no one really uses all of the datasets, but our CI has to take care of the combinations of all dependencies. This confuses pip as some dependencies are old and it hits weird combination, [sometimes it just stop CI to run completely](https://app.circleci.com/pipelines/github/kedro-org/kedro/19877/workflows/8bf11726-51c3-46e6-9a19-67b9bc41780b/jobs/238318) and it's difficult to figure out which package is the root cause due to 2nd or even 3rd level dependencies i.e. C depends on B depends on A, and CI complain about A. I can only follow a trial & error pattern and bump one library at a time.


2. How do we handle release for `kedro-datasets` when some libraries are not ready?
Is this considered a breaking change and we will bump the major version immediately? [Semantic Versioning](https://semver.org/). It also add more complexity to CI, we will need to skip certain test to run in a particular Python version. #2433 

3. Is there a way to handle (1) better?
It would be great if the team can share some wisdom, or how to approach this problem systematically. I haven't find a better way to make small changes incrementally. It could be just a quick discussion, if there is no better way it's fine too.

4. How do we handle versions conflict?
[...]

PR #2417, opened 2023-03-13
Attempt to bump python version to 3.11 

noklam:
Signed-off-by: Nok Chan <nok.lam.chan@quantumblack.com>

> **NOTE:** Kedro datasets are moving from `kedro.extras.datasets` to a separate `kedro-datasets` package in
> [`kedro-plugins` repository](https://github.com/kedro-org/kedro-plugins). Any changes to the dataset implementations
> should be done by opening a pull request in that repository.
## Description

See what's failing in CI and estimate the effort of actually supporting newer version.

Investigation of #2048 #2270 

## Checklist

- [ ] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [ ] Added tests to cover my changes

PR #2742, opened 2023-06-28
Allow any version of Python (including 3.11) in metadata

astrojuanlu:
See https://iscinumpy.dev/post/bound-version-constraints/#pinning-the-python-version-is-special, and discussion in https://discuss.python.org/t/requires-python-upper-limits/12663?u=astrojuanlu, as well as practical problems with leaving the version cap on in https://github.com/kedro-org/kedro/issues/2270#issuecomment-1611641812.

Deliberately _not_ adding anything to the changelog, and not touching our CI systems either.

## Checklist

### DIFF
--- a/tests/test_import.py
+++ b/tests/test_import.py
@@ -1,11 +1,32 @@
 import importlib
 import warnings
 import pytest
-from unittest.mock import patch
 
 def test_import_kedro_with_no_official_support_raise_error(mocker, monkeypatch):
     """Test importing kedro should fails with python>=3.11 should fails"""
     import kedro
+    with mocker.patch("kedro.sys"):
+        kedro.__loader__.exec_module(kedro)
+
+@pytest.mark.filterwarnings("default:Kedro")
+def test_import_kedro_success(mocker):
+    """Test importing kedro should fails with python>=3.11 should fails"""
+    import kedro
+    with mocker.patch("kedro.sys"):
+        kedro.__loader__.exec_module(kedro)
+
+def test_import_kedro_success_with_env_variable(mocker, monkeypatch):
+    """Test importing kedro should fails with python>=3.11 should fails"""
+    import kedro
+    monkeypatch.setenv("PYTHONWARNINGS", "default:Kedro")
+    with mocker.patch("kedro.sys"):
+        kedro.__loader__.exec_module(kedro)
+
+def test_import_kedro_success_with_env_variable(mocker, monkeypatch):
+    """Test importing kedro should fails with python>=3.11 should fails"""
+    import kedro
+    with mocker.patch("kedro.sys"):
+        kedro.__loader__.exec_module(kedro)
     
 @pytest.mark.filterwarnings("default:Kedro")
 def test_import_kedro_slient_warning(mocker, monkeypatch):
--- a/tests/test_import.py
+++ b/tests/test_import.py
@@ -1,0 +1,13 @@
+import importlib
+import warnings
+import pytest
+from unittest.mock import patch
+
+def test_import_kedro_with_no_official_support_raise_error(mocker, monkeypatch):
+    """Test importing kedro should fails with python>=3.11 should fails"""
+    import kedro
+    
+@pytest.mark.filterwarnings("default:Kedro")
+def test_import_kedro_slient_warning(mocker, monkeypatch):
+    """Test importing kedro when warning is silent"""
+    import kedro
--- a/tests/test_import.py
+++ b/tests/test_import.py
@@ -1,34 +1,27 @@
-import importlib
-import warnings
 import pytest
 
-def test_import_kedro_with_no_official_support_raise_error(mocker, monkeypatch):
-    """Test importing kedro should fails with python>=3.11 should fails"""
-    import kedro
-    with mocker.patch("kedro.sys"):
+import kedro
+
+
+def test_import_kedro_with_no_official_support_raise_error(mocker):
+    """Test importing kedro with python>=3.11 should fail"""
+    mocker.patch("kedro.sys.version_info", (3, 11))
+
+    # We use the parent class to avoid issues with `exec_module`
+    with pytest.raises(UserWarning) as excinfo:
         kedro.__loader__.exec_module(kedro)
 
-@pytest.mark.filterwarnings("default:Kedro")
-def test_import_kedro_success(mocker):
-    """Test importing kedro should fails with python>=3.11 should fails"""
-    import kedro
-    with mocker.patch("kedro.sys"):
+    assert "Kedro is not yet fully compatible" in str(excinfo.value)
+
+
+def test_import_kedro_with_no_official_support_emits_warning(mocker):
+    """Test importing kedro python>=3.11 and controlled warnings should work"""
+    mocker.patch("kedro.sys.version_info", (3, 11))
+    mocker.patch("kedro.sys.warnoptions", ["default:Kedro is not yet fully compatible"])
+
+    # We use the parent class to avoid issues with `exec_module`
+    with pytest.warns(UserWarning) as record:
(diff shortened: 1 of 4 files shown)


astrojuanlu:
We've decided to wait a bit to hear from @merelcht and @idanov

astrojuanlu:
Prompted by @idanov, I added a warning that by default is treated like an error:

```
❯ kedro info                                                                                                                                                                                                                                                                       (kedro311-dev) 
  File "/Users/juan_cano/.micromamba/envs/kedro311-dev/bin/kedro", line 5, in <module>
    from kedro.framework.cli import main
  File "/Users/juan_cano/Projects/QuantumBlack Labs/kedro/kedro/__init__.py", line 20, in <module>
    warnings.warn(
kedro.KedroPythonVersionWarning: Kedro is not yet fully compatible with this Python version.
To proceed at your own risk and ignore this warning,
run Kedro with `python -W "default:Kedro is not yet fully compatible" -m kedro ...`
or set the PYTHONWARNINGS environment variable accordingly.
```

And can be treated as a normal warning or ignored in two different ways (see https://docs.python.org/3/library/warnings.html#describing-warning-filters):

```
❯ python -W "default:Kedro is not yet fully compatible" -m kedro info                                                                                                                                                                                                              (kedro311-dev) 
/Users/juan_cano/Projects/QuantumBlack Labs/kedro/kedro/__init__.py:20: KedroPythonVersionWarning: Kedro is not yet fully compatible with this Python version.
To proceed at your own risk and ignore this warning,
run Kedro with `python -W "default:Kedro is not yet fully compatible" -m kedro ...`
or set the PYTHONWARNINGS environment variable accordingly.
  warnings.warn(

 _            _
| | _____  __| |_ __ ___
| |/ / _ \/ _` | '__/ _ \
|   <  __/ (_| | | | (_) |
|_|\_\___|\__,_|_|  \___/
v0.18.10

Kedro is a Python framework for
creating reproducible, maintainable
and modular data science code.

No plugins installed
❯ PYTHONWARNINGS="ignore:Kedro is not yet fully compatible" kedro info                                                                                                                                                                                                             (kedro311-dev) 

 _            _
| | _____  __| |_ __ ___
| |/ / _ \/ _` | '__/ _ \
|   <  __/ (_| | | | (_) |
|_|\_\___|\__,_|_|  \___/
v0.18.10

Kedro is a Python framework for
creating reproducible, maintainable
and modular data science code.

No plugins installed
```

(custom categories can't be specified, see https://github.com/python/cpython/issues/66733)

merelcht (approved):
LGTM! 👍

noklam:
Do we need to merge this in #2417? Not sure how we can test this without adding python 3.11 first, as the test will never be triggered in the CI and basically do nothing for 3.7 - 3.10.

astrojuanlu:
Merging this into gh-2417 doesn't really solve the bootstrapping problem (that there has to be a 3.11-enabled version to unblock the CI).

On the other hand, the plan is to test these changes - it's just that we'll have to mock the `sys.version_info` check, rather than using a real Python 3.11 environment.

noklam:
@astrojuanlu It turns out to be tricky with `sys` modules, I haven't found a way to handle version switch nicely.

A few challenges here
1. To use `importlib.reload`, it takes a module, so you need to imported the module already (which we don't if we expect the test fail). So I use a `setup_environment` fixture to bypass it and reset the `PYTHONWARNING`, which I think is fine.
2. mocking `sys` directly will cause error because this interfere with `pytest`, the correct way to do it is always mock the local module which is `kedro.sys`. This is fine, but when we do `importlib.reload(kedro)` we effectively remove `kedro.sys` so we the mock that we insert is removed.

Any idea how it could work?

astrojuanlu:
Ugh, I see what you mean. I'll give it some thought.

astrojuanlu:
The key is to not use `reload`:

```
In [1]: import importlib

In [2]: import kedro

In [3]: kedro.__loader__.exec_module(kedro)

In [4]: from unittest.mock import patch

In [5]: with patch("kedro.sys.version_info", (3, 11)):
   ...:     kedro.__loader__.exec_module(kedro)
   ...: 
---------------------------------------------------------------------------
KedroPythonVersionWarning                 Traceback (most recent call last)
Cell In[5], line 2
      1 with patch("kedro.sys.version_info", (3, 11)):
----> 2     kedro.__loader__.exec_module(kedro)

File <frozen importlib._bootstrap_external>:883, in exec_module(self, module)

File <frozen importlib._bootstrap>:241, in _call_with_frames_removed(f, *args, **kwds)

File ~/Projects/QuantumBlack Labs/kedro/kedro/__init__.py:22
     19     warnings.simplefilter("error", KedroPythonVersionWarning)
     21 if sys.version_info >= (3, 11):
---> 22     warnings.warn(
     23         """Kedro is not yet fully compatible with this Python version.
     24 To proceed at your own risk and ignore this warning,
     25 run Kedro with `python -W "default:Kedro is not yet fully compatible" -m kedro ...`
     26 or set the PYTHONWARNINGS environment variable accordingly.""",
     27         KedroPythonVersionWarning,
     28     )

KedroPythonVersionWarning: Kedro is not yet fully compatible with this Python version.
To proceed at your own risk and ignore this warning,
run Kedro with `python -W "default:Kedro is not yet fully compatible" -m kedro ...`
or set the PYTHONWARNINGS environment variable accordingly.
```

noklam:
Weird, `pytest` seem failing to catch the warning as an exception and the test just fail. I temporarily switch the GitPod version to 3.11 so it's easier to test.

tests/test_import.py
```diff
@@ -0,0 +1,34 @@
+import importlib
+import warnings
+import pytest
+
+def test_import_kedro_with_no_official_support_raise_error(mocker, monkeypatch):
+    """Test importing kedro should fails with python>=3.11 should fails"""
+    import kedro
+    with mocker.patch("kedro.sys"):
```
astrojuanlu, line 8: As they stand, the result of these tests depends on the Python version they're ran on, so we are forced to patch `kedro.sys.version_info` I'd say (both with the `(3, 11)` case and the `(3, 10)` one)

noklam: I understand this, two questions I have here:
1. I expect `mocker.patch` is same as `unittest.mock.patch`, but `mocker.patch("kedro.sys.version_info")` is not allowed but `patch("kedro.sys.version_info")` is allowed. I guess I won't be too obsess with `pytest` here and just use `unittest.mock`
2. All tests are failing which I am not sure why.

I will make parmaterize it with success(3.10) and fail case (3.11) later, but need to make 2. works first.

astrojuanlu: TIL: https://pytest-mock.readthedocs.io/en/latest/remarks.html#usage-as-context-manager

noklam: Ah yes, but I keep switching between mock and mocker so I mess this up.

I don't think the context manager is the problem here since removing the with have the same result, i could explain more next Monday.

astrojuanlu: I pushed a proposal, let me know what you think.

Indeed there were some weird things with reloading the module, and the interaction between the various warning control options. For example, `pytest.mark.filterwarnings` was not setting `sys.warnoptions`, so it was not working as intended and I had to mock instead. Setting the environment variable with monkeypatch didn't work for me either. And finally, catching the `KedroPythonVersionWarning` didn't work, so I used the parent class instead.

SajidAlamQB (approved):
This PR looks great to me! 🚀 I think changing it to a warning when using unsupported Python versions is a good approach.

I appreciate your attention to detail in this PR. Thank you @astrojuanlu!

astrojuanlu:
Thanks a bunch @SajidAlamQB ! This is going in

noklam:
🚀🚀🚀🚀

Issue #2123, opened 2022-12-14
Investigate if `micropkg` spec in `pyproject.toml` is used


merelcht:
## Description
In https://github.com/kedro-org/kedro/pull/2119 it was discovered that the feature to add `micropkg` specs in `pyproject.toml` hasn't worked since March 2022. Not a single user reported this.

There are still some open issues regarding the `micropkg` functionality in `pyproject.toml`:
- https://github.com/kedro-org/kedro/issues/1304
- https://github.com/kedro-org/kedro/issues/1478

Before we spend more time implementing and expanding this functionality we need to investigate if `micropkg` spec in `pyproject.toml` is used at all. 

## Context
- It would also be worth talking to the Alloy team and hearing how they do packaging and if they find the micro-packaging functionality in Kedro useful or not.
- Look at the `micropkg` usage stats on telemetry

lamhm:
I am currently affected by the bug #2119. I hope the fix will be released soon, so that I can convince my team to switch to `micropkg` instead of using our in-house method (which is just a half-baked solution).

Here is why I only start trying `micropkg` now, and not before. I have been using Kedro since late 2020, starting with only one-person projects. In 2021, when I got assigned into a new team of Data Scientists, I realised that all my co-workers only coded in Jupyter notebooks and it was very hard to replicate their work outside their local environment. So, I adopted Kedro for the whole team. Since then, Kedro has played a very important role in our workflow.

If I recall correctly, the first time I heard about `micropkg` was late 2021. But I did not pay much attention to it since the standard method (`wheel`) was good enough for us. This is because my team at the time only worked on developing and training ML models. When our models reached an acceptable quality, we packaged them and sent them to the Software Engineering Team. The Engineering team will take care of the deployment. Therefore:

- Packaging was not a bottleneck as we did not do it very frequently. And we only needed to package whole projects (not parts of them).
- It was not easy for us to change our packaging method as we would need an agreement with the Engineering Team.

But now I am working in an R&D team that develop ML solutions from start to finish. I.e. we are in charge of not only developing and training models but also deploying models as software applications and continuously upgrading them. So, we have total control of the packaging and deployment methods for our ML models/pipelines. I can see an opportunity to streamline our model deployment (which happens very frequently now) by using `micropkg`. That is why I am about to convince my team to switch to `micropkg`. But I have to postpone this plan because of the bug above (I know how to work around it, but I don't want my team to go that way).

lamhm:
[...]

PR #2761, opened 2023-07-03
Make `kedro micropkg package` use `pyproject.toml`

astrojuanlu:
## Description
Fix gh-2414.

Initially, I tried to be smart and also rework how `requirements.txt` files are handled. But in the end I opted for this simple solution (just replacing `setup.py` with `pyproject.toml`) and it looks like it's simple and it works.

Thanks to gh-2614, not many code changes were required because essentially the `pull` code is now packaging-agnostic. The only part I had to touch was the on-the-fly generation of the file. Not even the tests needed to be touched.

## Checklist

### DIFF
--- a/kedro/framework/cli/micropkg.py
+++ b/kedro/framework/cli/micropkg.py
@@ -36,17 +36,19 @@ from kedro.framework.cli.utils import (
 )
 from kedro.framework.startup import ProjectMetadata
 
-_SETUP_PY_TEMPLATE = """# -*- coding: utf-8 -*-
-from setuptools import setup, find_packages
+_PYPROJECT_TOML_TEMPLATE = """
+[build-system]
+requires = ["setuptools"]
+build-backend = "setuptools.build_meta"
 
-setup(
-    name="{name}",
-    version="{version}",
-    description="Micro-package `{name}`",
-    packages=find_packages(),
-    include_package_data=True,
-    install_requires={install_requires},
-)
+[project]
+name = "{name}"
+version = "{version}"
+description = "Micro-package `{name}`"
+dependencies = {install_requires}
+
+[tool.setuptools.packages]
+find = {{}}
 """
 
 
@@ -212,7 +214,7 @@ def _pull_package(
         # without making assumptions on the project metadata.
         library_meta = project_wheel_metadata(project_root_dir)
 
-        # Project name will be `my-pipeline` even if `setup.py` says `my_pipeline`
+        # Project name will be `my-pipeline` even if `pyproject.toml` says `my_pipeline`
         # because standards mandate normalization of names for comparison,
         # see https://packaging.python.org/en/latest/specifications/core-metadata/#name
         # The proper way to get it would be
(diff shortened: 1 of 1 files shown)


noklam (commented):
I don't have many experience with `micropkg` before, I have few comments but not necessary related to this PR.

First,  I expect `kedro pipeline create` has symmetry to `kedro micropkg package`. i.e. `kedro pipeline create nok` then `kedro micropkg package nok` should work, but it doesn't because I need to do `kedro micropkg pipelines.nok` which isn't trivial to find in the doc. I end up have to look at the e2e test. The documentation doesn't tell me I need to do this https://docs.kedro.org/en/latest/nodes_and_pipelines/micro_packaging.html

2.  Error message is not very helpful for `--all`, I am pretty lost when it talks about manifest and

> (kedro_core) pattern main % kedro micropkg package              
> Please specify a micro-package name or add '--all' to package all micro-packages in the 'pyproject.toml' package manifest section.
> (kedro_core) pattern main % kedro micropkg package --all
> Nothing to package. Please update the 'pyproject.toml' package manifest section.

I also did a manual test to create micropkg and install it
```
kedro micropkg pipelines.nok
pip install dist/nok-0.1.tar.gz
```

This works fine but when I try to uninstall it I notice it is also in `test` folder which I didn't expect.

> (kedro_core) pattern main % pip uninstall nok              
> Found existing installation: nok 0.1
> Uninstalling nok-0.1:
>   Would remove:
>     /Users/Nok_Lam_Chan/miniconda3/envs/kedro_core/lib/python3.8/site-packages/nok-0.1.dist-info/*
>     /Users/Nok_Lam_Chan/miniconda3/envs/kedro_core/lib/python3.8/site-packages/nok/*
>     /Users/Nok_Lam_Chan/miniconda3/envs/kedro_core/lib/python3.8/site-packages/tests/*  <- what is this?
>

astrojuanlu:
Thanks a lot @noklam ! 

> First, I expect kedro pipeline create has symmetry to kedro micropkg package. i.e. kedro pipeline create nok then kedro micropkg package nok should work, but it doesn't because I need to do kedro micropkg pipelines.nok which isn't trivial to find in the doc.

This is good feedback, we should add that to https://github.com/kedro-org/kedro/issues/1536

> Error message is not very helpful for --all, I am pretty lost when it talks about manifest and

This works on `main` as well right?

> I also did a manual test to create micropkg and install it

Couldn't reproduce this 🤔 but I have an idea of how it could have happened.

merelcht (approved):
Nice and simple solution 👍 

I think @noklam raised some very good points, which we need to address. If no tickets exist yet for these issues, let's add them to: https://github.com/kedro-org/kedro/milestone/21

astrojuanlu:
Spread these comments over gh-1536, gh-2123, gh-2766 👍🏽

noklam:
> > Error message is not very helpful for --all, I am pretty lost when it talks about manifest and
> 
> This works on `main` as well right?

I test it with `main`, the behavior is the same so it is not introduced by this PR, it's just a minor documentation improvement suggestion.

SajidAlamQB:
Will this still work for users who are still using setup.py in their existing Kedro projects? Do we need to make any considerations for them?

astrojuanlu:
Yes indeed - in fact, currently the templates still use `setup.py` (because #2280 isn't closed yet) and this still works.

That's for `kedro micropkg package` with older projects. For `kedro micropkg pull` that pulls older micropackages, it will work because after #2614 the machinery is packaging-agnostic. ✨

SajidAlamQB (approved):

PR #2768, opened 2023-07-05
Safe guard KedroSession for Runner argument

noklam:
## Description

This is created as a quick attempt to fix thing as it was asked in Slack today.

Currently when user passed a Runner class instead of Runner(), Kedro gives a very subtle error which is not helpful. We can add a simple checks here to make sure user don't fall into the trap.

This is the error you get if you pass `session.run(runner=SequentialRunner)` instead of `session.run(runner=SequentialRunner())`
```
     54 """Run the ``Pipeline`` using the datasets provided by ``catalog``
     55 and save results back to the same objects.
     56 
   (...)
     70 
     71 """
     73 hook_manager = hook_manager or _NullPluginManager()
---> 74 catalog = catalog.shallow_copy()
     76 unsatisfied = pipeline.inputs() - set(catalog.list())
     77 if unsatisfied:

AttributeError: 'PluginManager' object has no attribute 'shallow_copy'
```

Same goes for Hook when we expect a instance `Hook()` instead of `Hook`, this confuse user sometimes.
## Checklist

### DIFF
--- a/kedro/framework/session/session.py
+++ b/kedro/framework/session/session.py
@@ -417,6 +417,10 @@ class KedroSession:
         # Run the runner
         hook_manager = self._hook_manager
         runner = runner or SequentialRunner()
+        if not isinstance(runner, AbstractRunner):
+            raise KedroSessionError(
+                "KedroSession expect an instance of Runner, make sure you are passing an instance instead of a class."
+            )
         hook_manager.hook.before_pipeline_run(  # pylint: disable=no-member
             run_params=record_data, pipeline=filtered_pipeline, catalog=catalog
         )
--- a/kedro/framework/session/session.py
+++ b/kedro/framework/session/session.py
@@ -406,7 +406,7 @@ class KedroSession:
             "runner": getattr(runner, "__name__", str(runner)),
         }
 
-        catalog = context._get_catalog(
+        catalog = context._get_catalog(  # pylint: disable=protected-access
             save_version=save_version,
             load_versions=load_versions,
         )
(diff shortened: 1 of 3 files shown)


kedro/framework/session/session.py
```diff
@@ -417,6 +417,10 @@ def run(  # pylint: disable=too-many-arguments,too-many-locals
         # Run the runner
         hook_manager = self._hook_manager
         runner = runner or SequentialRunner()
+        if not isinstance(runner, AbstractRunner):
+            raise KedroSessionError(
+                "KedroSession expect an instance of Runner, make sure you are passing an instance instead of a class."
```
datajoely, line 422:
```suggestion
                "KedroSession expect an instance of Runner, make sure you are passing an instance instead of a class. Have you forgotten the `()` at the end of the statement?"
```

Is this useful or overkill?

datajoely (commented):
Really really like it

datajoely:
Should we do the same for hooks too?

kedro/framework/session/session.py
astrojuanlu: I'd say let's keep it! Mostly harmless, but can go a long way unpacking the jargon

noklam: Maybe
```
KedroSession expect an instance of Runner instead of a class. Have you forgotten the `()` at the end of the statement?
```

noklam:
Let's keep this PR small, I will add Hooks later since I think there are something more to consider (or I am just overthinking)

kedro/framework/session/session.py
datajoely: Love it!

merelcht (approved):
Thanks for improving this @noklam ! I think this is a much nicer experience 👍

datajoely (approved):

PR #2770, opened 2023-07-05
Add TypeAlias for datasets pattern and refactoring

noklam:
## Checklist

### DIFF
--- a/kedro/io/data_catalog.py
+++ b/kedro/io/data_catalog.py
@@ -14,6 +14,7 @@ from collections import defaultdict
 from typing import Any, Iterable
 
 from parse import parse
+from typinh import TypeAli
 
 from kedro.io.core import (
     AbstractDataSet,
@@ -29,6 +30,8 @@ from kedro.io.memory_dataset import MemoryDataset
 CATALOG_KEY = "catalog"
 CREDENTIALS_KEY = "credentials"
 WORDS_REGEX_PATTERN = re.compile(r"\W+")
+
+Patterns = dict[str, dict[str, Any]]
 
 
 def _get_credentials(
@@ -186,8 +189,8 @@ class DataCatalog:
         self.layers = layers
         # Keep a record of all patterns in the catalog.
         # {dataset pattern name : dataset pattern body}
-        self._dataset_patterns = dict(dataset_patterns or {})
-        self._load_versions = dict(load_versions or {})
+        self._dataset_patterns = dataset_patterns or {}
+        self._load_versions = load_versions or {}
         self._save_version = save_version
 
         if feed_dict:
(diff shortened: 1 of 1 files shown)


kedro/io/data_catalog.py
```diff
@@ -30,6 +30,8 @@
 CREDENTIALS_KEY = "credentials"
 WORDS_REGEX_PATTERN = re.compile(r"\W+")
 
+Patterns = dict[str, dict[str, Any]]
```
noklam, line 33: Is this an overkill? I found that we don't do this in `kedro` other than datasets which use `DI_` and `DO_` for data input/output. In this case I do think it has value as it helps a lot to read the code especially when we have a lot of dict of dict flowing around everywhere, I think this put some context.

@astrojuanlu WDYT about using type alias?

astrojuanlu: I think this is a good use of a type alias! Not sure the lowercase `dict` will work for them (but if they do, fantastic - I sometimes get lost with Python typing)

kedro/io/data_catalog.py
```diff
@@ -319,14 +321,10 @@ class to be loaded is specified with the key ``type`` and their
     @staticmethod
     def _is_pattern(pattern: str):
         """Check if a given string is a pattern. Assume that any name with '{' is a pattern."""
-        if "{" in pattern:
-            return True
-        return False
+        return True if "{" in pattern else False
```
astrojuanlu, line 324:
```suggestion
        return "{" in pattern
```

🎩

kedro/io/data_catalog.py
```diff
@@ -186,8 +188,8 @@ def __init__(  # pylint: disable=too-many-arguments
         self.layers = layers
         # Keep a record of all patterns in the catalog.
         # {dataset pattern name : dataset pattern body}
-        self._dataset_patterns = dict(dataset_patterns or {})
-        self._load_versions = dict(load_versions or {})
+        self._dataset_patterns = dataset_patterns or {}
```
astrojuanlu, line 191: This means that `dataset_patterns` don't get converted to `dict` anymore right? Hope there's not any weird corner case

kedro/io/data_catalog.py
noklam: lol True

kedro/io/data_catalog.py
noklam: dataset_patterns is already a dict[str, dict[str, Any]], no?

astrojuanlu: Indeed, I meant to say that I don't know how this function is used - if it's internal and we _know_ that the inputs are dictionaries already, then all is fine 👍🏽

Issue #2776, opened 2023-07-06
Introduce Type Alias for Patterns and more


noklam:
# Background
As a follow up action, we agreed to introduce type like `Patterns = Dict[str, Dict[str, Any]]` to improve the readability of the code a lot. 

An example can be found [here](https://github.com/kedro-org/kedro/pull/2770/files#diff-83c76b6d924034cc585ba09c9fcb10161887b633b74e8b09a91e4803200f0ed9R33)

We can also do this beyond just `Patterns` but to more complicated types within the code base.

> 
>   As discussed, let's separate the Type Alias proposed in #2770.
>  I will close #2770 and approve this PR now. :)
> 
> _Originally posted by @noklam in https://github.com/kedro-org/kedro/issues/2635#issuecomment-1623710496_
>

noklam:
This could be nicely rewritten to:
```
matches = (parse(pattern, data_set_name) for pattern in data_set_patterns.keys())
return next(filter(None, matches), None)
```

_Originally posted by @idanov in https://github.com/kedro-org/kedro/pull/2635#discussion_r1255521107_
            

There are some minor refactoring work linked here, we can address this with this change.

noklam:
Closed with #2779

PR #2779, opened 2023-07-07
Add type alias for dataset factory patterns 

ankatiyar:
## Description

Related to #2776 

Including @noklam's suggestion about using type alias for patterns and @idanov's suggestions on the merged PR for dataset factories #2635 

## Checklist

### DIFF
--- a/kedro/io/data_catalog.py
+++ b/kedro/io/data_catalog.py
@@ -11,7 +11,7 @@ import difflib
 import logging
 import re
 from collections import defaultdict
-from typing import Any, Iterable
+from typing import Any, Dict, Iterable
 
 from parse import parse
 
@@ -25,6 +25,8 @@ from kedro.io.core import (
     generate_timestamp,
 )
 from kedro.io.memory_dataset import MemoryDataset
+
+Patterns = Dict[str, Dict[str, Any]]
 
 CATALOG_KEY = "catalog"
 CREDENTIALS_KEY = "credentials"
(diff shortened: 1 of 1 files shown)


kedro/io/data_catalog.py
```diff
@@ -322,20 +324,17 @@ def _is_pattern(pattern: str):
...
-        for pattern, _ in data_set_patterns.items():
-            result = parse(pattern, data_set_name)
-            if result:
-                return pattern
-        return None
+        matches = [
+            pattern
+            for pattern in data_set_patterns.keys()
+            if parse(pattern, data_set_name)
+        ]
```
idanov, line 333: This one is important to be a generator, so you compute things lazily, i.e. only until you find a match. Also no need to filter here, the filter is applied in the return.
```suggestion
        matches = (
            parse(pattern, data_set_name)
            for pattern in data_set_patterns.keys()
        )
```

kedro/io/data_catalog.py
```diff
@@ -322,20 +324,17 @@ def _is_pattern(pattern: str):
...
+        return next(filter(None, matches), None)
```
idanov, line 334: It's better to make it a generator, so we only parse until we find a match. Also the second filter is not needed here anymore because of the filter you've added in the generator.
```suggestion
        matches = (
            pattern
            for pattern in data_set_patterns.keys()
            if parse(pattern, data_set_name)
        )
        return next(matches, None)
```

idanov (approved):
⭐️

noklam (approved):
Looks good! It's more clean and easier to read!

Issue #2669, opened 2023-06-12
Update documentation around credentials management


noklam:
## Description
Short description of the problem here.
```yml
parquet_dataset:
  type: dask.ParquetDataSet
  filepath: "s3://bucket_name/path/to/folder"
  credentials:
    client_kwargs:
      aws_access_key_id: YOUR_KEY
      aws_secret_access_key: "YOUR SECRET"
```
This is the way we mentioned how to provide credentials in Kedro's doc. However `fsspec` has update the API for quite a while and if you are using newer version of fsspec, you should use `key`,`secret` instead of `aws_access_key_id` instead.

It could be only affecting `s3fs` (This is how I bump into error), but potentially affect `gcs` and more.

## Context
The docs on credentials are out of date and mention wrong key names. All doc chapters mentioning credentials should be updated to use the correct keys.

astrojuanlu:
Today I was helping @ricardopicon-mck and it was not clear how to use Google Cloud credentials. There are excellent examples of how to set up the `catalog.yml`:

https://docs.kedro.org/en/stable/data/data_catalog_yaml_examples.html#load-an-excel-file-from-google-cloud-storage

But how does `credentials.yml` look in that case?

For the record, this did the trick for me:

```yaml
gcp_credentials:
  token: gcp_credentials.json
```

But this only worked with a flat file structure. When having a full-fledged Kedro project with `conf/base` and `conf/local`, I had to specify the absolute path:

```yaml
gcp_credentials:
  token: /Users/juan_cano/Projects/QuantumBlack Labs/tmp/test-credentials/conf/local/gcp_credentials.json
```

I'm sure there is a better way.

In general, the credentials page is not very useful: https://docs.kedro.org/en/stable/configuration/credentials.html

It places a lot of emphasis in how to load them from code, but I'd consider this "advanced" or "programmatic" usage, which is not how most users experience Kedro. 

(see also https://github.com/fsspec/gcsfs/issues/583)

stichbury:
That's a good point and this page needs a clean up to bring up to the same standards as the recent data catalog updates.

datajoely:
See this for reference
https://github.com/kedro-org/kedro/issues/3164

astrojuanlu:
[...]

Issue #1646, opened 2022-06-23
Credentials in kedro


antonymilne:
# How do credentials currently work in kedro?

The basic pattern is as follows:
```yaml
# conf/base/catalog.yml
dataset_name:
  ...
  credentials: credentials_key

# conf/local/credentials.yml
credentials_key: 
  kwarg1: value1
  kwarg2: value2
  ...
```

To be concrete, here's an example for Azure Blob Storage:
```yaml
# conf/base/catalog.yml
shuttles:
  type: pandas.CSVDataSet
  filepath: abfs://somewhere/shuttles.csv
  credentials: abs_credentials

# conf/local/credentials.yml
abs_credentials:
  account_name: antonymilne
  account_key: verysecretpassword
```

The `credentials` key is injected into the call that instantiates `pandas.CSVDataSet` when kedro is run. Specifically, here: https://github.com/kedro-org/kedro/blob/a925fd59187a642e124527f0f1097e92ea8d1819/kedro/io/data_catalog.py#L276

Note:
* `credentials` is a special reserved keyword. This doesn't work for any other key name
* this is one of the very few customisations that kedro's `ConfigLoader` makes to how yaml is parsed. In-file variable injection is (kind of) supported in yaml using anchors, but injecting a variable from another file is not. The mechanism that does the injection here is entirely defined by kedro
* the reason for enabling this custom behaviour is that credentials should not be committed to source control. Hence they need to be stored in a separate file outside the data catalog that lives in `local` and injected into the catalog at runtime

# Does this work well?

In my experience and from talking to users: _in the case that credentials can be stored in a file_, yes. Very little confusion is caused by the custom behaviour of injecting credentials.

# What are the problems with this?

The biggest problem is that credentials might not be stored in a file. Alternatives are:
* very common: storing credentials in environment variables rather than files. kedro can [deal with this through `TemplatedConfigLoader`](https://web.archive.org/web/20210921074814/https://kedrozerotohero.com/programming-patterns/how-to-inject-secrets-into-your-kedro-configuration). This works ok but feels hacky and is so common it shouldn't really require a workaround
[...]

PR #2787, opened 2023-07-11
Update on credentials.md

jmnunezd:
Updating example code in credentials.md to make it usable when copy/pasting. Since as of now it will generate a TypeError

## Description

I was reading the docs on how to load credentials directly in code, I tried using the copy button and paste it in my local to test it. 

Screenshot 2023-07-11 at 2 40 36 PM

The problem is that even if you change `project_path` you will still get an TypeError since Python won't allow to use "/" between two strings.

**For example:**
`conf_path = str(project_path / settings.CONF_SOURCE)`     changing project_path:
`conf_path = str("my_project_path" /  "conf")`                        <----- Will raise TypeError

**Specifically:**
`TypeError: unsupported operand type(s) for /: 'str' and 'str'`

## Development notes

I changed the credentials.md from:

`conf_path = str(project_path / settings.CONF_SOURCE)`

To:
`conf_path = f"{project_path}/{settings.CONF_SOURCE}"`

## Checklist

- [N/A] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [N/A] Updated the documentation to reflect the code changes
- [N/A] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [N/A] Added tests to cover my changes


docs/source/configuration/credentials.md
```diff
@@ -15,7 +15,7 @@ The following examples all use the default `ConfigLoader` class.
 from kedro.config import ConfigLoader
 from kedro.framework.project import settings
 
-conf_path = str(project_path / settings.CONF_SOURCE)
+conf_path = f"{project_path}/{settings.CONF_SOURCE}"
```
astrojuanlu, line 18: Now that I look at this, where does `project_patg` come from?

Is the `TypeError` because you filled a string and this was expecting a `pathlib.Path` ?

jmnunezd: In my case I'm using project_path = os.getcwd() so it is a str

jmnunezd: > Now that I look at this, where does `project_patg` come from?
> 
> Is the `TypeError` because you filled a string and this was expecting a `pathlib.Path` ?

ohh ok! I thought that project_path was a string, so you say it should be a pathlib.Path object?

astrojuanlu: Correct! But the fact that you got it wrong is a documentation bug 😁 could you revert these changes and add a line at the beginning that assigns project_path as a Path object?

stichbury: Hi @jmnunezd -- would you be able to make the change suggested so I can approve and merge please?

jmnunezd: so it will be adding to the docs something like:

```
from pathlib import Path
project_path = Path(my_project_path)

conf_path = str(project_path / settings.CONF_SOURCE)

...
```

right?

astrojuanlu: Correct @jmnunezd! Please give it a try just to verify if it works. It should be a matter of adding the first 2 lines, as I understand that the third one was already there.

jmnunezd: Done! It's now ready for a new review

jmnunezd: I imported Path before kedro to make it compliant with [PEP8 import's order ](https://peps.python.org/pep-0008/#imports)

```
from pathlib import Path

from kedro.framework.project import settings
```

docs/source/configuration/credentials.md
```diff
@@ -12,10 +12,12 @@ Credentials configuration can be loaded the same way as any other project config
 The following examples all use the default `ConfigLoader` class.
 
 ```python
+from pathlib import Path
+
 from kedro.framework.project import settings
 
-conf_path = str(project_path / settings.CONF_SOURCE)
+conf_path = str(Path(project_path) / settings.CONF_SOURCE)
```
astrojuanlu, line 20: `project_path` needs to be assigned

astrojuanlu (changes requested):
The code snippet still fails because `project_path` is not defined, right?

jmnunezd:
> The code snippet still fails because `project_path` is not defined, right?

That's right! but again, I think the user should be able to manually give the path they want to use and assign it to project_path, does that makes sense? what would be the alternative?

noklam:
@jmnunezd Late to this conversation - Maybe we should just keep it consistent with `<project_root>` to indicate that this is something user need to define themselve?

stichbury:
It'd be great to get this in @jmnunezd as we are close to making a release and these changes have been on the cards a while. How can I help push it over the line? cc @astrojuanlu @noklam

noklam:
@stichbury I think it's good to check our terminology, is it `project_root` or `project_path`? The code snippet here is not working because `project_path` does not exist. Would it be enough to change it to `<project_root>` to indicate that this is something user need to put in themselves since it is different for every case.

stichbury:
> @stichbury I think it's good to check our terminology, is it `project_root` or `project_path`? The code snippet here is not working because `project_path` does not exist. Would it be enough to change it to `<project_root>` to indicate that this is something user need to put in themselves since it is different for every case.

Thanks @noklam yes I agree. We use this term reasonably consistently and define it here https://docs.kedro.org/en/stable/tutorial/spaceflights_tutorial.html#terminology so I think we should change it (and maybe put a comment in the code to explain?

docs/source/configuration/credentials.md
stichbury, line 20:
```suggestion
# Substitute <project_root> with the [root folder for your project](https://docs.kedro.org/en/stable/tutorial/spaceflights_tutorial.html#terminology)
conf_path = str(Path(<project_root>) / settings.CONF_SOURCE)
```

stichbury (approved):

noklam (approved):

astrojuanlu (approved):

astrojuanlu:
Thanks @jmnunezd for your patience and for kicking this off!

astrojuanlu:
After merging this we should probably add a more practical example on how credentials are normally used, see also #2669 (and more broadly, #1646)

PR #2793, opened 2023-07-13
Update `kedro catalog list` command to account for dataset factories

AhdraMeraliQB:
## Description
With the introduction of dataset factories to the data catalog (#2635) our catalog CLI commands need to be updated to reflect the changes. This PR updates `kedro catalog list` to include datasets that make use of the dataset factories.

## Development notes
I have tested this manually with a modified version of spaceflights. In the catalog, both entries for `preprocessed_shuttles` and `preprocessed_companies` have been replaced with the following:

```yaml
preprocessed_{name}:
  type: pandas.ParquetDataSet
  filepath: data/02_intermediate/preprocessed_{name}.pq
```

Before the changes made in this PR, this project setup would yield the following output using `kedro catalog list -p data_processing`:

```
Datasets in 'data_processing' pipeline:
  Datasets mentioned in pipeline:
    CSVDataSet:
    - companies
    - reviews
    DefaultDataset:
    - preprocessed_shuttles
    - preprocessed_companies
    ExcelDataSet:
    - shuttles
    ParquetDataSet:
    - model_input_table
  Datasets not mentioned in pipeline:
    PickleDataSet:
    - regressor
```

With the changes this now looks like:

```
Datasets in 'data_processing' pipeline:
  Datasets generated from factories:
    pandas.ParquetDataSet:
    - preprocessed_shuttles
    - preprocessed_companies
  Datasets mentioned in pipeline:
    CSVDataSet:
    - companies
    - reviews
    ExcelDataSet:
    - shuttles
    ParquetDataSet:
    - model_input_table
  Datasets not mentioned in pipeline:
    PickleDataSet:
    - regressor
```

### Questions for reviewers
This is more of a nit but I'm unsure "Datasets generated from factories" is the best way to describe this section, any other ideas/comments?

## Checklist


kedro/framework/cli/catalog.py
```diff
@@ -73,15 +77,25 @@ def list_datasets(metadata: ProjectMetadata, pipeline, env):
...
+        default_ds = default_ds - set(chain.from_iterable(factory_ds_by_type.values()))
+
         unused_by_type = _map_type_to_datasets(unused_ds, datasets_meta)
         used_by_type = _map_type_to_datasets(used_ds, datasets_meta)
 
         if default_ds:
             used_by_type["DefaultDataset"].extend(default_ds)
 
-        data = ((not_mentioned, dict(unused_by_type)), (mentioned, dict(used_by_type)))
+        data = ((mentioned, dict(used_by_type)), (factories, dict(factory_ds_by_type)), (not_mentioned, dict(unused_by_type)))
```
AhdraMeraliQB, line 96: It's worth noting that PyYaml will automatically sort what it dumps to terminal. To configure this we would need to update to PyYAML 5.1.

Currently, generated datasets will be listed first, then those mentioned in the pipeline, and lastly those that aren't. A similar alphabetical ordering is applied to the pipelines themselves (as opposed to in order of input). I'm not the biggest fan of this, but I'm unsure it's worth bumping  PyYAML for. In any case, I have re-ordered them, so that if/when we do update the PyYAML lower bound we can make the changes easily.

ankatiyar:
QQ: any reason we're distinguishing between datasets explicitly in the catalog vs generated from factories?

AhdraMeraliQB:
> QQ: any reason we're distinguishing between datasets explicitly in the catalog vs generated from factories?

@ankatiyar 

The idea spawned from a conversation with @merelcht - it is probably useful to distinguish these not just for clarity but also debugging purposes.

ankatiyar:
> The idea spawned from a conversation with @merelcht - it is probably useful to distinguish these not just for clarity but also debugging purposes.

This sounds good but I think then the heading for the "non factories" datasets could be changed to be consistent. Something like "Datasets generated from catalog files:"? What do you think'?

Also another thing, I was trying it out with my dummy project. But I noticed that this wasn't working for "Datasets not mentioned in pipeline" when these datasets are factories -
### When all the datasets are explicitly mentioned : 
```
Datasets in '__default__' pipeline:
  Datasets mentioned in pipeline:
    CSVDataSet:
    - france_companies
    - germany_companies
    - france_preprocessed_companies
    - germany_preprocessed_companies
    - switzerland_companies
    - switzerland_preprocessed_companies
    ParquetDataSet:
    - germany_final
    - switzerland_final
    - france_final
Datasets in 'pipe1' pipeline:
  Datasets mentioned in pipeline:
    CSVDataSet:
    - france_companies
    - germany_companies
    - france_preprocessed_companies
    - germany_preprocessed_companies
    - switzerland_companies
    - switzerland_preprocessed_companies
  Datasets not mentioned in pipeline:
    ParquetDataSet:
    - germany_final
    - switzerland_final
    - france_final

```
### vs When all datasets are factories
```
Datasets in '__default__' pipeline:
  Datasets generated from factories:
    pandas.CSVDataSet:
    - france_preprocessed_companies
    - switzerland_companies
    - germany_companies
    - germany_preprocessed_companies
    - france_companies
    - switzerland_preprocessed_companies
    pandas.ParquetDataSet:
    - switzerland_final
    - france_final
    - germany_final
Datasets in 'pipe1' pipeline:
  Datasets generated from factories:
    pandas.CSVDataSet:
    - france_preprocessed_companies
    - switzerland_companies
    - germany_companies
    - germany_preprocessed_companies
    - france_companies
    - switzerland_preprocessed_companies
```

AhdraMeraliQB:
@ankatiyar 

The "Datasets not mentioned in pipeline" looks at entries included in the catalog that aren't included in the pipeline. Because the factory datasets aren't included in the catalog, they wouldn't show up here - for them to be resolved they would have to be explicitly defined in the pipeline, which would then mean they show up in the "Datasets mentioned in pipeline" section instead.

> This sounds good but I think then the heading for the "non factories" datasets could be changed to be consistent. Something like "Datasets generated from catalog files:"? What do you think'?

I'm unsure the distinction here would be necessary as it is the default case, but it's worth considering 🤔

ankatiyar:
> The "Datasets not mentioned in pipeline" looks at entries included in the catalog that aren't included in the pipeline. Because the factory datasets aren't included in the catalog, they wouldn't show up here - for them to be resolved they would have to be explicitly defined in the pipeline, which would then mean they show up in the "Datasets mentioned in pipeline" section instead.

Could this be resolved like we do in `runner.py` -
https://github.com/kedro-org/kedro/blob/c1822a2be6d1649f858d38cf56ffc5da3889fbb6/kedro/runner/runner.py#L79

Like -> `catalog_ds` = Union(`set(catalog.list())`, `set(registered_ds)`)

AhdraMeraliQB:
@ankatiyar I see, you're suggesting any datasets that make use of factories within the default pipeline should be considered in the "Datasets not mentioned in pipeline" section?

To be completely honest, I feel like it might be overkill - to me "Datasets generated from factories" are generated by the pipeline therefore there would be no need to list the ones generated by a different pipeline, but it would be good to see if @merelcht / any one else have any thoughts on this.

RELEASE.md
```diff
@@ -15,6 +15,7 @@
 * Activated all built-in resolvers by default for `OmegaConfigLoader` except for `oc.env`.
 
 ## Bug fixes and other changes
+* Updated `kedro catalog list` to work with dataset factories.
```
merelcht, line 18: Nit: I would change this to 
```suggestion
* Updated `kedro catalog list` to show datasets generated with factories.
```

kedro/framework/cli/catalog.py
```diff
@@ -73,15 +77,30 @@ def list_datasets(metadata: ProjectMetadata, pipeline, env):
...
+        factory_ds_by_type = defaultdict(list)
+        for ds_name in default_ds:
+            matched_pattern = data_catalog._match_pattern(
+                data_catalog._dataset_patterns, ds_name
+            )
+            if matched_pattern:
+                ds_config = data_catalog._resolve_config(ds_name, matched_pattern)
+                factory_ds_by_type[ds_config["type"]].append(ds_name)
+
+        default_ds = default_ds - set(chain.from_iterable(factory_ds_by_type.values()))
```
merelcht, line 90: What does `chain.from_iterable()` do here?

tests/framework/cli/test_catalog.py
```diff
@@ -150,6 +162,47 @@ def test_default_dataset(
         assert yaml_dump_mock.call_count == 1
         assert yaml_dump_mock.call_args[0][0][key] == expected_dict[key]
 
+    def test_generated_dataset(
```
merelcht, line 165:
```suggestion
    def test_list_factory_generated_datasets(
```

merelcht:
> @ankatiyar I see, you're suggesting any datasets that make use of factories within the default pipeline should be considered in the "Datasets not mentioned in pipeline" section?
> 
> To be completely honest, I feel like it might be overkill - to me "Datasets generated from factories" are generated by the pipeline therefore there would be no need to list the ones generated by a different pipeline, but it would be good to see if @merelcht / any one else have any thoughts on this.

I don't really understand this. If a dataset isn't mentioned in the pipeline it wouldn't be picked up by a dataset factory, so how could a dataset generated by a factory every belong to the "non mentioned in pipeline" section?

merelcht (commented):
I've left a couple of suggestions and questions, but overall it looks good. I think "datasets generated from factories" is a good way to describe this section.

kedro/framework/cli/catalog.py
AhdraMeraliQB: Here `chain.from_iterable()` takes the collection of lists from `factory_ds_by_type.values()` (as the value type of this dict is lists) and flattens them into a single, flattened iterable.

SajidAlamQB (dismissed):
This looks great! 🌟 This change makes a lot of sense and improves the `kedro catalog list` to add visibility into datasets generated from factories well.

merelcht (approved):
Nice work 👍

SajidAlamQB (approved):

kedro/framework/cli/catalog.py
```diff
@@ -50,10 +51,15 @@ def list_datasets(metadata: ProjectMetadata, pipeline, env):
...
+    factories = "Datasets generated from factories"
 
     session = _create_session(metadata.package_name, env=env)
     context = session.load_context()
-    datasets_meta = context.catalog._data_sets  # noqa: protected-access
+
+    data_catalog = context.catalog
+    datasets_meta = data_catalog._data_sets
+    catalog_ds = set(data_catalog.list())
+    datasets_meta = context.catalog._data_sets
```
merelcht, line 62: In line 60 `datasets_meta` is defined as `data_catalog._data_sets` but then in line 62 it's immediately overwritten as `datasets_meta = context.catalog._data_sets`, is that on purpose? And what exactly do you mean with "meta" here?

AhdraMeraliQB: Thanks for catching this! I suspect this was introduced by a poor resolve on the merge conflicts, I've removed the redundant lines. 

The variable `datasets_meta` predates the changes in this PR, from my understanding using meta refers to the protected dictionary `catalog._data_sets` as opposed to the public _FrozenDataSets object `catalog.datasets`

merelcht: Ah! Thanks for clarifying!

merelcht (approved):
LGTM! ⭐

PR #2670, opened 2023-06-12
Add documentation for dataset factories feature

ankatiyar:
> **NOTE:** Kedro datasets are moving from `kedro.extras.datasets` to a separate `kedro-datasets` package in
> [`kedro-plugins` repository](https://github.com/kedro-org/kedro-plugins). Any changes to the dataset implementations
> should be done by opening a pull request in that repository.
## Description

Resolve #2666 
To be merged with/after https://github.com/kedro-org/kedro/pull/2635

## Development notes

Documentation for the dataset factories feature to the "The Data Catalog" page. 

## Checklist

- [ ] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [ ] Added tests to cover my changes

Issue #2791, opened 2023-07-12
Add `kedro catalog resolve` CLI command


AhdraMeraliQB:
## Description
Child of #2603

Issue #2723, opened 2023-06-23
Rectify "modular pipelines" terminology


astrojuanlu:
## Description
We're making various distinctions in our documentation about "Pipelines" and "Modular pipelines", for example in the TOC:

![image](https://github.com/kedro-org/kedro/assets/316517/07667e88-c6e1-449a-b88c-8b03a0b1f26d)

And in our wording:

> In many typical Kedro projects, a single (“main”) pipeline increases in complexity as the project evolves. To keep your project fit for purpose, we recommend that you create modular pipelines, which are logically isolated and can be reused.

> This wrapper really unlocks the power of modular pipelines.
> `from kedro.pipeline.modular_pipeline import pipeline`

To the point that I believed namespaces were the same as modular pipelines.

However, it turns out that Pipelines and Modular Pipelines are mostly the same thing, and that `kedro.pipeline.modular_pipeline.pipeline` is not a wrapper over `kedro.pipeline.pipeline`: they're the same function.

https://github.com/kedro-org/kedro/blob/160fd6bc66d4d780652dd77499abeb23e13f1c09/kedro/pipeline/__init__.py#L5

This is also related from this comment that I didn't fully understand back then: https://github.com/kedro-org/kedro/issues/2402#issuecomment-1460087177

## Context

It's a key concept for reusability that many users use.

## Possible Implementation
- Remove mentions of "modular pipelines", just talk about "pipelines", some of which are _registered_ (in `pipeline_registry.register_pipelines`) and some of which aren't.
- In the pages that currently talk about "modular pipelines", replace the wording by "namespaced pipelines" or just "namespaces".
- Remove mentions to `kedro.pipeline.modular_pipeline.pipeline` and just use `kedro.pipeline.pipeline` everywhere. (xref #712)

## Possible Alternatives
There are maybe less disruptive paths but I can't think of alternative ways of rectifying the current terminology.

astrojuanlu:
> **Maybe** ❓ Move `pipeline` from `modular_pipeline.py` to `pipeline.py` and delete `modular_pipeline.py`. This would break any imports `from kedro.pipeline.modular_pipeline import pipeline` but not `from kedro.pipeline import pipeline`

This!

stichbury:
[...]

Issue #2402, opened 2023-03-08
Merge `kedro pipeline` and `kedro registry`


astrojuanlu:
## Description
(Comes from [Slack](https://linen.dev/s/kedro/t/9712189/hi-folks-is-there-a-cli-command-to-show-the-available-pipeli#e0983ec7-211d-4cc9-9188-3c699deb2470))

Since `kedro pipeline` and `kedro registry` are both somewhat related to the pipeline registry, I propose we merge both to have

- `kedro pipeline list` (formerly in `kedro registry`)
- `kedro pipeline describe` (formerly in `kedro registry`)
- `kedro pipeline create`
- `kedro pipeline delete`

and therefore deprecate, and eventually remove, `kedro registry`.

@merelcht points out that "We’d need to clearly define what the difference is between a “registered” pipeline and just pipelines that exist, but aren’t part of any run command e.g. if the user has changed what’s run by default".

## Context
I was trying to get a list of available pipelines, which was available behind `kedro registry` but not `kedro pipeline`.

astrojuanlu:
I now understand that these two should not be _merged_. If anything, we could have a `kedro pipeline list` that lists _all_ the pipelines (even non-registered ones), but we should assess how useful this would be.

PR #2796, opened 2023-07-14
Add `kedro catalog factory list` CLI command

AhdraMeraliQB:
Ignore the target branch for now, #2793 to be merged first.

## Description
This PR introduces the command `kedro catalog factory list \<edited\> which lists out all dataset factories in the catalog config in order of how they are matched (i.e. if an explicitly named dataset could match several of the factories in the catalog, it would be resolved with the first on the list.)

The priority is determined as such:
1. Decreasing specificity (number of characters outside the curly brackets)
2. Decreasing number of placeholders (number of curly bracket pairs)
3. Alphabetically

### Question for reviewers
An alternative suggestion for the name of the command was `kedro catalog patterns`. I went with factories to attempt to streamline the terminology used in the docs and other communications, but am otherwise impartial. Any thoughts on this?

#### Added note on streamlining the terminology
It looks like `dataset factory` is equiv to `factory pattern`, `catalog factory`, `dataset pattern`, `dataset factory pattern` and others that have been used interchangeably. As per #2670 any mentions of a catalog entry that makes use of placeholders will be referred to as a `dataset factory`. `Catalog factories` refer to all dataset factories within a specific catalog.

## What this means in practice / Why this is useful
As seen in the test case, a catalog is created with the following patterns:

1. `an_example_{place}_{holder}`,
2.  `an_example_{placeholder}`,
3.  `an_{example_placeholder}`,
5.  `on_{example_placeholder}`,

The explicitly declared `an_example_data_set` could match any of patterns 1, 2, or 3: `an_example_{data}_{set}`, `an_example_{data_set}`, `an_{example_data_set}`. Priority matching means that it will be resolved with the first pattern.
`kedro catalog factories` allows users to confirm the order in which factory patterns are considered for matching.

## Checklist

### DIFF
--- a/kedro/framework/cli/catalog.py
+++ b/kedro/framework/cli/catalog.py
@@ -179,7 +179,7 @@ def create_catalog(metadata: ProjectMetadata, pipeline_name, env):
         click.echo("All datasets are already configured.")
 
 
-@catalog.command("factories")
+@catalog.command("factory list")
 @env_option
 @click.pass_obj
 def list_patterns(metadata: ProjectMetadata, env):
--- a/kedro/framework/cli/catalog.py
+++ b/kedro/framework/cli/catalog.py
@@ -193,7 +193,10 @@ def list_factories(metadata: ProjectMetadata, env):
     context = session.load_context()
 
     catalog_factories = context.catalog._dataset_patterns
-    click.echo(yaml.dump(list(catalog_factories.keys())))
+    if catalog_factories:
+        click.echo(yaml.dump(list(catalog_factories.keys())))
+    else:
+        click.echo("There are no dataset factories in the catalog.")
 
 
 def _add_missing_datasets_to_catalog(missing_ds, catalog_path):
--- a/kedro/framework/cli/catalog.py
+++ b/kedro/framework/cli/catalog.py
@@ -179,7 +179,11 @@ def create_catalog(metadata: ProjectMetadata, pipeline_name, env):
         click.echo("All datasets are already configured.")
 
 
-@catalog.command("factory list")
+@catalog.group()
+def factory():
+    """Command for working with catalog factories"""
+
+@factory.command("list")
 @env_option
 @click.pass_obj
 def list_patterns(metadata: ProjectMetadata, env):
--- a/kedro/framework/cli/catalog.py
+++ b/kedro/framework/cli/catalog.py
@@ -179,6 +179,18 @@ def create_catalog(metadata: ProjectMetadata, pipeline_name, env):
         click.echo("All datasets are already configured.")
 
 
+@catalog.command("factories")
+@env_option
+@click.pass_obj
+def list_patterns(metadata: ProjectMetadata, env):
+    "Show all factory patterns in catalog, ranked by priority by which they are matched"
+    session = _create_session(metadata.package_name, env=env)
+    context = session.load_context()
+
+    catalog_factories = context.catalog._dataset_patterns
+    click.echo(yaml.dump(list(catalog_factories.keys())))
+
+
 def _add_missing_datasets_to_catalog(missing_ds, catalog_path):
     if catalog_path.is_file():
         catalog_config = yaml.safe_load(catalog_path.read_text()) or {}
--- a/kedro/framework/cli/catalog.py
+++ b/kedro/framework/cli/catalog.py
@@ -181,13 +181,14 @@ def create_catalog(metadata: ProjectMetadata, pipeline_name, env):
 
 @catalog.group()
 def factory():
-    """Command for working with catalog factories"""
+    """Commands for working with catalog dataset factories"""
+
 
 @factory.command("list")
 @env_option
 @click.pass_obj
-def list_patterns(metadata: ProjectMetadata, env):
-    "Show all factory patterns in the catalog, ranked by priority by which they are matched."
+def list_factories(metadata: ProjectMetadata, env):
+    """Show all dataset factories in the catalog, ranked by priority by which they are matched."""
     session = _create_session(metadata.package_name, env=env)
     context = session.load_context()
 
--- a/kedro/framework/cli/catalog.py
+++ b/kedro/framework/cli/catalog.py
@@ -183,7 +183,7 @@ def create_catalog(metadata: ProjectMetadata, pipeline_name, env):
 @env_option
 @click.pass_obj
 def list_patterns(metadata: ProjectMetadata, env):
-    "Show all factory patterns in catalog, ranked by priority by which they are matched"
+    "Show all factory patterns in the catalog, ranked by priority by which they are matched."
     session = _create_session(metadata.package_name, env=env)
     context = session.load_context()
 
--- a/tests/framework/cli/test_catalog.py
+++ b/tests/framework/cli/test_catalog.py
@@ -419,3 +419,28 @@ def test_list_catalog_factories(
 
     assert yaml_dump_mock.call_count == 1
     assert yaml_dump_mock.call_args[0][0] == expected_patterns_sorted
+
+
+@pytest.mark.usefixtures(
+    "chdir_to_dummy_project",
+    "fake_load_context",
+)
+def test_list_factories_with_no_factories(
+    fake_project_cli, fake_metadata, fake_load_context
+):
+    mocked_context = fake_load_context.return_value
+
+    catalog_data_sets = {
+        "iris_data": CSVDataSet("test.csv"),
+        "intermediate": MemoryDataset(),
+        "not_used": CSVDataSet("test2.csv"),
+    }
+    mocked_context.catalog = DataCatalog(data_sets=catalog_data_sets)
+
+    result = CliRunner().invoke(
+        fake_project_cli, ["catalog", "factory", "list"], obj=fake_metadata
+    )
+
+    assert not result.exit_code
+    expected_output = "There are no dataset factories in the catalog."
+    assert expected_output in result.output
--- a/tests/framework/cli/test_catalog.py
+++ b/tests/framework/cli/test_catalog.py
@@ -406,7 +406,7 @@ def test_list_catalog_factories(
     )
 
     result = CliRunner().invoke(
-        fake_project_cli, ["catalog", "factory list"], obj=fake_metadata
+        fake_project_cli, ["catalog", "factory", "list"], obj=fake_metadata
     )
     assert not result.exit_code
 
--- a/tests/framework/cli/test_catalog.py
+++ b/tests/framework/cli/test_catalog.py
@@ -397,7 +397,9 @@ def test_list_catalog_factories(
 ):
     yaml_dump_mock = mocker.patch("yaml.dump", return_value="Result YAML")
     mocked_context = fake_load_context.return_value
-    mocked_context.catalog = DataCatalog.from_config(fake_catalog_with_overlapping_factories)
+    mocked_context.catalog = DataCatalog.from_config(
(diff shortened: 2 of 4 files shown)


ankatiyar:
Implementation wise, manually tested and it works as expected! 💯 
Suggestion on the name of the commands -
Could we do something like `kedro catalog factories --list` and `kedro catalog factories --resolve` for https://github.com/kedro-org/kedro/issues/2791? IMO `kedro catalog factories` does not explain what it does enough.
We could even do `kedro catalog list --factories` for this. Would like to hear what everyone else thinks?

merelcht:
> Implementation wise, manually tested and it works as expected! 💯 Suggestion on the name of the commands - Could we do something like `kedro catalog factories --list` and `kedro catalog factories --resolve` for #2791? IMO `kedro catalog factories` does not explain what it does enough. We could even do `kedro catalog list --factories` for this. Would like to hear what everyone else thinks?

I think this is an interesting point. I tend to agree that `kedro catalog factories` isn't very self explanatory. It would be good to get @amandakys's designer point of view. We need to ensure this is consistent with the other CLI commands, just like the exercise we did for `kedro run` and all the flags there.

amandakys:
I had a look at the existing 'list' style commands and we have:

`kedro catalog list` - list datasets per pipeline per type 
-  also has the optional argument `--pipelines` that allows you to further narrow down the list of datasets returned by pipeline name. 
`kedro registry list` - list all registered pipelines in your project 
we also have `kedro registry describe` which outputs all the nodes in a specified pipeline 
___
As a note here - it could be more intuitive to reduce the number of keywords by changing `kedro registry describe` to be `kedro registry list --pipelines=<pipeline-name>` or `kedro register list --nodes --pipelines=<pipeline-name>`. There is potential for discussion here on our use of arguments, when to use <pipeline-name> as a input vs as an argument to a `--pipeline` flag. If there was interest in this work i'd be happy to do an audit 
___

The convention appears to be putting the verb at the end of the command with additional parameters being used to narrow down the output. So initial thoughts would be that `kedro catalog factory list` is the most consistent. 

However, it does depend on whether there are future plans for providing different types of outputs to allow users to better understand their catalog i.e. returning a full 'expanded' catalog, or providing a dataset name and returning which pattern it matched. In this case we could consider the addition of another key word like `resolve` or adding arguments to the `list` command like `--datasets`

astrojuanlu:
xref https://github.com/kedro-org/kedro/issues/2402 and https://github.com/kedro-org/kedro/issues/2723

merelcht:
Thanks for the detailed analysis @amandakys ! I think reading this and keeping in mind that we have https://github.com/kedro-org/kedro/issues/2791 coming up I would suggest going for `kedro catalog factory list` and then the command for https://github.com/kedro-org/kedro/issues/2791 can be `kedro catalog factory resolve`.

kedro/framework/cli/catalog.py
```diff
@@ -179,6 +179,18 @@ def create_catalog(metadata: ProjectMetadata, pipeline_name, env):
         click.echo("All datasets are already configured.")
 
 
+@catalog.command("factories")
+@env_option
+@click.pass_obj
+def list_patterns(metadata: ProjectMetadata, env):
+    "Show all factory patterns in catalog, ranked by priority by which they are matched"
```
merelcht, line 186: Nit:
```suggestion
    "Show all factory patterns in the catalog, ranked by priority by which they are matched."
```

tests/framework/cli/test_catalog.py
```diff
@@ -360,3 +383,35 @@ def test_bad_env(self, fake_project_cli, fake_metadata):
 
         assert result.exit_code
         assert "Unable to instantiate Kedro session" in result.output
+
+
+@pytest.mark.usefixtures(
+    "chdir_to_dummy_project", "fake_load_context", "mock_pipelines"
+)
+def test_list_catalog_factories(
```
merelcht, line 391: I think it would also be useful to show where an exact match will show up (should be at the top).

merelcht: E.g. a catalog entry without any placeholders.

merelcht (commented):
This looks good! I left some small comments. When those are resolved + the name of the command this will be pretty much ready for merging 🙂

AhdraMeraliQB:
Coverted back to draft as currently only works with `kedro catalog "factory list"` and not `kedro catalog factory list`

kedro/framework/cli/catalog.py
```diff
@@ -179,6 +179,18 @@ def create_catalog(metadata: ProjectMetadata, pipeline_name, env):
         click.echo("All datasets are already configured.")
 
 
+@catalog.command("factory list")
+@env_option
+@click.pass_obj
+def list_patterns(metadata: ProjectMetadata, env):
+    "Show all factory patterns in the catalog, ranked by priority by which they are matched."
```
SajidAlamQB, line 186:
```suggestion
    """Show all factory patterns in the catalog, ranked by priority by which they are matched."""
```

kedro/framework/cli/catalog.py
```diff
@@ -179,6 +179,18 @@ def create_catalog(metadata: ProjectMetadata, pipeline_name, env):
...
+    session = _create_session(metadata.package_name, env=env)
+    context = session.load_context()
+
+    catalog_factories = context.catalog._dataset_patterns
+    click.echo(yaml.dump(list(catalog_factories.keys())))
```
SajidAlamQB, line 191:
```suggestion
    dataset_patterns = context.catalog._dataset_patterns
    click.echo(yaml.dump(list(dataset_patterns.keys())))
```
minor nitpick since we are using `_dataset_patterns` might be more fitting to call this `dataset_patterns`.

tests/framework/cli/test_catalog.py
```diff
@@ -360,3 +387,35 @@ def test_bad_env(self, fake_project_cli, fake_metadata):
...
```
SajidAlamQB, line 395: Should we add some failure test cases for example, testing empty or invalid DataCatalog configuration?

SajidAlamQB (commented):
I think the implementation looks good and concise the test are also well-structured and verifies the core functionality. I've just left some minor comments and nitpicks. I think it will be ready to merge once the the other reviewer comments are also addressed. Happy to approve once thats done. 😄

kedro/framework/cli/catalog.py
AhdraMeraliQB: I think there will still be some inconsistency, with the change we will still be dumping "dataset patterns" under the name catalog factories 🤔. How about `catalog_factory_patterns`? A tad verbose but more readable/understandable.

tests/framework/cli/test_catalog.py
AhdraMeraliQB: Good catch!

Following the test setup for `kedro catalog list` neither of these cases are tested for. Behaviour-wise an invalid catalog produces an error for both `kedro catalog list` and `kedro catalog factory list`. An empty catalog leaves all datasets in the pipeline to be described as a default dataset for `kedro catalog list`, and an empty set `[ ]` to be output by `kedro catalog factory list`. The same empty set is output for a valid catalog with no patterns.

 I think catching it and returning a "There are no dataset factory patterns in the catalog" message should suffice, and then adding a test case to ensure it's output where it should be.

AhdraMeraliQB:
Another question for reviewers - I've refactored `test_get_cli_structure_help` with two approaches, [iterative](https://github.com/kedro-org/kedro/pull/2796/commits/49bf326645b90b5668fa233f200fe580581d941c) and [recursive](https://github.com/kedro-org/kedro/pull/2796/commits/e7723e912ea2e190d6c41372806ff542fe5a86a5). I'm not particularly partial to either method; whilst iteration is faster and more memory-efficient, we are not handling any considerable amounts of data with this test. 

Does anyone have any opinions on this?

tests/framework/cli/test_catalog.py
AhdraMeraliQB: Added in [a8657db](https://github.com/kedro-org/kedro/pull/2796/commits/a8657db7e146de1e7b0a0667d13ba642d0ef30d8)

kedro/framework/cli/catalog.py
AhdraMeraliQB: Added clarification to the docstring, and some comments on consistency to the PR description. Let me know if you think the variables should still be updated!

kedro/framework/cli/catalog.py
AhdraMeraliQB: Changed in [4aa5bf8](https://github.com/kedro-org/kedro/pull/2796/commits/4aa5bf870d1737e1e9cd18bdfd6a90ef30913742)

AhdraMeraliQB:
Also to note - the code that is tested by `test_get_cli_structure_help` (and the tests themselves) are duplicated in `kedro-telemetry` which brings two things:
1. Any refactoring decided here should be applied there
2. Why do we have duplicate code? Can it be removed from Kedro framework?

tests/tools/test_cli.py
```diff
@@ -139,17 +139,18 @@ def test_get_cli_structure_help(self, mocker, fake_metadata):
...
-                    )
-            elif isinstance(v, str):
-                assert v.startswith("Usage:  [OPTIONS]")
+        self.recursive_cli_structure_test(help_cli_structure["kedro"])
 
         assert sorted(list(help_cli_structure["kedro"])) == sorted(
             DEFAULT_KEDRO_COMMANDS
         )
+
+    def recursive_cli_structure_test(self, structure):
```
merelcht, line 148: Nice one! 👍 Maybe just rename it to something like `recursively_check_cli_structure`

merelcht (approved):
Great work @AhdraMeraliQB ! ⭐ The recursive approach in the cli test is great 👍

kedro/framework/cli/catalog.py
SajidAlamQB: Happy to leave as is.

tests/tools/test_cli.py
AhdraMeraliQB: Renamed in [0b04fa4](https://github.com/kedro-org/kedro/pull/2796/commits/0b04fa477da7f6d62662c7b243a1bf61bbb26943)

noklam (approved):
Implementations look all good so I approve it.

IMO I am not a fan of `kedro catalog factory list`, this is hard to remember and I have never run any cli that is 4 levels deep.  While `data factory` is something we called, I really think this should be just an advance feature of `catalog`. Similarly we have `kedro pipeline` instead of `kedro modular_pipeline`.

astrojuanlu:
I agree with @noklam that `kedro catalog factory list` seemed unusually long for me as well. @amandakys point above about consistency is important though 🤔 @noklam do you have any alternatives in mind, taking into account the discussion above?

I agree we should try to limit bikeshedding

noklam:
In my mind I prefer just kedro catalog list and introduce new arguments - - resolve for the addition functions

Pro:
* consistent with the current command, less intrusive. 

Cons:
* May blurred things a bit but I don't think this is too important. 

If we are going to change these CLI, it's a breaking change and we cannot add it later. Introducing a new command is not breaking but it is more intrusive we introduce a new one and later reactor it into other commands.

AhdraMeraliQB:
After discussion with @merelcht, @noklam, @amandakys, @ankatiyar, and @astrojuanlu we have decided to implement this functionality under the command `kedro catalog rank`, and keep #2791 as `kedro catalog resolve`. As this affects several choices made on this branch, I will be closing this in favour of a new, clean PR.

PR #2568, opened 2023-05-08
Remove mentions to `.egg` and stop producing them from `kedro package`

astrojuanlu:
## Description
Fix gh-2273.

~~In addition, I fixed some low-hanging fruit for gh-2414, but found some complex issues along the way (gh-2542, gh-2567) so I didn't go any further. On the other hand, the `micropkg` workflow is somewhat self-contained and is not affected by how starters look, so it can be refactored separately.~~ Reverted, see below.

## Development notes
Added https://pypi.org/project/build/ as runtime dependency.

Even though _in principle_ we should be ready to move forward with gh-2350, just to avoid any issues I think it will be better to wait until https://github.com/kedro-org/kedro/milestone/36 is complete.

## Checklist

- [x] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [x] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [x] Updated the documentation to reflect the code changes
- [x] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [ ] Added tests to cover my changes

PR #2375, opened 2023-03-01
Use `fsspec` in `OmegaConfigLoader` to read config from compressed file

merelcht:
## Description
Solves part of: [Add support to read Kedro config from a zip/tar.gz file#2146](https://github.com/kedro-org/kedro/issues/2146)

This PR introduces the reading of config from `zip` and `tar.gz` files for the `OmegaConfigLoader`. One or more PRs will follow to add the functionality to the `ConfigLoader` and `TemplatedConfigLoader` as well. 

## Development notes 🏗

After a lot of trial and error it turned out that `fsspec` does allow reading both zip and tar.gz for all the config loaders. The only requirement is that the latest version of `fsspec` should be used in which they solved an issue with tar thread safety (https://filesystem-spec.readthedocs.io/en/latest/changelog.html#id1). 

**Notes on testing 🧪**
- I tested the functionality manually by creating a tar file using: `tar --exclude=local/*.yml -czf tar_conf.tar.gz --directory="/Users/merel_theisen/Projects/Testing/iris-omega" conf` where `iris-omega` is a Kedro project based on the pandas-iris starter. And then doing `kedro run --conf-source tar_conf.tar.gz` and similarly for zip by running: `zip -r zip_conf.zip conf` from my project and then `kedro run --conf-source zip_conf.zip`
- I intend to add end to end tests when all config loaders have this reading functionality and when [Package config as a tar.gz#2212](https://github.com/kedro-org/kedro/issues/2212) is also completed. 

## Checklist

- [ ] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [x] Updated the documentation to reflect the code changes
- [x] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [x] Added tests to cover my changes

PR #2816, opened 2023-07-18
Pin `pip<23.2` and skip one micropkg test to unblock CI from breaking changes

noklam:
## Description

Partly fix #2813 
## Development notes

- bump version of pytest-xdist
- change --numprocess 4 -> -n logical, let pytest-xdist to figure out how many cores can be used. See the documentations https://pytest-xdist.readthedocs.io/en/latest/distribution.html?highlight=numprocess#running-tests-across-multiple-cpus
- version from 2.2.1 -> 3.3.1 (not bound) - [Changelog](https://pytest-xdist.readthedocs.io/en/latest/changelog.html)
## Checklist

It could be `psutils` or `pytest-xdist`, root cause unclear.


RELEASE.md
```diff
@@ -16,6 +16,7 @@
 
 ## Bug fixes and other changes
 * Consolidated dependencies and optional dependencies in `pyproject.toml`.
+* Pin `pip<23.2` due to a breaking change. See https://github.com/kedro-org/kedro/pull/2813
```
astrojuanlu, line 19: Hopefully we will not have to release Kedro like this 😢 

Also, notice that we only changed the CI requirements, Kedro `pyproject.toml` is intact

noklam:
https://app.circleci.com/pipelines/github/kedro-org/kedro/24640/workflows/2ca3adec-dbbf-45f7-a9a3-0133ee7bc981/jobs/284236

Interesting, it seems that our window test didn't run this step "Install pip setuptools" at all.

astrojuanlu:
Weird, there's still a stack overflow on Windows. We'll have to look more closely.

noklam:
I suspect it is still pip issue but I can confirm later.

noklam:
Noted Window tests are failing in different way.

Python 3.7 fails with 1 specific test
> FAILED 
> tests/framework/cli/micropkg/test_micropkg_package.py::TestMicropkgPackageCommand::test_package_micropkg

Python >3.7 fails with StackOverflow and pytest just panic.

noklam:
image
On my laptop everything passed even with pip==23.2

noklam:
Good sign that win e2e test are passing, I bumped the pytest-xdist which is very old and only support pytest<=3.9

noklam:
only win_unit_test 3.7 is failing now. https://app.circleci.com/pipelines/github/kedro-org/kedro/24745/workflows/d07eff5f-f9a7-4537-b812-f47450a7f7ca/jobs/285418

I will try to make it run sequentially now.

noklam:
Temporarily for 3.7 tests run in sequential.

astrojuanlu:
Good insight @noklam. Does it happen if only the offending tests are run? e.g. doing `pytest -k micropkg`.

noklam:
I'll open a separate PR as I want to keep this CI job running.

noklam:
CI still failed - https://github.com/kedro-org/kedro/pull/2828

merelcht:
@noklam So it's just the windows unit tests on 3.7 failing? If you feel like it's sensible, should we just change this check to not be mandatory?

noklam:
Ok, looking into the details of the test and it's failing in Window only, it reminds me of an issue when we work on the `conf_source` feature it also create some bugs in Window. I wonder if [this](https://github.com/kedro-org/kedro/pull/2375#pullrequestreview-1324484262) is related. I remember the tar/zip file is acting weird on Window Py37

```
        with tarfile.open(sdist_file, "r") as tar:
            sdist_contents = set(tar.getnames())
```

noklam:
* Suppress few more test Win Python 3.7 test_micropkg_packge.py - `TestMicropkgPackageCommand::test_micropkg_package_same_name_as_package_name`
* Recover a passing test from last week - pin the exact version for python 3.7

AhdraMeraliQB:
> It may be related to these lines that file handler are not closed properly and blocked other process?
> 
> ```
>         with tarfile.open(sdist_file, "r") as tar:
>             sdist_contents = set(tar.getnames())
> ```

Just a note that the `with` context should be automatically closing the files after the lines are run

noklam:
The issue seems to be slightly random, Win 3.7 consistently fail, while the 3.8, 3.9 fails occasionally.

- This fails in 3.7, 3.8, 3.9 [cc3c6a9](https://github.com/kedro-org/kedro/pull/2816/commits/cc3c6a9ed3e97d089c90a7517c0f38a13244ef13)
- It was only failing in Win 3.7 [27fcd1a](https://github.com/kedro-org/kedro/pull/2816/commits/27fcd1a6029d3376efe93df5f02db7f6902ae00f)
- Now it fails for Win 3.8 too - [46a34e4](https://github.com/kedro-org/kedro/pull/2816/commits/46a34e40ad5ea7cbb249369a39f4792cb70e34fb)

AhdraMeraliQB:
Looks like all 3.8 and 3.9 fails are a result of the node crashing and the process timing out, not any particular tests failing. I would expect that on a re-run or with raising the time allowance they would pass more consistently. However this does not address the node crashing (looks like a seg fault) or 3.7's error

noklam:
> Just a note that the with context should be automatically closing the files after the lines are run

@AhdraMeraliQB This is probably true. I only say that because when I was working on reading the compressed config `.tar.gz`, I recall there was similar issue with Window. However that is because of an implementation issue with `fsspec`, in this case it is pure `tarfile` which is Python Standard Library and I don't expect it to be buggy.

AhdraMeraliQB:
Could be the case, after playing around, the file permissions conflict is arising from the lines 
```
result = CliRunner().invoke(
            fake_project_cli, ["pipeline", "create", pipeline_name], obj=fake_metadata
        )
```
My guess at the moment is that several tests are trying to `kedro pipeline create` the same modular pipeline simultaneously. Unsure why it has only started failing _now_, an intermediate fix could be to specify that these should not be run in parallel (essentially calling pytest twice, parallel for those unmarked, and sequentially for anything marked as such). It is puzzling that it's only just arisen 🤔

astrojuanlu:
Wait, that looks familiar: https://github.com/kedro-org/kedro/pull/2568#issuecomment-1539949117

This is my attempt at a diagnose then: https://github.com/kedro-org/kedro/pull/2568#issuecomment-1543546691

Essentially, it seemed like not all the tests were getting their own temporary directory, and under some circumstances this seems to fail. Maybe it's the same problem?

I was not able to fix it back then, I just sidestepped it or reverted the changes.

noklam:
I suspect is our custom fixture which creates problem, so I use pytest fixture but the problem remains. [51986d5](https://github.com/kedro-org/kedro/pull/2816/commits/51986d557f3c18c75ce18bd12c18288021a04cbd)

astrojuanlu:
If I understand correctly, there are two issues:

(1) Incompatibility between latest pip-tools and latest pip on Python 3.7. This is because pip-tools 7 dropped support for Python 3.7, and latest pip-tools 6.* is incompatible with latest pip. I managed to fix that issue with this diff:

```diff
diff --git a/features/environment.py b/features/environment.py
index 218ff097..671bee01 100644
--- a/features/environment.py
+++ b/features/environment.py
@@ -103,7 +103,9 @@ def _setup_minimal_env(context):
             "pip",
             "install",
             "-U",
-            "pip>=21.2",
+            # pip==23.2 breaks pip-tools<7.0, and pip-tools>=7.0 does not support Python 3.7
+            "pip>=21.2,<23.2; python_version < '3.8'",
+            "pip>=21.2; python_version >= '3.8'",
             "setuptools>=65.5.1",
             "wheel",
         ],
diff --git a/pyproject.toml b/pyproject.toml
index ae5fc961..50bf4a84 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ dependencies = [
     "more_itertools~=9.0",
     "omegaconf~=2.3",
     "parse~=1.19.0",
-    "pip-tools~=6.5",
+    "pip-tools>=6.5,<8",
     "pluggy~=1.0",
     "PyYAML>=4.2, <7.0",
     "rich>=12.0, <14.0",
```

(2) Problems with the micropackaging tests only on Windows and only when run in parallel, unknown origin. As @noklam said [above](https://github.com/kedro-org/kedro/pull/2816#issuecomment-1649905606), suppressing 1 test doesn't work, there seems to be a systematic issue with those.

Maybe it's related to this observation by @SajidAlamQB in #1328:

https://github.com/kedro-org/kedro/blob/48d94a6972bcd8f9e7d381479e9914279721a054/tests/framework/cli/conftest.py#L147-L155

> an intermediate fix could be to specify that these should not be run in parallel (essentially calling pytest twice, parallel for those unmarked, and sequentially for anything marked as such).

@AhdraMeraliQB I agree. Could be something like this:

```diff
diff --git a/.circleci/continue_config.yml b/.circleci/continue_config.yml
index 1632c4bd..bc7cf756 100644
--- a/.circleci/continue_config.yml
+++ b/.circleci/continue_config.yml
@@ -285,8 +285,11 @@ jobs:
             equal: [ "3.10", <<parameters.python_version>> ]
           steps:
             - run:
-                name: Run unit tests without spark in parallel
-                command: conda activate kedro_builder; make test-no-spark
+                name: Run unit tests without spark and without micropkg in parallel
+                command: conda activate kedro_builder; pytest --no-cov --ignore tests/extras/datasets/spark -k "not micropkg" --numprocesses 4 --dist loadfile
+            - run:
+                name: Run micropkg unit tests sequentially
+                command: conda activate kedro_builder; pytest --no-cov --ignore tests/extras/datasets/spark -k "micropkg" --dist no
       - when:
           condition:
             equal: [ "3.10", <<parameters.python_version>> ]
```

astrojuanlu:
In https://github.com/astrojuanlu/kedro/pull/1 I experimented with the two ideas above and I still got a test failure with the dreaded "The process cannot access the file because it is being used by another process". The funniest thing is that the tests were running sequentially, and this test was the first to run. On the other hand, after my changes in #2614, `kedro micropkg` is slower, and the CI times can run out of hand.

Fundamentally, the fact that we are testing the micropkg workflow almost exclusively through the CLI makes debugging more difficult:

- The actual errors are hidden, because we're calling `kedro micropkg` as a subprocess.
- The way we mock and create temporary directories on the fly contributes to this.

Addressing this requires a significant rewrite of the tests.

I don't have very good ideas left on how to proceed beyond marking these tests as `xfail` on Windows + Python 3.7 and move on. I honestly don't think there's anything wrong with the code, but rather with our tests.

noklam:
@astrojuanlu Agree with most of it, just want to note that it also cause this flaky stackoverflow issue (not show up everytime) - See https://app.circleci.com/pipelines/github/kedro-org/kedro/24923/workflows/99bc8992-4935-44bb-9aa4-75dfc2402e23/jobs/286915

In this PR the stackoverflow issue fails in 3.8 & 3.9 as well, I suspect it is more related to `pytest-xdist` base on what @AhdraMeraliQB found.  If it's only ""The process cannot access the file because it is being used by another process", then I agree with `xfail`, otherwise we may need `xskip`.

So on solution, I think there are 2 options
1. `xskip` on Window Python 3.7-3.9
2. `xfail` on Window Python 3.7 + change all the test to run sequentially.

On refactoring - I agree on this. https://github.com/kedro-org/kedro/pull/2816#issuecomment-1650148254 is an attempt to improve how we create and clean up tests. We should leverage `fixture` as much as we could to reduce the chance that it conflicts with other pytest plugin. From my experience, touching the `sys.modules` often create unexpected issues, we made a lot of effort to remove most of these tests recently.

noklam:
On the other hand, it may worth to enable Github Action on `kedro`, it would not take too much effort since we already have the more complicated setup in `kedro-plugins`. The reason for this is I believe there is no bug in the code or dependency itself, so one possibility is CircleCI did something funny behind the scene.

noklam:
Closed in favor of moving to Github Action which solves the problem. Still unsure what's the root cause, could be CircleCI infrastructure as we didn't make any code change.

Issue #1995, opened 2022-11-01
Bump s3fs version number


datajoely:
## Description
> Is your feature request related to a problem? A clear and concise description of what the problem is: "I'm always frustrated when ..."

Our current setup.py has the following restriction `S3FS = "s3fs>=0.3.0, <0.5"`

[s3fs has been on calendar versioning](https://s3fs.readthedocs.io/en/latest/changelog.html#version-2021-04-0) since April 2021 so we're quite out of date.

## Implementation
- Move to the same calendar versioning scheme we apply to related fsspec project.
- Fix tests using `moto` or `aiobotocore`

noklam:
See #2017 for more details. This may not be trivial as bumping `s3fs` immediately cause a lot of tests fails. I suspect this is a pure test issue (Kedro will still work, but the tests are not working because of `moto` or `aiobotocore`).

Estimate: 5/8

astrojuanlu:
xref https://github.com/kedro-org/kedro-plugins/issues/337, which has an open PR https://github.com/kedro-org/kedro-plugins/pull/348

PR #2851, opened 2023-07-27
Add Python 3.11 support to Kedro

SajidAlamQB:
## Description

This is a continuation of https://github.com/kedro-org/kedro/pull/2417 building on it since we moved from `circleci` to `github actions`.

This PR will aim to add support for Python 3.11 to kedro, try consolidate GitHub actions CI and make it work with 3.11.

Related to: https://github.com/kedro-org/kedro/discussions/2048 https://github.com/kedro-org/kedro/issues/2270 https://github.com/kedro-org/kedro/pull/2417

## Development notes

- Added MSBuild to PATH, github actions windows os doesn't come with visual studio which is needed to build `matplotlib` for `freetype`

-  Pinned `pytables` to 3.8.0 for python >= 3.8, `"tables~=3.8.0; platform_system == 'Windows' and python_version>='3.8'"` there were some issues with python 3.8 on windows by pinning to 3.8.0 it fixes this, see: https://github.com/PyTables/PyTables/issues/933#issuecomment-1555917593

- @noklam comment: Regarding the SPARK pin in `setup.py`, `SPARK = "pyspark>=2.2, <3.4`, Set a strict lower cap because 3.4 breaks a lot of things, it's very hard to fix it without changing implementation. We shouldn't do it in kedro, will leave this to kedro-datasets instead.

- The new `moto` version now allows for parallel test execution for Python 3.10.

## Checklist


astrojuanlu:
Is this a separate effort from #2417 ?

SajidAlamQB:
> Is this a separate effort from #2417 ?

Part of the same effort just trying some things here separately.

tests/extras/datasets/spark/test_spark_dataset.py
```diff
@@ -528,7 +539,7 @@ def test_versioning_existing_dataset(
     sys.platform.startswith("win"), reason="DBFS doesn't work on Windows"
 )
 class TestSparkDataSetVersionedDBFS:
-    def test_load_latest(  # noqa: too-many-arguments
+    def test_load_latest(  # pylint: disable=too-many-arguments
```
merelcht, line 542: Why are these `pylint: disable` again? I thought with the move to `ruff` they were all changed to `noqa`?

.github/workflows/unit-tests.yml
```diff
@@ -45,7 +48,7 @@ jobs:
       - name: Run unit tests
         if: inputs.os == 'ubuntu-latest' && inputs.python-version != '3.10'
         run: make test
-      - name: Run unit tests without spark sequentially (Windows)
+      - name: Run unit tests without spark sequentially
```
merelcht, line 51: Might still be useful to have `(Windows)` in the name here, or does that not show on the build?

setup.py
```diff
@@ -162,23 +163,26 @@ def _collect_requirements(requires):
     "jupyter~=1.0",
     "lxml~=4.6",
     "matplotlib>=3.0.3, <3.4; python_version < '3.10'",  # 3.4.0 breaks holoviews
-    "matplotlib>=3.5, <3.6; python_version == '3.10'",
+    "matplotlib>=3.5, <3.6; python_version >= '3.10'",
     "memory_profiler>=0.50.0, <1.0",
     "moto==1.3.7; python_version < '3.10'",
     "moto==3.0.4; python_version == '3.10'",
+    "moto==4.1.12; python_version == '3.11'",
```
merelcht, line 170: Can we also use this `moto` version for Python 3.10? It would be better if we don't have a so many separate pins for the versions..

tests/extras/datasets/spark/test_spark_dataset.py
SajidAlamQB: Good catch, some of these files I copied over from previous PR attempting Python 3.11, https://github.com/kedro-org/kedro/pull/2417. This PR was a continuation of that it seems I copied some outdated files.

SajidAlamQB: Fixed

setup.py
```diff
@@ -143,14 +143,15 @@ def _collect_requirements(requires):
     "compress-pickle[lz4]~=2.1.0",
     "coverage[toml]",
     "dask[complete]~=2021.10",  # pinned by Snyk to avoid a vulnerability
-    "delta-spark~=1.2.1",  # 1.2.0 has a bug that breaks some of our tests: https://github.com/delta-io/delta/issues/1070
+    "delta-spark>=1.2.1; python_version >= '3.11'",  # 1.2.0 has a bug that breaks some of our tests: https://github.com/delta-io/delta/issues/1070
+    "delta-spark~=1.2.1; python_version < '3.11'",
```
merelcht, line 147: Doesn't `delta-spark>=1.2.1` mean the same as `delta-spark~=1.2.1`? https://peps.python.org/pep-0440/#compatible-release

SajidAlamQB: From what I understand `delta-spark>=1.2.1` will allow `python 3.11` to use any future version of delta-spark even outside the 1.2.x series such as 1.3.x etc. While `delta-spark~=1.2.1` will limit all other python versions to only anything in the 1.2.x series starting from `1.2.1`.

setup.py
```diff
@@ -162,23 +163,25 @@ def _collect_requirements(requires):
...
+    "pandas-gbq>=0.12.0, <0.18.0; python_version < '3.11'",
+    "pandas-gbq>=0.18.0; python_version >= '3.11'",
     "pandas~=1.3  # 1.3 for read_xml/to_xml",
     "Pillow~=9.0",
     "plotly>=4.8.0, <6.0",
     "pre-commit>=2.9.2, <3.0",  # The hook `mypy` requires pre-commit version 2.9.2.
-    "psutil~=5.8",
-    "pyarrow>=6.0",
+    "pyarrow>=1.0; python_version <= '3.10'",
+    "pyarrow>=7.0; python_version > '3.10'",
```
noklam, line 180: I forgot what's the reason that requires pyarrow >=7.0, can we add a comment here?

In addition, a styling problem. We use two pattern inconsitently, will be good to keep one.
```
xxx <= "3.10"
xxx > "3.10"
```

or 
```
xxx < "3.11"
xxx >= "3.11"
```

Actually, this is a bug if I didn't mess up.  <=3.10 mean anything before 3.10 which doesn't include 3.10.x. I think it's better to stick with pattern 2 here.

setup.py
```diff
@@ -187,12 +190,14 @@ def _collect_requirements(requires):
...
+    "scikit-learn>=1.0.2,<2",
+    "scipy>=1.7.3",
+    "semver",
     "SQLAlchemy~=1.2",
-    "tables~=3.6.0; platform_system == 'Windows' and python_version<'3.9'",
+    "tables~=3.6.0; platform_system == 'Windows' and python_version<'3.8'",
+    "tables~=3.8.0; platform_system == 'Windows' and python_version>='3.8'",  # Import issues with python 3.8 with pytables pinning to 3.8.0 fixes this https://github.com/PyTables/PyTables/issues/933#issuecomment-1555917593
     "tables~=3.6; platform_system != 'Windows'",
-    "tensorflow~=2.0; platform_system != 'Darwin' or platform_machine != 'arm64'",
+    "tensorflow>=2.0; platform_system != 'Darwin' or platform_machine != 'arm64'",
```
noklam, line 200: Does this makes any difference? Tensorflow is still on 2.x series why do we change from ~=2.0 -> >=2.0?

setup.py
SajidAlamQB: I think it was to avoid the numpy build errors, pinning pyarrow >=7.0 seems to have resolved it.

noklam (approved):
LGTM!

astrojuanlu:
OMG it's happening 🥲

tests/test_import.py
```diff
@@ -16,7 +16,7 @@ def test_import_kedro_with_no_official_support_raise_error(mocker):
 
 def test_import_kedro_with_no_official_support_emits_warning(mocker):
     """Test importing kedro python>=3.11 and controlled warnings should work"""
```
noklam, line 18:
```suggestion
    """Test importing kedro python>=3.12 and controlled warnings should work"""
```

tests/test_import.py
```diff
@@ -5,7 +5,7 @@
 
 def test_import_kedro_with_no_official_support_raise_error(mocker):
     """Test importing kedro with python>=3.11 should fail"""
```
noklam, line 7:
```suggestion
    """Test importing kedro with python>=3.12 should fail"""
```

noklam (approved):

.github/workflows/unit-tests.yml
```diff
@@ -30,6 +30,9 @@ jobs:
         with:
           path: ~\AppData\Local\pip\Cache
           key: ${{inputs.os}}-python-${{inputs.python-version}}
+      - name: Add MSBuild to PATH
+        if: inputs.os == 'windows-latest'
+        uses: microsoft/setup-msbuild@v1
       - name: Install dependencies
         run: |
             make install-test-requirements
```
noklam, line 38: I am surprise you don't need to add 3.11 here. Does that means test can be run in parallel for 3.11? Or is it because the `moto` version is now upgrade so we don't need to handle 3.10 specially anymore?

SajidAlamQB: The new `moto` version now allows for parallel test execution for Python 3.10. I believe before when we [added 3.10 support](https://github.com/kedro-org/kedro/pull/1275#:~:text=moto%20had%20to,run%20in%20parallel.) `moto` newer version broke parallel tests for 3.10 hence it was made sequential but I believe with the newer `moto` version we have pinned this has resolved the issue.

merelcht: That's amazing! 🤩

setup.py
merelcht: The `~` syntax still confuses me, but what you're saying sounds right.

merelcht (approved):
🥳 This should definitely be added to the release notes! It might also be useful to add any "migration" tips if there's any specific things that are different with python 3.11. Even though they're not Kedro specific we could share our own learning.

.github/workflows/unit-tests.yml
noklam: Wow! Good that I asked. This is great, I have look into the s3fs repository and basically they have to use the server mode for `moto` to run tests. The newer version of `moto` seems supporting this better and potentially we can unblock this ticket.

- https://github.com/kedro-org/kedro/issues/1995

PR #2854, opened 2023-07-28
Remove non-user-facing release notes about pip pin

deepyaman:
## Description

Super minor, but checking the change, I don't think there's anything user-facing here? As in, it's a change on the CI.

## Checklist


merelcht:
I think it would be useful to have it in the release notes for our own reference, so we know exactly when it was introduced. But you're right that it's not user facing, so perhaps it shouldn't be there.. I'm fine with either.

astrojuanlu (approved):
I'm leaning more towards keeping the changelog for user-facing changes, so I'm +1 on this PR. We can always do a `git blame` for internal stuff.

deepyaman:
> I think it would be useful to have it in the release notes for our own reference, so we know exactly when it was introduced. But you're right that it's not user facing, so perhaps it shouldn't be there.. I'm fine with either.

@merelcht I'm aligned with what @astrojuanlu said; since users read the release notes (hopefully), we should avoid things that are dev-only/won't necessarily mean anything to them, and use other tools for tracking those changes.

If you're fine either way, do you mind approving? :)

merelcht (approved):

Issue #2624, opened 2023-05-31
Document the use of custom resolvers with `OmegaConfigLoader`


ankatiyar:
## Description
To be done after #2622. Add a Kedro specific use case for custom resolvers with `OmegaConfigLoader` in the docs.
This will require some creativity in trying out ways of using custom resolvers in a Kedro project that's useful for our users rather than just addition which they show in the omegaconf docs. 

We also need to clarify what "custom resolvers" are in our own words. 

## Context
Follow up task from the tech design discussion https://github.com/kedro-org/kedro/issues/2407 (24/05)

astrojuanlu:
To note: we have some docs on how to register custom resolvers, but using a custom class, which is probably not necessary

https://github.com/kedro-org/kedro/blob/4a4bd7396668887309dd2352e72ce5b822c24b48/docs/source/configuration/advanced_configuration.md?plain=1#L243-L277

Also, rather than a simple `add` resolver, we could use my Polars use case for instance.

ankatiyar:
Also, mention `replace` is set to `True` in the docs when new resolvers are registered.

noklam:
Cross-post Slack answer

> 
> Ankita Katiyar
>   [25 minutes ago](https://kedro-org.slack.com/archives/C03RKP2LW64/p1691068792855539?thread_ts=1690997737.947639&cid=C03RKP2LW64)
> [@Nok Lam Chan](https://kedro-org.slack.com/team/U03RJ2PH79D)
>  is right - after the next release (or if you use the main branch of the source code), you would be able to set custom resolvers through CONFIG_LOADER_ARGS in your settings.py like this :
> def mydate():
>     return whatever
> 
> CONFIG_LOADER_ARGS = {
>     "custom_resolvers": {
>         "date_resolver": lambda: mydate(),
>     }
> }
> And then in your parameters.yml
> my_date: "${date_resolver:}"
> ---
> Right now you can also do in your settings.py
> def mydate():
>     return whatever
> 
> from omegaconf import OmegaConf
> if not OmegaConf.has_resolver("date_resolver"):
>     OmegaConf.register_new_resolver("date_resolver, lamda: mydate())

PR #2642, opened 2023-06-05
Decouple `logging` from ConfigLoader

noklam:
> **NOTE:** Kedro datasets are moving from `kedro.extras.datasets` to a separate `kedro-datasets` package in
> [`kedro-plugins` repository](https://github.com/kedro-org/kedro-plugins). Any changes to the dataset implementations
> should be done by opening a pull request in that repository.
## Description

Fix #2426

related:
- #2644 This has to be done later as we cannot change starters now.
> **Note**. Breaking, goes in 0.19.0.

Following https://github.com/kedro-org/kedro/issues/2281 we have decided that users will no longer be able to change logging configuration dependent on kedro run environment. This is a huge simplification and means that logging only gets configured _once_ in the kedro process with `LOGGING = _ProjectLogging()`.

There's a few things to be done here, but they all need to be done at the same time or it will be very confusing.

- [x] Remove `KedroSession._setup_logging` and `KedroSession._get_logging_config` 🥳 
- [x] Remove everything to do with logging from all config loaders (e.g. in `config_patterns` and docstring examples)
- [x] Double check you've removed everything that needs removing by searching `logging` throughout the whole codebase. Note that `configure_logging` must _stay_ since it's needed by the parallel runner
- [x] Make sure you explain in the release notes migration guide that a custom logging file is now independent of kedro run environment and only used if `KEDRO_LOGGING_CONFIG` is set to point to it
- [ ] Related doc changes and example of `logging.yml` for advance users who may want to customise logging.

## Checklist

- [ ] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [ ] Added tests to cover my changes

PR #2869, opened 2023-07-31
Allow registering of custom resolvers to `OmegaConfigLoader`

### LINKED ISSUE #2622: Allow registering of custom resolvers to OmegaConfigLoader through `CONFIG_LOADER_ARGS`
## Description
Allowing users to register custom resolvers will make it easy for users to do more advanced interpolation in their configs.
OmegaConf documentation about custom resolvers - https://omegaconf.readthedocs.io/en/2.3_branch/custom_resolvers.html

## Context
Follow up task for #2407

## Possible Implementation
Users should be able to add custom resolvers in their `settings.py` files -
```python
from kedro.config import OmegaConfigLoader


CONFIG_LOADER_CLASS = OmegaConfigLoader

CONFIG_LOADER_ARGS = {
    "custom_resolvers": {
        "add": lambda *x: sum(x),
    }
}
```
The config files then could look like this -
`parameters.yml`
```yaml
model_options:
  test_size: ${add:1,2,3}
  random_state: 3
```

* The implementation of this on the `OmegaConfigLoader` side should be a very thin wrapper for `OmegaConf.register_new_resolver()` so that any future changes to `OmegaConf` don't affect anything on Kedro side.
* Users can overwrite built-in resolvers using `replace=True`

ankatiyar:
## Description

Resolve #2622 

## Development notes

* Add a new parameter `custom_resolvers` and a new function `_register_new_resolvers` to `OmegaConfigLoader`
`custom_resolvers` is a dict of resolver name -> resolver and can be passed through the project's `settings.py`
```python
from kedro.config import OmegaConfigLoader
CONFIG_LOADER_CLASS = OmegaConfigLoader

CONFIG_LOADER_ARGS = {
    "custom_resolvers": {
        "add_2": lambda x: x+2,
        "plus_10": lambda y: y+10,
    }
}
```   

`oc.env` can be activated like this -
```python
from kedro.config import OmegaConfigLoader
CONFIG_LOADER_CLASS = OmegaConfigLoader

from omegaconf.resolvers import oc

CONFIG_LOADER_ARGS = {
    "custom_resolvers": {
        "oc.env": oc.env,
    }
}
```   
* The function `OmegaConfigLoader._register_new_resolvers()` uses the `omegaconf` function `register_new_resolver()` under the hood. The `replace` parameter is set to `True` so that builtin resolvers can be overwritten.
* Added a test for custom resolvers to `test_omegaconf_config.py`
* Documentation ticket to follow separately - #2624  
* TODO - add this to release notes but holding off till Kedro 0.18.12 release.
## Checklist


kedro/config/omegaconf_config.py
```diff
@@ -302,6 +307,12 @@ def _is_valid_config_path(self, path):
             ".json",
         ]
 
+    @staticmethod
+    def _register_new_resolvers(resolvers: dict[str, Callable]):
+        """Register custom resolvers"""
+        for name, resolver in resolvers.items():
+            OmegaConf.register_new_resolver(name, resolver, replace=True)
+
```
noklam, line 315: Does it generates any log when a resolve is replaced?

noklam: Why do we want to override built-in resolver? Maybe I am missing some context in the issue.

tests/config/test_omegaconf_config.py
```diff
@@ -649,3 +649,21 @@ def test_variable_interpolation_in_catalog_with_separate_templates_file(
...
+        }
+        _write_yaml(base_params, param_config)
+        custom_resolvers = {
+            "add": lambda *x: sum(x),
+            "plus_2": lambda x: x + 2,
+        }
+        conf = OmegaConfigLoader(str(tmp_path), custom_resolvers=custom_resolvers)
+        conf.default_run_env = ""
+        assert conf["parameters"]["model_options"]["test_size"] == 7
+        assert conf["parameters"]["model_options"]["random_state"] == 3
```
noklam, line 669: if `replace=True` is needed we should have a test for overriding resolver.

tests/config/test_omegaconf_config.py
```diff
@@ -649,3 +649,21 @@ def test_variable_interpolation_in_catalog_with_separate_templates_file(
...
+                "test_size": "${add: 3, 4}",
+                "random_state": "${plus_2: 1}",
+            }
+        }
+        _write_yaml(base_params, param_config)
+        custom_resolvers = {
+            "add": lambda *x: sum(x),
+            "plus_2": lambda x: x + 2,
+        }
+        conf = OmegaConfigLoader(str(tmp_path), custom_resolvers=custom_resolvers)
```
noklam, line 666: is `str(tmp_path)` strictly needed?

kedro/config/omegaconf_config.py
```diff
@@ -97,6 +98,7 @@ def __init__(  # noqa: too-many-arguments
                 the configuration paths.
             default_run_env: Name of the default run environment. Defaults to `"local"`.
                 Can be overridden by supplying the `env` argument.
+            custom_resolvers:
```
noklam, line 101: missing some docstring here? Maybe a link to OmegaConf docs

noklam (commented):
Leave some comments, thanks for the PR the description is clear and help a lot for review :)

kedro/config/omegaconf_config.py
ankatiyar: I think we decided in the tech design to have this flag set to true so users have the option to overwrite the built-in resolvers.
No logs generated, I think it might show up twice per resolver because config loader is initialised twice during one run (once when logging is set up and once again when context is created)

merelcht (commented):
I think this generally looks good. I agree with @noklam on having a test to check the overwriting works. We should also mention clearly in the docs that `replace=True` is set when the register new resolvers is called.

tests/config/test_omegaconf_config.py
ankatiyar: Added a test

tests/config/test_omegaconf_config.py
ankatiyar: I think so, `conf_source` is required argument

kedro/config/omegaconf_config.py
```diff
@@ -302,6 +308,12 @@ def _is_valid_config_path(self, path):
             ".json",
         ]
 
+    @staticmethod
+    def _register_new_resolvers(resolvers: dict[str, Callable]):
+        """Register custom resolvers"""
+        for name, resolver in resolvers.items():
+            OmegaConf.register_new_resolver(name, resolver, replace=True)
```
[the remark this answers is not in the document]
ankatiyar: Done

ankatiyar:
I've added a note to #2624 to document the fact that `replace` flag is set to True

tests/config/test_omegaconf_config.py
```diff
@@ -649,3 +649,40 @@ def test_variable_interpolation_in_catalog_with_separate_templates_file(
...
+            }
+        }
+        _write_yaml(base_params, param_config)
+        custom_resolvers = {
+            "custom": lambda x: x + 20,
+        }
+        conf = OmegaConfigLoader(str(tmp_path), custom_resolvers=custom_resolvers)
+        conf.default_run_env = ""
+        # test_size should be calculated using overwritten custom resolver (x + 20)
+        assert conf["parameters"]["model_options"]["test_size"] == 30
```
merelcht, line 688: Is it possible to add a check/assert first to show that "test_size" is set to 20 and then after the overwriting it will be 30? Maybe just by calling omegaconf directly on the `param_config`?

ankatiyar: Updated the test

merelcht (approved):
LGTM! 😄

tests/config/test_omegaconf_config.py
noklam: Sorry, I mean if `tmp_path` alone works? I check the definition of `AbstractConfigLoader` have a signature of `conf_source: str`, but I expect it should be `conf_source : str | Path`.
image

A quick experiment suggests that it should work

image

cc @merelcht

kedro/config/omegaconf_config.py
noklam: I think it would be nice to produce some log here. Either via OmegaConf builtin logging (if any) or we should have log when it is overwritten.

I think it should be an INFO level log because normally you won't overwrite a built-in resolver. Is there any risk if one want to overwrite built-in?

However, if it is producing duplicate logs, maybe it should be a DEBUG level.  

Note that Logging no longer need config_loader, see https://github.com/kedro-org/kedro/pull/2642 This is merged into `develop` only (0.19)

ankatiyar:
@merelcht, I've turned off "replace=True" after discussion with @noklam. The only use case for that would be overwriting the built-in resolvers (oc.create, oc.decode, oc.deprecate, oc.select, oc.dict) which probably not a lot of people would want to do. To switch on `oc.env` does not require it being replaced since we clear it entirely at initialisation of `OmegaConfigLoader`. `OmegaConf` also does not allow replacing resolvers by default. Either way, it doesn't make too much of a difference to the implementation. What do you think?

Also updated the test to check if registering `oc.env` works properly.

merelcht (approved):
Don't forget to update the release notes! Otherwise, this looks all good from my side 👍

ankatiyar:
For advanced users that want to overwrite the built-in resolvers, they can always do so directly in `settings.py` -
```python
from omegaconf import OmegaConf
OmegaConf.register_new_resolver(name, resolver, replace=True)
```

noklam:
> > @merelcht, I've turned off "replace=True" after discussion with @noklam. The only use case for that would be overwriting the built-in resolvers (oc.create, oc.decode, oc.deprecate, oc.select, oc.dict) which probably not a lot of people would want to do. To switch on `oc.env` does not require it being replaced since we clear it entirely at initialisation of `OmegaConfigLoader`. `OmegaConf` also does not allow replacing resolvers by default. Either way, it doesn't make too much of a difference to the implementation. What do you think?
> 
> I think that's a fair point. If it turns out that users want to be able to replace resolvers we could potentially add another argument to `CONFIG_LOADER_ARGS` where they can overwrite it, e.g. `replace_resolvers`. Thanks for discussing this in more detail. I'm happy with this approach 🙂 
>  
> > Also updated the test to check if registering `oc.env` works properly.
> 
> 👍 
> 
> 

@merelcht We also discussed this point. Adding a new replace argument isn't hard, but it  clutters the already long argument list.

Two reasons why I think we shouldn't do it now.
1. If we want to add replace argument, arguably we may need to add a few more since omegaconf allows few extra options. If so should we provide a omegaconf_kwargs instead of just one argument? It's not clear for now.
2. We can always add a new one, and it wouldn't be a breaking change. However if we decide to add replace as an arguments, but later we want to change it to kwargs, it will be a breaking change and we cannot do it easily.

So for now I think we have a good workaround for advance use case which won't be blocking. If there are more evidence coming in later we can then discuss more about 1. and have a hollistic view of what other arguments we may need to support.

noklam:
I also open a new issue to fix the tests and signature in one go. Will refine the description later.

kedro/config/omegaconf_config.py
```diff
@@ -302,6 +308,13 @@ def _is_valid_config_path(self, path):
             ".json",
         ]
 
+    @staticmethod
+    def _register_new_resolvers(resolvers: dict[str, Callable]):
+        """Register custom resolvers"""
+        for name, resolver in resolvers.items():
+            if not OmegaConf.has_resolver("name"):
+                OmegaConf.register_new_resolver(name=name, resolver=resolver)
```
noklam, line 316: Do you still think it makes sense to put a debug log here?

ankatiyar: I added it now!

noklam (approved):
⭐️

PR #2881, opened 2023-08-01
Consolidate two `ruff-pre-commit` entries into one

deepyaman:
## Checklist


tests/pyproject.toml
```diff
@@ -0,0 +1,8 @@
+[tool.ruff]
```
astrojuanlu, line 1: Is there a way to avoid creating another `pyproject.toml` file? This makes `tests` pip-installable

noklam: I don't like the idea of another `pyproject.toml` , does ruff automatically refers to more local scope of `pyproject.toml`? 

I think the better solution of this is using [tool.ruff.per-file-ignores] - See https://beta.ruff.rs/docs/settings/#per-file-ignores
```
[tool.ruff.per-file-ignores]
# Allow these outside of the package
"{tests,docs,tools}/*" = [
rules...
]
```

deepyaman: > I don't like the idea of another `pyproject.toml` , does ruff automatically refers to more local scope of `pyproject.toml`?

Yes, it does; see https://beta.ruff.rs/docs/configuration/#pyprojecttoml-discovery.

> I think the better solution of this is using [tool.ruff.per-file-ignores] - See https://beta.ruff.rs/docs/settings/#per-file-ignores
> 
> ```
> [tool.ruff.per-file-ignores]
> # Allow these outside of the package
> "{tests,docs,tools}/*" = [
> rules...
> ]
> ```

Seems reasonable. I didn't know about this syntax (not sure where you found it? :) I eventually found it in https://github.com/ing-bank/popmon/blob/v1.4.6/pyproject.toml#L131-L133, after searching very specifically, since I didn't find it in the Ruff docs).

I'll make the change.

noklam: It was actually pointed out by a contributor, the comment was in the PR.

noklam (approved):

tests/pyproject.toml
deepyaman: I see you got to fixing it before I did; thanks!

astrojuanlu (approved):
> Ruff's pre-commit hook should be placed after other formatting tools, such as Black and isort, unless you enable autofix, in which case, Ruff's pre-commit hook should run before Black, isort, and other formatting tools, as Ruff's autofix behavior can output code changes that require reformatting.

https://beta.ruff.rs/docs/usage/#pre-commit

LGTM!

PR #2891, opened 2023-08-03
Add kedro catalog resolve command

### LINKED ISSUE #2791: Add `kedro catalog resolve` CLI command

AhdraMeraliQB:
## Description

The `kedro catalog resolve` command resolves pipeline datasets against the dataset factories to get explicit configuration for these datasets, which is then collected and output with the explicit dataset entries in the data catalog. It was agreed that the first version of this command should not introduce options for filtering this output yet, and as such it resolves the catalog against the default project pipeline. Resolves #2791

## Development notes

To test these changes, navigate to a local kedro project and run `kedro catalog resolve`. For maximum demonstration you should have some dataset factories in your catalog with explicit entries in the project pipeline. 

I have used the spaceflights starter, replacing the preprocess_companies and preprocess_shuttles datasets in the catalog with:

```yaml
preprocessed_{name}:
  type: pandas.ParquetDataSet
  filepath: data/02_intermediate/preprocessed_{name}.pq
```

The expected output should be as follows:

```yaml
companies:
  filepath: data/01_raw/companies.csv
  type: pandas.CSVDataSet
model_input_table:
  filepath: data/03_primary/model_input_table.pq
  type: pandas.ParquetDataSet
preprocessed_companies:
  filepath: data/02_intermediate/preprocessed_companies.pq
  type: pandas.ParquetDataSet
preprocessed_shuttles:
  filepath: data/02_intermediate/preprocessed_shuttles.pq
  type: pandas.ParquetDataSet
regressor:
  filepath: data/06_models/regressor.pickle
  type: pickle.PickleDataSet
  versioned: true
reviews:
  filepath: data/01_raw/reviews.csv
  type: pandas.CSVDataSet
shuttles:
  load_args:
    engine: openpyxl
  type: pandas.ExcelDataSet
```

## Checklist


kedro/framework/cli/catalog.py
```diff
@@ -208,3 +208,72 @@ def rank_catalog_factories(metadata: ProjectMetadata, env):
...
+        pl_obj = pipelines.get(pipe)
+        if pl_obj:
+            pipeline_ds = pl_obj.data_sets()
+        else:
+            existing_pls = ", ".join(sorted(pipelines.keys()))
+            raise KedroCliError(
+                f"'{pipe}' pipeline not found! Existing pipelines: {existing_pls}"
+            )
+
+        for ds_name in pipeline_ds:
```
AhdraMeraliQB, line 250: This is the second time this kind of loop through a set to find patterns is seen in this file (with the potential for a third given `kedro catalog create` needs to be updated for dataset factories as well) - I propose refactoring it out into a helper method, with the note that the surrounding implementation may have to be modified slightly. Any thoughts on this?

kedro/io/data_catalog.py
```diff
@@ -148,6 +148,7 @@ def __init__(  # noqa: too-many-arguments
         dataset_patterns: Patterns = None,
         load_versions: dict[str, str] = None,
         save_version: str = None,
+        raw_config: dict[str, Any] = None,
```
ankatiyar, line 151: I feel like there might be a way to do the resolving without saving the raw config here. We could use the config loader to get the raw catalog?

ankatiyar: + the raw config in `DataCatalog` also has params which might need to be filtered out

kedro/framework/cli/catalog.py
```diff
@@ -208,3 +208,51 @@ def rank_catalog_factories(metadata: ProjectMetadata, env):
...
+)
+@click.pass_obj
+def resolve_patterns(metadata: ProjectMetadata, pipeline, env):
+    """Resolve pipeline datasets with catalog factories."""
+
+    session = _create_session(metadata.package_name, env=env)
+    context = session.load_context()
+
+    data_catalog = context.catalog
+    catalog_config = data_catalog._raw_config
```
ankatiyar, line 232:
```suggestion
    raw_config = context.config_loader["catalog"]
```

kedro/framework/cli/catalog.py
```diff
@@ -208,3 +208,51 @@ def rank_catalog_factories(metadata: ProjectMetadata, env):
...
+
+    for ds_name in list(catalog_config.keys()):
+        if data_catalog._is_pattern(ds_name):
+            del catalog_config[ds_name]
```
ankatiyar, line 236:
```suggestion
    catalog_config = {ds_name: config for ds_name, config in catalog_config.items() if not data_catalog._is_pattern(ds_name)}
```

kedro/framework/cli/catalog.py
```diff
@@ -208,3 +208,51 @@ def rank_catalog_factories(metadata: ProjectMetadata, env):
...
+        if pl_obj:
+            pipeline_ds = pl_obj.data_sets()
+        else:
+            existing_pls = ", ".join(sorted(pipelines.keys()))
+            raise KedroCliError(
+                f"'{pipe}' pipeline not found! Existing pipelines: {existing_pls}"
+            )
+
+        for ds_name in pipeline_ds:
+            matched_pattern = data_catalog._match_pattern(
```
ankatiyar, line 251: Maybe a condition before this which is -
```python
if ds_name.startswith('params:'):
    continue
```

AhdraMeraliQB:
NOTE: Incomplete.

With the current implementation generated datasets and explicitly defined datasets have a discrepancy in their filepaths, with the former being absolute and the latter being relative. I am leaning towards having relative paths for all datasets returned by the command, which has been supported in an offline conversation with @merelcht. I am also roping in @amandakys in order to consider this from a design perspective as well.

kedro/io/data_catalog.py
AhdraMeraliQB: Added for documentation:

Spoke with @ankatiyar offline and whilst the `raw_config` will not have parameters that need filtering out, they will need to be filtered out of the pipeline entries to protect against a case where there is a catch-all pattern specified in the catalog.

AhdraMeraliQB:
Questions to come out of discussion with @amandakys:
1. Should we have some way of indicating which of the output datasets are resolved from dataset factories?
2. For future additions - should we introduce the ability to filter the output (e.g by pipeline, resolve factories only, e.t.c)

noklam:
> To test these changes, navigate to a local kedro project and run kedro catalog resolve. For maximum demonstration you should have some dataset factories in your catalog with explicit entries in the project pipeline.
> 
> I have used the spaceflights starter, replacing the preprocess_companies and preprocess_shuttles datasets in the catalog with:
> 
> preprocessed_{name}:
>   type: pandas.ParquetDataSet
>   filepath: data/02_intermediate/preprocessed_{name}.pq

GitPod will ease the setup for reviewer :)

AhdraMeraliQB:
@noklam 

Ah yes, we decided on relative paths (as the absolute paths got a bit chunky and difficult to read). Have also updated the description with the expected output.

AhdraMeraliQB:
@astrojuanlu 

Good point on the description - I struggled with the wording for the one-line explanation. Essentially dataset factories in the catalog are "resolved" against any applicable entries in the pipeline. Perhaps switching it around makes more sense "Resolve catalog factories against pipeline datasets" ?

docs/source/development/commands_reference.md
```diff
@@ -492,6 +493,14 @@ The command also accepts an optional `--pipeline` argument that allows you to sp
 kedro catalog list --pipeline=ds,de
 ```
 
+##### Resolve dataset factories in the catalog
+
+```bash
+kedro catalog resolve
+```
+
+This command resolves dataset factories in the catalog file with any explicit entries in the pipeline. The output is the aggregation of all explicit dataset entries in the catalog and factory datasets in the pipeline.
```
ankatiyar, line 502: I would say something like "This command shows a fully resolved catalog. It includes datasets explicitly mentioned in your catalog files and datasets mentioned in the project's pipelines resolved against corresponding dataset factory patterns". (Not exactly this but something to this effect)

kedro/framework/cli/catalog.py
```diff
@@ -207,3 +207,50 @@ def rank_catalog_factories(metadata: ProjectMetadata, env):
...
+
+    catalog_config = {
+        ds_name: ds_config
+        for ds_name, ds_config in catalog_config.items()
+        if not data_catalog._is_pattern(ds_name)
+    }
+
+    target_pipelines = pipelines.keys()
+
+    for pipe in target_pipelines:
```
ankatiyar, line 232: Could these nested loops be separated into two? 
* For all pipelines in `target_pipelines` add their datasets to a set
* Then for each `ds_name` in that set do the matching

AhdraMeraliQB: Definitely possible, but I am not able to see an advantage to the change? From my understanding the nested loops give us a time complexity of **number-of-pipelines * datasets-per-pipeline**, or **n * m/n** (if we assume an average distribution of datasets per pipeline, in any case we would still only iterate through m datasets regardless of distribution). This can be simplified to a linear time complexity just **m** of w.r.t the number of datasets. Having two separate loops gives us a time of **n+m** instead.

Is the aim just to move the matching logic outside of the nested loop? It would still be run on each dataset so I don't really see any optimisation to be gained from the change 🤔

ankatiyar: The reasoning for this being it would make the code easier to read. The factory datasets are only matched and added once in the current implementation, and they will be skipped in consequent checks so the time complexity wouldn't be affected much. This still loops through all datasets for all pipelines - which means all datasets are being checked at least twice (once for their own pipeline + once for the default pipeline + more if they belong to multiple pipelines)

AhdraMeraliQB: That makes sense! Changed in [8784a37](https://github.com/kedro-org/kedro/pull/2891/commits/8784a377b35135c7b36933f89737b682ab4c879e)

ankatiyar (approved):
Looks good to me! Manually tested it as well.

docs/source/development/commands_reference.md
```diff
@@ -492,6 +493,14 @@ The command also accepts an optional `--pipeline` argument that allows you to sp
 kedro catalog list --pipeline=ds,de
 ```
 
+##### Resolve dataset factories in the catalog
+
+```bash
+kedro catalog resolve
+```
+
+This command resolves dataset factories in the catalog file with any explicit entries in the pipeline. The output includes datasets explicitly mentioned in your catalog files and any datasets mentioned in the project's pipelines that resolve some dataset factory.
```
merelcht, line 502:
```suggestion
This command resolves dataset factories in the catalog file with any explicit entries in the pipeline. The output includes datasets explicitly mentioned in your catalog files and any datasets mentioned in the project's pipelines that match a dataset factory.
```

kedro/framework/cli/catalog.py
```diff
@@ -207,3 +207,51 @@ def rank_catalog_factories(metadata: ProjectMetadata, env):
...
+def resolve_patterns(metadata: ProjectMetadata, env):
+    """Resolve catalog factories against pipeline datasets"""
+
+    session = _create_session(metadata.package_name, env=env)
+    context = session.load_context()
+
+    data_catalog = context.catalog
+    catalog_config = context.config_loader["catalog"]
+
+    catalog_config = {
```
merelcht, line 224: It's a bit confusing that these variables have the exact name. In general it is best practice not to use the same variable as input and reassign the value in the same step. Maybe the second `catalog_config` can be re-named to `explicit_dataset`. That's what's stored there right?

tests/framework/cli/test_catalog.py
```diff
@@ -67,6 +67,23 @@ def fake_catalog_with_overlapping_factories():
     return config
 
 
+@pytest.fixture
+def fake_catalog_config_with_overwrite():
```
merelcht, line 71: What do you mean with "config_with_overwrite"?

tests/framework/cli/test_catalog.py
```diff
@@ -441,3 +458,121 @@ def test_rank_catalog_factories_with_no_factories(
...
+)
+def test_no_overwrite(
+    fake_project_cli,
+    fake_metadata,
+    fake_load_context,
+    mocker,
+    mock_pipelines,
+    fake_catalog_config_with_overwrite,
+):
+    """Test that explicit catalog entries are not overitten by factory config."""
```
merelcht, line 515:
```suggestion
    """Test that explicit catalog entries are not overwritten by factory config."""
```

tests/framework/cli/test_catalog.py
```diff
@@ -441,3 +458,121 @@ def test_rank_catalog_factories_with_no_factories(
...
+    )
+
+    assert not result.exit_code
+    assert yaml_dump_mock.call_count == 1
+
+    # 'parameters' and 'params:data_ratio' should not appear in the output
+    output = yaml_dump_mock.call_args[0][0]
+
+    assert "parameters" not in output.keys()
+    assert "params:data_ratio" not in output.keys()
```
merelcht, line 578: To demonstrate the right datasets are included in output it might be worth adding two more asserts for `iris_data` and `intermediate`

merelcht (commented):
I left some questions and suggestions, but other than that this is almost there 🙂

astrojuanlu:
The new one liner looks good to me ⭐

tests/framework/cli/test_catalog.py
AhdraMeraliQB: In this context this means a catalog configuration where one of the explicit entries in the catalog also matches a pattern in the catalog, and the test is ensuring the explicit entry is not overwritten. Perhaps `fake_catalog_config_with_overwritable_dataset()` could be more intuitive? Or `fake_catalog_config_with_resolvable_dataset()`?

kedro/framework/cli/catalog.py
AhdraMeraliQB: Yep! Changed in [779a286](https://github.com/kedro-org/kedro/pull/2891/commits/779a286f31c317ceb07591ff3c85d70ffcf40dd6)

tests/framework/cli/test_catalog.py
merelcht: I think `fake_catalog_config_with_resolvable_dataset` makes more sense to me.

merelcht (approved):
Great work! ⭐ 😄

PR #2899, opened 2023-08-04
Make unit tests compatible with cookiecutter 2.3.0

deepyaman:
## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ dependencies = [
     "build",
     "cachetools~=5.3",
     "click<9.0",
-    "cookiecutter>=2.1.1, <3.0",
+    "cookiecutter>=2.3.0, <3.0",
     "dynaconf>=3.1.2, <4.0",
     "fsspec>=2021.4, <2024.1",  # Upper bound set arbitrarily, to be reassessed in early 2024
     "gitpython~=3.0",


pyproject.toml
```diff
@@ -17,7 +17,7 @@ dependencies = [
     "build",
     "cachetools~=5.3",
     "click<9.0",
-    "cookiecutter>=2.1.1, <3.0",
+    "cookiecutter>=2.3.0, <3.0",
```
deepyaman, line 20: I will undo this; this is just to force CI to pick up latest cookiecutter.

noklam: Seems all tests pass. Is there anything failing?

deepyaman: I thought it would fail (since have something extremely similar in a private project that's failing), but it seems to pass. Maybe the difference is that Kedro uses Click, while we use Typer. :shrug:

FWIW what fixes our issue is adding a `""` to the end of the list in https://github.com/kedro-org/kedro/blob/0.18.12/tests/framework/cli/test_starters.py#L48:

```python
    return "\n".join([project_name, repo_name, python_package, ""])
```

That way, you get a new line after the last one, too.

Not sure what caused the changed behavior in cookecutter 2.3.0, but I'd imagine it has to do with the switch to Rich prompts.

PR #2591, opened 2023-05-18
Refactor the OmegaConfigLoader load_and_merge_from_dir method

noklam:
> **NOTE:** Kedro datasets are moving from `kedro.extras.datasets` to a separate `kedro-datasets` package in
> [`kedro-plugins` repository](https://github.com/kedro-org/kedro-plugins). Any changes to the dataset implementations
> should be done by opening a pull request in that repository.
## Description

The OCL has a big function `load_and_merge_from_dir` which does a lot of things. It makes it hard to reason about and write test for it.

We can simplify the test and extend the class by breaking it down. The PR is a quick effort of #2584 @ankatiyar  as we keep adding arguments to this function, making it hard to maintain. I think it's easier to have a draft PR to show what I mean, ultimately the PR should also update the test and `__getitem__` will be orchestrating the logic which simply call these functions (Not done in this draft).

## Checklist

- [ ] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [ ] Added tests to cover my changes

Issue #2481, opened 2023-03-29
[Proposal] - Replace ConfigLoader with ConfigLoader and ConfigResolver


noklam:
## Introduction
The current `ConfigLoader` is a big class that does more than loading configurations. It is responsible to load, resolve, validate and merge configurations for all `parameters`, `catalog`, `logging`.

## Background
The ConfigLoader starts to become too big and tests become more complicated. Kedro is a framework that consists of a few major components, `DataCatalog`, `ConfigLoader`, `Pipeline` and `Runner`. Addition classes like `KedroSession` and `KedroContext` are mainly for orchestrating the `kedro run`.

Currently, the most standalone component is `DataCatalog`, which can be used separately without the Kedro project template. However, it's not the case for `ConfigLoader` and there is potential to make it more modular.

## Problem
`ConfigLoader` is a big class and has many assumptions. It assumes you are working with a Kedro project, and it does all loading, resolving and merging configuration. It also assumes your configurations are file-based. There are workarounds that you can use hooks to override non-file base configuration, but that's not ideal. Why can't we have a non-file base ConfigLoader?

Many of these assumptions may not hold true, the easiest way to test it is to try to create a `ConfigLoader` in a Jupyter notebook and load some configuration file, it's not straightforward and it indicates there are strong coupling with the Kedro project structure.

Be as concrete as you can about:

## Design
Explain your design to the solution here. Diagrams could help.
The main idea is to separate `ConfigLoader` into `ConfigLoader` and `ConfigResolver`, potentially we can have a `ConfigManager` to orchestrate them but that may be going too far.

Once we realise the current `ConfigLoader` has two jobs, load and resolve (Templating, interpolating etc), we can split them into smaller classes. Here are some pseudo interfaces for the classes.

```python
class AbstractConfigLoader(UserDict):
    """Abstract base class for all ConfigLoader implementations."""
    # We can keep the Dict-like interface

class FileConfigLoader(AbstractConfigLoader):
    """ConfigLoader implementation for loading configuration from a file."""
    # This will be the same as the current `ConfigLoader`
           
class DatabaseConfigLoader(AbstractConfigLoader):
    """ConfigLoader implementation for loading configuration from a database."""
    def load(self, connection_string):
          ...

class ConfigResolver:
    """Abstract base class for all ConfigResolver implementations."""

    def resolve(self) -> Dict[str, Any]:
        """Resolves the configuration data."""
        raise NotImplementedError

[...]

PR #2921, opened 2023-08-11
Add globals feature for `OmegaConfigLoader` using a globals resolver

### LINKED ISSUE #2794: Add globals functionality to `OmegaConfigLoader` using custom "globals" resolver
# Description
After the introduction of variable interpolation in `OmegaConfigLoader`, two issues still stand -
    * Sharing variables across different configs (catalog, parameters etc)
    * Sharing variables across different envs (base, local etc)

This will be resolved by allowing global variables for configs. 

# Possible Implementation
Proposed by @marrrcin 🥇 

To use globals, in your project -
* Update `settings.py`
```python
CONFIG_LOADER_ARGS = {
    "globals_pattern": "*globals.yml"
}
```
* Add `globals.yml` to `conf/base/` or `conf/local`
* To use global values in `parameters.yml` or `catalog.yml` 
`globals.yml`
```yaml
datasets: 
    csv: pandas.CSVDataSet

test_size: 0.3
```
`catalog.yml`
```yaml
reviews:
  type: ${globals:"datasets.csv"}
  filepath: <path_to_file>
```
`parameters.yml`
```yaml
model_options:
    test_size: ${globals:"test_size"}
```
Here "`globals`" is a custom resolver - https://omegaconf.readthedocs.io/en/latest/custom_resolvers.html#id9

> **Q: Configurable `globals_pattern`?**
> A: Allow users to pass `globals_pattern` through `settings.py` but have it be `globals_pattern: {"*globals.yml"}` by default.
> 
> **Q: Where would `globals.yml` go?**
> A: The `globals.yml` files would stay within `conf/base/` and `conf/local/` instead of a top level `conf/globals.yml`. The common keys in `local/globals.yml` will overwrite the ones in `base/globals.yml`
> ```
> conf/
>   |_ base/
>     |_ globals.yml
>     |_ catalog.yml
>   |_ local/
>     |_ globals.yml
>     |_ catalog.yml
> ```
> 
> **Q: Soft or hard merge common keys?**
> A: Overwrite keys from `base/globals.yml` with keys from `local/globals.yml` for now but allow soft-merging in Kedro 0.19
> 
> **Q: Allow `globals_dict` like `TemplatedConfigLoader`?`**
> A: No, this use case for dynamically providing global values can be solved using "custom resolvers"
> 
> _Originally posted by @ankatiyar in https://github.com/kedro-org/kedro/issues/2175#issuecomment-1634269590_
>             

### TBD
Q: Do global variables need to start with "_"?
Q: What should the name of the custom resolver be? "`oc.global`", "`global`" etc?

ankatiyar:
## Description

Resolve #2794

## Development notes

* Added `globals` patterns to `config_patterns`
     * Can be changed/added in your project's `settings.py` through `CONFIG_LOADER_ARGS` NOTE: this is already present in `OmegaConfigLoader` and this step is only needed if users want a different pattern
    ```python
    CONFIG_LOADER_ARGS = {
        "config_patterns" = {
            "globals": ["globals*", <etc>]
         }
     }
    ```
     ❓ [QUESTION]  In the attached issue, I've mentioned passing globals patterns as separate argument "globals_pattern" and not as a part of "config_patterns". Does this make more sense? 

~~Try loading globals as a `dict` and assign to `self._globals` when `OmegaConfigLoader` is initialised.~~
  
SECTION BELOW NOT RELEVANT ANYMORE
>   * This needed me to do a some changes to the `__getitem__()` function since I couldn't use `self["globals"]` inside the `__init__` function without causing weird issues. So I separated the actual loading logic into `_get_config(key)` which is called through `__getitem__(key)` 
>     * Did some refactoring of `__getitem__()` and `load_and_merge_dir()` functions while I was at it. ❓ [QUESTION] Should I keep this or can also revert this? (I was trying a different approach which is why I separated the `_get_filtered_conf_paths()` and `_load_config_files()` initially but it is not strictly required for this implementation)

* In the `__init__` constructor -> `self._register_globals_resolver()` is called which registers a "globals" custom resolver to `OmegaConf` which calls the `_get_globals_value()` to resolve the value of the variable from `config_loader["globals"]` this happens when the config is loaded and merged by `load_and_merge_dir_config()`

 ### To Test this PR
* `settings.py` - Only if you want to change the default patterns
```python
from kedro.config import OmegaConfigLoader
CONFIG_LOADER_CLASS = OmegaConfigLoader
CONFIG_LOADER_ARGS = {
    "config_patterns": {
        "globals" : {"globals*"},
    },
}
```
* `conf/base/parameters.yml` (or any other conf file)
```yaml
my_param : "${globals:my_global_var}"
```
* `conf/base/globals.yml` 
```yaml
my_global_var: 34
```

❓ [QUESTION] Need feedback on the overwriting behaviour of globals - currently this is the same as `TemplatedConfigLoader`
* if I have globals in `local` they will override any globals in `base`
* If I have a global in `local` that doesn’t exist in the `globals.yml` in base then it should still be used in any `catalog.yml` or `parameters.yml` in `base`
* If I run `kedro run --env=base`  none of the `globals.yml` in `local` should be used because the `base` environment has been explicitly selected

### [TODO]
Waiting for feedback on the general approach before doing the following tasks -
- [ ] Release notes
- [x] More tests
- [x] Docs

## Checklist


kedro/config/omegaconf_config.py
```diff
@@ -136,10 +138,20 @@ def __init__(  # noqa: too-many-arguments
             env=env,
             runtime_params=runtime_params,
         )
+        # Read globals from patterns if the files exist
+        try:
+            self.globals = self._get_config("globals")
```
merelcht, line 143: What happens if a user tries to call `config_loaderr["globals"]`?

kedro/config/omegaconf_config.py
```diff
@@ -199,18 +208,19 @@ def __getitem__(self, key) -> dict[str, Any]:
...
-            f"config_patterns={self.config_patterns})"
-        )
+    def _get_conf_path(self, env):
+        if self._protocol == "file":
+            conf_path = str(Path(self.conf_source) / env)
+        else:
+            conf_path = str(Path(self._fs.ls("", detail=False)[-1]) / env)
+        return conf_path
 
     def load_and_merge_dir_config(  # noqa: too-many-arguments
```
merelcht, line 225: I think we need to have a closer look at which methods are now "public" and which "private". I know python doesn't strictly enforce these concepts, but initially the `OmegaConfigLoader` was built in a way that all "public" methods could easily be discovered and overwritten by users who want to build their own custom version on top of the class. Now `load_and_merge_dir_config` is still public but called from a private method `_get_config()`. Did you mark methods public/private with a specific reason in mind?

merelcht (commented):
I'm not quite sure I find the refactored code easier to follow. Especially extracting the loading code `_load_config_files` makes it harder for me to follow the code. @noklam might have a different view though. He started some refactoring a while back in https://github.com/kedro-org/kedro/pull/2591.

My main concern with the code though is around the "private" and "public" methods and the ease of customising `OmegaConfigLoader` for users.

I think it makes sense to have globals in `config_patterns` and not add a separate argument for `globals_patterns`.

kedro/config/omegaconf_config.py
ankatiyar: In the code if user does `config_loader["globals"]` it would return an empty dict `{}`

kedro/config/omegaconf_config.py
ankatiyar: My intention was to not change the private or public nature or the logic used by pre-existing functions. 
* `load_and_merge_dir` is still public and just certain parts of it have been taken out as helper methods. I don't think this would affect anyone who has overwritten `load_and_merge_dir()` method in their custom implementations.
    * The refactoring of `load_and_merge_dir()` is not necessarily needed to make globals work tbh, I can revert it but I did see @noklam had proposed something similar in #2591 
* `__getitem__()` is a magic function which is treated essentially like a private method (it's not meant to be called directly) so I would think it doesn't makes much of a difference if some of it's code is separated into another private method. It can still be overwritten by the users as well. 
    * This change though is required for this implementation to work

kedro/config/omegaconf_config.py
ankatiyar: It's not throwing a `MissingConfigException` because we don't require users to have `globals.yml` in their config necessarily

merelcht: Will it also return an empty dict if the user does have globals? That would be quite confusing..

noklam:
> I'm not quite sure I find the refactored code easier to follow. Especially extracting the loading code `_load_config_files` makes it harder for me to follow the code. @noklam might have a different view though. He started some refactoring a while back in #2591.
> 
> My main concern with the code though is around the "private" and "public" methods and the ease of customising `OmegaConfigLoader` for users.
> 
> I think it makes sense to have globals in `config_patterns` and not add a separate argument for `globals_patterns`.



In general I think some methods are too big, signified with the fact that we have a function called `load_and_merge_dir_config`. a `xxx_and_yyy` function usually is a hint that this is hard to test because it does multiple things at the same time. In https://github.com/kedro-org/kedro/pull/2591/files I try to break it into 3 steps
1. Build up the path to load config (aware of the pattern)
2. Load config (only aware of the path)
3. Merge config (aware of the config / dictionary)

- https://github.com/kedro-org/kedro/issues/2481
This issue further push the idea that ConfigLoader isn't really a "loader" because it does "load" and "resolve". We could then break it into two class so potentially we could take config anywhere (other than filebase). Conceptually I think this is the right thing to do but there isn't immediate use cases.

```
class FileConfigLoader(AbstractConfigLoader):
    """ConfigLoader implementation for loading configuration from a file."""
    # This will be the same as the current `ConfigLoader`
           
class DatabaseConfigLoader(AbstractConfigLoader):
    """ConfigLoader implementation for loading configuration from a database."""
    def load(self, connection_string):
          ...

class ConfigResolver:
    """Abstract base class for all ConfigResolver implementations."""

    def resolve(self) -> Dict[str, Any]:
        """Resolves the configuration data."""
        raise NotImplementedError

class JinjaConfigResolver(ConfigResolver):
    """ConfigResolver implementation for resolving Jinja templates."""

    def resolve(self) -> Dict[str, Any]:
        """Resolves the configuration data using Jinja templates."""
        # implementation here
```

Back to this PR specificially, I think I am not quite sure what's the constrains since @ankatiyar mention the refactor is necessary for the implementation.

Few comments:
- Moving logic from `__getitem__` -> `_get_config` doesn't seem to be clearer to me.
- `_get_conf_path` looks good to me
- How different `_get_filtered_conf_paths` from `_build_conf_path` we used to have?

ankatiyar:
@merelcht @noklam I've reverted the refactoring stuff, turns out it works fine without it. We can keep the scope of this PR just to adding the globals functionality.

merelcht:
> @merelcht @noklam I've reverted the refactoring stuff, turns out it works fine without it. We can keep the scope of this PR just to adding the globals functionality.

Okay cool, in that case let's indeed just keep this PR for adding the globals functionality and do refactoring separately. I think @noklam made a great point about our configuration loaders being more than just loaders and actually doing resolving as well. For the time being I wouldn't go as far as introducing the separation into two classes, but we can have another look at improving the methods and specifically the large `load_and_merge_dir_config`.

docs/source/configuration/advanced_configuration.md
```diff
@@ -262,6 +263,34 @@ Since both of the file names (`catalog.yml` and `catalog_globals.yml`) match the
 #### Other configuration files
 It's also possible to use variable interpolation in configuration files other than parameters and catalog, such as custom spark or mlflow configuration. This works in the same way as variable interpolation in parameter files. You can still use the underscore for the templated values if you want, but it's not mandatory like it is for catalog files.
 
+### How to use global variables with the `OmegaConfigLoader`
+From Kedro `0.18.13`, you can also use variable interpolation in your configurations using "globals" with `OmegaConfigLoader`.
```
ankatiyar, line 267: Note to self: Change version depending on when this is merged.

ankatiyar:
I've made some final changes + added docs and tests - 
* Removed `self.globals` from `__init__`. Globals are loaded when the resolver `_get_globals_value()` is called by the "globals" resolver. This means that the globals are loaded multiple times but now if the users bypass configuration and set it in their `settings.py`, it will still work -
```python
conf_path = str(project_path / settings.CONF_SOURCE)
conf_loader = OmegaConfigLoader(conf_source=conf_path)
config_loader["globals"] = {"my_global" : 23 }
```
* Trying `config_loader["globals"]` should not return `MissingConfigException` but return an empty dict if there are no globals files.

* You can have nested global variables - 
`globals.yml`
```yaml
nested:
  x: 23
```
 Can be used in parameters/catalog etc with dot notation -
```yaml
my_param: "${globals:nested.x}"
```

docs/source/configuration/advanced_configuration.md
merelcht, line 267:
```suggestion
From Kedro `0.18.13`, you can use variable interpolation in your configurations using "globals" with `OmegaConfigLoader`.
```

docs/source/configuration/advanced_configuration.md
```diff
@@ -262,6 +263,34 @@ Since both of the file names (`catalog.yml` and `catalog_globals.yml`) match the
...
+The benefit of using globals over regular variable interpolation is that the global variables are shared across different configurations.
```
merelcht, line 268:
```suggestion
The benefit of using globals over regular variable interpolation is that the global variables are shared across different configuration types, such as catalog and parameters.
```

tests/config/test_omegaconf_config.py
```diff
@@ -671,3 +672,89 @@ def test_custom_resolvers(self, tmp_path):
         assert conf["parameters"]["model_options"]["param1"] == 7
         assert conf["parameters"]["model_options"]["param2"] == 3
         assert conf["parameters"]["model_options"]["param3"] == "my_env_variable"
+
+    def test_globals(self, tmp_path):
```
merelcht, line 676: This test is doing quite a lot of things, I'd probably separate it into: 
1. Just checking that the omegaconfigloader has the resolver and you can load the globals with `assert conf["globals"] == globals_config`
2. Check that resolving is done properly (for both params & catalog)
3. Check that nested resolving is done properly

merelcht (commented):
This is looking so goooood! 👏  Great docs as well. I just left one comment about a test.

tests/config/test_omegaconf_config.py
ankatiyar: Done!

merelcht (approved):
Awesome work @ankatiyar ! 🤩 Don't forget to update the release notes!

kedro/config/omegaconf_config.py
```diff
@@ -109,6 +110,7 @@ def __init__(  # noqa: too-many-arguments
             "parameters": ["parameters*", "parameters*/**", "**/parameters*"],
             "credentials": ["credentials*", "credentials*/**", "**/credentials*"],
             "logging": ["logging*", "logging*/**", "**/logging*"],
+            "globals": ["globals*", "globals*/**", "**/globals*"],
```
noklam, line 113:
```
        >>> # in settings.py
        >>> from kedro.config import TemplatedConfigLoader
        >>>
        >>> CONFIG_LOADER_CLASS = TemplatedConfigLoader
        >>> CONFIG_LOADER_ARGS = {
        >>>     "globals_pattern": "*globals.yml",
        >>> }
```
```suggestion
            "globals": ["globals*"],
```

I suggest keep this pattern simple.
1. Align with the current `TemplatedConfigLoader` pattern, the above code block is copy from our template
2. Most people don't need nested `globals`,  putting globals in folder should be minority, if so they should change the settings instead.
3. `**/globals*` is quite dangerous, as this could match `parameters/globals_parameters.yml` - although we usually suggest people using `parameter_globals` instead.

I may even suggest a more conservative default  `"globals": "globals.yml"`, forcing the default to be just `globals.yml`. If we change this after release it will be breaking change.

WDYT? @ankatiyar @merelcht , cc @stichbury because I really think we should separate the terminology for `globals`, it's being overloaded a lot now and this can lead to weird bugs.

noklam (approved):
Excited to have this feature finally : ). Please include a release note to highlight this feature!

docs/source/configuration/advanced_configuration.md
```diff
@@ -262,6 +263,34 @@ Since both of the file names (`catalog.yml` and `catalog_globals.yml`) match the
 #### Other configuration files
 It's also possible to use variable interpolation in configuration files other than parameters and catalog, such as custom spark or mlflow configuration. This works in the same way as variable interpolation in parameter files. You can still use the underscore for the templated values if you want, but it's not mandatory like it is for catalog files.
 
+### How to use global variables with the `OmegaConfigLoader`
+From Kedro `0.18.13`, you can use variable interpolation in your configurations using "globals" with `OmegaConfigLoader`.
+The benefit of using globals over regular variable interpolation is that the global variables are shared across different configuration types, such as catalog and parameters.
+By default, these global variables are assumed to be in files called `globals.yml` in either of your environments. If you want to configure the naming patterns for the files that contain your global variables,
```
merelcht, line 269:
```suggestion
By default, these global variables are assumed to be in files called `globals.yml` in any of your environments. If you want to configure the naming patterns for the files that contain your global variables,
```

merelcht (approved):

ankatiyar:
Added a more helpful error message to address this problem - https://github.com/kedro-org/kedro/issues/2794#issuecomment-1682420305
> We found that `_` is not support for `globals.yml`. It will now give an clear message that you shouldn't use `_` for globals at all.
> 
> Noted this is not because of the implementation of the resolver, but rather `OmegaConfigLoader` filter out everything with `_`. If it turns out it is needed we can always add it and it would not be breaking. (if we enable it now and change later , it will be breaking).

Essentially, if you have keys starting with '_' in `globals.yml`, it gets filtered out by the `load_and_merge_dir_config()` when loading the "globals". Then if you try to reference it in your `parameters.yml` or `catalog.yml` like - `param1 : "${globals:_global_value}". It would just say "Globals key '_global_value' not found." Even if it does exist in the `globals.yml`.

ankatiyar:
Also added the ability to provide a default value for globals resolver to fall back on if the key does not exist. + updated test and docs

docs/source/configuration/advanced_configuration.md
```diff
@@ -262,6 +263,38 @@ Since both of the file names (`catalog.yml` and `catalog_globals.yml`) match the
...
+```yaml
+my_param : "${globals:my_global_value}"
+```
+`conf/base/catalog.yml`:
+```yaml
+companies:
+  filepath: data/01_raw/companies.csv
+  type: "${globals:dataset_type.csv}"
+```
+You can also provide a default value to be used incase the globals variable does not exist:
```
merelcht, line 290:
```suggestion
You can also provide a default value to be used in case the globals variable does not exist:
```

kedro/config/omegaconf_config.py
noklam: Offline conversation - agree with using default with `globals.yml`.

kedro/config/omegaconf_config.py
```diff
@@ -308,6 +311,34 @@ def _is_valid_config_path(self, path):
...
+        for k in keys:
+            value = value.get(k)
+            if not value:
+                if default_value:
+                    _config_logger.debug(
+                        f"Using the default value for the global variable {variable}."
+                    )
+                    return default_value
+                msg = f"Globals key '{variable}' not found and no default value provided. "
+                if variable.startswith("_"):
```
marrrcin, line 337: Shouldn't this check happen earlier (at the very first line of the `_get_globals_value`)?

tests/config/test_omegaconf_config.py
```diff
@@ -671,3 +672,128 @@ def test_custom_resolvers(self, tmp_path):
...
+        }
+        local_param_config = {
+            "param2": "${globals:x}",
+        }
+        base_globals_config = {
+            "x": 34,
+            "y": 25,
+        }
+        local_globals_config = {
+            "y": 99,
```
marrrcin, line 751: How this will work for nested keys in the following configuration?

```python
        base_globals_config = {
            "x": 34,
            "y": {
"a": 7,
"z": 6
}
        }
        local_globals_config = {
            "y": {
"z": -11
}
        }
```

marrrcin (commented):
Looks great, really clean implementation. Added 2 small questions.

tests/config/test_omegaconf_config.py
ankatiyar: Right now `OmegaConfigLoader` doesn't softmerge keys so the final "globals" variables will be -
```yaml
x: 34,
y: 
   z: -11
```

kedro/config/omegaconf_config.py
ankatiyar: Good point! I moved it.

marrrcin (approved):

PR #2922, opened 2023-08-13
PiP/pyproject.toml and Conda/Meta.yaml Sync

rxm7706:
https://github.com/conda-forge/kedro-feedstock/pull/42

## Checklist


rxm7706:
@noklam @astrojuanlu this is ready for review - But I cant seem to get past the DSO error 
Parallel PR https://github.com/conda-forge/kedro-feedstock/pull/42

noklam:
You need to sign-off your commit. If you click the DCO icon it shows you the instructions.

https://github.com/kedro-org/kedro/runs/15848030417

rxm7706:
> You need to sign-off your commit. If you click the DCO icon it shows you the instructions.
> 
> https://github.com/kedro-org/kedro/runs/15848030417

I tried - and it seems to show signed off on my branch - is there a way to overide, or for you to make a new PR using the pyproject.toml here. 

image

noklam:
@rxm7706 I retriggered the CI, DCO seems passing now.

pyproject.toml
```diff
@@ -12,30 +12,30 @@ authors = [
...
-    "gitpython~=3.0",
-    "importlib-metadata>=3.6; python_version >= '3.8'",
-    "importlib_metadata>=3.6, <5.0; python_version < '3.8'",  # The "selectable" entry points were introduced in `importlib_metadata` 3.6 and Python 3.10. Bandit on Python 3.7 relies on a library with `importlib_metadata` < 5.0
-    "importlib_resources>=1.3",  # The `files()` API was introduced in `importlib_resources` 1.3 and Python 3.9.
-    "jmespath>=0.9.5, <2.0",
-    "more_itertools>=9,<11",
-    "omegaconf~=2.3",
-    "parse~=1.19.0",
+    "anyconfig>=0.10.0,<1.0.0",
+    "attrs>=21.3,<24.0",
```
astrojuanlu, line 16: Do we know of any known incompatibility?

pyproject.toml
```diff
@@ -12,30 +12,30 @@ authors = [
...
+    "build>=0.7.0,<1.0.0",
```
astrojuanlu, line 17: Same as above

pyproject.toml
```diff
@@ -12,30 +12,30 @@ authors = [
...
+    "cachetools>=4.1,<6.0",
+    "click>=7.0,<9.0",
+    "cookiecutter>=2.1.1,<3.0",
+    "dynaconf>=3.1.2,<4.0",
+    "fsspec>=2021.4,<2024.1",  # Upper bound set arbitrarily, to be reassessed in early 2024
+    "gitpython>=3.0,<4.0",
+    "importlib_metadata>=3.6,<5.0; python_version < '3.8'",  # The "selectable" entry points were introduced in `importlib_metadata` 3.6 and Python 3.10. Bandit on Python 3.7 relies on a library with `importlib_metadata` < 5.0
+    "importlib-metadata>=3.6,<7.0; python_version >= '3.8'",
```
astrojuanlu, line 25: Wondering what happened with `bandit` here?

pyproject.toml
```diff
@@ -12,30 +12,30 @@ authors = [
...
+    "importlib_resources>=1.3,<7.0",  # The `files()` API was introduced in `importlib_resources` 1.3 and Python 3.9.
```
astrojuanlu, line 26: Are we aware of any known incompatibility?

pyproject.toml
```diff
@@ -12,30 +12,30 @@ authors = [
...
+    "parse>=1.19.0,<=2.0.0",
     "pip-tools>=6.5,<8",
-    "pluggy~=1.0",
-    "PyYAML>=4.2, <7.0",
-    "rich>=12.0, <14.0",
-    "rope>=0.21, <2.0",  # subject to LGPLv3 license
-    "setuptools>=65.5.1",
-    "toml~=0.10",
-    "toposort~=1.5",  # Needs to be at least 1.5 to be able to raise CircularDependencyError
+    "pluggy>=1.0,<2.0",
```
astrojuanlu, line 32: This is essentially equivalent if I'm not mistaken, should we leave it as it is?

pyproject.toml
```diff
@@ -12,30 +12,30 @@ authors = [
...
+    "PyYAML>=4.2,<7.0",
+    "rich>=12.0,<14.0",
+    "rope>=0.21,<2.0",  # subject to LGPLv3 license
+    "setuptools>=65.5.1,<70.0.0",
```
astrojuanlu, line 36: Are we using any deprecated functionality in `setuptools`?

astrojuanlu (commented):
Thanks @rxm7706 for the pull request. I'm in favour of relaxing the upper bounds when feasible, and loosening the lower bounds if we have tried Kedro with older versions of the dependencies.

What I'm generally not in favour of is turning `>=` into `>=,<` (hence introducing upper bounds when there were none). So I left a comment for every case to make sure we are doing it if there's a known incompatibility.

As you said in https://github.com/conda-forge/kedro-feedstock/pull/42#issue-1817531932, Kedro needs to support a broader range of package versions. Putting too many upper version caps ends up being a maintainability nightmare, so let's not add new ones unless strictly necessary IMHO.

rxm7706:
> What I'm generally not in favour of is turning `>=` into `>=,<` (hence introducing upper bounds when there were none). So I left a comment for every case to make sure we are doing it if there's a known incompatibility.

Thanks @astrojuanlu  Agree with the point above - there are two ways I have seen upper bounds done.. Arbitrary calendar version (e.g. like dask <2024.1) or semantic major version (pluggy>=1.0,<2.0") 

1. The packages that try to stay most compatible seem to use both  `>=,<` - then disallow versions that only broke functionality (or had a security vulnerability)
https://github.com/conda-forge/google-api-core-feedstock/blob/main/recipe/meta.yaml#L28C1-L28C111
2. Too many packages just look at dependabot / pyupgrade and keep updating their lower bound - so I tried to use the lower bounds from a kedro at 0.18.1 / 0.18.8 as a reference to set lower bounds (and not change lower bounds unless there is need to (kedro tests or features break - or a known vulnerability) 

Let me know if you want to remove "arbitrary upper bounds, after you have taken a look at 
https://github.com/conda-forge/google-api-core-feedstock/blob/main/recipe/meta.yaml - and I can make the recommended changes / proposals.

pyproject.toml
rxm7706: I dont see bandit in either the pyproject,toml or the conda-feedstock..

pyproject.toml
rxm7706: this was existing

rxm7706:
> Thanks for the extra context - yes I'd say let's remove the arbitrary upper bounds

@astrojuanlu - changes made, please review. Regards.

astrojuanlu (approved):
Thanks for your patience @rxm7706. My initial comment (should have been clearer) was about removing the upper bounds _introduced by this PR_. There were already some pre-existing upper bounds, and you removed some of those in your last commit.

I don't have the full context on all the dependencies, but all the tests seem to pass and I don't see anything that warrants extra caution.

We just have to be mindful that this shifts the problem from "the dependency contraints are too tight and I can't install Kedro with X" to "latest upgrade of X broke Kedro". The point is that users cannot do anything about the former, just wait for us to make a release, while they can, after some debugging, properly pin their dependencies in the latter case.

Having said that, I'm in favour of this change.

pyproject.toml
astrojuanlu: cc @merelcht (I don't have any experience with bandit)

rxm7706:
> My initial comment (should have been clearer) was about removing the upper bounds _introduced by this PR_. There were already some pre-existing upper bounds, and you removed some of those in your last commit.

@astrojuanlu I think your comment, was clear enough for me-- and I actually understood it as remove only the added upper bounds, and that's what I set out to do - the only upper bound I meant to remove was was the dask one which I understood as arbitrary. Let me know if you need me to fix anything else. Thank You.

rxm7706:
Hi @noklam - Referring back to https://github.com/conda-forge/kedro-feedstock/pull/42#issuecomment-1684186096
I believe we still have an open question on how to handle click -

noklam:
Thanks. I look back in the history and click never cause any problem

1. click==6.7 # First open source Kedro pinned version - https://github.com/kedro-org/kedro/blame/7d8687b7c74dc09b8dac62c1694f03c0c867a988/requirements.txt
2. Bump to 8.0  https://github.com/kedro-org/kedro/commit/065b7cc45da28b67394fb6205d73333c7095d962


>So looks I added a lower bound.. which we don't need.. and the "click<9.0" is an existing arbitrary upper bound... that can be removed / made consistent across pyproject.toml and conda recipe

> What would your preference be
> 
> click <9.0 - Arbitrary Upper Bound (current in pyproject main)
> click - No Bounds, equivalent to our Arbitrary Upper Bound but cleaner than current that keeps an arbitrary upper
> click >=7.0 - Arbitrary Lower Bound - remove arbitrary upper bound
> click >=7.0,<9.0 - Arbitrary Lower Bound added keeping current Arbitrary Lower Bound - currently in pyproject and conda-forge PR's

I test the minimal version of click is `4.0`, I don't have a problem with keeping an open bound.
`click >=4.0`

astrojuanlu:
Let's go with that, thanks @noklam 👍🏽

rxm7706:
> Let's go with that, thanks @noklam 👍🏽

@noklam @astrojuanlu Thank you. Ready for review.

astrojuanlu (approved):
LGTM! Thanks a lot for the patience @rxm7706 🙏🏽

astrojuanlu:
Pinging @merelcht again for review. @rxm7706 this will require one more rebase, please 🙏🏽

pyproject.toml
merelcht: It looks like `bandit` is in the test requirements. We use it in the `pre-commit` flow, but it shouldn't be a requirement of Kedro itself.

merelcht (approved):
Thank you so much for doing this @rxm7706 ⭐ 👍

rxm7706:
> Pinging @merelcht again for review. @rxm7706 this will require one more rebase, please 🙏🏽

@astrojuanlu @noklam - Done - Please review.

pyproject.toml
```diff
@@ -12,30 +12,30 @@ authors = [
...
+    "gitpython>=3.0",
+    "importlib_metadata>=3.6,<5.0; python_version < '3.8'",  # The "selectable" entry points were introduced in `importlib_metadata` 3.6 and Python 3.10. Bandit on Python 3.7 relies on a library with `importlib_metadata` < 5.0
+    "importlib-metadata>=3.6,<7.0; python_version >= '3.8'",
+    "importlib_resources>=1.3,<7.0",  # The `files()` API was introduced in `importlib_resources` 1.3 and Python 3.9.
+    "jmespath>=0.9.5",
+    "more_itertools>=8.14.0",
+    "omegaconf>=2.1.1",
+    "parse>=1.19.0",
+    "pip-tools>=6.5",
+    "pluggy>=1.0",
```
astrojuanlu, line 32: Please restore the `<1.3` cap, otherwise our tests fail

rxm7706:
@astrojuanlu @noklam fixed pluggy cap version. Thank you.

noklam (approved):
Thanks, appreciate it.

astrojuanlu:
I saw a test failing so I restarted the workflow just in case, but if it fails again, please take a look. Maybe it's something with the dependencies we changed.

astrojuanlu:
It passed :) Merging!

pyproject.toml
```diff
@@ -12,30 +12,30 @@ authors = [
...
-    "jmespath>=0.9.5, <2.0",
-    "more_itertools>=9,<11",
-    "omegaconf~=2.3",
-    "parse~=1.19.0",
-    "pip-tools>=6.5,<8",
-    "pluggy>=1.0, <1.3",  # TODO: Uncap when dropping Python 3.7 support, see https://github.com/kedro-org/kedro/issues/2979
-    "PyYAML>=4.2, <7.0",
-    "rich>=12.0, <14.0",
-    "rope>=0.21, <2.0",  # subject to LGPLv3 license
+    "build>=0.7.0,",
```
astrojuanlu, line 17: Woops we slipped a typo here. ⚠️

PR #2931, opened 2023-08-16
Enable overriding of config keys through run time parameters

### LINKED ISSUE #2531: Enable overriding configuration via variable interpolation for all configurations other than just `parameters` for OmegaConfigLoader
## Description
Is your feature request related to a problem? A clear and concise description of what the problem is: "I'm always frustrated when ..."

With #2467 we resolves OCL earlier and parameters propogate correctly for `parameters`. We cannot do the same for `catalog` or other configuration yet due to the problem of "residual" configurations. The problem is specific to `OmegaConfigLoader`, because it requires merging the dictionaries before resolving.

Take this simple example

```yaml
# catalog.yml
my_dataset:
   path: ${base_path}/my_file.txt
```

## Expected Result
`kedro run --params base_path:my_bucket` this should work and replace the value correctly.

## Current Result
The value will not propagate for configurations other than `parameters`. This is because we need to merge `aggregated_configs` with `runtime_params`, which produce invalid catalog entries.

The compiled catalog may look like this
```yaml
# catalog.yml
my_dataset:
  path: ${base_path}/my_file.txt


base_path: my_bucket # <-- This is not a dataset, ideally we need to remove it or bypass it.
```

## Context
This is essential to enable common use case like providing a bucket name for `catalog` via CLI.
`kedro run --params base_path:my_bucket` , with TemplatedConfigLoader it doesn't work either, but there are [workarounds](https://github.com/kedro-org/kedro/discussions/1782). In contrast, there is no workaround for OmegaConfigLoader.

## Possible Implementation
1. Figure out a way to remove the residual configurations

The implementation should make sure above case will work correctly. It should also work for arbitrary configurations, i.e. `config_loader["spark"] or `config_loader["my_random_config"]` as long as it's registered with `config_patterns`

## Possible Alternatives

> How does “global” fit in the picture? It will be clearer when we finished Change 1
> Antony thinks that with the validation change, it shouldn’t be matter even when the “residuals” are left in the config, Python config also don’t do anything special to extra configurations.
> Ivan think this doesn’t solve the problem in a generic way, config_lodaer[“global”] is purposed, but this doesn’t solve all the problem because OmegaConfigLoader requires to merge dictionary before interpolating. This shouldn’t be a limitation, but will requires more work to implement a solution
> Merel has some concern about the messaging of “meta_params”, will people really understand which argument to be used.
> kedro run --params=base_path:s3_bucket will not work for catalog yet, but agree that it would be nice to support.

ankatiyar:
## Description

Resolve #2531 

## Checklist

### DIFF
--- a/kedro/config/omegaconf_config.py
+++ b/kedro/config/omegaconf_config.py
@@ -7,7 +7,7 @@ import io
 import logging
 import mimetypes
 from pathlib import Path
-from typing import Any, Callable, Iterable, Mapping
+from typing import Any, Callable, Iterable
 
 import fsspec
 from omegaconf import OmegaConf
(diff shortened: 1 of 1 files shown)


ankatiyar:
Closing in favour of the discussed solution using resolvers

PR #2983, opened 2023-08-29
Improve global resolvers for missing keys

noklam:
## Description

- Refactor the variable nested_key
- remove lambda, use the function, move the default to the underlying function
- Add a test that should fail (passing now)

Currently when access a non-exist global, a default value is return (I believe this is undesired)

## Checklist


noklam:
I think we should have a sentinel value `NOVALUE` rather than `None` to differentiate a user define default `None` versus a no-op

`${globals: DUMMY_KEY, None}` should be different from `${globals: DUMMY_KEY}`

astrojuanlu:
Is this still relevant after merging #2976?

noklam:
closing in favor or #2976,  I discussed with @ankatiyar and the changes are included in the same PR already.

PR #2989, opened 2023-08-31
Load micropackage dependencies statically

astrojuanlu:
## Description

This avoids building the wheel in an isolated environment just to extract the dependencies.

Cons: It relies on a PEP-621 compliant `pyproject.toml` file to be present, which only happens for Kedro projects created after https://github.com/kedro-org/kedro/issues/2280 was closed. Therefore, it is backwards incompatible.

Pros: It makes `kedro micropkg pull` fast again, cutting the time to run the full test suite in ~half (from 12 to 7 minutes on Linux, from 14 to 6 minutes on Windows 🎉)

## Development notes

This uses `pyproject-metadata` instead of PyPA/build, see https://pypi.org/project/pyproject-metadata/

## Checklist

### DIFF
--- a/kedro/framework/cli/micropkg.py
+++ b/kedro/framework/cli/micropkg.py
@@ -17,13 +17,13 @@ import click
 from omegaconf import OmegaConf
 from packaging.requirements import InvalidRequirement, Requirement
 from packaging.utils import canonicalize_name
+from pyproject_metadata import StandardMetadata
 from rope.base.project import Project
 from rope.contrib import generate
 from rope.refactor.move import MoveModule
 from rope.refactor.rename import Rename
 from setuptools.discovery import FlatLayoutPackageFinder
 
-from build.util import project_wheel_metadata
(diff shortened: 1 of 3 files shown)


noklam:
Is it possible to keep both, using the Metadata when available and fallback if not?

astrojuanlu:
Yes indeed!

astrojuanlu:
I don't have time to pursue the backwards-compatible approach, leaving this as a POC in case anybody wants to pick it up. Essentially this would need

- To retain `build` as a dependency,
- Conditionally try to locate the PEP-621 metadata in `pyproject.toml` and otherwise use the slow codepath, and
- Test both things.

merelcht:
@astrojuanlu Would it make sense to write this up in a ticket instead of keeping a PR open that will become more and more outdated? It's not super likely that this will be merged as is since it's from August and solving conflicts might be a real pain.

astrojuanlu:
Closing in favour of #3457

PR #2993, opened 2023-08-31
Make dataset factory resolve nested dict properly

### LINKED ISSUE #2992: Dataset factories doesn't resolve nested config properly
## Description
Reported by @m-gris
Dataset factories does not work properly for config that is nested.

## Context
The problem is in `DataCatalog._resolve_config()`. It fills out the placeholder on top level keys properly but doesn't go into nested dicts.

## Steps to Reproduce
Message quoted from slack - 

> Hi everyone,
> I’ve started to use kedro-mlflow and am under the impression that there is some conflict / incompatibility with the dataset factory.
> Having moved from
> ```yaml
> "models.{model}.embeddings":
>     type: pickle.PickleDataSet
>     filepath: "data/07_model_output/{model}_embeddings.pkl"
>     backend: joblib
>     versioned: True
>     layer: model_output
> ```
> to
> ```yaml
> "models.{model}.embeddings":
>   type: kedro_mlflow.io.artifacts.MlflowArtifactDataSet
>   data_set:
>     type: pickle.PickleDataSet
>     filepath: "data/07_model_output/{model}_embeddings.pkl"
>     backend: joblib
>     versioned: True
>   layer: model_output
> ```
> results in artifacts being saved without the model named properly interpolated, i.e “litteraly” saved as {model}_embeddings.pkl
> Any comment ? Anything I’m missing ?
>

ankatiyar:
## Description

Fix #2992 
- resolve config properly for nested dict
- resolve config properly for lists

## Development notes

* I have updated the `DataCatalog._resolve_config()` function -
    * To recursively call itself if the config is a nested dict.
    * Recursively call itself for each item if the config is a list
    * if it is a string, try to replace the placeholder
    * Into a class method since it doesn't need access to `DataCatalog._dataset_patterns` but takes the config dict as an argument
* Added a test for nested dict config, list, list + dict cases
* Updated dataset factories CLI commands to use the updated `_resolve_config()`
* Updated release notes
## Checklist


datajoely:
Super speedy!

kedro/io/data_catalog.py
```diff
@@ -436,16 +439,21 @@ def __contains__(self, data_set_name):
...
         data_set_name: str,
         matched_pattern: str,
+        config_copy: dict,
     ) -> dict[str, Any]:
         """Get resolved AbstractDataset from a factory config"""
         result = parse(matched_pattern, data_set_name)
-        config_copy = copy.deepcopy(self._dataset_patterns[matched_pattern])
         # Resolve the factory config for the dataset
         for key, value in config_copy.items():
+            if isinstance(value, Dict):
```
noklam, line 453: Can we keep `dict` here? Unsure why we need typing here.

noklam (approved):
Approve with small comments.

kedro/io/data_catalog.py
ankatiyar: fixed!

kedro/io/data_catalog.py
```diff
@@ -436,27 +439,33 @@ def __contains__(self, data_set_name):
...
+        if isinstance(config, dict):
+            for key, value in config.items():
+                config[key] = cls._resolve_config(data_set_name, matched_pattern, value)
+        elif isinstance(config, list):
+            config = [
+                cls._resolve_config(data_set_name, matched_pattern, value)
+                for value in config
+            ]
+        elif isinstance(config, Iterable) and "}" in config:
+            try:
```
noklam, line 461: Why list and Iterable are treated differently? You mentioned there is a new example from user maybe it worth to document it in the PR/issue.

Gabriel2409: Hi @noklam, yes I provided an example in #2992

Iterable here basically means string because of the second condition. 
In fact, I think having `and "}" in config:` is actually misleading if you expect the condition to work with nested dictionaries and lists. 

Indeed, while strings, lists and dicts are Iterable, the second check fails for dict and list, see example below

```python
from typing import Iterable

config = "my{param}" 
print(isinstance(config, Iterable)) # True
print("}" in config) # True

config = ["my{param}"]
print(isinstance(config, Iterable)) # True
print("}" in config) # False

config = {"mykey":"my{param}"}
print(isinstance(config, Iterable)) # True
print("}" in config) # False
```

Which is why the solution is to only go inside the base case when you actually substitute a string. 
To make things clearer, maybe `Iterable` should be replaced by `str`.

ankatiyar: Yep this should be string

tests/io/test_data_catalog.py
```diff
@@ -840,7 +863,10 @@ def test_unmatched_key_error_when_parsing_config(
     ):
         """Check error raised when key mentioned in the config is not in pattern name"""
         catalog = DataCatalog.from_config(**config_with_dataset_factories_bad_pattern)
-        pattern = "Unable to resolve 'filepath' for the pattern '{type}@planes'"
+        pattern = (
+            "Unable to resolve 'data/01_raw/{brand}_plane.pq' from the pattern '{type}@planes'. "
+            "Keys used in the config should be present in the dataset factory pattern."
```
merelcht, line 868:
```suggestion
            "Keys used in the configuration should be present in the dataset factory pattern."
```

kedro/io/data_catalog.py
```diff
@@ -436,27 +439,33 @@ def __contains__(self, data_set_name):
...
+                cls._resolve_config(data_set_name, matched_pattern, value)
+                for value in config
+            ]
+        elif isinstance(config, str) and "}" in config:
+            try:
+                config = str(config).format_map(result.named)
+            except KeyError as exc:
+                raise DatasetError(
+                    f"Unable to resolve '{config}' from the pattern '{matched_pattern}'. Keys used in the config "
+                    f"should be present in the dataset factory pattern."
```
merelcht, line 466:
```suggestion
                    f"Unable to resolve '{config}' from the pattern '{matched_pattern}'. Keys used in the configuration "
                    f"should be present in the dataset factory pattern."
```

kedro/io/data_catalog.py
noklam: Thanks for explaining! It make perfect sense and I agree it makes more sense with `str`.

merelcht (approved):
LGTM! 👍

noklam (approved):

Issue #2940, opened 2023-08-17
Add a section to data docs about the `metadata` attribute used by plugins


stichbury:
In working on #2888 @astrojuanlu made the comment that we needed to add some content about the `metadata` attribute (https://github.com/kedro-org/kedro/pull/2537) that can be used by plugins, such as the preview stuff for viz? We do need this in the `data_catalog.md` file that introduces settings in the catalog YAML. @merelcht says "I don't think there's any existing docs other than the part about "preview" datasets in viz. I'd suggest doing that in a separate task because it requires new content".

This is that ticket 👯‍♀️

merelcht:
I just realised there was already a ticket for this: https://github.com/kedro-org/kedro/issues/2689

astrojuanlu:
Indeed, let's close in favor of that one.

PR #2888, opened 2023-08-02
Reorganise and improve the data catalog documentation

stichbury:
> **NOTE:** Kedro datasets are moving from `kedro.extras.datasets` to a separate `kedro-datasets` package in
> [`kedro-plugins` repository](https://github.com/kedro-org/kedro-plugins). Any changes to the dataset implementations
> should be done by opening a pull request in that repository.
## Description
Created this PR in response to https://github.com/kedro-org/kedro/issues/2488

## Development notes
I've made a large set of changes to the Data Catalog documentation, which are best viewed at high-level as built docs, since the diffs are going to be impossible.

There are still some improvements to do (such as adding valid anchor links in the example page) but I am interested in feedback and help in fixing anything that is out of date now.

### Main changes
* Added a basic introduction to the index page
* Made the first main page of Data Catalog docs more comprehensive and more comprehensible. It breaks down all the elements of a YAML `catalog` to explain them. It's a Diataxis Explanation page.
* Created a page that is just a set of example YAML. 
* Split out the factories into its own page that explains them with how-to code 
* Added an advanced usage page for transcoding and `kedro.io.DataCatalog`
* Added an advanced usage page about partitioned and incremental datasets (this has had the least work)
* Brought the tutorial about creating a custom dataset from the `extend_kedro` section into the `data` section. This has also had little revision.
* Fixed the broken links throughout the docs to portions of the DataCatalog docs. 
* Temporarily removed the "Warnings->Errors" in `build-docs.sh` for the docs

## How to review

Please read through the section titled 'The Kedro Data Catalog" and comment/suggest improvements.

* [edit: updated since latest commits to reflect changes. I've built the docs to GitHub pages here](https://stichbury.github.io/data/index.html) or
* edit: this is still out of date as the RTD build is failing, so don't use this.[<s>You can review the RTD docs build](https://kedro--2888.org.readthedocs.build/en/2888/data/index.html) - Read “the Kedro Data Catalog”</s>

Both are the same. I was working with the first until I managed to get the docs building with all internal linking working. 

## Checklist

- [ ] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [ ] Added tests to cover my changes

PR #2998, opened 2023-09-01
Add hook example to access `metadata` 

### LINKED ISSUE #2676: Document datasets' metadata attribute in catalog definition
## Introduction

As of #2440 users are now able to define metadata for their datasets within the catalog. We should add a section to the hooks documentation that provides an example of how to consume this information through hooks. This should be added to this page: https://docs.kedro.org/en/stable/hooks/common_use_cases.html

Related: https://github.com/kedro-org/kedro-plugins/pull/189#issuecomment-1523740768

noklam:
## Description

fix #2676

## Checklist


noklam:
I will keep the PR simple and I created the follow up in #2999

docs/source/hooks/common_use_cases.md
```diff
@@ -200,3 +200,22 @@ HOOKS = (AzureSecretsHook(),)
...
+class MetadataHook:
+    @hook_impl
+    def after_catalog_created(
+        self,
+        catalog: DataCatalog,
+        conf_catalog: Dict[str, Any],
+        conf_creds: Dict[str, Any],
+        feed_dict: Dict[str, Any],
+        save_version: str,
+        load_versions: Dict[str, str],
```
astrojuanlu, line 217: All these parameters are not used right? Maybe we can remove them from the function. Pluggy allows this

docs/source/hooks/common_use_cases.md
```diff
@@ -200,3 +200,22 @@ HOOKS = (AzureSecretsHook(),)
 ```{note}
 Note: `DefaultAzureCredential()` is Azure's recommended approach to authorise access to data in your storage accounts. For more information, consult the [documentation about how to authenticate to Azure and authorize access to blob data](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-quickstart-blobs-python).
 ```
+
+## Use Hook to read `metadata` from `DataCatalog`
```
stichbury, line 204:
```suggestion
## Use a Hook to read `metadata` from `DataCatalog`
```

docs/source/hooks/common_use_cases.md
```diff
@@ -200,3 +200,22 @@ HOOKS = (AzureSecretsHook(),)
...
+We recommend to use the `after_catalog_created` hook if you need to access `metadata` to extend Kedro.
```
stichbury, line 205:
```suggestion
Use the `after_catalog_created` Hook to access `metadata` to extend Kedro.
```

stichbury (approved):
Approved with minor changes. What's Vale saying about this page?

docs/source/hooks/common_use_cases.md
noklam: Fair point. We do this inconsistently, sometimes we do include signature that wasn't used (I was asked to add that before, can't find it exactly which PR it is). 

For example, `after_node_run` has a redundant `node` argument.
https://docs.kedro.org/en/stable/hooks/examples.html#add-observability-to-your-pipeline

noklam: @stichbury do you think we should keep all our example minimals? We can do a clean up if this is agreed.

In this case
```
   def after_catalog_created(
        self,
        catalog: DataCatalog,
        conf_catalog: Dict[str, Any],
        conf_creds: Dict[str, Any],
        feed_dict: Dict[str, Any],
        save_version: str,
        load_versions: Dict[str, str])
```

or 
```
   def after_catalog_created(
        self,
        catalog: DataCatalog)
```

will both work.

astrojuanlu (commented):
I like this example 👍🏽 but I'm wondering, is this a full solution for https://github.com/kedro-org/kedro/issues/2676? Wondering if we should have some prose somewhere else in the data catalog documentation, otherwise I fear it might be a bit difficult to find. Essentially what @stichbury wrote in https://github.com/kedro-org/kedro/issues/2940 as a response to the reviews on https://github.com/kedro-org/kedro/pull/2888.

What do you think?

noklam:
@astrojuanlu Maybe we split the ticket too much, that should  be https://github.com/kedro-org/kedro/issues/2689 ?

astrojuanlu (approved):
Too many tickets 🥲 Thanks @noklam !

PR #3023, opened 2023-09-08
doc: fix typo return vs yield node generator

inigohidalgo:
## Description

Fixing a small error in new generator documentation.

A line read:
> Update the `make_prediction` function definition to use `return` instead of `yield`.

When it should be

> Update the `make_prediction` function definition to use `yield` instead of `return`.

## Checklist

NA: documentation change only


stichbury (approved):
Approve from docs PoV but I've no idea if it's correct technically.

stichbury:
Thanks @inigohidalgo !

noklam (approved):
🫣Thanks for picking this up.

AhdraMeraliQB:
Well spotted @inigohidalgo! Could you check the DCO failure by signing your commit message? https://github.com/kedro-org/kedro/pull/3023/checks?check_run_id=16616352181

PR #3026, opened 2023-09-11
Improve resume suggestions for SequentialRunner

### LINKED ISSUE #3002: Improve resume pipeline suggestions
## Description

Currently, when the pipeline crashes (using `SequentialRunner`), a suggestion is provided for a `--from-nodes` value to resume the pipeline using persisted inputs. However, if pipeline nodes use _parameters_ as input values, they are considered transitory (as they are represented as `MemoryDatasets`). As a result, the suggested pipelines are often a lot larger than necessary (assuming the parameters do not change between runs).

Consider the following pipeline, assuming `first_output`, `second_output` and `third_output` all defined in the catalog and `value` defined in a parameters config file:

<details><summary>Pipeline definition</summary>
<p>

```python
from typing import Dict
from kedro.pipeline import Pipeline, node

def first_func():
    return 1

def second_func(x, y):
    return x + y

def third_func(x, y):
    raise RuntimeError

    pipeline = Pipeline(
        [
            node(
                name="first_node",
                func=first_func,  # ok
                inputs={},
                outputs="first_output",
            ),
            node(
                name="second_node",
                func=second_func,  # ok
                inputs={"x": "first_output", "y": "params:value"},
                outputs="second_output",
            ),
            node(
                name="third_node",
                func=third_func,  # fails
                inputs={"x": "second_output", "y": "params:value"},
                outputs="third_output",
            ),
        ]
    )

    return {
        "__default__": pipeline,
    }
```

</p>
</details> 

The pipeline crashes at `third_node`. Because both `third_func` and `second_func` use a parameterized input value, the runner suggests rerunning both of them, suggesting re-running the pipeline with `--from-nodes "first_node"`.

Since I don't change the parameters between runs, fixing `third_func` and re-running `third_node` only should be enough to finish the full pipeline – so `--from-nodes "third_node"` would be a desired suggestion here. 

If parameters were considered persistent inputs, the suggested resumed pipelines could be smaller – unless this violates any expectations regarding pipeline params?

## Context

In my team, we use often parameterized node inputs and our pipelines can run for 1+ hours, so this can have a significant impact – especially for interactive work. I implemented a hook to generate an alternative resume suggestion, but I believe this change could be useful to other people too. I can open a PR if this is a welcome addition.

## Possible Implementation

- in `kedro.runner.runner._has_persistent_inputs`, skip evaluation of parameters.

## Possible Alternatives

If the default behavior is desired, this alternative suggestion could also be provided as an additional hook.

ondrejzacha:
## Description

Fixes #3002
(considers parameters as persistent inputs for calculation of nodes to continue crashed pipeline from)

## Development notes

This PR makes three small changes:
- It adds a special case for paramater "datasets" to be considered persistent.
- It makes the node selection algorithm less conservative, suggesting less nodes to rerun (I believe I found a small bug; in case an input – persistent or not – is shared across all nodes of the pipeline, the suggestion would be to rerun the _whole_ pipeline if _any_ node crashes).
- Adjusts related test to be more specific and cover more scenarios.

⚠️ Upon cloning (before any changes), I didn't get the tests to run successfully (neither `make test` nor `make test-no-spark`) on Python 3.9 on a Mac; hoping to see them pass on CI.

This is my first PR at this repo, let me know if I missed anything important!

## Checklist


noklam:
@ondrejzacha Thanks a lot for the PR! I am glad that someone is using this for interactive workflow and get it a try. Please let me know if you have more suggestion. IMO the session constrains is sometimes too rigid for interactive purpose, but we need to gather more feedback before we come up with any implementation.

Sorry for the slow response, at the moment the team is very busy at something else. I have a quick look at the PR. It makes sense and I agree parameters should be treated as persisted. I'd suggest not to change the original test but add a new one. It helps to understand what's failing and help prevent regression in the future.

ondrejzacha:
>  I'd suggest not to change the original test but add a new one. It helps to understand what's failing and help prevent regression in the future.

Makes sense! I kept the original test, with a comment – it adds some duplication but hopefully should help with consistency

kedro/runner/runner.py
```diff
@@ -273,25 +272,47 @@ def _enumerate_parents(pipeline: Pipeline, child: Node) -> list[Node]:
...
+        pipeline: the ``Pipeline`` to search for initial ``Node``s in.
+        nodes: the ``Node``s to find initial group for.
+
+    Returns:
+        A list of initial ``Node``s to run given inputs (in topological order).
+
+    """
+    node_names = set(n.name for n in nodes)
+    sub_pipeline = pipeline.only_nodes(*node_names)
+    initial_nodes = sub_pipeline.grouped_nodes[0]
```
merelcht, line 312: TIL this method exists!

merelcht (commented):
Thank you for opening this PR @ondrejzacha ! On first review the implementation looks good. I do need some more time to think about this though. I feel like the assumption that `parameters` are not changed makes sense, but I wonder if there are cases where actually they do change and what consequences this change would have for those types of pipelines. 

I'd be interested to hear what @idanov and @deepyaman have to say about this, because they've known Kedro for longer.

ondrejzacha:
Hi Kedro team, this PR has been open for a while; is this something you'd like to see merged? My team has successfully been using this adjustment (in the form of a hook) and it saved us hours of pipeline runtime – but I understand this may potentially negatively affect other use-cases. Let me know if there's interest, I'm happy to adjust/add more tests if needed

noklam:
@ondrejzacha I personally think this is a good change, and is a step towards more efficient pipeline re-run beyond a single session run. I do find that the leaky definition of `persistent_input` is a bit weird and as a result in this PR you have to use something like `impersistent`.

kedro/runner/runner.py
```diff
@@ -242,54 +241,86 @@ def _find_persistent_ancestors(
...
+
+    """
+    # We use _data_sets because they pertain parameter name format
+    catalog_datasets = catalog._datasets
+    missing_inputs: set[str] = set()
+    for node_input in node.inputs:
+        # Important difference vs. Kedro approach
+        if node_input.startswith("params:"):
+            continue
+        if isinstance(catalog_datasets[node_input], MemoryDataset):
```
noklam, line 284: Can we use `catalog.load` instead of `catalog._datasets`? If i understand you want to perserve `params:` and `catalog.load` should work

merelcht: Is it necessary to load the dataset though? Isn't that a bit of overkill?

merelcht (commented):
Apologies on the delay of addressing your PR @ondrejzacha. There's been a lot going on  and it slipped my mind. Let's try get it merged soon now 🙂 

I've left some small comments and questions. But also wanted to discuss what @noklam said: 

> I do find that the leaky definition of persistent_input is a bit weird and as a result in this PR you have to use something like impersistent.

@noklam Do you mean here you don't like the terminology of the word `impersistent` or do you think the implementation itself is leaky and should be changed?

If it's about terminology, I do agree that `impersistent` is a tricky word. It's of course the correct opposite of `persistent`, but for a non-native speaker (me) it makes it harder to read the code for some reason. Maybe just call it `non_persistent`? 😅  In any case I think it would be helpful to clarify what we mean with `impersistent`/`non_persistent` in the context of Kedro. From reading the code (specifically `_enumerate_impersistent_inputs`), I interpret the definition being anything that is a `MemoryDataset`. Even if it's defined in the catalog.

ondrejzacha:
Hi both @noklam, @merelcht – thanks for the renewed interest for this PR! I wasn't sure if there are still intentions to merge it so I haven't pushed any changes here, but I did experiment a bit on a side and came up with what I believe is a better solution here: [see diff here.](https://github.com/kedro-org/kedro/compare/main...ondrejzacha:kedro:improve-resume-suggestions-v2?expand=1)

It's a bit more elaborate approach which separates the logic into a separate function. On one hand it's a larger change, on the other it's easier to test. Overall it adds these benefits:
- if dataset (or param) is persistent & shared, it won't keep looking for ancestors
- it only looks for ancestors producing _impersistent_ inputs
- it minimizes the number of suggested nodes (= shorter message for the same pipeline)
- it's testable (therefore more applicable to other runners too)

Do you think this is a good direction? I could also create a new PR and make sure all commits are properly signed etc.

kedro/runner/runner.py
```diff
@@ -242,54 +241,86 @@ def _find_persistent_ancestors(
...
+    Args:
+        node: the ``Node`` to check the inputs of.
+        catalog: the ``DataCatalog`` of the run.
 
+    Returns:
+        Set of names of impersistent inputs of given ``Node``.
+
+    """
+    # We use _data_sets because they pertain parameter name format
+    catalog_datasets = catalog._datasets
```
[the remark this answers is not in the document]
noklam: > @noklam Do you mean here you don't like the terminology of the word impersistent or do you think the implementation itself is leaky and should be changed?

> If it's about terminology, I do agree that impersistent is a tricky word. It's of course the correct opposite of persistent, but for a non-native speaker (me) it makes it harder to read the code for some reason. Maybe just call it non_persistent? 😅 In any case I think it would be helpful to clarify what we mean with impersistent/non_persistent in the context of Kedro. From reading the code (specifically _enumerate_impersistent_inputs), I interpret the definition being anything that is a MemoryDataset. Even if it's defined in the catalog.

@merelcht Mostly terminology, it's just hard to remember as previously it's just MemoryDataset, now is basically memoryDataset + params.

New thought: https://github.com/kedro-org/kedro/pull/3520 - any chance we can utilise the new flag that we added? Can we load `parameters` as a "special" kind of MemoryDataset that will have `self._EPHEMERAL = False`? We can take this PR first and refactor this afterward if needed.

merelcht:
Oh amazing, thanks @ondrejzacha ! The approach definitely sounds very well thought out 🙂  I'll need a bit of time to review, but I'll prioritise this for next week. Thanks for your patience 🙏

kedro/runner/runner.py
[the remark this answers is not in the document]
ondrejzacha: `catalog.datasets` members need to be accessible as attributes, so colons and periods are replaced with `__`.  I agree this is very far from an elegant solution, so I'm definitely open to suggestions

kedro/runner/runner.py
ondrejzacha: In this approach I need to check the Dataset instance, not exactly its data contents

ondrejzacha:
Opened a new PR here: #3719

PR #3036, opened 2023-09-15
Add "runtime_params" resolver to allow overriding of config with CLI params

### LINKED ISSUE #2531: Enable overriding configuration via variable interpolation for all configurations other than just `parameters` for OmegaConfigLoader

ankatiyar:
## Description

Resolve #2531 

### `runtime_params` resolver
- similar to globals - registered using `_registed_runtime_params_resolver()`
- Uses `_get_runtime_value()` to get the interpolated value, also similar to globals' `_get_globals_value()`
### Changes to `globals`
- Read globals and populate `self._globals` in the `__init__` function. 
- When global values are read by `_get_globals_value()` it uses the `self._globals` dict. 
- This is so when globals are first read, the `runtime_params` resolver is ***cleared*** and then ***registered again*** in the `__getitem__` fn
- Also helps that we don't need to read `self["globals"]` each time to resolve wherever globals is used.
### On the topic of nested resolvers
***Need feedback from reviewers on this***
- I disable the use of `runtime_params` in `globals.yml` by disabling the `runtime_params` in `__getitem__` if the key is "globals" and then registering it again. Throws an `UnsupportedInterpolationType` error when `runtime_params:` is used in globals.
- Disabling globals within `runtime_params` is tricky, I haven't found a clean way to do it.

However, I think allowing globals to be used as a default value for parameters with `runtime_params` resolver is valid and might be useful and we probably should allow this - 
```yaml
my_param : "${runtime_params: y, ${globals: x, 34}}"
```

### TODO

- [x] Test
- [x] Docs
- [x] Release Notes

## Checklist


kedro/config/omegaconf_config.py
```diff
@@ -178,9 +186,15 @@ def __getitem__(self, key) -> dict[str, Any]:
...
-            base_path, patterns, key, processed_files, read_environment_variables
-        )
+        try:
+            base_config = self.load_and_merge_dir_config(
+                base_path, patterns, key, processed_files, read_environment_variables
+            )
+        except UnsupportedInterpolationType:
+            raise UnsupportedInterpolationType(
+                "The `runtime_params:` resolver is not supported for globals."
+            )
```
merelcht, line 196: This could potentially be raised because of something else right? So we should make sure that this specific message only shows when the user is trying to use `runtime_params` with globals.

merelcht (commented):
This looks great @ankatiyar ! I have also tried some simple use cases and it works perfectly 👍  I agree that it's should be allowed for users to use `globals` as a default inside the `runtime_params` resolver. But definitely the other way around doesn't make sense. 

Are you adding docs as part of this PR or separately?

kedro/config/omegaconf_config.py
ankatiyar: I've updated the condition to only show this message if `runtime_params` is used.

RELEASE.md
```diff
@@ -11,6 +11,7 @@
 
 ## Major features and improvements
 * Allowed using of custom cookiecutter templates for creating pipelines with `--template` flag for `kedro pipeline create` or via `template/pipeline` folder.
+* Allowed overriding of configuration keys with runtime parameters using the `runtime_params` resolver.
```
merelcht, line 14: It would be good to still specify this is for the `OmegaConfigLoader` since the other two config loaders are still around.

merelcht (approved):
This is great! I think the docs are very clear too 👍

docs/source/configuration/advanced_configuration.md
```diff
@@ -300,9 +296,31 @@ You can also provide a default value to be used in case the global variable does
 ```yaml
 my_param: "${globals: nonexistent_global, 23}"
 ```
-If there are duplicate keys in the globals files in your base and run time environments, the values in the run time environment
+If there are duplicate keys in the globals files in your base and runtime environments, the values in the runtime environment
 will overwrite the values in your base environment.
```
stichbury, line 300:
```suggestion
overwrite the values in your base environment.
```

docs/source/configuration/advanced_configuration.md
```diff
@@ -300,9 +296,31 @@ You can also provide a default value to be used in case the global variable does
...
 
+### How to override configuration with runtime parameters with the `OmegaConfigLoader`
+
+Kedro allows you to [specify runtime parameters for the `kedro run` command with the `--params` CLI option](parameters.md#how-to-specify-parameters-at-runtime). These runtime parameters
+are added to the `KedroContext` and merged with parameters from the configuration files to be used in your project's pipelines and nodes. From Kedro `0.18.14`, you can use the
+`runtime_params` resolver to indicate that you want to override values of certain keys in your configuration with runtime parameters provided through the CLI option.
+This resolver can be used across different configuration types, such as parameters, catalog, and more, except for "globals".
+Consider this `catalog.yml` file :
```
stichbury, line 308:
```suggestion

Consider this `catalog.yml` file:
```

stichbury (approved):
This looks good to me, thanks @ankatiyar!

I made a commit to the branch to move the table of contents that links to the "How to " subsections to the top of the page (as there was some repetition with it in the place previously sited).

RELEASE.md
ankatiyar, line 14:
```suggestion
* Allowed overriding of configuration keys with runtime parameters using the `runtime_params` resolver with `OmegaConfigLoader`.
```

docs/source/configuration/advanced_configuration.md
```diff
@@ -300,9 +296,32 @@ You can also provide a default value to be used in case the global variable does
...
+This resolver can be used across different configuration types, such as parameters, catalog, and more, except for "globals".
+
+Consider this `catalog.yml` file:
+```yaml
+model_options:
+  random_state: "${runtime_params:random}"
+```
+This will allow you to pass a runtime parameter named `random` through the CLI to specify the value of `model_options.random_state` in your project's parameters:
+```bash
+kedro run --params="random=3"
```
noklam, line 316:
```suggestion
kedro run --params random=3
```
Nit-pick

docs/source/configuration/advanced_configuration.md
```diff
@@ -300,9 +296,32 @@ You can also provide a default value to be used in case the global variable does
...
+overwrite the values in your base environment.
+
+### How to override configuration with runtime parameters with the `OmegaConfigLoader`
 
+Kedro allows you to [specify runtime parameters for the `kedro run` command with the `--params` CLI option](parameters.md#how-to-specify-parameters-at-runtime). These runtime parameters
+are added to the `KedroContext` and merged with parameters from the configuration files to be used in your project's pipelines and nodes. From Kedro `0.18.14`, you can use the
+`runtime_params` resolver to indicate that you want to override values of certain keys in your configuration with runtime parameters provided through the CLI option.
+This resolver can be used across different configuration types, such as parameters, catalog, and more, except for "globals".
+
+Consider this `catalog.yml` file:
```
noklam, line 309:
```suggestion
Consider this `parameters.yml` file:
```

kedro/config/omegaconf_config.py
```diff
@@ -170,6 +175,9 @@ def __getitem__(self, key) -> dict[str, Any]:
             )
         patterns = [*self.config_patterns[key]]
 
+        if key == "globals":
+            # "runtime_params" resolver is not allowed in globals.
+            OmegaConf.clear_resolver("runtime_params")
```
noklam, line 180: Does that means running `config_loader["global"]` twice could get different results? User may use `config_loader["global"]` explicitly to perform logging etc. https://kedro-org.slack.com/archives/C03RKP2LW64/p1695380623784409

kedro/config/omegaconf_config.py
```diff
@@ -339,6 +375,19 @@ def _get_globals_value(self, variable, default_value=_NO_VALUE):
...
+            runtime_oc, variable, default=default_value
+        )
+        if interpolated_value != _NO_VALUE:
+            return interpolated_value
+        else:
+            raise InterpolationResolutionError(
+                f"Runtime parameter '{variable}' not found and no default value provided."
+            )
+
     @staticmethod
```
noklam, line 391: 👍🏼

tests/config/test_omegaconf_config.py
```diff
@@ -860,7 +860,7 @@ def test_not_hidden_config(self, tmp_path, hidden_path):
         assert not conf._is_hidden(hidden_path)
 
     def test_ignore_ipynb_checkpoints(self, tmp_path, mocker):
-        conf = OmegaConfigLoader(str(tmp_path), default_run_env=_BASE_ENV)
+        conf = OmegaConfigLoader(str(tmp_path), default_run_env="base")
```
noklam, line 863: Why do you need to change it? isn't the default value `base` already?

kedro/config/omegaconf_config.py
```diff
@@ -189,9 +206,19 @@ def __getitem__(self, key) -> dict[str, Any]:
...
+            )
+        except UnsupportedInterpolationType as exc:
+            if "runtime_params" in str(exc):
+                raise UnsupportedInterpolationType(
+                    "The `runtime_params:` resolver is not supported for globals."
+                )
+            else:
+                raise exc
+        # Re-register runtime params resolver incase it was deactivated
+        self._register_runtime_params_resolver()
```
noklam, line 221: I find this a bit hard to follow - I am not sure if this is registered or not. If I follow the logic, it assumes that it is always on and only turn off if key==global. Maybe this should move to the beginning instead? Then the `__init__` register_runtime_params_resolvers can be removed optionally.

In fact, I found this is inconsistent and I think we should treat this same as the `read_environment_variable` flag.

kedro/config/omegaconf_config.py
ankatiyar: I don't think running `config_loader["globals"]` would return anything different each time, it just turns off the `runtime_params` resolver.

kedro/config/omegaconf_config.py
ankatiyar: Since `replace=True` is set in `_register_runtime_params_resolver()`, I didn't add an if condition here. It'll re-register it anyway.

tests/config/test_omegaconf_config.py
ankatiyar: Accidentally left this in, reverted.

kedro/config/omegaconf_config.py
ankatiyar: I've moved it to the beginning and removed the call from `__init__`.

merelcht (approved):
Btw the docs failure for the databricks was fixed in `develop` with https://docs.databricks.com/en/dev-tools/cli/authentication.html

kedro/config/omegaconf_config.py
```diff
@@ -141,8 +141,17 @@ def __init__(  # noqa: too-many-arguments
...
+        try:
+            self._globals = self["globals"]
+        except MissingConfigException:
+            self._globals = {}
 
-    def __getitem__(self, key) -> dict[str, Any]:
+    def __setitem__(self, key, value):
+        if key == "globals":
+            self._globals = value
+        super().__setitem__(key, value)
```
noklam, line 152: I think we should have a comment explaining why we treat global differently.

noklam (approved):
Approved with minor comments. It's a great feature to add!

kedro/config/omegaconf_config.py
ankatiyar, line 152:
```suggestion
    def __setitem__(self, key, value):
        if key == "globals":
        # Update the cached value at self._globals since it is used by the globals resolver
            self._globals = value
        super().__setitem__(key, value)
```

PR #2987, opened 2023-08-30
MVP for "add-ons" flow within kedro new CLI command

SajidAlamQB:
> **NOTE:** Kedro datasets are moving from `kedro.extras.datasets` to a separate `kedro-datasets` package in
> [`kedro-plugins` repository](https://github.com/kedro-org/kedro-plugins). Any changes to the dataset implementations
> should be done by opening a pull request in that repository.
## Notes for reviewers - @AhdraMeraliQB 

Thank you for taking the time to review almost 600 new lines of code - we hope you enjoy your stay.

**Just a few notes before diving right in:**

##### Implementation

`kedro/templates/project/hooks/post_gen_project.py`: The code run by `cookiecutter` after the project template is generated.

`kedro/templates/project/hooks/utils.py`: Functions called by `post_gen_project.py` to setup the add-ons. Includes parsing user input, tear-down of unnecessary files, injection of necessary requirements, and alphabetic sorting of requirements.

`kedro/templates/project/prompts.yml`: The new add-on related prompt output when running `kedro new` with no starter and no config path.

`kedro/framework/cli/starters.py`: Additions to the output generated after project creation and regex validation for add-ons passed through from config

`kedro/templates/project/{{ cookiecutter.repo_name }}/pyproject.toml`, `kedro/templates/project/{{ cookiecutter.repo_name }}/src/pyproject.toml`, and `kedro/templates/project/{{ cookiecutter.repo_name }}/src/requirements.txt`: Removal of testing, linting, and documentation requirements from the base template.

`kedro/templates/project/cookiecutter.json`: Updated to include add-ons

`kedro/templates/project/{{ cookiecutter.repo_name }}/conf/logging.yml`: The return of the logging config

`pyproject.toml`: A fix to a somewhat recurring issue; see [this stackoverflow post](https://stackoverflow.com/questions/32160439/preventing-python-coverage-from-including-virtual-environment-site-packages)

##### Tests

`tests/framework/cli/test_starters.py`: Modified old tests to work with new flow, added requirements check helper function, and added tests to check success of including add-ons via CLI prompt and config file. 

`features/*`: Any tests that make use of `kedro new` or any processes that are now part of add-ons have been updated in accordance to the new flow

#### 3. If you've got the time, try and break it! After checking out this PR and installing the local version of Kedro, testing the changes is as simple as running `kedro new`. This is what it should look like if you do:

```
(kedrotest) a/path/on/your/computer % kedro new

Project Add-Ons
===============
[...]

PR #3081, opened 2023-09-25
Add addon flags to kedro new

lrcouto:
## Description

Part of the add-ons flow work. Adds an `--addons` flag to `kedro new` that allows the used to select which addons they would like to have on their new project through the CLI.

Examples:

`kedro new --addons=lint,test,log,docs,data`
`kedro new --addons=all`
`kedro new --addons=none`

## Checklist


kedro/framework/cli/starters.py
```diff
@@ -93,6 +93,24 @@ class KedroStarterSpec:  # noqa: too-few-public-methods
     "An optional directory inside the repository where the starter resides."
 )
 
+# TODO; Insert actual link to the documentation.
```
stichbury, line 96: Please don't forget this :)

kedro/framework/cli/starters.py
```diff
@@ -93,6 +93,24 @@ class KedroStarterSpec:  # noqa: too-few-public-methods
...
+ADDON_ARG_HELP = """
+Select which add-ons you'd like to include. By default, none are included.
+To read more about these add-ons and what they do visit: kedro.org/{insert-documentation}
```
stichbury, line 99:
```suggestion
To found out more about these add-ons visit: kedro.org/{insert-documentation}
```

stichbury: Is it possible to have it like this? It is better because it's got the action first that the reader needs to take.

kedro/framework/cli/starters.py
```diff
@@ -93,6 +93,24 @@ class KedroStarterSpec:  # noqa: too-few-public-methods
...
+
+Add-Ons
+1) Linting : Provides a basic linting set up with Black and ruff
```
stichbury, line 102:
```suggestion
1) Linting : Provides a basic linting setup with Black and Ruff
```

kedro/framework/cli/starters.py
```diff
@@ -93,6 +93,24 @@ class KedroStarterSpec:  # noqa: too-few-public-methods
...
+2) Testing : Provides basic testing set up with pytest
```
stichbury, line 103:
```suggestion
2) Testing : Provides basic testing setup with pytest
```

kedro/framework/cli/starters.py
stichbury: I think you can eliminate the "provides" on each line.

kedro/framework/cli/starters.py
```diff
@@ -93,6 +93,24 @@ class KedroStarterSpec:  # noqa: too-few-public-methods
...
+3) Custom Logging : Provides more logging options
+4) Documentation: Provides basic documentations setup with Sphinx
```
stichbury, line 105:
```suggestion
4) Documentation: Basic documentation setup with Sphinx
```

stichbury (approved):
Nice! Just a few suggested changes ⭐

kedro/framework/cli/starters.py
```diff
@@ -93,6 +93,24 @@ class KedroStarterSpec:  # noqa: too-few-public-methods
...
+
+Add-Ons
+1) Linting : Provides a basic linting setup with Black and Ruff
+2) Testing : Provides basic testing setup with pytest
+3) Custom Logging : Provides more logging options
+4) Documentation: Basic documentation setup with Sphinx
+5) Data Structure: Provides a directory structure for storing data
+
+Example usage:
+kedro new --addons=lint,test,log,docs,data (or any subset of these options)
```
AhdraMeraliQB, line 109: Wouldn't we be specifying the number selection here instead?

merelcht: @amandakys designed it as implemented, so not with numbers. But it would be good to get her confirmation!

AhdraMeraliQB: Worried about the inconsistency between specifying by flag (through add-on name) and through the prompting (through add-on number), maybe we can look at changing the latter post-#2987?

merelcht: The numbers will be very hard to remember though, so from a usability point of view I prefer the way it's done now. But as I said, this is definitely a design detail, so @amandakys has the final say 🙂

kedro/framework/cli/starters.py
stichbury: Is there a reason this can't be done before merging?

kedro/framework/cli/starters.py
```diff
@@ -93,6 +93,24 @@ class KedroStarterSpec:  # noqa: too-few-public-methods
...
+kedro new --addons=all
+kedro new --addons=none
```
merelcht, line 111: To have this print like this when you do `kedro new --help` you'll need to add `\n` at the end of every line to get a new line in between, e.g:

```
Add-Ons \n
1) Linting: Provides a basic linting setup with Black and Ruff \n
2) Testing: Provides basic testing setup with pytest \n
``` 

etc..

merelcht (commented):
I tried the code, but found that only `kedro new --addons=all` or `kedro new --addons=none` works. If I enter any of the add ons using the names e.g. `lint`, `test` etc. I get an error. This actually makes sense when looking at your test `def test_valid_add_ons(self, fake_kedro_cli, add_ons):` because there you use the numbers as entry.

kedro/framework/cli/starters.py
lrcouto: For this text block, I've tried adding the line breaks before, and the result is a bit strange. It adds a larger break between lines, like this:

image

I've also tried using `\b` instead of `\n`, and formatting it with methods like `click.wrap_text()`, but none of that gives me the formatting I actually want. I'll leave it with the `\n`s for now because it's easier to read.

tests/framework/cli/test_starters.py
```diff
@@ -933,3 +951,40 @@ def test_invalid_add_ons(self, fake_kedro_cli):
...
+    )
+    def test_valid_add_ons(self, fake_kedro_cli, add_ons):
+        result = CliRunner().invoke(
+            fake_kedro_cli,
+            ["new", "--addons", add_ons],
+            input=_make_cli_prompt_input_without_addons(),
+        )
+        add_ons = _convert_addon_names_to_numbers(selected_addons=add_ons)
+        _assert_template_ok(result, add_ons=add_ons)
+        _assert_requirements_ok(result, add_ons=add_ons)
```
merelcht, line 981:
```suggestion
        _assert_requirements_ok(result, add_ons=add_ons)
        _clean_up_project(Path("./new-kedro-project"))
```

You need to clean up the project after creation to fix the coverage issues.

merelcht (commented):
The `addons` flag now works well if you provide valid add-ons, however if you try something like `kedro new --addons=random` it will still prompt for the project name and then you get an error like:
```
(dev) M-C02XN0V5JHD5:Testing merel_theisen$ kedro new --addons=random

Project Name
============
Please enter a human readable name for your new project.
Spaces, hyphens, and underscores are allowed.
 (New Kedro Project): bla
  File "/var/folders/_3/kzkd3jjj31s24wjnsg7qgnk80000gn/T/tmp02ezo4dy.py", line 27, in <module>
    main()
  File "/var/folders/_3/kzkd3jjj31s24wjnsg7qgnk80000gn/T/tmp02ezo4dy.py", line 18, in main
    selected_add_ons_list = parse_add_ons_input(selected_add_ons)
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/merel_theisen/Projects/kedro/kedro/templates/project/hooks/utils.py", line 90, in parse_add_ons_input
    _validate_selection(selected)
  File "/Users/merel_theisen/Projects/kedro/kedro/templates/project/hooks/utils.py", line 58, in _validate_selection
    if int(add_on) < 1 or int(add_on) > 5:
       ^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: 'random'
[10/19/23 13:26:09] ERROR    Stopping generation because post_gen_project hook script didn't exit         hooks.py:159
                             successfully
kedro.framework.cli.utils.KedroCliError: Failed to generate project when running cookiecutter.
Run with --verbose to see the full exception
Error: Failed to generate project when running cookiecutter.
```

Ideally, the add on input gets validated before it promts for the project name and the message shouldn't be `ValueError: invalid literal for int() with base 10: 'random'` but more something along the lines of "`random` isn't a valid option for `addons`, the possible options are ..."

SajidAlamQB (commented):
Thank you @lrcouto, I've tested it manually and it looks great and on the right track!

There were a few things I noticed though. When we separate the add-ons with spaces after each comma the command fails. i.e `lint,test,log` works but not `lint, test, log` maybe we need some logic that sanitises the whitespace from inputs.

Also I noticed if we do `lint,all` we get unhelpful/confusing error:

```
ValueError: invalid literal for int() with base 10: 'all'
[10/23/23 16:14:58] ERROR    Stopping generation because post_gen_project hook script didn't exit successfully                  generate.py:256
kedro.framework.cli.utils.KedroCliError: Failed to generate project when running cookiecutter.
Run with --verbose to see the full exception
Error: Failed to generate project when running cookiecutter.
```

We might want a more clear error message telling users they have to select between a combination of add ons or none/all.

merelcht (approved):
I've tested this again after your changes and I think this looks really good 👏 . I can't think of any other edge cases at the moment. I would add this change to the release notes for visibility of the feature ✍️ 

@SajidAlamQB you made a very good point about the spaces. The way it's implemented now is consistent with e.g. `kedro run --tags=....` spaces can only be "allowed" if they are inside a string so: `kedro new --addons="test, docs, log"`, but not like `kedro new --addons=test, docs`.

lrcouto:
> I've tested this again after your changes and I think this looks really good 👏 . I can't think of any other edge cases at the moment. I would add this change to the release notes for visibility of the feature ✍️
> 
> @SajidAlamQB you made a very good point about the spaces. The way it's implemented now is consistent with e.g. `kedro run --tags=....` spaces can only be "allowed" if they are inside a string so: `kedro new --addons="test, docs, log"`, but not like `kedro new --addons=test, docs`.

The way it is right now, trying to pass them outside of quotation marks will return an "unexpected argument" error. Using them inside quotation marks as in `kedro new --addons="lint, test, log"` should work.

lrcouto:
> I confirm @SajidAlamQB 's case above still raises an unhelpful error:
> 
> ```
> ❯ kedro new --addons=lint,all                                                                       (kedro38-dev) 
> 
> Project Name
> ============
> Please enter a human readable name for your new project.
> Spaces, hyphens, and underscores are allowed.
>  [New Kedro Project]: test-spaceflights
> Traceback (most recent call last):
>   File "/var/folders/r7/ywj0_kvj0mxfkdkx0jrgh73r0000gn/T/tmphs49kn1j.py", line 28, in <module>
>     main()
>   File "/var/folders/r7/ywj0_kvj0mxfkdkx0jrgh73r0000gn/T/tmphs49kn1j.py", line 19, in main
>     selected_add_ons_list = parse_add_ons_input(selected_add_ons)
>   File "/Users/juan_cano/Projects/QuantumBlack Labs/kedro/kedro/templates/project/hooks/utils.py", line 91, in parse_add_ons_input
>     _validate_selection(selected)
>   File "/Users/juan_cano/Projects/QuantumBlack Labs/kedro/kedro/templates/project/hooks/utils.py", line 59, in _validate_selection
>     if int(add_on) < 1 or int(add_on) > 5:
> ValueError: invalid literal for int() with base 10: 'lint'
> ```

When I try to reproduce this, I get the folllwing result:

image

astrojuanlu:
Turns out I was using an old commit 🤦🏽 All good!

astrojuanlu:
Only missing bit then is adding this to the release notes

PR #3085, opened 2023-09-26
Added new docs section that warns against using dot notation for pipe…

MehdiNV:
## Description
Related to https://github.com/kedro-org/kedro-viz/issues/1522 

There is no warning given to users, in code or docs, from my understanding that advises them to avoid using `.` notation (which ordinarily indicates namespace pipeline). If `.` is used, it leads to disconnected and erratic behaving Kedro structures.

## Development notes

- Added text in the md file, advising users to not add `.` when creating pipelines (and alternatively suggesting `_`)

## QA notes

Documentation website should now have a new section at the bottom, warning users against creating pipelines w/ dot notation.

### DIFF
--- a/docs/source/nodes_and_pipelines/pipeline_introduction.md
+++ b/docs/source/nodes_and_pipelines/pipeline_introduction.md
@@ -249,5 +249,6 @@ pipeline([node(lambda x: x, inputs: "input1", outputs: "outputs1.kedro")])
 
 The output tends to be a disconnected pipeline and / or improperly formatted Kedro structure. 
 
-This is because `.` has a special meaning and indicates a namespace pipeline; here, the outputs segment should be disconnected as Kedro will think that there is a "outputs1" namespace pipeline.
+This is because `.` has a special meaning internally and indicates a namespace pipeline; here, the outputs segment should be disconnected as Kedro will think that there is a "outputs1" namespace pipeline.
 
+Alternatively, consider using characters like `_` - similar visual effect, without the downside. 
--- a/docs/source/nodes_and_pipelines/pipeline_introduction.md
+++ b/docs/source/nodes_and_pipelines/pipeline_introduction.md
@@ -244,10 +244,10 @@ Nodes named with dot notation may behave strangely.
 <summary><b>Click to expand</b></summary>
 
 ```python
-pipeline([node(lambda x: x, inputs = "input1kedro", outputs = "output1.kedro")])
+pipeline([node(lambda x: x, inputs="input1kedro", outputs="output1.kedro")])
 ```
 
-Nodes that are created with input or output names that contain this character risk a disconnected pipeline and/or improperly formatted Kedro structure. 
+Nodes that are created with input or output names that contain this character risk a disconnected pipeline and/or improperly formatted Kedro structure.
 
 This is because `.` has a special meaning internally and indicates a namespace pipeline. In the example, the outputs segment should be disconnected as the name implies there is an "output1" namespace pipeline. The input is not namespaced, but the output is via its dot notation. This leads to Kedro processing each separately. For this example, a better approach would've been writing both as `input1_kedro` and `output1_kedro`.
 
--- a/docs/source/nodes_and_pipelines/pipeline_introduction.md
+++ b/docs/source/nodes_and_pipelines/pipeline_introduction.md
@@ -247,7 +247,7 @@ Nodes named with dot notation may behave strangely.
 pipeline([node(lambda x: x, inputs="input1kedro", outputs="output1.kedro")])
 ```
 
-Nodes that are created with input or output names that contain this character risk a disconnected pipeline and/or improperly formatted Kedro structure.
+Nodes that are created with input or output names that contain `.` risk a disconnected pipeline and/or improperly formatted Kedro structure.
 
 This is because `.` has a special meaning internally and indicates a namespace pipeline. In the example, the outputs segment should be disconnected as the name implies there is an "output1" namespace pipeline. The input is not namespaced, but the output is via its dot notation. This leads to Kedro processing each separately. For this example, a better approach would've been writing both as `input1_kedro` and `output1_kedro`.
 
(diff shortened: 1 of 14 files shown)


docs/source/nodes_and_pipelines/pipeline_introduction.md
```diff
@@ -236,3 +236,19 @@ The output is as follows:
 Circular dependencies exist among these items: ['first node: <lambda>([x]) -> [y]', 'second node: <lambda>([y]) -> [x]']
 ```
 </details>
+
+### Pipeline with improperly named nodes
+Nodes named with improper characters, such as the dot notation, may exhibit strange behaviour than intended.
```
stichbury, line 241:
```suggestion
Nodes named with improper characters, such as the dot notation, may behave strangely.
```

docs/source/nodes_and_pipelines/pipeline_introduction.md
```diff
@@ -236,3 +236,19 @@ The output is as follows:
...
+
+<details>
+<summary><b>Click to expand</b></summary>
+
+```python
+pipeline([node(lambda x: x, inputs: "input1", outputs: "outputs1.kedro")])
+```
+
+The output tends to be a disconnected pipeline and / or improperly formatted Kedro structure.
```
stichbury, line 250:
```suggestion
The output tends to be a disconnected pipeline and/or improperly formatted Kedro structure. 
```

stichbury: Is the actual node name "outputs1.kedro" or is that the name of the output of a node?

stichbury: Cos it would be better to be precise if it's the latter and say something like "Nodes that are created with input or output names that contain improper characters..." above if the latter.

docs/source/nodes_and_pipelines/pipeline_introduction.md
```diff
@@ -236,3 +236,19 @@ The output is as follows:
...
+<details>
+<summary><b>Click to expand</b></summary>
+
+```python
+pipeline([node(lambda x: x, inputs: "input1", outputs: "outputs1.kedro")])
+```
+
+The output tends to be a disconnected pipeline and / or improperly formatted Kedro structure. 
+
+This is because `.` has a special meaning internally and indicates a namespace pipeline; here, the outputs segment should be disconnected as Kedro will think that there is a "outputs1" namespace pipeline.
```
stichbury, line 252:
```suggestion
This is because `.` has a special meaning internally and indicates a namespace pipeline. In the example, the outputs segment should be disconnected as the name implies there is a "outputs1" namespace pipeline.
```

stichbury: What does "outputs segment should be disconnected" mean? What should the reader do? I think this needs to be clearer.

docs/source/nodes_and_pipelines/pipeline_introduction.md
```diff
@@ -236,3 +236,19 @@ The output is as follows:
...
+
+Alternatively, consider using characters like `_` - similar visual effect, without the downside.
```
stichbury, line 254:
```suggestion
We recommend use of characters like `_` instead of `.` as name separators. 
```

stichbury (changes requested):
Looks good to me but a few minor changes requested

MehdiNV:
> Looks good to me but a few minor changes requested

Good call! Have just committed new changes up

It's only the dot notation that breaks Kedro, other characters like ()!@£ etc should be completely fine. Is it better to be specific and say "_Don't use dot notation_" or keep it a bit expansive like it is now and say "_Don't use improper characters like dot notation_"?

docs/source/nodes_and_pipelines/pipeline_introduction.md
```diff
@@ -236,3 +236,19 @@ The output is as follows:
...
 </details>
+
+### Pipeline with improperly named nodes
+Nodes named with improper characters, such as the dot notation, may behave strangely.
+
+<details>
+<summary><b>Click to expand</b></summary>
+
+```python
+pipeline([node(lambda x: x, inputs: "input1kedro", outputs: "outputs1.kedro")])
```
astrojuanlu, line 247:
```suggestion
pipeline([node(lambda x: x, inputs="input1kedro", outputs="outputs1.kedro")])
```

?

astrojuanlu (commented):
Wondering if this could be made shorter and turned into a warning instead of a whole section of its own, something like

`````
````warning
Kedro treats dots `.` in input and output names as namespace separators. If you do not want to use namespaces, avoid the use of dots in your input and output names, and use underscores `_` or other separators instead.

Example:

```python
pipeline([
  node(
    func=lambda x: x,
    inputs="input1kedro",
    outputs="outputs1.kedro",  # Will assume a namespace that doesn't exist!
  )
])
```
````
`````

**Comment by astrojuanlu** [comment]
Maybe put it under a "Troubleshooting" section then?

**Comment by MehdiNV** [comment]
> Maybe put it under a "Troubleshooting" section then?

@stichbury @astrojuanlu I'm personally fine w/ either, is there a preference? Docs is not my strong suit so not sure which way to personally lean myself

I assume with the Troubleshooting approach, we'd just create a new page named 'Troubleshooting' and move some of these texts there instead?

**Comment by stichbury** [comment]
> I assume with the Troubleshooting approach, we'd just create a new page named 'Troubleshooting' and move some of these texts there instead?

Probably not, we don't need a page with a generic name because readers won't know what it contains. It's fine as it is -- I don't think we need to rename the section you've included it under. I'm happy with this now, so approving.

**Review by stichbury — APPROVED** [review]

File: docs/source/nodes_and_pipelines/pipeline_introduction.md [hunk 6]
```diff
@@ -236,3 +236,19 @@ The output is as follows:
...
+Nodes named with dot notation may behave strangely.
+
+<details>
+<summary><b>Click to expand</b></summary>
+
+```python
+pipeline([node(lambda x: x, inputs="input1kedro", outputs="output1.kedro")])
+```
+
+Nodes that are created with input or output names that contain this character risk a disconnected pipeline and/or improperly formatted Kedro structure.
```
stichbury, line 250:
```suggestion
Nodes that are created with input or output names that contain `.` risk a disconnected pipeline and/or improperly formatted Kedro structure.
```

astrojuanlu (approved):

Issue #3042, opened 2023-09-18
Subclass `KedroDeprecationWarning` to ensure dataset deprecation warnings show up


merelcht:
### Description

Follow up on [Replace "DataSet" with "Dataset" in Kedro-Datasets](https://github.com/kedro-org/kedro/issues/2751). Subclass `DeprecationsWarning` to show them by default in Kedro.

### Context

Currently `DeprecationWarning`s are not shown unless the user enables `default::DeprecationWarning`. It's important for users to see the deprecation warnings about the datasets that are being renamed. 

It seems to also be the approach taken by libraries like SymPy (first link I found while searching was https://docs.sympy.org/latest/modules/utilities/exceptions.html#:~:text=The%20SymPyDeprecationWarning%20class%20is%20a,a%20link%20to%20the%20documentation.)

PR #3095, opened 2023-09-28
Implement and use custom `KedroDeprecationWarning`

deepyaman:
## Description

Partially addresses https://github.com/kedro-org/kedro/issues/3042

## Checklist


kedro/__init__.py
```diff
@@ -9,11 +9,16 @@
 __version__ = "0.18.13"
 
 
+class KedroDeprecationWarning(DeprecationWarning):
+    """Custom class for warnings about deprecated Kedro features."""
+
+
+warnings.filterwarnings("default", category=KedroDeprecationWarning)
```
astrojuanlu, line 16: Since the global warning registry overrides any CLI option https://discuss.python.org/t/how-to-silence-pkg-resources-warnings/28629/7?u=astrojuanlu, could we have this behind a conditional?

```suggestion
if not sys.warnoptions:
    warnings.filterwarnings("default", category=KedroDeprecationWarning)
```

see also https://docs.python.org/3.7/library/warnings.html#overriding-the-default-filter

astrojuanlu: (Actually there's a `if not sys.warnoptions` below already)

merelcht (approved):
I agree with @astrojuanlu 's suggestion, but otherwise LGTM! 👍

kedro/__init__.py
deepyaman: Done! Also changed it to a `simplefilter`, since not using `message` anyway.

ankatiyar:
Made some additional changes -
- Replace `FutureWarning` with `KedroDeprecationWarning`
- Skip the jupyter convert end to end test that was failing on windows only because it'll be removed soon anyway

PR #2977, opened 2023-08-25
Stop OmegaConfigLoader from reading config from hidden directory like `ipynb_checkpoints`

noklam:
> **NOTE:** Kedro datasets are moving from `kedro.extras.datasets` to a separate `kedro-datasets` package in
> [`kedro-plugins` repository](https://github.com/kedro-org/kedro-plugins). Any changes to the dataset implementations
> should be done by opening a pull request in that repository.
## Description

Fix #2583 

This is a bug because ConfigLoader and TemplatedConfigLoader does not exhibit the same behavior.

## Development notes

The root cause of this is because `glob.glob` != `fsspec.filesystem.glob`. 
image

I also check
- Path.glob("*") - include hidden by default
- fs.glob("*") - include hidden by default
- glob.glob("*") - ignore hidden by default

See the signature
```python
glob.glob(pathname, *, root_dir=None, dir_fd=None, recursive=False, include_hidden=False)
```
This `include_hidden` argument was only added in Python 3.11 (but it is always the default).

Noted that this is not so trivial because of the inconsistent between different `glob`. fsspec.glob gives you `str` while `Path.glob` gives you `Path`. In addition I need to add `resolve` method because `.` can be confused with `../` which is relative path.

## Implementation
- Add addition step to check if the path is a hidden folder.
Reference of `glob.glob` implemenation: https://github.com/python/cpython/blob/0c79fabb036ccbe71298b141b0a9ff0506c67ae6/Lib/glob.py#L229-L230

# reviewer:

- Please check if I have missed any edge case.
- `../` will be resolve before the check - ideally it should be done in the upstream. I am not sure what's the need for all the `as_posix` call. But I think it's out of the scope so I keep the change minimal. 

It was slightly tricky to deal with `.` because `fsspec.glob` is more limited compare to `glob.glob` or `Path.glob`, but it's necessary to make it work for any filesystem.

## Checklist

- [ ] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [ ] Added tests to cover my changes
- [ ] Checked if this change will affect Kedro-Viz, and if so, communicated that with the Viz team

PR #3145, opened 2023-10-09
Make `OmegaConfigLoader` accept paths containing dots outside of conf_source

### LINKED ISSUE #3004: Fix Configuration Loader to Handle Paths Containing dots
## Description

When working within Databricks workflows, it's possible to create a job that utilizes a notebook hosted inside a repository. Each time the job is executed, the code is automatically cloned to a path that depends on the commit. In my case, I have hosted a notebook that runs the Kedro project within the `.azuredevops` directory.

To locate the current project root, we can employ a regular expression (regex) pattern:

```python
pattern = r"^/Workspace/Repos/.internal/(?!./.azuredevops)."
```

With this regex, we can effectively identify the project root. For example, one possible path is:

```
/Workspace/Repos/.internal/02476ba86f_commits/20adcb96c8bec99ac2fad8b78b25158e7d968fa4/
```

However, there's an issue that arises when a period (dot) appears within the path. This problem is not limited to Databricks; it can also occur locally. It's important to note that the OmegaConfigLoader did not encounter any issues with this path until Kedro version 0.18.12.

## Context
Just running a normal kedro project. Which was working untill 0.18.12

## Steps to Reproduce
1. modify you project directory adding some `.`, for example if you a have a kedro project in `Users/yourname/kedro_project` you can create a new folder with some point `Users/yourname/.somefolder/kedro_project` 
2. [Second Step] Enable OmageConfigLoader
3. [And so on...]  Kedro run

## Expected Result
kedro should load the session

## Actual Result
The config loader is not be able to find any config file

```
`No files of YAML or JSON format found in /Workspace/Repos/.internal/02476ba86f_commits/20adcb96c8bec99ac2fad8b78b25158e7d968fa4/conf/base or /Workspace/Repos/.internal/02476ba86f_commits/20adcb96c8bec99ac2fad8b78b25158e7d968fa4/conf/databricks_dev matching the glob pattern(s): ['catalog*', 'catalog*/**', '**/catalog*']`
```

## Your Environment

* Kedro version used (`pip show kedro` or `kedro -V`): 0.18.13
* Python version used (`python -V`): 3.10.12
* Operating system and version: mac

ankatiyar:
## Description

This PR is to fix #3004 

- Changes in #2977 which was a bug fix to let `OmegaConfigLoader` ignore hidden files inside the `conf/` folder (or `conf_source`) created this issue with Kedro ignoring files entirely if any part of the path contains a dot.

## Development notes

- Updated `_is_hidden` function to only return `True` when the hidden folder/file is inside the `conf_source`
- Added a test case

EDITED TO ADD:

- Turns out `pathlib.is_relative_to()` and `pathlib.relative_to()` are methods that do not work for python <3.9. And, they're already deprecated and will be removed for 3.12.
- `os.path.rel_path()` is an alternate for this functionality but it works slightly differently from `pathlib.Path.relative_to()`. So some of the relative paths contain stuff like `../../some/path` which doesn't pass the "is_hidden()" test

#### Implementation info
* I first tried to remove `resolve()` since the `path` for files that is passed to this function is already a relative path (and not do any of the resolving with `conf_path`. But this doesn't work if something like `"../user1/catalog*"` has been added to `config_patterns` (as is the case in one of the tests)
```
    def _is_hidden(self, path: str):
        """Check if path contains any hidden directory or is a hidden file"""
        path = Path(path).as_posix()
        parts = path.split(self._fs.sep)  # filesystem specific separator
        HIDDEN = "."
        # Check if any component (folder or file) starts with a dot (.)
        return any(part.startswith(HIDDEN) for part in parts)
```
* Relative path resolving with `conf_path` trips up incase the `conf_source` is a zip/tar file. So if the protocol is zip/tar, it simply checks the "relative" path passed on to this function without `resolve()`

## Checklist


merelcht (commented):
The approach looks good to me, but I see there's some issues with the tests. I'll have another look when they're fixed 🙂 Also don't forget to add this to the release notes!

ankatiyar:
@merelcht The CI failures are because turns out the methods I used here `path.is_relative_to()/relative_to()` are not available for lower python versions(<3.9).

kedro/config/omegaconf_config.py
```diff
@@ -6,6 +6,7 @@
 import io
 import logging
 import mimetypes
+import os
```
merelcht, line 9: We actually try to not use `os` if it's possible.. Is there no other way to achieve this with using `pathlib`?

merelcht (approved):
Thanks for fixing this and explaining the approach so well. There's definitely a lot of small edge cases to account for. 👍

SajidAlamQB (approved):
Thank you @ankatiyar!

Issue #2984, opened 2023-08-30
Create new `spaceflights-pyspark` starter


merelcht:
## Description
Sub-task of https://github.com/kedro-org/kedro/issues/2838

## Context
[Introducing Utility Modules to Kedro #2388](https://github.com/kedro-org/kedro/issues/2388)

Issue #2506, opened 2023-04-12
Design full Project Creation CLI Flow 


amandakys:
Parent Issue: #2388 
## Description
Design the full revamped Project Creation CLI flow which includes:
- Add-ons selection (https://github.com/kedro-org/kedro/issues/2850)
- Template selection (e.g. pandas, pyspark, viz)
- Example selection

## Context

- https://github.com/kedro-org/kedro/issues/2388 Outlines the idea of the revamped CLI Flow
- https://github.com/kedro-org/kedro/issues/2838 Describes the new starters that will be created that can be selected as templates and examples during the project creation flow.
- https://github.com/kedro-org/kedro/issues/2850 Contains the design for the add-ons selection

yetudada:
Another requirement from @NeroOkwa: "Making sure Kedro-Viz is discoverable at CLI creation".

datajoely:
Some nice libraries for interactive command line interfaces:

- [InquirePy](https://inquirerpy.readthedocs.io/en/latest/)
- [Textual](https://github.com/Textualize/textual)

yetudada:
**Up for review; will be modified by @amandakys, @astrojuanlu and @merelcht**

## Design considerations 

We will work on the assumption that the user will be able to create a project template in the following ways: 
 - Using the project creation CLI wizard
 - Using the project creation CLI flags
 - Using their custom starters 

These assumptions mean that we should: 

 - Rework what happens when a user runs `kedro new` because the project creation CLI wizard starts running
 - Create new CLI flags to match the project creation CLI wizard 
 - Maintain the current CLI flags that allow users to choose a starter, namely `--starter=<path-to-starter>`, ` --starter=name_of_starter` and `--checkout=version_number`

### What's out of scope for V1?

 - We are not including Airflow and Databricks in this current flow; we will look at this after 0.19.0
 - Perhaps later, we can deprecate the CLI command group, `kedro starter list`, if we assume that users will not want to see _our_ list, but this remains to be verified

## What does the project creation wizard look like? 

The user should: 
 - Run `kedro new`
 - Be asked to provide a project name
 - Be asked if they want example code, i.e. spaceflights
 - Be asked for project add-ons (or tools)

> _Note:_ I've called add-ons, tools. But this can be changed back. 

```
(my-virtual-environment) ➜  kedro new 

Project name
===========
Please enter a human-readable name for your new project.
Spaces, hyphens, and underscores are allowed.

 [New Kedro project]: Customer segmentation 


[...]

PR #3169, opened 2023-10-12
Add PySpark to the add-ons flow

SajidAlamQB:
## Description

PySpark should be added to the add-ons options, allowing users to get setup for pyspark (see https://github.com/kedro-org/kedro/issues/2506#issuecomment-1717432995). This option should make use of the new spaceflights-pyspark starter: https://github.com/kedro-org/kedro/issues/2984

When a user select pyspark they should get the config needed to run Kedro project with pyspark. This shouldn't include the example pipelines. This will happen in a follow up ticket.

## Checklist


merelcht (commented):
The general approach looks good, but instead of adding the `spark` specific files in the basic template, these should be fetched from the new starter: https://github.com/kedro-org/kedro-starters/tree/main/spaceflights-pyspark. You might have to also create a `post_gen_project` file there..

kedro/framework/cli/starters.py
```diff
@@ -210,14 +211,17 @@ def _validate_range(start, end):
...
                 click.secho(message, fg="red", err=True)
                 sys.exit(1)
 
     if add_ons_str == "all":
         return list(ADD_ONS_DICT)
     if add_ons_str == "none":
         return []
+    # Guard clause if add_ons_str is None, which can happen if prompts.yml is removed
+    if not add_ons_str:
+        return []
```
SajidAlamQB, line 224: When `prompts.yml` is empty or doesn't exist we should default add-ons to be `"none"` hence we return an empty dictionary. We don't want `add_ons_str` to be `null`.

kedro/framework/cli/starters.py
```diff
@@ -493,6 +500,17 @@ def _make_cookiecutter_args(
     return cookiecutter_args
 
 
+def fetch_template_based_on_add_ons(template_path, cookiecutter_args: dict[str, Any]):
```
SajidAlamQB, line 503: If Pyspark option is selected we need to switch the template to the `spaceflights-pyspark` starter.

kedro/framework/cli/starters.py
```diff
@@ -210,14 +212,17 @@ def _validate_range(start, end):
...
                 click.secho(message, fg="red", err=True)
                 sys.exit(1)
 
     if add_ons_str == "all":
         return list(ADD_ONS_DICT)
     if add_ons_str == "none":
         return []
+    # Guard clause if add_ons_str is None, which can happen if prompts.yml is removed
+    if not add_ons_str:
+        return []
```
noklam, line 225:
```suggestion
        return []  # pragma: no cover
```

noklam (approved):
Thank you, it looks good to me. There is a coverage issue, feel free to ignore it. I am aware of this issue and I will try to fix this in my PR or a separate one later.

noklam:
I see all tests are passing but linting, I take the liberty to fix it.

SajidAlamQB:
> I see all tests are passing but linting, I take the liberty to fix it.

lol we did it at the same time.

merelcht:
If you combine `pyspark` with any of the other add-ons you potentially get duplicate entries in your `pyproject.toml`, because the `pyproject.toml` in the starter already contains all settings for e.g. docs, linting etc. I guess the `pyproject.toml` will need to be stripped.

Not related to just pyspark: I also noticed that `docs` requirements are already present in the base `pyproject.toml` and will also end up being duplicated even if pyspark isn't selected.

merelcht:
Yeah part of the stuff was stripped from the `pyproject.toml` in the empty template in https://github.com/kedro-org/kedro/pull/2987/files, but the starters were merged first so I didn't realise this had to happen.. 😅

SajidAlamQB:
> If you combine `pyspark` with any of the other add-ons you potentially get duplicate entries in your `pyproject.toml`, because the `pyproject.toml` in the starter already contains all settings for e.g. docs, linting etc. I guess the `pyproject.toml` will need to be stripped.
> 

We could just add checks to make sure that the entries we're adding are not already present.

noklam:
Our tests should have caught this - I will make a separate PR to fix the test. Essentially I think we need to parse the pyproject.toml properly instead of reading it as text file. 

```
        with open(pyproject_file_path) as pyproject_file:
            requirements = pyproject_file.read()
```

noklam:
I created the test in https://github.com/kedro-org/kedro/pull/3230, it catches an invalid combination that we need to fix.

kedro/framework/cli/starters.py
```diff
@@ -524,8 +549,8 @@ def _create_project(template_path: str, cookiecutter_args: dict[str, Any]):
     )
     add_ons = extra_context.get("add_ons")
 
-    # Only non-starter projects have configurable add-ons
-    if template_path == str(TEMPLATE_PATH):
+    # Only core template and spaceflights-pyspark have configurable add-ons
+    if template_path == str(TEMPLATE_PATH) or add_ons == "6":
```
merelcht, line 553: In `fetch_template_based_on_add_ons` there's the check `if add_ons and "Pyspark" in add_ons:` and here we check that `add_ons == "6"`. It seems that in both cases the add_ons comes from `add_ons = extra_context.get("add_ons")`, so could we make the checks consistent and either check for 6 or pyspark?

tests/framework/cli/test_starters.py
```diff
@@ -291,12 +299,12 @@ def test_parse_add_ons_invalid_range(input, capsys):
 
 @pytest.mark.parametrize(
     "input,first_invalid",
-    [("0,3,5", "0"), ("1,3,6", "6"), ("0-4", "0"), ("3-6", "6")],
+    [("0,3,5", "0"), ("1,3,7", "7"), ("0-4", "0"), ("3-7", "7")],
```
merelcht, line 302: Why not just make the invalid number "100"? That way we don't have to keep bumping this with every new add on.

merelcht (commented):
I've left some minor comments but I think this is almost ready to be merged. I suggest fixing the other issues I found in a separate PR, because they aren't caused by the changes here and that shouldn't block the work for adding the `viz` option.

kedro/framework/cli/starters.py
merelcht: And shouldn't this check if `6` (or pyspark) is in the selected rather than an exact match?

SajidAlamQB:
> I've left some minor comments but I think this is almost ready to be merged. I suggest fixing the other issues I found in a separate PR, because they aren't caused by the changes here and that shouldn't block the work for adding the `viz` option.

Oh, I've already added some changes that attempt to fix this but I can revert and open it up on new PR? @merelcht

merelcht:
> Oh, I've already added some changes that attempt to fix this but I can revert and open it up on new PR? @merelcht

Ah no, that's great!! I'll have a look 😄 I just didn't want to block this PR by my earlier comments.

tests/framework/cli/test_starters.py
SajidAlamQB: I believe its because the `_validate_selection` function checks each number in the list of selected add-ons and on the first invalid selection it raises an error which for now is 7, the test will always have to know the first invalid selection.

kedro/templates/project/hooks/utils.py
```diff
@@ -60,42 +63,127 @@ def setup_template_add_ons(selected_add_ons_list, requirements_file_path, pyproj
...
+        with open(requirements_file_path, 'w') as file:
+            for line in content:
+                if line.strip() not in lint_requirements:
+                    file.write(line)
+
+        with open(pyproject_file_path, 'r') as file:
+            content = file.read()
+        with open(pyproject_file_path, 'w') as file:
+            content = content.replace(lint_pyproject_requirements, "")
+            file.write(content)
```
merelcht, line 79: I'm not quite sure I understand what's happening here. Is this checking the `requirements.txt` and `pyproject.toml` files from the starter and then rewriting only the content that's not linting? 

I think we should just make sure that the `pyproject.toml` and `requirements.txt` files in the starters contain the minimum so we can assume they don't contain any "add-ons" related stuff by default.

SajidAlamQB: Yes, as you said it rebuilding pyproject and requirements while removing the add-ons that are not selected. But if we are happy to strip pyproject.toml on starters then this won't be needed.

kedro/framework/cli/starters.py
SajidAlamQB: @noklam's refactor made use of the `_parse_add_ons_input` in `_make_cookiecutter_args` so we have the selected add-ons in the cookiecutter context.

kedro/templates/project/hooks/utils.py
SajidAlamQB: I will revert it for now then.

kedro/framework/cli/starters.py
```diff
@@ -524,8 +549,8 @@ def _create_project(template_path: str, cookiecutter_args: dict[str, Any]):
     )
     add_ons = extra_context.get("add_ons")
 
-    # Only non-starter projects have configurable add-ons
-    if template_path == str(TEMPLATE_PATH):
+    # Only core template and spaceflights-pyspark have configurable add-ons
+    if template_path == str(TEMPLATE_PATH) or add_ons == "Pyspark":
```
merelcht, line 553:
```suggestion
    if template_path == str(TEMPLATE_PATH) or "Pyspark" in add_ons:
```

merelcht (approved):
Left one more comment about the `pyspark` check. `add_ons` is a list there so `==` will always fail. But otherwise this looks good to be merged! 👍

kedro/framework/cli/starters.py
SajidAlamQB: oops sorry misread this, yes it shouldn't be an exact match.

PR #3262, opened 2023-11-02
Document Python version support policy

### LINKED ISSUE #2815: Adopt less conservative Python support policy
## Description
Supporting old Python versions is important for corporate users that want newer features, but adds toil on the Kedro team. Maybe we could explore adopting a less conservative Python support policy.

## Context
We are among the last ones to drop support for old Python versions, and this creates problems at the tail of the support window.

For example, pip-tools recently dropped support for Python 3.7, and it's affecting us:

https://github.com/kedro-org/kedro/pull/2813

Notice that, whatever we do, should be starting on 0.19, or on 0.19+.

Also important to bear in mind how this interacts with other issues:

- The fact that upgrading Kedro is hard https://github.com/kedro-org/kedro/issues/1887
- The fact that we don't have a fixed or very predictable release cadence

## Possible Implementation
https://numpy.org/neps/nep-0029-deprecation_policy.html

image

## Possible Alternatives
NEP 29 might be a tad too conservative. Maybe something more specific, like

> Once a Python version is EOL, we're removing it from our CI, and we do not guarantee support for it going forward.

Which is maybe the smallest advance that can be made from the current position, but would still go a long way.

merelcht:
## Description
Closes #2815 

## Development notes
Added the section to the README.md + docs.

I wasn't sure what the right place is to add this paragraph. Options I considered:

- [First steps: Setup Kedro.](https://docs.kedro.org/en/stable/get_started/install.html) This is where I've added it now, right below the installation prereqs where we talk about Python version. 
- [Project setup](https://docs.kedro.org/en/stable/kedro_project_setup/index.html) It might fit here with the chapter on dependencies? 
- [Development](https://docs.kedro.org/en/stable/development/index.html) Also a possible place because this is where we talk about developing with Kedro. 

I was also thinking about whether this should be added to the FAQ page?

@stichbury and @astrojuanlu what do you think? 

## Checklist


README.md
```diff
@@ -93,3 +93,7 @@ We keep a list of [technical FAQs in the Kedro documentation](https://docs.kedro
 ## How can I cite Kedro?
 
 If you're an academic, Kedro can also help you, for example, as a tool to solve the problem of reproducible research. Use the "Cite this repository" button on [our repository](https://github.com/kedro-org/kedro) to generate a citation from the [CITATION.cff file](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files).
+
+## Python version support policy
+* The core [Kedro Framework](https://github.com/kedro-org/kedro) supports all Python versions that are actively maintained by [the Python Software Foundation](https://www.python.org/psf-landing/). When a [Python version reaches end of life](https://devguide.python.org/versions/#versions), support for that version is dropped from Kedro. This is not considered a breaking change.
```
astrojuanlu, line 98: Splitting hairs here but I'm not 100 % sure that the PSF maintains CPython, maybe we should be saying here "by the CPython core team". May I summon @pablogsal ? 🖤

astrojuanlu (approved):
Other than my comment about who maintains CPython this looks good 🙏🏽

README.md
pablogsal: The most correct statement is that "The CPython team maintains CPython" as the PSF mainly holds the trademarks rights.

astrojuanlu: Thanks a bunch 🙏🏽

stichbury (approved):
LGTM!

README.md
deepyaman: Probably splitting hairs, but we _attempt_ to support the newer versions, but sometimes it takes a while. Does it make sense to distinguish that we're talking about policy for older versions?

astrojuanlu: I think going forward and especially since splitting datasets it will be much easier to support newer versions right? it's just that we haven't gotten around adding 3.12 to our CI yet, I believe

Issue #3185, opened 2023-10-17
Bug in kedro new add-ons flows 


noklam:
## Description

This is in `develop` branch only.

See 
![image](https://github.com/kedro-org/kedro/assets/18221871/8e8095d7-ffd1-4fb7-8ea2-0ffadc0e5c44)

I try two different CLI and the result is unexpected.

1. `kedro new`, then `1,33`
2. `kedro new`, then `1, 6` 

Both are invalid options, the first one fail immediately but the second didn't. I believe the 1. is the correct behavior. Cc @amandakys Can you confirm this?

noklam:
Noted I have a space in the screenshot with `1, 33`, I try again and it fails without space.
![image](https://github.com/kedro-org/kedro/assets/18221871/c7defc99-0a55-43e9-b62d-b674a25ae71d)

merelcht:
Great catch! I have been able to replicate this behaviour. I think it's as soon as you add a double digit choice, so anything above 10.

AhdraMeraliQB:
This issue has been extended to include extending the flexibility of the case and whitespace sensitivity

merelcht:
Fixed in https://github.com/kedro-org/kedro/pull/3263

PR #3263, opened 2023-11-02
Adjust add-on input validation

AhdraMeraliQB:
## Description

Following #3185, this PR ensures that invalid inputs to the add-on prompt fail immediately. Additionally, case sensitivity has been removed, whitespaces are now ignored, and selections larger than 9 are processed as invalid selections instead of improper formatting.

The case sensitivity changes have also been made to the add-on flag inputs (they were already whitespace robust prior to this PR).

## Checklist


merelcht:
> Note: Values provided to the add-on flag remain case sensitive after these changes.

I would expect this to also have some processing to handle cases, the flows aren't entirely consistent, but I think this behaviour should be the same.

merelcht:
I also noticed that if I put in a negative number I still get the unhelpful message: `ValueError: invalid literal for int() with base 10: ''`

AhdraMeraliQB:
@merelcht 
> I would expect this to also have some processing to handle cases, the flows aren't entirely consistent, but I think this behaviour should be the same.

Done! Addressed in [a1fb0c6](https://github.com/kedro-org/kedro/pull/3263/commits/a1fb0c60813de674c0f1ab2994c4338abdf512be)


> I also noticed that if I put in a negative number I still get the unhelpful message: `ValueError: invalid literal for int() with base 10: ''`

Fixed in [906f63a](https://github.com/kedro-org/kedro/pull/3263/commits/906f63ae05a66aab59971d94468a54ee72b0bb48)

tests/framework/cli/test_starters.py
```diff
@@ -314,17 +314,6 @@ def test_parse_add_ons_invalid_range(input, capsys):
     assert message in capsys.readouterr().err
 
 
-@pytest.mark.parametrize(
-    "input,first_invalid",
-    [("0,3,5", "0"), ("1,3,7", "7"), ("0-4", "0"), ("3-7", "7")],
-)
-def test_parse_add_ons_invalid_selection(input, first_invalid, capsys):
```
AhdraMeraliQB, line 321: This test was removed as this validation step is no longer performed by `_parse_add_ons_input`

tests/framework/cli/test_starters.py
```diff
@@ -906,7 +895,29 @@ def test_directory_flag_with_starter_alias(self, fake_kedro_cli):
...
+            "  1,  2, 3  ",
+            " 1-3 ",
+            "  1 - 3  ",
+            "ALL",
+            "All",
+            "AlL",
+            "NONE",
+            "None",
+            "nonE",
+        ],
```
noklam, line 920: Some of these tests are repetitive, can it be reduce some of these to one example. For example
```
            "ALL",
            "All",
            "AlL" 
 ```
can be reduced to one since `AlL` is probably the most weird case?

noklam (commented):
Looks good in general, leave a minor comment to reduce the test cases.

tests/framework/cli/test_starters.py
AhdraMeraliQB: I do agree `AlL` is probably the strangest case, though I'd avoid simplifying to just either `ALL` or `All` as I'd want to ensure all small letters, all capitals, and a mix of the two are all processed correctly. `AlL` and `nonE` have both been dropped in [5ad8bab](https://github.com/kedro-org/kedro/pull/3263/commits/5ad8babe5a2f23eba1fe9ef30866f9cd6af254ff)

merelcht:
There's still a difference in user experience when I enter a negative number vs a positive invalid number:

Screenshot 2023-11-03 at 13 50 04

This might be fine, because `-` is also used for range here. @amandakys do you have any thoughts on this?

amandakys:
@merelcht Do you mean to say that a negative value should be accepted and auto corrected to be a positive value? Only when it falls within correct range, i.e we should validate the absolute value?

noklam:
I suggest test it after solving the conflicts. I suspect this maybe solved already as a side-effect of https://github.com/kedro-org/kedro/pull/3265. I introduced a new dictionary and use it everywhere, this avoids an addition rules because the condition it checking if value `in` the dictionary thus handle edge cases like negative value.

merelcht:
I found one more interesting edge case, what if you put a range that's way beyond the acceptable number, e.g. "5-12", now you'll get:

Screenshot 2023-11-07 at 17 32 15

Maybe I'm going too far into what users could reasonably enter, but I don't see why this shouldn't just throw: "'12' is not a valid selection.
Please select from the available add-ons: 1, 2, 3, 4, 5, 6, 7."

noklam:
@merelcht This is a reasonable expectation, I have made the changes and updated the tests.

SajidAlamQB:
Should we ignore white spaces? :

image

image

noklam:
@SajidAlamQB Did you test with latest branch? There are test cases covering that already and it seems working fine on my end.

image

SajidAlamQB:
> @SajidAlamQB Did you test with latest branch? There are test cases covering that already and it seems working fine on my end.

I created a new environment and it seemed to work there. Perhaps my old environment had some issues... 😅

merelcht (approved):
LGTM! 👍

SajidAlamQB (approved):
Thank you, @noklam!

noklam:
Thank you @AhdraMeraliQB for implementing the solution 😸 I just help to fix a test.

Issue #2668, opened 2023-06-12
Use dataset factories to register default datasets


ankatiyar:
## Description
After the introduction of dataset factories in #2423, the creation of default datasets which currently is handled by runners should be handled in the `DataCatalog`.
This can be done adding a catch all pattern to the `DataCatalog` if no such pattern exists in the project's catalog config.

noklam:
Ccing @marrrcin for comments as I think overwriting default datasets is done across the collection of plugins you created.

ankatiyar:
## Current behaviour
* creation of default datasets is handled in the `runner` -
* Each runner has a `create_default_dataset()` class which returns an `MemoryDataset`/`SharedMemoryDataset`(for parallel runner) object for each "unregistered dataset"(not mentioned in the catalog as an explicit entry but mentioned in the pipeline inputs or outputs) and adds it to the catalog.


Updating this behaviour to use a **dataset factories** instead for `SequentialRunner` and `ThreadRunner` is straight forward -
- Just add a catch-all pattern to the catalog with `MemoryDataset` : 
```
"{default}":
  type: MemoryDataset
```
- Add a function to catalog `catalog.add_pattern(name, config)` which can be called either -
   - With the current behaviour, in `runner.run()` after the "free outputs" have been calculated as these datasets are returned as the output for the session
   - Before `runner.run()` is called in the session, but then some logic to determine which datasets are "free outputs" needs to be added.

### Issues with `ParallelRunner`
- The default dataset for `ParallelRunner` is `SharedMemoryDataset` which is a wrapper around `MemoryDataset`. It is **not** an `AbstractDataset` -> therefore, doesn't have a `from_config()` function that the catalog uses in `catalog._get_dataset()/from_config()` to initialise datasets on creation
- `ParallelRunner` has a `_manager` attribute which is passed to the `SharedMemoryDataset` object on initialisation (can't be done in a declarative way by simply adding a catalog entry)

### Solution 
I've managed to hack together a solution in the draft PR #3269 

Some of these things can be cherry-picked and implemented regardless of whether we decide to go for the whole thing -

- Moved `SharedMemoryDataset` to `kedro/io` - it's currently defined in `kedro/runner/parallel_runner.py`
- Make `SharedMemoryDataset` a subclass of `AbstractDataset` - this allows the catalog to register it also through a "catalog entry" (catch-all pattern) but without initialising the underlying `shared_memory_dataset` 
[...]

PR #3269, opened 2023-11-03
[Draft] Trying to use dataset factories for default datasets 

ankatiyar:
## Description

This PR is a quick prototype for #2668 

## Current behaviour
* creation of default datasets is handled in the `runner` -
* Each runner has a `create_default_dataset()` class which returns an `MemoryDataset`/`SharedMemoryDataset`(for parallel runner) object for each "unregistered dataset"(not mentioned in the catalog as an explicit entry but mentioned in the pipeline inputs or outputs) and adds it to the catalog.

Updating this behaviour to use a **dataset factories** instead for `SequentialRunner` and `ThreadRunner` is straight forward -
- Just add a catch-all pattern to the catalog with `MemoryDataset` : 
```
"{default}":
  type: MemoryDataset
```
- Add a function to catalog `catalog.add_pattern(name, config)` which can be called either -
   - With the current behaviour, in `runner.run()` after the "free outputs" have been calculated as these datasets are returned as the output for the session
   - Before `runner.run()` is called in the session, but then some logic to determine which datasets are "free outputs" needs to be added.

### Issues with `ParallelRunner`
- The default dataset for `ParallelRunner` is `SharedMemoryDataset` which is a wrapper around `MemoryDataset`. It is **not** an `AbstractDataset` -> therefore, doesn't have a `from_config()` function that the catalog uses in `catalog._get_dataset()/from_config()` to initialise datasets on creation
- `ParallelRunner` has a `_manager` attribute which is passed to the `SharedMemoryDataset` object on initialisation (can't be done in a declarative way by simply adding a catalog entry)

### Solution 
I've managed to hack together a solution in the draft PR #3269 

Some of these things can be cherry-picked and implemented regardless of whether we decide to go for the whole thing -

- Moved `SharedMemoryDataset` to `kedro/io` - it's currently defined in `kedro/runner/parallel_runner.py`
- Make `SharedMemoryDataset` a subclass of `AbstractDataset` - this allows the catalog to register it also through a "catalog entry" (catch-all pattern) but without initialising the underlying `shared_memory_dataset` 
- In the code for `ParallelRunner` -> it initialises the `shared_memory_dataset` attribute of the `SharedMemoryDataset` with the (`ParallelRunner`'s)`manager.MemoryDataset()` object if a dataset is a `SharedMemoryDataset`
- The way i've implemented it right now is first the runner checks for the existence of all pipeline datasets with `catalog.exists(dataset)` which calls `catalog._get_dataset(dataset)` -> the dataset factory datasets are registered to the Catalog. And then for the `SharedMemoryDataset`, initialise the underlying dataset. (This could be improved in the final implementation)

### DIFF
--- a/kedro/io/__init__.py
+++ b/kedro/io/__init__.py
@@ -14,7 +14,7 @@ from .core import (
 )
 from .data_catalog import DataCatalog
 from .lambda_dataset import LambdaDataset
-from .memory_dataset import MemoryDataset
+from .memory_dataset import MemoryDataset, SharedMemoryDataset
 
 __all__ = [
     "AbstractDataset",
@@ -26,5 +26,6 @@ __all__ = [
     "DatasetNotFoundError",
     "LambdaDataset",
     "MemoryDataset",
+    "SharedMemoryDataset",
     "Version",
 ]
--- a/kedro/io/data_catalog.py
+++ b/kedro/io/data_catalog.py
@@ -438,6 +438,12 @@ class DataCatalog:
             return True
         return False
 
+    def add_default_pattern(self, dataset_type):
+        for pattern, _ in self._dataset_patterns.items():
+            if self._specificity(pattern) == 0:
+                return
+        self._dataset_patterns["{default}"] = {"type": dataset_type}
+
     @classmethod
     def _resolve_config(
         cls,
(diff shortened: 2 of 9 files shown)


merelcht (commented):
On first review, this doesn't look as messy as I thought it might be! I'd like to do a deeper dive into this to understand the flow better.

I'm wondering why the `_SharedMemoryDataset` didn't inherit from `AbstractDataset` when it was introduced.. @idanov do you remember this? It was added by Kiyo: https://github.com/quantumblacklabs/private-kedro/pull/310

ankatiyar:
Closing this as #3332 has been merged

Issue #1721, opened 2022-07-21
Improve DataCatalog and ConfigLoader with autocompletion and meaningful representation when it get printed


noklam:
# Dataset discovery
![Image](https://user-images.githubusercontent.com/18221871/179968511-7d6fea56-db93-4e92-8d7b-4799b5506bf7.png)

Running the variable in a notebook cell is the most common way to inspect a variable in Jupyter. Currently `catalog` give us an useless memory address, it would be much nicer if it prints out what's available, potentially just wrap it as `catalog.list()` for simplicity.

demo:

![Image](https://user-images.githubusercontent.com/18221871/179978660-2ad873bc-d37d-4b00-8d3f-e948150d6485.png)

# Autocompletion
```python
# Current Way - Autocompletion not possible
catalog.load("example_iris_data").head()

# Slightly more verbose way - but much easier to type with Tab-completion
catalog.datasets.example_iris_data.load().head()
```

Under the hood, auto-completion works with checking the dir(object) method

Possible solutions:
1. Modify __dir__ to shows `datasets`, so user can do `catalog.` + `Tab`, it will shows other method available in `catalog` though. We can probably makes `datasets` show up at the top of autocompletion too and others method at the bottom. I think it just has to be a list most likely. Of course we have to implement the `__getattr__` so it actually looks at `catalog.datasets` too when we do `catalog.dataset_name`.

demo:
![Image](https://user-images.githubusercontent.com/18221871/180262642-b31ad3d2-72fe-446b-ae92-5d52e1e58ad3.png)

3. Implement a dict-like interface so when user do `catalog[` + `Tab`, it will shows all the datasets.
# Bonus
`session` and `context` also has useless `__repr__`, but they are not too useful in the interactive workflow so it has lower priority.

AntonyMilneQB:
LOVE this idea. I've added it to the interactive workflow milestone which I still need to populate a bit more since I've had related thoughts that I haven't had a chance to write down properly yet.

Just a few notes for now:
* the correct mechanism to access `dataset_name` through `catalog` needs to be formalised anyway as per my comments in https://github.com/kedro-org/kedro/issues/1691. Easy tab completion in the interactive workflow should definitely be a factor here. Note there's a complication that namespaced datasets have `.` in their name which might make access through `getattr` awkward. So we need to think quite carefully about what are the different options and the pros and cons here
[...]

PR #3299, opened 2023-11-11
Add meaningful representation when printing a DataCatalog 

Galileo-Galilei:

## Description
 This PR focuses on solving (very partially) #1721. I focus only on making catalog printing more meaningful, instead of the current ``<kedro.io.data_catalog.DataCatalog at 0x5x231>``. 

The "best" representation is still not clear, and this PR aims sharing publicly trials and errors to make it more meaningful. Some requirements I'd like to meet for the "best representation" of the printed objects: 
- it should leverage existing underlying code as much as possible
- it should be "copy-pasteable" to generate the object itself
- it should be easy to read: 
  - for datasets, I think that "easy to read" means "look like a black formatted string" 
  - for the catalog, I think the information we are looking for is *mostly* the datasets names. I personally like having one dataset per line, even if it gives very long lines and breaks black formatting, because the details fo the datasets is mostly a detail.  I am afraid than formatting the entire catalog with black can lead to hundreds of lines of printing. Such a cluttered output would be against what we are trying to achieve.

**In scope** : Making ``print(catalog)`` informative

**Out of scope** : 
  - Autocompletion
  - it is likely worth leveraging the ``AbstractDataset.__repr__`` method to generate the ``DataCatalog.__repr__``.  However, fixing potential issues in each dataset ``_describe`` is out of scope. 

## Development notes

1. ``AbstractDataset.__repr__``
  - I removed the ``__str__`` method and fully replaced it by ``__repr__``. When ``__str__`` is not user defined, ``__repr__`` is used instead which is exactly what we want: consistency between the 2 methods. 
  - I renamed the ``_to_str`` as ``_prettify_dict_to_str`` (which looks slightly more informative, but not very nice either TBH)
  - I also changed the behaviour of this ``_to_str`` method : it used to customize the string representation of dict. We now rely on the default ``__str__`` method of ``dict``. The main change is that we keep quotes around strings, which is necessary to ensure the output can be copy pasted and remain valid python code. 
- I created a separated  ``_build_str_representation`` which builds the string on one line for ease of integration in the catalog. The ``__repr__`` function only calls this methods and format it with black, which often renders it no several lines.

```python 
# ⚙️ Internal representation: output of _build_str_representation
CSVDataSet(filepath="temp.csv", protocol={"sep": ",", "decimal": ".", "header": True}, save_args={"index": False, "sep": ";", "decimal": ",", "header": False})

# ✅ output of __repr__
CSVDataSet(
    filepath="temp.csv",
    protocol={"sep": ",", "decimal": ".", "header": True},
    save_args={"index": False, "sep": ";", "decimal": ",", "header": False},
)
```

2. ``DataCatalog.__repr__``
- Implement a ``__repr__`` method which prints one line per dataset beginning by the dataset name

```python
# ✅Approach 1 (suggested)
DataCatalog(
    data_sets={
        'ds': CSVDataSet(filepath='temp.csv', protocol={'sep': ';'}, save_args={'index': False}),
        'ds2': CSVDataSet(filepath='temp.csv', protocol={'sep': ',', 'decimal': '.', 'header': True}, save_args={'index': False, 'sep': ';', 'decimal': ',', 'header': False}),
        'ds3': JSONDataSet(filepath='temp.csv')
})
```

For comparison, here is how it would have look like if we use black to render the string:

```python
# ❌Approach 2 : black-like representation

DataCatalog(
    data_sets={
        "ds": CSVDataSet(
            filepath="temp.csv", protocol={"sep": ";"}, save_args={"index": False}
        ),
        "ds2": CSVDataSet(
            filepath="temp.csv",
            protocol={"sep": ",", "decimal": ".", "header": True},
            save_args={"index": False, "sep": ";", "decimal": ",", "header": False},
        ),
        "ds3": JSONDataSet(filepath="temp.csv"),
    }
)
```

⚠️ Points to discuss : 
- [ ] This introduces black as a dependency again while we've jsut moved to ruff. Do we want this? 
- [ ] Each dataset representation is very long and may be hard to read. Do people prefer approach  1 or 2?
- [ ] "as is", the representation is incorrect and cannot be copy-pasted to generate a new catalog because the ``_describe()`` method of many dataset is incorrect (but this is out of scope), see the ``protocol`` argument above in ``CSVDataset``. Is it ok to release it with incorrect representation? 

📝 TODO: 
- [ ] make it more robust to missing _describe
- [ ] make it more robust to potentially invalid code which may crashes black
- [ ] add black as a dependency
- [ ] write tests

## Developer 

Certificate of Origin

## Checklist

### DIFF
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -17,6 +17,7 @@ from pathlib import Path, PurePath, PurePosixPath
 from typing import Any, Callable, Generic, TypeVar
 from urllib.parse import urlsplit
 
+from black import Mode, format_str
 from cachetools import Cache, cachedmethod
 from cachetools.keys import hashkey
 
@@ -227,10 +228,37 @@ class AbstractDataset(abc.ABC, Generic[_DI, _DO]):
             message = f"Failed while saving data to data set {str(self)}.\n{str(exc)}"
             raise DatasetError(message) from exc
 
-    def __str__(self):
-        return f"{type(self).__name__}({_prettify_dict_to_str(self._describe(), True)})"
+    # def __str__(self):
+    #     return format_str(self.build_str_representation(), mode=Mode())
 
     def __repr__(self):
+        return format_str(self._build_str_representation(), mode=Mode())
+
+    def _build_str_representation(self):
+        def _prettify_dict_to_str(obj, is_root=False):
+            """Returns a string representation where
+            1. The root level (i.e. the Dataset.__init__ arguments) are
+            formatted like Dataset(key=value).
+            2. Root arguments are alphabetically sorted.
+            3. None / empty dict values are not shown.
+            """
+
+            fmt = "{}={}" if is_root else "'{}': {}"  # 1
+
+            if is_root:
+                sorted_dict = sorted(obj.items(), key=lambda pair: str(pair[0]))  # 2
+
+                text = ", ".join(
+                    fmt.format(key, _prettify_dict_to_str(value))  # 2
+                    for key, value in sorted_dict
+                    if value  # 3
+                )
+
+                return text if is_root else "{" + text + "}"  # 1
+
+            # not the root
+            return f"'{str(obj)}'" if isinstance(obj, PurePath) else str(obj)
+
         return f"{type(self).__name__}({_prettify_dict_to_str(self._describe(), True)})"
 
     @abc.abstractmethod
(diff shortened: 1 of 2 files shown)


Galileo-Galilei:
@astrojuanlu Any thoughts of the points to discuss above ? I find it quite satisfying right now but I want to double check before creating tests.

datajoely:
Nice nice work! I think I like Black representation because it will mostly be viewed in a terminal with limited width

datajoely:
Are we doing anything to mask credentials?

astrojuanlu:
Thanks @Galileo-Galilei! My main points are

1. I see that this is basically a "proper" `repr` for the `DataCatalog` in the sense that copy-pasting its output on the interpreter would rebuild the same object. that's fine by me for now, as long as
2. we are properly masking credentials or not showing them at all, and
3. we are not introducing a runtime dependency with `black`, which I think would be problematic

noklam:
I will also be cautious to introduce `black` for just formatting the __repr__, on the other hand, does `rich` handle this already or does it offer some kind of `prettify` method?

noklam:
@Galileo-Galilei Are you still interested to finish this PR?

Galileo-Galilei:
Nope, sorry, I close it. Very little time at the moment, and I did not come up with something totally satisfying for now so it still requires a bit of thinking.

PR #3300, opened 2023-11-13
Modify `KedroContext` with frozen attributes instead of frozen class

### LINKED ISSUE #1459: Revisit: Make KedroContext a dataclass and add `config_loader` as a property

noklam:
Signed-off-by: Nok <nok.lam.chan@quantumblack.com>

## Description

Fix #1459. Initially we plan to freeze the class to prevent users changing the internal details. After some discussion, this can be too restrictive and we decide to only freeze existing attributes. Users are free to create their own as long as it doesn't conflict with the internal one.

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -478,7 +478,7 @@ We are grateful to the following for submitting PRs that contributed to this rel
 
 ## Bug fixes and other changes
 * Removed fatal error from being logged when a Kedro session is created in a directory without git.
-* `KedroContext` is now an `attrs`'s frozen class and `config_loader` is available as public attribute.
+* `KedroContext`'s attributes are now frozen and `config_loader` is available as public attribute.
 * Fixed `CONFIG_LOADER_CLASS` validation so that `TemplatedConfigLoader` can be specified in settings.py. Any `CONFIG_LOADER_CLASS` must be a subclass of `AbstractConfigLoader`.
 * Added runner name to the `run_params` dictionary used in pipeline hooks.
 * Updated [Databricks documentation](https://docs.kedro.org/en/0.18.1/deployment/databricks.html) to include how to get it working with IPython extension and Kedro-Viz.


noklam:
@Galileo-Galilei Any chance you can test with this branch?

Galileo-Galilei:
I'll do it tonight

merelcht:
Generally this looks good to me. In the description you asked: 

> Looking at the current interface, should we re-order the argument as it seems arbitary, (and) maybe make it keyword only.

I 100% agree this order looks arbitrary now, especially because there's a mix of "internal"/"private" arguments starting with `_` and not. I think now would be the right moment to re-order, because it's a breaking change.

Galileo-Galilei:
I haven't looked at the code in detail yet, but I have tested the behaviour and **it works as expected**:
- The attributes are indeed frozen, and I can add attributes (just with ``context.next_attr="value"``, not need to use a workaround with ``object.__setattr__``)
- ``kedro-mlflow`` does work with this version 👍 

I agree that it is likely time to reorder all arguments to make clearer which one are public. 

However, **I still have more generic concerns on whether all (or even *any*) attributes should be frozen**. The ``after_context_hook`` is particularly useful to **distribute the logic without relying on the template**, which makes it much easier to update and maintain. For instance, we have a plugin at work which register a custom ``ConfigLoader`` (basically because we want to *fetch* the conf from another place, but we want to benefit from all the capabilities of kedro for *resolving* - related to #2481, but this is another story). If we cannot modify it through the hook, we will need to tell our users to update their templates on their which is basically a nightmare in maintenance. 

I can also think of use cases in which I want to update the ``_extra_params`` dictionary, but since the underlying object is mutable, ``context._extra_params.update({"new_param":"value"})`` works. 

By order of preference: 
1. Do not freeze anything
2. Freeze all attributes except ConfigLoader
3. Stay as is (but I am afraid that it will break some of our internal workflows at work, I need to check but I am not sure I could make it before the release)

tests/framework/context/test_context.py
```diff
@@ -209,9 +209,29 @@ def test_attributes(self, tmp_path, dummy_context):
...
+    )
+    def test_internal_attributes(self, dummy_context, internal_attr):
+        getattr(dummy_context, internal_attr)
+
+    def test_immutable_class_attribute(self, dummy_context):
+        with pytest.raises(FrozenAttributeError):
+            dummy_context.project_path = "dummy"
+
+    def test_set_new_attribute(self, dummy_context):
+        dummy_context.mlflow = 1
```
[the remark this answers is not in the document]
noklam: This test attribute are mutable without throwing erros - but I agree asserting the attributed has been set properly is better.

noklam:
I have decided to not freeze any attribute at the end. Choosing some attributes to freeze and not others feels arbitrary to me, let's just fallback to Python convention with `_`. On the other hand, we should still respect that this is something internal to Kedro, so this is a "do it at your own risk", but of course we should't break it easily without offering a valid alternative.

Additionally, I re-arrange the order of the argument. Noted that `package_name` [will be removed](#1892 ) soon so it's irrelevant.

RELEASE.md
```diff
@@ -479,7 +479,7 @@ We are grateful to the following for submitting PRs that contributed to this rel
 
 ## Bug fixes and other changes
 * Removed fatal error from being logged when a Kedro session is created in a directory without git.
-* `KedroContext` is now an `attrs`'s frozen class and `config_loader` is available as public attribute.
+* `KedroContext`'s attributes are now frozen and `config_loader` is available as public attribute.
```
merelcht, line 482: Don't forget to update this again.

kedro/framework/context/context.py
```diff
@@ -158,18 +158,38 @@ def _expand_full_path(project_path: str | Path) -> Path:
     return Path(project_path).expanduser().resolve()
 
 
-@frozen
+@define(slots=False)  # Enable setting new attributes to `KedroContext`
```
merelcht, line 161: Is this still needed now we're not freezing anything?

tests/framework/session/test_session.py
```diff
@@ -601,7 +640,9 @@ def test_run(
                 "__default__": mocker.Mock(),
             },
         )
+        print(f"{mock_context_class=}")
         mock_context = mock_context_class.return_value
+        print(f"{mock_context=}")
```
merelcht, line 645: I'm guessing these print statement were just for debugging?

tests/framework/session/test_session.py
```diff
@@ -74,7 +111,9 @@ def mock_settings(mocker):
 
 @pytest.fixture
 def mock_settings_context_class(mocker, mock_context_class):
+    # mocker.patch("dynaconf.base.LazySettings.unset")
```
merelcht, line 114:
```suggestion
```

merelcht (commented):
Thanks @Galileo-Galilei for testing and explaining your pain points. This is really insightful. I agree with the approach to not freeze anything in the end @noklam 👍 

I was just wondering if you then still need all the tests for checking attributes?

kedro/framework/context/context.py
noklam: > I was just wondering if you then still need all the tests for checking attributes?

It is needed, and for this reason I keep the test. If you try to set `slots=True` (or remove this argument since it's default), you will see the test fail.  [The default for standard Python library is `False`](https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass)

https://www.attrs.org/en/stable/api.html#attrs.define - The reason for this is slotted class is more efficient and make sense if it is used as "dataclass" (think of pydantic model that get convert to JSON etc), in this case it's irrelevant

merelcht (approved):
Thank you @noklam for revisiting this change, good thing we didn't release it the way it was 👍

Galileo-Galilei:
Thank you everyone and especially @noklam for this change, and sorry for not having been as available as I had wanted to to help for this!

PR #3311, opened 2023-11-15
Moved the environment assumption from `ConfigLoader` to `_ProjectSettings`

### LINKED ISSUE #2971: Remove the environment default for ConfigLoader
### Context
See #2819 and more

Currently ConfigLoader assume `base` and `local` exist, but for single config use cases ( a notebook) it's not relevant.

### Requirements
- Backward compatible
- Enable use of OmegaConfigLoader without environment

### Current workaround
```python
OmegaConfigLoader(conf_source=".", base_env=".", default_run_env=".")  # To use this in a notebook
```

### Potential Options:
1. Change defaults to "." and change `settings.py` default to "base" and "local" (and starters)
2. Change defaults to "." and change `KedroSession.create`

```python
settings.CONFIG_LOADER_CLASS
return config_loader_class(
    conf_source=self._conf_source,
    env=env,
    runtime_params=extra_params,
)
```

```python
# Maybe update this in `session.py`?
config_loader_args = {"base_env"="base", "default_run_env"="local"}
config_loader_args.update(settings.CONFIG_LOADER_ARGS) # Overwrite if any
return config_loader_class(
            conf_source=self._conf_source,
            env=env,
            runtime_params=extra_params,
            **config_loader_args,
        )

# Original
settings.CONFIG_LOADER_CLASS
return config_loader_class(
    conf_source=self._conf_source,
    env=env,
    runtime_params=extra_params,
)
```

I prefer 2. and we can do it now instead of waiting 0.19 cc @merelcht

_Originally posted by @noklam in https://github.com/kedro-org/kedro/issues/2819#issuecomment-1690265666_

Option 3:  
Use `_ProjectSettings` to bake framework assumption.

https://github.com/kedro-org/kedro/blob/d3171f1fc12292adc90a59fbb9ad357cbab0a906/kedro/framework/project/__init__.py#L102-L107

### Additionally
We should clean up the test for config when `local` isn't a necessary requirement. This can be consider as a separate ticket but probably best to do it together.

noklam:
## Description

Fix #2971. This enable the use of ConfigLoader as a standalone component while keeping existing behavior unchanged.

## Development notes

- [ ] Update `settings.py` to declare the defaults explicitly
- [ ] Move the `base` and `local` assumption from ConfigLoader -> `_ProjectSettings`
- [ ] Add test to make sure standalone configloader work as expected.

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -8,6 +8,7 @@
 * Added the functionality to choose a merging strategy for config files loaded with `OmegaConfigLoader`.
 * Modified the mechanism of importing datasets, raise more explicit error when dependencies are missing.
 * Added validation for configuration file used to override run commands via the CLI.
+* Moved the default environment `base` and `local` from config loader to `_ProjectSettings`. This enable the use of config loader as a standalone class without affecting existing Kedro Project users.
 
 ## Bug fixes and other changes
 * Added a new field `add-ons` to `pyproject.toml` when a project is created.


tests/framework/context/test_context.py
```diff
@@ -189,8 +195,13 @@ def extra_params(request):
...
 def dummy_context(tmp_path, prepare_project_dir, env, extra_params):
-    configure_project(MOCK_PACKAGE_NAME)
-    config_loader = OmegaConfigLoader(str(tmp_path / "conf"), env=env)
+    bootstrap_project(tmp_path)
+
+    session = KedroSession.create(
+        project_path=tmp_path, env=env, extra_params=extra_params
+    )
+    context = session.load_context()
+    config_loader = context.config_loader
```
noklam, line 204: The test was never updated after session is created. This is causing fail test because settings will not propagate  correctly if we try to construct KedroContext directly.

astrojuanlu (commented):
I confirm that this works as expected from a usage point of view:

```
$ tree                                       (kedro38-dev) 
.
└── catalog.yml

1 directory, 1 file
$ ipython --no-banner

In [1]: from kedro.config import OmegaConfigLoader

In [2]: conf_loader = OmegaConfigLoader(conf_source=".")

In [3]: conf_loader["catalog"]
Out[3]: {'test_csv': {'type': 'pandas.CSVDataset', 'filepath': 'companies.csv'}}
```

Also did a quick check on an old `spaceflights` starter and nothing broke 👍🏽

astrojuanlu (approved):

ankatiyar (approved):
LGTM! Tried it locally. Do the docs need to be updated for this too?

tests/config/test_omegaconf_config.py
```diff
@@ -978,34 +1067,25 @@ def test_runtime_params_missing_default(self, tmp_path):
...
-                default_run_env="local",
+                base_env=_BASE_ENV,
+                default_run_env=_DEFAULT_RUN_ENV,
                 runtime_params=runtime_params,
             )
-        with pytest.raises(
-            UnsupportedInterpolationType,
-            match=r"The `runtime_params:` resolver is not supported for globals.",
-        ):
-            OmegaConfigLoader(tmp_path, runtime_params=runtime_params)
```
ankatiyar, line 1008: This needs to be reverted for code coverage I think

noklam:
> LGTM! Tried it locally. Do the docs need to be updated for this too?

@ankatiyar Good point, we can do that in a separate PR. I am not sure where would be the best place to do this, maybe https://docs.kedro.org/en/stable/notebooks_and_ipython/notebook-example/add_kedro_to_a_notebook.html? 

Cc @stichbury What do you think? This related to the change of default arguments, thus make it easier to use ConfigLoader as a standalone component. Should we also update this in the [Configuration](https://docs.kedro.org/en/stable/configuration/configuration_basics.html#configuration-environments) page? The default of a Kedro project is unchanged (base and local), but it wouldn't be required if someone just import `OmegaConfigLoader`

I will fix the test coverage later since I want to wait for @merelcht review as well.

stichbury:
> Cc @stichbury What do you think? This related to the change of default arguments, thus make it easier to use ConfigLoader as a standalone component. Should we also update this in the [Configuration](https://docs.kedro.org/en/stable/configuration/configuration_basics.html#configuration-environments) page? The default of a Kedro project is unchanged (base and local), but it wouldn't be required if someone just import `OmegaConfigLoader`

Thanks for the catch @noklam -- yes, I think the best option is to update the configuration docs to highlight the change. Then illustrate its usage in the notebook example as @ankatiyar suggests.

merelcht (commented):
The implementation looks good to me! 👍 One thing I didn't immediately see was if we have test that explicitly tests different environments from `base`, `local`, and `""`. And if we have a test where we only set the `default_run_env` and not the base?

noklam:
> The implementation looks good to me! 👍 One thing I didn't immediately see was if we have test that explicitly tests different environments from base, local, and "". And if we have a test where we only set the default_run_env and not the base?

@merelcht I have added a test for the latter. One thing that I am not sure is the only `""` case.

Assuming a structure like this and the `parameters.yml` has the same content.
```
--base/parameters.yml
--local/parameters.yml
```

You can either use both environments, or move everything to base. If you choose to keep this structure while setting no environment. You get "duplicate key error" because setting `base_env="."` is going to read both files due to this  the default config pattern `**/parameters*` (match "local/parameters.yml" and "base/parameters.yml").

> E           ValueError: Duplicate keys found in /tmp/pytest-of-gitpod/pytest-5/test_load_config_only_default_0/local/parameters.yml and /tmp/pytest-of-gitpod/pytest-5/test_load_config_only_default_0/base/parameters.yml: dummy

We can avoid this by conditionally checking with something like
```python
if base_env or default_run_env:
    self.config_patterns = {
                "catalog": ["catalog*", "catalog*/**", "**/catalog*"],
                "globals": ["globals.yml"],
            }
else:
    self.config_patterns = {
                "catalog": ["catalog*", "catalog*/**",],
                "parameters": ["parameters*", "parameters*/**"],
                "credentials": ["credentials*", "credentials*/**"],
                "globals": ["globals.yml"],
            }
```

It will avoid the problem but I don't like it because it is very special to this one case. Maybe what we can do is to enhance the error message, so it tells you which "env" is causing this error, additionally printing the related "config_pattern"?

tests/config/test_omegaconf_config.py
```diff
@@ -722,9 +722,7 @@ def test_ignore_hidden_keys(self, tmp_path):
         _write_yaml(tmp_path / _BASE_ENV / "catalog1.yml", {"k1": "v1", "_k2": "v2"})
         _write_yaml(tmp_path / _BASE_ENV / "catalog2.yml", {"k3": "v3", "_k2": "v4"})
 
-        conf = OmegaConfigLoader(
-            str(tmp_path), base_env=_BASE_ENV, default_run_env=_DEFAULT_RUN_ENV
-        )
+        conf = OmegaConfigLoader(str(tmp_path), base_env=_BASE_ENV)
         conf.default_run_env = ""
```
merelcht, line 726: This line isn't needed anymore right?

merelcht:
> > The implementation looks good to me! 👍 One thing I didn't immediately see was if we have test that explicitly tests different environments from base, local, and "". And if we have a test where we only set the default_run_env and not the base?
> 
> @merelcht I have added a test for the latter. One thing that I am not sure is the only `""` case.

Aahh sorry reading my own comment again, I realise it's not very clear 😅 What I meant to say, is that I'm not sure if there's a test which checks if this all works when you have a random environment, so `base_env="my_env"` and `default_run_env="my_other_env"`. I guess this isn't actually different from "base" and "local", but might be good to have an explicit test.

tests/config/test_omegaconf_config.py
noklam: you mean ` conf.default_run_env = ""`? I have removed it.

noklam:
> Aahh sorry reading my own comment again, I realise it's not very clear 😅 What I meant to say, is that I'm not sure if there's a test which checks if this all works when you have a random environment, so base_env="my_env" and default_run_env="my_other_env". I guess this isn't actually different from "base" and "local", but might be good to have an explicit test.

@merelcht Ah, I see what you mean now. I have updated the original tests `test_load_config_only_default_run_environment` and `test_load_config_only_base_environment` for this.

merelcht (approved):
Great work! 👍

PR #3325, opened 2023-11-21
Add E2E test for tools

SajidAlamQB:
## Description
For most of the add-ons flow I've been testing the work manually: create a new project with a combination of add-ons, install requirements and try a kedro run. This process has uncovered several bugs already.

## Context
We should add automated tests to test at least some of the add-ons flow.

Implementation
At a minimum have tests for:

```
kedro new --addons=none
kedro new --addons=lint,docs,data,logging,... (everything but not viz and pyspark)
kedro new --addons=all
kedro new --addons=pyspark
kedro new --addons=viz
```
All of the above should have an example so kedro run can be tested.

## Development Notes

- Added relevant e2e tests for addons.
- Updated GitHub Actions e2e config to support running PySpark

## Checklist


merelcht (commented):
The structure of the tests looks great, but I would make the check in `check_created_project_structure_from_addons` more specific to the selected tools. So if `pyspark` is selected, also check if `spark.yml` exists and for `data` that there's a data directory etc. For viz you might just have to check the run logs for the `reporting` pipeline.

SajidAlamQB:
Having a bit of trouble getting the windows e2e tests passing on running the `PySpark` starters. I believe there's some issue with the environment wrt hadoop and PySpark. I have tried several methods to get it working but to no success does anyone have any ideas.. @merelcht @astrojuanlu @AhdraMeraliQB @ankatiyar 

Error:

```
.csv, load_args={'header': True, 'inferSchema': True}, save_args={'header': True, 'mode': overwrite, 'sep': ,}).
An error occurred while calling o86.save.
: ExitCodeException exitCode=-1073741515: 
```

AhdraMeraliQB:
QQ: How is the `--addons=all` test passing with #3334 being unresolved?

SajidAlamQB:
> QQ: How is the `--addons=all` test passing with #3334 being unresolved?

The e2e tests are running via the `--config` not the `--addons` cli flags, where the `--config` uses the same code the normal cli wizard flow uses to parse `all`. The `--addons` flag uses `_convert_addon_names_to_numbers` which is where I think the problem for `--addons=all` is coming from see, https://github.com/kedro-org/kedro/issues/3334#issuecomment-1830292861.

AhdraMeraliQB (commented):
Thanks for adding these!

Porting over our conversation offline - it would be nice to see these tests ensure that the tools we provide are ready to go for the projects we add them to

e.g. `pytest` works for the test tool, `black .` and `ruff check .` work for linting, `kedro viz` should work for viz. Docs, logs, data structure, and pyspark are fine though - I'm not sure there's much more to test for those.

What do you think?

merelcht (approved):
These tests are great, thanks for adding them @SajidAlamQB ! ⭐ 

@AhdraMeraliQB raises an interesting point, although I would personally say these are a "nice to have", because it's not technically our responsibility to ensure the behaviour of `pytest` and `ruff` and `black`. In the unit tests we check that the right requirements are added right? So then I think it's an okay assumption that when installed these work okay. But would be good to hear other opinions 🙂

ankatiyar:
^^ We could just run `pip show ruff/black/pytest` as a step to see if they were installed for the scenarios with lint/test tools but I also think these are nice to haves.

ankatiyar (approved):
LGTM! 💯

.github/workflows/e2e-tests.yml
```diff
@@ -42,6 +42,23 @@ jobs:
       - name: Add MSBuild to PATH
         if: inputs.os == 'windows-latest'
         uses: microsoft/setup-msbuild@v1
+      - name: Install Microsoft Visual C++ Redistributable
+        if: inputs.os == 'windows-latest'
+        run: |
+          choco install vcredist-all
+        shell: powershell
```
deepyaman, line 49:
```suggestion
```
Isn't this unnecessary? Default shell on Windows runners is `pwsh` anyway, which I assume should behave similarly; if it's not available, it would fall back to `powershell`. See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell.

.github/workflows/e2e-tests.yml
```diff
@@ -42,6 +42,23 @@ jobs:
...
+        run: |
+          Invoke-WebRequest "https://github.com/steveloughran/winutils/blob/master/hadoop-2.7.1/bin/winutils.exe?raw=true" -OutFile winutils.exe
+          Invoke-WebRequest "https://github.com/steveloughran/winutils/blob/master/hadoop-2.7.1/bin/hadoop.dll?raw=true" -OutFile hadoop.dll
+          Move-Item .\hadoop.dll C:\Windows\System32
+          New-Item -ItemType directory -Path C:\hadoop\bin
+          Move-Item .\winutils.exe C:\hadoop\bin
+          [System.Environment]::SetEnvironmentVariable('HADOOP_HOME', 'C:\hadoop', [System.EnvironmentVariableTarget]::Machine)
+          echo "HADOOP_HOME=C:\hadoop" | Out-File -Append -Encoding ascii -FilePath $env:GITHUB_ENV
+          echo "PATH=$env:HADOOP_HOME\bin;$env:PATH" | Out-File -Append -Encoding ascii -FilePath $env:GITHUB_ENV
+        shell: powershell
```
deepyaman, line 61:
```suggestion
```

.github/workflows/e2e-tests.yml
```diff
@@ -42,6 +42,23 @@ jobs:
...
+        if: inputs.os == 'windows-latest'
+        run: |
+          Invoke-WebRequest "https://github.com/steveloughran/winutils/blob/master/hadoop-2.7.1/bin/winutils.exe?raw=true" -OutFile winutils.exe
+          Invoke-WebRequest "https://github.com/steveloughran/winutils/blob/master/hadoop-2.7.1/bin/hadoop.dll?raw=true" -OutFile hadoop.dll
+          Move-Item .\hadoop.dll C:\Windows\System32
+          New-Item -ItemType directory -Path C:\hadoop\bin
+          Move-Item .\winutils.exe C:\hadoop\bin
+          [System.Environment]::SetEnvironmentVariable('HADOOP_HOME', 'C:\hadoop', [System.EnvironmentVariableTarget]::Machine)
+          echo "HADOOP_HOME=C:\hadoop" | Out-File -Append -Encoding ascii -FilePath $env:GITHUB_ENV
+          echo "PATH=$env:HADOOP_HOME\bin;$env:PATH" | Out-File -Append -Encoding ascii -FilePath $env:GITHUB_ENV
```
deepyaman, line 60: You're a better man than me. :clap: I tried all day to get Spark working in https://github.com/kedro-org/kedro-plugins/pull/457, but eventually gave up.

Nit: I used https://github.com/cdarlint/winutils, because it's maintained and has 2.7.3 also available (as well as newer versions). Wonder if it would work? But there's the potential drawback, that those are not signed or something, FWIW.

features/steps/cli_steps.py
```diff
@@ -453,6 +478,49 @@ def is_created(name):
...
+    if "4" in tools_list:  # docs tool
+        assert is_created("docs"), "docs directory does not exist"
+
+    if "5" in tools_list:  # data tool
+        assert is_created("data"), "data directory does not exist"
+
+    if "6" in tools_list:  # PySpark tool
+        assert is_created("conf/base/spark.yml"), "spark.yml does not exist"
+
+    if "7" in tools_list:  # 'viz' tool
```
deepyaman, line 515:
```suggestion
    if "7" in tools_list:  # viz tool
```
Nit: For consistency :)

deepyaman (approved):
Minor suggestions.

SajidAlamQB:
> it's not technically our responsibility to ensure the behaviour of `pytest` and `ruff` and `black`. In the unit tests we check that the right requirements are added right? So then I think it's an okay assumption that when installed these work okay. But would be good to hear other opinions 🙂

I agree with this as well, I feel it is a bit out of the scope of this PR.

.github/workflows/e2e-tests.yml
SajidAlamQB: Good spot.

DimedS (approved):
@SajidAlamQB , excellent work! LGTM

features/steps/cli_steps.py
```diff
@@ -453,6 +478,49 @@ def is_created(name):
...
+    # Base checks for any project
+    for path in ["README.md", "src", "pyproject.toml", "requirements.txt"]:
+        assert is_created(path), f"{path} does not exist"
+
+    tools_list = (
+        tools.split(",") if tools != "all" else ["1", "2", "3", "4", "5", "6", "7"]
+    )
+
+    if "1" in tools_list:  # lint tool
+        pass
```
DimedS, line 498: Could you explain please why 1 is empty? We cannot check file?

PR #3326, opened 2023-11-21
Explicitly set the environment defaults in `settings.py`

merelcht:
## Description
I think [Remove the environment default for ConfigLoader](https://github.com/kedro-org/kedro/issues/2971#top) might have an unexpected side effect. If you set `CONFIG_LOADER_ARGS` it will override the default set in `_ProjectSettings` and so it will override `base_env` and `default_run_env`. I noticed this because in the starters using spark we have:
```
CONFIG_LOADER_ARGS = {
#       "base_env": "base",
#       "default_run_env": "local",
    "config_patterns": {
        "spark": ["spark*", "spark*/**"],
    }
}
```
If I don’t uncomment base/default env, I will get an error saying it can’t find spark config inside myproject/conf: https://app.circleci.com/pipelines/github/kedro-org/kedro-starters/889/workflows/f1376cb4-be57-4429-8a1b-a8b9a3a31150/jobs/10413 

## Development notes
**Option 1:**
[Make sure the env defaults don't get overwritten when not explictly a…](https://github.com/kedro-org/kedro/pull/3326/commits/a6d360002db025ed03c9d3cc6550a2b13b9deaaf)
This is a fix to check inside the session if `base_env` or `default_run_env` has been specified and if not it shouldn’t overwrite the default.

**Option 2:**
https://github.com/kedro-org/kedro/pull/3326/commits/7ad31c8eeaafb42b5e7036ce987ed3b5d36104d2
Alternatively, we can just add the default `base_env` and `default_run_env` to `settings.py`. If a user removes that, it's less unexpected that this will change the behaviour IMO. That way we don't need to change the session. 

## Checklist


kedro/framework/session/session.py
```diff
@@ -253,11 +253,16 @@ def _get_config_loader(self) -> AbstractConfigLoader:
         extra_params = self.store.get("extra_params")
 
         config_loader_class = settings.CONFIG_LOADER_CLASS
+        config_loader_args = settings.CONFIG_LOADER_ARGS
+        if not (
+            "base_env" in config_loader_args or "default_run_env" in config_loader_args
+        ):
+            config_loader_args.update({"base_env": "base", "default_run_env": "local"})
```
noklam, line 260: In the original issue, I propose to use session to set the default. Technically, `_ProjectSettings` feels like the right place to set it, but I don't think there are any real difference between having the defaults in `_ProjectSettings` or `KedroSession`. I think this is fine, but basically this will always override the `_ProjectSettings` and the default there isn't relevant (either override by user provided CONFIG_LOADER_ARGS or session).

@idanov was the one pointing out that we could have the defaults in `_ProjectSettings`, so I'd love to see if you have any thought on this.

https://github.com/kedro-org/kedro/issues/2971#top
```python
# Maybe update this in `session.py`?
config_loader_args = {"base_env"="base", "default_run_env"="local"}
config_loader_args.update(settings.CONFIG_LOADER_ARGS) # Overwrite if any
return config_loader_class(
            conf_source=self._conf_source,
            env=env,
            runtime_params=extra_params,
            **config_loader_args,
        )

# Original
settings.CONFIG_LOADER_CLASS
return config_loader_class(
    conf_source=self._conf_source,
    env=env,
    runtime_params=extra_params,
)
```

merelcht: Yeah I remember.. I don't think we considered this situation though where you override some args but not explicitly the environments..

merelcht:
My preference goes to option 2 in the PR description, because IMO it’s more transparent to the user what happens here. No more magic treating of `base` and `local` environment.

DimedS (approved):
Option 2 also seems to me more clear, thanks for PR @merelcht

astrojuanlu (approved):
💯 "Explicit is better than implicit"

PR #3332, opened 2023-11-22
Remove `create_default_dataset()`

merelcht:
## Description
An alternative approach to https://github.com/kedro-org/kedro/pull/3269 for https://github.com/kedro-org/kedro/issues/2668

## Development notes

- The main difference between this PR and https://github.com/kedro-org/kedro/pull/3269 is that it removes the `create_default_dataset()` method altogether, decreasing the API surface. 
- In this approach the `shallow_copy()` method on the catalog is updated to take in optional extra dataset patterns. The default pattern is hardcoded on the runners, but can be overwritten for a custom runner.
- I've kept @ankatiyar 's change to move the `SharedMemoryDataset` from `ParellelRunner` to `io/shared_memory_dataset.py` and the change to make it inherit from `AbstractDataset`.
- I also kept the change to set the manager on the `SharedMemoryDataset` in the `ParallelRunner`.
- Removed the skipping of `test_parallel_runner.py` tests on Windows. This was put in forever ago and it was causing test coverage to not be reached on Windows. Turns out these tests are fine now 😄 
- Considered making `"{default}"` a reserved pattern name and enforcing it to always match last, but decided against it when @idanov pointed out that that would make the DataCatalog implicitly coupled with the runners by reserving the default pattern for them.
- Regarding `kedro catalog resolve`: I also considered doing some clever resolution for the default pattern in this command, but the problem is that we don't know what runner the user will use at the time of running `kedro catalog resolve`. In future, we could add a flag to resolve for a specific runner, but for now I decided to not do anything with the default pattern and added explanations about that to the docs as well as the command description.

## Checklist


kedro/io/data_catalog.py
```diff
@@ -730,16 +730,22 @@ def list(self, regex_search: str | None = None) -> list[str]:
             ) from exc
         return [dset_name for dset_name in self._datasets if pattern.search(dset_name)]
 
-    def shallow_copy(self) -> DataCatalog:
+    def shallow_copy(
+        self, default_dataset_pattern: Patterns | None = None
```
idanov, line 734:
```suggestion
        self, extra_dataset_patterns: Patterns | None = None
```

kedro/runner/sequential_runner.py
```diff
@@ -20,27 +20,17 @@ class SequentialRunner(AbstractRunner):
...
-        Args:
-            ds_name: Name of the missing data set
-
-        Returns:
-            An instance of an implementation of AbstractDataset to be used
-            for all unregistered data sets.
-
-        """
-        return MemoryDataset()
+        default_dataset_pattern = {"{default}": {"type": "MemoryDataset"}}
```
marrrcin, line 30: How can I change this if I inherit from `SequentialRunner`?

merelcht: I see what you mean. So I guess you'd prefer it if the `__init__` of all runners take the `extra_dataset_patterns` which you can then overwrite?

marrrcin: Yup, that would be nice.

kedro/io/data_catalog.py
```diff
@@ -730,16 +730,22 @@ def list(self, regex_search: str | None = None) -> list[str]:
...
+    def shallow_copy(
+        self, extra_dataset_patterns: Patterns | None = None
+    ) -> DataCatalog:
         """Returns a shallow copy of the current object.
 
         Returns:
             Copy of the current object.
         """
+        if extra_dataset_patterns:
+            dataset_patterns = {**self._dataset_patterns, **extra_dataset_patterns}
```
AhdraMeraliQB, line 742: Could we look at preventing over-writing if users have specified their own default pattern in their catalog?  Like a left join mechanism of sorts

noklam (commented):
> The main difference between this PR and [Draft] Trying to use dataset factories for default datasets  #3269 is that it removes the create_default_dataset() method altogether, decreasing the API surface.

I have a question that does it really reduce the API surface. This PR move this from a method to an Runner argument. Arguably it increases the API surface because now one can set pattern not just for "default", you can set as many as you want. Is this something expect users to do?

kedro/runner/runner.py
```diff
@@ -90,11 +95,11 @@ def run(
...
         free_outputs = pipeline.outputs() - set(registered_ds)
-        unregistered_ds = pipeline.datasets() - set(registered_ds)
 
-        # Create a default dataset for unregistered datasets
-        for ds_name in unregistered_ds:
-            catalog.add(ds_name, self.create_default_dataset(ds_name))
+        # Register the default dataset pattern with the catalog
+        catalog = catalog.shallow_copy(
+            extra_dataset_patterns=self._extra_dataset_patterns
+        )
```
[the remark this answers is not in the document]
merelcht: If this is moved up we can't accurately determine the `unsatisfied` inputs that aren't defined in the catalog and don't match a pattern or the `free_outputs` because both would be captured under the default dataset which isn't supposed to happen at that stage yet.

kedro/io/memory_dataset.py
```diff
@@ -136,3 +138,52 @@ def _copy_with_mode(data: Any, copy_mode: str) -> Any:
         )
 
     return copied_data
+
+
+class SharedMemoryDataset(AbstractDataset):
```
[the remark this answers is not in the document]
merelcht: Done!

kedro/runner/sequential_runner.py
merelcht: Done!

kedro/runner/parallel_runner.py
```diff
@@ -244,6 +199,16 @@ def _validate_catalog(cls, catalog: DataCatalog, pipeline: Pipeline):
                 f"MemoryDatasets"
             )
 
+    def _set_manager_datasets(self, catalog, pipeline):
+        for dataset in pipeline.datasets():
+            try:
+                catalog._get_dataset(dataset)
+            except DatasetNotFoundError:
+                pass
```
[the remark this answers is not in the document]
merelcht: `if dataset in catalog` wouldn't catch any datasets that match a pattern, because those are only materialised at access (load/save/exists). But I'll improve the rest.

ankatiyar:
The new changes look good to me, though, I think we need to either prohibit the use of `"{default}"` pattern in the catalog by erroring out. And the dataset factory CLI commands need to be updated. If the user has "{default}" pattern in their catalog - `kedro catalog resolve` will still consider it but during a session run, the catch-all pattern isn't used.

merelcht:
> The new changes look good to me, though, I think we need to either prohibit the use of `"{default}"` pattern in the catalog by erroring out. And the dataset factory CLI commands need to be updated. If the user has "{default}" pattern in their catalog - `kedro catalog resolve` will still consider it but during a session run, the catch-all pattern isn't used.

Done in [d3cece8](https://github.com/kedro-org/kedro/pull/3332/commits/d3cece82f9ab618ed61658ccf9a2c83734f2c410)

kedro/io/shared_memory_dataset.py
```diff
@@ -0,0 +1,54 @@
...
+                pickle.dumps(data)
+            except Exception as serialisation_exc:  # SKIP_IF_NO_SPARK
+                raise DatasetError(
+                    f"{str(data.__class__)} cannot be serialised. ParallelRunner "
+                    "implicit memory datasets can only be used with serialisable data"
+                ) from serialisation_exc
+            raise exc  # pragma: no cover
+
+    def _describe(self) -> dict[str, Any]:
+        return {"shared_memory_dataset": self.shared_memory_dataset}
```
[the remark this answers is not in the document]
idanov: We can return just an empty dictionary then.

idanov (approved):

kedro/io/shared_memory_dataset.py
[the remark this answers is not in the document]
merelcht: Done

Issue #3334, opened 2023-11-23
`--addons=all` doesn't behave as expected


merelcht:
## Description
When I use `--addons=all` or `--addons all` in `kedro new` directly I don't get all addons as expected.
You can also see in the message: 
```
You have selected the following add-ons:
```

## Steps to Reproduce
1. Checkout the `develop` branch
2. Run `kedro new --addons=all` or `kedro new --addons all`
3. Follow the rest of the prompts adding the name and example yes or no
4. Check the resulting project

```
(dev) M-C02XN0V5JHD5:Testing merel_theisen$ kedro new --addons=all --example=yes --name=allbug

You have selected the following add-ons:

The project name 'allbug' has been applied to:
- The project title in /Users/merel_theisen/Projects/Testing/allbug/README.md
- The folder created for your project in /Users/merel_theisen/Projects/Testing/allbug
- The project's python package in /Users/merel_theisen/Projects/Testing/allbug/src/allbug

A best-practice setup includes initialising git and creating a virtual environment before running 'pip install -r requirements.txt' to install project-specific dependencies. Refer to the Kedro documentation: https://kedro.readthedocs.io/

Change directory to the project generated in /Users/merel_theisen/Projects/Testing/allbug by entering 'cd /Users/merel_theisen/Projects/Testing/allbug'
```
## Expected Result
The resulting project should have tests folder and requirements for linting, testing and logging etc. 

## Actual Result
It doesn't have all the expected files and folder structures

❗ If I follow the prompts when doing `kedro new` and use `all` the project gets created as expected.

merelcht:
Should also test: `kedro new --addons=lint,log,data,docs,test,viz,pyspark`

lrcouto:
> Should also test: `kedro new --addons=lint,log,data,docs,test,viz,pyspark` and why didn't the test pick this up?

just tested it and it generates the files, the issue is happening only with the `all` option. Looking into it now.

SajidAlamQB:
Completed in: https://github.com/kedro-org/kedro/pull/3361

PR #3361, opened 2023-11-29
Fix `--tools=all` not working properly

SajidAlamQB:
## Description

Related to: https://github.com/kedro-org/kedro/issues/3334

When you use `--tools=all` or `--tools all` in kedro new directly you don't get all addons as expected.
You can also see in the message:

`You have selected the following tools:`

## Development notes

The problem looks to be coming from `_convert_tool_names_to_numbers`. https://github.com/kedro-org/kedro/blob/7d295e878e2b663a0f24d51685de95c1656a5632/kedro/framework/cli/starters.py#L469-L490

We don't have a check for the "all" flag. We should probably also have "none" check here as well.

## Checklist


merelcht (commented):
Do we have any tests that should have caught this?

deepyaman:
> Do we have any tests that should have caught this?

+1 was going to say the same thing, then saw @merelcht already did :D

SajidAlamQB:
> Do we have any tests that should have caught this?

Yes, I believe we have the tests already, @lrcouto, found that this problem wasn't originally picked up by the tests as it was using the faulty function which shouldn't be a problem now that it's fixed. Right @lrcouto?

kedro/framework/cli/starters.py
```diff
@@ -481,6 +481,10 @@ def _convert_tool_names_to_numbers(selected_tools: str | None) -> str | None:
     """
     if selected_tools is None:
         return None
+    if selected_tools.lower() == "none":
+        return ""
+    if selected_tools.lower() == "all":
+        return ",".join(TOOLS_SHORTNAME_TO_NUMBER.values())
```
DimedS, line 487: Maybe we just need to update the TOOLS_SHORTNAME_TO_NUMBER dictionary? Put "none" and "all" there? Or it will break something else?

AhdraMeraliQB:
> > Do we have any tests that should have caught this?
> 
> Yes, I believe we have the tests already, @lrcouto, found that this problem wasn't originally picked up by the tests as it was using the faulty function which shouldn't be a problem now that it's fixed. Right @lrcouto?

😬 The unchecked point of failure is a little concerning 

I noticed in `test_starters.py` we only really test the helper function `_parse_tools_input`. I think we can mitigate something like this again by testing the others as well - would it be in scope to add some quick unit tests for `_parse_yes_no_to_bool` and `_validate_selection`, as well as `_convert_tools_names_to_numbers`?

DimedS (approved):
Thanks, @lrcouto and @SajidAlamQB , great work!

tests/framework/cli/test_starters.py
```diff
@@ -1324,3 +1324,92 @@ def test_invalid_names(self, fake_kedro_cli, name):
             "Kedro project names must contain only alphanumeric symbols, spaces, underscores and hyphens and be at least 2 characters long"
             in result.output
         )
+
+
+class TestParseYesNoToBools:
```
AhdraMeraliQB, line 1329: Could we move the tests for `parse_tools_input` into their own class as well? For consistency

tests/framework/cli/test_starters.py
```diff
@@ -1324,3 +1324,92 @@ def test_invalid_names(self, fake_kedro_cli, name):
...
+
+class TestValidateSelection:
+    def test_validate_selection_valid(self):
+        tools = ["1", "2", "3", "4"]
+        assert _validate_selection(tools) is None
+
+    def test_validate_selection_invalid_single_tool(self):
+        tools = ["8"]
+        with pytest.raises(SystemExit):
+            _validate_selection(tools)
```
AhdraMeraliQB, line 1356: Could we add an assert for the correct error message?

AhdraMeraliQB (commented):
Love the new tests, just a few comments:

kedro/framework/cli/starters.py
```diff
@@ -489,6 +489,10 @@ def _convert_tool_names_to_numbers(selected_tools: str | None) -> str | None:
...
```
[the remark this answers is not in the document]
lrcouto: Changed it. Thanks!

merelcht (approved):
LGTM! 👍

PR #3426, opened 2023-12-14
Tools: Minor issues/improvements to `pyproject.toml`

SajidAlamQB:
## Description

Context: #3333

## Development notes

- QoL changes and adding examples to `pyproject.toml`

## Developer Certificate of Origin
We need all contributions to comply with the [Developer Certificate of Origin (DCO)](https://developercertificate.org/). All commits must be signed off by including a `Signed-off-by` line in the commit message. [See our wiki for guidance](https://github.com/kedro-org/kedro/wiki/Guidelines-for-contributing-developers/).

If your PR is blocked due to unsigned commits, then you must follow the instructions under "Rebase the branch" on the GitHub Checks page for your PR. This will retroactively add the sign-off to all unsigned commits and allow the DCO check to pass.

## Checklist

- [x] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [x] Signed off each commit with a [Developer Certificate of Origin (DCO)](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository)
- [x] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [ ] Added tests to cover my changes
- [x] Checked if this change will affect Kedro-Viz, and if so, communicated that with the Viz team

PR #3462, opened 2023-12-22
Add tests to all new project creation flows

AhdraMeraliQB:
## Description

I noticed after #3426 supplying `--tools=none` to kedro new no longer resulted in the message "You have selected no project tools". This PR fixes this, and adds automated tests for this test case, and for providing tools through the config file, both of which were previously untested.

## Checklist


kedro/framework/cli/starters.py
```diff
@@ -849,7 +849,7 @@ def _create_project(template_path: str, cookiecutter_args: dict[str, Any]):
 
     # we can use starters without tools:
     if tools is not None:
-        if tools in ("[]", ""):  # TODO: This should be a list
+        if tools in ("['None']", "[]"):  # TODO: This should be a list
```
noklam, line 852: Where is this ["None"] come from? At the very end of the process I expected all the tool should be converted to a list of names that is ready to be copy to `pyproject.toml`.

What's the expected result of for the following cases? 
1. `kedro new --tools=none
2. `kedro new` -> Prompt enter
```
# pyproject.toml
tools = 
```

DimedS: Looks that right now it will be "[]" after Prompt and ['None'] after CLI. We should unify that

AhdraMeraliQB: It comes from starters.py L521;

Previously, when storing the selected tools as a string, when the input "none" was parsed through CLI or config, an empty list wrapped in a string would be returned `"[]"` , and if the selection for no tools was made through the prompts, an empty string `""` would be returned instead. 

We now store the tools selection by passing through to cookiecutter a list wrapped in a string, that we then unwrap in the template `pyproject.toml`. This resulted in errors as unwrapping an empty string (from no tools selection) would result in an invalid toml file.

The implementation results in two different values depending on how the selection for no tools is made, after #/3426 the values are as such:
1. "None" is provided through the flag --tools, or config -> the result of parsing this selection will end in L521 in starters.py storing tools as str(['None'])
2. Selecting "none" through the interactive flow -> the prompt will be parsed by `_parse_tools_input` instead, and will be registered as an empty list `[]`, which is then wrapped in a string and stored as "[ ]"

As such, after #/3426, the condition has changed from `("[]", "")` to `("['None']", "[]")`

Undoubtedly this code requires some refactoring 😬; I am happy to make changes to `_parse_tools_input` within this PR, but an overarching refactor (or code evaluation) is still necessary

DimedS: Thank you, @AhdraMeraliQB . The current situation with the tool's processing is really unclear. I agree it's a good idea to add that refactoring task to minor tech debts ticket.

kedro/framework/cli/starters.py
```diff
@@ -516,9 +516,10 @@ def _get_extra_context(  # noqa: PLR0913
             NUMBER_TO_TOOLS_NAME[tool]
             for tool in _parse_tools_input(tools)  # type: ignore
         ]
-        extra_context["tools"] = str(extra_context["tools"])
+    if extra_context.get("tools"):
```
AhdraMeraliQB, line 519: Extra condition to catch: 
- User prompts tools:"none"

merelcht: What does this additional check do on top of this check above?

```python
 tools = extra_context.get("tools")
    if tools:
```

merelcht (commented):
This mostly looks good to me, just one comment about the extra `if statement`

kedro/framework/cli/starters.py
AhdraMeraliQB: If the user goes through the prompts and selects none, `_parse_tools_input(tools)` on line 517 will return an empty list, which was causing the discrepancy of none resulting in either `[]` or `['None']` being returned. This additional check ensures this is handled without adding `{"none":'None'} `to `NUMBER_TO_TOOLS_NAME` which felt incongruous (as "none" isn't a number or tool selection)

DimedS: I think I understand the idea, but it seems somewhat tricky and not immediately clear, despite its functionality. Perhaps we should consider little bit refactoring lines 513-521 to clarify the logic?

AhdraMeraliQB: I agree, especially because this is where the three streams of input converge. I have tried to clarify in more procedural detail:

There are several ways of selecting no tools:
1. Using `--config` and omitting `tools:` from config
2. ` kedro new --tools=none`
3. User prompts and using the default value or inputing none
4. Config file entry `tools: "none"`

The condition at line 514, is triggered by cases 3 and 4, which will result in `extra_context['tools'] = []`.
Line 519 then checks against this, only converting the list to a string to be stored if the list is **not** empty. 
All cases 1-4 will skip to line 521 as `extra_context['tools']` will either be `None` or an empty list `[]`, unifying the stored result as `['None']`

I am unsure we could refactor only these lines without affecting the rest of the file, did you have any suggestions?

DimedS: @AhdraMeraliQB , I agree, it's really a lot of ways to select tools.
And I also agree that we should split some code before lines 513-521. I think we should clarify steps:
1. Finalise the value of extra_context["tools"] as a list of numbers, or [] by default anyway. It should be done before lines 513.
2. Clarify that now in lines 513 we want to convert that list of numbers to list of Readable names and [] (empty list) will be converted to a ["None"]

So split the collection and merging step from conversion step.

merelcht: I would be in favour of doing the refactoring in this PR, because the change introduced here makes the code even harder to follow. I understand why it's needed now, but it's hard to grasp it just from reading the code.

noklam (approved):

AhdraMeraliQB:
Unifying `extra_context["tools"]="none"` and `extra_context["tools"] = None` requires the logic on starters to be updated, which has been done in [this PR](https://github.com/kedro-org/kedro-starters/pull/204) - note that tests on this branch will fail until that PR is merged. Appended below is the PR description:

> When a starter is selected, any removal of tools should not be applied to the template project. To prevent this, the post project generation hook would check that tools have been specified to gauge if this was a starter selection or example pipeline selection (where the latter would continue with the removal process).
> 
> However, this overlooked the case of no tools selected and an example pipeline being requested - the resulting project should have the unwanted tools stripped of it, but instead the project template was treated like a starter and no modifications were made. This is addressed by testing for an example-pipeline instead of a tools selection - this check is more robust, particularly because example-pipeline cannot be true when a starter flag is used.

AhdraMeraliQB:
@noklam re-requested review due to logic refactor

merelcht (approved):
Thanks @AhdraMeraliQB, this is much clearer now! 👍

AhdraMeraliQB:
[Here are the changes](https://github.com/kedro-org/kedro/pull/3462/files/445e28e55d6f7604a4e9b7b34e1eb3c3b3c6d4c4..c909055a69bc41ed717aeaf987e3a215d648242e) between the refactor and the string wrapper, had to revert for testing purposes, but no logic changes (just linting)

DimedS (approved):
@AhdraMeraliQB thank you, amazing work!

DimedS:
@AhdraMeraliQB , looks like now we receive a message "You have selected no project tools" even if using just a starter, where no tools needed.

AhdraMeraliQB:
Interesting...

PR #3469, opened 2024-01-02
Format plugin name to fix CI

### LINKED ISSUE #3461: ci: Nightly build failure on `main`
GitHub Actions workflow [Run nightly tests on Kedro #98](https://github.com/kedro-org/kedro/actions/runs/7294464197) failed.

AhdraMeraliQB:
## Description

Resolves #3461 and #3464 

## Development notes

Tests were failing because they were attempting to disable `test-plugin`, however within #3329 the plugin name was changed to `test_plugin` and therefore was not being properly disabled. In this PR the name is change back, following package naming convention and fixing the e2e tests.

## Checklist

### DIFF
--- a/features/steps/test_plugin/pyproject.toml
+++ b/features/steps/test_plugin/pyproject.toml
@@ -1,10 +1,10 @@
 [project]
-name = "test_plugin"
+name = "test-plugin"
 description = "Dummy plugin with hook implementations and custom starters"
 version = "0.1"
 
 [project.entry-points."kedro.hooks"]
-test_plugin = "plugin:hooks"
+test-plugin = "plugin:hooks"
 
 [project.entry-points."kedro.starters"]
 starter = "plugin:starters"
--- a/features/steps/test_plugin/pyproject.toml
+++ b/features/steps/test_plugin/pyproject.toml
@@ -4,7 +4,7 @@ description = "Dummy plugin with hook implementations and custom starters"
 version = "0.1"
 
 [project.entry-points."kedro.hooks"]
-test-plugin = "plugin:hooks"
+test_plugin = "plugin:hooks"
 
 [project.entry-points."kedro.starters"]
 starter = "plugin:starters"


merelcht (commented):
It's surprising to me that this change fixes things, because it seems to be the opposite of what we describe in the docs: https://docs.kedro.org/en/stable/extend_kedro/plugins.html#cli-hooks

DimedS:
> It's surprising to me that this change fixes things, because it seems to be the opposite of what we describe in the docs: https://docs.kedro.org/en/stable/extend_kedro/plugins.html#cli-hooks

Looks like we should modify our codes to use "test_plugin" name, not revert name back

AhdraMeraliQB:
CC: @DimedS @merelcht 

There are two avenues for the fix, either changing the name to match what the test looks for, or modify the test to look for the different name. I went for the former as using a hyphen instead of an underscore seems to match the convention we use in for our plugins [kedro-airflow](https://github.com/kedro-org/kedro-plugins/blob/2928ccab7d8ab45179f06abeaebd5116731cfe95/kedro-airflow/pyproject.toml#L6), [kedro-datasets](https://github.com/kedro-org/kedro-plugins/blob/2928ccab7d8ab45179f06abeaebd5116731cfe95/kedro-datasets/pyproject.toml#L6), [kedro-docker](https://github.com/kedro-org/kedro-plugins/blob/2928ccab7d8ab45179f06abeaebd5116731cfe95/kedro-docker/pyproject.toml#L6), and [kedro-telemetry](https://github.com/kedro-org/kedro-plugins/blob/2928ccab7d8ab45179f06abeaebd5116731cfe95/kedro-telemetry/pyproject.toml#L6), but happy to go with the latter.

AhdraMeraliQB:
> It's surprising to me that this change fixes things, because it seems to be the opposite of what we describe in the docs: https://docs.kedro.org/en/stable/extend_kedro/plugins.html#cli-hooks

I don't see how our docs conflict with this change, it's ultimately (I believe) a choice of formatting, and the docs make no mention of `settings.py` as far as I can see, which is where we reference the plugin name in our tests

DimedS:
> CC: @DimedS @merelcht
> 
> There are two avenues for the fix, either changing the name to match what the test looks for, or modify the test to look for the different name. I went for the former as using a hyphen instead of an underscore seems to match the convention we use in for our plugins [kedro-airflow](https://github.com/kedro-org/kedro-plugins/blob/2928ccab7d8ab45179f06abeaebd5116731cfe95/kedro-airflow/pyproject.toml#L6), [kedro-datasets](https://github.com/kedro-org/kedro-plugins/blob/2928ccab7d8ab45179f06abeaebd5116731cfe95/kedro-datasets/pyproject.toml#L6), [kedro-docker](https://github.com/kedro-org/kedro-plugins/blob/2928ccab7d8ab45179f06abeaebd5116731cfe95/kedro-docker/pyproject.toml#L6), and [kedro-telemetry](https://github.com/kedro-org/kedro-plugins/blob/2928ccab7d8ab45179f06abeaebd5116731cfe95/kedro-telemetry/pyproject.toml#L6), but happy to go with the latter.

Yes, you right, that package naming is our common situation, but I think it's better to leave entry point "test_plugin", I think previously it was the same.

DimedS (approved):
Thank you, @AhdraMeraliQB ! I think we should fix only name, not entry-point?

merelcht:
> > It's surprising to me that this change fixes things, because it seems to be the opposite of what we describe in the docs: https://docs.kedro.org/en/stable/extend_kedro/plugins.html#cli-hooks
> 
> I don't see how our docs conflict with this change, it's ultimately (I believe) a choice of formatting, and the docs make no mention of `settings.py` as far as I can see, which is where we reference the plugin name in our tests

What I meant is that in the docs we talk about a plugin called "plugin-name" e.g. "my-plugin", should then be referenced in the `pyproject.toml` as:

```yml
[project.entry-points."kedro.cli_hooks"]
plugin_name = "plugin_name.plugin:cli_hooks"
```

So for "my-plugin" this should be:
```
[project.entry-points."kedro.cli_hooks"]
my_plugin = "plugin_name.plugin:cli_hooks"
```

So the change here to
```
[project.entry-points."kedro.hooks"]
test-plugin = "plugin:hooks"
```

seems conflicting to me.

AhdraMeraliQB:
> > > It's surprising to me that this change fixes things, because it seems to be the opposite of what we describe in the docs: https://docs.kedro.org/en/stable/extend_kedro/plugins.html#cli-hooks
> > 
> > 
> > I don't see how our docs conflict with this change, it's ultimately (I believe) a choice of formatting, and the docs make no mention of `settings.py` as far as I can see, which is where we reference the plugin name in our tests
> 
> What I meant is that in the docs we talk about a plugin called "plugin-name" e.g. "my-plugin", should then be referenced in the `pyproject.toml` as:
> 
> ```yaml
> [project.entry-points."kedro.hooks"]
> plugin_name = "plugin_name.plugin:hooks"
> ```
> 
> So for "my-plugin" this should be:
> 
> ```
> [project.entry-points."kedro.hooks"]
> my_plugin = "my_plugin.plugin:hooks"
> ```
> 
> So the change here to
> 
> ```
> [project.entry-points."kedro.hooks"]
> test-plugin = "plugin:hooks"
> ```
> 
> seems conflicting to me.

Interesting, it works both ways, I've changed it back to align with the docs but it's intriguing that there's no consistency requirement. Maybe @astrojuanlu can provide more insight on how toml processes this?

merelcht (approved):
Thanks for fixing this @AhdraMeraliQB ! ⭐  

I agree it would be good to clarify if there's a best practice when it comes to the naming convention for the entry point in `pyproject.toml` and make sure we adopt it consistently in our docs and code.

PR #3481, opened 2024-01-04
Unpin `fsspec`

### LINKED ISSUE #3102: Unpin `fsspec` in test requirements
## Description
In the migration of test requirements from `setup.py` to `pyproject.toml` (See #3078), the version of `fsspec` was pinned to `<2023.9` in the test requirements. (on the `develop` branch)

The latest versions of `fsspec` are causing the test "`test_no_versions_with_cloud_protocol`" in `tests/io/test_data_catalog.py` to fail. 

This is a follow up ticket to investigate the test failure and remove the `fsspec` pin from the test requirements.

merelcht:
## Description
Fixes #3102 

## Development notes
In the newer `fsspec` versions permissions are checked at a different stage so the following line already threw an error: 
`The AWS Access Key Id you provided does not exist in our records.`

https://github.com/kedro-org/kedro/blob/main/kedro/io/core.py#L539
```python 
version_paths = sorted(self._glob_function(pattern), reverse=True)
```

I added a try except to handle this.

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,7 +58,7 @@ test = [
     "blacken-docs==1.9.2",
     "black~=22.0",
     "coverage[toml]",
-    "fsspec<2023.9", # Temporary, newer version causing "test_no_versions_with_cloud_protocol" to fail
+    "fsspec>=2021.4",
     "import-linter==1.12.1",
     "ipython>=7.31.1, <8.0; python_version < '3.8'",
     "ipython~=8.10; python_version >= '3.8'",
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -536,7 +536,14 @@ class AbstractVersionedDataset(AbstractDataset[_DI, _DO], abc.ABC):
         # When load version is unpinned, fetch the most recent existing
         # version from the given path.
         pattern = str(self._get_versioned_path("*"))
-        version_paths = sorted(self._glob_function(pattern), reverse=True)
+        try:
+            version_paths = sorted(self._glob_function(pattern), reverse=True)
+        except Exception as exc:
+            message = (
+                f"Did not find any versions for {self}. This could be "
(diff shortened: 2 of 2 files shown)


noklam:
Why do we want to catch the error if it explicitly saying it's a permission error?

merelcht:
> Why do we want to catch the error if it explicitly saying it's a permission error?

My thinking was that `fsspec` is used for all filesystem types so it could be a different error is thrown if it's not a cloud system.

pyproject.toml
```diff
@@ -58,7 +58,7 @@ test = [
     "blacken-docs==1.9.2",
     "black~=22.0",
     "coverage[toml]",
-    "fsspec<2023.9", # Temporary, newer version causing "test_no_versions_with_cloud_protocol" to fail
+    "fsspec>=2021.4",
```
AhdraMeraliQB, line 61: Is the new error handling compatible with `2021.4<=version<2023.9`?

AhdraMeraliQB (approved):
LGTM! Just a QQ:

pyproject.toml
merelcht: In the older versions the error will be caught later on in this check https://github.com/kedro-org/kedro/blob/main/kedro/io/core.py#L544 

It's not possible to keep it fully backwards compatible, because the test-coverage would then fail and in different ways depending on which version of `fsspec` is installed. It's just a test-requirement though, so backward compatibility isn't as important as for the regular dependencies.

AhdraMeraliQB: Could we then pin the test requirement at `>2023.9`?

merelcht, line 61:
```suggestion
```

merelcht: I ended up just removing `fsspec` from the test requirements, because it wasn't included before the pin was needed.

SajidAlamQB (approved):
LGTM 👍

PR #3484, opened 2024-01-04
Check if starter is used before printing tools message

AhdraMeraliQB:
## Description

As pointed out by @DimedS: [xref](https://github.com/kedro-org/kedro/pull/3462#issuecomment-1877214413)

We don't allow users to select tools when using starters, so the message "you have selected no project tools" should not be displayed if a starter is used. This check was previously implicit, relying on the different values of tools to differentiate (previously `'[]'` and `['None']`). These have since been unified, and as such, an explicit condition is required to ensure the message isn't printed when a starter is used.

## Development notes

There are several alternatives for implementing this:
1. Move up the tool and example selection messages from L865 to L301, to access `starter_alias`  _<- current approach_
2. Pass `starter_alias` through to `_create_project` as an argument
3. Pass `starter_alias` through as part of the `extra_config` to be accessed in `_create_project` _<- original approach_
4. Overhaul and refactor `starters.py` 

## Checklist


kedro/framework/cli/starters.py
```diff
@@ -851,9 +851,12 @@ def _create_project(template_path: str, cookiecutter_args: dict[str, Any]):
         f"\nYour project '{project_name}' has been created in the directory \n{result_path}\n"
     )
 
-    # we can use starters without tools:
+    # End here if a starter was used
+    if template_path != TEMPLATE_PATH:
+        return
+
     if tools is not None:
-        if tools == "['None']":  # TODO: This should be a list
```
AhdraMeraliQB, line 856: Removed the comment as the string to list conversion is done later in the execution timeline, and the final item stored in `pyproject.toml` is indeed a list object. This conversion cannot happen before the cookiecutter generation, as `cookiecutter` does not store lists but rather parses them as a selection of options, only storing one of the values within the list.

kedro/framework/cli/starters.py
```diff
@@ -851,9 +851,12 @@ def _create_project(template_path: str, cookiecutter_args: dict[str, Any]):
         f"\nYour project '{project_name}' has been created in the directory \n{result_path}\n"
     )
 
-    # we can use starters without tools:
+    # End here if a starter was used
+    if template_path != TEMPLATE_PATH:
+        return
+
     if tools is not None:
```
AhdraMeraliQB, line 858: I am unsure if tools can ever be `None`, but have left this condition as a guardrail

kedro/framework/cli/starters.py
```diff
@@ -498,6 +502,9 @@ def _get_extra_context(  # noqa: PLR0913
             prompts_required, cookiecutter_context
         )
 
+    if starter_alias:
+        extra_context["starter_alias"] = starter_alias
```
DimedS, line 506: I'm not sure that it's a good idea to put starter_alias to the extra_context, because the extra context dict will be sent to cookiecutter where starter_alias is not needed?

AhdraMeraliQB: That's true, though this doesn't have any consequence I don't think it's the optimal approach here. Of the alternatives I do think it's the most minimal, but would opt for 1 in its stead. What do you think?

DimedS: Yes, I saw it was already done with some message for the same reason in lines 302-307), so it's not critical to do one more.
But maybe for consistency we should wrap all that messages that we throw after project creation in one func.

merelcht:
The `new` function has become quite long. Maybe we can refactor this a bit. For example, extract lines 250-261 https://github.com/kedro-org/kedro/blob/30ce26c4a72532bdeba29bd047e745745d835cab/kedro/framework/cli/starters.py#L250-L261 
into a separate function that validates the chosen arguments.

merelcht (commented):
The approach looks good to me 👍 I left some comments to tidy the code some more.

kedro/framework/cli/starters.py
```diff
@@ -299,7 +323,17 @@ def new(  # noqa: PLR0913
...
 
-    if prompts_required and not config_path and not starter_alias:
+    # End here if a starter was used
+    if starter_alias:
+        return
+
+    # If not a starter, print tools and example selection
+    if not starter_alias:
+        _print_tools_selection(extra_context.get("tools"))
+        _print_example_selection(extra_context.get("example_pipeline"))
```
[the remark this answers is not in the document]
AhdraMeraliQB: Addressed in [7160bcd](https://github.com/kedro-org/kedro/pull/3484/commits/7160bcd3246fdecc46c3fefea40ba1f6858bd8e4)

kedro/framework/cli/starters.py
```diff
@@ -48,6 +47,8 @@
 kedro new --tools=lint,test,log,docs,data,pyspark,viz (or any subset of these options)\n
 kedro new --tools=all\n
 kedro new --tools=none
+
+For more information on using tools, see https://docs.kedro.org/en/latest/starters/new_project_tools.html
```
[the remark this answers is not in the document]
AhdraMeraliQB: Addressed in [5500b95](https://github.com/kedro-org/kedro/pull/3484/commits/5500b956ab5e5e35da0b647a8a6da8c369cbb8bc)

kedro/framework/cli/starters.py
```diff
@@ -187,6 +188,29 @@ def _validate_selected_tools(selected_tools):
             sys.exit(1)
 
 
+def _print_tools_selection(selected_tools):
```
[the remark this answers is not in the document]
AhdraMeraliQB: Added in [b2d92f5](https://github.com/kedro-org/kedro/pull/3484/commits/b2d92f58bab8f87b549aea7b36a8d9773555e703)

kedro/framework/cli/starters.py
```diff
@@ -187,6 +188,29 @@ def _validate_selected_tools(selected_tools):
...
+                fg="green",
+            )
+        else:
+            click.secho(
+                f"You have selected the following project tools: {selected_tools}",
+                fg="green",
+            )
+
+
+def _print_example_selection(example_pipeline):
```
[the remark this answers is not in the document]
AhdraMeraliQB: Added in same commit, [b2d92f5](https://github.com/kedro-org/kedro/pull/3484/commits/b2d92f58bab8f87b549aea7b36a8d9773555e703)

kedro/framework/cli/starters.py
```diff
@@ -223,19 +264,16 @@ def new(  # noqa: PLR0913
...
-        raise KedroCliError(
-            "Cannot use the --directory flag without a --starter value."
-        )
-
-    if (selected_tools or example_pipeline) and starter_alias:
-        raise KedroCliError(
-            "Cannot use the --starter flag with the --example and/or --tools flag."
-        )
-
+    flag_inputs = {
```
merelcht, line 267: Why create a dict and not just directly pass the arguments to `_validate_flag_inputs`?

merelcht (commented):
Some minor comments from my side, but in general this is looking good!

kedro/framework/cli/starters.py
```diff
@@ -299,7 +337,16 @@ def new(  # noqa: PLR0913
 
     _create_project(project_template, cookiecutter_args)
 
-    if prompts_required and not config_path and not starter_alias:
+    # End here if a starter was used
+    if starter_alias:
```
DimedS, line 341: @AhdraMeraliQB, thank you for the excellent work, the code is much clearer now! 
Also, to further streamline our approach, perhaps we could consider this modification:
```
if not starter_alias:
   print_selection_and_prompt_info()
```
By consolidating all printing operations into a single function, I believe it will make the overall flow in the main create_cli() function easier to understand."

kedro/framework/cli/starters.py
DimedS: I think it's in general a good idea - to put all CLI inputs to one dictionary, because currently there are a lot of them, but it should be done not only for one _validate_flag_inputs() function, but to the whole subsequent code

kedro/framework/cli/starters.py
```diff
@@ -130,6 +131,23 @@ class KedroStarterSpec:  # noqa: too-few-public-methods
 }
 
 
+def _validate_flag_inputs(flag_inputs: dict[str, Any]):
```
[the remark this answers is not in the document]
AhdraMeraliQB: Addressed in [3c72ba5](https://github.com/kedro-org/kedro/pull/3484/commits/3c72ba5e8201eea163627f37b345247b2477d9ca)

kedro/framework/cli/starters.py
AhdraMeraliQB: Done in [12ba18c](https://github.com/kedro-org/kedro/pull/3484/commits/12ba18cd82766ccc36cf0fecbe69ff081549ef0d) !

kedro/framework/cli/starters.py
AhdraMeraliQB: I agree, the initial reason was due to how many inputs there are, and for future implementation/refactor as there is no guarantee that we won't need to pass these through later. I couldn't find any strong support for either convention, however, so I am mostly ambivalent. Though whatever we do choose, we should be sure to implement consistently.

DimedS (approved):
Thank you, @AhdraMeraliQB , awesome job! 👍

Issue #2156, opened 2022-12-22
Revisit our `mypy` setup


idanov:
## Description

Currently we are using `mypy` only in our `pre-commit` hook and we even don't have it as an explicit `test_requirements.txt` dependency. If you are using `pyright` as a language server to explore Kedro's codebase, you will see a lot of type errors and warnings that you wouldn't catch through our pre-commit hook. Also Python's type annotations have evolved since we first started using it and we should catch up with these developments, e.g. the change of PEP 484 which disallows now inexplicit optional definitions:

```
session_id: str = None
```
now has to become:

```
session_id: Optional[str] = None
```

The goal of this is to end up as close as possible to be able to run `mypy --strict` on our codebase without any complains. 

## Context

To improve the quality of our codebase and possibly uncover some bugs hidden bugs. 

## Possible Implementation

- Ensure `pre-commit` and running `mypy` locally produce the same output
- Bump up to the latest `mypy` version in dev requirements (`test_requirements.txt`)
- Run `mypy --strict` and fix as much errors as possible
- Configure `mypy` to ignore the errors we cannot fix, e.g. the ones introduced to us by third-party packages

Note: don't run `mypy` on `kedro.extras` because we're removing the datasets in 0.19.0

## Possible Alternatives

Another option is to move to `pyright` entirely and use it as a type checker instead of `mypy`, or maybe even use both?

noklam:
Second to that, PyLance (which is better than PyRight) have built-in extension for these type hints which is better. Our mypy is very old (v0.961). Simply bumping it to 1.4.1 I get 267 errors. Not all of the errors are important but I think it helps a lot when we get these type correct for the functions that nested in many layers. It will also give us better insight to refactor.

```
kedro/ipython/__init__.py:78: note: Use https://github.com/hauntsaninja/no_implicit_optional to automatically upgrade your codebase
Found 283 errors in 57 files (checked 268 source files)
```

astrojuanlu:
Echoing a comment I made somewhere else about MyPy and pre-commit, see https://jaredkhan.com/blog/mypy-pre-commit

An alternative is to use it as a pytest plugin rather than a linter https://github.com/realpython/pytest-mypy or just call it separately.

merelcht:
Resolved in the following PRs:
1. #3485 
2. #3490 
3. #3497 
4. #3498

PR #3485, opened 2024-01-05
Revisit `mypy` setup

merelcht:
## Description
Part of https://github.com/kedro-org/kedro/issues/2156

## Development notes

- Removed `mypy` from `pre-commit` setup and run it as part of `make lint` in a separate step instead
- Fixed or ignored all issues flagged when running `mypy kedro`
- Quite a lot of issues came from re-assigning the same variable name making the type incorrect.
- Next step is to run `mypy --strict`, see follow up PRs: 
   - #3490 
   - #3497 
   - #3498 

## Checklist


docs/source/hooks/examples.md
```diff
@@ -368,7 +368,7 @@ class NodeInputReplacementHook:
     @hook_impl
     def before_node_run(
         self, node: Node, catalog: DataCatalog
-    ) -> Optional[Dict[str, Any]]:
+    ) -> Optional | Dict[str, Any]:
```
sbrugman, line 371: `None | Dict[str, Any]`?

Makefile
```diff
@@ -9,6 +9,7 @@ clean:
 
 lint:
 	pre-commit run -a --hook-stage manual $(hook)
+	mypy kedro
```
noklam, line 12: Is there any reason that we cannot keep this in pre-commit? For `ruff` we don't include this in `[test]` because pre-commit has its own virtual environment, could be useful to avoid dependencies problem.

kedro/config/omegaconf_config.py
```diff
@@ -431,7 +433,7 @@ def _check_duplicates(seen_files_to_keys: dict[Path, set[Any]]):
             raise ValueError(f"{dup_str}")
 
     @staticmethod
-    def _resolve_environment_variables(config: dict[str, Any]) -> None:
+    def _resolve_environment_variables(config: DictConfig | ListConfig) -> None:
```
noklam, line 436: This may be a bit controversial, technically OmegaConf support both but Kedro only expect the former one.

noklam (approved):
Nice work! Approved with one non-blocking comment about dependencies.

Makefile
merelcht: I did it because of this comment by @astrojuanlu https://github.com/kedro-org/kedro/issues/2156#issuecomment-1623648689

astrojuanlu: My comment came from https://jaredkhan.com/blog/mypy-pre-commit

> We have seen many potential issues of running Mypy in Pre-commit:
> - Changing one file may cause a type error in another file, so we need to run Mypy on all files, not just those that have changed
> - We need to give Mypy access to the installed dependencies of the code it is type-checking, otherwise it can’t check the usages of those dependencies
> - Flags that reduce the strictness of Mypy like --ignore-missing-imports can give us false confidence

kedro/config/omegaconf_config.py
merelcht: No that's a good point. I've changed it.

Makefile
astrojuanlu: In other words, pre-commit seems best suited to run linters that are stateless, have little configuration, and don't need the code to be installed. Neither of these applies to MyPy.

AhdraMeraliQB (approved):
Looks great! Love that we're not overwriting variable names anymore, it's a lot easier to read. Just a (non-blocking) QQ - sometimes we have `var: Type | None` and other times `var: None | Type` - is there a rule or convention we should apply in terms of what should go first? Or alternatively, is consistency not a priority for typing?

DimedS (approved):
Thank you, @merelcht !

astrojuanlu:
+1 for `var: Type | None` rather than `None | Type`!

merelcht:
> Looks great! Love that we're not overwriting variable names anymore, it's a lot easier to read. Just a (non-blocking) QQ - sometimes we have `var: Type | None` and other times `var: None | Type` - is there a rule or convention we should apply in terms of what should go first? Or alternatively, is consistency not a priority for typing?

Ha yeah good spot. I started on this with `None | Type` and later decided I preferred the other way around, but forgot to change that everywhere. Will do that before merging!

PR #3490, opened 2024-01-08
Fix issues flagged by `mypy` strict 

merelcht:
## Description
Part of https://github.com/kedro-org/kedro/issues/2156 and continuation of #3485 

I'm running `mypy kedro --strict --allow-any-generics` to find the issues and fix them. There's a lot of them, so I'll split the work up in different PRs. This being the first.

## Development notes

- Running with `--allow-any-generics`, because disallowing it made mypy flag a lot of errors that weren't possible to fix https://mypy.readthedocs.io/en/stable/error_code_list2.html#check-that-type-arguments-exist-type-arg
- Made all `# type: ignore` specific to a error code
- Disabled error code `['misc']` because it was flagging stuff that couldn't be fixed like:
```
kedro/framework/cli/micropkg.py:307: error: Untyped decorator makes function "package_micropkg" untyped  [misc]
```

## Checklist


astrojuanlu (commented):
Upon quick inspection I didn't find anything problematic but I'll defer to others to do a proper review

DimedS (approved):
Thank you, @merelcht ! A lot of work has been done, LGTM.

kedro/ipython/__init__.py
```diff
@@ -6,10 +6,11 @@
 
 import logging
 import sys
+import typing
 from pathlib import Path
 from typing import Any
 
-from IPython import get_ipython
+import IPython
```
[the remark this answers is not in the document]
merelcht: The former import results in:
```
kedro/ipython/__init__.py:13: error: Module "IPython" does not explicitly export attribute "get_ipython"  [attr-defined]
```

So instead I've updated line 110 to be `IPython.get_ipython()`.

kedro/framework/cli/starters.py
```diff
@@ -592,11 +586,12 @@ def _fetch_config_from_file(config_path: str) -> dict[str, str]:
...
         ) from exc
 
-    return config
+    # The return type defined is more specific than the "Any" type config return from yaml.safe_load
+    return config  # type: ignore[no-any-return]
 
 
 def _fetch_config_from_user_prompts(
-    prompts: dict[str, Any], cookiecutter_context: OrderedDict
+    prompts: dict[str, Any], cookiecutter_context: OrderedDict | None
```
noklam, line 594: I found this a bit odd and run the same `mypy` check locally, I didn't find mypy complains the original signature.

Here we don't really expect `None` to be passed, I see that you have to add another guard clause to check `if not cookiedcutter_context`. If we apply this logic everywhere, we will almost always check for every single argument for `None`, thus I am curious what make this function a special case.

> kedro/framework/cli/starters.py:393: error: Call to untyped function "_validate_selected_tools" in typed context  [no-untyped-call]
> kedro/framework/cli/starters.py:397: error: Call to untyped function "_validate_regex" in typed context  [no-untyped-call]
> kedro/framework/cli/starters.py:401: error: Call to untyped function "_validate_regex" in typed context  [no-untyped-call]
> kedro/framework/cli/starters.py:404: error: Returning Any from function declared to return "dict[str, Any]"  [no-any-return]
> kedro/framework/cli/starters.py:445: error: Call to untyped function "_safe_load_entry_point" in typed context  [no-untyped-call]
> kedro/framework/cli/starters.py:523: error: Unused "type: ignore" comment  [unused-ignore]
> kedro/framework/cli/starters.py:595: error: Returning Any from function declared to return "dict[str, str]"  [no-any-return]
> kedro/framework/cli/starters.py:635: error: Function is missing a return type annotation  [no-untyped-def]
> kedro/framework/cli/starters.py:643: error: Function is missing a return type annotation  [no-untyped-def]
> kedro/framework/cli/starters.py:705: error: Function is missing a return type annotation  [no-untyped-def]
> kedro/framework/cli/starters.py:742: error: Function is missing a return type annotation  [no-untyped-def]

merelcht:
Marking this as draft again while investigating the test errors.

kedro/framework/cli/starters.py
merelcht: Yeah this one took me a while to fix. The original error is:
```
kedro/framework/cli/starters.py:287: error: Argument "cookiecutter_context" to "_get_extra_context" has incompatible type "OrderedDict[Any, Any] | None"; expected "OrderedDict[Any, Any]"  [arg-type]
```

The problem is that `mypy` can't figure out that `_fetch_config_from_user_prompts` will never be called when `cookiecutter_context` is `None`. I don't know if there's a more elegant solution to fix this.

AhdraMeraliQB: The only fix I could see is changing the condition that calls _fetch_config_from_user_prompts to be `if cookiecutter_context`, which would work given we have a cookiecutter context if and only if we don't have a config path. However, this isn't representative of the actual logic and just exploits the iff condition - I don't think satisfying mypy is worth it. I'm happy to keep `Ordered Dict | None` in the type hint and add a line to the docstring below explaining that the cookiecutter context cannot be None (and hence a guardrail condition has been omitted).

Alternatively we could add a guardrail condition for consistency.

AhdraMeraliQB (approved):
These are a lot of changes, thank you @merelcht, can't wait to see this go in!

Left some comments and questions, but nothing blocking! 🌟

kedro/framework/cli/jupyter.py
```diff
@@ -30,27 +31,27 @@
...
+def jupyter() -> None:
     """Open Jupyter Notebook / Lab with project specific variables loaded, or
     convert notebooks into Kedro code.
     """
 
 
 @forward_command(jupyter, "setup", forward_help=True)
 @click.pass_obj  # this will pass the metadata as first argument
-def setup(metadata: ProjectMetadata, args, **kwargs):  # noqa: unused-argument
+def setup(metadata: ProjectMetadata, /, args: Any, **kwargs: Any) -> None:  # noqa: unused-argument
```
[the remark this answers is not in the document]
merelcht: See https://martinxpn.medium.com/positional-only-and-keyword-only-arguments-in-python-37-100-days-of-python-310c311657b0 and specifically the section "Positional-only Function Arguments in Python
"

kedro/framework/cli/starters.py
merelcht: > I don't think satisfying mypy is worth it

So would you prefer to add an `type: ignore` here over the current solution?

tests/framework/cli/test_starters.py
```diff
@@ -2,6 +2,7 @@
 """
 from __future__ import annotations
 
+import collections
```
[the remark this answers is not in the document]
merelcht: Changed!

kedro/framework/cli/starters.py
AhdraMeraliQB: I should clarify, I don't think satisfying `mypy` is worth restructuring how we perform our condition checks in `_get_extra_context `. I'd prefer to leave the type hint as is with a note in the docstring that `cookiecutter_context` is never expected to be None / `_fetch_config_from_user_prompts` will only be called with a `cookiecutter_context`

PR #3498, opened 2024-01-09
Fix `mypy` error postitional only

merelcht:
## Description
Part of #2156 and follow up on #3497 and https://github.com/kedro-org/kedro/pull/3490

## Development notes
This PR specifically addresses the following error which was flagged for all CLI commands where we call `@click.pass_obj`
```
kedro/framework/cli/pipeline.py:139: error: Argument 1 to "pass_obj" has incompatible type "Callable[[ProjectMetadata, str, str, bool, KwArg(Any)], None]"; expected "Callable[[Any, str, str, bool, KwArg(Any)], None]"  [arg-type]
kedro/framework/cli/pipeline.py:139: note: This is likely because "delete_pipeline" has named arguments: "metadata". Consider marking them positional-only
```

## Checklist


astrojuanlu:
```
Argument 1 to "pass_obj" has incompatible type "Callable[[ProjectMetadata, str, str, bool, KwArg(Any)], None]"; expected "Callable[[Any, str, str, bool, KwArg(Any)], None]"  [arg-type]
```

MyPy error messages always so helpful... isn't `ProjectMetadata` compatible with `Any`?

---

```
This is likely because "delete_pipeline" has named arguments: "metadata". Consider marking them positional-only
```

does this solution work to avoid adding the `# type: ignore` comments?

merelcht:
> ```
> This is likely because "delete_pipeline" has named arguments: "metadata". Consider marking them positional-only
> ```
> 
> does this solution work to avoid adding the `# type: ignore` comments?

Ah it does! I couldn't figure out at first how to do this. The only thing I'm not 100% sure about is whether the argument `args: Any` in some of the CLI commands should be marked as positional-only as well?

astrojuanlu:
> The only thing I'm not 100% sure about is whether the argument args: Any in some of the CLI commands should be marked as positional-only as well?

Good question 🤔 Not sure how the current `args` are being used, I'm used to see `*args` instead but not sure it makes sense in this case (and `*args` are positional-only by definition).

SajidAlamQB (approved):

ankatiyar (approved):

DimedS (approved):
Thanks, @merelcht , LGTM!

Issue #3335, opened 2023-11-23
Inconsistency in `kedro new` with `--config`


ankatiyar:
## Description
Creating a new project with the `--config` flag doesn't work properly. There's multiple inconsistencies in the flow. 

## Problem 1:
#3342 

## Problem 2:
#3343 

## Problem 3:
* There's inconsistency between the flag names `--example` and the fields you have to use in `config.yml` and prompts instead, i.e. `example_pipeline`.

According to the discussion below, we decided to leave it as is

## Problem 4:
* `tools` in `config.yml` and prompts takes numeric values `1,2,3` but `--tools` takes `lint,test,docs` 

According to the discussion below, we decided to use short text in `config.yml` instead of numbers to be consistent with `--tools`

DimedS:
Problems 3 and 4 are still actual, but I'm uncertain about solution.

- For problem 3 I think it will be better if we will unify user interface and always use "example" - for prompts, CLI and config. But if I understood correctly it will be a breaking change and I don't sure is it possible now.

- For problem 4, looks like we decided to use numbers in prompts because user can see explanation of that numbers just before writing them in prompts and numbers allows to write faster. And for CLI it's more convenient to use short words. I think it's also more convenient to use short words in config for the same reason, but again it looks like a breaking change.

I think it should be good to discuss solution first, do you have any thoughts on that: @ankatiyar , @SajidAlamQB , @lrcouto , @AhdraMeraliQB , @noklam , @merelcht

AhdraMeraliQB:
I agree, that these are both breaking, I think the idea was these were nice-to-haves to be done before 0.19.0 that didn't make the cut.

For 3, I don't see much of an issue here, I assume `--example` is short for `example_pipeline`, not unsimilar from how `--nodes` is used in the cli command, but it's `node_names` in config: see [`kedro run` config docs](https://docs.kedro.org/en/latest/nodes_and_pipelines/run_a_pipeline.html#configure-kedro-run-arguments)

Regarding 4, I would agree that it would be nice to have a single standard way of providing this information, but I do think this would require input from design (and given it's a breaking change, we might get user feedback on what they'd like, or if they mind, before the time for implementing this comes around)

merelcht:
[...]

Issue #3333, opened 2023-11-23
Add-ons: Minor issues/improvements


ankatiyar:
## Description
Collecting stuff I found during bug bash related to the add-ons/tools workflow here -

### Add ons info in `pyproject.toml`
This currently gets added to the project's `pyproject.toml` as a string -
```toml
add_ons = "['Linting', 'Custom Logging', 'Documentation']"
```
Could this be a list for easier analytics on Heap side -
```toml
add_ons = ['Linting', 'Custom Logging', 'Documentation']
```

Additionally, could the selection of "example pipeline" be included in this?

merelcht:
Quickly discussed this and it would be nicer to have "example" as a separate entry.

ankatiyar:
Also found this while testing -
`kedro new --addons=lint,test,lint,test`
or entering repeat numbers in the prompt does not fail but ends up adding this to the `pyproject.toml`:
```toml
add_ons = "['Linting', 'Testing', 'Linting', 'Testing']"
```

Issue #3264, opened 2023-11-03
Minor tech debts in the project creation flow


noklam:
## Description

Clean out some tech debts in the new project creation flow. This ticket documents the issues I observed, it can be tackled in one PR or separately depending how many issues we found at the end.

## Issue 1 ->  ❌  won't be done: see discussion below
- [x] 1. the type passing should be a list, it's str now but everything is functional because we use `in` everywhere so either `list` or `str` works. It makes more sense to keep it as a `list[str]` nonetheless.

## Issue 2 -> fixed ✅ in #3429 
- [x] 2. Inconsistent logic updating cookiecutter arguments

<details>

In the `kedro new` interactive flow, the cookiecutter argument is set with this logic.
```python
    # Map the selected add on lists to readable name
    add_ons = config.get("add_ons")
    if add_ons:
        config["add_ons"] = [
            ADD_ONS_DICT[add_on] for add_on in _parse_add_ons_input(add_ons)  # type: ignore
        ]
        config["add_ons"] = str(config["add_ons"])
```
While in `kedro new --addon`, we use another logic.

```python
    if selected_addons is not None:
        addons = selected_addons.split(",")
        for i in range(len(addons)):
            addon = addons[i].strip()
            if addon in string_to_number:
                addons[i] = string_to_number[addon]
        return ",".join(addons)
```

</details>

## Issue 3 -> `addons` will be renamed to `tools`, so this isn't an issue anymore. -> fixed ✅
- [x] 3. Use of `addon`, `add_on`, `add_ons`. Particular, in the cli we use `kedro new --addons`, but pyproject.toml stores `add_ons`, we should keep this consistent.

## Issue 4. -> fixed ✅
- [x] The code isn't using best practices when it comes to keeping functions pure and not mutating input/output. Let's try and clean this up where possible. 

## Issue 5 - Not relevant ❌ , [see this discussion](https://github.com/kedro-org/kedro/pull/3533#issuecomment-1903852852)
- [x] Add exception handling for file and directory operations in case of issues like permission errors or files being used by another process. 

## Issue 6. -> fixed ✅ in #3429 
- [x] Looks like _validate_config_file_inputs function duplicates the prompt validation process described in Prompt.validate()  function, but with additional hardcoded duplicated values from prompts.yml. The same could be also for _select_prompts_to_display() func. Or if we think that we need to check some 'kedro new' options syntax in any case without depending what written in prompts.yml I think we need to clarify that and hardcode only once in starters.py.

[...]

PR #3188, opened 2023-10-17
Add `add_ons` metadata to `pyproject.toml` for project creation

noklam:
> **NOTE:** Kedro datasets are moving from `kedro.extras.datasets` to a separate `kedro-datasets` package in
> [`kedro-plugins` repository](https://github.com/kedro-org/kedro-plugins). Any changes to the dataset implementations
> should be done by opening a pull request in that repository.
## Development notes

I added more comments in the changed files. It's a bit tricky to get everything works

Overall there are few changes and refactoring. There are few benefit of using names over number:
- Easy to read from telemetry
- Flexible to re-order/change the number. This can be seen from the PR as I remove more code than writing new code.
1. Move the parsing logic back to `starter`, cookiecutter will store the parsed list i.e ["Linting","Documentation"]

## Developer Certificate of Origin
We need all contributions to comply with the [Developer Certificate of Origin (DCO)](https://developercertificate.org/). All commits must be signed off by including a `Signed-off-by` line in the commit message. [See our wiki for guidance](https://github.com/kedro-org/kedro/wiki/Guidelines-for-contributing-developers/).

If your PR is blocked due to unsigned commits, then you must follow the instructions under "Rebase the branch" on the GitHub Checks page for your PR. This will retroactively add the sign-off to all unsigned commits and allow the DCO check to pass.

## Checklist

- [ ] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [ ] Signed off each commit with a [Developer Certificate of Origin (DCO)](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository)
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [ ] Added tests to cover my changes
- [ ] Checked if this change will affect Kedro-Viz, and if so, communicated that with the Viz team

PR #3499, opened 2024-01-10
Change to short names in config

DimedS:
## Description
This PR addresses Issue 4 from #3335, implementing the following changes:

1. Altered the tools format derived from the config file to use short names instead of numbers.
2. Updated both unit and E2E tests to reflect these changes. Combined `test_tools_flag_none() `with `test_valid_tools_flag() `as the only difference between them is a single line.
3. Updated or added missing comments to ensure they are current.
4. Consolidated the functions for fetching, validating, and parsing the configuration from a file into a single function.
5. Modified the `_convert_tool_names_to_numbers()` function to return a list directly instead of string, eliminating the need for additional processing. The `_parse_tools_input()` function was utilized for this purpose.
6. Restructured the `_get_extra_context()` function for enhanced clarity. It's divided into distinct sections, with a specific section for default values.
7. Improved the `_validate_range(start, end)` function, which checks the range interval of tools. Previously, entering an excessively large end value would cause the prompt to freeze. An additional check for the upper limit of the range has been introduced to prevent this issue.

## Checklist


ankatiyar (approved):
Manually tested, works well! Thanks @DimedS!

AhdraMeraliQB (commented):
Could you update the tools docs pages too within this PR?
- https://docs.kedro.org/en/stable/starters/new_project_tools.html#specify-tools-configuration-using-kedro-new-config
- https://docs.kedro.org/en/stable/starters/index.html#:~:text=from%20comma%2Dseparated-,values%2C%20ranges%2C,-or%20%E2%80%98all/none


Additionally don't forget to put these changes in `RELEASE.md` with a note about them breaking tools config

features/steps/cli_steps.py
```diff
@@ -189,7 +189,7 @@ def create_config_file_with_tools(context, tools):
     It takes a custom tools list and sets example prompt to `y`.
     """
 
-    tools_str = tools if tools != "none" else ""
+    tools_str = tools
```
merelcht, line 192: Is this still necessary now we don't check for "none" ?

kedro/framework/cli/starters.py
```diff
@@ -487,71 +487,66 @@ def _get_extra_context(  # noqa: PLR0913
...
-    extra_context.setdefault("kedro_version", version)
-
-    tools = _convert_tool_names_to_numbers(selected_tools)
-
-    if tools is not None:
-        extra_context["tools"] = tools
-
+    # Update extra_context, if CLI inputs are available
+    if selected_tools is not None:
+        extra_context["tools"] = _convert_tool_names_to_numbers(selected_tools)  # type: ignore
```
merelcht, line 504: I've put a lot of effort into making these `type: ignore` statements specific (see https://github.com/kedro-org/kedro/pull/3490) so please update that here as well. I'm seeing quite a lot though, so I'm wondering if there's anything that can be done to not just ignore but actually fix the issues?

merelcht (commented):
The code is starting to look a lot better with these refactorings, but I think there's still a bit more we can do. I've left comments with some suggestions.

kedro/framework/cli/starters.py
DimedS: Sounds good, I'll hold off until you merge #3490. After that, I'll update my code to align with our new --strict standards.

DimedS: fixed all type: ignore

kedro/framework/cli/starters.py
```diff
@@ -487,71 +487,66 @@ def _get_extra_context(  # noqa: PLR0913
...
+
+    # Convert tools numbers to list of readable names
+    extra_context["tools"] = [
+        NUMBER_TO_TOOLS_NAME[tool]
+        for tool in extra_context["tools"]  # type: ignore
+    ]
+    # use "None" in case of no tools selected
+    if extra_context["tools"] == []:
+        extra_context["tools"] = ["None"]  # type: ignore
+    # convert to string for PUT THE REASON HERE
```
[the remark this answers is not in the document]
DimedS: fixed, I've moved the explanation that was previously a comment at the top of the function into the DocString. Thanks to @AhdraMeraliQB for providing me with this information.

kedro/framework/cli/starters.py
```diff
@@ -592,10 +594,24 @@ def _fetch_config_from_file(config_path: str) -> dict[str, str]:
...
         ) from exc
 
+    _validate_config_file_against_prompts(config, prompts_required)
+
+    if starter_alias and ("tools" in config or "example_pipeline" in config):
+        raise KedroCliError(
+            "The --starter flag can not be used with `example_pipeline` and/or `tools` keys in the config file."
+        )
+
+    _validate_regex("project_name", config.get("project_name", "New Kedro Project"))
```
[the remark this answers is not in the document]
DimedS: fixed, changed to validate_input_with_regex_pattern()

kedro/framework/cli/starters.py
```diff
@@ -592,10 +594,24 @@ def _fetch_config_from_file(config_path: str) -> dict[str, str]:
             f"Failed to generate project: could not load config at {config_path}."
         ) from exc
 
+    _validate_config_file_against_prompts(config, prompts_required)
+
+    if starter_alias and ("tools" in config or "example_pipeline" in config):
+        raise KedroCliError(
+            "The --starter flag can not be used with `example_pipeline` and/or `tools` keys in the config file."
+        )
```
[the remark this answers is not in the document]
DimedS: good point, fixed

kedro/framework/cli/starters.py
```diff
@@ -628,7 +644,12 @@ def _fetch_config_from_user_prompts(
         user_input = read_user_variable(str(prompt), cookiecutter_variable)
         if user_input:
             prompt.validate(user_input)
+            if variable_name == "tools":
+                # convert tools input to list of numbers and validate
+                user_input = _parse_tools_input(user_input)
+                _validate_selection(user_input)
```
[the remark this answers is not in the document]
DimedS: I agree, fixed

DimedS:
> Could you update the tools docs pages too within this PR?
> 
> * https://docs.kedro.org/en/stable/starters/new_project_tools.html#specify-tools-configuration-using-kedro-new-config
> * https://docs.kedro.org/en/stable/starters/index.html#:~:text=from%20comma%2Dseparated-,values%2C%20ranges%2C,-or%20%E2%80%98all/none
> 
> Additionally don't forget to put these changes in `RELEASE.md` with a note about them breaking tools config

thank you for the links @AhdraMeraliQB , fixed

DimedS:
> The code is starting to look a lot better with these refactorings, but I think there's still a bit more we can do. I've left comments with some suggestions.

Thank you, @merelcht  , for your valuable suggestions. I've implemented them and they indeed make a lot of sense.

Could you help me understand the purpose of the test_fetch_config_from_user_prompts_with_context() function? I've temporarily removed it as I'm unsure how to modify it after updating the fetch_config_from_user_prompts() function. Your guidance on this would be greatly appreciated.

kedro/framework/cli/starters.py
```diff
@@ -353,8 +351,8 @@ def new(  # noqa: PLR0913
         # If interactive flow used, print hint
         interactive_flow = prompts_required and not config_path
         _print_selection_and_prompt_info(
-            extra_context.get("tools"),
-            extra_context.get("example_pipeline"),
+            extra_context["tools"],
+            extra_context["example_pipeline"],
```
merelcht, line 355: Why did you change it from `.get()` to this direct access? I always feel like `.get()` is the safer way in case the key doesn't exist in the dictionary.

kedro/framework/cli/starters.py
```diff
@@ -543,64 +542,51 @@ def _get_extra_context(  # noqa: PLR0913
...
-        extra_context["tools"] = str([NUMBER_TO_TOOLS_NAME[tool] for tool in tools])
-
-    extra_context["example_pipeline"] = _parse_yes_no_to_bool(
-        example_pipeline
-        if example_pipeline is not None
-        else extra_context.get("example_pipeline", "no")
-    )
+    # set defaults for required fields, will be used mostly for starters
+    extra_context.setdefault("kedro_version", version)
+    extra_context.setdefault("tools", str(["None"]))
```
merelcht, line 567: `str(["None"])` still seems like a bit of an odd value to me.. We don't have to change that in this PR, but I think it would be good to discuss.

DimedS: Yes, using .get() is indeed a safer method for retrieving values from a dictionary. However, I noticed we were overusing it in every step of our processing. This made the code less precise and harder to comprehend, particularly in discerning whether and where default values were being provided. 
Upon analyzing the code, I realized the default values were actually set in the _get_extra_context() function. During the refactoring of this function, I made the inclusion of default values more explicit by introducing a 'default' section and provided comments about it in DocString. Subsequently, I replaced all instances of variable access using get() with direct access. This change aims to enhance the clarity and strictness of our code. 
Now, I hope it's clear that values must be specified, and to understand which values are used, one should review the previous processing steps instead of worrying about default values at each stage.

features/steps/cli_steps.py
DimedS: tools == "" is not acceptable from config file, only "none/all" or short names, so I changed that line

kedro/framework/cli/starters.py
DimedS: Yes, using .get() is indeed a safer method for retrieving values from a dictionary. However, I noticed we were overusing it in every step of our processing. This made the code less precise and harder to comprehend, particularly in discerning whether and where default values were being provided.
Upon analyzing the code, I realized the default values were actually set in the _get_extra_context() function. During the refactoring of this function, I made the inclusion of default values more explicit by introducing a 'default' section and provided comments about it in DocString. Subsequently, I replaced all instances of variable access using get() with direct access. This change aims to enhance the clarity and strictness of our code.
Now, I hope it's clear that values must be specified, and to understand which values are used, one should review the previous processing steps instead of worrying about default values at each stage.

kedro/framework/cli/starters.py
```diff
@@ -631,21 +635,46 @@ def _fetch_config_from_file(config_path: str) -> dict[str, str]:
...
-    return config  # type: ignore[no-any-return]
+    config = (
+        {str(key): str(val) for key, val in config_input.items()}
+        if config_input
+        else {}
+    )
+
+    if starter_alias and ("tools" in config or "example_pipeline" in config):
+        raise KedroCliError(
+            "The --starter flag can not be used with `example_pipeline` and/or `tools` keys in the config file."
```
AhdraMeraliQB, line 656: Is this check not redundant as `_validate_flag_inputs` already checks for this?

kedro/framework/cli/starters.py
```diff
@@ -631,21 +635,46 @@ def _fetch_config_from_file(config_path: str) -> dict[str, str]:
...
-            click.echo(yaml.dump(config, default_flow_style=False))
+            click.echo(yaml.dump(config_input, default_flow_style=False))
     except Exception as exc:
         raise KedroCliError(
             f"Failed to generate project: could not load config at {config_path}."
         ) from exc
 
-    # The return type defined is more specific than the "Any" type config return from yaml.safe_load
-    return config  # type: ignore[no-any-return]
+    config = (
```
AhdraMeraliQB, line 648: Why do we introduce this dictionary comprehension here?

kedro/framework/cli/starters.py
```diff
@@ -543,64 +542,51 @@ def _get_extra_context(  # noqa: PLR0913
...
+    extra_context.setdefault("kedro_version", version)
+    extra_context.setdefault("tools", str(["None"]))
+    extra_context.setdefault("example_pipeline", "False")
 
     return extra_context
 
 
-def _convert_tool_names_to_numbers(selected_tools: str | None) -> str | None:
-    """Prepares tools selection from the CLI input to the correct format
+def _convert_tool_short_names_to_numbers(selected_tools: str | None) -> list:
```
AhdraMeraliQB, line 573: With the current changes, this function should never be invoked with selected_tools as `None`. We can remove `None` from the argument type hint

kedro/framework/cli/starters.py
```diff
@@ -847,6 +853,10 @@ def _validate_range(start: Any, end: Any) -> None:
             message = f"'{start}-{end}' is an invalid range for project tools.\nPlease ensure range values go from smaller to larger."
             click.secho(message, fg="red", err=True)
             sys.exit(1)
+        if int(end) > max([int(num) for num in NUMBER_TO_TOOLS_NAME]):
```
AhdraMeraliQB, line 856: Isn't this check redundant? The purpose of _validate_range is to ensure that the range provided can be parsed into a list of numbered options. The validity of these individual options will be validated by `_validate_tools_selection`, which currently always follows `_parse_tools_input`. It feels redundant to have this check, and incomplete that we would only check for the larger number in the range, but we don't check that the smaller number is > 0. I'd opt for removing this additional condition.

AhdraMeraliQB (commented):
Leaving my comments on the implementation here, will add another review if necessary as I go through the test file changes

kedro/framework/cli/starters.py
DimedS: > Is this check not redundant as `_validate_flag_inputs` already checks for this?

Looks like _fetch_validate_parse_config_from_file() checks input from config file, and _validate_flag_inputs() validates inout from CLI?

kedro/framework/cli/starters.py
DimedS: > Why do we introduce this dictionary comprehension here?

I made this to be consistent with mypy --strict rules: 
yaml.safe_load(config_file) returns object with Any type. From our function we want to return dict[str, str] object. I think a good way to do so - made an explicit conversion, after that we can remove "type ignore"

kedro/framework/cli/starters.py
DimedS: Looks like in the main code - yes, but it used in tests also:
`result = _convert_tool_short_names_to_numbers(None)`
we can rewrite tests also but but maybe it's not so critical

docs/source/starters/index.md
```diff
@@ -32,7 +32,7 @@ The configuration file must contain:
 
 Additionally, the configuration file may contain:
 
-* `tools`: The tools to customise your project setup with. Select from comma-separated values, ranges, or 'all/none'. Omitting this from your configuration file will result in the default selection of `none`.
+* `tools`: The tools to customise your project setup with. Select from comma-separated values `lint, test, log, docs, data, pyspark, viz` or `all/none`. Omitting this from your configuration file will result in the default selection of `none`.
 * `example`: Indicate `yes` or `no` to select whether you would like your project to be populated with example code. Omitting this from your configuration file will result in the default selection of `no`.
```
ankatiyar, line 36: This is `example_pipeline` still, isn't it?

kedro/framework/cli/starters.py
AhdraMeraliQB: I'd rather rewrite the test, as it's not indicative of the actual execution and was probably introduced for test coverage instead of representing a user story

kedro/framework/cli/starters.py
AhdraMeraliQB: I see, could you add a comment above the condition noting that this is a safeguarding clause and not strictly part of the validation? 

As a side note, you could skip the list comprehension by using `len(NUMBER_TO_TOOLS_NAME )`

kedro/framework/cli/starters.py
AhdraMeraliQB: Instead of this, could we add typing `config_input: dict[str,str] = yaml.safe_load(config_file)` on line 638 instead?

kedro/framework/cli/starters.py
AhdraMeraliQB: Absolutely correct! I got my wires crossed

docs/source/starters/index.md
DimedS: Thank you, @ankatiyar , fixed!

kedro/framework/cli/starters.py
DimedS: that's right, fixed!

kedro/framework/cli/starters.py
DimedS: correct, fixed!

kedro/framework/cli/starters.py
DimedS: great, it works, fixed!

kedro/framework/cli/starters.py
AhdraMeraliQB: Idea, just noting this here but not adding to the scope of the PR - could we save it as an unwrapped list `["None"]` instead?

Context:

The reason we pass through a list wrapped in a string is because `cookiecutter` interprets a list as a provision of options, from which it selects one to save. To preserve all the chosen tools, we wrap the list in a string. Though the option for no tools is processed separately, `pyproject.toml` has no knowledge of what is passed through and so we cannot pass through just the string "None" (which will be unwrapped).

However, none tools cannot be selected with any other tools, and so passing through just a list shouldn't have any adverse effects; `cookiecutter` would pick the only option from the list - "None" - and pass that through to `pyproject.toml`. 

Would we run into the same problem with unwrapping strings? If so we could always pass through `[["None"]]` but it doesn't seem much better than `str(["None"])` 🤔 

Additionally, for telemetry, should we be consistent with what we store (a list for several tools, but this would save a string if no tools are selected)?

Alternatively, I believe using dictionaries to pass a collection of objects through `cookiecutter` is possible, but it would require significant changes to pyproject.toml - I'm unsure we'd be able to preserve the list type as we'd have to somehow process these values within the `cookiecutter` execution.

Any opinions on this? CC: @DimedS @merelcht

DimedS: @AhdraMeraliQB , and why we need a string of a list? Why just a string is not enough, like:
tools = "Linting, Testing, ..." or tools = "None"

tests/framework/cli/test_starters.py
```diff
@@ -1395,51 +1332,27 @@ class TestToolsAndExampleFromCLI:
             "ALL",
             "TEST, LOG, DOCS",
             "test, DATA, liNt",
```
AhdraMeraliQB, line 1334: This is the other test where we could have test cases for both accepted tool names (`test` and `tests` e.t.c.)

AhdraMeraliQB (commented):
This looks amazing!! Thank you for all the changes - only one significant question (re: the error messages) from me 😄

kedro/framework/cli/starters.py
AhdraMeraliQB: @DimedS 
If I remember correctly, the reason is tied to telemetry - it's easier (and more intuitive) to query tools data when stored as a list than as a string. We did originally pass this through as a string ([xref](https://github.com/kedro-org/kedro/pull/3188/files#:~:text=cookiecutter.kedro_version%20%7D%7D%22-,add_ons,-%3D%20%7B%7B%20cookiecutter.add_ons)), but was suggested to be changed as part of #3264 and in #3333

The fix in #3426 introduced using [str["None"]](https://github.com/kedro-org/kedro/pull/3426/files#:~:text=else%3A-,extra_context,-%5B%22tools%22) to workaround `cookiecutter` limitations ([see Nok's comment](https://github.com/cookiecutter/cookiecutter/issues/1975#issuecomment-1805044493)) and the rest is history 😅

tests/framework/cli/test_starters.py
```diff
@@ -321,14 +320,25 @@ def test_parse_tools_invalid_range(self, input, capsys):
...
         assert message in capsys.readouterr().err
 
+    @pytest.mark.parametrize(
+        "input,right_border",
+        [("3-9", "9"), ("3-10000", "10000")],
+    )
+    def test_parse_tools_range_too_high(self, input, right_border, capsys):
+        with pytest.raises(SystemExit):
+            selected = _parse_tools_input(input)
+            _validate_tool_selection(selected)
```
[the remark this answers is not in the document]
DimedS: I think it will protect us from "0-5"

tests/framework/cli/test_starters.py
```diff
@@ -1164,24 +1136,28 @@ class TestToolsAndExampleFromConfigFile:
...
-            "2,3,4",
-            "3-5",
-            "1,2,4-6",
-            "1,2,4-6,7",
-            "4-6,7",
-            "1, 2 ,4 - 6, 7",
-            "1-3,5-7",
+            "lint",
+            "test",
+            "tests",
```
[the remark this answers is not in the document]
DimedS: maybe I misunderstood, but looks like that test cases are already there?

kedro/framework/cli/starters.py
```diff
@@ -852,6 +851,11 @@ def _validate_range(start: Any, end: Any) -> None:
             message = f"'{start}-{end}' is an invalid range for project tools.\nPlease ensure range values go from smaller to larger."
             click.secho(message, fg="red", err=True)
             sys.exit(1)
+        # safeguard to prevent passing of excessively large intervals that could cause freezing:
+        if int(end) > len(NUMBER_TO_TOOLS_NAME):
+            message = f"'{start}-{end}' is an invalid range for project tools.\n{end} is too large."
```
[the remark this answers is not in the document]
DimedS: I agree, fixed

tests/framework/cli/test_starters.py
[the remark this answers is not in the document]
AhdraMeraliQB: Ah I see, this is only tested when testing `_convert_tool_short_names_to_numbers`
As a side note, I think the test class `TestConvertToolNamesToNumbers` could be refactored, but not needed in this PR

tests/framework/cli/test_starters.py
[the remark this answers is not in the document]
AhdraMeraliQB: I'm not sure that should be a test case for this test as we're checking that the error is thrown when parsing a range that extends beyond available tools. We'd want to ensure this error is being correctly raised by `_parse_tools_input` instead of later in `_validate_tool_selection `.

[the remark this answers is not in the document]
DimedS: yes, we can make a full validation inside of _parse_tools_input() and remove _validate_tool_selection(selected), I agree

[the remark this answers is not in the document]
DimedS: Apologies, I misunderstood and thought you were suggesting to completely remove the _validate_tool_selection(selected) function. I've addressed your comment now.

tests/framework/cli/test_starters.py
```diff
@@ -1049,60 +1029,52 @@ def test_valid_tools_and_example(self, fake_kedro_cli, tools, example_pipeline):
...
-        _clean_up_project(Path("./new-kedro-project"))
 
     @pytest.mark.parametrize(
-        "bad_input",
-        ["bad input", "-1", "3-", "1 1"],
+        "input,right_border",
+        [("3-9", "9"), ("3-10000", "10000")],
     )
-    def test_invalid_tools(self, fake_kedro_cli, bad_input):
+    def test_invalid_tools_range_too_high(self, fake_kedro_cli, input, right_border):
```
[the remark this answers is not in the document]
DimedS: agree, fixed

AhdraMeraliQB (approved):
Fantastic work @DimedS!! Thanks for making all the changes, this will make the file so much easier to read 🌟 

The PR has changed quite a bit since the first approval, I've re-requested @ankatiyar to take a new look at this before we give it the final send-off 🚀

ankatiyar (approved):
Thanks @DimedS for the fix as well as a refactoring! 👍🏾

DimedS:
Thank you for your contribution @AhdraMeraliQB , @ankatiyar , @merelcht !

Issue #1828, opened 2022-09-05
kedro run CLI incorrectly splits the names of nodes at commas


jmholzer:
## Description
Running kedro with the command `kedro run --from-nodes "two_inputs([A0,B0]) -> [C1]"` will cause an error: `Pipeline does not contain nodes named ['B0]) -> [C1]', 'two_inputs([A0'].`

Kedro has incorrectly split the name of the intended `Node` into the names of two `Node`s that do not exist.
## Context
This bug prevents the user from re-running kedro from any `Node` with a comma in its name. This is the case for any `Node` that is not explicitly given a name and has more than one input.

## Steps to Reproduce
1. Create a kedro project
2. Attempt to run the pipeline from a node with a comma in its name (`kedro run --from-nodes 'My node, a good node'`)
3. Receive the error `Pipeline does not contain nodes named ['a good node', 'My node'].`

## Expected Result
Kedro should try to run from the node with the specified name.

## Actual Result
Kedro attempts to run from multiple nodes, with names drawn from the result of splitting the intended `Node` name at commas.

noklam:
Linking this issue to #1795

AntonyMilneQB:
As discussed in backlog grooming, this is a very long-standing issue bug and it would be nice to get it fixed. I've looked through our old Jira board to get the previous discussion for context too.

Possible solutions here:

1. some syntax where you can escape a comma in a node name with a backslash so it doesn't get split by `_split_string`
2. alter node name syntax to use `;` rather than `,` - quick but hacky fix. This is a breaking change though (won't affect anyone using `--from-nodes` etc. arguments because those don't work, but would affect someone using an automatically-generated node name in `--node` though)
3. @jmholzer idea of a fine-state machine csv-parsing type solution to do the string splitting rather than just blindly splitting on `,`. This would basically mean that `,` inside `()` would basically not split the string. It wouldn't solve the case of a comma explicitly being used inside a node name, but it would solve the automatically-named node case, which is way  more important
4. change the resume scenario suggestion to use `from-inputs` instead. This should completely circumvent problems around commas (unless you choose to use a comma in your catalog entry name... but no one does that)
5. change `--from-nodes` to be a `multiple=True` argument in click, just like `--node` is. This seems pretty inconsistent currently actually: you do `--node node_1 --node node_2` but `--from-nodes node_1, node_2`. Why not `--from-nodes node_1 --from-nodes node_2`? At a glance this reads a bit weird to me though 🤔 

[...]

PR #3510, opened 2024-01-15
 create line magic to debug a node in notebook workflow

noklam:
## Description

You can manually test it by creating a new `spaceflight` project and run `%load_node split_data_node`.

there are some skipped tests that we aim to improve on separte PR. The current scope of the PR is huge already.

### To reviewer
- Try to run it and see if there are bugs, there are some skipped tests that we are awared not support yet (nested function/lambda)
- Does the format make sense to you? The `return` statement isn't handled properly yet because I am not sure how should we treat it, should we simply comment it out? It would be a naive solution because it wouldn't work nicely if a function have multiple returns.
- Play with different platforms (databricks/sagemakers etc) - I haven't played with any of this, but I will test on databricks when I have time.

IMO the code to load catalog is the most needed part, the import statement can be handled with `from xxx import *` which guarantee the code can be run (good enough for debugging). The return statement is also not too important because for a debugging use case the error should happen before that. It's also not too much work for the user to just comment it out or handle edge cases themselves. Ideally it would be nice if user can just copy back the entire block back to the codebase (or even automated)

I didn't go with sophisticated solutions such as parsing the abstract syntax tree, there are many small things that we can improve but should be prioritised properly.

I also think some refactor is needed to make tests simpler, it's troublesome to write these tests and easy to miss a newlike character somewhere. I will do this but maybe a bit later.

## Demo
debug

## Note
- Add support for `kedro ipython`?
- Support VS Notebook (?)
- Support write back to script?

## Development notes

One of the challenge of this feature is that there are no solution that works across platform. It sounds like a simple task, but there are very limited support on creating **multiple cells** in notebook. The standard `%load` method only support creating ONE cell but not more. I end up choosing `ipylab` since this is the only option that support create multiple cell which offers a nicer user experience. It works only for notebook >7.0 and JupyterLab.

If we want to support VSCode notebook, we may have to settle with a single cell solution, which is still an improvement but not ideal. For the initial MVP, I think it's fine to support Notebook / Lab. If we end up supporting more, it's likely we will need to maintain platform specific code until there are nice solutions that works well everywhere.

- Databricks Notebook
- VSCode Notebook
- Notebook / lab

Edge Case improvement:
- Handle imports that are not from top of the script, i.e. functions define in the same file as the node function which lives in `locals()`.
- Intermediate MemoryDataset 

Reference List: 
- https://stackoverflow.com/questions/65379879/define-a-ipython-magic-which-replaces-the-content-of-the-next-cell
- https://github.com/ipython/ipykernel
- IPython Docs - https://ipython.readthedocs.io/en/stable/config/custommagics.html
- `%load` source-code https://github.com/ipython/ipython/blob/b14df3808a86bb1c9565816d15be4a52bc694c2a/IPython/core/magics/code.py#L316-L326
- ipylab

## Checklist


AhdraMeraliQB:
Left to do:
- [ ] Fix naive import fetching
- [ ] Fix weird node function edge cases
- [ ] Test with parameters as node input
- [ ] Test if node names and func.__name__ always match
- [ ] Make find node recursive if node function makes call to helper function (unsure if in scope/consequences)

noklam:
Requested @AhdraMeraliQB for comments only since we worked together on this ticket

astrojuanlu:
I tried this on https://github.com/astrojuanlu/talk-kedro-huggingface and I get a `ValueError: Node clean_statuses is not found in any pipelines.` for `%load_node clean_statuses`.

This is my pipeline:

```
    return pipeline(
        [   
            node(
                func=clean_statuses,
                inputs="statuses_table",
                outputs="statuses_clean",
            )   
        ]   
    )
```

Maybe it's failing because the node has no `name=`?

kedro/ipython/__init__.py
```diff
@@ -178,3 +184,124 @@ def _find_kedro_project(current_dir: Path) -> Any:  # pragma: no cover
...
+
+    cells: list[str] = []
+    cells.append(node_inputs)
+    cells.append(imports)
+    cells.append(function_text)
+    return cells
+
+
+def _find_node(node_name: str) -> Node:
+    for pipeline in pipelines.values():
```
astrojuanlu, line 245: To make this easier to debug, is there a chance we turn `pipelines` into an actual parameter?

```suggestion
def _find_node(node_name: str, pipelines: ...) -> Node:
    for pipeline in pipelines.values():
```

the caller would have to be modified as well of course.

In general I'm not a fan of how `pipelines` comes `from kedro.framework.project`, but that's another story.

kedro/ipython/__init__.py
```diff
@@ -178,3 +184,124 @@ def _find_kedro_project(current_dir: Path) -> Any:  # pragma: no cover
...
+    default=None,
+)
+def magic_load_node(node: str) -> None:
+    """The line magic %load_node <node_name>
+    Currently it only support Jupyter Notebook (>7.0) and Jupyter Lab. This line magic
+    will generate code in multiple cells to load dataest from Data Catalog, import
+    relevant functions and modules, and the function body.
+    """
+    cells = _load_node(node)
+    from ipylab import JupyterFrontEnd
```
astrojuanlu, line 205: Trying this on IPython (after naming my node, see above) and got an error:

```
ModuleNotFoundError: No module named 'ipylab'
```

I would find it confusing if `kedro.ipython` didn't work on IPython, just because of the name. On the other hand, I don't think it makes sense for this to be in IPython at all, because the point is to bring the source code and be able to edit it on a cell - something that can't happen on a REPL.

If this is to be a Jupyter-only extension, I think we should have it somewhere else. Maybe `kedro.jupyter`? Or even a separate `kedro-jupyter` package?

astrojuanlu: Sorry, there are two different things here:

- I did not reinstall kedro so I didn't have `ipylab`, but I see it's there when the user does `pip install kedro[jupyter]` 👍🏽 
- After installing it, I did `%load_node clean_statuses` on IPython and nothing happened:

image

My node function is not in `dir()`:

```
In [4]: "clean_statuses" in dir()
Out[4]: False
```

so I stand by my previous point: since this is pointless in IPython, I'd move it somewhere else.

noklam:
@astrojuanlu Good question around the name of the node. [We had some discussion for node name quite a while ago](https://github.com/kedro-org/kedro/issues/1828). The default name may surprise you a bit, because it is used for the logs that get print when you do `kedro run`

image

kedro/ipython/__init__.py
noklam: https://github.com/kedro-org/kedro/issues/3536

There is a separate ticket for IPython support. Originally I thought the same, but it appears that some people, particular anti-notebook user would still like to have IPython support. It will work slightly different because IPython doesn't have the concept of "cells". It will still be useful if data can get pre-load (or just the code to load the corresponding data from catalog). If you have comments about the design of the feature, feel free to drop comments over there instead.

Re: `kedro.ipython` vs `kedro.jupyter` - I agree it's a bit confusing, but I think it's okay to leave it in `kedro.ipython` since ultimately I plan to add support for IPython. In additionally, we almost use IPython/Jupyter interchangeably because the feature parity is always the same, until now.

kedro/ipython/__init__.py
noklam: Good call, I will do this shortly.

kedro/ipython/__init__.py
astrojuanlu: Sounds good. I don't know how feasible this is, but if it's possible to show a warning like "IPython is not yet supported", that would be golden.

merelcht:
When trying to run the cell that contains the function body I see:
Screenshot 2024-01-24 at 16 23 39


I guess the question is if it should contain the full function definition instead of just the body?

astrojuanlu:
Tested this on VSCode. I was first prompted to download some widget files, which is fine:

image

But then, apparently the IDE wasn't able to actually download the widget files:

image

kedro/ipython/__init__.py
```diff
@@ -178,3 +185,124 @@ def _find_kedro_project(current_dir: Path) -> Any:  # pragma: no cover
...
+        list[str]: A list of string which is the generated code, each string represent a
+        notebook cell.
+    """
+    node = _find_node(node_name, pipelines)
+    node_func = node.func
+
+    node_inputs = _prepare_node_inputs(node)
+    imports = _prepare_imports(node_func)
+    function_text = _prepare_function_body(node_func)
+    function_text = "# Function Body\n" + function_text
```
merelcht, line 236: Can we distinguish between these variables instead of overwriting the same one?

pyproject.toml
```diff
@@ -26,7 +26,7 @@ dependencies = [
     "more_itertools>=8.14.0",
     "omegaconf>=2.1.1",
     "parse>=1.19.0",
-    "pluggy>=1.0",
+    "pluggy>=1.0, <1.4.0",
```
merelcht, line 29: Why does `pluggy` need to be pinned with an upper bound?

tests/ipython/test_ipython.py
```diff
@@ -297,3 +379,159 @@ class MockKedroContext:
...
+            "_get_pipelines_registry_callable",
+            return_value=my_register_pipeline,
+        )
+
+        node_to_load = "dummy_node"
+        try:
+            magic_load_node(node_to_load)
+        except Exception:
+            # Fail if an error is thrown
+            assert False
```
merelcht, line 408: What does this try/catch do?

merelcht (commented):
I left some questions, but in general the implementation looks good 👍

kedro/ipython/__init__.py
```diff
@@ -178,3 +185,124 @@ def _find_kedro_project(current_dir: Path) -> Any:  # pragma: no cover
...
+        except ValueError:
+            continue
+    # If reached the node was not found in the project
+    raise ValueError(f"Node with name='{node_name}' not found in any pipelines.")
+
+
+def _prepare_imports(node_func: Callable) -> str:
+    """Prepare the import statements for loading a node."""
+    python_file = inspect.getsourcefile(node_func)
+    logger.info(f"Loading node defintion from {python_file}")
```
ankatiyar, line 259:
```suggestion
    logger.info(f"Loading node definition from {python_file}")
```
typo

ankatiyar (commented):
This is super cool, tried it locally but will also try it on databricks and get back to you but the implementation is looking good! I think we should probably comment out the return statement.
Would it also be possible to copy the code for the helper functions used in the node function?

tests/ipython/test_ipython.py
noklam: @AhdraMeraliQB I think this can be removed? If a test run successfully `pytest` treat it as a success so `assertion` is not needed.

noklam:
@astrojuanlu Does it works for you in VSCode? It doesn't seem to find the widget that I need and it doesn't work for me with VSCode Notebook. 

There are some compatibility issues that I would like to brought up in tech design next. For now the expectation it works for Jupyter Notebook/Lab only (maybe Sagemaker would work, Databricks doesn't work for me).

astrojuanlu:
It does not: https://github.com/kedro-org/kedro/pull/3510#issuecomment-1908540889

noklam:
> When trying to run the cell that contains the function body I see:
Screenshot 2024-01-24 at 16 23 39

>I guess the question is if it should contain the full function definition instead of just the body?

Missed this comment earlier. My view on this is that it should just show the body, because we already try to match the function signature to catalog's name. We expect it it something that user can run.

To your questions, the block should be runnable if `return` is commented out. We can simply comment it out, but we will need to handle some edge case better.


For example:
This is a valid case
```python
x = 123
y = func(x)
# return y
```


```py
x = yyy
if <some_flag):
  # return x
else:
  # return None
```

noklam:
Thanks @AhdraMeraliQB for fixing the tricky unit tests!

AhdraMeraliQB:
### Tech Design outcome - Actions to get this PR completed:

- [ ] Warn users when preparing node inputs that inputs need to be explicitly declared in the catalog
- [ ] Handle return statements by commenting out and substituting display()
- [ ] Add a line when node not found about node name vs node function
- [ ] Add a platform agnostic warning that this is an experimental feature and is only supports jupyter and ipython, linking parent issue for feature discussion

### Other blocking actions:
- [ ] Handle multi-line function definitions and add test case

astrojuanlu:
I flagged an issue about manipulating `return` statements https://github.com/kedro-org/kedro/issues/3535#issuecomment-1919690287

RELEASE.md
```diff
@@ -1,6 +1,7 @@
 # Upcoming Release 0.19.3
 
 ## Major features and improvements
+* Create the debugging line magic `%load_node` for Jupyter Notebook/IPython.
```
merelcht, line 4: This works for `ipython` now as well?

kedro/ipython/__init__.py
```diff
@@ -178,3 +185,133 @@ def _find_kedro_project(current_dir: Path) -> Any:  # pragma: no cover
...
+@argument(
+    "node",
+    type=str,
+    help=("Name of the Node."),
+    nargs="?",
+    default=None,
+)
+def magic_load_node(node: str) -> None:
+    """The line magic %load_node <node_name>
+    Currently it only support Jupyter Notebook (>7.0) and Jupyter Lab. This line magic
```
merelcht, line 201:
```suggestion
    Currently it only supports Jupyter Notebook (>7.0) and Jupyter Lab. This line magic
```

kedro/ipython/__init__.py
```diff
@@ -178,3 +185,133 @@ def _find_kedro_project(current_dir: Path) -> Any:  # pragma: no cover
...
+    will generate code in multiple cells to load dataest from Data Catalog, import
```
merelcht, line 202:
```suggestion
    will generate code in multiple cells to load datasets from `DataCatalog`, import
```

merelcht (approved):
I've tried it again and it works great! I think @marrrcin 's suggestion is a real improvement 👍 I'm very excited to see this go in 😄

kedro/ipython/__init__.py
```diff
@@ -178,3 +185,133 @@ def _find_kedro_project(current_dir: Path) -> Any:  # pragma: no cover
...
+    relevant functions and modules, and the function body.
```
ankatiyar, line 203: This is now function and a function call, right?

kedro/ipython/__init__.py
```diff
@@ -178,3 +185,133 @@ def _find_kedro_project(current_dir: Path) -> Any:  # pragma: no cover
...
+        node_name (str): The name of the node.
+
+    Returns:
+        list[str]: A list of string which is the generated code, each string represent a
+        notebook cell.
+    """
+    warnings.warn(
+        """This is an experimental feature, only Jupyter Notebook (>7.0) & Jupyter Lab are supported. If you find any unexpected issue, please report
+                  in https://github.com/kedro-org/kedro/issues/3580"""
+    )
```
ankatiyar, line 233: I think this should say something like "if you encounter unexpected behaviour or would like to suggest feature enhancements, add it under this github issue <link>"

ankatiyar (approved):
Excited to see this feature released soon! 💯

marrrcin (approved):

kedro/ipython/__init__.py
noklam: You are right!

RELEASE.md
noklam: Not yet, I've revised

kedro/ipython/__init__.py
```diff
@@ -178,3 +185,134 @@ def _find_kedro_project(current_dir: Path) -> Any:  # pragma: no cover
...
+
+def _find_node(node_name: str, pipelines: _ProjectPipelines) -> Node:
+    for pipeline in pipelines.values():
+        try:
+            found_node: Node = pipeline.filter(node_names=[node_name]).nodes[0]
+            return found_node
+        except ValueError:
+            continue
+    # If reached the node was not found in the project
+    raise ValueError(
```
datajoely, line 259: I wonder if we want to support `func_name` at some point too? 🤔

noklam: This was briefly discussed in last TD. @AhdraMeraliQB originally have an implementation. 

It runs into issue if function is used twice, nodes are unique but functions do not have to. Fundamentally we want to fix this with the default node name but this was decided to be fixed later.

PR #3518, opened 2024-01-16
Make e2e tests pull `main` branch of `kedro-starters` when using `tools`

SajidAlamQB:
## Description
Our CI will use the same tag version of `kedro-starters` as `kedro` but this causes a problem when we want to do a new `kedro` release as the new release tag won't exist on starters until we do the release. So we should make our e2e test run on the `main` branch of `kedro-starters` for tests involving the `tools` cli or find an alternative solution.

#### Method 3:

After pairing with @merelcht we decided to move the tests that are causing issues to `kedro-starters` see https://github.com/kedro-org/kedro-starters/pull/209 for more info.

#### Method 2: 
I tried using environment variables in GitHub actions to change the checkout branch to main if it detects that a CI is running it. This doesn't feel like the best solution as it modifies the core code to get tests working. Update: after speaking with @merelcht we decided this method is not the best idea as it modifies production code to make tests pass.

#### Method 1: 
I've added try and catch block for `RepositoryCloneFailed` from `cookiecutter`. If it catches `RepositoryCloneFailed` it will modify it to return the `main` branch of `kedro-starters`. This will effect both `--tools` when using `--example` and `--starters` when it uses `--checkout`. In both instances users will be warned that the version they were looking for was not found and it will default to using the `main` branch. This will effectively solve the issue but does change the behaviour of using non existing versions instead of outright failing it will have a fallback mechanic. Would love to get some thoughts on this.

Ultimately this wouldn't work for all cases only when new releases happen so will forgo this method.

```
NOTE: The fix for unit tests has been done in the ruff PR: https://github.com/kedro-org/kedro/pull/3435 (I added a minor fix that was introduced in that PR that I overlooked, it just makes the `cookiecutter_args` into a `fixture` and uses on `class` level instead of `function`)
```

## Checklist


features/tools.feature
```diff
@@ -4,38 +4,8 @@ Feature: New Kedro project with tools
...
     When I run a non-interactive kedro new without starter
     Then the expected tool directories and files should be created with "none"
-    Given I have installed the project dependencies
-    When I execute the kedro command "run"
-    Then I should get a successful exit code
 
   Scenario: Create a new Kedro project with all tools except 'viz' and 'pyspark'
     Given I have prepared a config file with tools "lint, test, log, docs, data"
     When I run a non-interactive kedro new without starter
     Then the expected tool directories and files should be created with "lint, test, log, docs, data"
```
ankatiyar, line 11: Is there any reason to keep this bit?

SajidAlamQB: I guess these are just testing the templates instead of any of the starters this will just be a sanity check.

ankatiyar: Would it make sense to then still have the check for exit code?

SajidAlamQB: You are right, added. 👍

ankatiyar (approved):
LGTM, thanks @SajidAlamQB :)

features/tools.feature
SajidAlamQB: Oh looks like that failed there isn't a 'result' I think the assertions should fine.

Issue #1910, opened 2022-10-06
Add an attribute to dataset classes to flag persistence


jmholzer:
## Description
Adding a boolean attribute (e.g. `IS_PERSISTENT`) to dataset objects would create a clean, consistent interface to check whether a dataset is persistent or not. This will help the development of new features in the long term and will have a direct positive impact on at least two open issues (#1802, #1830).

Currently, the best-possible check for persistence is to check if the dataset object is an instance of `MemoryDataSet`. This is a problem because not all non-persistent DataSets are necessarily instances of `MemoryDataSet`; there are user-defined dataset objects to consider as well as some defined elsewhere in Kedro (`_SharedMemoryDataSet`, for instance).

## Context
The context for this change is that it is similar to other design decisions that we made in the past. For instance, the attribute `_SINGLE_PROCESS` is used to flag whether a dataset can be used with `ParallelRunner`.

This problem has been important in several contexts recently:
- #1802
- #1795
- #1830

## Possible Implementation
I see two ways to implement this interface.

1. We selectively implement the flag on any datasets that are not persistent.
2. We add it to `AbstractDataSet`, with inheriting classes overriding the default definition as necessary.

I think that option 2 is the better one, since it would define a consistent interface across all dataset objects. It also presents a cleaner interface than option 1, which would require all checks for persistence to take the form `if getattr(data_set, "IS_PERSISTENT", False): ...`.

MerelTheisenQB:
Suggestion from @idanov: Change `IS_PERSISTENT` to `_EPHEMERAL` instead, because most datasets are persistent.

yetudada:
This will be needed to support being able to debug a node in a Jupyter notebook because you need to know if the previous dataset in the pipeline is available to be loaded or not, and if not then re-run the pipeline up to the point. 

@noklam We might be able to delay this, it's an improvement on having the line magic for loading a node implemented.

PR #3520, opened 2024-01-17
Add attribute to flag persistence in Dataset classes

lrcouto:

## Checklist

### DIFF
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -113,6 +113,13 @@ class AbstractDataset(abc.ABC, Generic[_DI, _DO]):
             # param2 will be True by default
     """
 
+    """
+    Datasets are persistent by default. user-defined Datasets that
+    are not made to be persistent, such as instances o MemoryDataset,
+    need to change the `_EPHEMERAL` attribute to 'True'.
+    """
+    _EPHEMERAL = False
+
     @classmethod
     def from_config(
         cls: type,
--- a/kedro/io/memory_dataset.py
+++ b/kedro/io/memory_dataset.py
@@ -12,7 +12,8 @@ _EMPTY = object()
 
 class MemoryDataset(AbstractDataset):
     """``MemoryDataset`` loads and saves data from/to an in-memory
-    Python object.
+    Python object. The `_EPHEMERAL` attribute is set to True to
+    indicate MemoryDataset's non-persistence.
 
     Example:
     ::
@@ -33,6 +34,8 @@ class MemoryDataset(AbstractDataset):
         >>> assert reloaded_data.equals(new_data)
 
     """
+
+    _EPHEMERAL = True
 
     def __init__(
         self,
--- a/tests/io/test_memory_dataset.py
+++ b/tests/io/test_memory_dataset.py
@@ -54,6 +54,9 @@ class TestMemoryDataset:
         loaded_data = MemoryDataset(None).load()
         assert loaded_data is None
 
+    def test_ephemeral_attribute(self, memory_dataset):
+        assert memory_dataset._EPHEMERAL is True
+
     def test_load_infer_mode(
         self, memory_dataset, input_data, mocked_infer_mode, mocked_copy_with_mode
     ):


kedro/io/core.py
```diff
@@ -113,6 +113,13 @@ class AbstractDataset(abc.ABC, Generic[_DI, _DO]):
             # param2 will be True by default
     """
 
+    """
+    Datasets are persistent by default. user-defined Datasets that
+    are not made to be persistent, such as instances o MemoryDataset,
+    need to change the `_EPHEMERAL` attribute to 'True'.
```
merelcht, line 119:
```suggestion
    Datasets are persistent by default. User-defined datasets that
    are not made to be persistent, such as instances of `MemoryDataset`,
    need to change the `_EPHEMERAL` attribute to 'True'.
```

kedro/io/lambda_dataset.py
```diff
@@ -100,6 +100,8 @@ def __init__(  # noqa: PLR0913
 
         """
 
+        self._EPHEMERAL = False
```
merelcht, line 103: Why should this be explicitly set to False for the `LambdaDataset`?

kedro/io/shared_memory_dataset.py
```diff
@@ -10,6 +10,8 @@
 class SharedMemoryDataset(AbstractDataset):
     """``SharedMemoryDataset`` is a wrapper class for a shared MemoryDataset in SyncManager."""
 
+    _EPHEMERAL = True
```
merelcht, line 13: For `MemoryDataset` you've only set `self._EPHEMERAL = True` but here we have this line as well as `self._EPHEMERAL = True` in the __init__, why is that?

merelcht (commented):
I've left some questions, but the general gist looks good. I think you'll also need to check if there's any datasets in `kedro-datasets` that will need `_EPHEMERAL` set to True.

kedro/io/shared_memory_dataset.py
lrcouto: That's a leftover from a previous commit, I'll remove it. Thanks for the catch!

astrojuanlu:
One question: What prevents users from creating non-ephemeral datasets with `_EPHEMERAL = True`? Or the other way around?

Also, I don't see an issue linked to this, but I guess this is related to https://github.com/kedro-org/kedro-viz/issues/1709 ?

merelcht:
@astrojuanlu It's [Add an attribute to dataset classes to flag persistence](https://github.com/kedro-org/kedro/issues/1910)

lrcouto:
> One question: What prevents users from creating non-ephemeral datasets with `_EPHEMERAL = True`? Or the other way around?
> 
> Also, I don't see an issue linked to this, but I guess this is related to [kedro-org/kedro-viz#1709](https://github.com/kedro-org/kedro-viz/issues/1709) ?

That is an interest question. Do we want this to be something users have control over (with the possibility for error)? We could use the other implementation suggestion and define a dataset as ephemeral if it inherits from `MemoryDataset`. I had understood that this would be something that users would set up themselves, to allow for this option to be used in user-created datasets.

merelcht (commented):
Don't forget to add this change to the release notes. I also think it would be helpful to add a sentence here https://docs.kedro.org/en/stable/data/how_to_create_a_custom_dataset.html about the `_EPHEMERAL` attribute. 

Otherwise this LGTM 👍

docs/source/data/how_to_create_a_custom_dataset.md
```diff
@@ -38,6 +38,8 @@ At the minimum, a valid Kedro dataset needs to subclass the base [AbstractDatase
 `AbstractDataset` is generically typed with an input data type for saving data, and an output data type for loading data.
 This typing is optional however, and defaults to `Any` type.
 
+The `_EPHEMERAL` boolean attribute present in `AbstractDataset` is used to indicate whether a dataset is meant to be persistent or not, such as in the case of [MemoryDataset](/kedro.io.MemoryDataset). By default, `_EPHEMERAL` is set to False.
```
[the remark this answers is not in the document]
stichbury: Just a bit simpler with a clear example of what ephemeral means (by showing a case when it's true).

stichbury (approved):
Docs changes are spot on, thanks! 🌟

lrcouto:
> Docs changes are spot on, thanks! 🌟

Thank you! I just added one extra sentence to clarify the example using MemoryDataset, as it's meant to be an example of a dataset that's not persistent. The phrasing was a little ambiguous IMO.

PR #3533, opened 2024-01-19
minor tech debts

DimedS:
## Description
Last issues from #3264 :

5. It appears that all file and directory operations are encapsulated within try-except blocks. @SajidAlamQB, have you identified any cases that are not covered?

7. The issue has been addressed in #3499. A default values section was introduced for tools and the example_pipeline. Here, I have inserted a comment in that section stating that the default value for the project name will be sourced from the cookiecutter.json file located in the project's directory.

8. The possibility of integrating flag_inputs into the _get_extra_context() function exists, but such a change would require code modifications. I am not convinced that this alteration would simplify the maintenance of _get_extra_context(). Therefore, I suggest keeping it as is for now.

## Checklist

### DIFF
--- a/kedro/framework/cli/starters.py
+++ b/kedro/framework/cli/starters.py
@@ -560,6 +560,7 @@ def _get_extra_context(  # noqa: PLR0913
         extra_context["example_pipeline"] = str(_parse_yes_no_to_bool(example_pipeline))
 
     # set defaults for required fields, will be used mostly for starters
+    # project_name defaut value will be taken from cookiecutter.json file
     extra_context.setdefault("kedro_version", version)
     extra_context.setdefault("tools", str(["None"]))
     extra_context.setdefault("example_pipeline", "False")


kedro/framework/cli/starters.py
```diff
@@ -560,6 +560,7 @@ def _get_extra_context(  # noqa: PLR0913
         extra_context["example_pipeline"] = str(_parse_yes_no_to_bool(example_pipeline))
 
     # set defaults for required fields, will be used mostly for starters
+    # project_name defaut value will be taken from cookiecutter.json file
```
merelcht, line 563: The name will come from `cookiecutter.json` or from the user input right?

DimedS: Yes, user can provide input from prompt or config file, if not - default value from cookiecutter.json will be used

SajidAlamQB:
> 5. It appears that all file and directory operations are encapsulated within try-except blocks.

In that case no worries, happy to proceed. Thank you @DimedS.

kedro/framework/cli/starters.py
AhdraMeraliQB: To add more context, the project name **must** be provided by the user in one way or another - if omitted from provided config the project creation will fail. If not provided by flag and no config is given, then the user will be required to provide it in the user prompts. The value from cookiecutter.json is used if the user hits enter for this prompt, using the default value that is shown as part of the prompt (currently New Kedro Project) - but this will never be populated directly (or invisibly, rather) from cookiecutter.json

merelcht: Yes, I agree with @AhdraMeraliQB this isn't actually a "default" value. I feel like this comment creates more confusion than clarification so I'd probably vote to close the PR and not merge it.

AhdraMeraliQB:
Closing after brief discussion. It was agreed that there wasn't much to gain from this comment - instead we should concentrate explanation efforts in developer docs. [Xref similar decision](https://github.com/kedro-org/kedro-starters/pull/205).

PR #3543, opened 2024-01-22
Improve error message for invalid example pipeline value

merelcht:
## Description
I noticed that the error message on entering an invalid example pipeline value wasn't very descriptive:

Screenshot 2024-01-22 at 16 19 37

## Development notes
Updated the message to repeat the entered value to the user and mention that the value was invalid specifically for the example pipeline flag:

Screenshot 2024-01-22 at 16 19 14

## Checklist

### DIFF
--- a/tests/framework/cli/test_starters.py
+++ b/tests/framework/cli/test_starters.py
@@ -1133,7 +1133,9 @@ class TestToolsAndExampleFromUserPrompts:
         )
 
         assert result.exit_code != 0
-        assert "is an invalid value for example pipeline." in result.output
+        assert (
+            f"'{bad_input}' is an invalid value for example pipeline." in result.output
+        )
         assert (
             "It must contain only y, n, YES, or NO (case insensitive).\n"
             in result.output
@@ -1311,6 +1313,9 @@ class TestToolsAndExampleFromConfigFile:
         )
 
         assert result.exit_code != 0
+        assert (
+            f"'{bad_input}' is an invalid value for example pipeline." in result.output
+        )
         assert (
             "It must contain only y, n, YES, or NO (case insensitive).\n"
             in result.output
--- a/tests/framework/cli/test_starters.py
+++ b/tests/framework/cli/test_starters.py
@@ -1135,7 +1135,8 @@ class TestToolsAndExampleFromUserPrompts:
         assert result.exit_code != 0
         assert "is an invalid value for example pipeline." in result.output
         assert (
-            "It must contain only y, n, YES, NO, case insensitive.\n" in result.output
+            "It must contain only y, n, YES, or NO (case insensitive).\n"
+            in result.output
         )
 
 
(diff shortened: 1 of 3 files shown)


tests/framework/cli/test_starters.py
```diff
@@ -1311,7 +1312,8 @@ def test_invalid_example(self, fake_kedro_cli, bad_input):
 
         assert result.exit_code != 0
         assert (
-            "It must contain only y, n, YES, NO, case insensitive.\n" in result.output
+            "It must contain only y, n, YES, or NO (case insensitive).\n"
```
AhdraMeraliQB, line 1315: Should we test for the full expected error message here? Like we do for the test above

```
assert f"{bad_input} is an invalid value for example pipeline." in result.output`
```

AhdraMeraliQB (commented):
Looks good to me - just a small comment on one of the tests. We should probably add a test to `TestToolsAndExampleFromCLI` for an invalid `--example` input

tests/framework/cli/test_starters.py
merelcht: Done!

merelcht:
> Looks good to me - just a small comment on one of the tests. We should probably add a test to `TestToolsAndExampleFromCLI` for an invalid `--example` input

This is a good point. However, looking at `test_starters.py` I feel like the full file can use some refactoring. It contains far too many tests IMO and it's really hard to figure out what test should be where. E.g. without reading all tests I don't understand the difference between the classes `TestToolsAndExampleFromUserPrompts` and `TestNewFromUserPromptsValid`. There's also inconsistency in the invalid/valid format. So rather than adding more tests now I think the test file must be refactored.

AhdraMeraliQB (approved):
LGTM - we can add that test case in/after the refactor!

lrcouto (approved):

lrcouto:
> > Looks good to me - just a small comment on one of the tests. We should probably add a test to `TestToolsAndExampleFromCLI` for an invalid `--example` input
> 
> This is a good point. However, looking at `test_starters.py` I feel like the full file can use some refactoring. It contains far too many tests IMO and it's really hard to figure out what test should be where. E.g. without reading all tests I don't understand the difference between the classes `TestToolsAndExampleFromUserPrompts` and `TestNewFromUserPromptsValid`. There's also inconsistency in the invalid/valid format. So rather than adding more tests now I think the test file must be refactored.

Maybe we could separate the tools-related tests to their own file, since there's a decent amount of them.

merelcht:
> Maybe we could separate the tools-related tests to their own file, since there's a decent amount of them.

I've been trying some things here: https://github.com/kedro-org/kedro/pull/3550 
But separating them isn't as easy as I thought.. It creates some issues with the `cookiecutter` cache. I think the main question is how we properly test the use cases where the real starters need to be accessed.

Issue #3560, opened 2024-01-27
CLI Command 'kedro catalog resolve' fails on dataset factories that use PartitionedDataset


MosaicMan:
## Description
Running the cli command 'kedro catalog resolve' fails.

## Context
The error occurs when trying to use a dataset factory pattern with a PartitionedDataset. It appears that the **resolve_patterns** method of the **kedro.framework.cli.catalog** module assumes that every dataset will have a **'filepath'** property explicitly defined. That is not true for partitioned datasets.

## Steps to Reproduce
1. Define a dataset factory pattern in your catalog.yml:
```
"items_{timeframe}":
  type: partitions.PartitionedDataset
  path: /data/models/02_intermediate/items/{timeframe}
  filename_suffix: ".pq"
  overwrite: False
  dataset:
    type: pandas.ParquetDataset
```
2. Make it a part of a pipeline:
```
...
node(
    func=nodes.extract_daily_items,
    inputs='items',
    outputs='items_daily',
    name='extract_daily_items',
),
node(
    func=nodes.extract_weekly_items,
    inputs='items',
    outputs='items_weekly',
    name='extract_weekly_items',
),
...
```
3. Run **'kedro catalog resolve'**

## Expected Result
The different variations produced by the dataset factory should be displayed on screen:

```
items:
  dataset:
    type: pandas.ParquetDataset
  filename_suffix: .pq
  overwrite: true
  path: /data/models/02_intermediate/items
  type: partitions.PartitionedDataset
items_daily:
  dataset:
    type: pandas.ParquetDataset
  filename_suffix: .pq
  overwrite: false
  path: /data/models/02_intermediate/items/daily
  type: partitions.PartitionedDataset
items_weekly:
  dataset:
    type: pandas.ParquetDataset
  filename_suffix: .pq
  overwrite: false
  path: /data/models/02_intermediate/items/weekly
  type: partitions.PartitionedDataset
```

## Actual Result
The command throws an exception:

```
Traceback (most recent call last):
  File "/home/juancq/.pyenv/versions/3.11.7/envs/draco/bin/kedro", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/juancq/.pyenv/versions/3.11.7/envs/draco/lib/python3.11/site-packages/kedro/framework/cli/cli.py", line 198, in main
    cli_collection()
[...]

PR #3561, opened 2024-01-27
Adding a list of "path" keys to check dataset config against.

MosaicMan:
Signed off by MosaicMan <34198823+MosaicMan@users.noreply.github.com>

## Description
Addressing bug as described in #3560. Ran test suite as well as linter. Test status unchanged.

## Development notes
Tweaked the path normalization logic in the resolve_patterns method of the **framework.cli.catalog** module

## Checklist

- [N/A] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [N/A] Updated the documentation to reflect the code changes
- [N/A] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [N/A] Added tests to cover my changes
- [N/A] Checked if this change will affect Kedro-Viz, and if so, communicated that with the Viz team


kedro/framework/cli/catalog.py
```diff
@@ -265,9 +265,14 @@ def resolve_patterns(metadata: ProjectMetadata, env: str) -> None:
...
-                str(context.project_path) + "/", ds_config["filepath"]
-            )
+            path_keys_to_check = ["filepath", "path"]
+
+            for key in path_keys_to_check:
+                if key in ds_config:
+                    ds_config[key] = _trim_filepath(
+                        str(context.project_path) + "/", ds_config[key]
+                    )
+
```
ankatiyar, line 275: @AhdraMeraliQB @merelcht  Could we get rid of this whole check instead of hard-coding this? 
I see this was a design decision (https://github.com/kedro-org/kedro/pull/2891#issuecomment-1665306293)
I think this was before the code went through a few changes. Even for explicit datasets, we just get the config as it is written in the catalog using the config loader so relative path instead of the absolute path so everything is uniform even without this check.

AhdraMeraliQB: Works for me - tried removing this check locally and it all the tests still pass.

merelcht: Ah good point @ankatiyar. If everything still works as expected and the problem with `PartitionedDataset` is solved when the check is removed, then that definitely seems like the best solution.

ankatiyar: in that case, @MosaicMan could you get rid of these lines entirely and the `_trim_filepath()` method below? And add this change to the release notes as well! :)

ankatiyar (approved):
Thank you @MosaicMan for this PR! 💯

merelcht (approved):
Thanks for the contribution @MosaicMan ⭐

Issue #3581, opened 2024-01-31
Completely remove CLI flag deprecation warnings from documentation


AhdraMeraliQB:
## Description

[This section of the docs](https://docs.kedro.org/en/stable/development/commands_reference.html#modifying-a-kedro-run) was missed in #3279, the old flag versions should be completely removed.

PR #3584, opened 2024-02-01
Remove leftover deprecated commands from docs

AhdraMeraliQB:
## Description

Context: #3581 

An oversight from the 0.19.0 release grooming, these commands can be fully removed from our documentation as they have been deprecated for several releases and are no longer supported as of 0.19.0.


stichbury (approved):
LGTM 🌟

merelcht (approved):
Good catch! 🦅

PR #3633, opened 2024-02-19
improve `%load_node` support full node syntax

### LINKED ISSUE #3629: `%load_node` support full Kedro Node syntax, with better *args, **kwargs handling
Copy from https://github.com/kedro-org/kedro/issues/3580#issuecomment-1952285132
> Better handle of `*args` and `**kwargs`, currently the `%load_node` have a simple logic to map node's input to function parameters, it fails with edge cases. Kedro node syntax: https://docs.kedro.org/en/latest/nodes_and_pipelines/nodes.html#syntax-for-input-variables

The idea is to use `inspect.Signature.bind` and `inspect.Parameters`. Additionally, we can identify the special arguments ([VAR_POSITIONAL](https://docs.python.org/3/library/inspect.html#inspect.Parameter.kind)) if needed.

We are already using similar logic to validate node inputs, so this should make sure we support valid inputs only.

https://github.com/kedro-org/kedro/blob/b3637db3f4d07a89076d064307e1fcd777b31d39/kedro/pipeline/node.py#L482-L491

For example:
```python
def dummy(a,b,c, *args, **kwargs):
    ...

node(dummy, ["data_1", "data_2", "data_3", "dummy1","dummy2","dummy3"], ...)
```
should translate to

```python
a = catalog.load("data_1")
b = catalog.load("data_2")
c = catalog.load("data_3")
dummy1 = catalog.load("dummy1")
dummy2 = catalog.load("dummy2")
dummy3 = catalog.load("dummy3")
args = [dummy1, dummy2, dummy3] # Noted here the name of the "dummy_x" variable are arbitrary
dummy(a, b, c, *args)
```

# Requirements
- Include richer sets of test case that support all node input syntax that kedro support.
- 

# Supplementary
- https://github.com/kedro-org/kedro/pull/3623/ is a marginal improvement, but I found it will be difficult to have generic support. This issue propose to use the `inspect.Signature.bind` method to get generic support

noklam:
## Description

Fix #3629
- #3629 
## Development notes

2 main changes:
1. [Refactor to use a `NodeBoundArguments`, which inherit from `inspect.BoundArguments`. ](https://github.com/kedro-org/kedro/pull/3633/commits/bb1574017529934c5b4d39dea340d8342e15b5e3). This allow the helper function to return a proper data structure for testing. The structure will also be useful for generating all the necessary text so our tests can focus on testing on the data structure, not the exact string.
2. Using `BoundArguments`, we can easily generate the necessary function signature according to the node function, the only thing we need to do is taking care of the "*args" to handle a specific edge case.
3. 

## Checklist


noklam:
@lrcouto It's ready for test, I have refactor the logic and manual tested and unit tests are all passed.

lrcouto:
@astrojuanlu here's the issue I mentioned today at standup.

image

This function at `inspect.BoundNode` has a docstring that contains `*args` and `**kwargs`. Sphinx is considering the `*` characters to be a bold text tag that was opened and not closed, and I can't change the docstring since it's on an external lib. =p

noklam:
Was the issue Sphinx is trying to render that method in docs? I have not idea what sphinx cares about it.

lrcouto:
> Was the issue Sphinx is trying to render that method in docs? Can we somehow suppress this, tbh there is not much value to document the class in the API docs, we can even convert it to a private class `_NodeBoundArguments` if it helps.

I think it is. I've been trying to find a way to either suppress warnings from external libraries, or suppress this specific warning, but I don't think I can do that without ignoring the whole file. I'll try to see if converting it to a private class helps.

lrcouto:
Converted NodeBoundArguments to private and that solved this specific instance of the problem. It'd still be a good idea to find a more permanent solution for this issue though.

noklam:
@lrcouto Great! I think we have seen this issue in the past, I don't really remember how we solved it, maybe @AhdraMeraliQB or @astrojuanlu knows more? The solution is to not use some of the `autodoc` that trigger upstream dependencies, turning into private class is sufficient for now.

noklam:
Tagging @DmitriiDeriabinQB and @ankatiyar for review

astrojuanlu:
The other @DimedS :)

astrojuanlu:
The way I've usually solved it is by committing the template generated by sphinx.ext.autosummary during the docs building process and removing the classes or properties I don't want to have documented. Sphinx will then abstain from overwriting the file.

Good that you solved the problem for now 👍🏽 I don't think this was expected to be public API anyway?

astrojuanlu (approved):
Tested this, and it works!

Issue #3591, opened 2024-02-03
Consider changing the default setting of logging.yml


noklam:

astrojuanlu:
I've said it in https://github.com/kedro-org/kedro/issues/3446#issuecomment-1936278169, but for the record, +1 to this. From https://12factor.net/logs:

> **A twelve-factor app never concerns itself with routing or storage of its output stream.** It should not attempt to write to or manage logfiles. Instead, each running process writes its event stream, unbuffered, to stdout.

astrojuanlu:
Another point in favour of having `kedro -vv`: sometimes you just want to know what kedro is doing without having to configure the whole stdlib logging. This is how most other tools work.

Here is an example of a user that's puzzled because `kedro` just hangs https://linen-slack.kedro.org/t/16630828/i-m-encountering-issues-when-running-the-load-ext-kedro-ipyt#3758284a-34fb-435c-b054-dd3713cd78c9

Issue #3446, opened 2023-12-19
Use of `.env` and enable logging automatically


noklam:
## Introduction

Previous discussion:
- https://github.com/kedro-org/kedro/issues/2428
- https://github.com/kedro-org/kedro/issues/2281#issuecomment-1510914273

In previous discussion we banned the idea of introducing `.env` to enable logging, we introduce a new environment variable `KEDRO_LOGGING_CONFIG` that user need to set in their environment (maybe in terminal or `settings.py`).

With the new `kedro new` flow, when `logging` is selected, it is not automatically used until user set `KEDRO_LOGGING_CONFIG` manually.

However, setting it in `settings.py` against the Twelve factor apps principles. I found that it was quite easy to enable the `.env` feature because Dynaconf support this natively (it's just one argument).

Basically, 
`settings = _ProjectSettings()` become `settings = _ProjectSettings(load_dotenv=True)`

# Discussion
Should we enable automatically loading up a `.env` file? I have seen people loading the .env file in `settings.py` manually, some are confused where should they put this file because it's not clear to them which file in a Kedro project get loaded first.

Putting it in `settings.py` is not ideal either because logging is configured before settings, thus it will have no effect on logging.

Pro:
- I find that it is quite awkward to find a place to hold this environment variables. Kedro has `conf` for holding project config so naturally users want to keep these configuration in their workspace.
- Often people need to set their credentials such as S3 etc. They can export it manually in terminal, but eventually this become a tedious process and the automation step is using `.env` which is a common practice.
- This may also help with handling credentials across different env (base/local/prod/dev) because you don't really want to commit any of these credentials in the codebase. Kedro's credentials doesn't solve this problem completely beyond `local`.
Cons:
- https://github.com/kedro-org/kedro/issues/2281#issuecomment-1510914273

One major change is that we introduce the add-ons (or tools) to add logging with more flexibility. 

# Summary
To summarise, the problem I had is how to enable logging automatically, `.env` is an option that could help with this problem and potentially helps with managing credentials.

astrojuanlu:
Not sure about this. There's a myriad mechanisms to manage environment variables, for example:

- direnv https://direnv.net/ uses `.envrc` (and, only optionally, `.env`)
- rtx https://github.com/jdx/rtx uses its own config file:

```
$ cat .rtx.toml
[tools]
python = {version="3.10.11", virtualenv=".venv"}

[env]
[...]

Issue #2208, opened 2023-01-16
Work out to show DEBUG level messages emitted early in kedro process


AntonyMilneQB:
Currently any `DEBUG` level messages emitted by logging calls before `session._setup_logging` will never be shown, even if you alter your project logging.yml to show `DEBUG` level messages. This is because framework-side logging sets kedro logging info to level `INFO` (which is the right default, we shouldn't change that to `DEBUG`). This isn't a big deal but would be nice if users had a way to show the messages if they want.

Options:
- preferred: https://github.com/kedro-org/kedro/issues/2206
- some sort of a verbosity flag to change logging level (note we already have a verbosity option on some commands, but it looks like it only changes how detailed `KedroCliError` message is) e.g. https://hydra.cc/docs/tutorials/basic/running_your_app/logging/

AntonyMilneQB:
This will be fixed by #2206 since you'll then be able to specify `DEBUG` as a level when framework logging config is used.

Issue #1732, opened 2022-07-28
default logging configuration doesn't log other packages than kedro


Debbby57:
## Description
since the 0.18.2 release, I lost the logs for the other packages than kedro in my projects, if I don't specify them in the logging.yaml
I tested with multiple packages like custom kedro plugins like kedro-mlflow (@Galileo-Galilei) and other packages like paramiko. 
If I want to have logs for thoses packages I need to update logging.yml

```yaml
loggers:
    kedro:
        level: INFO
    <my_package>:
        level: INFO
    kedro-mlflow:
        level: INFO
    paramiko:
        level: INFO
```
I think it's a bug because for me  it's difficult to ask users to add all their packages in the logging.yaml. I think the default behavior we want is to have all logs by default and use logging.yaml to remove logs we don't want.

## Context
I need all my logs by default to control the correct behavior of my projects. And if I don't want some of them I remove them

## Steps to Reproduce
1. install kedro 0.18.2
2. install kedro plugin like kedro-mlflow or another package like paramiko
3. run your project. You will see that you don't have kedro-mlflow logs or paramiko's logs
4. update your logging.yaml like above and run your project. You will have logs

## Expected Result
By default, we should have all logs.

## Actual Result
We just have kedro logs.

## Your Environment
Include as many relevant details about the environment in which you experienced the bug:

* Kedro version used (`pip show kedro` or `kedro -V`): 0.18.2
* Python version used (`python -V`): 3.9.13
* Operating system and version: linux

Thank you for your help :)

Galileo-Galilei:
This is a quite annoying issue I am facing too : [All tests of this PR](https://github.com/Galileo-Galilei/kedro-mlflow/runs/7508041379?check_suite_focus=true) on ``kedro-mlflow`` were working before the release and are failing since ``kedro==0.18.2``. The logs shows that the test fail because ``kedro-mlflow`` logging is no longer displayed in the console. 

The fix in ``logging.yml`` works, but like @Debbby57 I don't feel it is the right default behaviour (and it is a breaking change comapred to ``0.18.1``, so I think this is not intentional).

Debbby57:
thank you for your reply. 
I need to think about this but for now : 
* I still don't like option 1, because as a user I don't want to search and list packages for which I want infos logs. I could be very time-consuming.
* short term, I'm going to use option 2
* I don't like option 3 but I'm just a user of kedro-plugins so I think my opinion is really facultative ;)
[...]

PR #3657, opened 2024-02-27
[Draft] - logging

noklam:
## Description

- https://github.com/kedro-org/kedro/issues/3591
- https://github.com/kedro-org/kedro/issues/3446

This draft PR is not ready for merge, but rather facilitate the discussion of the issues above.

## Development notes

Changes:
1. Read logging.yml automatically - be3d28d981ecf51d8fef96a256d9c0ef74d5d8c4
2. Change default of `logging.yml`- 65cf2a0b82afdf18a4f9e1f56af061b6a80f25ee . This challenge the idea of "project logging". Why is `logger` necessary? It's more common to have module level logger, which propagate log to the "root logger", and set the level for root logger only. The exception here is "info logger" because it should only filter `INFO` level. If anything `logger` should be optional, not default. I also think it's more likely you want to keep the level consistent for any `logger` you have manually configured, it's rare you want `kedro` at `INFO` but project at `DEBUG` (This become less important if we end up introduce `-v` or `-vv`
3. Add kedro run -v 2ec8bc1b218b608f094f1c81d2b1a37c44cceda8 . The main problem for this is CLI argument is slightly too late, there are chance that some log is produced before a "run" start. There is also question of extending this to all CLI command, because `logging.yml` affect any Kedro CLI. While this introduce a slight inconsistency, I don't think this is a big problem. As I envision this is mainly for debug use.

I imagine `-v`, `-vv` is a runtime config that used mostly for debugging. If users has sophisticated configuration, they can continue to use `logging.yml` to have different group of settings (I haven't seen much tbh)

Alternatlive, I found `pip` utilise something called a [VerboseLogger](https://github.com/pypa/pip/blob/0ad4c94be74cc24874c6feb5bb3c2152c398a18e/src/pip/_internal/utils/_log.py#L16) 

## Checklist

### DIFF
--- a/kedro/framework/cli/project.py
+++ b/kedro/framework/cli/project.py
@@ -196,6 +196,13 @@ def package(metadata: ProjectMetadata) -> None:
     help=PARAMS_ARG_HELP,
     callback=_split_params,
 )
+@click.option(
+    "-v",
+    "verbose",
+    type=click.UNPROCESSED,
+    is_flag=True,
+    default="",
+)
 def run(  # noqa: PLR0913
     tags: str,
     env: str,
@@ -212,8 +219,13 @@ def run(  # noqa: PLR0913
     conf_source: str,
     params: dict[str, Any],
     namespace: str,
+    verbose: bool,
 ) -> None:
     """Run the pipeline."""
+    if verbose:
+        from kedro.framework.project import LOGGING
+        LOGGING.data["root"]["level"] = "DEBUG"
+        LOGGING.configure(LOGGING.data)
 
     runner_obj = load_obj(runner or "SequentialRunner", "kedro.runner")
     tuple_tags = tuple(tags)
--- a/kedro/framework/project/__init__.py
+++ b/kedro/framework/project/__init__.py
@@ -217,9 +217,14 @@ class _ProjectLogging(UserDict):
     def __init__(self) -> None:
         """Initialise project logging. The path to logging configuration is given in
         environment variable KEDRO_LOGGING_CONFIG (defaults to default_logging.yml)."""
-        path = os.environ.get(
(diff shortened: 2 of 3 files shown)


noklam:
Quick notes:
@antonymilne give some good insight:

- Automatic loading logging is fine, but there are edge cases with `CONF_SOURCE`. I think it's still an improvement and it doesn't make anything worse in the current states.
- We should never change `root` logger level by default, it's too disruptive. We can however, try to catch `kedro-xxx` pattern, but always respect the package logging level if it exists.
  - Alternative is we required user to add a new logger in logging.yml manually, would be annoying for plugins. 
- `kedro run` should still default to `INFO`, so nothing change in terms of default user behavior, `-v` to change to `DEBUG` level. 

Should `kedro run -v` only affect the project logging, or `kedro`(and plugins)?

antonymilne:
Some more information now I've found the old issues I wanted to refer to....

* https://github.com/kedro-org/kedro/issues/1732 has a very thorough explanation of why it's a bad idea to set the `root` logger level to `INFO`. I don't think it's possible to catch `kedro-*` as a pattern, but `kedro.*` is already caught by the `kedro` logger. This means a plugin called `kedro-plugin` logging with namespace `kedro.plugin` which is a bit weird though. Better is to just set the logging level to `INFO` in `kedro-plugin` itself, as @Galileo-Galilei did in https://github.com/Galileo-Galilei/kedro-mlflow/pull/354
* https://github.com/kedro-org/kedro/issues/2208 is the only comment I could find that I made on a verbose mode - I thought I had written something else more useful about this but couldn't find it. Basically I think `-v/--verbose` is a great idea though
* milestones https://github.com/kedro-org/kedro/milestone/5 and https://github.com/kedro-org/kedro/milestone/14 well carefully documented and tracked so definitely worth looking back on the parent issues there if you need more context on how we ended up with the current solution

Issue #3626, opened 2024-02-17
Plain logging contains rich color codes


sbrugman:
## Description
Following the instructions [here](https://docs.kedro.org/en/stable/logging/index.html#how-to-use-plain-console-logging), the logging will still outputs color codes.

## Expected Result

```txt
Loading data from datasetname (MyDataset).
```

## Actual Result
```txt
Loading data from [dark_orange]datasetname[/dark_orange] (MyDataset).
```

## Your Environment

* Kedro version used (0.19.2):

sbrugman:
The markup should either be added conditional on that the rich handler is used, or stripped afterwards.

Instead of including the markup directly in the log, it could be parametrized:
```python
def fmt(value: str, color: str):
    return f"[{color}]{value}[/{color}]"

...

# Configurable, inferred from the logging handler, or conditional on `rich` installed.
markup = False

  self._logger.info(
      "Loading data from %s (%s)...",
      markup(name, color="dark_orange") if markup else name,
      type(dataset).__name__,
      extra={"markup": markup},
  )
```

instead of 
```python
        self._logger.info(
            "Loading data from [dark_orange]%s[/dark_orange] (%s)...",
            name,
            type(dataset).__name__,
            extra={"markup": True},
        )
```
https://github.com/kedro-org/kedro/blob/main/kedro/io/data_catalog.py#L487

For my use case it would be enough to disable the markup if `rich` is not installed, but no strong preference.

Alternatively, you could include a NoRichHandler that strips out the formatting. This would be unintuitive (why need to do additional work for requiring _fewer_ features), but it works.
 
Rich markup reference: https://rich.readthedocs.io/en/stable/markup.html

noklam:
I think a simple condition is enough here, #3657 will not be merged soon feel free to open a PR @sbrugman.

astrojuanlu:
I assume this was fixed in #3682.

PR #3682, opened 2024-03-04
Disable rich markup if rich handler is not present

sbrugman:
## Description
Fix for https://github.com/kedro-org/kedro/issues/3626

Only used in Data Catalog (2x)

Done in Github UI to start the discussion.
Will add unit tests, and add release notes.

## Checklist


kedro/logging.py
```diff
@@ -58,3 +58,13 @@ def __init__(self, *args: Any, **kwargs: Any):
             # fixed on their side at some point, but until then we disable it.
             # See https://github.com/Textualize/rich/issues/2455
             rich.traceback.install(**traceback_install_kwargs)  # type: ignore[arg-type]
+
+
+def has_rich_handler(logger: logging.Logger) -> bool:
+    """Returns true if the logger has a RichHandler attached."""
+    return any(isinstance(handler, RichHandler) for handler in logger.handlers)
+
```
noklam, line 66: Is it possible to save this result to avoid calling this hundred times? It's probably have insignificant performance cost.

sbrugman: Added `functools.cache`. Side questions: has there been any profiling of Kedro? What are the most expensive parts of the framework?

josephfinlayson:
I would like this to go in please

kedro/logging.py
noklam: @sbrugman Sorry I missed the notification. I think it depends on what you are targeting.

#1476 is trying to optimise the CLI performance, those are insignificant time for a `kedro run` but significant for CLI UX.

For a `kedro run`, "it depends". The slowest one is always the node computation itself or the dataset I/O. Kedro itself does not have too much penalty, it can be even faster than a script if you can use some of the tricks to speed up (parallel runner, async I/O etc).

The main "penalty" I would say is saving & loading the same data, which could be reduced by using `CacheDataset` or something similar to avoid the unnecessary loading/blocking time.

noklam:
Just leaving a comment here we didn't forget about this PR as we discussed internally.

I will shortly review this PR.

sbrugman:
Thanks, will do

lrcouto:
> minor comments:
> 
>     * `fmt_rich` -> `_format_rich`, not a big improvement but I don't think we use `fmt` anywhere in the codebase
> 
>     * modify the method to internals `_` one, I don't think any of these should be public API at this point.

Seconding this.

lrcouto:
Reverting the previous merge I did because it broke some tests. Probably gonna need a rebase because of the DCO issue though.

kedro/logging.py
```diff
@@ -54,3 +55,15 @@ def __init__(self, *args: Any, **kwargs: Any):
             # fixed on their side at some point, but until then we disable it.
             # See https://github.com/Textualize/rich/issues/2455
             rich.traceback.install(**traceback_install_kwargs)  # type: ignore[arg-type]
+
+
+# @cache is supported from 3.9 onwards
```
deepyaman, line 60: Nit: I would just leave out this comment; pyupgrade will take care of it when drop 3.8.

noklam (approved):

lrcouto (approved):

PR #3683, opened 2024-03-05
Make kedro commands work from inside subdirectories in project

### LINKED ISSUE #1831: Make kedro project commands work from any directory inside a project
> **Note**. This ticket does not handle the question of what happens if you're outside the project altogether. That is done in #1829. This ticket is just about what happens if you're inside the project but not at project root.

> **Warning**. This shouldn't be difficult, but the correct implementation needs a bit of careful thought and testing to get right!

When a kedro project command (e.g. `kedro run`) is run from inside a project but not the project root it doesn't work. There doesn't seem to be any good reason for this. Other tools (.e.g black, git) work from project subfolders, so let's do the same with kedro.

* iterate up through directory structure to see if it's a kedro project. If yes then use that as the project path; if no then behave as now (i.e. will say "run command not found")
* we have code to do this already in `_find_kedro_project` in the ipython extension. This would likely be made more performant if it used iterated through `current_dir.parents` rather than the while loop that's currently there. Do a quick bit of profiling to see what the best way to do this is
* once the best way is found, we should put the function in a common place like `utils` so the ipython extension and CLI can both use it
* for a start, for `kedro` to show all the project commands from inside a subdirectory you'll need to change `KedroCLI(project_path=Path.cwd())` to `KedroCLI(project_path=_find_kedro_project(Path.cwd()))`
* in addition, anywhere that starts a `KedroSession` will need to use the correct project root. This could be done by providing the right `project_path` whenever `KedroSession` is instantiated or (probably better) altering the `KedroSession` itself from `self._project_path = Path(project_path or Path.cwd()).resolve()` to `self._project_path = Path(project_path or _find_kedro_project(Path.cwd()).resolve())`
* there are very likely other places that need to change too! This needs some care to get right

> **Warning**. The change _shouldn't_ actually `cd` the user to a different directory. i.e. this is the correct behaviour:
> ```
> pwd # /Users/antony_milne/kedro_stuff/projects/iris/data
> kedro run # should work as if I ran it from /Users/antony_milne/kedro_stuff/projects/iris/
> pwd # /Users/antony_milne/kedro_stuff/projects/iris/data - hasn't changed!
> ```
> 
>

ankatiyar:
## Description

Fix #1831 
## Development notes

- Move `_find_kedro_project()` and `_is_project()` to `kedro.utils` so it can be used both by `KedroCLI` and `kedro.ipython` extension
- Update `_find_kedro_project()` to use `Path.parents` 
- Update `project_path` for `KedroCLI` and `KedroSession` to use `_find_kedro_project()`
- Added e2e test for this change

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,6 +1,7 @@
 # Upcoming Release 0.19.4
 
 ## Major features and improvements
+* Kedro commands now work from any subdirectory within a Kedro project.
 
 ## Bug fixes and other changes
 * Updated `kedro pipeline create` and `kedro pipeline delete` to read the base environment from the project settings.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -7,6 +7,7 @@
 * Updated `kedro pipeline create` and `kedro pipeline delete` to read the base environment from the project settings.
 
 ## Breaking changes to the API
+* Methods `_is_project` and `_find_kedro_project` have been moved to `kedro.utils`. We recommend not using private methods in your code, but if you do, please update your code to use the new location.
 
 ## Documentation changes
 
--- a/kedro/framework/session/session.py
+++ b/kedro/framework/session/session.py
@@ -27,6 +27,7 @@ from kedro.framework.project import (
 from kedro.framework.session.store import BaseSessionStore
 from kedro.io.core import generate_timestamp
 from kedro.runner import AbstractRunner, SequentialRunner
+from kedro.utils import _find_kedro_project
 
 
 def _describe_git(project_path: Path) -> dict[str, dict[str, Any]]:
@@ -104,7 +105,9 @@ class KedroSession:
         save_on_close: bool = False,
         conf_source: str | None = None,
     ):
-        self._project_path = Path(project_path or Path.cwd()).resolve()
+        self._project_path = Path(
+            project_path or _find_kedro_project(Path.cwd()).resolve() or Path.cwd()
+        )
         self.session_id = session_id
         self.save_on_close = save_on_close
         self._package_name = package_name
--- a/kedro/framework/session/session.py
+++ b/kedro/framework/session/session.py
@@ -106,8 +106,8 @@ class KedroSession:
         conf_source: str | None = None,
     ):
         self._project_path = Path(
-            project_path or _find_kedro_project(Path.cwd()).resolve() or Path.cwd()
-        )
+            project_path or _find_kedro_project(Path.cwd()) or Path.cwd()
+        ).resolve()
         self.session_id = session_id
         self.save_on_close = save_on_close
         self._package_name = package_name
--- a/features/run.feature
+++ b/features/run.feature
@@ -60,3 +60,11 @@ Feature: Run Project
     When I execute the kedro command "run --params extra1=1,extra2=value2"
     Then I should get a successful exit code
     And the logs should show that 4 nodes were run
+
+  Scenario: Run kedro run from within a sub-directory
+    Given I have prepared a config file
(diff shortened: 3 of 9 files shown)


noklam:
@datajoely #3680

noklam (approved):
I did a manual test with arbitary deep directory with `kedro run`,`kedro ipython`, all works well. This is such a small change but big improvement in the developer experience. 

It checks all the parents, which I think it's good. Alternatively we are try to traverse the directory up, we run into risks that we execute other kedro project accidentally.

image





## Out of scope:
@rashidakanchwala I test viz command and it's failing.
image

noklam:
Since this breaks `kedro-viz`, I also send a question in [Slack](https://kedro-org.slack.com/archives/C03R8N2M8KT/p1709740227090809). In some sense we may consider this as a breaking change, mainly to plugin developer. Though I don't think this should stop us from releasing this in 0.19, but just in case we should check with some plugin developers particularly if they rely on the fact that command are running from project root.

ravi-kumar-pilla:
> Since this breaks `kedro-viz`, I also send a question in [Slack](https://kedro-org.slack.com/archives/C03R8N2M8KT/p1709740227090809). In some sense we may consider this as a breaking change, mainly to plugin developer. Though I don't think this should stop us from releasing this in 0.19, but just in case we should check with some plugin developers particularly if they rely on the fact that command are running from project root.

Hi Nok, Currently the commands `kedro viz run` and `kedro viz build` default to current working directory as Kedro project path. We are introducing an option for `--project-path` in `kedro viz run` [here](https://github.com/kedro-org/kedro-viz/pull/1792). We can also plan on extending this to other cli commands. Thank you

kedro/framework/session/session.py
```diff
@@ -104,7 +105,9 @@ def __init__(  # noqa: PLR0913
         save_on_close: bool = False,
         conf_source: str | None = None,
     ):
-        self._project_path = Path(project_path or Path.cwd()).resolve()
+        self._project_path = Path(
+            project_path or _find_kedro_project(Path.cwd()).resolve() or Path.cwd()
```
merelcht, line 109: Previously we called `.resolve()` on all options `project_path` as well as `Path.cwd()` and now only on `_find_kedro_project(Path.cwd()).resolve()`, why did that change?

RELEASE.md
```diff
@@ -1,6 +1,7 @@
 # Upcoming Release 0.19.4
 
 ## Major features and improvements
+* Kedro commands now work from any subdirectory within a Kedro project.
```
merelcht, line 4: `_is_project` and `_find_kedro_project` are private methods, but it might be worth mentioning in the release notes section "Breaking changes to the API" that these were moved to `kedro.utils`, just in case people are using them.

merelcht (commented):
The implementation looks all good and I've tested it successfully as well 👍 I agree with @noklam that even though it can be considered breaking, it should be shipped in the 0.19.x series. 

On the viz issue we could either change the logic there, or at the least make the error message clearer when running Viz in a subdirectory.

kedro/framework/session/session.py
ankatiyar: Updated!

RELEASE.md
ankatiyar: Updated!

merelcht (approved):
LGTM! 😄

astrojuanlu:
Wow, this was so fast I didn't even have time to review it. Well done everyone ❤️

noklam:
run-kedro-anywhere

PR #3719, opened 2024-03-18
Improve resume suggestions

ondrejzacha:
## Description

Supersedes #3026 

Addresses #3002:
- Parameter inputs are considered persistent, therefore a failed node with a parameter input won't necessarily trigger re-run of their ancestor nodes.

Additionally:
- Only nodes producing non-persistent inputs are suggested for rerun (as opposed to all parent nodes of a node with 1+ non-persistent inputs)
- Number of suggested nodes to re-run from is minimised (-> shorter message for the same pipeline)
- Refactored into a testable function (therefore more applicable to other runners too)

## Development notes

Additional tests added in `test_sequential_runner.py` – both to check the message in logs for sequential runner specifically, as well as a new suite of tests for the suggestion logic itself. I appreciate that the logic is not exactly straightforward, so please double check the tests / let me know if more tests are needed.

## Checklist


kedro/runner/runner.py
```diff
@@ -198,39 +198,62 @@ def _suggest_resume_scenario(
...
-                "argument to your previous command:\n%s",
-                len(remaining_nodes),
-                postfix,
+                f"argument to your previous command:\n{postfix}"
             )
 
 
-def _find_persistent_ancestors(
-    pipeline: Pipeline, children: Iterable[Node], catalog: DataCatalog
+def find_nodes_to_resume_from(
```
noklam, line 222: Is there a reason why some of the methods are private methods, but the `find_xxxx` are all public?

ondrejzacha: You're right, that's inconsistent. I find that this method could be useful also outside of this context so I'd keep it public but I made the others private

kedro/runner/runner.py
```diff
@@ -242,54 +265,113 @@ def _find_persistent_ancestors(
...
+        of nodes of interest.
 
     """
-    parent_pipeline = pipeline.only_nodes_with_outputs(*child.inputs)
-    return parent_pipeline.nodes
+    p_nodes_of_interest = pipeline.only_nodes(*(n.name for n in nodes_of_interest))
+    p_nodes_with_external_inputs = p_nodes_of_interest.only_nodes_with_inputs(
+        *p_nodes_of_interest.inputs()
+    )
+    return set(p_nodes_with_external_inputs.nodes)
```
noklam, line 310: The function name is `_node_with_external_inputs` but its arguments is a list of node. If you are trying to search for pipeline node that has external_inputs, can you use `pipeline.inputs` and then `pipeline.only_nodes_with_inputs`?

kedro/runner/runner.py
```diff
@@ -242,54 +265,113 @@ def _find_persistent_ancestors(
...
-        if _has_persistent_inputs(current_node, catalog):
-            ancestor_nodes_to_run.add(current_node)
-            continue
-        for parent in _enumerate_parents(pipeline, current_node):
-            if parent in visited:
+        nodes_to_run.add(current_node)
+        non_persistent_inputs = _enumerate_non_persistent_inputs(current_node, catalog)
+        # Look for the nodes that produce non-persistent inputs (if those exist)
+        for node in _enumerate_nodes_with_outputs(pipeline, non_persistent_inputs):
+            if node in visited:
```
noklam, line 280: ✨

kedro/runner/runner.py
```diff
@@ -198,39 +198,62 @@ def _suggest_resume_scenario(
...
+        pipeline: the ``Pipeline`` to find starting nodes for.
+        unfinished_nodes: collection of ``Node``s that have not finished yet
+        catalog: the ``DataCatalog`` of the run.
+
+    Returns:
+        Set of node names to pass to pipeline.from_nodes() to continue
+        the run.
+
+    """
+    all_nodes_that_need_to_run = find_all_required_nodes(
```
noklam:
```suggestion
    nodes_to_be_run = find_all_required_nodes(
```
?

kedro/runner/runner.py
```diff
@@ -198,39 +198,62 @@ def _suggest_resume_scenario(
...
+        pipeline, unfinished_nodes, catalog
+    )
+
+    # Find which of the remaining nodes would need to run first (in topo sort)
+    persistent_ancestors = find_initial_node_group(pipeline, all_nodes_that_need_to_run)
```
noklam: Is this needed for "Number of suggested nodes to re-run from is minimised (-> shorter message for the same pipeline)" this?

kedro/runner/runner.py
```diff
@@ -198,39 +198,64 @@ def _suggest_resume_scenario(
...
+
+    """
+    all_nodes_that_need_to_run = _find_all_required_nodes(
+        pipeline, unfinished_nodes, catalog
+    )
+
+    # Find which of the remaining nodes would need to run first (in topo sort)
+    persistent_ancestors = _find_initial_node_group(
+        pipeline, all_nodes_that_need_to_run
+    )
```
noklam, line 247: I find this a bit hard to understand, is it similar to find
```python
p = pipeline(all_nodes_tht_need_to_run)
p_inputs = p.inputs()
input_nodes = p.only_nodes_with_input(p_inputs)
input_nodes_names = [n.name for n in input_nodes]
```
It's just pseudocode so I don't guarantee it works

noklam (commented):
I left some comments, and manually tested and it works well.

I have some concerns to open up a lot of public methods which is unclear to me how will it be useful outside of this context (Maybe for using it in notebook? Would love to see some example of this if that's the case)

kedro/runner/runner.py
ondrejzacha: Good point, simplified that

kedro/runner/runner.py
ondrejzacha: This is not exactly the same. The difference appears when nodes share the same external inputs:
```
Pipeline(
    node(
        name="first_node",
        func=...,
        inputs=["external_input"],
        outputs="intermediate_output",
    ),
    node(
        name="second_node",
        func=...,
        inputs=["external_input", "intermediate_output"],
        outputs="final_output",
    )
)
```
Your suggestion would produce `["first_node", "second_node"]` here, whereas with the topo sort group approach, only the `["first_node"]` is produced (as this pipeline will have two topo sort groups, each containing a single node).

ondrejzacha:
> I have some concerns to open up a lot of public methods which is unclear to me how will it be useful outside of this context (Maybe for using it in notebook? Would love to see some example of this if that's the case)

Sure, I don't have a strong preference here. In my work I've found myself pulling these methods out of the runner code to figure out how to continue a run of a pipeline which did not fail gracefully (ran into an OOM error and never showed this suggestion) – but I can imagine that can very much be an exception

kedro/runner/runner.py
```diff
@@ -198,39 +198,64 @@ def _suggest_resume_scenario(
...
 ) -> set[Node]:
     """Breadth-first search approach to finding the complete set of
-    persistent ancestors of an iterable of ``Node``s. Persistent
-    ancestors exclusively have persisted ``Dataset``s as inputs.
+    ``Node``s which need to run to cover all unfinished nodes,
+    including any additional nodes that should be re-run if their outputs
+    are not persisted.
 
     Args:
         pipeline: the ``Pipeline`` to find ancestors in.
```
merelcht, line 261: I can't comment on the line below, but the doc string is outdated and still mentions `children` as one of the input arguments.

kedro/runner/runner.py
```diff
@@ -242,54 +267,110 @@ def _find_persistent_ancestors(
...
+    # Make sure no downstream tasks are skipped
+    nodes_to_run = pipeline.from_nodes(*(n.name for n in nodes_to_run)).nodes
+
+    return set(nodes_to_run)
 
 
-def _enumerate_parents(pipeline: Pipeline, child: Node) -> list[Node]:
-    """For a given ``Node``, returns a list containing the direct parents
-    of that ``Node`` in the given ``Pipeline``.
+def _nodes_with_external_inputs(nodes_of_interest: Iterable[Node]) -> set[Node]:
```
merelcht, line 291: I'm struggling a bit to understand this method. With "external inputs", do you mean those not produced by nodes in the pipeline?

kedro/runner/runner.py
```diff
@@ -242,54 +267,110 @@ def _find_persistent_ancestors(
...
+    # We use _datasets because they pertain parameter name format
+    catalog_datasets = catalog._datasets
+    non_persistent_inputs: set[str] = set()
     for node_input in node.inputs:
-        if isinstance(catalog._datasets[node_input], MemoryDataset):
-            return False
-    return True
+        if node_input.startswith("params:"):
+            continue
+        if node_input not in catalog_datasets or isinstance(
```
merelcht, line 327: Quoting the comment @noklam left on the previous PR:

"New thought: https://github.com/kedro-org/kedro/pull/3520 - any chance we can utilise the new flag that we added?"

merelcht (commented):
Thanks for this improved PR @ondrejzacha ! 🙂 On first review, things look generally good to me. I've left a couple of clarification questions. 

The tests look good as well, but I agree with your comment on moving it to a separate test file. The last set of tests aren't actually using any specific runner so it might confuse others coming across them to understand whether this is `SequentialRunner` specific or not.

kedro/runner/runner.py
ondrejzacha: Indeed, I added that to the docstring. If there's better terminology for this I'll be happy to update it

noklam (approved):
Thanks for making all the changes and it is taking so long. This is a great improvement and it's one step closer to rerunning pipeline in an efficient way.

kedro/runner/runner.py
```diff
@@ -198,98 +198,178 @@ def _suggest_resume_scenario(
...
+    """
+    nodes_to_be_run = _find_all_required_nodes(pipeline, unfinished_nodes, catalog)
+
+    # Find which of the remaining nodes would need to run first (in topo sort)
+    persistent_ancestors = _find_initial_node_group(pipeline, nodes_to_be_run)
+
+    return {n.name for n in persistent_ancestors}
+
+
+def _find_all_required_nodes(
```
DimedS, line 248: Perhaps it would be better to use a more descriptive name, such as _find_all_nodes_to_resume_from().

kedro/runner/runner.py
```diff
@@ -198,98 +198,178 @@ def _suggest_resume_scenario(
...
+            if node in visited:
                 continue
-            visited.add(parent)
-            queue.append(parent)
-    return ancestor_nodes_to_run
+            visited.add(node)
+            queue.append(node)
 
+    # Make sure no downstream tasks are skipped
+    nodes_to_run = set(pipeline.from_nodes(*(n.name for n in nodes_to_run)).nodes)
```
DimedS, line 282: Is it necessary? It seems that the previous algorithm already identifies all the required nodes. Moreover, in the subsequent step, we'll only be selecting nodes from the beginning anyway.

kedro/runner/runner.py
```diff
@@ -198,98 +198,178 @@ def _suggest_resume_scenario(
...
+        A set containing all input unfinished ``Node``s and all remaining
+        ``Node``s that need to run in case their outputs are not persisted.
 
     """
-    ancestor_nodes_to_run = set()
-    queue, visited = deque(children), set(children)
+    nodes_to_run = set(unfinished_nodes)
+    initial_nodes = _nodes_with_external_inputs(unfinished_nodes)
+
+    queue, visited = deque(initial_nodes), set(initial_nodes)
```
DimedS, line 269: Do we actually need `visited` here, or could `nodes_to_run` be used instead? They appear to be similar.

DimedS (approved):
Thank you, @ondrejzacha , for the significant update to the resume suggestions algorithm and the development of high-quality tests! It's fantastic that the suggestions are now effectively minimised, excluding nodes that have already been successfully processed. I've left some minor comments, but overall, the PR looks great to me!

kedro/runner/runner.py
ondrejzacha: It is not strictly necessary (the output would be equivalent), but using `visited` prevents from processing certain nodes multiple times – parent nodes with multiple children.

kedro/runner/runner.py
ondrejzacha: It wouldn't matter for the final pipeline created with `pipeline.from_nodes()`, but in some cases (see test case with `pipeline_triangular`), this could produce an inaccurate output of `_find_all_required_nodes`.

kedro/runner/runner.py
ondrejzacha: I agree that this is not very specific – however I find this alternative a little misleading – it returns all nodes in the resumed pipeline, not just those suggested for `--from-nodes`. Perhaps `_find_all_nodes_for_resumed_pipeline` could work?

kedro/runner/runner.py
DimedS: Thank you for explanation, it's a tricky example. I think it really depends on what you expect from _find_all_required_nodes(). Looks like node2 in that example shouldn’t run again because it already did its job well and saved its results. So, it's not clear for me why we're including that node to results of that func.

kedro/runner/runner.py
DimedS: I mean looks like we can use `nodes_to_run` for the same purpose, not duplicating mostly the same values in `visited`, but it's ok to leave it as it is.

kedro/runner/runner.py
DimedS: thanks, sound great!

DimedS (approved):
Thank you for addressing comments, @ondrejzacha !

kedro/runner/runner.py
ondrejzacha: My thinking is that if `node1` func is not deterministic and `node2` depends on its output, when we don't rerun `node2`, we end up with inconsistent pipeline outputs.

ondrejzacha:
I think from my side this is ready; the docs build is still failing so let me know if there's anything I can still do. Excited to see this merged, thanks for the reviews! 🤩

noklam:
@ondrejzacha amazing, thanks again really appreciate the effort🙏 I will take over and merge this.

DimedS (approved):
Thank you for responding to the comments, @ondrejzacha. Fantastic work on the PR!

Issue #2912, opened 2023-08-08
Replace `.absolute()` with `.resolve()`


noklam:
## Description

Potentially helps
- #2059 
- https://github.com/kedro-org/kedro/issues/2819
## Context

See https://github.com/kedro-org/kedro/issues/2819#issuecomment-1669748465
See docs:
> Path.resolve(strict=False)[¶](https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve)
Make the path absolute, resolving any symlinks. A new path object is returned:
Python Discussion Thread: https://discuss.python.org/t/pathlib-absolute-vs-resolve/2573/11

![image](https://github.com/kedro-org/kedro/assets/18221871/95ff5625-7dd4-4c68-97bc-2014e53fe2f4)

The only way to remove `..` in a path is using `.resolve()`, this is useful when user need to instantiate `DataCatalog` in a notebook, because the project root is in `../` one level up.

astrojuanlu:
About symlinks, notice that in https://github.com/kedro-org/kedro/issues/2346 I was advocating for the opposite: replacing `.resolve()` with `.absolute()`.

> this is useful when user need to instantiate DataCatalog in a notebook, because the project root is in ../ one level up.

If we do away with the magic path conversion as proposed in https://github.com/kedro-org/kedro/issues/2965, is this still needed?

noklam:
Actually I have a second look. This is the only case that in our codebase use `absolute()`
https://github.com/kedro-org/kedro/blob/a91ccdbecb5541490b5146b6179ab4c621b9ea73/kedro/framework/cli/utils.py#L204-L218

This is the only leftover we used `.absolute()` in the codebase.


https://github.com/kedro-org/kedro/issues/2819#issuecomment-1669771871
Additionally, we may always do a `resolve` in ConfigLoader so even if user provide a relative path it would work. I will add this comment to #2971, depends on the implementation we may avoid this completely with the new `root` argument.

This ticket will remains as a clean up ticket.

merelcht:
Do we still need to do this?

astrojuanlu:
I actually think we should close this and do the opposite: https://github.com/kedro-org/kedro/issues/2346

noklam:
@merelcht It didn't change by #3742. 

https://github.com/kedro-org/kedro/blob/a91ccdbecb5541490b5146b6179ab4c621b9ea73/kedro/framework/cli/utils.py#L204-L218

[...]

Issue #2346, opened 2023-02-20
Symlink resolution during configuration detection might break some CLI commands


astrojuanlu:
## Description
While debugging another issue, I found an interesting behavior that only happens in very specific circumstances. Since [on macOS `/tmp` is a symlink to `/private/tmp`](https://apple.stackexchange.com/questions/1043/why-is-tmp-a-symlink-to-private-tmp), while trying to instantiate a `KedroCLI` object I was getting a long error:

```
In [2]: from kedro.framework.cli.cli import KedroCLI

In [3]: cli = KedroCLI("/tmp/test-kedro-ipython-regular/")
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
...
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
ValueError: Source path '/private/tmp/test-kedro-ipython-regular/src' has to be relative to your project root 
'/tmp/test-kedro-ipython-regular/'.
```

which of course was addressed by using the resolved path instead:

```
In [10]: cli = KedroCLI("/private/tmp/test-kedro-ipython-regular/")

In [11]: cli
Out[11]: <KedroCLI None>
```

## Context
I guess this never happens in real life because of how the codebase calls the initializer, but looks like the brittleness could be averted by replacing `.resolve()` with `.absolute()` here:

https://github.com/kedro-org/kedro/blob/116ddd015e81d2a6930a0dfbe83e630e526634f4/kedro/framework/startup.py#L64

I'm not sure if there are other consequences potentially hidden. This line has been there "forever" (dc39a96fa386).

## Steps to Reproduce
(given above)

## Expected Result
The `KedroCLI` can be instantiated in paths regardless of symlinks.

## Your Environment
Include as many relevant details about the environment in which you experienced the bug:

* Kedro version used (`pip show kedro` or `kedro -V`): 0.18.4
* Python version used (`python -V`): 3.10.9
* Operating system and version: macOS Ventura 13.2.1

merelcht:
Test if this also breaks datasets or only the CLI.

noklam:
Opposite idea, what about we make sure both `project_path` and `source_path` are `resolved`?

noklam:
Windows do have something similar to symlink, used to do that with Powershell

astrojuanlu:
[...]

PR #3742, opened 2024-03-26
Resolve all path in Kedro

noklam:
## Description

Related:
- https://github.com/kedro-org/kedro/issues/2912
- #2346 
- 
## Development notes

```python

def _add_src_to_path(source_dir: Path, project_path: Path) -> None:
   ... 

def bootstrap_project(project_path: Path) -> ProjectMetadata:
    """Run setup required at the beginning of the workflow
    when running in project mode, and return project metadata.
    """
    project_path = Path(project_path).resolve() # New
    metadata = _get_project_metadata(project_path)
    _add_src_to_path(metadata.source_dir, project_path)
    return metadata
```

If you look at the `bootstrap_project`, it expect a path object but in fact sometimes string is passed and it is okay about this. Same go as `_add_src_to_path` that accepts `Path` only but actually taking `str` sometimes.

The fix here is simple, make sure `project_path` is resolved. It is already done in many place, just not consistently (sometimes with `expanduser()` too):
https://github.com/kedro-org/kedro/blob/35ee3eadb1a30557d2c4bf28a96c7369a2e4c4a0/kedro/framework/startup.py#L52-L53

and

https://github.com/kedro-org/kedro/blob/35ee3eadb1a30557d2c4bf28a96c7369a2e4c4a0/kedro/framework/startup.py#L93-L95

- [ ] Todo: add a test

## Checklist

### DIFF
--- a/kedro/framework/startup.py
+++ b/kedro/framework/startup.py
@@ -1,4 +1,6 @@
 """This module provides metadata for a Kedro project."""
+from __future__ import annotations
+
 import os
 import sys
 from pathlib import Path
--- a/kedro/framework/startup.py
+++ b/kedro/framework/startup.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 import os
 import sys
 from pathlib import Path
-from typing import NamedTuple, Union
+from typing import NamedTuple
 
 import toml
 
@@ -36,7 +36,7 @@ def _version_mismatch_error(kedro_init_version: str) -> str:
     )
 
 
-def _get_project_metadata(project_path: Union[str, Path]) -> ProjectMetadata:
+def _get_project_metadata(project_path: Path) -> ProjectMetadata:
     """Read project metadata from `<project_root>/pyproject.toml` config file,
     under the `[tool.kedro]` section.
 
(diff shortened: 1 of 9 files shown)


noklam:
[3831e82](https://github.com/kedro-org/kedro/pull/3742/commits/3831e826900b1c0037339d359e33a7f317ebbe80) this seems to work already

kedro/framework/startup.py
```diff
@@ -34,7 +36,7 @@ def _version_mismatch_error(kedro_init_version: str) -> str:
     )
 
 
-def _get_project_metadata(project_path: Union[str, Path]) -> ProjectMetadata:
```
noklam, line 37: I hesitate a bit on whether it should accept both `str | Path`. As you can see it breaks one of our test.

On the other hand it's much cleaner to keep this path conversion happen at the beginning, i.e. `bootstrap_project`. Otherwise we are doing the same conversion over and over, and this is the only public function.

merelcht (approved):
LGTM! 👍 It makes total sense to do the path resolution as early on in the flow as possible.

ankatiyar (approved):
Thanks @noklam! Also came across this issue on window tests in https://github.com/kedro-org/kedro-plugins/pull/599.

Issue #1271, opened 2022-02-21
Agree and document the best practices we recommend for how to write tests for nodes and pipelines


yetudada:
## Description

Our users have expressed a need to learn how to write tests for their nodes and pipelines. We encourage learning about software-engineering best practice and want to include a guide in our documentation for this. 

## Possible Implementation
This guide must focus on explaining:

- How `kedro test` works
- How to write a test for a node, using an example
- How to write a test for a pipeline, using an example

This guide can fall into our **Development** chapter. And remember to follow our guidelines for [contributing to the documentation](https://kedro.readthedocs.io/en/stable/14_contribution/04_documentation_contributor_guidelines.html).

datajoely:
FYI @benhorsburgh

deepyaman:
> How to write a test for a pipeline, using an example

I strongly believe that end-to-end pipeline tests should not be written as unit tests, as they retest the behavior of each contained node. Happy to discuss if contentious (I think a lot of people ascribe to running pipelines with data in pytest), and further willing to contribute to writing this part.

datajoely:
@deepayman end-to-end pipeline tests can be written as integration tests right?

datajoely:
If at all useful, please [steal what you can from here](https://protocol.quantumblack.com/docs/alloy/docs/source/03_tutorial/testing.html).

daBlesr:
I'm not able to access those links, but in any case, in my experience writing tests for pipeline came to its full power when doing operations on Spark DataFrames. Usually, for pipelines, one test with data suffices. Here is a suggestion for what this page could look like using Spark DataFrames:

### Writing Tests for Nodes and Pipelines

In this section we introduce the way you write unit tests and integration tests for Nodes and Pipelines respectively. Each Node in a pipeline should have its own (parametrised) unit tests. Tests for pipelines on the other hand should test a number of sequential nodes. For such a test, the input of the first node is tested against the expected output of the last node. Imagine the following nodes, were we compute the total equity of a store's inventory:
```python
# pipelines/inventory/nodes.py
def process_total_price_per_product(inventory: DataFrame):
    return (
        inventory
            .withColumn("total_prices", col("in_store") * col("price"))
            .select("name", "total_price")
    )

def process_total_equity(total_price_per_product: DataFrame):
    return (
        total_price_per_product
            .groupBy()
            .sum('total_price')
            .first()
            [0]  # first col
    ) or 0.0
```
The pipeline:
```python
[...]

Issue #3418, opened 2023-12-12
Describe how to use custom code with Kedro


astrojuanlu:
## Description
> In general, i struggle to find any documentation on how to place/use/import custom libraries with kedro

https://linen-slack.kedro.org/t/16143547/hey-all-very-new-and-excited-user-of-kedro-here-hopefully-i-#600585f1-60b6-48e3-b0b6-6d1efe01480a

## Context

This is natural for people who understand that Kedro projects are mostly Python libraries, but

1. This was not the case in Kedro<0.19, and
2. The fact that `kedro run` injects the project source into the path on startup hides the fact that users can, and should, install their own code.

https://github.com/kedro-org/kedro/blob/44a3d170fc15be2139f8fafcc647431c5f506aeb/kedro/framework/startup.py#L147-L148

stichbury:
@astrojuanlu If possible, pleasecould you add a bit more detail of where this content should go in the docset and who could contribute it. It looks like something that engineering should write at first draft. What's your priority here -- high-ish?

astrojuanlu:
I'd say this is Medium-Low priority, and something that definitely engineering should start.

I'd need a bit more time to propose an outline. There are some loose ends in our user journeys that I think would impact this content:

- Are we encouraging users to install their own code _for development_ anywhere? The only place where `pip install [-e] .` is mentioned is in the packaging section of the tutorial https://docs.kedro.org/en/stable/tutorial/package_a_project.html which probably needs revisiting already:

> In addition, it was always weird to have those instructions in the "Package an entire Kedro project" page - `kedro package` is about distributing the code (wheel + config), docs have nothing to do with it.

_Originally posted by @astrojuanlu in https://github.com/kedro-org/kedro/issues/3376#issuecomment-1836354148_

- Are we encouraging users to configure their IDEs assuming that they will install their own code? Looking at https://docs.kedro.org/en/stable/development/set_up_vscode.html and https://docs.kedro.org/en/stable/development/set_up_pycharm.html, I'd say the answer is "no".

We could tackle this task without revisiting these user journeys, but I think we risk introducing some inconsistent information, or making the page too long or complex.

I'll give this some more thought in the near future.

yuryfedotov-mck:
Starting on this per [conversation](https://kedro-org.slack.com/archives/C03RKP2LW64/p1711294524985879) with @noklam

rashidakanchwala:
Maybe we add examples in our docs/kedro-academy to showcase this.

[...]

PR #3782, opened 2024-04-05
Document best practices  for writing tests for nodes and pipelines

AhdraMeraliQB:
## Description

Addresses #1271 

Added the page "Test a Kedro project" to the spaceflights tutorial. The page details the structure of writing a positive and negative unit test for a node function, and an integration test for (part of) a pipeline, with examples. It then discusses some best practices for testing, including the directory setup, naming conventions, and use of fixtures.

The page on automated testing was updated to reflect the test directory structure we use in our template projects, and to include instructions on locally installing the Kedro project before testing, resolving the issues seen around imports from the project into the tests - [xref on Slack](https://kedro-org.slack.com/archives/C03RKP2LW64/p1710950639001399?thread_ts=1710948385.470859&cid=C03RKP2LW64)

These changes will need to be updated on starters (PR paused - pending docs approval) - https://github.com/kedro-org/kedro-starters/pull/216

## Development notes

[See the built docs here](https://kedro--3782.org.readthedocs.build/en/3782/tutorial/test_a_project.html)

The example tests have been tested in a spaceflights project (both unrefined and refactored versions) and pass as expected. These docs have pulled inspiration from some of the internal pages linked on the original issue.

## Checklist


docs/source/tutorial/test_a_project.md
```diff
@@ -0,0 +1,335 @@
...
+                 }
+    return parameters
+```
+
+</details>
+
+We can then access these through the test arguments.
+
+```python
+def test_split_data(self, dummy_data, dummy_parameters):
```
astrojuanlu, line 239: No `self` here, I believe

```suggestion
def test_split_data(dummy_data, dummy_parameters):
```

docs/source/tutorial/test_a_project.md
```diff
@@ -0,0 +1,335 @@
...
+def dummy_parameters():
+    parameters = {"model_options":
+                     {"test_size": 0.2,
+                      "random_state": 3,
+                      "features": ["engines", "passenger_capacity", "crew"]}
+                 }
+    return parameters
+
+
+class TestDataScienceNodes:
```
astrojuanlu, line 290: Can we _not_ use classes? 😄

docs/source/tutorial/test_a_project.md
```diff
@@ -0,0 +1,335 @@
...
+        assert len(y_train) == 2
+        assert len(X_test) == 1
+        assert len(y_test) == 1
+
+class TestDataSciencePipeline:
+    def test_data_science_pipeline(self, dummy_data, dummy_parameters):
+        pipeline = create_pipeline().from_nodes("split_data_node").to_nodes("evaluate_model_node")
+        catalog = DataCatalog()
+        catalog.add("model_input_table", MemoryDataset(dummy_data))
+        catalog.add_feed_dict({"params:model_options" : dummy_parameters["model_options"]})
```
astrojuanlu, line 303: Maybe merge these two calls into one `add_feed_dict`?

docs/source/tutorial/test_a_project.md
```diff
@@ -0,0 +1,335 @@
...
+
+        output = SequentialRunner().run(pipeline, catalog)
```
astrojuanlu, line 305: What is this output supposed to have? Maybe a bit more discussion on the runner in the previous paragraph would help

astrojuanlu (commented):
Thanks a lot for putting this together @AhdraMeraliQB !

I gave a very quick first pass

docs/source/development/automated_testing.md
```diff
@@ -96,6 +96,17 @@ Tests should be named as descriptively as possible, especially if you are workin
...
+
+Specify your source directory in `pyproject.toml` for pytest to properly discover your Kedro project. This is necessary due to the nested directory structure.
+
+```toml
+[tool.pytest.ini_options]
+pythonpath = [
+  "src"
+]
+```
+
```
noklam, line 109: I am not particular familiar with this config, but looks like kedro starter don't have this config?

AhdraMeraliQB: Opened a PR to add it, it resolves the issue of "Module spaceflights not found" when importing from the project

docs/source/tutorial/test_a_project.md
```diff
@@ -0,0 +1,335 @@
...
+    y = data["price"]
+    X_train, X_test, y_train, y_test = train_test_split(
+        X, y, test_size=parameters["test_size"], random_state=parameters["random_state"]
+    )
+    return X_train, X_test, y_train, y_test
+```
+
+</details>
+
+The function takes a pandas DataFrame and dictionary of parameters as input, and splits the input data into 4 different data objects as per the parameters provided. We recommend following [pytest's anatomy of a test](https://docs.pytest.org/en/7.1.x/explanation/anatomy.html#anatomy-of-a-test) which breaks a test down into 4  steps: arrange, act, assert, and cleanup. For this specific function, these steps will be:
```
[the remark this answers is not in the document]
AhdraMeraliQB: I used the split data function from the tutorial and didn't catch the divergence, I'll update both docs pages accordingly

docs/source/tutorial/test_a_project.md
```diff
@@ -0,0 +1,335 @@
...
+
+We define our node functions as [pure functions](https://realpython.com/python-functional-programming/#what-is-functional-programming); a pure function is one whose output value follows solely from its input values, without any observable side effects such as changes to state or mutable data. When testing these functions we check that for a given set of input values, a node will produce the expected output. This type of test is referred to as a unit test.
+
+Let us explore what this looks like in practice. Consider the node function `split_data` defined in the data science pipeline:
+
+<details>
+<summary><b>Click to expand</b></summary>
+
+```python
+def split_data(data: pd.DataFrame, parameters: Dict) -> Tuple:
```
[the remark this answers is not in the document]
datajoely: I don't have a huge problem with `dict` vs `Dict` but...

- I do think we shouldn't encourage the generic `parameter` block over the `params:specific.key` pattern 
- We should also have a specific type hint, i.e. `Dict[str, Union[str,int]` no `Any`!

docs/source/tutorial/test_a_project.md
```diff
@@ -0,0 +1,335 @@
...
+
+        # Act
+        X_train, X_test, y_train, y_test = split_data(dummy_data, dummy_parameters["model_options"])
+
+        # Assert
+        assert len(X_train) == 2
+        assert len(y_train) == 2
+        assert len(X_test) == 1
+        assert len(y_test) == 1
+```
```
datajoely, line 88: Let's have a unhappy case and some failure cases too. @Nab-0 has a good tutorial on this

docs/source/tutorial/test_a_project.md
```diff
@@ -0,0 +1,335 @@
...
+        catalog.add("model_input_table", MemoryDataset(dummy_data))
+
+        dummy_parameters = {"model_options":
+                     {"test_size": 0.2,
+                      "random_state": 3,
+                      "features": ["engines", "passenger_capacity", "crew"]}}
+        catalog.add_feed_dict({"params:model_options" : dummy_parameters["model_options"]})
+
+        # Act
+        output = SequentialRunner().run(pipeline, catalog)
```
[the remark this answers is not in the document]
datajoely: This is also why I think we need that `kedro-test` framework to bootstrap these internal for users.

docs/source/tutorial/test_a_project.md
```diff
@@ -0,0 +1,335 @@
...
+        {"engines": [1, 2, 3],
+         "crew": [4, 5, 6],
+         "passenger_capacity": [5, 6, 7],
+         "price": [120, 290, 30]})
+        catalog.add("model_input_table", MemoryDataset(dummy_data))
+
+        dummy_parameters = {"model_options":
+                     {"test_size": 0.2,
+                      "random_state": 3,
+                      "features": ["engines", "passenger_capacity", "crew"]}}
```
datajoely, line 170: If we use explicit (`params:x`) this is much more readable

docs/source/tutorial/test_a_project.md
AhdraMeraliQB: Is it poor practice? I was following the convention of the pre-existing example tests in our starters

docs/source/tutorial/test_a_project.md
AhdraMeraliQB: The pipeline has no outputs, this was a trivial check to have an explicit assert that would confirm the pipeline ran successfully. I'll take a look at how to better capture that

docs/source/development/automated_testing.md
astrojuanlu: This works, but I think the better solution is to tell users to install their own code:

```
pip install -e .
```

See also https://github.com/kedro-org/kedro/issues/3418

docs/source/tutorial/test_a_project.md
astrojuanlu: I don't think it's a bad practice per se but I think it's inherited from the old `unittest` (itself inspired in JUnit) way of doing things, and I think it's unnecessary

docs/source/tutorial/test_a_project.md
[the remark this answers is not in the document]
astrojuanlu: For unit tests of pipelines like these though it should suffice to do `runner.run` right?

docs/source/tutorial/test_a_project.md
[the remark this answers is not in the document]
AhdraMeraliQB: Updated the type hint to `dict[str, Any]` as we use in our own tests in [221c13a](https://github.com/kedro-org/kedro/pull/3782/commits/221c13a8f3a491ab77e307227550d4235331d4d3).

Re: `parameters` Vs `params:model_options` - this example isn't the most clear, and conflates the node function definition in nodes.py (above) and the node definition in the pipeline - I believe the latter is where we recommend using `params:...` over `parameters`. In the spaceflights tutorial we do employ this practice, using `params:model_options` for `split_data_node`'s inputs. 

The argument in the node function above is named `parameters` (which is confusing) but could be anything. I've avoided changing it as I wanted to preserve the spaceflights starter code as much as possible, but I'm open to suggestions. I considered `model_parameters` but I'm not sure if that accurately captures what the parameters are for - any thoughts? Maybe `node_parameters`? Or `splitting_parameters`?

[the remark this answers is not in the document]
astrojuanlu: To clarify, this is more about going from

```python
def split_data(data: pd.DataFrame, parameters: dict[str, Any]) -> Tuple:
    X = data[parameters["features"]]
    y = data["price"]
        X, y, test_size=parameters["test_size"], random_state=parameters["random_state"]
    )
    ...

            node(
                func=split_data,
                inputs=["model_input_table", "params:model_options"],
```

to

```python
def split_data(data: pd.DataFrame, features: list[str], test_size: float, random_state: int) -> Tuple:
    X = data[features]
    y = data["price"]
    )
    ...

            node(
                func=split_data,
                inputs=["model_input_table", "params:model_options.features", "params:model_options.test_size", "params:model_options.random_state"],
```

does it make sense?

docs/source/development/automated_testing.md
AhdraMeraliQB: Changed in [9339ad1](https://github.com/kedro-org/kedro/pull/3782/commits/9339ad14acc54c88f28e290fb85795e273322701)

docs/source/tutorial/test_a_project.md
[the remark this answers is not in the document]
AhdraMeraliQB: Just bumping this - @noklam

docs/source/tutorial/test_a_project.md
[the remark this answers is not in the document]
AhdraMeraliQB: Using SequentialRunner is in line with [our other documented recommendations](https://docs.kedro.org/en/stable/nodes_and_pipelines/run_a_pipeline.html#run-pipelines-with-io) - though I'm open to suggestions on other recommendations (with the priority of maintaining consistency across our docs)

docs/source/tutorial/test_a_project.md
AhdraMeraliQB: Added an unhappy/negative test case in [91e9ec0](https://github.com/kedro-org/kedro/pull/3782/commits/91e9ec07627f8e45082b8b5143510a64c72df645). 

I'm wary about adding errors to `split_data` to be found by failing cases as users may stop the tutorial before going through testing, incorrectly assuming that the implemented logic is sound. 

I considered adding a section about Test Driven Development instead, but it's difficult to recommend TDD in the tutorial, as the tutorial itself does not practice TDD 😅. Any thoughts?

docs/source/tutorial/test_a_project.md
AhdraMeraliQB: I believe all the calls in the tests use the explicit `params:model_options`, is there somewhere else this could be used?

docs/source/tutorial/test_a_project.md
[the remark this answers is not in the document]
astrojuanlu: @AhdraMeraliQB this is my only outstanding comment I believe. I appreciate that writing the code in this way might not be consistent with the starters, or how we're doing it in other places. Should we open a separate issue to track this?

docs/source/development/automated_testing.md
```diff
@@ -19,6 +19,17 @@ There are many testing frameworks available for Python. One of the most popular
...
+Before getting started with `pytest`, it is important to ensure you have installed your project locally. This allows you to test different parts of your project by importing them into your test files.
+
+To install your project, navigate to your project root and run the following command:
+
+```bash
+pip install -e .
+```
+
+>**NOTE**: The option `-e` installs an editable version of your project, allowing you to make changes to the project files without needing to re-install them each time.
```
astrojuanlu, line 32: Maybe let's use native MyST syntax

````suggestion
:::{note}
The option `-e` installs an editable version of your project, allowing you to make changes to the project files without needing to re-install them each time.
:::
````

https://myst-parser.readthedocs.io/en/latest/syntax/admonitions.html

(fenced admonitions should also work)

**Review by astrojuanlu — APPROVED** [review]
Other than a minor formatting comment and another one that probably should become an issue in itself, this looks great ⭐

**Review by ankatiyar — APPROVED** [review]
Looks good to me! Thanks @AhdraMeraliQB 💯

File: docs/source/tutorial/test_a_project.md [hunk 12]
```diff
@@ -0,0 +1,429 @@
...
+
+* How to test a Kedro node
+* How to test a Kedro pipeline
+* Testing best practices
+
+
+This section does not cover:
+
+* Automating your tests - instead read our [automated testing documentation](../development/automated_testing.md).
+* More advanced features of testing, including [mocking](https://realpython.com/python-mock-library/#what-is-mocking) and [parametrizing tests](https://docs.pytest.org/en/7.1.x/example/parametrize.html).
```
merelcht, line 15:
```suggestion
* More advanced features of testing, including [mocking](https://realpython.com/python-mock-library/#what-is-mocking) and [parameterising tests](https://docs.pytest.org/en/7.1.x/example/parametrize.html).
```

docs/source/tutorial/test_a_project.md
```diff
@@ -0,0 +1,429 @@
...
+
+
+## Writing tests for Kedro nodes: Unit testing
+
+Kedro expects nodes functions to be  [pure functions](https://realpython.com/python-functional-programming/#what-is-functional-programming); a pure function is one whose output follows solely from its inputs, without any observable side effects. Testing these functions checks that a node will behave as expected - for a given set of input values, a node will produce the expected output. These tests are referred to as unit tests.
```
merelcht, line 20:
```suggestion
Kedro expects node functions to be [pure functions](https://realpython.com/python-functional-programming/#what-is-functional-programming); a pure function is one whose output follows solely from its inputs, without any observable side effects. Testing these functions checks that a node will behave as expected - for a given set of input values, a node will produce the expected output. These tests are referred to as unit tests.
```

docs/source/tutorial/test_a_project.md
```diff
@@ -0,0 +1,429 @@
...
+    y = data["price"]
+    X_train, X_test, y_train, y_test = train_test_split(
+        X, y, test_size=parameters["test_size"], random_state=parameters["random_state"]
+    )
+    return X_train, X_test, y_train, y_test
+```
+
+</details>
+
+The function takes a pandas DataFrame and dictionary of parameters as input, and splits the input data into 4 different data objects as per the parameters provided. We recommend following [pytest's anatomy of a test](https://docs.pytest.org/en/7.1.x/explanation/anatomy.html#anatomy-of-a-test) which breaks a test down into 4  steps: arrange, act, assert, and cleanup. For this specific function, these steps will be:
```
merelcht, line 47:
```suggestion
The function takes a pandas `DataFrame` and dictionary of parameters as input, and splits the input data into four different data objects as per the parameters provided. We recommend following [pytest's anatomy of a test](https://docs.pytest.org/en/7.1.x/explanation/anatomy.html#anatomy-of-a-test) which breaks a test down into four steps: arrange, act, assert, and cleanup. For this specific function, these steps will be:
```

docs/source/tutorial/test_a_project.md
```diff
@@ -0,0 +1,429 @@
...
+
+1. Arrange: Prepare the inputs `data` and `parameters`.
+2. Act: Make a call to `split_data` and capture the outputs with `X_train`, `X_test`, `Y_train`, and `Y_test`.
+3. Assert: Ensure that the length of the outputs are the same as the expected lengths
+
+Cleanup is not required for this test. Remember to import the function being tested and any necessary modules at the top of the file.
```
merelcht, line 53: Perhaps explain why cleanup isn't needed?

docs/source/tutorial/test_a_project.md
merelcht: I think explaining TDD here goes a bit beyond just best practices for coding, because it goes into actual ways of working which is a whole other story and not really something I feel like we have a firm stance on as the Kedro team.

docs/source/tutorial/test_a_project.md
```diff
@@ -0,0 +1,429 @@
...
+                inputs=["regressor", "X_test", "y_test"],
+                outputs=None,
+                name="evaluate_model_node",
+            ),
+        ]
+    )
+```
+</details>
+
+The pipeline takes a DataFrame and dictionary of parameters as input, splits the data in accordance to the parameters, and uses it to train and evaluate a regression model. With an integration test, we can validate that this sequence of nodes runs as expected. As we did with our unit tests, we break this down into several steps:
```
merelcht, line 181:
```suggestion
The pipeline takes a pandas `DataFrame` and dictionary of parameters as input, splits the data in accordance to the parameters, and uses it to train and evaluate a regression model. With an integration test, we can validate that this sequence of nodes runs as expected. As we did with our unit tests, we break this down into several steps:
```

docs/source/tutorial/test_a_project.md
```diff
@@ -0,0 +1,429 @@
...
+                "model_input_table" : dummy_data,
+                "params:model_options": dummy_parameters["model_options"],
+            }
+        )
+
+        # Act
+        SequentialRunner().run(pipeline, catalog)
+
+        # Assert
+        assert True     # trivial assert
```
merelcht, line 236: Is there maybe a logging statement we could check?

docs/source/tutorial/test_a_project.md
```diff
@@ -0,0 +1,429 @@
...
+We can then access these through the test arguments.
+
+```python
+def test_split_data(dummy_data, dummy_parameters):
+        ...
+```
+
+### Pipeline slicing
+
+In the test `test_data_science_pipeline` we test the that outputs of the data science pipeline are those that are expected. However, as pipelines are not static, this test is not robust. Instead we should be specific with how we define the pipeline to be tested; we do this by using [pipeline slicing](../nodes_and_pipelines/slice_a_pipeline.md#slice-a-pipeline-by-running-specified-nodes) to specify the pipeline's start and end:
```
merelcht, line 310: "we test the that outputs of the data science pipeline are those that are expected", at the moment we don't do this, but test that the pipeline runs without errors, that's why you needed to add the trivial `assert` statement. I think it would be better if we could indeed check some sort of output, even if it's just a logging statement.

docs/source/development/automated_testing.md
AhdraMeraliQB: The MyST syntax doesn't seem to render properly on RTD. Additionally, the syntax of `>**NOTE**` is currently consistent across the docs - I'm unsure of the motivation to change it

docs/source/tutorial/test_a_project.md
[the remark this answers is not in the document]
AhdraMeraliQB: I'd prefer to track this in a separate issue - the change on starters would be somewhat minimal but I believe it would also require some significant reworking of the suggestions in our [parameters docs](https://docs.kedro.org/en/stable/configuration/parameters.html), which would fall out of this ticket's scope.

docs/source/tutorial/test_a_project.md
```diff
@@ -0,0 +1,446 @@
...
+
+        # Arrange the log testing setup
+        caplog.set_level(logging.DEBUG, logger="kedro") # Ensure all logs produced by Kedro are captured
+        successful_run_msg = "Pipeline execution completed successfully."
+
+        # Act
+        SequentialRunner().run(pipeline, catalog)
+
+        # Assert
+        assert successful_run_msg in caplog.text
```
merelcht, line 247: 👍

docs/source/tutorial/test_a_project.md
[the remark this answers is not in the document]
astrojuanlu: Opened https://github.com/kedro-org/kedro/issues/3818 to track this

merelcht (approved):
This looks great! Thanks @AhdraMeraliQB ✨

docs/source/development/automated_testing.md
astrojuanlu: True, we needed the `colon_fence` extension

PR #3795, opened 2024-04-09
Suggested fix for `ParallelRunner`

ElenaKhaustova:

## Checklist

### DIFF
--- a/kedro/runner/runner.py
+++ b/kedro/runner/runner.py
@@ -3,6 +3,7 @@ implementations.
 """
 from __future__ import annotations
 
+import copy
 import inspect
 import itertools as it
 import logging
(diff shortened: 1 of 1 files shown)


kedro/runner/runner.py
```diff
@@ -17,6 +18,7 @@
 )
 from typing import Any, Collection, Iterable, Iterator
 
+import pandas as pd
```
merelcht, line 21: This would mean having to add `pandas` as a dependency and especially after the discussion we had today I don't think that will be a good idea 😅

ElenaKhaustova: Yeah, that's why it doesn't seem like a good solution. Otherwise, we need to copy objects in all cases.

From https://github.com/scikit-learn/scikit-learn/issues/28781 it seems like a `scikit-learn` problem when multiprocessing. So the current plan is to report our case - just need to create an example without kedro-related code.

ElenaKhaustova:
Closing as agreed to open an issue on the `scikit-learn` side instead.

astrojuanlu:
Please xref from here when you open the issue or share the link, would love to subscribe there

Issue #3801, opened 2024-04-11
Make `logging.yml` read by default


noklam:
> Quick summary:
> - :white_check_mark:There are at least one obvious user problem - which is logging get chosen as a tool but has no effect, and the consensus is that we should fix it. - https://github.com/kedro-org/kedro/issues/3446
>   - One possible solution is hard coding the path to conf/logging.yml with option to override with KEDRO_LOGGINGI_CONFIG, very similar to https://github.com/kedro-org/kedro/commit/be3d28d981ecf51d8fef96a256d9c0ef74d5d8c4
> - It's unclear whether we should do more for logging,[ such as adding a new -v flag](https://github.com/kedro-org/kedro/issues/3591)
> - It's mentioned logging in development and logging in production are very different, and users probably have different needs:
>   - It was mentioned we may want to de-couple rich as a plugin, pip-uninstallable to deactivate it or at least provide an easy enough way to opt out from it.
> - We may consider improving the messaging of logging and people should seek for production grade logging.
> - It was mentioned even though there are users need, Kedro can still say NO as we had for many other things.
> 
> Antony shared the old user research here in case you miss it at the end: https://github.com/kedro-org/kedro/issues/2281#issuecomment-1471805422
> 
> _Originally posted by @noklam in https://github.com/kedro-org/kedro/discussions/3755#discussioncomment-9009731_

# Context
Since 0.19, we introduced a few changes:
- #2637 
- #2535 
- A new kedro project creation flow, i.e. `kedro new --tools` (with logging)

This create a bad UX because:
-  New users expect `logging.yml` will be used by default
-  Existing users migrate from 0.18 -> 0.19 things start working unexpectedly

The scope of ticket is to implement a fix to make sure the default file is read. `KEDRO_LOGGING_CONFIG` will take priority when it is provided. (See https://github.com/kedro-org/kedro/commit/be3d28d981ecf51d8fef96a256d9c0ef74d5d8c4 for inspiration)

# Solution
https://github.com/kedro-org/kedro/commit/be3d28d981ecf51d8fef96a256d9c0ef74d5d8c4 is a good starting point. At the minimal case, the happy path of
- `kedro new --tools=log`
- `kedro run` should auotmatically read `logging.yml`

There are things to consider:
- Should we consider accepting `logging*` or searching for subfolder? 
[...]

PR #3831, opened 2024-04-23
Make logging.yml read by default

SajidAlamQB:
## Description

Related to: https://github.com/kedro-org/kedro/issues/3801

Implement a fix to make sure the default file is read. KEDRO_LOGGING_CONFIG will take priority when it is provided.

## Development notes

- It will default to `conf/logging.yml` if `KEDRO_LOGGING_CONFIG` isn't set. 
- If `conf/logging.yml` does not exist, it falls back to `default_logging.yml`. 

### CONF_SOURCE

Investigated the possibility of adding custom `CONF_SOURCE` support during the logging setup, allowing the `logging.yml` to be dynamically set. Below are the results of this investigation:

#### Code changes:

```
class _ProjectLogging(UserDict):
    def __init__(self, settings: _ProjectSettings) -> None:
        """Initialise project logging. The path to logging configuration is given in
        environment variable KEDRO_LOGGING_CONFIG (defaults to conf/logging.yml)."""
        # Fetch CONF_SOURCE from settings
        conf_source = getattr(settings, 'CONF_SOURCE', 'conf')

        # Check if a user path is set in the environment variable
        user_logging_path = os.environ.get("KEDRO_LOGGING_CONFIG")

        # Check if the default logging configuration exists
        default_logging_path = Path(conf_source) / "logging.yml"
        if not default_logging_path.exists():
            default_logging_path = Path(__file__).parent / "default_logging.yml"

        # Use the user path if available, otherwise, use the default path
        if user_logging_path and Path(user_logging_path).exists():
            path = Path(user_logging_path)
        else:
            path = default_logging_path

        # Load and apply the logging configuration
        logging_config = Path(path).read_text(encoding="utf-8")
        self.configure(yaml.safe_load(logging_config))

PACKAGE_NAME = None

settings = _ProjectSettings()

LOGGING = _ProjectLogging(settings)

pipelines = _ProjectPipelines()
```

#### Explanation:
The above code attempted to fetch the `CONF_SOURCE` setting from the Kedro settings and use this setting to find the path of the logging config file, allowing the logging setup to change automatically depending on where the user specifies.

#### Issues:
During the implementation I ran into several problems mostly due to timings:

The dynamic fetching of settings led to circular imports because the logging module depended on settings, which in turn depended on other things that required logging. This dependency cycle caused runtime errors preventing the project from starting.

This approach also complicated the creation of the project. It was quite challengin to initialise logging before all configurations are loaded without causing import loops.

So for now the current approach of using the happy path for logging will only remain. For the users that need more dynamic logging they will have to manually configure logging. I've enhanced the documentation to outline this.

## Checklist

### DIFF
--- a/kedro/framework/project/__init__.py
+++ b/kedro/framework/project/__init__.py
@@ -237,7 +237,9 @@ class _ProjectLogging(UserDict):
         self.configure(yaml.safe_load(logging_config))
 
         # Log info about the logging configuration
-        if not user_logging_path and path.resolve() == (Path("conf/logging.yml").resolve()):
+        if not user_logging_path and path.resolve() == (
+            Path("conf/logging.yml").resolve()
+        ):
             logger = logging.getLogger(__name__)
             logger.info(
                 "Using `conf/logging.yml` as logging configuration. "
--- a/kedro/framework/project/__init__.py
+++ b/kedro/framework/project/__init__.py
@@ -231,6 +231,11 @@ class _ProjectLogging(UserDict):
             path = Path(user_logging_path)
         else:
             path = default_logging_path
+            logger = logging.getLogger(__name__)
+            logger.warning(
+                "Using `conf/logging.yml` as logging configuration. You can change this by setting the "
+                "KEDRO_LOGGING_CONFIG environment variable accordingly."
+            )
 
         # Load and apply the logging configuration
         logging_config = Path(path).read_text(encoding="utf-8")
--- a/kedro/framework/project/__init__.py
+++ b/kedro/framework/project/__init__.py
@@ -217,7 +217,8 @@ class _ProjectLogging(UserDict):
     def __init__(self) -> None:
         """Initialize project logging. The path to logging configuration is given in
         environment variable KEDRO_LOGGING_CONFIG (defaults to conf/logging.yml)."""
-        # Check if a user-specified path is set in the environment variable
+
+        # Check if a user path is set in the environment variable
         user_logging_path = os.environ.get("KEDRO_LOGGING_CONFIG")
 
         # Check if the default logging configuration exists
@@ -225,7 +226,7 @@ class _ProjectLogging(UserDict):
         if not default_logging_path.exists():
             default_logging_path = Path(__file__).parent / "default_logging.yml"
 
-        # Use the user-specified path if available; otherwise, use the default path
+        # Use the user path if available, otherwise, use the default path
         if user_logging_path and Path(user_logging_path).exists():
             path = Path(user_logging_path)
         else:
--- a/kedro/framework/project/__init__.py
+++ b/kedro/framework/project/__init__.py
@@ -215,19 +215,26 @@ class _ProjectPipelines(MutableMapping):
 
 class _ProjectLogging(UserDict):
     def __init__(self) -> None:
-        """Initialise project logging. The path to logging configuration is given in
-        environment variable KEDRO_LOGGING_CONFIG (defaults to default_logging.yml)."""
-        path = os.environ.get(
-            "KEDRO_LOGGING_CONFIG", Path(__file__).parent / "default_logging.yml"
-        )
+        """Initialize project logging. The path to logging configuration is given in
+        environment variable KEDRO_LOGGING_CONFIG (defaults to conf/logging.yml)."""
+        # Check if a user-specified path is set in the environment variable
+        user_logging_path = os.environ.get("KEDRO_LOGGING_CONFIG")
+
+        # Check if the default logging configuration exists
+        default_logging_path = Path("conf/logging.yml")
+        if not default_logging_path.exists():
+            default_logging_path = Path(__file__).parent / "default_logging.yml"
+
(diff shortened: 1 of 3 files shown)


kedro/framework/project/__init__.py
```diff
@@ -216,10 +216,23 @@ def configure(self, pipelines_module: str | None = None) -> None:
...
-        path = os.environ.get(
-            "KEDRO_LOGGING_CONFIG", Path(__file__).parent / "default_logging.yml"
-        )
+        environment variable KEDRO_LOGGING_CONFIG (defaults to conf/logging.yml)."""
+
+        # Check if a user path is set in the environment variable
+        user_logging_path = os.environ.get("KEDRO_LOGGING_CONFIG")
+
+        # Check if the default logging configuration exists
+        default_logging_path = Path("conf/logging.yml")
```
noklam, line 225: Have you considered the CONF_SOURCE or is it only possible to support the happy path?

If this is the case, we should at least update the documentation to reflect that.

SajidAlamQB: I will try adding support for custom conf_source.

SajidAlamQB: I've been trying out dynamically supporting `CONF_SOURCE` for logging but ran into some challenges. The main issues are related to the timing of loading configuration. So, as you suggested I will update our documentation to reflect this.

astrojuanlu (commented):
Tested this locally and it works.

Wondering if we could add some logging message like

```
WARNING Using `conf/logging.yml` as logging configuration, you can change it by setting the KEDRO_LOGGING_CONFIG variable accordingly
```

to minimise surprises and clarify what's happening. What do you folks think?

kedro/framework/project/__init__.py
```diff
@@ -216,10 +216,28 @@ def configure(self, pipelines_module: str | None = None) -> None:
...
+        # Use the user path if available, otherwise, use the default path
+        if user_logging_path and Path(user_logging_path).exists():
+            path = Path(user_logging_path)
+        else:
+            path = default_logging_path
+            logger = logging.getLogger(__name__)
+            logger.warning(
+                "Using `conf/logging.yml` as logging configuration. You can change this by setting the "
+                "KEDRO_LOGGING_CONFIG environment variable accordingly."
+            )
```
astrojuanlu, line 238: I'm on the fence on whether this should be `logger.warning` or `warnings.warn`... From https://docs.python.org/3/howto/logging.html#when-to-use-logging,

> [warnings.warn()](https://docs.python.org/3/library/warnings.html#warnings.warn) in library code if the issue is avoidable and the client application should be modified to eliminate the warning
> 
> A logger’s [warning()](https://docs.python.org/3/library/logging.html#logging.Logger.warning) method if there is nothing the client application can do about the situation, but the event should still be noted

astrojuanlu (commented):
Works:

```
❯ kedro run
Using `conf/logging.yml` as logging configuration. You can change this by setting the KEDRO_LOGGING_CONFIG environment variable accordingly.
```

I'll let others comment on the implementation. Leaving just one comment.

kedro/framework/project/__init__.py
SajidAlamQB: I see, then I think then it makes sense to use `warnings.warn` because it lets the users know they can make modifications to avoid this.

ankatiyar: I don't think this should be a warning at all - could be an info level log. If a user is not bothered about the logging stuff, it might be annoying if this keeps popping up with every `kedro run`. There's nothing technically wrong with just using the default logging settings. 

Additionally, since this warning is emitted before the logging settings are configured, it looks kinda out of place. It should be shown after the `self.configure(..)` step. WDYT?

Screenshot 2024-05-07 at 11 14 00

astrojuanlu: > If a user is not bothered about the logging stuff, it might be annoying if this keeps popping up with every kedro run. There's nothing technically wrong with just using the default logging settings.

Good point @ankatiyar, agreed 👍🏼

ankatiyar (approved):
Manual tested, works well! 👍🏾

SajidAlamQB:
Not sure why the windows test are getting a permission error. Anyone have any ideas?

kedro/framework/project/__init__.py
```diff
@@ -216,13 +216,36 @@ def configure(self, pipelines_module: str | None = None) -> None:
...
+        else:
+            path = default_logging_path
+
+        # Load and apply the logging configuration
         logging_config = Path(path).read_text(encoding="utf-8")
         self.configure(yaml.safe_load(logging_config))
 
+        # Log info about the logging configuration
+        if not user_logging_path and path.resolve() == (
+            Path("conf/logging.yml").resolve()
```
ankatiyar, line 241: @SajidAlamQB I suspect the windows test is failing because on windows this path is not being resolved properly, it happens on GHA. try absolute path - `(project_path / Path("conf/logging.yml")).resolve()`

noklam (approved):

PR #3883, opened 2024-05-21
Lazy load click subcommands

### LINKED ISSUE #1476: [spike] Improve Kedro CLI startup time

ankatiyar:
## Description

Fix #1476 

## Development notes

Lazy Loading of subcommands docs on `click`: https://click.palletsprojects.com/en/8.1.x/complex/#lazily-loading-subcommands

- Added two lazy groups: `kedro/framework/cli/cli.py`
    - `project_commands`: Point to project sub commands such as `registry`, `catalog`, `package` etc
    - `global_commands`: Point to the implementation of `new` and `starter` commands
- Changed the order of command groups for `project_groups` and `global_groups` properties of `KedroCLI` -> This allows the commands from the plugins to be found first (incase a plugin has overwritten one of the Kedro core commands)
- Implementation of `LazyGroup` in `kedro/framework/cli/utils.py`
- For our custom `CommandCollection` class in `kedro/framework/cli/utils.py`, removed `_dedupe_commands()` method since the ordering of the command groups means that the commands from plugins are found first. **

** This means that the help text looks a bit different.

Before: 
image
After:
Screenshot 2024-07-08 at 16 47 36

KEY DIFFERENCES:
- Kedro project commands go to the bottom, this is because of the re-ordering of the command groups
- Command from `kedro-override` which is a dummy plugin I have installed which overwrites the `kedro catalog` command groups shows separately in both cases. But in the first one, since de-duplication takes place, `catalog` help text is removed from under the "Project specific commands from Kedro" section but after this change it remains.

❓ For reviewers: ^ Is the change in the help text acceptable?

## Checklist


merelcht:
> Command from kedro-override which is a dummy plugin I have installed which overwrites the kedro catalog command groups shows separately in both cases. But in the first one, since de-duplication takes place, catalog help text is removed from under the "Project specific commands from Kedro" section but after this change it remains.

Is this happening because the de-duplication process would mean you can't do lazy loading? IMO it's not ideal, but it's also not the worst. I think it's becoming clear that performance gains on the CLI loading will always have some downsides. We can always add some clarification in the docs on overwriting of commands and what a user should expect to see on the CLI side.

tests/framework/cli/test_cli.py
```diff
@@ -333,15 +332,8 @@ def test_project_commands_no_clipy(self, mocker, fake_metadata):
...
-        assert len(kedro_cli.project_groups) == 6
-        assert kedro_cli.project_groups == [
-            catalog_cli,
-            jupyter_cli,
-            pipeline_cli,
-            micropkg_cli,
-            project_group,
-            registry_cli,
-        ]
+        assert len(kedro_cli.project_groups) == 1
```
merelcht, line 335: It might be helpful to update the test name or leave a comment explaining that the 1 command group loaded is everything that's not doing lazy loading.

ankatiyar:
> Is this happening because the de-duplication process would mean you can't do lazy loading? IMO it's not ideal, but it's also not the worst. I think it's becoming clear that performance gains on the CLI loading will always have some downsides. We can always add some clarification in the docs on overwriting of commands and what a user should expect to see on the CLI side.
@merelcht 

Yeah, for a `LazyGroup`, the `group.commands` is an empty `{}` so it'll have to be loaded for the de-duplication. It's also going to be tricky if we advocate the plugins to define their own commands as a `LazyGroup`. The order of the sources of commands will already determine which commands get discovered first, and if commands from plugins are prioritised, this should happen on its own.

We could add information to this part of the help text (up top) saying if there's duplicate commands, the commands from the plugins will be used : 
Screenshot 2024-07-09 at 12 40 58


* To be fair, making our own commands lazy doesn't have too much of a performance gain (atleast on my system). It would be more beneficial if the plugins implemented it, starting with kedro-viz. 
* However, I think if the plugins implemented their own `LazyGroup` commands **and** had overwriting Kedro commands in there, probably the de-duplication code we have now won't catch it anyway.

merelcht:
> We could add information to this part of the help text (up top) saying if there's duplicate commands, the commands from the plugins will be used : Screenshot 2024-07-09 at 12 40 58

That's a great idea!

> * To be fair, making our own commands lazy doesn't have too much of a performance gain (atleast on my system). It would be more beneficial if the plugins implemented it, starting with kedro-viz.
> * However, I think if the plugins implemented their own `LazyGroup` commands **and** had overwriting Kedro commands in there, probably the de-duplication code we have now won't catch it anyway.

Got it! In that case, this solution is all fine with me 👍

tests/framework/cli/test_cli.py
```diff
@@ -371,22 +363,23 @@ def test_project_commands_valid_clipy(self, mocker, fake_metadata):
...
-            pipeline_cli,
-            micropkg_cli,
-            project_group,
-            registry_cli,
+            # catalog_cli,
+            # jupyter_cli,
+            # pipeline_cli,
+            # micropkg_cli,
+            # project_group,
+            # registry_cli,
```
merelcht, line 373: Do you want to keep these comments or is that still a left-over clean up?

tests/framework/cli/test_cli_hooks.py
```diff
@@ -97,7 +97,7 @@ def test_kedro_cli_should_invoke_cli_hooks_from_plugin(
     ):
         caplog.set_level(logging.DEBUG, logger="kedro")
 
-        Module = namedtuple("Module", ["cli"])
+        # Module = namedtuple("Module", ["cli"])
```
merelcht, line 100: Same here re: comments and the mocker part below.

merelcht (commented):
One question regarding testing: is it possible to trigger the lazy loading in the tests to make sure the lazy loaded commands to get loaded eventually?

tests/framework/cli/test_cli.py
ankatiyar: cleanup, removed!

tests/framework/cli/test_cli_hooks.py
ankatiyar: removed!

ankatiyar:
> One question regarding testing: is it possible to trigger the lazy loading in the tests to make sure the lazy loaded commands to get loaded eventually?

I added a check if all the commands are being listed properly for the lazy group in 37764a7a25698835341e1d9d64ee5a8f73971ec0

docs/source/kedro_project_setup/session.md
```diff
@@ -67,4 +67,8 @@ This function reads `settings.py` and `pipeline_registry.py` and registers the c
...
 
-If you are using `multiprocessing`, you need to be careful about this. Depending on your Operating System, you may have [different default](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods). If the processes are `spawn`, Python will re-import all the modules in each process and thus you need to run `configure_project` again at the start of the new process. For example, this is how Kedro handle this in `ParallelRunner`(https://github.com/kedro-org/kedro/blob/9e883e6a0ba40e3db4497b234dcb3801258e8396/kedro/runner/parallel_runner.py#L84-L85)
+If you are using `multiprocessing`, you need to be careful about this. Depending on your Operating System, you may have [different default](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods). If the processes are `spawn`, Python will re-import all the modules in each process and thus you need to run `configure_project` again at the start of the new process. For example, this is how Kedro handles this in `ParallelRunner`:
+```python
+if multiprocessing.get_start_method() == "spawn" and package_name:
+        _bootstrap_subprocess(package_name, logging_config)
+```
```
ankatiyar, line 74: The secret scan test was complaining about the link ^ I added the code snippet instead. I hope that's okay? @noklam

merelcht (approved):
Don't forget to add this to the release notes. Otherwise looks all good to me 👍 🎉

ElenaKhaustova (approved):
Looks great 🚀

docs/source/kedro_project_setup/session.md
ElenaKhaustova: I had this problem locally as well but it's all good now and the link seems correct and working.

PR #3896, opened 2024-05-28
Refactor the logging setup loggic

noklam:
## Description

I was pairing with @jitu5 and found that the logic of logging become quite complicated. @lrcouto @SajidAlamQB because you have made changes to this file recently. There is an opportunity to remove duplicate checking and overload variable `default_logging_config`.

## Development notes

- logger.info will always emit logs for which logging file is read
- simplify the logic

## Checklist


lrcouto:
I like the changes. The messages about the logging itself are rightly informative.

astrojuanlu:
> `logger.info` will always emit logs for which logging file is read

👍🏼

SajidAlamQB (approved):
This makes sense, thank you @noklam!

kedro/framework/project/__init__.py
```diff
@@ -217,40 +217,32 @@ class _ProjectLogging(UserDict):
...
+        logger = logging.getLogger(__name__)
         user_logging_path = os.environ.get("KEDRO_LOGGING_CONFIG")
+        project_logging_path = Path("conf/logging.yml")
+        default_logging_path = Path(
+            Path(__file__).parent / "rich_logging.yml"
+            if importlib.util.find_spec("rich")
+            else Path(__file__).parent / "default_logging.yml",
+        )
+        msg = ""
+        if user_logging_path:
```
merelcht, line 229: Do we still need to check whether this path exists? Previously we did: 

```
if user_logging_path and Path(user_logging_path).exists():
```

noklam: Good question. Why should we explicitly check for the user? If they specify a path that is invalid, it's seem magic to be to fallback to default logging automatically.

noklam:
Need a bit more thinking here:
image

It's a nice idea but this is a chicken and egg problem. The log would not show up until we set the logging level. By default python show WARNINGs only so `logger.info` did log something but there is no easy way to actually see the log.

noklam:
image
Fix the annotation from ``` to `'` (for rich), delay the log after self.configure is run.

kedro/framework/project/__init__.py
noklam: I still feel it's a bit weird we handle this error, is this fallback mechanism necessary? If one put in a value there with the intend to use their own config, it's more intuitive to give an error when the file does not exist.

@merelcht What do you think? Should I revert this to the fallback logic?

SajidAlamQB: I think its still beneficial for users who might not know they need to provide a custom configuration or for those who accidentally gave the wrong path.

noklam: >who might not know they need to provide a custom configuration

I think this make sense and it's the primary reason we introduce the logic. This PR doesn't change this logic.

>for those who accidentally gave the wrong path.

This is different because for these users they know their intend and they are using `KEDRO_LOGGING_CONFIG` explicitly. I don't feel this is more special than other things (i.e. if `CONF_SOURCE` is wrong we wouldn't fallback to `base`, if `KEDRO_ENV` doesn't exist we wouldn't fallback to `base`), what make `KEDRO_LOGGING_CONFIG` more special?

The only place I can remember is `find_pipeline`, and that has a very specific workflow in mind. And we do add back the `find_pipeline(raise_error=True)` option as some people find the magic behavior confusing, particularly for production environment.

merelcht (approved):
Thanks @noklam ⭐

kedro/framework/project/__init__.py
noklam: Thanks! I will keep the PR as it

PR #3900, opened 2024-05-29
Add fallback to starters pull on kedro new

lrcouto:
## Description

During our current project creation and test setup, Kedro looks up to the `kedro-starters` repo and always uses the latest released version to get project templates. This can cause a problem during releases that depends on changes on the `kedro-starters` repo, as they won't be acknowledged by the current flow until after they are released.

This PR implements a fallback for when this situation happens. When the version of Kedro installed on your environment does not match the latest `kedro-starters`release, it will pull the `main` branch instead.

## Checklist


lrcouto:
A couple notes:

- I've used the [requests_mock](https://pypi.org/project/requests-mock/) library to run tests for this PR because I  was having a lot of trouble mocking requests to the Github API with fixtures we already had.
- In agreement with [this response](https://kedro-org.slack.com/archives/C0694PW10NS/p1717505604753899?thread_ts=1717470729.998539&cid=C0694PW10NS) on Slack, I've preferred to not touch the `kedro checkout` function, since it is user-facing and this fallback feature is just for us.

kedro/framework/cli/starters.py
```diff
@@ -773,7 +812,11 @@ def _make_cookiecutter_args_and_fetch_template(
 
     tools = config["tools"]
     example_pipeline = config["example_pipeline"]
-    starter_path = "git+https://github.com/kedro-org/kedro-starters.git"
+    starter_path = (
+        "git+https://github.com/kedro-org/kedro-starters.git"
+        if _kedro_and_starters_version_identical()
+        else "https://github.com/kedro-org/kedro-starters.git@main"
+    )
```
merelcht, line 819: Can't this just be the global `_STARTERS_REPO`?

kedro/framework/cli/starters.py
```diff
@@ -95,7 +96,45 @@ class KedroStarterSpec:
...
+                "https://api.github.com/repos/kedro-org/kedro-starters/releases/latest",
+                headers=headers,
+                timeout=10,
+            )
+            response.raise_for_status()  # Raise an HTTPError for bad status codes
+            latest_release = response.json()
+        except requests.exceptions.RequestException as e:
+            raise RuntimeError(
+                f"Error fetching kedro-starters latest release version: {e}"
+            )
```
merelcht, line 118: This piece of code is now also called in a regular `kedro new` flow outside of the whole release process, so instead of raising an exception here, I think it would be better to only log the error and just return an empty string. Otherwise users wouldn't be able to do `kedro new` when something is wrong with the GH API.

merelcht (commented):
The approach looks good to me 👍  I left some small suggestions.

kedro/framework/cli/starters.py
lrcouto: That's a good point

kedro/framework/cli/starters.py
```diff
@@ -95,7 +97,44 @@ class KedroStarterSpec:
 KEDRO_PATH = Path(kedro.__file__).parent
 TEMPLATE_PATH = KEDRO_PATH / "templates" / "project"
 
-_STARTERS_REPO = "git+https://github.com/kedro-org/kedro-starters.git"
+
+def _get_latest_starters_version() -> str:
+    if "STARTERS_VERSION" not in os.environ:
```
astrojuanlu, line 102:
```suggestion
    if "KEDRO_STARTERS_VERSION" not in os.environ:
```

? (and everywhere else)

kedro/framework/cli/starters.py
```diff
@@ -773,7 +812,7 @@ def _make_cookiecutter_args_and_fetch_template(
 
     tools = config["tools"]
     example_pipeline = config["example_pipeline"]
-    starter_path = "git+https://github.com/kedro-org/kedro-starters.git"
+    starter_path = _STARTERS_REPO
```
ElenaKhaustova, line 815: Cannot comment on the proper line as it was not changed but the suggestion is what if we do not introduce the `starter_path` variable as it complicates the logic? So instead we can do:
```
    else:
        # Use the default template path for non PySpark, Viz or example options:
        cookiecutter_args, template_path
    return cookiecutter_args, _STARTERS_REPO
```

kedro/framework/cli/starters.py
```diff
@@ -95,7 +97,44 @@ class KedroStarterSpec:
...
+
+def _kedro_and_starters_version_identical() -> bool:
+    starters_version = _get_latest_starters_version()
+    return True if version == starters_version else False
+
+
+_STARTERS_REPO = (
+    "git+https://github.com/kedro-org/kedro-starters.git"
+    if _kedro_and_starters_version_identical()
+    else "https://github.com/kedro-org/kedro-starters.git@main"
```
ElenaKhaustova, line 134: What happens if we point to the main branch (`_STARTERS_REPO="https://github.com/kedro-org/kedro-starters.git@main"`) but then below in `_make_cookiecutter_args_and_fetch_template` we set the checkout version as well? For example in this case:
```
if "PySpark" in tools and "Kedro Viz" in tools:
        # Use the spaceflights-pyspark-viz starter if both PySpark and Kedro Viz are chosen.
        cookiecutter_args["directory"] = "spaceflights-pyspark-viz"
        # Ensures we use the same tag version of kedro for kedro-starters
        cookiecutter_args["checkout"] = version
```
Will it then work as we expect I mean creating a project and passing the branch and the checkout version to the cookiecutter?
https://github.com/kedro-org/kedro/blob/15ba4dae69231b5f92921ddab879f4df43233e61/kedro/framework/cli/starters.py#L932

ElenaKhaustova (commented):
Thank you, @lrcouto, like the solution!

Left one question and suggestion.

kedro/framework/cli/starters.py
lrcouto: That's a good point. To be honest, I'm not sure why this variable was there in the first place.

kedro/framework/cli/starters.py
lrcouto: This will lead us to cookiecutter shenanigans, which is where these parameters are ultimately used. You have to dig down four or five functions to see it, but cookiecutter is using the template parameter to determine the repo address and the checkout parameter to determine the branch, tag or commit ID. What it does is the following:

```py
    if clone:
        try:
            subprocess.check_output(  # nosec
                [repo_type, 'clone', repo_url],
                cwd=clone_to_dir,
                stderr=subprocess.STDOUT,
            )
            if checkout is not None:
                checkout_params = [checkout]
                # Avoid Mercurial "--config" and "--debugger" injection vulnerability
                if repo_type == "hg":
                    checkout_params.insert(0, "--")
                subprocess.check_output(  # nosec
                    [repo_type, 'checkout', *checkout_params],
                    cwd=repo_dir,
                    stderr=subprocess.STDOUT,
                )
```

So if you passed the repo_url as the address to the main branch and, for example, passed the checkout value as a different branch, cookiecutter would clone the main branch and then checkout to the other branch.

kedro/framework/cli/starters.py
ElenaKhaustova: Yes, I've seen it changed but I still prefer to return the original one instead. See the edited suggestion (I've missed a "return" and looked confusing)

kedro/framework/cli/starters.py
ElenaKhaustova: Thank you for explaining, that's what I thought as well. Doesn't it mean that in this case instead of `main` branch (which we aim for) we again will have the released branch (as the checkout argument is the installed kedro version in case it's not passed https://github.com/kedro-org/kedro/blob/15ba4dae69231b5f92921ddab879f4df43233e61/kedro/framework/cli/starters.py#L344) which is not desired?

lrcouto: That's a good point, let me test that.

ElenaKhaustova: I was thinking of refactoring the logic in `_make_cookiecutter_args_and_fetch_template` because now we have two places where we define the template which is a bit confusing. But that can be done in a separate PR.

Now, we can just identify the case when repos versions are not matched (as you do for setting the `_STARTERS_REPO`). In this case, we only need to use the template path provided but skip setting up the checkout parameter in `_make_cookiecutter_args_and_fetch_template`.

ElenaKhaustova: Or we can just do something like 
```
checkout = checkout or version if _kedro_and_starters_version_identical() else None
```

https://github.com/kedro-org/kedro/blob/15ba4dae69231b5f92921ddab879f4df43233e61/kedro/framework/cli/starters.py#L399

kedro/framework/cli/starters.py
```diff
@@ -766,6 +805,8 @@ def _make_cookiecutter_args_and_fetch_template(
         "extra_context": config,
     }
 
+    kedro_version_match_starters = _kedro_and_starters_version_identical()
+
     if checkout:
         cookiecutter_args["checkout"] = checkout
```
merelcht, line 811: Is this still needed? I looks like this will be overwritten below in all branches.

kedro/framework/cli/starters.py
```diff
@@ -779,21 +820,29 @@ def _make_cookiecutter_args_and_fetch_template(
...
     elif "PySpark" in tools:
         # Use the spaceflights-pyspark starter if only PySpark is chosen.
         cookiecutter_args["directory"] = "spaceflights-pyspark"
-        cookiecutter_args["checkout"] = version
+        cookiecutter_args["checkout"] = (
+            version if kedro_version_match_starters else "main"
+        )
     elif "Kedro Viz" in tools:
         # Use the spaceflights-pandas-viz starter if only Kedro Viz is chosen.
         cookiecutter_args["directory"] = "spaceflights-pandas-viz"
```
merelcht, line 834: Shouldn't `checkout` be set here as well?

kedro/framework/cli/starters.py
```diff
@@ -766,35 +805,39 @@ def _make_cookiecutter_args_and_fetch_template(
...
     elif example_pipeline == "True":
         # Use spaceflights-pandas starter if example was selected, but PySpark or Viz wasn't
         cookiecutter_args["directory"] = "spaceflights-pandas"
-        cookiecutter_args["checkout"] = version
+        cookiecutter_args["checkout"] = checkout_version
     else:
         # Use the default template path for non PySpark, Viz or example options:
         starter_path = template_path
-
+        cookiecutter_args["checkout"] = (
```
ElenaKhaustova, line 838: Related to the previous comment.

If the `checkout` argument is provided and kedro version doesn't match starters, do we indeed want to sick to `main`?

In this case, there's also a discrepancy with `_get_cookiecutter_dir` method, where we use the provided `checkout` argument.

DimedS (commented):
Thank you for the PR, @lrcouto!

I have a few questions regarding it:

1. I'm trying to understand the algorithm in this PR. Could you please help with that?

In my opinion, we need to check the following:

If `current_kedro_version` is not in the `kedro_starters_tags_list`, we should not use checkout in the starters repo (since that tag won't exist). Instead, we should use the default starters repo.
Otherwise, if `current_kedro_version` is in the list, we should set `kedro_starters_tag` to `current_kedro_version`.

From what I see in the current PR, we are checking `current_kedro_version` against `latest_kedro_starters_tag`. If they don't match, we use the default repo. It seems that if a user is using an older version of Kedro, it won't match with `latest_kedro_starters_tag`, and the user will receive the latest default repo. However, they should receive the version of starters that matches `current_kedro_version` (I believe this is a current logic that we don't want to change).

2. Why are we using `os.environ["KEDRO_STARTERS_VERSION"]` to store the latest starters version? Why isn't a global variable sufficient?

lrcouto:
Thank you for your reviews, @ElenaKhaustova and @DimedS ! From what I'm getting, the logic of the version/branch/tag selection for starters itself is a little confusing, so I'm gonna look into refactoring it a little bit.

As for the question regarding the environment variable, it came from [this](https://kedro-org.slack.com/archives/C0694PW10NS/p1717470729998539) discussion. We were having problems with repeated requests to Github being rejected, which happened every time tests were ran on out CI. My idea is that setting it on the environment once and then checking it would avoid this large number of requests, and it would not need to be re-declared every time the file is loaded.

lrcouto:
> If `current_kedro_version` is not in the `kedro_starters_tags_list`, we should not use checkout in the starters repo (since that tag won't exist). Instead, we should use the default starters repo. Otherwise, if `current_kedro_version` is in the list, we should set `kedro_starters_tag` to `current_kedro_version`.

Regarding this point, would we have to get all of the existing Kedro versions from Git? Or does that exist already somewhere? I'm asking because it'd be another request that we'd have to make.

DimedS:
> > If `current_kedro_version` is not in the `kedro_starters_tags_list`, we should not use checkout in the starters repo (since that tag won't exist). Instead, we should use the default starters repo. Otherwise, if `current_kedro_version` is in the list, we should set `kedro_starters_tag` to `current_kedro_version`.
> 
> Regarding this point, would we have to get all of the existing Kedro versions from Git? Or does that exist already somewhere? I'm asking because it'd be another request that we'd have to make.

I believe we can use the `latest_kedro_starters_tag` that you already received and compare it with the `current_kedro_version`. If the `current_kedro_version` is greater than the `latest_kedro_starters_tag` (indicating that Kedro has been released with a new version, but the starters have not yet been updated), then we should use the default starters repository (i.e., the latest starters from the main branch). Otherwise, we should maintain the current logic and use the same version of starters that matches the Kedro version.

kedro/framework/cli/starters.py
```diff
@@ -766,35 +806,44 @@ def _make_cookiecutter_args_and_fetch_template(
...
```
merelcht, line 844: Shouldn't this just be `cookiecutter_args["checkout"] = checkout_version` as well?

merelcht (commented):
I left one more question, but otherwise the approach with checking if the version is smaller or equal to the starters version looks good!

DimedS (commented):
Thank you for changing the logic, @lrcouto. It looks mostly good now! I've left a few additional questions.

kedro/framework/cli/starters.py
```diff
@@ -766,35 +806,44 @@ def _make_cookiecutter_args_and_fetch_template(
...
+    kedro_version_match_starters = _kedro_version_equal_or_lower_to_starters(version)
+
     if directory:
         cookiecutter_args["directory"] = directory
 
     tools = config["tools"]
     example_pipeline = config["example_pipeline"]
     starter_path = "git+https://github.com/kedro-org/kedro-starters.git"
 
+    if checkout:
```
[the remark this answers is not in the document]
ElenaKhaustova: Should this logic be global within `new` command?

lrcouto:
Pushed some changes based on the conversation earlier today with @ElenaKhaustova and @DimedS.
- The logic that selects which checkout argument will be passed to the Cookiecutter function was moved to is own function.
- `_STARTERS_REPO` now points only to the main starters repo.
- Removed redundant logic from `_make_cookiecutter_args_and_fetch_template`.

merelcht (approved):
LGTM 👍

DimedS (approved):
Thank you for the PR and addressing comment, @lrcouto ! Looks nice without that `cookiecutter_args["checkout"] = version` repetition in every `if else` line.

PR #3901, opened 2024-05-30
Lazy load commands from plugins

### LINKED ISSUE #1476: [spike] Improve Kedro CLI startup time

ankatiyar:
## Description
Partly resolve #1476 

As discussed in #1476, the initialisation of `KedroCLI` takes up a significant chunk of time. This is especially evident if you have a lot of kedro plugins installed in the environment as `KedroCLI` which is a `CommandCollection`, loads up all the commands from the plugins before the command is processed. This PR is to add a lazy way to load the commands from the plugins.

## Development notes

- Add a property called `plugin_groups` which returns a `dict` of `entry_point_name` to the `EntryPoint` object.
- This group is passed to the super class `CommandCollection` which is a custom version of `click.CommandCollection` used by Kedro and defined in `kedro/framework/cli/utils.py`

Update to custom `CommandCollection`:
- Add a new `main` function which receives the command to be run as `args`, eg `[ "catalog", "list"]` or `["airflow", "create"]`
- if the first element of the `args` is not in the command, then load commands from plugins
- run the command by passing to `super().main()`

Loading of plugins:

I tried to implement a sort of "smart" loading where if the command arg, i.e. `airflow` or `mlflow` or `viz` partially matches the entry point names in the `lazy_group` dict keys, load the plugin, check if the command now exists and exit.
Note: The entry point names are decided by the plugins, eg. `kedro-airflow`'s ep name is `airflow` which fully matches the command name too but for `kedro-viz` and `kedro-mlflow` the entry point names don't match the commands. eg. Command: `viz` Entry point name: `kedro-viz`

Otherwise, load all plugins one by one, exit if the command exists after loading plugins.

### TODO
- Fix code coverage on unit tests, which is tricky since this code is only used when a plugin is installed but we don't have tests like these in our unit tests. 

## Checklist

### DIFF
--- a/kedro/framework/cli/cli.py
+++ b/kedro/framework/cli/cli.py
@@ -105,7 +105,7 @@ class KedroCLI(CommandCollection):
         super().__init__(
             ("Global commands", self.global_groups),
             ("Project specific commands", self.project_groups),
-            plugin_groups=self.plugin_groups,
+            plugin_entry_points=self.plugin_groups,
         )
 
     def main(
--- a/kedro/framework/cli/cli.py
+++ b/kedro/framework/cli/cli.py
@@ -105,6 +105,7 @@ class KedroCLI(CommandCollection):
         super().__init__(
             ("Global commands", self.global_groups),
             ("Project specific commands", self.project_groups),
+            lazy_groups=list(self.lazy_global_groups) + list(self.lazy_project_groups),
         )
 
     def main(
@@ -173,12 +174,22 @@ class KedroCLI(CommandCollection):
             sys.exit(exc.code)
 
     @property
+    def lazy_global_groups(self):
+        eps = _get_entry_points("global")
+        return eps
+
+    @property
+    def lazy_project_groups(self):
+        eps = _get_entry_points("project")
+        return eps
+
+    @property
     def global_groups(self) -> Sequence[click.MultiCommand]:
         """Property which loads all global command groups from plugins and
         combines them with the built-in ones (eventually overriding the
         built-in ones if they are redefined by plugins).
         """
-        return [cli, create_cli, *load_entry_points("global")]
+        return [cli, create_cli]
 
     @property
     def project_groups(self) -> Sequence[click.MultiCommand]:
@@ -201,7 +212,7 @@ class KedroCLI(CommandCollection):
             registry_cli,
         ]
 
-        plugins = load_entry_points("project")
+        # plugins = load_entry_points("project")
 
         try:
             project_cli = importlib.import_module(f"{self._metadata.package_name}.cli")
@@ -209,7 +220,7 @@ class KedroCLI(CommandCollection):
         except ModuleNotFoundError:
             # return only built-in commands and commands from plugins
             # (plugins can override built-in commands)
-            return [*built_in, *plugins]
+            return [*built_in]
 
         # fail badly if cli.py exists, but has no `cli` in it
         if not hasattr(project_cli, "cli"):
@@ -219,7 +230,7 @@ class KedroCLI(CommandCollection):
         user_defined = project_cli.cli
         # return built-in commands, plugin commands and user defined commands
         # (overriding happens as follows built-in < plugins < cli.py)
-        return [*built_in, *plugins, user_defined]
+        return [*built_in, user_defined]
 
 
 def main() -> None:  # pragma: no cover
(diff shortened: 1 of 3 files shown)


astrojuanlu:
To note, as @ankatiyar noted internally, "bad" lookups will take a lot of time: for example, `kedro infoo` (notice typo) took 17 seconds in my computer and tried to import kedro-viz, kedro-mlflow, and more.

Going forward, is there a smarter way we can declare new CLI commands, so that we can look them up in entry points without actually importing anything? (Feel free to open a new issue about this)

ankatiyar:
The code coverage is not complete but opening this up for review anyway to gather feedback on the method.

merelcht (changes requested):
I added some suggestions/questions to the code, but I actually don't think it's a good idea that now the plugin commands aren't loaded when you do `kedro --help`. I've built a small plugin myself following https://docs.kedro.org/en/stable/extend_kedro/plugins.html#example-of-a-simple-plugin and the problem you get here is that the commands belong to the `kedro` group, but don't show at all. In this case there's no other top-level group you can call to see what sub-commands exist. On top of that the overwriting behaviour is changes, so if I now create my own `micropkg` command inside this kedrojson plugin it doesn't overwrite the original `kedro micropkg` where in the non lazy-loading version it does do the overwriting. 

While this solution does lead to performance gain, it also degrades the plugin command discoverability and therefore IMO usability of the plugins as a whole. Features like https://docs.kedro.org/en/stable/extend_kedro/plugins.html#global-and-project-commands will not exist anymore after this change. 

All in all, I don't think the performance gains are worth losing the plugin features.

ankatiyar:
Fair point @merelcht! 🤔 
I was also trying to implement a way where `--help` does trigger the loading of the plugins. 
About overwriting of core Kedro commands eg. `micropkg`, that might also be possible with lazy loading. Let me explore these functionalities and update the PR!

ankatiyar:
Unfortunately, I don't think the overriding behaviour of CLI is possible with lazy loading of plugins - i.e. if a plugin has overriding CLI commands for Kedro core commands such as `kedro catalog list` etc since the premise of the proposal is to only load plugins if a command does not exist in KedroCLI.
I think it's possible to trigger the loading of plugins and display plugin commands with `kedro --help` but I would like to get the team's opinions on if we should be going ahead with this at all 
cc @astrojuanlu @noklam @merelcht

astrojuanlu:
The `python` CLI has `--help` and `--help-all`. Maybe for the next breaking release we could make `--help` display only the "core" commands, and `--help-all` display everything.

For retaining the current behavior while attaining much faster load times for all the rest of commands, I'd be 👍🏼 on force-triggering loading of all plugins in `kedro --help`

ankatiyar:
Moving this back to the drafts, will close it if no other comments come in. I'll focus on getting https://github.com/kedro-org/kedro-viz/pull/1920 and https://github.com/kedro-org/kedro/pull/3883 ready for review first!

ankatiyar:
Closing this in favour of the other solution!

Issue #3233, opened 2023-10-26
The pipeline registry is difficult to understand


astrojuanlu:
The current code for `pipeline_registry.py` in the default template is as follows:

https://github.com/kedro-org/kedro/blob/df9f174864640de193b2b85f04d0c3e8aee7d22c/kedro/templates/project/%7B%7B%20cookiecutter.repo_name%20%7D%7D/src/%7B%7B%20cookiecutter.python_package%20%7D%7D/pipeline_registry.py#L1-L16

Apart from #2526, this is fine and works well. The magic is in `kedro.framework.project.find_pipelines`, which scans different directories searching for a `create_pipeline` function:

https://github.com/kedro-org/kedro/blob/df9f174864640de193b2b85f04d0c3e8aee7d22c/kedro/framework/project/__init__.py#L292

This is so magical though, that the moment users want to manually register pipelines, they go crazy. For example, this is a user that was trying something like `kedro run --pipeline=data_science+evaluation`, which is a _beautiful_ syntax by the way https://linen-slack.kedro.org/t/15697047/i-have-a-quick-question-on-running-selected-pipelines-only-i#b93fe172-d54f-4f51-a8a6-b85f9dbcec32

to which I replied, how would I _subtract_ a pipeline?

```python
def register_pipelines() -> dict[str, Pipeline]:
    """Register the project's pipelines.

    Returns:
        A mapping from pipeline names to ``Pipeline`` objects.
    """
    pipelines = find_pipelines()
    pipelines["__default__"] = sum(pipelines.values())
    pipelines["except-train"] = ???
    return pipelines
```

in the end I did this:

```python
from .pipelines.model_training import create_pipeline as create_model_training_pipeline

...
pipelines["all"] = sum(pipelines.values())
pipelines["all_except_eval"] = pipelines["all"] - create_model_training_pipeline()
```

but @noklam suggested this instead

```
pipelines["all_except_eval"] = pipelines["all"] - pipelines["eval"]
```

This week I saw a user do something similar, but they renamed the functions instead: 

https://github.com/pablovdcf/TFM_HADO_Cares/blob/28d5a024b915169a039a5a84996b9ee11ee1f3ee/hado/src/hado/pipeline_registry.py#L5-L7

and since their pipeline creation functions were not named `create_pipeline` but something else, this completely broke the automagic `find_pipelines` for them.

astrojuanlu:
[...]

Issue #770, opened 2021-05-19
Universal Kedro deployment (Part 1) - Separate external and applicative configuration to make Kedro cloud native


Galileo-Galilei:
# Preamble

Dear Kedro team, 

I've been using Kedro since July 2019 (`kedro==0.14.3`, which is quite different from what kedro is now) and my team has deployed in the past 2 years a few dozens of machine learning pipelines in production with Kedro. I want to give you some feedback on my Kedro experience along this journey, and the advantages and drawbacks of Kedro from my team's point of view with the current versions (0.16.x and 0.17.x):

**Advantages:**
- the level of standardisation across projects has tremendously increased. As a consequence:
   - Maintenance is much easier, because in the worst case, if you open an undocumented project you can easily find out how to launch it and modify it beacuase each object has one single and well identified place in the template
  - Collaboration and reuse are much better because anyone can read/modify any other project even if it is poorly documented
- the team's software engineering skills have improved a lot and data scientists have much better habits: even if they have to work within a a non-kedro project, they are more inclined to spontaneously separate configuration from code execution.

**Drawbacks:** 
- Kedro's is still early stage and the frequent changes (especially in the template) make hard to upgrade between versions.  
- Kedro's deployment is not standard and is still hard to deploy. You have made a bunch of either plugins (kedro-airflow, kedro-docker) or tutorials (argo, prefect, kubeflow...) to explain how to deploy a kedro project. However, these examples suffer from several drawbacks:  
- **Project orchestration:** you assume we will map kedro ``nodes`` to the orchestrator nodes. This is not realistic, and in a discussion with @limdauto [here](https://github.com/Galileo-Galilei/kedro-mlflow/issues/44#issuecomment-772076223) we agreed on the fact that the conversion to the pipeline's nodes is complicated and must be thoroughly thought by the person in charge of the deployment

> In a nutshell, I don't think there is an exact mapping between kedro nodes (designed by a data scientist for code readibility, easy debugging and partial execution...) and orchestrator nodes (designed for system robustness, ease of asynchronous execution, retry strategies, efficient compute...). kedro nodes are much more low-level in my opinion than orchestrator nodes.

> The reason why we don't make all of them into plugins were precisely because of what you say here: how you slice your pipeline and map them to the orchestrator's nodes is up to you. A good pattern is using tag.[... or] you can map them based on namespace if you construct your pipeline out of modular pipelines.	

[...]

Issue #650, opened 2020-12-21
Define pipelines in YAML files


Sitin:
First of all, thank you very much for the great framework. I'd found myself not working in notebooks for weeks since I've switched to Kedro. Now I can debug everything and still have interactive access to my data and models. You've changed the quality of my professional life being and I am deeply grateful for this.

## Description

Now, my proposal. I think that in addition to existing Python based API for pipeline definition it would be nice to have a possibility to describe pipelines in YAML files. Moreover, the more I am thinking about this the more I'm convinced that this should be the default way of pipeline definition.

## Context

I've found that in 95% of cases my `pipeline.py` files are "static". Meaning that I don't need to dynamically define pipeline nodes and their IO. Also (my subjective opinion) most "dynamical" use cases can be handled by config templates and we already have this feature.

By putting pipeline definitions into `conf` we'll make modular pipelines easier to adjust and configure since any pipeline consumer will have a clear picture of used inputs and outputs which is extremely important since we have "global" dataset naming. We can also mix nodes from different pipelines on a consumer level (check the possible implementation below).

Another reason to use YAML-first pipeline definitions is that it could be much easier to integrate pipeline consistency checks into modern IDEs since we just need to parse YAML files (parameters, pipelines and catalog) and check that the corresponding paths to python modules is correct.

## Possible Implementation

We can use the same structure as in the current `pipeline.py` and `hooks.py` created by starters and same agreement as in `catalog.yml`.

### Current API

Let's say I have the following pipeline in `src/my_package/analysis/pipelines.py`:

```python3
from kedro.pipeline import Pipeline, node

from .nodes import select_important_features, visualize

def create_pipeline(**kwargs):
    return Pipeline(
        [
            node(
                select_important_features,
                inputs=dict(
                    dataset='05_model_input/inliers.parquet',
                    cluster_labels='07_model_output/cluster-labels.parquet',
                ),
                outputs='08_reporting/important-features.yml',
            ),
            node(
                visualize,
                inputs=dict(
                    dataset='05_model_input/inliers.parquet',
                    important_features='08_reporting/important-features.yml',
                ),
                outputs=dict(
                    charts='08_reporting/important-features-charts.html',
                    histograms='08_reporting/important-features-histograms.html',
                ),
                tags=['html_reports']
            ),
        ],
        tags=['cluster_analysis'],
    )
```

I also have to register it in `src/my_package/hooks.py`:

```python3
from my_package.pipelines import analysis

class ProjectHooks:
[...]

PR #3904, opened 2024-05-30
Define modular pipelines in config

bpmeek:
## Description
As a Kedro user I have always wanted to be able to define modular pipelines in a config file. I believe that doing so will reduce the likelihood of a user inadvertently impacting a pipeline other than the one intended.

## Development notes
I have adjusted two files:
`comegaconf_config.py` now has `"pipelines": ["pipelines.yml"],` added to its `config_patterns`
`kedro/framework/project/__init__.py` has an additional function `from_config` that can read a config entry and build the entry into a modular pipeline which is then returned.
A portion of `find_pipelines` was abstracted into a helper function `_get_pipeline_obj` for use in both functions.
This was tested using `make test` and manually tested with the default example_pipeline created when creating a new Kedro project.

## Checklist

- [X] Signed off each commit with a [Developer Certificate of Origin (DCO)](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository)

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -16,6 +16,7 @@
 * Added `raise_errors` argument to `find_pipelines`. If `True`, the first pipeline for which autodiscovery fails will cause an error to be raised. The default behaviour is still to raise a warning for each failing pipeline.
 * It is now possible to use Kedro without having `rich` installed.
 * Updated custom logging behavior: `conf/logging.yml` will be used if it exists and `KEDRO_LOGGING_CONFIG` is not set; otherwise, `default_logging.yml` will be used.
+* Can now define modular pipeline in `conf/pipelines.yml`
 
 ## Bug fixes and other changes
 * User defined catch-all dataset factory patterns now override the default pattern provided by the runner.
--- a/kedro/config/omegaconf_config.py
+++ b/kedro/config/omegaconf_config.py
@@ -114,6 +114,7 @@ class OmegaConfigLoader(AbstractConfigLoader):
             "parameters": ["parameters*", "parameters*/**", "**/parameters*"],
             "credentials": ["credentials*", "credentials*/**", "**/credentials*"],
             "globals": ["globals.yml"],
+            "pipelines": ["pipelines.yml"],
         }
         self.config_patterns.update(config_patterns or {})
 
--- a/kedro/framework/project/__init__.py
+++ b/kedro/framework/project/__init__.py
@@ -347,6 +347,27 @@ def _create_pipeline(pipeline_module: types.ModuleType) -> Pipeline | None:
     return obj
 
 
+def _get_pipeline_obj(pipeline_name: str, raise_errors: bool = False) -> Pipeline | None:
+    pipeline_obj = None
+    pipeline_module_name = f"{PACKAGE_NAME}.pipelines.{pipeline_name}"
+    try:
+        pipeline_module = importlib.import_module(pipeline_module_name)
+        pipeline_obj = _create_pipeline(pipeline_module)
+    except Exception as exc:
+        if raise_errors:
+            raise ImportError(
+                f"An error occurred while importing the "
+                f"'{pipeline_module_name}' module."
+            ) from exc
+
+        warnings.warn(
+            IMPORT_ERROR_MESSAGE.format(
+                module=pipeline_module_name, tb_exc=traceback.format_exc()
+            )
+        )
+    return pipeline_obj
+
+
 def find_pipelines(raise_errors: bool = False) -> dict[str, Pipeline]:  # noqa: PLR0912
     """Automatically find modular pipelines having a ``create_pipeline``
     function. By default, projects created using Kedro 0.18.3 and higher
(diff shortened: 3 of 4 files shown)


datajoely:
Hi @bpmeek thank you for this hard work. The topic of what can be declarative and what should be imperative is something that the team have wrestled with for a long long time. 

I _think_ my personal preference is that this should be a plug-in rather than in kedro core, but keen to get other's perspective

merelcht:
Hi @bpmeek, echoing @datajoely in thanking you the work you've put in this PR! I think in the future it might be better to open an issue or discussion first to talk about fundamental changes like this.

As Joel mentioned, this is a topic we've talked about a lot and we even had an internal version of Kedro where you could define pipelines and nodes in yml instead of python. 

With the solution you're proposing I'm wondering why you'd like to define _modular pipelines_ in config, but not regular pipelines. And what about nodes? I'd also like to understand more what the problem is that will be solved by this?

bpmeek:
Really appreciative, @datajoely and @merelcht for the recognition on my work.
Off the bat, you're 100% right @merelcht, and normally I would have gone ahead and opened an issue as I have in the past, but I was in a unique situation where I had the time to do this work so went ahead and did it instead of starting with an issue as I will most likely be using this either way in the future.

@merelcht, with this solution I'm defining modular pipelines, and pipeline chains, because of a recent project I was working on that kept growing and my pipeline registry ended up becoming a huge file that was hard to manage. Had I been able to take a couple of the larger more complex pipelines out and moved them into a config file I think it would've been much easier to work with.

That being said, I actually _do_ really like the idea of also defining pipelines in a config file as well, as I think this could simplify the problem of "I already have this pipeline but I need to skip this node"

@datajoely I am amenable to the idea of having this in a plug-in instead of kedro core, especially considering I couldn't come up with a simpler way to access `pipelines.yml`

datajoely:
> That being said, I actually do really like the idea of also defining pipelines in a config file as well, as I think this could simplify the problem of "I already have this pipeline but I need to skip this node"

now this is interesting - the canonical Kedro way of doing this is to use pipeline algebra:

```python 

    return {
         "my_pipeline" : pipeline_1
         "my_pipeline2" : pipeline_1 - nodes_to_exclude
    }

```

It's not super clear what your YAML should look like from the PR, what do you imagine this look like?

bpmeek:
@datajoely thanks, I wasn't aware it was something that simple!

> It's not super clear what your YAML should look like from the PR, what do you imagine this look like?

I have a really basic example in the docstring for `from_config` but if your modular pipeline looked like:

```python
modular_pipeline = pipeline(
    pipe=data_processing,
    inputs={
        input_a: input_b
    },
    namespace=modular_example
)
```
then your `pipelines.yml` would look like:

```yaml
<name_doesn't_matter>:
  pipe:
    - data_processing
  inputs:
    input_a: input_b
  namespace: modular_example  
```

datajoely:
Okay interesting - I think I have a better way that we could this (@noklam @astrojuanlu  would love your view too)

So you can actually run a [1:1 from the Kedro CLI with a YAML](https://docs.kedro.org/en/stable/nodes_and_pipelines/run_a_pipeline.html#configure-kedro-run-arguments) using `kedro run --config some_file.yml`

- What we're fighting now, is that the `kedro run --pipeline` syntax is too limiting. 
- We actually had someone on [slack last week ask something very similar](https://linen-slack.kedro.org/t/18836398/hi-everyone-wave-i-am-new-to-kedro-and-have-been-exploring-t#56ae9e4c-3483-4bb3-a77d-705aae4f7986).

In my view, if we were to make the Pipeline CLI slicing commands more expressive then configuring through YAML would improve automatically through the existing mechamnism

- Perhaps if we made the following possible: `kedro run --pipeline="a,b,c" --exclude=d` 
- It would solve the issue you're talking about (since it would be inherited by the CLI `--config` option) as well as the person who I mentioned on Slack above.

astrojuanlu:
Thanks @bpmeek for this contribution!

On the topic of pipelines as YAML, xref #650, https://github.com/kedro-org/kedro/discussions/3872

On the topic of pipeline algebra, I must say this is something most of our users struggle with #3233 

Without having looked at the code, I have two comments to make:

1. I don't think `pipelines.yml` should be in `conf/` at all. Pipeline structure is hardly "configuration" in the way we usually think about it (so much so that we are debating whether even dataset _types_ should be config, see #770). Instead, if we wanted users to define pipelines in YAML, I'd rather see those yamls shipped with the source:

```
❯ tree
.
├── README.md
├── conf
│   ├── README.md
│   ├── base
│   ├── local
│   └── logging.yml
...
├── src
│   ├── package_name
│   │   ├── __init__.py
│   │   ├── __main__.py
│   │   ├── pipelines
│   │   │   ├── __init__.py
│   │   │   ├── data_processing
│   │   │   │   ├── __init__.py
│   │   │   │   ├── nodes.py
│   │   │   │   └── pipeline.yaml  # <--------------
```

2. I 💯 % agree with @datajoely, this should start its life as a Kedro plugin. I'll drop some ideas in https://github.com/kedro-org/kedro-plugins/issues/713

noklam:
>Had I been able to take a couple of the larger more complex pipelines out and moved them into a config file I think it would've been much easier to work with.

I think I don't get this part well. Would breaking `pipeline_registry.py` into smaller files in import them from `pipeline_registry.py` solve this problem already? I am curious to see how the `pipeline_registry.py` looks like as I haven't seen crazily complex registry before, or do you mean `pipeline.py` instead? 

In my mind, the py -> YAML is mostly 1:1 translation from a dictionary to YAML and does not simplify anything. It may make editing easier(?), but worse for reading (you lose all the IDE features that can brings you to the definition/references).

> Pipeline slicing

This is more interesting to me and I think it does have some value. The slicing is already supported by Kedro Pipeline API, but  the CLI only limited to certain expression.

Issue #2749, opened 2023-06-30
dict unpacking with "**params:foo.bar" in `node(inputs=)`


m-gris:
When defining a node, it would be really nice and convenient if the params dictionary could be unpacked with something like: 

```python
node(foo, inputs=['dataset', '**params:bar'] ...
```

This would not only make the node definition more concise than 

```python
inputs=['dataset', "params:bar.a", "params:bar.b", "params:bar.c" ... 
``` 

But it would also result in a more dynamic and flexible node definition,  "immune" to changes in the function's signature. 
(in which case all that would require an update would be the params file, as opposed to both the node def. and the params file in the other case) 

I hope that this request / suggestion does not go against Kedro philosophy :)  

Thx & regards 
M.

astrojuanlu:
Thanks @m-gris for opening this feature request! We'll discuss about it soon.

m-gris:
> Thanks @m-gris for opening this feature request! We'll discuss about it soon.

Thanks for considering it ! :)

noklam:
I have evidence this could be useful too. But it may requires something more than just unpacking

```
# model_predict/functions.py
def make_prediction(model, a,b,c,d,e):
   ...

# model_predict_pipeline/nodes.py
def make_prediction_node(model, params: Dict):
    make_prediction(model, params.get("something", None), 
 params.get("something", None), 
 params.get("something", []), 
 params.get("something", []))
```

deepyaman:
In short, I think it makes sense to pass a parameters dictionary to a node (i.e. the current supported behavior).

While I agree that dict unpacking sounds nice, I wonder how often people would mess up because configuration keys don't match parameter names. Also, how often would the value of supporting this syntax be realized? I feel like it's most beneficial when you have a node (say, `def make_prediction(model, a, b, c, d, e): ...` from the example posted by @noklam) that you want to call both by passing a parameter dictionary and ALSO in cases by explicitly passing different parameters to the various arguments. Does this happen often in a codebase? One could argue that Python supports the unpacking syntax, but I don't think it's fair to think Kedro nodes should be near as flexible.

On the last example of assigning a default, I tend to agree that you could just use keyword arguments with default.

noklam:
I tend to think we shouldn't add arbitrary constrains unless there are good reason. And often users will try to find their way to do it anyway (e.g. Using environment variable for everything).


> @Nok I agree about using keyword arguments, and think you could then simplify it to:

[...]

PR #3905, opened 2024-05-31
Unpack dictionary parameters

bpmeek:
## Description
In response to this [ticket](https://github.com/kedro-org/kedro/issues/2749)

## Development notes
Can now unpack dictionaries in the `inputs` argument of a node, example below.

_unpack_params added to `kedro/kedro/pipeline/node.py`, this function iterates over node inputs and updates `node.inputs` if needed.

*note*: It is assumed that the dictionary entries and kwarg have the same name

### Changes:

`kedro/kedro/pipeline/modular_pipeline.py` was updated to not throw errors when mapping dataset names.
`_is_single_parameter` returns true if name starts with `**params:`
`_normalize_param_name` does not append `params:` if the name already begins with `**params:`
`_validate_datasets_exist` removes datasets that begin with `**` from `non_existent`, I have confirmed that `kedro/runner/runner.py` will catch these missing datasets.

### Undesirable behavior

Currently modular pipelines will still namespace the unpacked parameters, I'll need assistance with either updating the rules list or adjusting the `rename` function in a way that makes sense.

### Examples:

`nodes.py`
```python
def split_data(data: pd.DataFrame, features: List[str], test_size, random_state) -> Tuple:
    X = data[features]
    y = data["price"]
    )
    return X_train, X_test, y_train, y_test
```

`pipeline.py`
```python
node(
    func=split_data,
    inputs=["model_input_table", "**params:model_options"],
    outputs=["X_train", "X_test", "y_train", "y_test"],
    name="split_data_node",
),
```

## Checklist

- [X] Signed off each commit with a [Developer Certificate of Origin (DCO)](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository)


astrojuanlu:
Thanks for this PR @bpmeek! We'll review it shortly 🙏🏼

kedro/pipeline/node.py
```diff
@@ -681,3 +683,27 @@ def _get_readable_func_name(func: Callable) -> str:
...
+    Returns:
+        Either original inputs if no input was unpacked or a list of inputs if an input was unpacked.
+    """
+    use_new = False
+    new_inputs = []
+    for idx, _input in enumerate(_to_list(inputs)):
+        if _input.startswith("**"):
+            use_new = True
+            dict_root = _input.split(":")[-1]
+            _func_arguments = [arg for arg in inspect.signature(func).parameters]
```
datajoely, line 702: So there is a soft assumption that the dictionary keys match the function arguments. Do we want to throw a warning or error if:

(a) No arguments match
(b) Partial arguments match

bpmeek: So that is an assumption, I suppose we could also implement tuple unpacking if a user wanted to use index instead of key-value pairs.

The current behavior of no match/partial match is consistent I think with what users would expect. If you pass the dictionary `model_options` and the function expects a key `features` and that key is missing you would get the following error before the run.

`ValueError: Pipeline input(s) {'params:model_options.features'} not found in the DataCatalog`

datajoely: true! interested to see if others have an opinion though

ElenaKhaustova: I would vote for adding at least basic validation because there are definitely cases in which this method will not work correctly, for example, when having a `**kwargs'—only node function. Then, we could also extend tests with negative samples so it's clear which cases we cover and which we do not.

ElenaKhaustova (commented):
Thank you for the PR!

Left some comments/suggestion regarding the implementation. My main push is to clarify which cases do we cover by adding basic validation, negative tests and small usage example [here](https://docs.kedro.org/en/stable/nodes_and_pipelines/nodes.html)

kedro/pipeline/node.py
bpmeek: I'm a little confused about how we would implement any validation at this point in the process, this function is called when the Node object is being initialized and I don't believe the catalog is available to verify against is it?


> there are definitely cases in which this method will not work correctly, for example, when having a `**kwargs`—only node function.

Can you help me understand this?  if your function was something like `def foo(float_1, integer_2, string_3):` you could pass `**params:dictionary` and it would work. If your function was `def foo(**kwargs)` then this implementation wouldn't work but I wasn't under the impression that was what was being asked for, because you could just change it to `def foo(kwargs)` and then access the values normally from there.

ElenaKhaustova: Sure. I don't mean that this is the expected usage example and we aim to make it work. But it doesn't mean that someone cannot make that (or something else that breaks the current unpacking method) by mistake if syntax allows. There might be some other cases as well. To simplify the validation, we can follow the approach proposed above - check no/partial argument match and add some possible incorrect test cases to ensure the validation catches them. At the same time extend the docs with the expected usage example.

astrojuanlu: I think this is the main outstanding discussion right?

noklam:
@bpmeek Do you want to continue this PR? I am eager to see this being merged finally.

bpmeek:
@noklam, I can continue it, I think I'm just stuck right now on the best way to implement checking for partial matches, since the current process doesn't have access to the catalog when it's populating the required datasets and the verification doesn't know what dictionaries have been unpacked.

kedro/pipeline/node.py
```diff
@@ -681,3 +683,28 @@ def _get_readable_func_name(func: Callable) -> str:
...
+            for param in _func_arguments[idx:]:
+                new_inputs.append(f"params:{dict_root}.{param}")
+else:
+     new_inputs.append(_input)
+        new_inputs.append(_input)
+    if use_new:
+        return new_inputs
+    return inputs
```
noklam, line 710: This seems to be completely broken when I try to test

noklam: Feel free to ping me again if it's ready for review, I cannot fix the error so it's a bit hard to review the logic.

bpmeek: It is, I'm pushing a new update, I think it got broken when I accepted a change in the online editor.

kedro/pipeline/node.py
```diff
@@ -681,3 +683,27 @@ def _get_readable_func_name(func: Callable) -> str:
...
+
+def _unpack_params(
+    func: Callable, inputs: None | str | list[str] | dict[str, str]
+) -> None | str | list[str] | dict[str, str]:
+    """Iterate over Node inputs to see if they need to be unpacked.
+
+    Returns:
+        Either original inputs if no input was unpacked or a list of inputs if an input was unpacked.
+    """
+    use_new = False
```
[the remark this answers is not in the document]
bpmeek: done!

kedro/pipeline/node.py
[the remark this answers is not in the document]
bpmeek: I think this makes sense and went ahead and made the change, but I'm not sure which would be more efficient, creating a list of every node's function or making the list multiple times _if_ the input is a dictionary.

kedro/pipeline/node.py
```diff
@@ -681,3 +683,27 @@ def _get_readable_func_name(func: Callable) -> str:
...
+    new_inputs = []
+    for idx, _input in enumerate(_to_list(inputs)):
+        if _input.startswith("**"):
+            use_new = True
+            dict_root = _input.split(":")[-1]
+            _func_arguments = [arg for arg in inspect.signature(func).parameters]
+            for param in _func_arguments[idx:]:
```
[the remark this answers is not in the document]
bpmeek: I'm not sure I understand what you're recommending.

bpmeek:
Are the additional checks/tests here sufficient or is there still more to be done?

astrojuanlu:
Thanks for your patience @bpmeek! Let us look into this again.

noklam:
I have a long list of PRs so I will only have time to look at this next week earliest.

astrojuanlu:
> Hi @bpmeek , we're slowly catching up with backlog items, sorry about our slow response.
>
> I'm not up to date with https://github.com/kedro-org/kedro/pull/3905 but my understanding is that Kedro-Viz, as consumer of Kedro, shouldn't break - otherwise that would be a sign of a breaking change, that should not have happened until the next Kedro release, 0.20.*.
> 
> In other words, the PR should be backwards compatible.
> 
> The heads up is appreciated 🙏🏼 but for now there's nothing for us to do here.

_Originally posted by @astrojuanlu in https://github.com/kedro-org/kedro-viz/issues/1933#issuecomment-2368645041_

ElenaKhaustova (changes requested):
I'm sorry some comments on the validation required were not clear. I think to move forward with this PR we need to:

1. Address comments regarding the implementation;
2. Update the implementation description and clarify what cases are supported and what are not, so it was clear for the reviewers whether something is skipped intentionally;
3. Add a validation based on 2. and make sure the rest cases are failing gracefully;
4. Extend documentation with the example of added interface: https://docs.kedro.org/en/stable/nodes_and_pipelines/nodes.html#nodes

Thank you @bpmeek for the solution! Please let us know if you want to continue with it or need some help from our side.

kedro/pipeline/node.py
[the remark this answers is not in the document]
ElenaKhaustova: Could we please rename it?

astrojuanlu:
hi @bpmeek do you still intend to pursue this?

merelcht:
This PR has been open for a long time and diverged from the `main`/`develop` branch. I'll close it for now and suggest that if @bpmeek or anyone else wants to continue, they open a new PR. Thanks for being a contributor to Kedro and helping us make the product better. We appreciate your help a lot! ⭐

PR #3952, opened 2024-06-13
Update error message when `kedro-datasets` is not installed or `DataSet` spelling is used

### LINKED ISSUE #3909: [DataCatalog]: Error message is confusing when using `DataSet` instead of `Dataset`
## Description
There is confusion between `DataSet` and `Dataset` terminology, and the error message is not informative when using old naming. They have been renamed in 0.19, but people miss that fact when switching to the new version.

Relates to https://github.com/kedro-org/kedro/issues/2401

## Context
Example of the current error message:

```shell
DatasetError: An exception occurred when parsing config for dataset 'companies':
Class 'pandas.CSVDataSet' not found, is this a typo?
```

### LINKED ISSUE #3911: [DataCatalog]: Error message is confusing when kedro-dataset is not installed
## Description
When `kedro-datasets` is not installed the error message one gets is not informative.

We propose enhancing the error message to provide a clear message on the root cause of the failure - when dataset dependencies are missing.

Relates to https://github.com/kedro-org/kedro/issues/2401

## Context
Currently, users are required to install all dependencies even for unused datasets (in case you want to run pipeline partially or do not load some datasets when standalone catalog usage). The error message generated when some datasets are not installed is unclear, making it difficult for users to understand why the pipeline fails. 

Example of the current error message:

```
DatasetError: An exception occurred when parsing config for dataset 'companies':
Class 'pandas.CSVDataset' not found, is this a typo?
```

This error occurs during dataset configuration parsing and lacks clarity, not suggesting the straightforward solution of installing the necessary package. This can lead to confusion and delays as users may not immediately realize that the issue is due to missing software rather than a typo in their configuration.

ankatiyar:
## Description
Resolve #3911, resolve #3909

## Development notes

The error message below shows up in a number of cases:
```
Class '{dataset_name}' not found, is this a typo?
```
When a user is trying to use a Kedro (not custom) dataset and -
* There is a typo in the dataset type name
* Old spelling "DataSet" is used instead of "Dataset"
* `kedro-dataset` is not installed

Since not having `kedro-datasets` is not the only way this error is triggered, I thought I'd update the error message with a "Hint" like in #3944 which suggests it could be a potential solution without assuming it is the ONLY solution. 
The hint is different (related to #3909) when "DataSet" spelling is used.
I didn't change the `"is this a typo?"` part because that very well might be a case, specially if it's a custom dataset which still should work with the "DataSet" spelling unless the user has changed it.

I'd like to see what the team thinks about the messaging of the error! There's still some ongoing discussion on #3909 so I can revert this part for now. 

## Checklist

### DIFF
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -155,7 +155,7 @@ class AbstractDataset(abc.ABC, Generic[_DI, _DO]):
         except Exception as exc:
             raise DatasetError(
                 f"An exception occurred when parsing config "
-                f"for dataset '{name}':\n{str(exc)}."
+                f"for dataset '{name}':\n{str(exc)}"
             ) from exc
 
         try:
@@ -406,7 +406,21 @@ def parse_dataset_definition(
                 class_obj = tmp
                 break
         else:
-            raise DatasetError(f"Class '{dataset_type}' not found, is this a typo?")
+            hint = ""
+            if "DataSet" in dataset_type:
+                hint = (
+                    "Hint: If you are trying to use a dataset from `kedro-datasets`>=2.0.0, "
+                    "make sure that the dataset name uses the `Dataset` spelling instead of `DataSet`."
+                )
+            else:
+                hint = (
+                    "Hint: If you are trying to use a dataset from `kedro-datasets`, "
+                    "make sure that the package is installed in your current environment. "
+                    "You can do so by running `pip install kedro-datasets`."
+                )
+            raise DatasetError(
+                f"Class '{dataset_type}' not found, is this a typo?" f"\n{hint}"
+            )
 
     if not class_obj:
         class_obj = dataset_type
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -416,7 +416,9 @@ def parse_dataset_definition(
                 hint = (
                     "Hint: If you are trying to use a dataset from `kedro-datasets`, "
                     "make sure that the package is installed in your current environment. "
-                    "You can do so by running `pip install kedro-datasets`."
+                    "You can do so by running `pip install kedro-datasets` or "
+                    "`pip install kedro-datasets[<dataset-group>]` to install `kedro-datasets` along with "
+                    "related dependencies for the specific dataset group."
                 )
             raise DatasetError(
                 f"Class '{dataset_type}' not found, is this a typo?" f"\n{hint}"
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -408,7 +408,7 @@ def parse_dataset_definition(
         else:
             hint = ""
             if "DataSet" in dataset_type:
-                hint = (
+                hint = (  # pragma: no cover # To remove when we drop support for python 3.8
                     "Hint: If you are trying to use a dataset from `kedro-datasets`>=2.0.0, "
                     "make sure that the dataset name uses the `Dataset` spelling instead of `DataSet`."
                 )
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -4,6 +4,7 @@
 
 ## Bug fixes and other changes
 * Updated error message for invalid catalog entries.
+* Updated error message for catalog entries when the dataset class is not found with hints on how to resolve the issue.
 
 ## Breaking changes to the API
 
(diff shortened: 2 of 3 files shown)


kedro/io/core.py
```diff
@@ -406,7 +406,21 @@ def parse_dataset_definition(
...
+            if "DataSet" in dataset_type:
+                hint = (
+                    "Hint: If you are trying to use a dataset from `kedro-datasets`>=2.0.0, "
+                    "make sure that the dataset name uses the `Dataset` spelling instead of `DataSet`."
+                )
+            else:
+                hint = (
+                    "Hint: If you are trying to use a dataset from `kedro-datasets`, "
+                    "make sure that the package is installed in your current environment. "
+                    "You can do so by running `pip install kedro-datasets`."
```
datajoely, line 419:
```suggestion
                    "You can do so by running `pip install kedro-datasets` or `pip install kedro-datasets[polars] for a specific back-end`."
```

wdyt?

kedro/io/core.py
```diff
@@ -406,7 +406,21 @@ def parse_dataset_definition(
...
+                )
+            raise DatasetError(
+                f"Class '{dataset_type}' not found, is this a typo?" f"\n{hint}"
```
datajoely, line 422: I wonder:

- if we could aid debugging by spitting out the full importlib classpath at this point
- if we could check if users are missing an `__int__.py` a common issue

merelcht: > if we could check if users are missing an __int__.py a common issue

Is that Kedro's responsibility though?

kedro/io/core.py
merelcht: I like @datajoely 's suggestion, but I wouldn't use the wording "back-end" and just say for a specific dataset.

merelcht:
I'm happy with this approach 👍

kedro/io/core.py
ankatiyar: I added this with a slightly different wording

noklam (commented):
>When a user is trying to use a Kedro (not custom) dataset and -
> There is a typo in the dataset type name
> Old spelling "DataSet" is used instead of "Dataset"
> kedro-dataset is not installed

This is missing one case where `load_obj` currently handle. Which is `kedro-datasets` is installed but the specific dependencies is not.

i.e. you are using `polars.EagerDataset` but only installed `kedro-datasets` instead of `kedro-datasets[polars]`.

The complexity of mainly comes from kedro.io.core where we try to be smart about the imports.
```python
_DEFAULT_PACKAGES = ["kedro.io.", "kedro_datasets.", ""]
```
The result is that when there is a ModuleNotFoundError, we don't know whether it is:
1. Typo (i.e. the module doesn't exist, wrong path etc)
2. kedro-datasets is not installed
3. kedro-datasets is installed, but the necessary dependencies are missing (handled by `load_obj` currently)

In the past we try to improve (3) but (1)/(2) remains confusing. I wonder if it is simpler to list out all 3 cases explicitly and avoid trying to be too smart?

datajoely:
>In the past we try to improve (3) but (1)/(2) remains confusing. I wonder if it is simpler to list out all 3 cases explicitly and avoid trying to be too smart?

Explicit is better than implicit

ankatiyar:
@noklam It seems like if an optional requirement is missing, this error is not triggered. It is when trying to run the project, the error that shows up is when the dataset is being loaded(or saved) -
```
kedro.io.core.DatasetError: Failed while loading data from data set ExcelDataset(filepath=/Users/ankita_katiyar/kedro_projects/demo/data/01_raw/shuttles.xlsx, load_args={'engine': openpyxl}, protocol=file, save_args={'index': False}, writer_args={'engine': openpyxl}).
Missing optional dependency 'openpyxl'.  Use pip or conda to install openpyxl.
```

datajoely:
This is great - in the past this was an issue because sometimes it was stupid to install all of spark just to just run the Pandas parts. 

I wonder if it's finally time for us to may something like `kedro run --dry` a 1st class command for validating if a session is 'runnable'. It would be useful for our own testing and for users to build an integration test.

astrojuanlu (approved):
I think this is a neat improvement over what we have now, and the phrasing is helpful enough while not too prescriptive to help folks figure out what else to try out (provided that they read it!)

LGTM!

noklam:
@ankatiyar yes, the error is triggered in different place. Since this issue only focus on the typo / kedro-datasets missing case, I have approved this now.

noklam (approved):

PR #3958, opened 2024-06-18
Add default for `env` in `KedroContext`

### LINKED ISSUE #3954: Minimal `kedro new` project fails starter test at `KedroContext` initialization
## Description
This thing:

```python
@pytest.fixture
def project_context(config_loader):
    return KedroContext(
        package_name="my_package_name",
        project_path=Path.cwd(),
        config_loader=config_loader,
        hook_manager=_create_hook_manager(),
    )

Fails the only unit test because a `KedroContext` requires `env` argument to be supplied.
```

## Context
I resolved it for my project by specifying `env=None` there, but it should work out of the box from the starter.

## Steps to Reproduce

1. Initialize `kedro new` with all defaults except pyspark
2. `pip install -e .`
3. `pytest`

## Your Environment

* Kedro version used (`pip show kedro` or `kedro -V`): 0.19.6
* Python version used (`python -V`): 3.10
* Operating system and version: MacOS

ankatiyar:
## Description
Fix #3954 

## Development notes
I checked the test with Kedro 0.18.14 and it was still working. The test breaks from 0.19 with this change - https://github.com/kedro-org/kedro/pull/3300

Add a default value `None` to `env`.

## Checklist

### DIFF
--- a/kedro/framework/context/context.py
+++ b/kedro/framework/context/context.py
@@ -167,12 +167,12 @@ class KedroContext:
 
     project_path: Path = field(init=True, converter=_expand_full_path)
     config_loader: AbstractConfigLoader = field(init=True)
-    _package_name: str = field(init=True)
-    _hook_manager: PluginManager = field(init=True)
+    env: str | None = field(init=True, default=None)
+    _package_name: str = field(init=True, default="")
+    _hook_manager: PluginManager | None = field(init=True, default=None)
     _extra_params: dict[str, Any] | None = field(
         init=True, default=None, converter=deepcopy
     )
-    env: str | None = field(init=True, default=None)
 
     @property
     def catalog(self) -> DataCatalog:
--- a/kedro/framework/context/context.py
+++ b/kedro/framework/context/context.py
@@ -167,12 +167,12 @@ class KedroContext:
 
     project_path: Path = field(init=True, converter=_expand_full_path)
     config_loader: AbstractConfigLoader = field(init=True)
-    env: str | None = field(init=True)
     _package_name: str = field(init=True)
     _hook_manager: PluginManager = field(init=True)
     _extra_params: dict[str, Any] | None = field(
         init=True, default=None, converter=deepcopy
     )
+    env: str | None = field(init=True, default=None)
 
     @property
     def catalog(self) -> DataCatalog:


yury-fedotov:
Thanks for looking into this!

ankatiyar:
@yury-fedotov Thanks for reporting the issue :)

noklam (commented):
Thanks! I overlooked this. Can we keep the change at the same line? This is changing the order of arguments, is it needed?

ankatiyar:
> Thanks! I overlooked this. Can we keep the change at the same line? This is changing the order of arguments, is it needed? 

@noklam It was erroring out saying arguments with default values cannot be before ones without default values

merelcht:
> > Thanks! I overlooked this. Can we keep the change at the same line? This is changing the order of arguments, is it needed?
> 
> @noklam It was erroring out saying arguments with default values cannot be before ones without default values

Unfortunately, changing the order of arguments is a breaking change. So in this case it's better to ignore the error about argument order instead.

ankatiyar:
I've gone back to the original order but added default values to subsequent arguments. Not sure if this has any implications but the tests seem to be passing. The other way to fix the issue linked would be to change the example test in starters. WDYT? 🤔 @noklam @merelcht

noklam:
```python
    env: str | None = field(init=True, default=None)
    _package_name: str = field(init=True, default="")
    _hook_manager: PluginManager | None = field(init=True, default=None)
```

Unfortunately I think there is no nice solution here. ^ I don't think there should be a default for `hook_manager`, does it even work when it is None?

For `package_name`, I did a quick search and find no reference, I don't really know why it is needed. The problem is we cannot add a default for `env` without also adding default for all subsequent arguments. If we add default to all of them it's also a breaking change. (adding only to env is OK because I consider that's a bug fix, but it's not possible as mentioned).

**My suggestion is not to fix anything in `kedro`, but remove the test**. In fact, I don't think the `KedroContext` has any use in tests. In a standard kedro run, we should create the session and let the session to create context, runner, config_loader etc. However, in our test template, we created the [config loader ](https://github.com/kedro-org/kedro-starters/blob/db79aec64c4a0f062321bd8c74ee782750bba527/spaceflights-pandas-viz/%7B%7B%20cookiecutter.repo_name%20%7D%7D/tests/test_run.py#L16-L18)and [runner](https://github.com/kedro-org/kedro-starters/blob/db79aec64c4a0f062321bd8c74ee782750bba527/spaceflights-pandas-viz/%7B%7B%20cookiecutter.repo_name%20%7D%7D/tests/pipelines/data_science/test_pipeline.py#L65) manually. In addition, the context actually use the [real hooks](https://github.com/kedro-org/kedro-starters/blob/db79aec64c4a0f062321bd8c74ee782750bba527/spaceflights-pandas-viz/%7B%7B%20cookiecutter.repo_name%20%7D%7D/tests/test_run.py#L27), but the tests that created run with no hook, so they are not consistent at all.

In our own [testing docs](https://docs.kedro.org/en/stable/tutorial/test_a_project.html), KedroContext is not mentioned at all.

ankatiyar:
@noklam yeah, i'll add your comment to the issue and put that in backlog and close this PR for now!

Issue #3914, opened 2024-06-03
[DataCatalog]: Autocompletion support for accessing datasets


ElenaKhaustova:
## Description
Users struggle to find datasets within the catalog, particularly when dealing with a large number of datasets. They express the need for autocomplete functionality when accessing datasets in the catalog.

We propose implementing autocompletion support for accessing datasets in the catalog, enabling users to receive suggestions for dataset names as they type.

Relates to https://github.com/kedro-org/kedro/issues/1721

## Context

- "We have a lot of catalog entries which is common because we store a lot of intermediate results, we go back and forth with the `YAML` file to find how they named the dataset."
- "I think that the tab completion would be nice to have."
- "The current separate configuration structure requires excessive navigation."

astrojuanlu:
Re-stating what I said about dynamic properties (aka "pandas .column access") in #1721:

> The problem with doing the dynamic properties is that some dataset names that are valid in YAML would become illegal in that way (same problem as with pandas columns) and also it would pollute the namespace of the DataCatalog (again, same problem)

Is this something that could be addressed with https://github.com/kedro-org/vscode-kedro @noklam ?

noklam:
I think address this with VSCode extension is possible, but I think we should exhaust solutions that work for most of the things first. I know inherit from `dict` is bad for a reason, but this is almost the most ideal solution that satisfy all my needs. WDYT?

![image](https://github.com/kedro-org/kedro/assets/18221871/300b29ce-9c1b-47f2-9cb2-c745156c8715)

astrojuanlu:
This looks fantastic, and if it works on IPython I'm sure it will work in other places. Wondering if `dict`s are special-cased or if it's enough for a class to implement `__getitem__` and `keys()`.

According to https://stackoverflow.com/a/38732914, it's supported since 2014 https://github.com/ipython/ipython/pull/5304

astrojuanlu:
This does not rely on LSP though, it's a special IPython functionality https://github.com/ipython/ipython/blob/1b4607fbee253a718df14419414f624dfde1164e/IPython/core/completer.py#L2488-L2510

noklam:
Quote from Slack discussion, I think we have a promising solution now! (TypedDict). Is this ready enough to put in a sprint? Do we want to discuss on the API? I proposed one and asked in [Slack](https://kedro-org.slack.com/archives/C03R8N2M8KT/p1717688522981599)

[...]

Issue #3893, opened 2024-05-25
Investigate performance of config loading for big projects


astrojuanlu:
## Description
Earlier this week a user reached out to me in private saying that it was taking 3 minutes for Kedro to load their configuration (`KedroContext._get_catalog`).

Today another user mentioned that "Looking at the logs, it gets stuck at the kedro.config.module for more than 50% of the pipeline run duration, but we do have a lot of inputs and outputs"

I still don't have specific reproducers, but I'm noticing enough qualitative evidence to open an issue about it.

datajoely:
I'd like to see us add a CLI command which users can run to produce a flamegraph. It would massively reduce the guesswork here.

`kedro profile {kedro command}` -> `.html` / `.bin`

yury-fedotov:
> I'd like to see us add a CLI command which users can run to produce a flamegraph. It would massively reduce the guesswork here.
> 
> 
> 
> `kedro profile {kedro command}` -> `.html` / `.bin` 

@datajoely flamegraph for the entire pipeline run (how much time each node takes) or just the config resolution / pipeline initialization?

datajoely:
In my mind, it would run the whole command as normal, but also generate the profiling data. 

Perhaps if we were to take this seriously, a full on [memray](https://github.com/bloomberg/memray) integration would incredible.

astrojuanlu:
Continuing the discussion on creating custom commands here https://github.com/kedro-org/kedro/discussions/3908

astrojuanlu:
> Many users have been complaining about the slowness of Kedro with big projects and that can be attributed to many different causes. However one of the most prevailing cause is big parameter files that get expanded into hundreds of datasets on their own. That process takes a lot of time and if the files become too big (a couple of MB), it presents as significant slowdown.

_Originally posted by @idanov in https://github.com/kedro-org/kedro/pull/3732#issue-2202743592_

The solution works, but couples the `DataCatalog` with `OmegaConf`. From the discussion in the PR:

> Shouldn't we redesign the `DataCatalog` API instead so that parameters are first class citizens, and not fake datasets?

There were a few thumbs up to the idea, and it was brought up again in https://github.com/kedro-org/kedro/pull/3973 (@datajoely please do confirm that this is what you had in mind 😄)

@merelcht pointed out that there's a pending research item on how users use parameters and for what #2240 

@ElenaKhaustova agreed that this is relevant in the context of the ongoing `DataCatalog` API redesign #3934.

[...]

PR #3732, opened 2024-03-22
Remove the copying hack and add proper params querying capabilities in the DataCatalog

idanov:
## Description

Many users have been complaining about the slowness of Kedro with big projects and that can be attributed to many different causes. However one of the most prevailing cause is big parameter files that get expanded into hundreds of datasets on their own. That process takes a lot of time and if the files become too big (a couple of MB), it presents as significant slowdown.

This Draft PR is a POC in dropping the hack we have built into Kedro to support the convenience syntax of `params:x.y.z` and replacing it with a proper query instead, powered by `OmegaConf.select`. This way all datasets which load into a Python dictionary can provide the same functionality out of the box.

A side effect is the removal of all those `params:xxxx` datasets from the output of `catalog.list()`, which is something people have been annoyed by anyways. Nevertheless, it still presents a breaking change, so we need to decide whether it will need a new breaking Kedro version or it can go in as a bugfix/performance fix.

This profiling revealed another potential source of slowness and that's the loading of the config files, which is something we should investigate further in the future.

## Development notes

Tested with an autogenerated 2.2MB parameters file. As you can see that nearly 2/3 of the time is shaved off.

Before (~120s and 1.17GB memory):

```
╰─❯ pyinstrument --show '*/kedro/*' -m kedro catalog list
............................
  _     ._   __/__   _ _  _  _ _/_   Recorded: 15:01:57  Samples:  80971
 /_//_/// /_\ / //_// / //_'/ //     Duration: 120.480   CPU time: 123.195
/   _/                      v4.6.2

Program: pyinstrument --show */kedro/* -m kedro catalog list

120.476 <module>  kedro/__main__.py:1
└─ 119.979 main  kedro/framework/cli/cli.py:192
   ├─ 118.307 KedroCLI.__call__  click/core.py:1155
   │  └─ 118.307 KedroCLI.main  kedro/framework/cli/cli.py:107
   │     └─ 118.306 KedroCLI.main  click/core.py:1010
   │           [6 frames hidden]  click
   │              118.306 new_func  click/decorators.py:44
   │              └─ 118.302 list_datasets  kedro/framework/cli/catalog.py:37
   │                 └─ 117.349 KedroContext.catalog  kedro/framework/context/context.py:177
   │                    └─ 117.338 KedroContext._get_catalog  kedro/framework/context/context.py:209
   │                       ├─ 74.947 DataCatalog.add_feed_dict  kedro/io/data_catalog.py:638
[...]

Issue #2240, opened 2023-01-23
What do our users use parameters for?


yetudada:
## Description
Parameters in a Kedro project has been a long-standing feature. Users are able to interact with parameters for the data engineering and data science pipelines by using: 
 - the `parameters.yml` file located in `conf/base` 
 - a `parameters` folder and adding relevant parameters according to their own naming scheme or because Kedro creates parameter files when you create a new pipeline
- the CLI and passing overriding parameters specified in the YAML files

We have yet to learn what our users have been using parameters for, and it's time to understand this so that we can better support their use cases. 

The scope of this ticket includes: 
  - A research exercise in unpacking the different use cases for parameters
    - Additionally, we may need to define the different types of parameters that our users interact with - this language is vague
  - A design exercise to recommend ways that we can address these use cases

## Data sources
There are a few data sources that can be leveraged to support this project: 
 - Qualitative: 
   - Interviews with users; we will need to compile the list
   - Observation studies; combing through existing Kedro projects to see how parameters were used or ask our users to show us what they've done 
 - Quantitative:
   - Have a look at telemetry data from `kedro-telemetry` to see the use case for `kedro run --params param_key1:value1,param_key2:2.0` in action

## Random thoughts

I have been curious about supporting the reporting use case, e.g. it appears that some teams will create Jupyter notebooks and include parameters that can be changed in the notebook. The idea is that an end user can change those parameters, run the pipeline from there, and then see the charts and generated tables. If this use case is prevalent (I have no data to say that it is), could this be supported on Kedro-Viz - users change the parameters, run the pipelines and see dashboards?

datajoely:
We have a lot of internal validation that passing the global `parameters` mega block should be discouraged and we should encourage users to provide as specific keys as possible. It makes typing much much easier and maintainable. 

I would love to see the `parameters` mega block be dropped in 1.0.0

astrojuanlu:
It's the 2 year anniversary of this issue and, while the questions haven't changed much, it's not a priority. There are other pressing issues with parameters that users are complaining about, for example #4437.

I'll close this for now, we can always reopen or create a fresh research plan.

PR #3973, opened 2024-06-28
Key completion for dataset access

ElenaKhaustova:
## Description
Solves https://github.com/kedro-org/kedro/issues/3914

## Development notes

- Implemented `IPython` and `Jupyter` key completions for accessing datasets in the catalog. The solution is based on defining `_ipython_key_completions_()` for `_FrozenDatasets` class.

Screenshot 2024-07-01 at 08 55 01

- Original names (with non-word characters) are suggested as keys but those that are replaced with `__` (as for properties) are accepted as well.

Screenshot 2024-07-01 at 08 53 32

- Manually tested with transcoded datasets and configuration patterns.

## Checklist


kedro/io/data_catalog.py
```diff
@@ -105,28 +105,36 @@ def __init__(
...
+            return
         msg = "Operation not allowed! "
         if key in self.__dict__:
             msg += "Please change datasets through configuration."
         else:
             msg += "Please use DataCatalog.add() instead."
         raise AttributeError(msg)
 
+    def _ipython_key_completions_(self) -> list[str]:
+        return list(self._original_names)
```
datajoely, line 133: So there is an argument that simplicity may be better, but what about including the `dataset.__class__.__name__` so you would see something like `shuttles (CSVDataset)`

datajoely: I would also artificially place `parameters` at the very bottom to discourage its use

ElenaKhaustova: > So there is an argument that simplicity may be better, but what about including the `dataset.__class__.__name__` so you would see something like `shuttles (CSVDataset)`?

I think it's a good idea and I see the value. The only concern is whether we are fine with those keys. Cause with tab completion we will get `catalog.datasets["shuttles (CSVDataset)"]` which will work but not sure it looks good from the syntax side.

ElenaKhaustova: I'll check if it's possible to provide some metadata which is not part of the key but still displayed with `IPython` and `Jupyter` UI.

kedro/io/data_catalog.py
```diff
@@ -105,28 +105,36 @@ def __init__(
...
+
+    def __getitem__(self, key: str) -> Any:
+        return self.__dict__[_sub_nonword_chars(key)]
```
noklam, line 136: Can we change `self._original_names` to a dictionary and modify the `__getitem__` to access data through this mapping?

i.e. `self.__dict__[self._original_names[key]]`

noklam (commented):
Generally looks good, I have some questions:
image

1. When I try to prompt the autocompletion, it seems to prompt me to use the single quote instead of the more common `"` in Python. Is this something we can control or it's how IPython completion works? If I type `catalog.datasets["]` and hit tab then it would finish the autocompletion without the finishing `"`, which means a few more keystrokes.  This is not a deal breaker but just want to check if we can make this experience smoother.

kedro/io/data_catalog.py
ElenaKhaustova: I aimed to make both original names (with non-word characters) - `catalog.datasets["params:model_option"]` and transformed names - `catalog.datasets["params__model_option"]` work, so ended up with the current solution. So there's less discrepancy with by attribute access `catalog.datasets.params__model_option`

ElenaKhaustova:
@datajoely, regarding your question about placing parameters at the bottom - sorting is applied internally in `ipython` `completer.py` after `_ipython_key_completions_()` is called, so I cannot force a custom order. Keys are always in lexicographic order.

ElenaKhaustova:
> Generally looks good, I have some questions: image
> 
> 1. When I try to prompt the autocompletion, it seems to prompt me to use the single quote instead of the more common `"` in Python. Is this something we can control or it's how IPython completion works? If I type `catalog.datasets["]` and hit tab then it would finish the autocompletion without the finishing `"`, which means a few more keystrokes.  This is not a deal breaker but just want to check if we can make this experience smoother.

Yesk, that's the way [IPython completion](https://github.com/ipython/ipython/blob/a499dbd507a92ea0087eca0f53e550fc838f0580/IPython/core/completer.py#L2489) works and there's no any API to customise it from the outside. I see [some logic](https://github.com/ipython/ipython/blob/a499dbd507a92ea0087eca0f53e550fc838f0580/IPython/core/completer.py#L2558) for matching the closing quote but it's not straightforward why it's not working. So there's not much that we can do but create an issue and ask.

noklam (approved):

noklam:
Thanks for checking, I think this is good enough.

astrojuanlu:
> Sounds good! @ElenaKhaustova , I petition that we drop the `parameters` god object in `1.0.0` - cc @astrojuanlu

Not sure if related, but https://github.com/kedro-org/kedro/pull/3732#discussion_r1537828647

> Shouldn't we redesign the `DataCatalog` API instead so that parameters are first class citizens, and not fake datasets?

The recent research didn't surface any user problems around that, but if we're to change anything in that regard, probably the time to look at it is now. Shall we open a separate issue about it?

merelcht:
We still have the issue #2240 originally intended for a big research piece around parameter usage.

ElenaKhaustova:
> > Sounds good! @ElenaKhaustova , I petition that we drop the `parameters` god object in `1.0.0` - cc @astrojuanlu
> 
> Not sure if related, but [#3732 (comment)](https://github.com/kedro-org/kedro/pull/3732#discussion_r1537828647)
> 
> > Shouldn't we redesign the `DataCatalog` API instead so that parameters are first class citizens, and not fake datasets?
> 
> The recent research didn't surface any user problems around that, but if we're to change anything in that regard, probably the time to look at it is now. Shall we open a separate issue about it?

The idea sounds good, at least from the performance point of view mentioned in https://github.com/kedro-org/kedro/pull/3732#discussion_r1537828647. I think it also falls into the discussion about making datasets of first-class citizens, which we'll have anyway in terms of future `DataCatalog` redesign work. Vote for opening an issue.

astrojuanlu:
Left a comment at https://github.com/kedro-org/kedro/issues/3893#issuecomment-2203406999 and marking the off-topic comments as such.

astrojuanlu (approved):
Works like a charm! ✨ 

image

PR #3976, opened 2024-07-01
Make `kedro-telemetry` a core dependency

ElenaKhaustova:
## Description
Partly solves https://github.com/kedro-org/kedro-plugins/issues/726

## Development notes

- Removed `kedro-telemetry` dependency from the `requirements.txt` file in `Kedro` project template.
- Modified the `pyproject.toml` file to include `kedro-telemetry` in the dependencies section.

## Checklist


ElenaKhaustova:
Unit tests fail if `kedro-telemetry` is installed in the test environment.

astrojuanlu:
Key question from me: does this create a circular dependency? kedro depending on kedro-telemetry and viceversa

ElenaKhaustova:
> Key question from me: does this create a circular dependency? kedro depending on kedro-telemetry and viceversa

Yes, that's what we obviously missed 🙈

@DimedS shall we consider moving `kedro-telemetry` to framework instead?

DimedS:
I would be happy to incorporate the kedro-telemetry code into Kedro, and I believe it should be done. However, I thought that at this stage, it would be a significant change, so I decided to start with a minor step. As a temporary solution, I don't see any major issues with this type of circular dependency; it essentially means that the code is part of one package. I don't anticipate any usage problems with this approach. Nevertheless, I would be happy to discuss whether we are ready to integrate the telemetry code into Kedro, as this is definitely the correct long-term solution.

ElenaKhaustova:
As far as I understand, pip indeed can deal with circular dependencies as long as they are both not build dependencies of each other - which is our case.
https://discuss.python.org/t/handling-installation-of-circular-dependencies/25531/8

But from the code organisation perspective, it feels wrong to me unless we have reasonable arguments for keeping telemetry in a separate package. Now (with recent changes of making telemetry opt-out), we expect the `kedro-telemetry` package to be installed anyway, regardless of whether telemetry is collected, so it doesn't reduce framework dependencies. On the other side `kedro-telemetry` always goes with `kedro`, which means they probably should live in the same code base, otherwise, we need to understand what advantages we get from keeping them separate.

@DimedS, I agree with your points on long-term solution and am curious what others think.

DimedS:
I also tested by creating circular dependencies in two libraries and uploading them to PyPI. Everything worked well, with no warnings during the upload to PyPI or while installing with pip. Based on this, I believe that for the next Kedro release, it would be better to integrate kedro-telemetry as a dependency of Kedro. For the subsequent major release, we can consider integrating the telemetry directly into Kedro, possibly in conjunction with splitting Kedro into different modules.

kedro/framework/cli/cli.py
```diff
@@ -74,15 +74,11 @@ def info() -> None:
...
-    if plugin_versions:
-        click.echo("Installed plugins:")
-        for plugin_name, plugin_version in sorted(plugin_versions.items()):
-            entrypoints_str = ",".join(sorted(plugin_entry_points[plugin_name]))
-            click.echo(
-                f"{plugin_name}: {plugin_version} (entry points:{entrypoints_str})"
-            )
-    else:
-        click.echo("No plugins installed")
+    # plugin_versions now always include kedro-telemetry as it was moved to core kedro dependency
```
DimedS, line 77: I believe it is better to leave the code as it is now because it's more generic and users still can manually uninstall the `kedro-telemetry` which will produce "No plugins installed", and I believe in some time in future we will integrate `kedro-telemetry` inside of `kedro` and will need to change that back.

astrojuanlu:
> I believe that for the next Kedro release, it would be better to integrate kedro-telemetry as a dependency of Kedro.

Fantastic, thanks 👍🏼

> For the subsequent major release, we can consider integrating the telemetry directly into Kedro, possibly in conjunction with splitting Kedro into different modules.

I 100 % agree.

kedro/framework/cli/cli.py
ElenaKhaustova: Done

DimedS (approved):
Thank you for the PR, @ElenaKhaustova , tested it, everything works well!

pyproject.toml
```diff
@@ -22,6 +22,7 @@ dependencies = [
     "gitpython>=3.0",
     "importlib-metadata>=3.6,<9.0",
     "importlib_resources>=1.3,<7.0",  # The `files()` API was introduced in `importlib_resources` 1.3 and Python 3.9.
+    "kedro-telemetry>=0.3.1",
```
merelcht, line 25: Is `0.3.1` an acceptable version or should we just up this to `0.5.0` so we get the latest with all improvements on identifying users and projects?

merelcht (commented):
Do we have to wait for a breaking release to move the `kedro-telemetry` code to Kedro? Currently, telemetry is not a dependency so adding the code into Kedro isn't a breaking change.

pyproject.toml
ElenaKhaustova: IMO makes sense to update, indeed. @DimedS, wdyt?

ElenaKhaustova:
> Do we have to wait for a breaking release to move the `kedro-telemetry` code to Kedro? Currently, telemetry is not a dependency so adding the code into Kedro isn't a breaking change.



> Do we have to wait for a breaking release to move the `kedro-telemetry` code to Kedro? Currently, telemetry is not a dependency so adding the code into Kedro isn't a breaking change.

It's not a breaking change the only thing which is blocking it is some additional code restructuring that needs to be done before which @DimedS mentioned above https://github.com/kedro-org/kedro/pull/3976#issuecomment-2203341038. Otherwise, we can proceed with moving.

pyproject.toml
DimedS: Yes, I believe it makes sense to update it now, but it's more important to remember to update it with a new kedro-telemetry version after the next release of kedro-telemetry. This should be done after we complete all opt-out work, but before the Kedro release.

merelcht (approved):
Don't forget to add this to the release notes!

noklam (approved):
As I seen it in the discussion, it seems like circular dependencies wasn't a big problem, so I don't have problem with this approach.

I test it with uv to see what happens if I try to force an incompatible dependencies:
> (kedro) gitpod /workspace/kedro (feature/telemetry-core) $ uv pip install -e . kedro-tele
metry==0.3.1
  × No solution found when resolving dependencies:
  ╰─▶ Because only kedro==0.19.6 is available and kedro==0.19.6 depends on
      kedro-telemetry>=0.5.0, we can conclude that all versions of kedro depend on
      kedro-telemetry>=0.5.0.
      And because you require kedro-telemetry==0.3.


1. Do we have an issue to track the long term solution? Is it a breaking change or something that we can already implement in 0.19.x?

ElenaKhaustova:
> As I seen it in the discussion, it seems like circular dependencies wasn't a big problem, so I don't have problem with this approach.
> 
> I test it with uv to see what happens if I try to force an incompatible dependencies:
> 
> > (kedro) gitpod /workspace/kedro (feature/telemetry-core) $ uv pip install -e . kedro-tele
> > metry==0.3.1
> > × No solution found when resolving dependencies:
> > ╰─▶ Because only kedro==0.19.6 is available and kedro==0.19.6 depends on
> > kedro-telemetry>=0.5.0, we can conclude that all versions of kedro depend on
> > kedro-telemetry>=0.5.0.
> > And because you require kedro-telemetry==0.3.
> 
> 1. Do we have an issue to track the long term solution? Is it a breaking change or something that we can already implement in 0.19.x?

It's not a breaking change and we agreed to make it before the rest of the modularisation.

Here is an issue created: https://github.com/kedro-org/kedro/issues/3999

Issue #3980, opened 2024-07-02
Pretty printing: `AbstractDataset.__repr__`


ElenaKhaustova:
## Description
Parent ticket: https://github.com/kedro-org/kedro/issues/3913

Implement `__repr__` for `AbstractDataset` for better dataset representation and printing and further use it within `DataCatalog.__repr__`

## Context
https://github.com/kedro-org/kedro/issues/3913#issue-2330910213
https://github.com/kedro-org/kedro/issues/1721

## Possible Implementation

- We already have an implementation of `__str__` method for `AbstractDataset` based on the dataset's `_describe` which can be adjusted and moved to `__repr__`.
- Update `_describe` for `MemoryDataset`, `LambdaDataset`, `SharedMemoryDataset`, and `CachedDataset` if needed.
- One of the potential solutions is to extend the built-in `pprint.PrettyPrinter`.

ElenaKhaustova:
I've prototyped two different approaches for printing:

- both approaches implement `__repr__()` method for `AbstractDataset` based on the implementation of `_describe()` method for the specific dataset. Since the `__str__` method is not implemented `__repr__()` is called when converting object to string, so we have the same results when `print(obj)`, `repr(obj)`,` obj`.
- https://github.com/kedro-org/kedro/pull/3990 - the first approach prints datasets in one line in a format `module.Class(arg1=val_1, ..., arg_n=val_n)` where argument values are formatted with `pprint.pformat` and then joined as strings, so we construct the resulting string from formatted strings.

![Screenshot 2024-07-05 at 00 27 12](https://github.com/kedro-org/kedro/assets/157851531/d359381d-680d-4eac-b6fa-42b78b115afe)

- https://github.com/kedro-org/kedro/pull/3992 -  the second approach represents dataset in dict format `{'module.Class': {'arg_1:' val_1}, ..., {}}`, so at first the end dictionary is created and then it is formatted with `pprint.pformat` thus we can control indentation but it looks less compact.

![Screenshot 2024-07-05 at 00 16 11](https://github.com/kedro-org/kedro/assets/157851531/da576613-a90b-4e7d-99c8-2bc1ec9af4e5)

- for both approaches, we can control depth, so we can hide arguments after a certain indentation level
- they also both look representative enough compared to what we had before but the first is more compact, while the second keeps the indentation

I'm curious about what you think. Does it feel good enough? Do we want more or less information provided? Which approach seems better?

noklam:
[...]

PR #3987, opened 2024-07-03
Pretty printing dataset

ElenaKhaustova:
## Description
Implementation of one-line approach described here: https://github.com/kedro-org/kedro/issues/3980

## Checklist

### DIFF
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -228,16 +228,18 @@ class AbstractDataset(abc.ABC, Generic[_DI, _DO]):
             message = f"Failed while saving data to data set {str(self)}.\n{str(exc)}"
             raise DatasetError(message) from exc
 
-    def __repr__(self) -> str:
+    def pretty_repr(self, object_description: dict[str, Any]):
         str_keys = []
-        object_description = self._describe()
         for arg_name in sorted(object_description, key=lambda key: str(key)):
             if object_description[arg_name] is not None:
                 str_keys.append(
-                    f"{arg_name}={pprint.pformat(object_description[arg_name], compact=True)}"
+                    f"{arg_name}={pprint.pformat(object_description[arg_name], compact=True, width=300)}"
                 )
 
         return f"{type(self).__module__}.{type(self).__name__}({', '.join(str_keys)})"
+
+    def __repr__(self) -> str:
+        return self.pretty_repr(self._describe())
 
     @abc.abstractmethod
     def _load(self) -> _DO:
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -7,6 +7,7 @@ import abc
 import copy
 import logging
 import os
+import pprint
 import re
 import warnings
 from collections import namedtuple
@@ -253,6 +254,12 @@ class AbstractDataset(abc.ABC, Generic[_DI, _DO]):
             return str(obj)
 
         return f"{type(self).__name__}({_to_str(self._describe(), True)})"
+
+    def __repr__(self):
+        return (
+            f"{type(self).__module__}.{type(self).__name__}"
+            f"({pprint.pformat(self._describe(), compact=True, width=300)})"
+        )
 
     @abc.abstractmethod
     def _load(self) -> _DO:
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -9,6 +9,7 @@ import logging
 import os
 import pprint
 import re
+import sys
 import warnings
 from collections import namedtuple
 from datetime import datetime, timezone
@@ -237,7 +238,7 @@ class AbstractDataset(abc.ABC, Generic[_DI, _DO]):
                     sort_dicts=False,
                     compact=True,
                     depth=2,
-                    width=500,
+                    width=sys.maxsize,
                 )
                 str_keys.append(f"{arg_name}={descr}")
 
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -229,7 +229,7 @@ class AbstractDataset(abc.ABC, Generic[_DI, _DO]):
             raise DatasetError(message) from exc
 
     def __str__(self) -> str:
-        def _to_str(obj: Any, is_root: bool = False) -> str:
+        def _to_str(obj: Any) -> str:
             """Returns a string representation where
             1. The root level (i.e. the Dataset.__init__ arguments) are
             formatted like Dataset(key=value).
@@ -237,29 +237,31 @@ class AbstractDataset(abc.ABC, Generic[_DI, _DO]):
             3. None values are not shown.
             """
 
-            fmt = "{}={}" if is_root else "'{}': {}"  # 1
-
             if isinstance(obj, dict):
-                sorted_dict = sorted(obj.items(), key=lambda pair: str(pair[0]))  # 2
+                str_keys = []
+                for arg_name in sorted(obj, key=lambda key: str(key)):
+                    if obj[arg_name] is not None:
+                        str_keys.append(
+                            f"{arg_name}={pprint.pformat(obj[arg_name], compact=True)}"
+                        )
 
-                text = ", ".join(
-                    fmt.format(key, _to_str(value))  # 2
-                    for key, value in sorted_dict
-                    if value is not None  # 3
-                )
-
-                return text if is_root else "{" + text + "}"  # 1
+                return ", ".join(str_keys)
 
             # not a dictionary
             return str(obj)
 
-        return f"{type(self).__name__}({_to_str(self._describe(), True)})"
+        return f"{type(self).__name__}({_to_str(self._describe())})"
 
-    def __repr__(self):
-        return (
-            f"{type(self).__module__}.{type(self).__name__}"
-            f"({pprint.pformat(self._describe(), compact=True, width=300)})"
-        )
+    def __repr__(self) -> str:
+        str_keys = []
+        object_description = self._describe()
+        for arg_name in sorted(object_description, key=lambda key: str(key)):
+            if object_description[arg_name] is not None:
+                str_keys.append(
+                    f"{arg_name}={pprint.pformat(object_description[arg_name], compact=True)}"
+                )
+
+        return f"{type(self).__module__}.{type(self).__name__}({', '.join(str_keys)})"
 
     @abc.abstractmethod
     def _load(self) -> _DO:
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -232,9 +232,14 @@ class AbstractDataset(abc.ABC, Generic[_DI, _DO]):
         str_keys = []
         for arg_name in sorted(object_description, key=lambda key: str(key)):
             if object_description[arg_name] is not None:
-                str_keys.append(
-                    f"{arg_name}={pprint.pformat(object_description[arg_name], sort_dicts=False, compact=True, depth=2, width=500)}"
+                descr = pprint.pformat(
+                    object_description[arg_name],
+                    sort_dicts=False,
+                    compact=True,
+                    depth=2,
+                    width=500,
                 )
+                str_keys.append(f"{arg_name}={descr}")
 
         return f"{type(self).__module__}.{type(self).__name__}({', '.join(str_keys)})"
 
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -233,7 +233,7 @@ class AbstractDataset(abc.ABC, Generic[_DI, _DO]):
         for arg_name in sorted(object_description, key=lambda key: str(key)):
             if object_description[arg_name] is not None:
                 str_keys.append(
-                    f"{arg_name}={pprint.pformat(object_description[arg_name], compact=True)}"
+                    f"{arg_name}={pprint.pformat(object_description[arg_name], compact=True, depth=2, width=300)}"
                 )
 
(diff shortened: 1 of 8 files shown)


kedro/io/core.py
```diff
@@ -227,32 +228,18 @@ def save(self, data: _DI) -> None:
...
+            if object_description[arg_name] is not None:
+                str_keys.append(
+                    f"{arg_name}={pprint.pformat(object_description[arg_name], compact=True)}"
                 )
 
-                return text if is_root else "{" + text + "}"  # 1
-
-            # not a dictionary
-            return str(obj)
+        return f"{type(self).__module__}.{type(self).__name__}({', '.join(str_keys)})"
```
deepyaman, line 239: I think the issue with this is that all of the pretty-printed subkeys will be on a single line, and no formatting/wrapping is applied there. By using `PrettyPrinter` directly, you can get something pretty similar, and also get the formatting.

Happy to share some more context/pointers on this tomorrow or next week, if you'd like.

ElenaKhaustova: I played a bit with `PrettyPrinter` and different indentations but in the end, I preferred one line to it. There shouldn't be too many keys as they are some subset of class constructor input arguments passed via `_describe()`. I was planning to have this as a bare minimum but let me add a second option with the indentation, so we can choose which one looks better.

Happy to discuss them tomorrow.

kedro/io/core.py
```diff
@@ -227,32 +228,23 @@ def save(self, data: _DI) -> None:
...
+                    depth=2,
+                    width=500,
                 )
+                str_keys.append(f"{arg_name}={descr}")
 
-                return text if is_root else "{" + text + "}"  # 1
-
-            # not a dictionary
-            return str(obj)
+        return f"{type(self).__module__}.{type(self).__name__}({', '.join(str_keys)})"
```
deepyaman, line 244: If it comes from `kedro_datasets`, I feel like it would be much cleaner to have the short name users can provide (i.e. `pandas.CSVDataset` instead of `kedro_datasets.pandas.csv_dataset.CSVDataset`).

Similarly, from Kedro core, `MemoryDataset`, etc. seems much easier to read than the full path.

kedro/io/core.py
```diff
@@ -227,32 +228,23 @@ def save(self, data: _DI) -> None:
...
-            if isinstance(obj, dict):
-                sorted_dict = sorted(obj.items(), key=lambda pair: str(pair[0]))  # 2
-
-                text = ", ".join(
-                    fmt.format(key, _to_str(value))  # 2
-                    for key, value in sorted_dict
-                    if value is not None  # 3
+    def _pretty_repr(self, object_description: dict[str, Any]) -> str:
+        str_keys = []
+        for arg_name in sorted(object_description, key=lambda key: str(key)):
```
deepyaman, line 233: They key is already a string, right?

deepyaman:
```suggestion
        for key, value in sorted(object_description.items()):
```
feels a lot more Pythonic

Nit: `obj` is a perfectly Pythonic name for these cases, and easier than `object_description` IMO

kedro/io/core.py
```diff
@@ -227,32 +228,23 @@ def save(self, data: _DI) -> None:
...
+            if object_description[arg_name] is not None:
+                descr = pprint.pformat(
+                    object_description[arg_name],
+                    sort_dicts=False,
+                    compact=True,
+                    depth=2,
+                    width=500,
```
deepyaman, line 240:
```suggestion
                    width=float("inf"),
```
rather than random value?

kedro/io/core.py
```diff
@@ -227,32 +228,23 @@ def save(self, data: _DI) -> None:
...
-                    fmt.format(key, _to_str(value))  # 2
-                    for key, value in sorted_dict
-                    if value is not None  # 3
+    def _pretty_repr(self, object_description: dict[str, Any]) -> str:
+        str_keys = []
+        for arg_name in sorted(object_description, key=lambda key: str(key)):
+            if object_description[arg_name] is not None:
+                descr = pprint.pformat(
+                    object_description[arg_name],
+                    sort_dicts=False,
```
deepyaman, line 237:
```suggestion
```
Although, why sort the top-level key, but not subkeys?

deepyaman (changes requested):
What does the pretty repr for a `PartitionedDataset` look like? E.g. from the docs:

```yaml
# conf/base/catalog.yml

my_partitioned_dataset:
  type: partitions.PartitionedDataset
  path: s3://my-bucket-name/path/to/folder
  dataset:  # full dataset config notation
    type: pandas.CSVDataset
    load_args:
      delimiter: ","
    save_args:
      index: false
  credentials: my_credentials
  load_args:
    load_arg1: value1
    load_arg2: value2
  filepath_arg: filepath  # the argument of the dataset to pass the filepath to
  filename_suffix: ".csv"
```

datajoely:
Can you include an example of what it looks like?

kedro/io/core.py
ElenaKhaustova: I wanted to keep them in the order provided in the config, so removing top-level key sorting for consistency.

ElenaKhaustova:
> ```yaml
> my_partitioned_dataset:
>   type: partitions.PartitionedDataset
>   path: s3://my-bucket-name/path/to/folder
>   dataset:  # full dataset config notation
>     type: pandas.CSVDataset
>     load_args:
>       delimiter: ","
>     save_args:
>       index: false
>   credentials: my_credentials
>   load_args:
>     load_arg1: value1
>     load_arg2: value2
>   filepath_arg: filepath  # the argument of the dataset to pass the filepath to
>   filename_suffix: ".csv"
> ```

Here is how `_describe()` is implemented, so credentials are omitted: https://github.com/kedro-org/kedro-plugins/blob/be99fecf6cf5ac8f6a0a717c56b06dbc148b26eb/kedro-datasets/kedro_datasets/partitions/partitioned_dataset.py#L318

Screenshot 2024-07-09 at 12 23 18

ElenaKhaustova:
> Can you include an example of what it looks like?

We used one-line approach described here: https://github.com/kedro-org/kedro/issues/3980#issuecomment-2209639882

Here is the updated representation after removing sorting:
Screenshot 2024-07-09 at 12 30 27

kedro/io/core.py
ElenaKhaustova: I agree on your points but decided to keep the full path, so `__repr__` is as ambiguous as possible and one could easily understand where to look the implementation of the printed dataset.

Ideally, it would be good to have some `"short=True"` flag for the representation that can be set by user. But by default, I prefer to keep a full module name.

Curious what other think.

kedro/io/core.py
ElenaKhaustova: Two considered options: https://github.com/kedro-org/kedro/issues/3980

kedro/io/core.py
```diff
@@ -227,32 +229,23 @@ def save(self, data: _DI) -> None:
             message = f"Failed while saving data to data set {str(self)}.\n{str(exc)}"
             raise DatasetError(message) from exc
 
-    def __str__(self) -> str:
```
astrojuanlu, line 230: IIUC, we're removing the `__str__` format right? Shouldn't we keep it around for backwards compatibility?

kedro/io/core.py
deepyaman: Happy to hear what others think, but IMO we actually don't use the module name in the docs and examples. The standard way people write entries in catalog is the "short" version,  and I think that should be reflected.

deepyaman:
> > Can you include an example of what it looks like?
> 
> We used one-line approach described here: [#3980 (comment)](https://github.com/kedro-org/kedro/issues/3980#issuecomment-2209639882)
> 
> Here is the updated representation after removing sorting: Screenshot 2024-07-09 at 12 30 27

Would make more sense for `PartitionedDataset` to have a `dataset` key that is formatted like a dataset repr, for consistency with `CachedDataset`.

For `CachedDataset`, the wrapped dataset represents should also not be in quotes I think.

kedro/io/core.py
ElenaKhaustova: If `__str__` is not defined the default is to use the `__repr__`, so it shouldn't be a problem.

kedro/io/core.py
ElenaKhaustova: We can also consider adding `__str__` with a short representation, so that `__repr__` is actual representation but `__str__` adapted for printing

ElenaKhaustova:
> Would make more sense for `PartitionedDataset` to have a `dataset` key that is formatted like a dataset repr, for consistency with `CachedDataset`.

Happy to do that later on.

ElenaKhaustova:
> For `CachedDataset`, the wrapped dataset represents should also not be in quotes I think.

```
    def _describe(self) -> dict[str, Any]:
        return {
            "dataset": self._dataset._pretty_repr(self._dataset._describe()),
            "cache": self._cache._pretty_repr(self._cache._describe()),
        }
```

The idea was to keep `"dataset"` and `"cache"` representations aligned with dataset representation - `Class(arg_1=val_1, ...)` , so the values for `"dataset"` and `"cache"` keys are converted to the corresponding strings. That's why I cannot easily get rid of quotes.

kedro/io/core.py
astrojuanlu: Oh right. Still, we're changing __str__ right? Not sure if someone depends on str(dataset). it looks like a small, preventable breaking change.

ElenaKhaustova: So you mean keeping the old `__str__`?

astrojuanlu: Yes, at least until 0.20...

deepyaman:
@astrojuanlu @merelcht @noklam @datajoely would love to get your view on https://github.com/kedro-org/kedro/pull/3987#discussion_r1669430539 especially. 

Historically, I think the direct module name has been an implementation detail, and not something we surface to users (e.g. in the docs). If anything, it's the sort of thing you might stick under an underscore prefix, were we to adopt thay convention.

My personal preference would be for this aspect to remain closer to the shorter representation we have now, but would be great to get some thoughts. Think it would be good to not switch it and then change it back in another release.

datajoely:
I guess the argument in favour is so that someone can easily import class in a notebook and test things.

I sort of outline this in this prototype https://github.com/kedro-org/kedro/issues/3980#issuecomment-2217989257

noklam:
I am not a big fan of the shortcut, as this create burdens that Kedro need to be smart about guessing the import only to save some shortcut. This has more value in the past with a few different modules of dataset, but as most dataset moved to `kedro-datasets`, the main value is keep it non-breaking IMO.

I think, with the de-coupling of framework/library, it is better to not assume datasets is being used with framework. There are also side benefit of not trying to hide the import as @datajoely described.

TL;DR The standalone usage of dataset is the most important one to me and I slightly lean to the current implementation.

kedro/io/core.py
```diff
@@ -234,6 +236,8 @@ def _to_str(obj: Any, is_root: bool = False) -> str:
             formatted like Dataset(key=value).
             2. Dictionaries have the keys alphabetically sorted recursively.
             3. None values are not shown.
+
+            Will be removed in 0.20.0 release.
```
merelcht, line 240: Is this a deprecation meant to be seen by users? It should at least be mentioned in the release notes, but usually we throw a deprecation warning as well, see: https://github.com/kedro-org/kedro/blob/main/kedro/framework/cli/utils.py#L198-L200

merelcht:
I fully agree with @noklam and also prefer the longer represenation.

merelcht (approved):
I left one comment about the deprecation message, but otherwise looks great! 👍

kedro/io/core.py
ElenaKhaustova: This comment refers to a private method, so it's not for users to see but for us. I replaced it with TODO for clarity. The aim is to remove the current `__str__` implementation, so `__repr__` is used by default, so it's not a deprecation but an update.

kedro/io/core.py
astrojuanlu: > The standard way people write entries in catalog is the "short" version

Most users we've spoken to create their own datasets sooner or later, so people have a mix between short and long names in the catalog.

Also, the long name would work even for `kedro-datasets` right? The fact that we use the short one is a matter of convention and convenience. But arguably it obscures where the datasets come from, and maybe play a role in people not understanding their tracebacks (especially with `s` vs `S`).

So I'm slightly in favour of keeping the long names for all datasets to avoid introducing an exception to the logic written here, but I don't have a strong opinion.

(In either case, no matter what we do I don't think we should complicate the implementation with a flag.)

astrojuanlu (approved):
Approved, contingent on resolving the discussion on long vs short-form 👍🏼

deepyaman:
> > Would make more sense for `PartitionedDataset` to have a `dataset` key that is formatted like a dataset repr, for consistency with `CachedDataset`.
> 
> Happy to do that later on.

Can you create an issue for this, if not already done?

ElenaKhaustova:
> > > Would make more sense for `PartitionedDataset` to have a `dataset` key that is formatted like a dataset repr, for consistency with `CachedDataset`.
> > 
> > 
> > Happy to do that later on.
> 
> Can you create an issue for this, if not already done?

Issue created: https://github.com/kedro-org/kedro-plugins/issues/769

Issue #3612, opened 2024-02-09
Improve docstring for `add_feed_dict`


merelcht:
## Description
In the catalog we have a method called [`add_feed_dict`](https://github.com/kedro-org/kedro/blob/main/kedro/io/data_catalog.py#L638 ). Which is arguably one of the worst named methods in the public API. Unfortunately, renaming it is a breaking change. For `0.20.0` we should totally rename it, but in the mean time we should add a better docstring explanation so that engineers working on this code can better understand what it does.

## Context
This method was added a long time ago, likely even before Kedro was called Kedro, which git blame doesn't even know about. It was added before the catalog could be loaded from `yaml` and is supposed to describe how data can be "fed" into the catalog dictionary. The term "feed_dict" might come from clojure or tensorflow lingo, nobody really remembers 🤷

datajoely:
I would consider introducing a more intuitive name, especially with newbies in mind. We can introduce the new name as an alias to ease breaking APIs.

Some ideas of the top of my head:

- `from_objects`
- `from_dict`

astrojuanlu:
+1 to @datajoely 's idea because it also moves away from mutability

ElenaKhaustova:
Solved in https://github.com/kedro-org/kedro/pull/4009

PR #4009, opened 2024-07-15
Improve `add_feed_dict` docstrings

ElenaKhaustova:
## Description
Solves https://github.com/kedro-org/kedro/issues/3612

## Development notes

- Updated the docstrings for `DataCatalog`'s `add_feed_dict` method and extended them with an additional example.
- Created `form_dict` alias for `add_feed_dict` method and added a note to the docstring that the method will be renamed in the `0.20.0` release.

## Checklist

### DIFF
--- a/kedro/io/data_catalog.py
+++ b/kedro/io/data_catalog.py
@@ -681,11 +681,11 @@ class DataCatalog:
             self.add(name, dataset, replace)
 
     def add_feed_dict(self, feed_dict: dict[str, Any], replace: bool = False) -> None:
-        """This function adds datasets to the ``DataCatalog`` using the data provided through the `feed_dict`.
+        """Add datasets to the ``DataCatalog`` using the data provided through the `feed_dict`.
 
-        `feed_dict` dictionary key is used as a dataset name, and a value is used to create an instance of
-        ``MemoryDataset`` before adding to the ``DataCatalog`` for all the value types except of ``AbstractDataset``.
-        In the last case, the ``AbstractDataset`` is added as it is.
+        `feed_dict` keys are used as dataset names, and values can either be raw data or instances of
+        ``AbstractDataset``. In the former case, instances of ``MemoryDataset`` are automatically created before adding
+        to the ``DataCatalog``.
 
         Args:
             feed_dict: A feed dict with data to be added to the ``DataCatalog``.
--- a/kedro/io/data_catalog.py
+++ b/kedro/io/data_catalog.py
@@ -687,9 +687,6 @@ class DataCatalog:
         ``MemoryDataset`` before adding to the ``DataCatalog`` for all the value types except of ``AbstractDataset``.
         In the last case, the ``AbstractDataset`` is added as it is.
 
-        Has `from_dict` alias, so it can be called using `from_dict` name instead.
-        Will be fully renamed to `from_dict` in the `0.20.0` version.
-
         Args:
             feed_dict: A feed dict with data to be added to the ``DataCatalog``.
             replace: Specifies whether to replace an existing dataset with the same name in the ``DataCatalog``.
@@ -724,9 +721,6 @@ class DataCatalog:
                 dataset = MemoryDataset(data=feed_dict[dataset_name])
 
             self.add(dataset_name, dataset, replace)
-
-    # Alias for add_feed_dict method
-    from_dict = add_feed_dict
 
     def list(self, regex_search: str | None = None) -> list[str]:
         """
--- a/kedro/io/data_catalog.py
+++ b/kedro/io/data_catalog.py
@@ -683,12 +683,13 @@ class DataCatalog:
     def add_feed_dict(self, feed_dict: dict[str, Any], replace: bool = False) -> None:
         """Add datasets to the ``DataCatalog`` using the data provided through the `feed_dict`.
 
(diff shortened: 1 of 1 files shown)


ElenaKhaustova:
I would also suggest replacing `io` with `catalog` in the rest of the `DataCatalog`'s docstrings as the last feels more intuitive.

kedro/io/data_catalog.py
```diff
@@ -681,29 +681,41 @@ def add_all(
             self.add(name, dataset, replace)
 
     def add_feed_dict(self, feed_dict: dict[str, Any], replace: bool = False) -> None:
-        """Adds instances of ``MemoryDataset``, containing the data provided
-        through feed_dict.
+        """This function adds datasets to the ``DataCatalog`` using the data provided through the `feed_dict`.
```
astrojuanlu, line 684: I don't think we have a style guide for docstrings yet? In any case, following [Pydocstyle](http://www.pydocstyle.org/en/stable/error_codes.html) D404 and D401,

```suggestion
        """Add datasets to the ``DataCatalog`` using the data provided through the `feed_dict`.
```

kedro/io/data_catalog.py
```diff
@@ -713,6 +725,9 @@ def add_feed_dict(self, feed_dict: dict[str, Any], replace: bool = False) -> Non
 
             self.add(dataset_name, dataset, replace)
 
+    # Alias for add_feed_dict method
+    from_dict = add_feed_dict
```
astrojuanlu, line 729: Pedantic nitpick, but when I read `from_dict` I assume it's a classmethod, like

```python
class DataCatalog:
  @classmethod
  def from_dict(cls, data_dict: dict[str, AbstractDataset]):
    ...
    return cls
```

So not sure if I'd add this alias.

(I know this comes from https://github.com/kedro-org/kedro/issues/3612#issuecomment-2034961020 and that I asserted, but my response there is incomplete)

kedro/io/data_catalog.py
```diff
@@ -681,29 +681,41 @@ def add_all(
             self.add(name, dataset, replace)
 
     def add_feed_dict(self, feed_dict: dict[str, Any], replace: bool = False) -> None:
-        """Adds instances of ``MemoryDataset``, containing the data provided
-        through feed_dict.
+        """This function adds datasets to the ``DataCatalog`` using the data provided through the `feed_dict`.
+
+        `feed_dict` dictionary key is used as a dataset name, and a value is used to create an instance of
+        ``MemoryDataset`` before adding to the ``DataCatalog`` for all the value types except of ``AbstractDataset``.
+        In the last case, the ``AbstractDataset`` is added as it is.
```
astrojuanlu, line 688: What about

```suggestion
        `feed_dict` keys are used as dataset names, and values can either be raw data or instances of ``AbstractDataset``.
        In the former case, instances of ``MemoryDataset`` are automatically created before adding to the ``DataCatalog``.
```

?

kedro/io/data_catalog.py
merelcht: This is actually quite a significant change. I think we should discuss first what a good alternative is, before just adding an alias.

kedro/io/data_catalog.py
merelcht, line 684: I'm still missing a bit more explanation here about what `feed_dict` is. Adding that would make a huge difference to me.

datajoely: Add additional ?
```suggestion
        """Add additional datasets to the ``DataCatalog`` using the data provided through the `feed_dict`.
```

kedro/io/data_catalog.py
datajoely: I actually don't think mentioning `AbstractDataset` is useful here, we shouldn't really have instances of an interface right? Does the user gain anything for understanding this implementation detail?

kedro/io/data_catalog.py
datajoely: This has always been an undiscovered part of the API, I agree with @merelcht that we should think really hard / research what a better name should be.

ElenaKhaustova: I removed an alias until we are sure about adding it and agree on its name.

To @astrojuanlu's point, if we decide to keep an alias, we can use the `add_from_dict` name instead.

Or, as an alternative, we can add a new implementation (`add_from_dict` / `add_datasets`) and use it inside `add_feed_dict` with a deprecation warning.

kedro/io/data_catalog.py
ElenaKhaustova: I didn't fully understand why we need `additional` here, as we can add them to the empty catalog as well. So, it looks like we do not need to specify any adding conditions.

kedro/io/data_catalog.py
ElenaKhaustova: That's a good point. I rephrased it, but I still think it's useful to distinguish between the two cases when adding raw data and a dataset.

kedro/io/data_catalog.py
ElenaKhaustova: Done

astrojuanlu (approved):
LGTM!

ankatiyar (approved):
much clearer!

merelcht (approved):
LGTM! ⭐

Issue #3237, opened 2023-10-27
Improve Kedro run as a Package (2023)


noklam:
# Context 
I have sat down with @idanov today and try to recall our memory about https://github.com/kedro-org/kedro/pull/1423 that was made by @antonymilne. Solving this PR would make Kedro more compatible anywhere (particularly Databricks) and potentially simplify our documentation on Databricks.

#1423 summarize how Kedro run is supported currently

> A user can run their kedro project in several ways. Note that the run command executed can be defined on the framework side or overridden in turn by a plugin or a project cli.py (done by _find_run_command).
> 
> - Kedro CLI: kedro run. This is the only route that doesn't go through the project __main__.py; instead it goes through kedro.framework.main, which builds the CLI tree and does something like _find_run_command
> - CLI but not kedro CLI: python -m spaceflights. This hits the project __main__.py and will call main
> - CLI but not kedro CLI: python src/spaceflights. This is just a more unusual way of doing 2
> - Inside your own python script: from spaceflights.__main__ import main; main(), then run the script using python
> - Inside IPython/Jupyter with no kedro ipython extension: same as 4 but you run it in the notebook
> - Inside IPython/Jupyter with kedro ipython extension: same as 5 but you can also do session.run (which is what we have advertised as the way to do a kedro run in the past)
> 
> All the above must be run from within the project root or kedro won't be able to find your conf. Options 2 onwards needs you to have pip installed your project or to have src in your PYTHONPATH. Note that having the project pip installed could mean first doing kedro package and then pip install the resulting .whl file or it could mean just pip install ./src from your project root; it doesn't make a difference.

In summary, there are 3 things that #1423 attempt to fix and we can break it down.
- [x] https://github.com/kedro-org/kedro/issues/1807 summarise the 1st problem which is `click` is emitting a `sys.exit` which make it hard to integrate Kedro and causing Databricks Job "failure" despite a success run.
    - https://github.com/kedro-org/kedro/pull/3191 is a potential solution and @idanov has this PR which may solve this issue and removing the `databricks_run.py` to keep a single `__main__` entrypoint across project.
- [x] https://github.com/kedro-org/kedro/issues/3051 - The incomplete change in #1423 [`kedro/framework/project/__init__.py`](https://github.com/kedro-org/kedro/pull/1423/files#diff-51a3f12bf48820b22333e0ca633ccf09bc98ac566ecae11d6254c0cf251e5079) is trying to address this.
[...]

Issue #3051, opened 2023-09-19
Simplify Packaged Kedro Project entrypoint __main__.py, move `_find_run_command` to Framework


noklam:
## Description

`kedro new` ship a `__main__.py` which is the default way to run a packaged Kedro Project with `python -m <project>`. Currently we expose a function that loads the run command in the project. This should be move to the framework.

https://github.com/kedro-org/kedro-starters/blob/87715c48977bdfe1c64dd1d924a9f3e6c0933951/pandas-iris/%7B%7B%20cookiecutter.repo_name%20%7D%7D/src/%7B%7B%20cookiecutter.python_package%20%7D%7D/__main__.py#L11-L36

## Context

This make upgrading Kedro Project is hard because we leave code in user project.

noklam:
Cc @idanov Could you supplement what was the challenge that we need to solve to move this back to framework?

idanov:
I believe that the only challenge is to figure out what is the best and most user-friendly way to import the `run` function from `kedro.framework.project` and if we can make it behave in the same way as `kedro.framework.pipelines`, i.e. lazily loaded and configured only after `configure_project` is called before it is used.

PR #4012, opened 2024-07-16
Move `_find_run_command` from template to framework

merelcht:
## Description
https://github.com/kedro-org/kedro/issues/3051
Moved the `_find_run_command` method from the project template to the framework itself.

This is necessary refactoring to make running packaged Kedro better (https://github.com/kedro-org/kedro/issues/3237)

## Development notes
In the first attempt I moved the find run logic to `kedro.framework.project.__init__`, because that's also where `configure_project()` is. However, with that solution there are direct imports from `kedro.framework.project` to `kedro.framework.cli` which breaks the contract we've setup.
```
 from kedro.framework.cli.utils import (
        KedroCliError,
        load_entry_points,
    )
```
```
from kedro.framework.cli.project import run
```

Contract: https://github.com/kedro-org/kedro/blob/main/pyproject.toml#L159

In order not to break the import contract, I've moved the find run logic to `kedro/framework/cli/project.py`

> [!NOTE]
> The same change to `__main__.py` needs to be done on the starters side

## Checklist


kedro/framework/cli/project.py
```diff
@@ -234,3 +235,34 @@ def run(  # noqa: PLR0913
             pipeline_name=pipeline,
             namespace=namespace,
         )
+
+
+def _find_run_command(package_name: str) -> Any:  # pragma: no cover
```
noklam, line 240: Should this still be an internal method if we expect this to be user-facing function?

noklam (commented):
It looks good in general. I wonder if you found the reason why this was necessary to be implemented in the `starter` side instead of `kedro` at first place?

merelcht:
> It looks good in general. I wonder if you found the reason why this was necessary to be implemented in the `starter` side instead of `kedro` at first place?

It was implemented as part of making `cli.py` in the templates optional and the simple reason is that at the time this seemed the most logical place. When we later decided it should be moved, we never ended up prioritising because users didn't complain.

kedro/framework/cli/project.py
merelcht: Good question. Initially I called `_find_run_command()` from `run()` (https://github.com/kedro-org/kedro/pull/4012/commits/651bd2f694bc5fda3486c028acca725f18ccf2ed) which was cleaner, but now having moved the find run logic to `cli/project.py` it would clash with the existing click run command. So I guess now it makes sense to make `_find_run_command()` public.

merelcht:
> Shall we create a ticket on starters side?

Done: https://github.com/kedro-org/kedro-starters/issues/228

kedro/framework/cli/project.py
```diff
@@ -234,3 +235,34 @@ def run(  # noqa: PLR0913
...
+        _run = _find_run_command_in_plugins(plugins) if plugins else None
+        if _run:
+            # use run command from installed plugin if it exists
+            return _run
+        # use run command from `kedro.framework.cli.project`
+        return run
+    # fail badly if cli.py exists, but has no `cli` in it
+    if not hasattr(project_cli, "cli"):
+        raise KedroCliError(f"Cannot load commands from {package_name}.cli")
+    return project_cli.run
```
noklam, line 262: Can we move this block right after `project_cli`? If the code hit the `except` logic it will return eagerly and never reach here.

kedro/framework/cli/project.py
noklam: I think it makes sense to make it public.

kedro/framework/cli/project.py
merelcht: I'm not quite sure I follow, isn't it expected that if `cli.py` doesn't exist it goes to the `except` logic and never gets to the end?

kedro/framework/cli/project.py
merelcht: Done!

features/steps/test_starter/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/__main__.py
```diff
@@ -1,45 +1,16 @@
 """{{ cookiecutter.project_name }} file for ensuring the package is executable
 as `{{ cookiecutter.repo_name }}` and `python -m {{ cookiecutter.python_package }}`
 """
-import importlib
 from pathlib import Path
 
-from kedro.framework.cli.utils import KedroCliError, load_entry_points
+from kedro.framework.cli.project import find_run_command
```
idanov, line 6: This increases the API surface, so we should be mindful adding it here. If we are making it a public API, then I'd rather put it in `cli.utils`.

merelcht: Fair point. As I mentioned earlier, I didn't intent this to be public (https://github.com/kedro-org/kedro/pull/4012#discussion_r1688063853) If it's moved to `cli.utils` though I can bring back a `run` method that calls `find_run_command`.

kedro/framework/cli/project.py
noklam: I read the code incorrectly.

features/steps/test_starter/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/__main__.py
idanov:
```
from kedro.framework.cli.project import find_run_command
```
If it's used here, then it becomes public anyway. If we'd like to import the run, then maybe it makes sense to be put in `kedro.framework.cli.project`, else if it's the function `find_run_command`, it fits better under `kedro.framework.cli.utils` imo. I think both would be great improvements over the current state anyways.

merelcht: I've moved `find_run_command` to cli utils.

kedro/framework/cli/project.py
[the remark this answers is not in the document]
merelcht: It's required to have a `cli` group (https://docs.kedro.org/en/stable/development/commands_reference.html#customise-or-override-project-specific-kedro-commands) so if there's only a `run` command but no command group it will fail to load the custom commands.

kedro/framework/cli/utils.py
```diff
@@ -388,6 +389,47 @@ def load_entry_points(name: str) -> Sequence[click.MultiCommand]:
     return entry_point_commands
 
 
+def find_run_command(package_name: str) -> Any:
```
noklam, line 392:
```suggestion
def find_run_command(package_name: str) -> Callable:
```
nit

noklam (approved):
Thank you! Approved with a small comment.

ElenaKhaustova (approved):
LGTM, thank you!

PR #4014, opened 2024-07-16
`kedro-telemetry`: Improve performance by switching to after_command_run

lrcouto:

## Checklist

### DIFF
--- a/kedro/framework/cli/cli.py
+++ b/kedro/framework/cli/cli.py
@@ -5,6 +5,7 @@ This module implements commands available from the kedro CLI.
 from __future__ import annotations
 
 import importlib
+import logging
 import sys
 import traceback
 from collections import defaultdict
@@ -37,6 +38,8 @@ LOGO = rf"""
 |_|\_\___|\__,_|_|  \___/
 v{version}
 """
+
+logger = logging.getLogger(__name__)
 
 
 @click.group(context_settings=CONTEXT_SETTINGS, name="Kedro")
@@ -199,6 +202,13 @@ class KedroCLI(CommandCollection):
                 click.echo(message)
                 click.echo(hint)
             sys.exit(exc.code)
+        except Exception as error:
+            logger.error(f"An error has occurred: {error}")
+            pass
+        finally:
+            self._cli_hook_manager.hook.after_command_run(
+                project_metadata=self._metadata, command_args=args, exit_code=0
+            )
 
     @property
     def global_groups(self) -> Sequence[click.MultiCommand]:


kedro/framework/cli/cli.py
```diff
@@ -199,6 +202,13 @@ def main(
                 click.echo(message)
                 click.echo(hint)
             sys.exit(exc.code)
+        except Exception as error:
+            logger.error(f"An error has occurred: {error}")
+            pass
+        finally:
+            self._cli_hook_manager.hook.after_command_run(
+                project_metadata=self._metadata, command_args=args, exit_code=0
+            )
```
ankatiyar, line 211: I think you'll have to combine the call to `after_command_run` under the first except condition, since when the command runs fine, this hook executes twice.

kedro/framework/cli/cli.py
```diff
@@ -38,6 +39,13 @@
 v{version}
 """
 
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.ERROR)
```
noklam, line 43: This will hide `WARNING` level logs as I understand, what's the reasoning behind this?

kedro/framework/cli/cli.py
```diff
@@ -38,6 +39,13 @@
...
+handler = logging.StreamHandler(sys.stderr)
+formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
```
noklam, line 45: Is this not formatted by `kedro` already? Why is this needed?

kedro/framework/cli/cli.py
```diff
@@ -199,6 +213,19 @@ def main(
...
+            self._cli_hook_manager.hook.after_command_run(
+                project_metadata=self._metadata, command_args=args, exit_code=1
+            )
+            hook_called = True
+        finally:
+            if not hook_called:
+                self._cli_hook_manager.hook.after_command_run(
+                    project_metadata=self._metadata, command_args=args, exit_code=0
+                )
+            sys.exit(exit_code)
```
noklam, line 228: Is this `sys.exit` necessary? I think there is a high chance this will cause problem to use kedro CLI as `sys.exit` would terminate the program eagerly.

https://github.com/kedro-org/kedro/pull/4026 is trying to change `click` to `standalone=False` to make CLI

https://stackoverflow.com/questions/60780051/python-click-command-exit-flow

This [issue](https://github.com/kedro-org/kedro/issues/3237) document the previous investigation done by @antonymilne and explains more in depth.

noklam (commented):
Thanks for the PR. Could you add some notes to the PR? I think you did some testing before but I am particularly interested if we will continue to receive telemetry for long-running commands like once we move to `after_command_run`
- `kedro ipython`
- `kedro jupyter`

kedro/framework/cli/cli.py
```diff
@@ -199,6 +213,19 @@ def main(
...
             sys.exit(exc.code)
+        except Exception as error:
+            logger.error(f"An error has occurred: {error}")
+            exit_code = 1
+            self._cli_hook_manager.hook.after_command_run(
+                project_metadata=self._metadata, command_args=args, exit_code=1
+            )
+            hook_called = True
+        finally:
+            if not hook_called:
```
DimedS, line 224: Thank you for the PR, @lrcouto! Could you please explain how this is possible? It looks like `hook_called` will be set to `True` in the try block or, if it fails, in the except block, so this line will never be executed?

ElenaKhaustova: I guess that's in case of no exception

kedro/framework/cli/cli.py
lrcouto: My initial idea was that, since this would happen instead of the run stopping for an exception, it should appear only when an error occurs. But yeah it makes sense, we can lower it to a warning level.

kedro/framework/cli/cli.py
lrcouto: When I added this I had issues with the hook not being called unless this formatting was explicitly on the file, it was not using `kedro`'s. I'm doing some tests to see if I can make it just use `kedro`'s instead.

lrcouto:
Thanks for the reviews! I'm doing some tests here to try to address all of the comments.

kedro/framework/cli/cli.py
lrcouto: Ok, I recalled why I did this. Since it was not using the config from `kedro` correctly, it was putting the error messages on stdout instead of stderr.

kedro/framework/cli/cli.py
lrcouto: I've used it here as a way to ensure that it exits with the appropriate error code, because it was exiting with a 0 even when an error happened. It's not necessary to exit at this point, as long as when it exits, it's with the correct code.

kedro/framework/cli/cli.py
lrcouto: Yes, this finally block is there to solve that issue where the hook was called twice when there was not an exception.

kedro/framework/cli/cli.py
lrcouto: What I did for now moving the `sys.exit` into the block that handles the exception in case something goes wrong. So if no errors happen, click will exit normally.

ElenaKhaustova (approved):
Thank you, @lrcouto, LGTM!

kedro/framework/cli/cli.py
```diff
@@ -38,6 +39,9 @@
 v{version}
 """
 
+logger = logging.getLogger(__name__)
+logger.addHandler(logging.StreamHandler(sys.stderr))
```
DimedS, line 43: With that setting, will every logging message be sent to `stderr`? If the `INFO` logging level is configured in Kedro, will `INFO` messages also go to `stderr`, or are they not connected to Kedro's logging settings?

DimedS (approved):
Thank you, @lrcouto ! LGTM, I wrote one question about logging settings.

kedro/framework/cli/cli.py
lrcouto: It should affect only the messages generated from this instance of the logger, which is just the one error message on the main function.

lrcouto:
> Thanks for the PR. Could you add some notes to the PR? I think you did some testing before but I am particularly interested if we will continue to receive telemetry for long-running commands like once we move to `after_command_run`
> 
>     * `kedro ipython`
> 
>     * `kedro jupyter`
> 
> 
> The actual changes for telemetry to send event after command is [kedro-org/kedro-plugins#766](https://github.com/kedro-org/kedro-plugins/pull/766), so the description of this PR need to reflect the changes of this PR. I know there were some discussion offline, it would be beneficial to keep a summary here for future reference and know more about the context.

re: commands like kedro ipython or kedro jupyter, it'll send info to heap when you reload inside the shell and after you exit it:

image

Exited with ctrl+D here.

ankatiyar (approved):
tested with a few commands, LGTM!

PR #4022, opened 2024-07-20
Add 3 docstrings and specify return type in `kedro/utils.py`

yury-fedotov:
## Description
I decided to read Kedro codebase in detail to better understand how the framework works under the hood, and came across `kedro/utils.py`. I noticed that it has a few functions with quite clear objective and API that are missing docstrings, so decided to contribute their addition.

> [!NOTE]  
> Question to maintainers: are you in general supportive of PRs like that or prefer to keep control over docstrings yourself?

## Development notes
I did 2 things, hence 2 commits:

1. Add docstrings to 3 utility functions. This commit's diff is purely addition.
2. Detail return type hint of one of the functions. It used to be `Any`, but according to body, I infer that it's really a `Union[Path, None]`. So I change that and remove the `pragma: no cover` since this can now be covered by type checker I guess.

## Checklist

- [] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file


RELEASE.md
```diff
@@ -20,6 +20,7 @@
 ## Documentation changes
 * Improved documentation for configuring dataset parameters in the data catalog
 * Extended documentation with an example of logging customisation at runtime
+* Added docstrings and detailed type annotation in three private functions in `kedro/utils.py`
```
yury-fedotov, line 23: > `[Kedro.Spellings] Did you really mean 'docstrings'?`

I guess yes, per [this](https://peps.python.org/pep-0257/) PEP.

RELEASE.md
```diff
@@ -20,6 +20,7 @@
 ## Documentation changes
 * Improved documentation for configuring dataset parameters in the data catalog
 * Extended documentation with an example of logging customisation at runtime
+* Added docstrings in three private functions in `kedro/utils.py`
```
yury-fedotov, line 23: `docstrings` seem to be a valid plural: PEP [here](https://peps.python.org/pep-0257/).

kedro/utils.py
```diff
@@ -46,6 +62,17 @@ def _is_project(project_path: Union[str, Path]) -> bool:
 
 
 def _find_kedro_project(current_dir: Path) -> Any:  # pragma: no cover
```
datajoely, line 64:
```suggestion
def _find_kedro_project(current_dir: Path) -> Optional[bool]:
```

Thanks to your excellent documentation, we should do a better type hint right?

yury-fedotov: Return type here is actually this: `Union[Path, None]`.

I didn't change it in this PR because `mypy` starts complaining at another place:

```
kedro/ipython/__init__.py:173: error: Incompatible types in assignment (expression has type "Path | None", variable has type "Path")  [assignment]
```

So I decided to not resolve it here to focus this PR only on docstrings and thus simplify review.

yury-fedotov: @datajoely ok so I fixed that too, and pushed one commit on top of this branch. We can always revert if you prefer this PR do be `+++` in terms of diff and docstrings-focused only.

kedro/ipython/__init__.py
```diff
@@ -170,7 +170,7 @@ def _resolve_project_path(
         ):
             project_path = local_namespace["context"].project_path
         else:
-            project_path = _find_kedro_project(Path.cwd())
+            project_path: Path | None = _find_kedro_project(Path.cwd())
```
yury-fedotov, line 173: Without that, `mypy` complains that we assign whatever is returned from `_find_kedro_project`, which is `Path | None`, to `project_path` which is just `Path`

noklam (approved):

RELEASE.md
```diff
@@ -37,6 +37,7 @@
...
```
merelcht, line 40:
```suggestion
```

We don't add changes like these into the release notes, because they're not technically user facing changes. Especially because these docstrings have been added to private methods.

merelcht (approved):
Thanks @yury-fedotov ! I left a minor comment, but otherwise looks all good to be merged 👍

RELEASE.md
yury-fedotov: Agree, thanks @merelcht !

PR #4061, opened 2024-08-05
Fix broken tests after making telemetry a dependency

merelcht:
## Description
Fix broken hook test after releasing `0.19.7` which introduces the dependency to `kedro-telemetry`.

## Development notes
- Fix hook manager test now telemetry is included by default
- Fix checking of version in e2e tests now telemetry is included by default

## Checklist

### DIFF
--- a/tests/framework/cli/hooks/test_manager.py
+++ b/tests/framework/cli/hooks/test_manager.py
@@ -5,17 +5,18 @@ from kedro.framework.cli.hooks.specs import CLICommandSpecs
 
 
 @pytest.mark.parametrize(
-    "hook_specs,hook_name",
-    [(CLICommandSpecs, "before_command_run")],
+    "hook_specs,hook_name,hook_params",
+    [(CLICommandSpecs, "after_command_run", ("project_metadata", "command_args"))],
 )
-def test_hook_manager_can_call_hooks_defined_in_specs(hook_specs, hook_name):
+def test_hook_manager_can_call_hooks_defined_in_specs(
+    hook_specs, hook_name, hook_params
+):
     """Tests to make sure that the hook manager can call all hooks defined by specs."""
     cli_hook_manager = CLIHooksManager()
     hook = getattr(cli_hook_manager.hook, hook_name)
     assert hook.spec.namespace == hook_specs
-    # kedro-telemetry installed by default, this test runs the telemetry before_command_run hook, which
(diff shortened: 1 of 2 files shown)


tests/framework/cli/hooks/test_manager.py
```diff
@@ -5,18 +5,17 @@
...
     hook = getattr(cli_hook_manager.hook, hook_name)
     assert hook.spec.namespace == hook_specs
-    kwargs = {param: None for param in hook_params}
+    # kedro-telemetry installed by default, this test runs the telemetry before_command_run hook, which
+    # is expecting the command_args to be a list version of the command like ['kedro', 'run']
+    kwargs = {"command_args": ["kedro", "run"], "project_metadata": None}
     result = hook(**kwargs)
-    # since there hasn't been any hook implementation, the result should be empty
-    # but it shouldn't have raised
+    # since there hasn't been any hook implementation the result should be empty, but it shouldn't raise
```
ankatiyar, line 20: this comment might be outdated?

ankatiyar (approved):

tests/framework/cli/hooks/test_manager.py
merelcht: Good point. On second look I think it actually makes more sense to check a different hook, because the purpose of the test is to check that the hook manager can call all hooks defined by specs. It's not specific to before_command_run and since that is now implemented with `kedro-telemetry` it's not the best choice for this test anymore.

ankatiyar: That's fair though, I think telemetry now has an `after_command_run` hook as well which sends the heap event. Not 100% if this will cause tests to send useless data to heap 🤔

ElenaKhaustova: Confirm that telemetry uses `after_command_run` to send heap event: https://github.com/kedro-org/kedro-plugins/blob/e9878adae647656c9ca56486dbf9cf56639b82ea/kedro-telemetry/kedro_telemetry/plugin.py#L196

merelcht: Hmm okay, so I guess we don't really have any un-implemented hook specs for the `CLICommandSpecs` anymore. In that case I can revert to `before_command_run` and just update the comments in the test. We don't want to send unnecessary heap data.

merelcht: Done!

ElenaKhaustova (approved):

PR #4090, opened 2024-08-14
Enable alternative to run Kedro without Rich again.

lrcouto:
## Description
https://github.com/kedro-org/kedro/pull/3682 made our workaround to allow Kedro to be run without Rich stop working. This happened because one of the new functions implemented, `_has_rich_handler`, required importing `RichHandler`, which depended on Rich. The changes on this PR allow the workaround to function again in the same way it did before.

## Checklist

### DIFF
--- a/kedro/utils.py
+++ b/kedro/utils.py
@@ -2,9 +2,16 @@
 of kedro package.
 """
 import importlib
+import logging
 import os
+from functools import lru_cache
 from pathlib import Path
 from typing import Any, Union
+
+try:
+    from kedro.logging import RichHandler
+except ImportError:
+    pass
 
 _PYPROJECT = "pyproject.toml"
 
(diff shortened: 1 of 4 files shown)


merelcht:
Thanks @lrcouto ! I've tried it and with this fix it's possible to uninstall rich again ⭐ 

However, I think the change introduced in #3682 never really worked properly. The logic to detect the rich handler doesn't seem to work. When I have `rich` installed the orange colouring isn't there, and I added some print statements and it seems like the `RichHandler` isn't ever set. Maybe I'm missing something? @noklam @sbrugman

kedro/utils.py
```diff
@@ -78,3 +80,18 @@ def _find_kedro_project(current_dir: Path) -> Any:  # pragma: no cover
         if _is_project(parent_dir):
             return parent_dir
     return None
+
+
+@lru_cache(maxsize=None)
+def _has_rich_handler(logger: logging.Logger) -> bool:
+    """Returns true if the logger has a RichHandler attached."""
+    try:
+        from rich.logging import RichHandler
```
noklam, line 89: why do we need to explicitly import `rich` for testing?

lrcouto: My idea was checking if rich is installed at all, because if it isn't, the handler is not going to be there, making the function automatically return false. If it is installed, the user might still not be using it on their logging config, so we move to checking if it's present.

This approach is not working though, I'm verifying it right now.

lrcouto:
Closing because this issue will be solved by https://github.com/kedro-org/kedro/pull/4090

Issue #4007, opened 2024-07-12
DatasetAlreadyExistsError when using dataset factories and ThreadRunner


jaguirre16:
## Description

Kedro is trying to register same input dataset twice when used in multiple nodes and using ThreadRunner.
## Context

Im trying to use dataset factories with ThreadRunner to run independent nodes in parallel and reduce execution time. If using SecuentialRunner, pipeline works, when using ThreadRunner, DatasetAlreadyExistsError thrown.

## Steps to Reproduce
Minimal code to reproduce:
catalog.yml
````
in-{dataset_name}:
  type: "pandas.CSVDataset"
  filepath: "data/{dataset_name}.csv"

out-{dataset_name}:
  type: "pandas.CSVDataset"
  filepath: "data/{dataset_name}.csv"`
````
pipeline.py
````
 from kedro.pipeline import Pipeline, pipeline, node

def func(df):
    return df

def create_pipeline() -> Pipeline:
    return pipeline([
        node(
            func,
            inputs=["in-test1"],
            outputs="out-test2"
        ),
        node(
            func,
            inputs=["in-test1"],
            outputs="out-test3"
        ),
    ])
````

Run the project:
kedro run -r ThreadRunner

## Expected Result
Read same input on different nodes 

## Actual Result
Pipeline fails. Error log:
````
user@pc segcom % kedro run -r ThreadRunner
24/07/11 23:55:36 WARN Utils: Your hostname, pc.local resolves to a loopback address: 127.0.0.1; using 192.168.0.21 instead (on interface en0)
24/07/11 23:55:36 WARN Utils: Set SPARK_LOCAL_IP if you need to bind to another address
Setting default log level to "WARN".
To adjust logging level use sc.setLogLevel(newLevel). For SparkR, use setLogLevel(newLevel).
24/07/11 23:55:37 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /Users/user/opt/anaconda3/envs/segcom/bin/kedro:8 in <module>                                │
│                                                                                                  │
│ /Users/user/opt/anaconda3/envs/segcom/lib/python3.10/site-packages/kedro/framework/cli/cli.p │
│ y:233 in main                                                                                    │
│                                                                                                  │
│   230 │   cli_collection = KedroCLI(                                                             │
[...]

Issue #1006, opened 2021-10-29
Error to deploy spaceflights tutorial in AWS Steps Function 


andrea-cadeddu:
## Description
I tried to follow the tutorial to deploy the spaceflights template project on AWS Step Functions (https://antonymilneqb.github.io/kedro-docs/10_deployment/10_aws_step_functions.html). I follow the steps written in the tutorial but I get an error when I try to run the service. 

## Steps to Reproduce
1. python -m venv venv & source venv/bin/activate
2. pip install kedro==0.17.1 & kedro info
3. kedro new --starter=spaceflights
4. cd space (project folder)
5. create new configuration environment to prepare a compatible data catalog (conf/aws/catalog.yml) with the aws file paths
6. pip install -r src/requirements.txt
7. kedro package
8. create lambda_handler.py with a copy the code available in the tutorial and changed the row 7 in `configure_project("space")`
9. create Dockerfile with a copy the code available in the tutorial
10. push the project on ECR
11. pip install deploy_requirements.txt
12. go in the script deploy.py with a copy the code available in the tutorial and changed the rows 31 to `s3_data_bucket_name = ("<bucket-name>")` with my bucket name
13. create the file cdk.json with a copy the code available in the tutorial & run cdk deploy
14. go to aws and try to run the step function, which is deployed but it throws an error

## Expected Result
The service step function starts the execution correctly.

## Actual Result
The step functions throws an error in the first step, it seems that the lambda can't find the functions implemented.

-- Error Received.
```
{
  "errorMessage": "[Errno 38] Function not implemented",
  "errorType": "OSError",
  "requestId": "xxxxxxxxxxxxxxxxxxxxxxxxxxxx",
  "stackTrace": [
    "  File \"/home/app/lambda_handler.py\", line 7, in handler\n    configure_project(\"space\")\n",
    "  File \"/home/app/kedro/framework/project/__init__.py\", line 211, in configure_project\n    settings.configure(settings_module)\n",
    "  File \"/home/app/dynaconf/base.py\", line 223, in configure\n    self._wrapped = Settings(settings_module=settings_module, **kwargs)\n",
    "  File \"/home/app/dynaconf/base.py\", line 271, in __init__\n    self.validators.validate()\n",
    "  File \"/home/app/dynaconf/validator.py\", line 318, in validate\n    validator.validate(self.settings)\n",
[...]

PR #4093, opened 2024-08-15
Fixe using `ThreadRunner` with dataset factories

ElenaKhaustova:
## Description
Fixes https://github.com/kedro-org/kedro/issues/4007

The issue is not deterministic and appears depending on the order in which threads are executed.

The issue appears for `ThreadRunner` when loading the same pattern dataset. Since patterns are resolved at runtime, corresponding datasets are added upon execution: `load()` ->  `_get_dataset()` -> `add()`. Inside `_get_dataset()` https://github.com/kedro-org/kedro/blob/2a97dd4166ba20b5bf149cfeb2be6cebf9f5008c/kedro/io/data_catalog.py#L423 we check if dataset is in catalog and then call `add()` based on the state of the preceding check. The issue appears when first thread saved the state `dataset is not in catalog` but another thread added this dataset right after. So the first thread holds the outdated state based on which we try to add dataset to the catalog again.

This issue does not affect `ParallelRunner` because we never share objects between processes, so in each process, we have separate catalogs.

## Development notes
The main source of the issue is that `catalog.load()` does not only read but also write, and the catalog itself is not thread-safe now. So it feels like it's a major problem which we can temporarily address as suggested and make it properly when runners redesign work.

There are two solutions suggested:

1. The first solution includes adding [threading.Lock](https://docs.python.org/3/library/threading.html#lock-objects) for `load()` dataset (which does writing by calling `add()` inside) function which is applied to threads. With it, we check if `dataset in the catalog` and add dataset together as the whole `dataset.load()` is protected by the lock. The possible downside of this solution is affected performance but it should only affect the case when `load()` is done via calling an external API, because otherwise, we still cannot do simultaneous readings because of GIL.
2. The second solution includes resolving all patterns before passing the catalog to the runner, so we just avoid cases when `load()` calls `add()` as all patterns are resolved based on the pipeline datasets. This solution neglects lazy pattern resolution and will not work with dynamic pipelines.

I prefer the first solution cause it feels wrong to write to a shared object without locking it.

**Note:** if testing you need to run it multiple times to reproduce the error as it's not deterministic.

## Checklist


noklam (commented):
> The issue appears for both ThreadRunner and ParallelRunner when loading the same pattern dataset.

Does this issue also exist in `ParallelRunner`? I see all github issues are from `ThreadRunner`.

I am a bit concern about the performance penalty, the main issue here is a very simple check that causing duplicate dataset registration.

Even if we have to go with `Lock`, I think AbstractRunner is not the right place to add. See [issues](https://github.com/kedro-org/kedro/issues/1006#issuecomment-1330746265) like this, in the past we have something called `ShelveStore`, simply importing `multiprocess` will cause issues in more restricted environment.

ElenaKhaustova:
> > The issue appears for both ThreadRunner and ParallelRunner when loading the same pattern dataset.
> 
> Does this issue also exist in `ParallelRunner`? I see all github issues are from `ThreadRunner`.
> 
> I am a bit concern about the performance penalty, the main issue here is a very simple check that causing duplicate dataset registration.
> 
> Even if we have to go with `Lock`, I think AbstractRunner is not the right place to add. See [issues](https://github.com/kedro-org/kedro/issues/1006#issuecomment-1330746265) like this, in the past we have something called `ShelveStore`, simply importing `multiprocess` will cause issues in more restricted environment.

Thank you, @noklam!

1. It looked like it existed for `ParallelRunner` as it used the same method as `ThreadRunner` - `_run_node_sequential` where the problem appears. But I looked deeper into the implementation and after some tests, I can say that it `does not` affect `ParallelRunner` because we never share objects between processes, so in each process, we have separate catalogs.
2. Based on the point 1 and your comment about issues when importing `multiprocess` I replaced it with `threadinng.Lock` as `threading` module seems alright for AWS Lambda: https://docs.aws.amazon.com/prescriptive-guidance/latest/patterns/run-parallel-reads-of-s3-objects-by-using-python-in-an-aws-lambda-function.html#:~:text=Lambda%20functions%20support%20up%20to,data%2C%20in%20an%20S3%20bucket.
3. I agree that `AbstractRunner` might not be the right place for that, but given the current implementation, we can either do that or make a separate implementation of `_run_node_sequential` for `ThreadRunner`. But the main source of the issue is that `catalog.load()` does not only read but also write, and the catalog itself is not thread-safe now. So it feels like it's a major problem which we can temporarily address as suggested and make it properly when runners redesign work.
4. I see the point about the performance issues, but it feels wrong to write to a shared object without locking it. Theoretically, it should only affect the case when `load()` is done via calling an external API, because otherwise, we still cannot do simultaneous readings because of GIL. The solution to replace the dataset when registering it does not seem good as well since we will have to differ the cases when we replace it to bypass this issue and the rest.

noklam (commented):
> I see the point about the performance issues, but it feels wrong to write to a shared object without locking it. Theoretically, it should only affect the case when load() is done via calling an external API, because otherwise, we still cannot do simultaneous readings because of GIL. The solution to replace the dataset when registering it does not seem good as well since we will have to differ the cases when we replace it to bypass this issue and the rest.

I agree it's wrong to use shared object, it's definitely a bug and this is good enough for a temporary fix.

My question will be, would it make sense to resolve dataset pattern slightly earlier? We introduce dataset pattern to make `catalog.yml` simpler, this is resolved during pipeline execution. Most of the information is available once the pipeline is ready (roughly at `before_pipeline_run`). If we resolve the pattern here , it may be sufficient. (that is, of course ignoring the "dynamic pipeline" case where new dataset is being injected. But I think those case will suffer from the same problem anyway and most likely won't work with `ParallelRunner`/`ThreadRunner` from the first place.

astrojuanlu:
My worry is that, as already pointed out, both fixes touch parts of Kedro that should be unencumbered by this.

> The main source of the issue is that catalog.load() does not only read but also write, and the catalog itself is not thread-safe now.

What's the cost of keeping the issue unfixed and instead make the upcoming `DataCatalog` thread safe?

> So it feels like it's a major problem which we can temporarily address as suggested and make it properly when runners redesign work.

One thing is not clear to me: is it enough to redesign the `DataCatalog`? Or would the runners need to be redesigned as well?

ElenaKhaustova:
> My worry is that, as already pointed out, both fixes touch parts of Kedro that should be unencumbered by this.
> 
> > The main source of the issue is that catalog.load() does not only read but also write, and the catalog itself is not thread-safe now.
> 
> What's the cost of keeping the issue unfixed and instead make the upcoming `DataCatalog` thread safe?
> 
> > So it feels like it's a major problem which we can temporarily address as suggested and make it properly when runners redesign work.
> 
> One thing is not clear to me: is it enough to redesign the `DataCatalog`? Or would the runners need to be redesigned as well?

This particular issue will go in the new version if we implement this: https://github.com/kedro-org/kedro/issues/4110 (that's basically what we do in the second solution) and splitting reading from writing should address the rest of the cases.

But there might still be cases if we do some conditional writings within the catalog that require changes on the runners' side to make them safe. So, I would say that we will most probably need to redesign runners as well.

astrojuanlu:
Then my vote goes for the solution that touches the AbstractRunner and leaves the Session alone

ElenaKhaustova:
When working on DC2.0 I noted an issue second solution. Converted to draft until resolved.

ElenaKhaustova:
Solution 2 is fixed to consider only datasets from the target pipeline that match patterns.

merelcht:
I agree with @astrojuanlu and also prefer solution 1 in the runner. We should definitely take this into account when reworking the runners. I'm wondering if there's other scenarios aside from using the dataset factories that cause the same issues, but we haven't really heard about any afaik.

noklam:
Is it possible to push the `lock` to something like` _get_dataset` instead? The reason is ` `dataset.load()` takes long time and it could block the other thread. 

The actual thing that need to get protected is when dataset pattern write a new entry to `DataCatalog` right? Though it feels right that the lock should be in runner rather than `DataCatalog`

DimedS:
@ElenaKhaustova, thanks for the thorough analysis and proposals! These solutions seem temporary, and it looks like the long-term fix should come with the ongoing Catalog redesign. While Solution 1 seems like the better temporary fix, I have concerns that it could cause more harm than good. As I understand it, users rely on the ThreadRunner for handling long-term I/O operations and API calls. Introducing this lock would force the ThreadRunner to wait for the entire load operation to complete before proceeding to the next dataset, which could significantly diminish its utility and efficiency.

I suggest we leave the issue as it is temporarily and focus on the long-term solution that's currently in progress.

ElenaKhaustova:
Thank you, @astrojuanlu, @merelcht, @noklam and @DimedS, for sharing your opinion.

> Is it possible to push the lock to something like _get_dataset instead? The reason is dataset.load() takes long time and it could block the other thread.

I don't think the catalog should be aware of any external component using it or adapting to them.

I understand the concerns about performance, but it will only affect cases when reading from external sources, as local reads are not parallel with Python threads anyway. I don't know how common this case is.

I suggest keeping the solution with lock as temporal until we switch to a new catalog (the issue is not present there) or we get any evidence of performance issues from the user side.

noklam:
> I understand the concerns about performance, but it will only affect cases when reading from external sources, as local reads are not parallel with Python threads anyway. Though I don't know how common this case is.

https://docs.kedro.org/en/stable/integrations/pyspark_integration.html#tips-for-maximising-concurrency-using-threadrunner

Isn't this the case of Spark pipeline (the majority use case of ThreadRunner)?

ElenaKhaustova:
> > I understand the concerns about performance, but it will only affect cases when reading from external sources, as local reads are not parallel with Python threads anyway. Though I don't know how common this case is.
> 
> https://docs.kedro.org/en/stable/integrations/pyspark_integration.html#tips-for-maximising-concurrency-using-threadrunner
> 
> Isn't this the case of Spark pipeline (the majority use case of ThreadRunner)?

I see, the choice is:

1. Sacrifice load performance until we switch to the new catalog
2. Do nothing, so keep the issue until we switch to the new catalog
3. Apply solution 2 until we switch to the new catalog

noklam:
Sorry for not realising this earlier😅

With the new concern with Spark, my priorities are:
1. Solution 2 - Eagerly resolve pattern before pipeline.
2. Do nothing (because I assume this actually break the usage of parallelise spark pipeline with 0.19?). Is there any easy way we can confirm this?
3. Solution 1 (It should run without error, but I suspect this make it work more or less the same as `SequentialRunner`.)

I checked the telemetry but it's hard to confirm since most likely people disable telemetry on Databricks etc.
image

Cc @astrojuanlu

astrojuanlu:
Ugh... yeah difficult tradeoffs. I see this has been reported by at least 2 different users, but worsening the Spark/Databricks experience is not a great outcome either.

That, on top of the solutions being ugly already, makes me think that maybe we should indeed not do anything for now.

@noklam what are the implications of resolving

noklam:
@lrcouto Would it be possible to test this since you are working on the pipeline setup? I think this may be a good test case.

ElenaKhaustova:
Alright, let's wait until the performance tests are out and then decide on the temporal approach based on the results.

lrcouto:
I tried running a test project with spark, just to get an idea. Ran each 10 times or so.

Running it with the ThreadRunner before the changes:
- Took around 3 minutes.
- Used around 10~12% of my CPU.
- Used around 20~22% of my memory.

Running it with the ThreadRunner after the changes:
- Took around 4:30 minutes.
- Used around 12~14% of my CPU.
- Used around 23~25% of my memory.

I still want to try it again with something a little more complex, but I did perceive a small performance decrease. The test project I'm using is not super realistic and is not using any actual data, just generating and grouping 1.000.000 row dataframes to make the computer do things, so I still want to test it with something more akin to reality.

lrcouto:
I re-did the test making sure that there was nothing running in the background that could affect the results, and ran them in separate environments. The results were similar to what I've posted before.

Running 10x with regular Kedro, without the changes:
- Run times in seconds:
```
194.364, 193.176, 204.715, 196.733, 191.605, 195.560, 195.658, 198.238, 194.492, 193.692
```
- The longest run took 3m24.715s and the shortest run took 3m11.605s.
- The mean between all the run times was 195.8233 seconds (3min15.8233s).
- Standard deviation was 3.451.

Running 10x after the changes:
- Run times in seconds: 
```
266.357, 266.647, 269.864, 271.820, 264.308, 272.031, 270.103, 265.654, 261.948, 262.988
```
- The longest run took 4m32.031s and the shortest took 4m21.948s
- The mean between all the run times was 267.172 seconds (4min27.172s)
- Standard deviation was 3.425.

astrojuanlu:
```
In [1]: regular_kedro = [194.364, 193.176, 204.715, 196.733, 191.605, 195.560, 1
   ...: 95.658, 198.238, 194.492, 193.692]

In [2]: pr_4093 = [266.357, 266.647, 269.864, 271.820, 264.308, 272.031, 270.103
   ...: , 265.654, 261.948, 262.988]

In [3]: from scipy import stats

In [4]: stats.ttest_ind(regular_kedro, pr_4093)
Out[4]: TtestResult(statistic=np.float64(-44.027308662700634), pvalue=np.float64(8.798458120158517e-20), df=np.float64(18.0))
```

🤓 we have enough evidence to reject that the times didn't worsen (read that twice)

Stats jokes aside, it's still a 40 % increase in run time. Can potentially be very annoying for some users.

ElenaKhaustova:
Reverted to solution 2 after the test results for solution 1

tests/framework/session/test_session.py
```diff
@@ -693,6 +703,74 @@ def test_run(
             catalog=mock_catalog,
         )
 
+    @pytest.mark.usefixtures("mock_settings_context_class")
+    @pytest.mark.parametrize("fake_pipeline_name", [None, _FAKE_PIPELINE_NAME])
+    @pytest.mark.parametrize("match_pattern", [True, False])
+    def test_run_thread_runner(
```
merelcht, line 709: Does this test also verify that dataset factories are resolved? Or is it not possible to test that specifically because the behaviour is not deterministic?

ElenaKhaustova: It does not test that because we mock all the calls to runner, pipelines, catalog, etc, as for the rest of the session tests. It only tests that when using `ThreadRunner`, we enter into the branches to resolve the patterns.

The resolution logic is tested with the catalog tests, though it can be tested here. But for that, we'll have to properly initialise runner, pipelines, and catalog - which might be too complex given that the resolution logic is already tested separately.

merelcht: Makes sense. Thanks for clarifying! 🙂

merelcht (approved):
Thanks @ElenaKhaustova , great work investigating the issue and coming up with a suitable solution!

RELEASE.md
```diff
@@ -1,8 +1,10 @@
-# Upcoming Release
```
ankatiyar, line 1: Keep the heading here as it was, it should be changes during the release so the automatic release notes extraction works properly!

ankatiyar (approved):
Minor comment regarding the release notes but looks good! ⭐

noklam (approved):
⭐️great work on figuring out the mocking and it is great to see the pipeline testing already help us to make some decisions 🔥 @lrcouto

PR #4097, opened 2024-08-15
fix rich markup regression in 0.19.7

### LINKED ISSUE #4094: Rich markup regression in 0.19.7
# Context

The regression happened in 0.19.7.

before 0.19.7:
image

In 0.19.7:

> 
> However, I think the change introduced in #3682 never really worked properly. The logic to detect the rich handler doesn't seem to work. When I have `rich` installed the orange colouring isn't there, and I added some print statements and it seems like the `RichHandler` isn't ever set. 
> 
> _Originally posted by @merelcht in https://github.com/kedro-org/kedro/issues/4090#issuecomment-2291081005_
>

noklam:
## Description

fix #4094 

The root cause is the previous test aren't testing the correct thing. The problem is `RichHandler` is in the `root` logger but not `DataCatalog` internal one.

## Development notes

- Moved the test to `test_session.py`, as `DataCatalog` does not related to `RichHandler`. The `RichHandler` is more for Kedro project. This mean when someone use `DataCatalog` alone they won't see the markup log, but that's fine since they are not using `RichHandler` in that case.
- Remove `lru_cache`, the internal function is not used anywhere else and is not a user function. I change it to  a internal attribute to `DataCatalog`.

## Checklist


lrcouto (approved):
LGTM!

merelcht (approved):
I've tried this code and compared it to `0.19.7` and can confirm this now correctly shows highlighting again when using rich and also allows to run Kedro without `rich` installed! 👍

PR #4122, opened 2024-08-28
Exit with exit code when there's an exception

ankatiyar:

## Checklist

### DIFF
--- a/kedro/framework/cli/cli.py
+++ b/kedro/framework/cli/cli.py
@@ -210,7 +210,8 @@ class KedroCLI(CommandCollection):
                 project_metadata=self._metadata, command_args=args, exit_code=1
             )
             hook_called = True
-            raise
+            traceback.print_exception()
+            sys.exit(1)
         finally:
             if not hook_called:
                 self._cli_hook_manager.hook.after_command_run(


ankatiyar:
> Was the program not returning an 1 already in the case of failure? It should be failing both unit and e2e tests if it wasn't.

The kedro-docker tests were failing because of this https://github.com/kedro-org/kedro-plugins/issues/800. I think this branch of logic is executed when the exceptions come from outside like the kedro project itself, which I don't know if we test for.

kedro/framework/cli/cli.py
```diff
@@ -210,7 +210,8 @@ def main(
                 project_metadata=self._metadata, command_args=args, exit_code=1
             )
             hook_called = True
-            raise
+            traceback.print_exc()
+            sys.exit(1)
```
ElenaKhaustova, line 214: Now it looks like we don't need `hook_called` flag anymore as all the branches that set it to true now have `sys.exit(1)`. So maybe we can just move `self._cli_hook_manager.hook.after_command_run` to the `try` block and get rid of `finally`, so the logic is a bit cleaner?

merelcht (approved):
Tested and the behaviour is as expected and still solves #4073

kedro/framework/cli/cli.py
ElenaKhaustova: I don't want to block the PR, so we can pay attention to this in a separate one.

ankatiyar: Actually, I think you're right we don't need the `finally` block at all. If the the command executes successfully or the command doesn't exist - it goes into the first `except` block otherwise if there's an exception it is caught by the next `except`. Hook gets called either way.

ElenaKhaustova:
Tested the changes, the https://github.com/kedro-org/kedro/pull/4075 is still solved but the traceback view changed:

Before changes:
Screenshot 2024-08-30 at 17 34 43

After changes:
Screenshot 2024-08-30 at 17 34 31

kedro/framework/cli/cli.py
lrcouto: Check if it doesn't get called twice on a successful case too, it's why that `finally` block was introduced.

ankatiyar:
Did a little more investigation - the problem just turned out to be the indentation of `sys.exit()`, it was inside the `if` block so `KedroCliError` was exiting with `0`.

When there's a CLI error such as was the case with docker e2e test - https://github.com/kedro-org/kedro-plugins/issues/800 it goes into the first `except SystemExit` block and when there's any other kind of exceptions it should go into the `except Exception` block. In either cases we now have `after_command_hook` called so we don't need the `finally` block.

Please take another look @ElenaKhaustova @merelcht @lrcouto

ElenaKhaustova (approved):
Thank you, @ankatiyar!

Tested the updates - all works as expected 🚀

PR #4123, opened 2024-08-29
[DataCatalog2.0]: Move pattern resolution logic to the separate component

ElenaKhaustova:
## Description
Solves https://github.com/kedro-org/kedro/issues/4110

Relates to https://github.com/kedro-org/kedro/issues/3925

Please see the suggested order of work in [this comment](https://github.com/kedro-org/kedro/issues/3995#issuecomment-2329528221)

## Development notes
This PR includes the following:

1. Moving pattern resolution logic to a separate component - `CatalogConfigResolver`
2. Updating `DataCatalog` to use `CatalogConfigResolver` internally
3. Some refactoring to make `kedro run` command and `kedro catalog` commands work after the catalog updates
4.  `DataCatalog` interface remains the same, so no breaking changes introduced

## Developer Certificate of Origin
We need all contributions to comply with the [Developer Certificate of Origin (DCO)](https://developercertificate.org/). All commits must be signed off by including a `Signed-off-by` line in the commit message. [See our wiki for guidance](https://github.com/kedro-org/kedro/wiki/Guidelines-for-contributing-developers/).

If your PR is blocked due to unsigned commits, then you must follow the instructions under "Rebase the branch" on the GitHub Checks page for your PR. This will retroactively add the sign-off to all unsigned commits and allow the DCO check to pass.

## Checklist

- [ ] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [ ] Signed off each commit with a [Developer Certificate of Origin (DCO)](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository)
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [ ] Added tests to cover my changes
- [ ] Checked if this change will affect Kedro-Viz, and if so, communicated that with the Viz team

Issue #4110, opened 2024-08-20
[DataCatalog]: Move pattern resolution logic outside of `DataCatalog`


ElenaKhaustova:
## Description
Implement `DataCatalog` prototype with pattern resolution logic moved outside of it.

## Context
https://github.com/kedro-org/kedro/issues/3995#issuecomment-2298793761

Related to https://github.com/kedro-org/kedro/discussions/3746

## Possible Implementation

- Move pattern resolution logic to a separate component;
- Implement `KedroDataCatalog` ensuring #3925, #3926, #3916 and #3931
- Resolve configurations outside of the catalog and pass them to the catalog;

Issue #3995, opened 2024-07-08
Design `DataCatalog2.0`


merelcht:
## Description
The current `DataCatalog` in Kedro has served its purpose well but has limitations and areas for improvement identified through user research: https://github.com/kedro-org/kedro/issues/3934

As a result of `DataCatalog` user research interview we have created the list of tickets and split them into 3 categories:

1. Tickets that require low or medium effort and can be implemented for the current DataCatalog version without introducing any breaking changes. _(partially addressed)_
2. Tickets with all effort/impact levels that should be implemented for the new catalog version - DataCatalog2.0. These tickets include significant catalog redesign steps before the implementation.
3. Tickets mostly with a high effort level which didn't fall into the previous item because they are touching some significant conceptual changes or new features that require a decision on whether we want to implement them.

Addressing issues from 2. and 3. requires significant changes and the introduction of new features and concepts that go beyond the scope of incremental updates.

The objective is to design a new, robust, and modular `DataCatalog2.0` (a better name is welcomed) that incorporates feedback from the community, follows best practices, and integrates new features seamlessly.

While redesigning we plan for a smooth migration from the current `DataCatalog` to `DataCatalog2.0`, minimizing disruption for existing user.

## Context
Suggested prioritisation and tickets opened: https://github.com/kedro-org/kedro/issues/3934#issuecomment-2153342972

## Related topics
https://github.com/kedro-org/kedro-starters/tree/main/standalone-datacatalog
https://github.com/kedro-org/kedro/issues/2901
https://github.com/kedro-org/kedro/issues/2741

## Next steps

- [x] Diagram Current Design:

1. Create a detailed diagram of the current `DataCatalog` architecture.
2. Identify components, their interactions, and existing data flows.
3. Highlight pain points and limitations identified during user research.

- [x] Analyze Similar Projects:

1. Research and document how similar projects (such as `Unity Catalog`, `Polaris`, `dlthub`, other ?) address similar tasks and challenges.
2. Identify features, implementations, and best practices from these projects and map them with features requested/insights obtained during user research.
3. Analyze their architectural approaches and note down pros and cons.

- [x] Propose Initial Design:
1. Based on the research and analysis, propose an initial modular design for `DataCatalog2.0`.
2. Ensure the design addresses the tickets in group 2 and conceptual changes/new features from group 3.
3. Use the initial design as a starting point for the discussion of the redesign process with tsc.

datajoely:
[...]

PR #4124, opened 2024-08-29
[DataCatalog2.0]: Move pattern resolution logic - project cli

ElenaKhaustova:
## Description
This PR is a part of https://github.com/kedro-org/kedro/issues/4110 and is done on top of https://github.com/kedro-org/kedro/pull/4123

It serves as an example of changes required to keep both implementations together so that users can switch between them and use both, see the related [comment](https://github.com/kedro-org/kedro/issues/3995#issuecomment-2286943456)

**For the reviewers:** this PR does not include tests, please see the suggested order of work in [this comment](https://github.com/kedro-org/kedro/issues/3995#issuecomment-2319079315)

## Development notes
This PR includes modification of `context`, `session`, `runners` and `kedro run` command to be compatible with `KedroDataCatalog` and `DataCatalogConfigResolver`.

Currently, one can use old `DataCatalog` as usual without any changes or use `KedroDataCatalog` with `kedro run` by adding  the following flag:

```
kedro run --new_catalog
```

## Checklist


kedro/framework/cli/project.py
```diff
@@ -199,6 +200,7 @@ def package(metadata: ProjectMetadata) -> None:
     help=PARAMS_ARG_HELP,
     callback=_split_params,
 )
+@click.option("--new_catalog", "new_catalog", is_flag=True, help=NEW_CATALOG_ARG_HELP)
```
astrojuanlu, line 203: How does this flag play along with the `DATA_CATALOG_CLASS` defined in `settings.py`?

https://github.com/kedro-org/kedro/blob/7a16e1a48a2b1433a39b334787c534569abc4dbe/kedro/templates/project/%7B%7B%20cookiecutter.repo_name%20%7D%7D/src/%7B%7B%20cookiecutter.python_package%20%7D%7D/settings.py#L44-L46

idanov: Is this suggestion just for the PoC or to be included in the final code? Adding a temporary flag just for the new catalog has a very low API surface to benefit ratio. Removing ANY flag from the CLI is a breaking change, including this. Could we just utilise the existing methodology as pointed by @astrojuanlu ?

kedro/runner/runner.py
```diff
@@ -83,7 +83,9 @@ def run(
         """
 
         hook_or_null_manager = hook_manager or _NullPluginManager()
-        catalog = catalog.shallow_copy()
+
+        # Seems like it's not needed even for DataCatalog
+        # catalog = catalog.shallow_copy()
```
idanov, line 88: Will that stay in? Or just for the PoC?

kedro/runner/runner.py
```diff
@@ -99,9 +101,13 @@ def run(
             )
 
         # Identify MemoryDataset in the catalog
+        catalog_datasets = (
```
idanov, line 104: That's exactly the kind of thing an abstract DataCatalog (or a Protocol to this end) will prevent. Having code to check what type of instance we are working with is a bit of an anti-pattern in most cases (bar some exceptional cases emulating pattern matching). Could we somehow create a minimal interface which is used within Kedro, give it a name and program against it?

kedro/runner/runner.py
```diff
@@ -380,7 +398,7 @@ def _find_initial_node_group(pipeline: Pipeline, nodes: Iterable[Node]) -> list[
 
 def run_node(
     node: Node,
-    catalog: DataCatalog,
+    catalog: DataCatalog | KedroDataCatalog,
```
idanov, line 401: If this is for the PoC, could we (just temporarily) come up with something like:
```python
type AbstractDataCatalog = DataCatalog | KedroDataCatalog
```

And then simply use that everywhere until we design the `AbstractDataCatalog`?

idanov (changes requested):

kedro/framework/cli/project.py
ElenaKhaustova: I did this temporarily to simplify running the project with both catalogs without modifying `DATA_CATALOG_CLASS` for the project before running it.  This flag and `DATA_CATALOG_CLASS_NEW` will go away later on.

kedro/runner/runner.py
ElenaKhaustova: I should go IMO as it doesn't seem useful, but I was going to double-check why it was there

kedro/runner/runner.py
```diff
@@ -110,9 +116,10 @@ def run(
         free_outputs = pipeline.outputs() - (set(registered_ds) - memory_datasets)
 
         # Register the default dataset pattern with the catalog
```
merelcht, line 118: How does this work for the new Catalog?

ElenaKhaustova: All the patterns are managed by `DataCatalogConfigResolver` which has `dataset_patterns`, `default_pattern` and `runtime_patterns` and stores them explicitly - https://github.com/kedro-org/kedro/blob/506470aa2386d7d7d5209cde9b53e09e47c4dcf3/kedro/io/catalog_config_resolver.py#L98

Previously we used shallow copy just to add `runtime_patterns` (line that you're pointing to).  Now we have a dedicated method for that - https://github.com/kedro-org/kedro/blob/506470aa2386d7d7d5209cde9b53e09e47c4dcf3/kedro/io/catalog_config_resolver.py#L246

We add runtime patterns to the `DataCatalogConfigResolver` at the level of the session where runner is already initialised, so runner does not depend on `DataCatalogConfigResolver`.
https://github.com/kedro-org/kedro/blob/506470aa2386d7d7d5209cde9b53e09e47c4dcf3/kedro/framework/session/session.py#L409

ElenaKhaustova:
Replaced with https://github.com/kedro-org/kedro/pull/4123 and https://github.com/kedro-org/kedro/pull/4151

Issue #3931, opened 2024-06-05
[DataCatalog]: Revise methods and properties names for `DataCatalog`


ElenaKhaustova:
## Description
Users find the naming of public functions, such as `add_feed_dict,` unclear and confusing, making it difficult to understand their purpose.

We propose to:
1. Review and revise methods and properties names for `DataCatalog` to make them more descriptive and intuitive, clearly indicating their purpose and functionality.
2. Revisit our usage examples to make them more informative for users and replace `io` with `catalog`.

Relates to https://github.com/kedro-org/kedro/issues/3612

## Context

- Some user feedback: "I didn't even know what `add_feed_dict` did until someone explained it to me."
- Currently, in some of our examples, we use `io` instead of `catalog` while the last is meant.

https://github.com/kedro-org/kedro/blob/27f5405cefd6701ffac4c6243030486fb7d3c942/kedro/io/data_catalog.py#L681

![Screenshot 2024-06-05 at 18 15 46](https://github.com/kedro-org/kedro/assets/157851531/db6a4d41-66ce-4a61-81ca-6ef5a881d565)

merelcht:
We could tackle this either by changing the current `DataCatalog` function names in the next breaking release, or by making sure that the public API for a new `DataCatalog2` has better names. If we decide to build `DataCatalog2` (obviously with a better name) alongside `DataCatalog` we don't actually need to do anything with this ticket, just keep in mind we name our methods clearly 🙂

ElenaKhaustova:
Solved in https://github.com/kedro-org/kedro/pull/4218

Issue #3926, opened 2024-06-04
[DataCatalog]: Enhance `_FrozenDatasets` public API


ElenaKhaustova:
## Description
Users face challenges with understanding and effectively utilizing the `_FrozenDatasets` public API due to unclear documentation and limitations. They struggle to get dataset by name, iterate through datasets and get metadata. They express uncertainty about the advantages of using `_FrozenDatasets`, and find it unintuitive to work with due to its underscore prefix and limited functionality compared to the private API.

We propose:
1. Enhance the `FrozenDatasets` public API to provide more comprehensive functionality, including the ability to iterate over the datasets (https://github.com/kedro-org/kedro/issues/3916), access some metadata (type of dataset, type of file, filepath), and utilize methods like `get_by_name()` for flexible dataset retrieval.
2. Increase users' awareness of the `_FrozenDatasets` API through tutorials and documentation updates. Highlight the public API's capabilities and provide guidance on how to use it effectively for dataset management and retrieval.
3. Consider allowing `DataCatalog` modifications and getting rid of `_FrozenDatasets` - this is a broader question related to another issue that will be linked later.

## Context
Some quotes from the user feedback:

- "`_FrozenDataset` class is very confusing because we don't know exactly what's protected. I think the class itself starts with an underscore, so it doesn't really feel safe to loop over a `catalog.datasets` and to run into a private class. And I even don't know how to handle it whether when I use `catalog.datasets`, I think it's just a standard dictionary."
- "`_FrozenDataset` does not have the get accessor, so one cannot get dataset by name thus prefer using private `_get_datset()` method."
- "There's no straightforward way to iterate over frozen datasets like `for dataset in catalog.datasets`, so  you have to iterate via names and use private `_get_dataset()` method."

![Screenshot 2024-06-03 at 15 52 14](https://github.com/kedro-org/kedro/assets/157851531/02f5de5f-d238-4088-be23-0456a822ad33)

- "Users often don't even know that catalog api exists."
- "The public API primarily offers basic functions such as searching datasets by name and performing load and save operations. This restrictiveness often necessitates the use of private APIs to access more detailed metadata not available through the public API, so one has to break it."

![Screenshot 2024-06-04 at 23 43 09](https://github.com/kedro-org/kedro/assets/157851531/18abeaa9-95a5-4f1f-9e78-a391ad34e180)

ElenaKhaustova:
Solved in https://github.com/kedro-org/kedro/pull/4151

PR #4151, opened 2024-09-09
[DataCatalog2.0]: `KedroDataCatalog`

ElenaKhaustova:
## Description
In this PR we add a new catalog `KedroDataCatalog`, which uses the `DataCatalogConfigResolver` and addresses:

- https://github.com/kedro-org/kedro/issues/3926
- https://github.com/kedro-org/kedro/issues/3931

Please see the suggested order of work in https://github.com/kedro-org/kedro/issues/3995#issuecomment-2329528221 and comment below: https://github.com/kedro-org/kedro/pull/4151#issuecomment-2343700999

This PR is done on top of https://github.com/kedro-org/kedro/pull/4160 and relies on `CatalogProtocol`.

**For the reviewers:** this PR does not include unit-tests for `KedroDataCatalog`, they'll be added after the initial feedback.

## Development notes
- We kept some old `DataCatalog` API to avoid multiple if/else branches depending on catalog type used in `context`, `runner` and `session`
- Removed `_FrozenDatasets` and access datasets as properties
- Added get dataset by name feature: dedicated function and access by key
- Added iterate over the datasets feature
- `add_feed_dict()` simplified and renamed to `add_raw_data()`
- Datasets' init was moved out from `from_config()` method

To test `KedroDataCatalog` modify your `settings.py` and run commands as usual:
```
# settings.py

from kedro.io import KedroDataCatalog
DATA_CATALOG_CLASS = KedroDataCatalog
```

`kedro run`
`kedro catalog list/rank/resolve/create`

## Checklist


kedro/io/data_catalog_redesign.py
```diff
@@ -0,0 +1,321 @@
...
+            _format_rich(name, "dark_orange") if self._use_rich_markup else name,
+            type(dataset).__name__,
+            extra={"markup": True},
+        )
+
+        return dataset.load()
+
+    def add_from_dict(self, datasets: dict[str, Any], replace: bool = False) -> None:
+        # Consider changing to add memory datasets only, to simplify the method,
+        # adding AbstractDataset can be done via add() method
```
merelcht, line 276: Is this a todo?

kedro/io/data_catalog_redesign.py
```diff
@@ -0,0 +1,321 @@
...
+        self._logger.info(
+            "Loading data from %s (%s)...",
+            _format_rich(name, "dark_orange") if self._use_rich_markup else name,
+            type(dataset).__name__,
+            extra={"markup": True},
+        )
+
+        return dataset.load()
+
+    def add_from_dict(self, datasets: dict[str, Any], replace: bool = False) -> None:
```
merelcht, line 274: Does this need to be a public method?

kedro/runner/runner.py
```diff
@@ -94,7 +94,7 @@ def run(
 
         if unsatisfied:
             raise ValueError(
-                f"Pipeline input(s) {unsatisfied} not found in the DataCatalog"
+                f"Pipeline input(s) {unsatisfied} not found in the BaseDataCatalog"
```
merelcht, line 97: Can this instead mention the actual type of the catalog used?

kedro/io/data_catalog_redesign.py
ElenaKhaustova: That is a replacement for `add_feed_dict()`, but since we are going to modify it, I'll keep an old name for now to avoid confusion.

kedro/io/data_catalog_redesign.py
```diff
@@ -0,0 +1,321 @@
...
+)
+from kedro.io.memory_dataset import MemoryDataset
+from kedro.utils import _format_rich, _has_rich_handler
+
+
+class KedroDataCatalog:
+    def __init__(
+        self,
+        datasets: dict[str, AbstractDataset] | None = None,
+        feed_dict: dict[str, Any] | None = None,
```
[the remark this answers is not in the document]
ElenaKhaustova: Yes, good point I'll revise all TODOs.

ElenaKhaustova:
> This already looks very good. I've left some small questions. I'm also thinking that maybe it doesn't make sense to change every doc string from `DataCatalog` to `BaseDataCatalog`. Perhaps it's better to just use the word "catalog" not referring to any specific class type.
> 
> For example in a docstring like:
> 
> ```
> Any node outputs that cannot be processed by the ``BaseDataCatalog``.
> ```
> 
> This can be a bit confusing, because `BaseDataCatalog` isn't an actual class, just a Union of the existing catalog types. So instead I'd change that sentence to: "Any node outputs that cannot be processed by the catalog."

Thank you, @merelcht! I see the points regarding the confusion about `BaseDataCatalog`. We just discussed with @idanov that it probably makes sense to introduce `CatalogProtocol` before `KedroDataCatalog` so we do not add any union types that need to be removed after and use `CatalogProtocol` for typing instead. So I'm going to make the `CatalogProtocol` and typing change in a different PR preceding this. But changes made for `KedroDataCatalog` in this PR remain relevant.

ElenaKhaustova:
> > Thank you, @merelcht! I see the points regarding the confusion about `BaseDataCatalog`. We just discussed with @idanov that it probably makes sense to introduce `CatalogProtocol` before `KedroDataCatalog` so we do not add any union types that need to be removed after and use `CatalogProtocol` for typing instead. So I'm going to make the `CatalogProtocol` and typing change in a different PR preceding this. But changes made for `KedroDataCatalog` in this PR remain relevant.
> 
> That makes a lot of sense, thanks!

All done ☑️

kedro/io/kedro_data_catalog.py
```diff
@@ -0,0 +1,332 @@
...
+            pattern = re.compile(regex_search, flags=re.IGNORECASE)
+        except re.error as exc:
+            raise SyntaxError(
+                f"Invalid regular expression provided: '{regex_search}'"
+            ) from exc
+        return [ds_name for ds_name in self._datasets if pattern.search(ds_name)]
+
+    def save(self, name: str, data: Any) -> None:
+        """Save data to a registered dataset."""
+        dataset = self.get_dataset(name)
```
noklam, line 242:
```suggestion
        dataset = self[name]
```
Maybe?

kedro/io/kedro_data_catalog.py
```diff
@@ -0,0 +1,332 @@
...
+        else:
+            raise DatasetError(f"Dataset '{name}' does not have 'confirm' method")
+
+    def add_raw_data(self, data: dict[str, Any], replace: bool = False) -> None:
+        # This method was simplified to add memory datasets only, since
+        # adding AbstractDataset can be done via add() method
+        for ds_name, ds_data in data.items():
+            self.add(ds_name, MemoryDataset(data=ds_data), replace)  # type: ignore[abstract]
+
+    def add_feed_dict(self, feed_dict: dict[str, Any], replace: bool = False) -> None:
```
noklam, line 299: Are we still keeping this `feed_dict` terminology?

noklam (commented):
1. `catalog` now back to the default Python __repr__, is this by design?
image

2. One of the main goal of `FrozenDataset` is allow IDE autocompletion and protect users from overriding it. I see you using `@property` but most likely people are overriding a specific dataset instead of the whole dataset.

3. Dict-like interface, I think the current autocompletion is broken, the last time I try either `dict` or `TypeDict` seems to work well. https://github.com/kedro-org/kedro/issues/3914#issuecomment-2152858766

image

kedro/io/kedro_data_catalog.py
ElenaKhaustova: I think we should keep it as it is in case someone is trying to save an unregistered dataset. `get_dataset()` will resolve it in case it's a pattern or validate that the dataset is not in the catalog.

kedro/io/kedro_data_catalog.py
ElenaKhaustova: The terminology will change, please see the `add_raw_data()` method. `add_feed_dict()` is kept for now for compatibility with the old catalog

ElenaKhaustova:
> 1. `catalog` now back to the default Python **repr**, is this by design?
> 
> image
> 2. One of the main goal of `FrozenDataset` is allow IDE autocompletion and protect users from overriding it. I see you using `@property` but most likely people are overriding a specific dataset instead of the whole dataset.
> 3. Dict-like interface, I think the current autocompletion is broken, the last time I try either `dict` or `TypeDict` seems to work well. [[DataCatalog]: Autocompletion support for accessing datasets #3914 (comment)](https://github.com/kedro-org/kedro/issues/3914#issuecomment-2152858766)
> 
> image

Thank you, @noklam, good points.

1. Added `__repr__`
2. Based on the research findings, we decided to get rid of the `FrozenDataset` API. It was confusing for the users, and they almost never used it, preferring `_get_dataset()` instead. For now, I made `__getitem__` to return a `deep copy` for consistency with the `datasets` property. However, object assignment is not supported anyway, so users cannot override the specific dataset.
Screenshot 2024-09-16 at 11 45 42

3. Since the `FrozenDataset` API is gone completion now works only when accessing datasets like that:
Screenshot 2024-09-16 at 11 27 53

kedro/io/kedro_data_catalog.py
```diff
@@ -0,0 +1,335 @@
...
+    DatasetError,
+    DatasetNotFoundError,
+    Version,
+    generate_timestamp,
+)
+from kedro.io.memory_dataset import MemoryDataset
+from kedro.utils import _format_rich, _has_rich_handler
+
+
+class KedroDataCatalog:
```
merelcht, line 30: Should this be: `class KedroDataCatalog(CatalogProtocol)` ?

ElenaKhaustova: We only need it in case we share some logic between the implementations which we intentionally don't want to do to keep all the implementations independent from the protocol class.

> while it’s possible to subclass a protocol explicitly, it’s not necessary to do so for the sake of type-checking

https://peps.python.org/pep-0544/#explicitly-declaring-implementation

noklam (commented):
> Based on the research findings, we decided to get rid of the FrozenDataset API. It was confusing for the users, and they almost never used it, preferring _get_dataset() instead. For now, I made __getitem__ to return a deep copy for consistency with the datasets property. However, object assignment is not supported anyway, so users cannot override the specific dataset.

I guess it will be hard to prevent user assigned something that they are not gonna access later. We have similar issues previously where user try to assign something to `catalog` but find out it doesn't change anything because Kedro run always creates its own `catalog.

For deep-copy, would it works for datasets that are not "picklable" or dataset that comes with connection?




I left a comment in the other [PR](https://github.com/kedro-org/kedro/pull/4160#pullrequestreview-2304302289), I still found `CatalogProtocol` a bit weird but I am not sure what's convention did the Python community adopted, something like `CatalogLike`?

ElenaKhaustova:
> I guess it will be hard to prevent user assigned something that they are not gonna access later. We have similar issues previously where user try to assign something to `catalog` but find out it doesn't change anything because Kedro run always creates its own `catalog.
> 
> For deep-copy, would it works for datasets that are not "picklable" or dataset that comes with connection?

These two points totally make sense to me. However, `_FrozenDataset` didn't solve any of those problems as people were using `_get_dataset()`, which returns the reference to the dataset. Based on your second point, I returned to reference as well, so we do not require "pickable" datasets, though as far as I understand, `ParallelRunner` does.

ElenaKhaustova:
> I left a comment in the other [PR](https://github.com/kedro-org/kedro/pull/4160#pullrequestreview-2304302289), I still found `CatalogProtocol` a bit weird but I am not sure what's convention did the Python community adopted, something like `CatalogLike`?

As for naming convention, I see different examples: https://typing.readthedocs.io/en/latest/spec/protocol.html
But if looking at the existing ones they named like `SupportsDataCatalog`: https://docs.python.org/3/library/typing.html#protocols

So maybe the last is the better naming

kedro/io/kedro_data_catalog.py
```diff
@@ -0,0 +1,326 @@
...
+            raw_data: A dictionary with data to be added in memory as `MemoryDataset`` instances.
+                Keys represent dataset names and the values are raw data.
+            config_resolver: An instance of CatalogConfigResolver to resolve dataset patterns and configurations.
+            load_versions: A mapping between data set names and versions
+                to load. Has no effect on data sets without enabled versioning.
+            save_version: Version string to be used for ``save`` operations
+                by all data sets with enabled versioning. It must: a) be a
+                case-insensitive string that conforms with operating system
+                filename limitations, b) always return the latest version when
+                sorted in lexicographical order..
```
yury-fedotov, line 57:
```suggestion
                sorted in lexicographical order.
```

kedro/io/kedro_data_catalog.py
```diff
@@ -0,0 +1,326 @@
...
+
+        if raw_data:
+            self.add_raw_data(raw_data)
+
+    @property
+    def datasets(self) -> dict[str, Any]:
+        return copy.copy(self._datasets)
+
+    @datasets.setter
+    def datasets(self, value: Any) -> None:
```
yury-fedotov, line 77: I think that's a good use case for [NoReturn](https://peps.python.org/pep-0484/#the-noreturn-type)?

```suggestion
    def datasets(self, value: Any) -> NoReturn:
```

kedro/io/kedro_data_catalog.py
```diff
@@ -0,0 +1,326 @@
...
+            self.add(ds_name, MemoryDataset(data=ds_data), replace)  # type: ignore[abstract]
+
+    def add_feed_dict(self, feed_dict: dict[str, Any], replace: bool = False) -> None:
+        # TODO: remove when removing old catalog
+        return self.add_raw_data(feed_dict, replace)
+
+    def shallow_copy(
+        self, extra_dataset_patterns: Patterns | None = None
+    ) -> KedroDataCatalog:
+        # TODO: remove when old catalog
```
yury-fedotov, line 300: Based on other comments, it's probably:

```suggestion
        # TODO: remove when removing old catalog
```

kedro/io/kedro_data_catalog.py
```diff
@@ -0,0 +1,326 @@
...
+        ds = AbstractDataset.from_config(
+            ds_name,
+            ds_config,
+            self._load_versions.get(ds_name),
+            self._save_version,
+        )
+
+        self.add(ds_name, ds)
+
+    def get_dataset(
```
merelcht, line 165: Can you add a doc string to this method?

kedro/io/kedro_data_catalog.py
```diff
@@ -0,0 +1,326 @@
...
+    @staticmethod
+    def _validate_dataset_config(ds_name: str, ds_config: Any) -> None:
+        if not isinstance(ds_config, dict):
+            raise DatasetError(
+                f"Catalog entry '{ds_name}' is not a valid dataset configuration. "
+                "\nHint: If this catalog entry is intended for variable interpolation, "
+                "make sure that the key is preceded by an underscore."
+            )
+
+    def _add_from_config(self, ds_name: str, ds_config: dict[str, Any]) -> None:
```
merelcht, line 152: Perhaps name this `_add_dataset_from_config` to make it specific that this is calling the `AbstractDataset.from_config`

merelcht (commented):
I've done a first review and added (mostly nit) comments. I'll do another review and have a proper look at the tests tomorrow.

kedro/io/kedro_data_catalog.py
```diff
@@ -0,0 +1,326 @@
...
+        # TODO: Add lazy loading feature to store the configuration but not to init actual dataset
+        # TODO: Initialise actual dataset when load or save
```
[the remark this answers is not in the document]
ElenaKhaustova: Yep, all TODOs left are for later

kedro/io/kedro_data_catalog.py
ElenaKhaustova: Since we have `add()` method and `add_all()` (for `DataCatalog`) that add datasets without specifying that in their names I ended with the current naming. We only specify the type of data for `add_feed_dict` / `add_raw_data`.

I don't insist on the current naming though but it looks more aligned with what we already have.

kedro/io/kedro_data_catalog.py
```diff
@@ -0,0 +1,345 @@
...
+        """
+        self._logger.info("Confirming dataset '%s'", name)
+        dataset = self.get_dataset(name)
+
+        if hasattr(dataset, "confirm"):
+            dataset.confirm()
+        else:
+            raise DatasetError(f"Dataset '{name}' does not have 'confirm' method")
+
+    def add_raw_data(self, data: dict[str, Any], replace: bool = False) -> None:
```
idanov, line 306: Still unsure about the naming of this... Could we try some other names?

- `inject_data`
- `add_data`
- `load_data`
- `prepare_data`
- `put_data`
- `add_raw_data`

kedro/io/kedro_data_catalog.py
```diff
@@ -0,0 +1,345 @@
...
+                in the DatasetNotFoundError message.
+
+        Returns:
+            An instance of AbstractDataset.
+
+        Raises:
+            DatasetNotFoundError: When a dataset with the given name
+                is not in the collection and do not match patterns.
+        """
+        ds_config = self._config_resolver.resolve_dataset_pattern(ds_name)
```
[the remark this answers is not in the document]
ElenaKhaustova: Yeah, the point was to prevent you from complaining about nested ifs 😅 I moved it inside the condition now.

kedro/io/kedro_data_catalog.py
```diff
@@ -0,0 +1,345 @@
...
+
+        if ds_name not in self._datasets and ds_config:
+            self._add_from_config(ds_name, ds_config)
+
+        dataset = self._datasets.get(ds_name, None)
+
+        if dataset is None:
```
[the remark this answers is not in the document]
ElenaKhaustova: Well, we can only fail after we try to resolve. Otherwise, you get one more layer of `if` as the logic needs to go inside the `if fail [] else []` scenario.

Now the logic is like this:

- if not part of the materialised datasets, resolve the dataset pattern
- if resolved, add from config
- get the dataset
- if the dataset does not exist (basically if it cannot be resolved nor exists), go with the error scenario
- otherwise, continue with a non-error scenario

docs/source/conf.py
```diff
@@ -134,6 +134,7 @@
         "kedro.io.core.DatasetError",
         "kedro.io.core.Version",
         "kedro.io.data_catalog.DataCatalog",
+        "kedro.io.kedro_data_catalog.KedroDataCatalog",
```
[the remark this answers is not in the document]
ElenaKhaustova: Do you mean moving `KedroDataCatalog ` class into `data_catalog` instead of creating a new file?

kedro/io/kedro_data_catalog.py
ElenaKhaustova: I don't mind, but the point was to clearly show that we do not have a shared logic and addition of new catalog does not require an explicit declaration.

kedro/io/kedro_data_catalog.py
```diff
@@ -0,0 +1,345 @@
...
+            self.add_data(raw_data)
+
+    @property
+    def datasets(self) -> dict[str, Any]:
+        return copy.copy(self._datasets)
+
+    @datasets.setter
+    def datasets(self, value: Any) -> None:
+        raise AttributeError(
+            "Operation not allowed! Please use KedroDataCatalog.add() instead."
```
merelcht, line 79: Nit, but the exclamation mark seems a bit dramatic 😆 

```suggestion
            "Operation not allowed. Please use KedroDataCatalog.add() instead."
```

tests/io/test_kedro_data_catalog.py
```diff
@@ -0,0 +1,658 @@
...
+        def test_config_import_kedro_datasets(self, correct_config, mocker):
+            """Test kedro_datasets default path to the dataset class"""
+            # Spy _load_obj because kedro_datasets is not installed and we can't import it.
+
+            import kedro.io.core
+
+            spy = mocker.spy(kedro.io.core, "_load_obj")
+            parse_dataset_definition(correct_config["catalog"]["boats"])
+            for prefix, call_args in zip(_DEFAULT_PACKAGES, spy.call_args_list):
+                # In Python 3.7 call_args.args is not available thus we access the call
```
merelcht, line 337: This mentions python 3.7 which we haven't supported in a while. Is this still needed?

tests/io/test_kedro_data_catalog.py
```diff
@@ -0,0 +1,658 @@
...
+                # arguments with less meaningful index.
+                # The 1st index returns a tuple, the 2nd index return the name of module.
+                assert call_args[0][0] == f"{prefix}pandas.CSVDataset"
+
+        def test_config_import_extras(self, correct_config):
+            """Test kedro_datasets default path to the dataset class"""
+            correct_config["catalog"]["boats"]["type"] = "pandas.CSVDataset"
+            assert KedroDataCatalog.from_config(**correct_config)
```
merelcht, line 345: This looks like an old test when datasets still lived in "extras", can we remove it?

tests/io/test_kedro_data_catalog.py
```diff
@@ -0,0 +1,658 @@
...
+                "'version' attribute removed from data set configuration since it "
+                "is a reserved word and cannot be directly specified"
+            )
+            assert log_record.levelname == "WARNING"
+            assert expected_log_message in log_record.message
+
+        def test_from_correct_config_load_versions_warn(self, correct_config):
+            correct_config["catalog"]["boats"]["versioned"] = True
+            version = generate_timestamp()
+            load_version = {"non-boart": version}
```
merelcht, line 591: Nit: but this looks like it's always been a typo ⛵ 
```suggestion
            load_version = {"non-boat": version}
```

tests/io/test_kedro_data_catalog.py
```diff
@@ -0,0 +1,658 @@
...
+            pattern = (
+                r"\'load_versions\' keys \[non-boart\] are not found in the catalog\."
```
merelcht, line 593:
```suggestion
                r"\'load_versions\' keys \[non-boat\] are not found in the catalog\."
```

tests/io/test_kedro_data_catalog.py
```diff
@@ -0,0 +1,658 @@
...
+            )
+            with pytest.raises(DatasetNotFoundError, match=pattern):
+                KedroDataCatalog.from_config(
+                    **correct_config, load_versions=load_version
+                )
+
+        def test_compare_tracking_and_other_dataset_versioned(
+            self, correct_config_with_tracking_ds, dummy_dataframe
+        ):
+            """Test saving of tracking data sets from config results in the same
```
merelcht, line 603:
```suggestion
            """Test saving of tracking datasets from config results in the same
```

merelcht (approved):
Some final small comments, but otherwise I'm very happy with how this looks! Great work @ElenaKhaustova ⭐ 🌟

docs/source/conf.py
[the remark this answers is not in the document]
ElenaKhaustova: If yes, I'd prefer to keep them separately. Otherwise, `kedro.io.data_catalog` becomes really long, and it's hard to compare two implementations. Since we import both from `kedro.io`, there shouldn't be any problems when removing one of them later.

kedro/io/kedro_data_catalog.py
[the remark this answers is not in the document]
idanov: No, it was just a question, either is fine.

kedro/io/kedro_data_catalog.py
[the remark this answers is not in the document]
idanov: What I meant is that we should have as a first line something like:

```
if ds_name not in self._datasets and self._config_resolver.match_pattern():
    ...
```

and then continue with the error scenario, and then go on with everything else. It'll be much easier to follow this way. Btw while checking if this is possible, I saw a problem in the resolver - it can fail even if it matches, but that should not happen. 

https://github.com/kedro-org/kedro/blob/5147dfb24d8f1468daadacb8140406cb13485e8a/kedro/io/catalog_config_resolver.py#L149-L156

This ☝️ should have been checked at the config_resolver init time, basically we should not allow to create a config_resolver with unresolvable configs or add invalid configs that cannot be resolved.

Also there's other changes like e.g. `resolve_dataset_pattern` should be just `resolve_pattern` similar to all other public methods there, which never include the word `dataset` (rightfully). Hopefully the resolver API is not released yet, so we can change it now.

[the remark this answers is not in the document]
ElenaKhaustova: 1. Checking for a pattern match is not enough, as the config could already be resolved at the init time. `resolve_pattern` method encapsulates this, not to expose this logic outside the `config_resolver`. So we ask the `config_resolver` to provide a config for a pattern without bothering about how it's happening inside. I don't think we need to move any resolution logic (including any specific checks) to the catalog level.
2. The configs resolution makes sense to me. We can move this validation to the init time and simplify the resolution method. But would do that in a separate PR as it doesn't touch the catalog and will be done at the level of the config resolver.
3. Method was renamed.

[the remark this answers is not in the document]
idanov: 1. Why wouldn't it be enough? We are checking only for failing scenarios, what other failing scenarios would there be apart from not matching a concrete dataset or a pattern? Could we expect failure of instantiation?
2. ✅ 
3. ✅ 

In any case, this is a minor thing, let's merge it in as it is and then we can always come back and simplify it.

idanov (approved):

PR #4159, opened 2024-09-12
Fix template projects example tests

Manezki:
## Description
New project created by `kedro new --tools test` has example test that does not work.

## Development notes
* Changed ConfigLoader to the newly embraced OmegaConfigLoader.
* Added env argument to KedroContext initialization.

After the changes a new project can successfully run tests with `pytest .`

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -7,6 +7,7 @@
 * Enhanced `OmegaConfigLoader` configuration validation to detect duplicate keys at all parameter levels, ensuring comprehensive nested key checking.
 ## Bug fixes and other changes
 * Fixed bug where using dataset factories breaks with `ThreadRunner`.
+* Fixed template projects example tests.
 
 ## Breaking changes to the API
 * Removed `ShelveStore` to address a security vulnerability.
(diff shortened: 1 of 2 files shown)


merelcht (approved):
Thank you so much @Manezki ! 🌟 I can't believe we didn't spot that earlier 🙈

ankatiyar (approved):
Thanks @Manezki!

Manezki:
Thanks for the snappy review (and the positive comments).

I don't have permission to merge the PR. Is there still something I should do?
For the remaining "open points" I have the following:

* The failure of the language linter is a false negative on my alias.
* I haven't checked the Kedro-Viz impact, as I have no idea about the surface I would have to cover - but intuitively it would make sense that there's no impact.
* I haven't added a test-case for the error as that would probably require calling pytest from the tests. I'm not sure if the test harness would support that. Should I still look into it?

merelcht:
> Thanks for the snappy review (and the positive comments).
> 
> I don't have permission to merge the PR. Is there still something I should do? 

No, I can merge it for you 👍 

For the remaining "open points" I have the following:
> 
> * The failure of the language linter is a false negative on my alias.

This is more a suggestion, but in this case we can ignore it. 

> * I haven't checked the Kedro-Viz impact, as I have no idea about the surface I would have to cover - but intuitively it would make sense that there's no impact.

Indeed, no impact on Viz for this change.

> * I haven't added a test-case for the error as that would probably require calling pytest from the tests. I'm not sure if the test harness would support that. Should I still look into it?

Issue #4138, opened 2024-09-02
[DataCatalog2.0]: Implement `Protocol` abstraction for `DataCatalog`


ElenaKhaustova:
## Description
Implement `Protocol` abstraction for `DataCatalog` - `CatalogProtocol`

## Context
The logic behind: https://github.com/kedro-org/kedro/issues/3995#issuecomment-2298793761 and https://github.com/kedro-org/kedro/pull/4151#issuecomment-2343700999

As a part of this work pattern resolution logic was moved outside of the catalog: https://github.com/kedro-org/kedro/issues/4110

The full work planned: https://github.com/kedro-org/kedro/issues/3995#issuecomment-2319079315

PR #4160, opened 2024-09-12
[DataCatalog2.0]: `Protocol` abstraction for `DataCatalog`

ElenaKhaustova:
## Description
Solves https://github.com/kedro-org/kedro/issues/4138

This introduces the `Protocol` abstraction for the current `DataCatalog`, which will be used to add new catalog implementations. Currently, it fully relies on the existing `DataCatalog` implementation to avoid breaking changes.

## Development notes
Protocol class was excluded from test coverage since its method placeholders were marked as untested, see the related issue: https://github.com/nedbat/coveragepy/issues/1616

## Checklist


kedro/framework/context/context.py
```diff
@@ -178,13 +178,13 @@ class KedroContext:
     )
 
     @property
-    def catalog(self) -> DataCatalog:
-        """Read-only property referring to Kedro's ``DataCatalog`` for this context.
+    def catalog(self) -> CatalogProtocol:
```
noklam, line 181: Question: Is this how type hint supposed to be done with Protocol? I roughly understand how Protocol works like traits/interface but I haven't seen much in a real codebase.

noklam (commented):
My original understanding this is going to be a separate Catalog, but I see this PR has a lot of changes to existing API's typehint so I am slightly confused.

image

Does type hint suppose to work here? As it seems like it doesn't understand the catalog at all as all the IDE features seems broken.

kedro/framework/context/context.py
ElenaKhaustova: Yep, cause this is type in the end, see examples here: https://peps.python.org/pep-0544/#protocol-members

ElenaKhaustova:
> My original understanding this is going to be a separate Catalog, but I see this PR has a lot of changes to existing API's typehint so I am slightly confused.
> 
> image
> Does type hint suppose to work here? As it seems like it doesn't understand the catalog at all as all the IDE features seems broken.

Hmm, it seems working well in `PyCharm` and `mypy` do not complain as well.
Screenshot 2024-09-13 at 17 21 53

The new catalog is added in the following PR. But first we need this to fit two implementations together. Here is a context regarding this decision: https://github.com/kedro-org/kedro/pull/4151#issuecomment-2343700999

noklam (commented):
Overall looks good to me. I find `CatalogProtocol` slightly weird, I don't know what's the convention is for Python, I found a list here: https://github.com/python/typeshed/issues/4174

noklam (approved):
approved as non-blocking commenet

kedro/io/core.py
```diff
@@ -871,3 +881,70 @@ def validate_on_forbidden_chars(**kwargs: Any) -> None:
             raise DatasetError(
                 f"Neither white-space nor semicolon are allowed in '{key}'."
             )
+
+
+_C = TypeVar("_C")
```
merelcht, line 886: What is this for?

kedro/io/core.py
```diff
@@ -871,3 +881,70 @@ def validate_on_forbidden_chars(**kwargs: Any) -> None:
...
+        ...
+
+    @property
+    def config_resolver(self) -> CatalogConfigResolver:
+        """Return a copy of the datasets dictionary."""
+        ...
+
+    @classmethod
+    def from_config(cls, catalog: dict[str, dict[str, Any]] | None) -> _C:
+        """Create a ``KedroDataCatalog`` instance from configuration."""
```
merelcht, line 904:
```suggestion
        """Create a catalog instance from configuration."""
```

kedro/io/core.py
```diff
@@ -871,3 +881,70 @@ def validate_on_forbidden_chars(**kwargs: Any) -> None:
...
+        ...
+
+    def _get_dataset(
```
merelcht, line 907: Is it necessary to add all methods implemented by `DataCatalog` to the protocol? I thought with a protocol it's also possible to not implement all defined methods and add others?

merelcht (commented):
This mostly looks good, but I left a comment about the methods added here that will be removed in the future.

kedro/io/core.py
ElenaKhaustova: This type parameter for the catalog which is needed to define a generic protocol later in the code since `from_config()` and `shallow_copy()` return catalog type.

https://peps.python.org/pep-0544/#generic-protocols

kedro/io/core.py
ElenaKhaustova: I only added methods exposed to the framework components, so it is possible to use catalogs implementing this protocol with session, context, etc. So this is the bare minimum needed to make both catalogs work with the current implementation.

If I skip some of them, then we accept catalogs that do not implement some of the required methods, so they fail when used within the current framework implementation.

The same goes for adding methods — we can add them, but then all the catalogs should implement them, which is not needed for now.

merelcht: Ah right, that makes total sense. Thanks for clarifying!

merelcht (approved):
Don't forget to add this to the release notes ✍️ 

Otherwise, looks all good to me!

PR #4186, opened 2024-09-23
Revert `_local_exists` to mirror `fsspec` behavior

deepyaman:
## Description
The current implementation of `_local_exists` is different from how the `exists` method exposed by `fsspec` implementations works, in that it checks parent directories. This prevents the code path from reaching `NotADirectoryError` (e.g. if you try saving a versioned dataset on top of an unversioned one), resulting in a different/inconsistent error message from one of the `fsspec`-based dataset implementations.

## Development notes
This essentially reverts a very old change in https://github.com/kedro-org/kedro/commit/b0581f1d1276592b1dfe39896399a87cd44b3664, which predates `fsspec` use.

Existing dataset implementations (usually?) define their own `exists_function` using `fsspec`, so this doesn't normally get hit.

I ran into this while trying to implement basic versioning of `ibis.FileDataset`, because the `test_versioning_existing_dataset` was failing with a different error message than one would expect (due to the above note on `NotADirectoryError`).

## Checklist


noklam:
Thanks for catching this, do we know why it was implemented in that way? That surely looks a bit weird to me and the change looks correct.

deepyaman:
> Thanks for catching this, do we know why it was implemented in that way? That surely looks a bit weird to me and the change looks correct.

I can't really figure out, beyond the commit I linked above; maybe you can figure out from the QB-internal Jira, if KED-715 still exists?

noklam:
## Versioning failed if the same dataset name already exists in the file path directory

### Acceptance criteria

Documentation about what happens when you switch on versioning after you've run the pipeline, let users know that it will throw an error because the file path already exists

[If possible] Better error message in AbstractDataSet to show you how to resolve the problem i.e. delete the file

noklam:
Found a bit more at [here](https://github.com/McK-Private/private-kedro/pull/240)

> 
> Motivation and Context
> Original issue: quantumblack/asset-kernel-ai#683
> 
> What has been done:
> 
> _local_exists function in kedro/io/core.py now also checks for the existence of any files within filepath parents. This allows us to throw a DataSetError if versioned dataset attempts to save after unversioned one has already been saved.
> I did not change the [error message](https://github.com/quantumblacklabs/private-kedro/blob/c757d433cc9086254a11783edccbeb5cacbc913a/kedro/io/core.py#L505) since it applies not only to local datasets, but to S3 too, for example. S3 doesn't have this issue - it allows having a file and a 'folder' with the same name. WDYT?


- This allows us to throw a DataSetError if versioned dataset attempts to save after unversioned one has already been saved.

I think this is the main reason, but as far as I remember this doesn't seem to work nicely when a versioned dataset blocked by an existing unversion dataset.

merelcht (approved):
Since the old code pre-dates the use of `fsspec` and all tests pass, I'm happy with merging this in 👍

noklam (approved):
I didn't find anything alarming in the old issues/PR as well, let's trust our tests and if it breaks we know what regression tests are needed.

Issue #4191, opened 2024-09-24
[BUG] DataCatalog problem with ThreadRunner on kedro >=0.19.7


andresrq-mckinsey:
## Description

After kedro >= 0.19.7, there was a change introduced into the DataCatalog on PR #3990 with the addition of `__repr__` into the `_FrozenDatasets`. This is causing an `error: dictionary changed size during iteration` when using the `ThreadRunner` on some scenarios that handle big volumes of data.

## Context

So, there are 3 components that are acting to generate this error:
- The `pluggy/_tracing.py` has an inherit `logger.debug` that prints the DataCatalog
  - This always executes, even if the logger level is not debug
  - Runs for every node (thus, every created thread by the ThreadRunner) as it happens in `after_node_run [hook]` log from pluggy
  - Output example:
     ```
      2024-09-20 16:52:20,439 [DEBUG] - /databricks/python/lib/python3.10/site-packages/pluggy/_tracing.py @ kedro.framework.hooks.manager:34 23365 139967147775552 -       after_node_run [hook]
                node: some_func([input_dataset]) -> [output_dataset]
                catalog: {'dataset_name': "<dataset_class>", ...}
     ```
- MemoryDatasets not previously declared in the `catalog.yml` are added into the DataCatalog `self.datasets` property once the node finishes, to keep track of them
- ThreadRunner runs every node on a thread

The flow is as follows:

1. When using a ThreadRunner, every node is run in a thread, and the `before_node_run`, `on_node_error` and `after_node_run` get to also run in parallel.
2. After a node finishes, the debug logging that inherently happens on `after_node_run` from the `_tracing.py` will try to print the entire dataset object (see func `__repr__` from DataCatalog), which triggers the print (`__repr__`) of the `_FrozenDatasets`. 
3. Finally, the output dataset, if not previously declared in the `catalog.yml`, will be added into the DataCatalog as a MemoryDataset during the dataset saving process (runner.py:530 in `_run_node_sequential`: `catalog.save(name, data)` -> data_catalog.py:579 in `save`: `dataset = self._get_dataset(name)` -> data_catalog.py:453 in `_get_dataset`: `self.add(dataset_name, dataset)`)

[...]

PR #4210, opened 2024-10-07
Add runner benchmark

### LINKED ISSUE #4127: [Stress Testing] - Runners
## Description

The goal for this is test different runners,

In Kedro, there are 3 runners:
- SequentialRunner
- ThreadRunner (I/O bound, usually used for Spark pipeline)
- ParallelRunner (CPU bound)

The test should measure:
- Time
- Memory ([Memory Profiling](https://github.com/kedro-org/kedro/discussions/4111))

The test should be target for the runner, for example we need to test I/O heavy pipeline for `ThreadRunner`

## Context

https://github.com/kedro-org/kedro/issues/3957#issuecomment-2298861806

> ## Component stress test:
> - The main goal for this is to benchmark performance of individual component, this will inform if refactoring work has positive/negative impact. Currently we only check if test pass, so we have no idea if a change may slow down performance. We have done this in the past but usually ad-hoc basis, we should run this regularly (or at least per release).
> 
> The direction of this is simple, we want to make measure the change of time against # number of entries. We would start with Datasets and Catalog, as this fits in the `DataCatalog2.0` work and will be immediately useful.
> 
> 
> This can address:
> - https://github.com/kedro-org/kedro/discussions/4111

Should work with the setup decided in #4128

noklam:
## Description

Close #4127 

## Dev Notes:
- Rename `benchmarks` due to name clash with pyarrow: https://github.com/apache/arrow/issues/44398
- Add runner tests for peak memory, memory and time
- For time, each runner get a specific workload (IO/compute bound)
- `benchmarks` has been move to `kedro_benchmarks`, despite best effort to avoid installation of `pyarrow`, it will still causes a weird `benchmarks` module not found error. I try many different things, but ultimately I found that it works as long as the folder have any name other than `benchmarks. It's a bit strange but also not worth the effort to further investigate, so I end up renaming the folder.

## QA notes

pip install asv or `pip install -e ".[benchmark]"`
`asv run --quick`

## More notes
During test I discover `ThreadRunner` throws error like this occassionally, it's likely the same issue of https://github.com/kedro-org/kedro/issues/4191
> DatasetAlreadyExistsError: Dataset 'dummy_1' has already been registered

## Checklist


noklam:
> ============================================================Done============================================================
{'SequentialRunner': 7.8415021896362305, 'ThreadRunner': 7.56311297416687, 'ParallelRunner': 3.4262261390686035}


At the moment I only running test for a compute-bound workload, the result is expected only `ParallelRunner` should speed things up. I run into issue sometimes with dataset already registered error. I think it's likely related to: https://github.com/kedro-org/kedro/issues/4191

@ElenaKhaustova any thought? Should I focus on testing only the `KedroDataCatalog` instead?

kedro_benchmarks/pyproject.toml
```diff
@@ -0,0 +1,15 @@
+[project]
+name = "kedro_benchmarks"
```
noklam, line 2: The reason to do this is `multiprocessing` breaks imports. Without this I get a `benchmarks` module is not found error.

To make sure this always in `sys.path`, I add this into part of the installation step. Other suggestions are welcomed.

I have tried `PYTHONPATH` as well but it doesn't seem to work as new process doesn't inherit it.

kedro_benchmarks/benchmark_runner.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def time_thread_runner(self):
+        """IO bound pipeline"""
+        catalog = create_data_catalog()
+        test_pipeline = create_io_bound_pipeline()
+        runner_obj = ThreadRunner()
+        runner_obj.run(test_pipeline, catalog=catalog)
+
+
+if __name__ == "__main__":
```
ankatiyar, line 156: Is the main part needed here? I think the tests work without it, the params thing is handled by asv

noklam: removed - I have it for testing initially.

asv.conf.json
```diff
@@ -3,10 +3,16 @@
     "project": "Kedro",
     "project_url": "https://kedro.org/",
     "repo": ".",
-    "install_command": ["pip install -e ."],
-    "branches": ["main"],
+    "install_command": [
+        "pip install -e .  kedro-datasets[pandas-csvdataset]"
+    ],
+    "branches": [
+        "noklam/stress-testing-runners-4127"
```
noklam, line 10: Need to revert before merge.

kedro_benchmarks/README.md
```diff
@@ -0,0 +1,21 @@
...
+checkout the [official documentation](https://asv.readthedocs.io/en/stable/using.html#benchmarking)
+
+For example, `asv run main..mybranch` will run benchmark against every single commits since branching off from
+`main`.
+
+## Compare benchmark for two commits:
+Run this in the terminal:
+`asv compare v0.1 v0.2`
+
+This run benchmark against two different commits
```
merelcht, line 21:
```suggestion
This runs benchmark against two different commits.
```

merelcht (approved):
This looks good 👍 

One question I have is why the `SequentialRunner` and `ParallelRunner` are tested with the compute bound pipeline and `ThreadRunner` with io bound?

kedro_benchmarks/README.md
noklam: The special case is `ThreadRunner`, which was designed for I/O bound task so it doesn't make sense to test it with compute bound tasks (This is somewhat unique in Python since the GIL prevent thread doing anything useful for computation)

kedro_benchmarks/benchmark_runner.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def setup(self, *args, **kwargs):
+        data_dir = Path("benchmarks/data")
+        data_dir.mkdir(exist_ok=True, parents=True)
+
+        # Create a dummy csv
+        with open(data_dir / "data.csv", "w") as f:
+            f.write("col1,col2\n1,2\n")
+
+    def mem_runners(self, runner):
```
ElenaKhaustova, line 112: Shall we also test memory usage on various catalogs and actually return some objects that need sufficient memory rather than a "`dummy`" string?

ElenaKhaustova:
> This looks good 👍
> 
> One question I have is why the `SequentialRunner` and `ParallelRunner` are tested with the compute bound pipeline and `ThreadRunner` with io bound?

Maybe it makes sense to add time tests for each runner and each scenario (compute bound, io bound) so we can compare not only catalogs (when changes are applied) but also runners (in general, between each other).

noklam:
> > This looks good 👍
> > One question I have is why the `SequentialRunner` and `ParallelRunner` are tested with the compute bound pipeline and `ThreadRunner` with io bound?
> 
> Maybe it makes sense to add time tests for each runner and each scenario (compute bound, io bound) so we can compare not only catalogs (when changes are applied) but also runners (in general, between each other).

We can do that. I didn't do it because I think we don't get much for running `ThreadRunner` for compute bound task (maybe things will change with 3.13 but we also need to enable that flag explicitly). These tests are more costly as they takes some time to run.

I can add these tests, and if these tests start taking too long then I will suggest to remove the less useful one.

kedro_benchmarks/benchmark_runner.py
noklam: Do you mean just changing what it returns? We can do that but I don't know if it changes anything. Do I also need to change the structure of the pipeline?

Most data are not stored in `DataCatalog` unless `CacheDataset` is used. From my understanding they only get loaded within the node function and get recycled as soon as node is finished.

ElenaKhaustova: I mean varying the catalog and number of entries there.

Makes sense about returns, thank you.

noklam: Do you mean adding also the `KedroDataCatalog`? I am not sure what do you want to test by varying the number of entries, can you elaborate more?

noklam: @ElenaKhaustova I see, https://github.com/kedro-org/kedro/blob/main/benchmarks/benchmark_kedrodatacatalog.py, I think this is covered in the catalog test already?

ElenaKhaustova (approved):
Thank you, @noklam, LGTM!

There's a suggestion on possible extension of tests for memory profiling, but that can be done in the separate PR if decide to do that.

Issue #4175, opened 2024-09-18
[DataCatalog2.0]: Extend `KedroDataCatalog` with dict interface


ElenaKhaustova:
## Description
As a continuation of a work on `DataCatalog` redesign (https://github.com/kedro-org/kedro/issues/3995#issuecomment-2329528221) we suggest simplifying the interface for `KedroDataCatalog` and make it similar to [UserDict](https://docs.python.org/3/library/collections.html#collections.UserDict)

The reasoning: this way we want to make the catalog functioning as a collection of datasets with an interface familiar to users.

Related tickets that will be addressed:

- https://github.com/kedro-org/kedro/issues/3916
- https://github.com/kedro-org/kedro/issues/3931

## Context
[](https://github.com/kedro-org/kedro/pull/4151)

## Possible Implementation
Change the `KedroDataCatalog` API to support `__iter__`, `__getitem__`, `__setitem__`, `keys()`, `items()` similar to the dict.

Replace old API such as `save()`, `load()`, `list()`, etc.

The first iteration can be done without breaking changes if keep old interface and reuse newly added methods.

datajoely:
I'm actually less excited about this than a lot of the other proposed ideas for this redesign. 

What about putting it inside a property or something like a `.dict_like()` method

ElenaKhaustova:
> I'm actually less excited about this than a lot of the other proposed ideas for this redesign.
> 
> What about putting it inside a property or something like a `.dict_like()` method?

That's an attempt to align on the interface changes before introducing new features, so it doesn't cancel the rest of the redesign idea. But since this is a proposal of a possible interface, it will be helpful at this stage if you can expand your thoughts about why it feels not good.

noklam:
Related idea discussed in the past: https://github.com/kedro-org/kedro/issues/3914#issuecomment-2152697336

astrojuanlu:
We discussed this on backlog grooming. To clarify, the user impact of this one is that rather than doing

```
>>> catalog.list()
>>> catalog.load("my_dataset")
>>> catalog.save("my_dataset", data)
>>> for dataset in catalog.datasets: pass
```

the user would do

```
>>> list(catalog)  # __iter__
>>> catalog["my_dataset"]  # __getitem__
>>> catalog["my_dataset"] = data  # __setitem__
>>> for dataset in catalog: pass  # __iter__
```

[...]

Issue #3917, opened 2024-06-03
[DataCatalog]: Add functionality to search datasets in the catalog


ElenaKhaustova:
## Description
Users struggle to find datasets within the catalog, particularly when dealing with a large number of datasets. They express the need for search features to facilitate dataset discovery.

## Context
"As a user in my list object, I can filter by name but I can't filter by what. So it would be good to be able to say give me all the `sql` datasets and then the names of the tables that are attached."

Comment form @astrojuanlu: Kedro Viz has an item in their roadmap to include a table view of all the metadata, could help with this.

## Possible Implementation
 Integrate search functionality into the catalog, enabling users to search for datasets based on keywords, patterns and by kind. Include support for regex search to accommodate users with advanced search requirements.

datajoely:
Also search by kind - if I wanted to find all Parquet files today I'd have to get very creative. Retrieving paths associated with thos would be super complicated.

astrojuanlu:
@stephkaiser do you remember if we already opened an issue or discussion about the "metadata table view"?

(cc @rashidakanchwala for when you're back)

yury-fedotov:
> ... enabling users to search for datasets based on keywords or patterns...

Isn't it what `catalog.list()` already does?

IICR if you do e.g. `catalog.load("compani")`, it would raise an error with `did you mean one of ["companies", "processed_companies"]?`

astrojuanlu:
Related #3312

astrojuanlu:
But 

> Integrate search functionality into the catalog, enabling users to search for datasets based on keywords, patterns and by kind. Include support for regex search to accommodate users with advanced search requirements.

this is a bit more advanced, I'd say

merelcht:
When you say "search datasets in the catalog", what workflow are you talking about? Is this inside a notebook, on the CLI, directly in the IDE or on Kedro-Viz? Each of these user flows might have a different preferred solution.

ElenaKhaustova:
After the discussion at backlog grooming, we've decided to:
1. Revisit the docs on filtering with regex - to check if they're clear enough (as this feature exists but users were not able to find it);
2. Revisit our filtering approach and decide on whether we want to extend it for example with "search by kind" method/ other (?)

Galileo-Galilei:
(Just dropping a comment now for when this is ready to be tackled properly)

When implementing the dict like interface in #4218, several possibilities were proposed to filter on the values with a regex, ordered after by order of apparition in the PR: 
[...]

PR #4218, opened 2024-10-08
[DataCatalog2.0]: `KedroDataCatalog` with dict interface

ElenaKhaustova:
## Description
Solves https://github.com/kedro-org/kedro/issues/4175

## Development notes
See https://github.com/kedro-org/kedro/issues/4175#issuecomment-2400972641

## Checklist


kedro/io/kedro_data_catalog.py
```diff
@@ -112,6 +114,50 @@ def __eq__(self, other) -> bool:  # type: ignore[no-untyped-def]
             other.config_resolver.list_patterns(),
         )
 
+    def keys(self, regex_search: str | None = None) -> List[str]:  # noqa: UP006
+        return self._filter_keys(regex_search)
+
+    def values(self, regex_search: str | None = None) -> List[AbstractDataset]:  # noqa: UP006
```
noklam, line 120:
```suggestion
    def values(self, regex_search: str | None = None) -> list[AbstractDataset]:
```

noklam: You can use `list` directly

noklam (approved):

kedro/io/kedro_data_catalog.py
ElenaKhaustova: I had to change it to the current method because we have `list()` method defined for catalog and linter thinks you annotate using it rather than built-in `list`

Screenshot 2024-10-09 at 11 11 43

Screenshot 2024-10-09 at 11 14 43

kedro/io/kedro_data_catalog.py
```diff
@@ -112,6 +114,50 @@ def __eq__(self, other) -> bool:  # type: ignore[no-untyped-def]
...
+
+    def items(
+        self, regex_search: str | None = None
+    ) -> List[tuple[str, AbstractDataset]]:  # noqa: UP006
+        return [(key, self._datasets[key]) for key in self._filter_keys(regex_search)]
+
+    def __iter__(self) -> str:
+        yield from self._datasets.keys()
+
+    def __getitem__(self, ds_name: str) -> AbstractDataset | None:
```
ankatiyar, line 131: I think this should still error `catalog["non_existent"]` while `catalog.get("non_existent")` returns `None`

ankatiyar (approved):
Played around with the interface and everything works well, just had a small comment about the `__getitem__` method! 💯

kedro/io/kedro_data_catalog.py
ElenaKhaustova: Yeah, agree

kedro/io/kedro_data_catalog.py
```diff
@@ -112,6 +114,50 @@ def __eq__(self, other) -> bool:  # type: ignore[no-untyped-def]
             other.config_resolver.list_patterns(),
         )
 
+    def keys(self, regex_search: str | None = None) -> List[str]:  # noqa: UP006
+        return self._filter_keys(regex_search)
+
+    def values(self, regex_search: str | None = None) -> List[AbstractDataset]:  # noqa: UP006
+        return [self._datasets[key] for key in self._filter_keys(regex_search)]
```
Galileo-Galilei, line 121: Just a nitpick, because the syntax is neat,  but to be honest I can't think about any use case that would need to handle values (i.e. datasets objects) without the associated key (i.e. the dataset name). 

Say you have ``catalog.values(regex_search="^country_")`` :  it is very likely that I want to know the exact data (hence its name) I am loading / reassigning / whatever modification I do, and not just iterate over all datasets  matching the criteria. 

A much better filter search would be something specific to the dataset itself (e.g. its type, some metadata it has...), but I definitely understand it is out of scope here. I am not sure adding other filtering is better than just letting people do something like this : 

```python
[ds_name for ds_name in catalog if isinstance(catalog[ds_name], PandasDataset)] 
```

kedro/io/kedro_data_catalog.py
```diff
@@ -231,30 +273,31 @@ def _get_dataset(
...
+    def _filter_keys(self, regex_search: str | None) -> List[str]:  # noqa: UP006
         if regex_search is None:
             return list(self._datasets.keys())
 
-        if not regex_search.strip():
-            self._logger.warning("The empty string will not match any datasets")
-            return []
-
         try:
             pattern = re.compile(regex_search, flags=re.IGNORECASE)
```
Galileo-Galilei, line 302: if ``regex_search`` is forced to be a string, should we add ``regex_kwargs`` everywhere to handle these flags?

kedro/io/kedro_data_catalog.py
```diff
@@ -264,6 +307,11 @@ def list(self, regex_search: str | None = None) -> list[str]:
         return [ds_name for ds_name in self._datasets if pattern.search(ds_name)]
 
     def save(self, name: str, data: Any) -> None:
+        # TODO: remove when removing old catalog
+        """Save data to a registered dataset."""
+        self.save_data(name, data)
+
+    def save_data(self, name: str, data: Any) -> None:
```
Galileo-Galilei, line 314: I am not sure of what is the rationale of keeping ``load_data`` and ``save_data``, but I see this was decided in a TD. Could ou juste reexplain the expected benefit? (I suspect that we keep ``load_data`` to deal with version ; and we keep ``save_data`` for "symmetry")

Galileo-Galilei (approved):

kedro/io/kedro_data_catalog.py
```diff
@@ -112,6 +114,50 @@ def __eq__(self, other) -> bool:  # type: ignore[no-untyped-def]
...
+    def __getitem__(self, ds_name: str) -> AbstractDataset | None:
+        return self.get_dataset(ds_name)
+
+    def __setitem__(self, key: str, value: Any) -> None:
+        if key in self._datasets:
+            self._logger.warning("Replacing dataset '%s'", key)
+        if isinstance(value, AbstractDataset):
+            self._datasets[key] = value
+        else:
+            self._datasets[key] = MemoryDataset(data=value)  # type: ignore[abstract]
```
Galileo-Galilei, line 140: I really (really) think we should have at least logger.info() here to explain this behaviour. I expect silent error due to people assigning a malformed dataset manually which end up assigned as a MemoryDataset, opening a room for a lot of headache.

kedro/io/kedro_data_catalog.py
```diff
@@ -231,30 +273,31 @@ def _get_dataset(
     def add(
```
Galileo-Galilei, line 273: Can (should?) we add a method to remove a dataset? Either ``pop()`` or ``del`` (I saw elsewhere you did not ant to maintain all dict like methods, but I think this may be useful). I think it could be useful when one copy a catalog, filter a pipeline and just want to get some datasets associated to the filtered pipeline.

merelcht (approved):
This looks very neat! I think after this change it might be a good moment to add some documentation about using `KedroDataCatalog`. Of course, we'll make it clear it's still in "beta", but I think this will help the team itself get familiar with it and it's behaviour.

kedro/io/kedro_data_catalog.py
ElenaKhaustova: That totally makes sense.

We have a ticket on improving search options - https://github.com/kedro-org/kedro/issues/3917, so there's a high chance it will be modified in the future anyway. We also discussed whether we should provide any options for filtering since now it will be easy for users to apply it on top of the keys returned as in your example.

Currently, we kept it as part of the previous API - `list()` method. I added it for `values()` to align with `keys()` and `items()`. However, I see it leading to confusion, so I will remove it from the interface for `values()`.

kedro/io/kedro_data_catalog.py
ElenaKhaustova: I don't mind adding it, though previously, we didn't log it in `add_feed_dict()`, and there were no common issues with that. However, it could be because not many people used this method.

kedro/io/kedro_data_catalog.py
ElenaKhaustova: You're right about `load_data,` and we are keeping both for now to ensure compatibility with the old catalog. Version will be probably moved from the catalog methods interface soon and technically we won't need `load_data` and `save_data` as we can replace them with `catalog["dataset_name"].load()` and `catalog["dataset_name"].save()`.

kedro/io/kedro_data_catalog.py
ElenaKhaustova: We didn't plan to add them in the first version, but we can consider doing so if we see user demand. The main reason for not adding them now is that we still don't want to make catalog fully mutable.

kedro/io/kedro_data_catalog.py
ElenaKhaustova: We had one more discussion on this and decided that we will probably keep `load` and `save` at the catalog level, so framework components interact with the datasets via catalog abstraction.

ElenaKhaustova:
After initial feedback on the interface we removed `regex_search` argument from `keys()`, `values()` and `items()` as it differs from the standard `dict` interface and looked not intuitive to users. We added `filter()` method instead for compatibility with the old `list()` method.

kedro/io/kedro_data_catalog.py
Galileo-Galilei: Yes, I understand that we want to avoid users to modify the catalog inplace in a hook. As a user I'd assume I could handle it like a dict, but this deleting a key prgramatically seems a very rare use case.

kedro/io/kedro_data_catalog.py
Galileo-Galilei: Wow, I did not know it was already possible ; however as you say I think we target notebooks users with this design, so we can expect it to happen more often that it does now.

kedro/io/kedro_data_catalog.py
Galileo-Galilei: Yes, I was confused by the renaming but it makes more sense like this 👍

kedro/io/kedro_data_catalog.py
```diff
@@ -231,40 +272,77 @@ def _get_dataset(
...
+            return self.keys()
 
-        if not regex_search.strip():
+        if regex_search == "":
             self._logger.warning("The empty string will not match any datasets")
             return []
 
+        return self.filter(regex_search)
+
+    def filter(
```
idanov, line 299: Why I am usually in favour of simpler names, here `filter` is a bit ambiguous, since it's unclear what we are filtering exactly. If we are to keep a separate method, I'd rather keep `list()` with the regex.

kedro/io/kedro_data_catalog.py
```diff
@@ -231,40 +272,77 @@ def _get_dataset(
...
         except re.error as exc:
             raise SyntaxError(
                 f"Invalid regular expression provided: '{regex_search}'"
             ) from exc
-        return [ds_name for ds_name in self._datasets if pattern.search(ds_name)]
+        return [ds_name for ds_name in self.__iter__() if pattern.search(ds_name)]
 
     def save(self, name: str, data: Any) -> None:
-        """Save data to a registered dataset."""
+        # TODO: rename input argument when breaking change: name -> ds_name
```
idanov, line 317: We can still keep it a name, it's not too bad - this will avoid the breaking change.

kedro/io/kedro_data_catalog.py
```diff
@@ -277,7 +355,35 @@ def save(self, name: str, data: Any) -> None:
         dataset.save(data)
 
     def load(self, name: str, version: str | None = None) -> Any:
-        """Loads a registered dataset."""
+        # TODO: rename input argument when breaking change: name -> ds_name
```
idanov, line 358: Same as for save, let's keep it `name`.

kedro/io/kedro_data_catalog.py
```diff
@@ -277,7 +355,35 @@ def save(self, name: str, data: Any) -> None:
...
+        # TODO: remove version from input arguments when breaking change
```
idanov, line 359: 👏

kedro/io/kedro_data_catalog.py
```diff
@@ -112,6 +114,49 @@ def __eq__(self, other) -> bool:  # type: ignore[no-untyped-def]
             other.config_resolver.list_patterns(),
         )
 
+    def keys(self) -> List[str]:  # noqa: UP006
+        return list(self.__iter__())
```
idanov, line 118:
```suggestion
        return list(self)
```
Would this work?

kedro/io/kedro_data_catalog.py
ElenaKhaustova: We agreed to rename it for consistency with arguments for other methods.

kedro/io/kedro_data_catalog.py
ElenaKhaustova: We agreed to rename it for consistency with arguments for other methods.

kedro/io/kedro_data_catalog.py
ElenaKhaustova: Done

kedro/io/kedro_data_catalog.py
ElenaKhaustova: I made it for iteration cases, like

```
    def values(self) -> List[AbstractDataset]:  # noqa: UP006
        return [self._datasets[key] for key in self]
```

kedro/io/kedro_data_catalog.py
Galileo-Galilei: I quite disagree with this one, because "list" already existed in previous version and virtually no one new about it being able to filter with regex as shown in the interview, so I'd argue that's not helping discovery. 

I personally find `filter` more explicit, or even the original `.values(regex=...)`. What do other think?

ElenaKhaustova: @Galileo-Galilei we agreed to rename it based on the solution for https://github.com/kedro-org/kedro/issues/3917, so we don't rename it several times. That's why we kept the old name for now.

PR #4231, opened 2024-10-15
Add Github Actions workflow to trigger pipeline performance test

lrcouto:
## Description

Adds a workflow to trigger a pipeline performance test on the [performance test repo](https://github.com/kedro-org/pipeline-performance-test) if the user adds the "performance" label to the pull request they wish to test.

The test will run a comparison in elapsed time between the branch with the performance label against the latest kedro release.

## Checklist


ankatiyar:
My initial thought is that this should be triggered not on every push or based on the commit message because someone might want to trigger this _after_ making all the changes so making a random commit just to trigger a pipeline test seems counterintuitive. 
Alternatives - Adding a "performance" label to the PR: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request
WDYT?

lrcouto:
> My initial thought is that this should be triggered not on every push or based on the commit message because someone might want to trigger this _after_ making all the changes so making a random commit just to trigger a pipeline test seems counterintuitive. Alternatives - Adding a "performance" label to the PR: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request WDYT?

I like the idea!

noklam:
Agree on above, I like the idea about label better. How can we trigger it multiple times with label? Perhaps is it possible to trigger the workflow manually with parameter (commit hash/branch name?)

.github/workflows/pipeline-performance-test.yml
```diff
@@ -0,0 +1,39 @@
...
+      run: |
+        branch_name=${GITHUB_REF#refs/heads/}
+        echo "branch_name=$branch_name" >> $GITHUB_ENV
+
+    - name: Check if commit contains 'PERFORMANCE TEST'
+      if: contains(env.commit_message, 'PERFORMANCE TEST')
+      run: |
+        echo "Commit message contains 'PERFORMANCE TEST'. Triggering performance test."
+
+    - name: Trigger performance test workflow in test project
```
ankatiyar, line 32: Also I think you can use the `checkout` action to clone the performance test repo - https://github.com/actions/checkout

lrcouto:
> Agree on above, I like the idea about label better. How can we trigger it multiple times with label? Perhaps is it possible to trigger the workflow manually with parameter (commit hash/branch name?)

Maybe we can have an alternative way to trigger from the test repo itself with a specific branch.

lrcouto:
I've changed the trigger method to use a label, and created a "performance" label to test it out. It will trigger when the label is added, and you can repeat the test by removing the label and adding it again or running the job again.

Now for a little bit of a philosophical question - My intention was to use commits not only to trigger the test itself but also to set those hook/save/load delay parameters that the test has. If we use a label instead of the commit to trigger the run, how could we set the delays? Maybe we could have both methods available, so you can use the label for a simple test but a commit if you want to manually alter the delays?

merelcht:
> I've changed the trigger method to use a label, and created a "performance" label to test it out. It will trigger when the label is added, and you can repeat the test by removing the label and adding it again or running the job again.

I really like the label trigger! 

> Now for a little bit of a philosophical question - My intention was to use commits not only to trigger the test itself but also to set those hook/save/load delay parameters that the test has. If we use a label instead of the commit to trigger the run, how could we set the delays? Maybe we could have both methods available, so you can use the label for a simple test but a commit if you want to manually alter the delays?
> 
> This is the test branch I'm using, by the way: #4233

@lrcouto How do you set those parameters in the commit? And what are the default now?

lrcouto:
> @lrcouto How do you set those parameters in the commit? And what are the default now?

The default right now was no delays. My initial idea was passing the parameter line on the commit message and sending it on the POST payload, same as it's being done with the branch name right now. But I'm looking for alternatives that would work with the label solution.

lrcouto:
> > @lrcouto How do you set those parameters in the commit? And what are the default now?
> 
> The default right now was no delays. My initial idea was passing the parameter line on the commit message and sending it on the POST payload, same as it's being done with the branch name right now. But I'm looking for alternatives that would work with the label solution.

I thought of two possible alternatives for this issue that we could use. What I would like to do is allow the user to pass all of the parameters to `kedro run`, so not only you can use the delay values specific to the test project but you could also, for example, set it to run with a specific runner.

The first option would be using PR comments to pass parameters and run the workflow. It's easy to use but the disadvantage is that we would pollute the PRs with parameter comments, which makes me not like this option very much.

The second would be to have a manually dispatched workflow on the test repo, in which you could go and input the parameters. Then you could use the performance label one for a quick run with all default values, or you could do something with more customization by going to the test repo itself and running it manually. I'm leaning more towards this one.

Let me know what you think!

merelcht:
Maybe for a first iteration just triggering the test as is, is fine and we can create a follow up ticket to explore parameters? #4210 is already meant to check the different runners, so my question is whether it makes sense to also have that as a changing factor for this setup?

noklam:
Agree with @merelcht, I think having parameters in commit are also a bit overkill and quite easy to break. I would prefer a sensible default (that are non-zero, otherwise it doesn't shows anything). If there are more needs to extend the parameters, we can always clone the repository and run with source code directly.

lrcouto:
We can leave this as it is for now, and extend it in the future then.

merelcht (approved):
LGTM! 👍

ankatiyar:
I do think you could checkout the other repo in this workflow and time and run it so one could see the results in the PR itself instead of triggering the workflow on the repo that contains the project where we would have to look through the Actions tab to see what the results are.

.github/workflows/pipeline-performance-test.yml
noklam: Has this been addressed?

noklam (commented):
This looks good to me besides some comments left by @ankatiyar that haven't been addressed.

lrcouto:
> I do think you could checkout the other repo in this workflow and time and run it so one could see the results in the PR itself instead of triggering the workflow on the repo that contains the project where we would have to look through the Actions tab to see what the results are.

Yeah I think that would be more convenient. I'm rewriting the workflow to work in this way.

.github/workflows/pipeline-performance-test.yml
```diff
@@ -0,0 +1,79 @@
...
+      run: echo "Performance label detected. Running performance test."
+
+    - name: Clone test repo
+      if: github.event.action == 'labeled' && contains(github.event.label.name, 'performance')
+      run: |
+        git clone https://x-access-token:${{ secrets.GH_TAGGING_TOKEN }}@github.com/kedro-org/pipeline-performance-test.git
+
+    - name: Set up Python 3.11
+      if: github.event.action == 'labeled' && contains(github.event.label.name, 'performance')
+      uses: actions/setup-python@v4
```
ankatiyar, line 23: This should be v5 I think there's some deprecations in v4

ankatiyar (approved):
LGTM! Thanks @lrcouto 💯

Issue #4235, opened 2024-10-17
0.19.9 introduced error, output is saved only once when running the pipeline


adobrogows-ki:
## Description
 output is saved only once when running the pipeline, testing with sequential runner

## Context
fails testing where we need to run a pipeline many times and our server that relies on re-running pipeline many times

## Steps to Reproduce
1. make new project with kedro new and sample code
2. modify the test case `test_data_science_pipeline` to end with:
```
    a = SequentialRunner().run(pipeline, catalog)
    b = SequentialRunner().run(pipeline, catalog)

    assert a == b
    assert successful_run_msg in caplog.text
```
3. run the test 

## Expected Result
a == b

## Actual Result
b is empty

## Your Environment
* Kedro version used 0.19.9

Calychas:
https://kedro-org.slack.com/archives/C03RKP2LW64/p1729156896234549

astrojuanlu:
Adding the public link instead for archival purposes https://kedro.hall.community/support-lY6wDVhxGXNY/second-runner-run-fails-to-save-output-after-pipeline-upgrade-V5X74llZ8xD1

noklam:
I cannot reproduce the issue following the description, I can produce the same error with following:
```python
def test_data_science_pipeline(caplog, dummy_data, dummy_parameters):

    pipeline = (
        create_ds_pipeline()
        .from_nodes("split_data_node")
        .to_nodes("train_model_node")
    )
    catalog = DataCatalog()
    catalog.add_feed_dict(
        {
            "model_input_table" : dummy_data,
            "params:model_options": dummy_parameters["model_options"],
        }
    )
    
    a = SequentialRunner().run(pipeline, catalog)
    b = SequentialRunner().run(pipeline, catalog)
    assert a == b
```

PR #3475, opened 2024-01-03
Add `MemoryDataset` entries to `free_outputs`

SajidAlamQB:
## Description

Context: https://github.com/kedro-org/kedro/issues/1900

The `free_outputs` output from `session` isn't very clear we'll change it to return all free outputs and additionally any MemoryDataSets that are defined in the catalog.

## Developer Certificate of Origin
We need all contributions to comply with the [Developer Certificate of Origin (DCO)](https://developercertificate.org/). All commits must be signed off by including a `Signed-off-by` line in the commit message. [See our wiki for guidance](https://github.com/kedro-org/kedro/wiki/Guidelines-for-contributing-developers/).

If your PR is blocked due to unsigned commits, then you must follow the instructions under "Rebase the branch" on the GitHub Checks page for your PR. This will retroactively add the sign-off to all unsigned commits and allow the DCO check to pass.

## Checklist

- [x] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [x] Signed off each commit with a [Developer Certificate of Origin (DCO)](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository)
- [x] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [x] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [x] Added tests to cover my changes
- [ ] Checked if this change will affect Kedro-Viz, and if so, communicated that with the Viz team

PR #4236, opened 2024-10-17
Add method to remove runtime patterns after run

ElenaKhaustova:
## Description
After the pattern resolution logic refactoring we process all the patterns (`dataset`, `default`, `runtime`) together.

As a result of `run()` we return:

>       datasets that aren't in the catalog and don't match a pattern in the catalog and include MemoryDataset
https://github.com/kedro-org/kedro/blob/ba981350ad57dbcaabf5fd758a9a3d4399a91f20/kedro/runner/runner.py#L108

Before the `run()` we add a runtime pattern to the catalog, so we could process all intermediate outputs as `MemoryDataset`.

So currently when we do two consecutive runs https://github.com/kedro-org/kedro/issues/4235 runtime pattern `{default}` added after the first run affects the next runs, so that all datasets match it and we do not return anything as a result of `run()` cause we think these datasets are in the catalog.

## Development notes
We think that the resolution logic is correct and all patterns should be processed together as we do now. To avoid [this behaviour ](https://github.com/kedro-org/kedro/issues/4235) we added method to remove runtime patters after the run, so they only live within the run and do not affect other runs.

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,6 +1,10 @@
 # Upcoming Release
 
 ## Major features and improvements
+* Implemented dict-like interface for `KedroDataCatalog`.
+
+**Note:** ``KedroDataCatalog`` is an experimental feature and is under active development. Therefore, it is possible we'll introduce breaking changes to this class, so be mindful of that if you decide to use it already. Let us know if you have any feedback about the ``KedroDataCatalog`` or ideas for new features.
+
 ## Bug fixes and other changes
 ## Breaking changes to the API
 ## Documentation changes
--- a/asv.conf.json
+++ b/asv.conf.json
@@ -8,5 +8,11 @@
     "environment_type": "virtualenv",
     "show_commit_url": "http://github.com/kedro-org/kedro/commit/",
     "results_dir": ".asv/results",
-    "html_dir": ".asv/html"
+    "html_dir": ".asv/html",
+    "matrix": {
+        "req": {
+            "kedro-datasets": [],
+            "pandas": []
+        }
+    }
 }
--- a/benchmarks/benchmark_datacatalog.py
+++ b/benchmarks/benchmark_datacatalog.py
@@ -1,0 +1,82 @@
+import pandas as pd
+from kedro_datasets.pandas import CSVDataset
+
+from kedro.io import DataCatalog
+
+base_catalog = {
+    f"dataset_{i}": {
+        "type": "pandas.CSVDataset",
+        "filepath": f"data_{i}.csv",
+    } for i in range(1, 1001)
+}
+# Add datasets with the same filepath for loading
+base_catalog.update({
+    f"dataset_load_{i}": {
+        "type": "pandas.CSVDataset",
+        "filepath": "data.csv",
+    } for i in range(1, 1001)
+})
+# Add a factory pattern
+base_catalog.update({
+    "dataset_factory_{placeholder}": {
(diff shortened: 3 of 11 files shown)


noklam:
>   datasets that aren't in the catalog and don't match a pattern in the catalog and include MemoryDataset

If a dataset is defined in catalog or as a pattern, the catalog would not return as, with the exception of **MemoryDataset**. In this case it should be still return because the runtime pattern add a memory dataset. Can you clarify why changing the pattern resolution somehow breaks this?


https://github.com/kedro-org/kedro/pull/3475

ElenaKhaustova:
> > datasets that aren't in the catalog and don't match a pattern in the catalog and include MemoryDataset
> 
> If a dataset is defined in catalog or as a pattern, the catalog would not return as, with the exception of **MemoryDataset**. In this case it should be still return because the runtime pattern add a memory dataset. Can you clarify why changing the pattern resolution somehow breaks this?
> 
> #3475

As I tried to explain above after we add `runtime` pattern - which matches all datasets it remains in the catalog when the second run, this part of the condition is not valid `don't match a pattern in the catalog`. Previously there was a different logic when resolution and `runtime` patterns were processed separately.

noklam:
I get this in the second run it will be in the `registered_ds`. But the logic here is:
free_output =  output - register_ds (excluding if is a memory dataset)

So my expectation here is that, even if it's registered, it will still be return.

https://github.com/kedro-org/kedro/blob/ba981350ad57dbcaabf5fd758a9a3d4399a91f20/kedro/runner/runner.py#L110

Trace:
I run this twice in the test and add some printing statement in `runner.py`

```
pipeline.outputs()=set()
registered_ds=['params:model_options', 'model_input_table']
memory_datasets={'params:model_options', 'model_input_table'}
free_outputs=set()

pipeline.outputs()=set()
registered_ds=['X_test', 'params:model_options', 'X_train', 'y_test', 'y_train', 'model_input_table', 'regressor']
memory_datasets={'params:model_options', 'model_input_table'}
free_outputs=set()
```

ElenaKhaustova:
> This is the result I get with other test, I think the problem is where we define Memory dataset, in 2nd run I expected, y_test, X_test, regressor in the memory_datasets

They are not in the catalog because we make

```
        catalog = catalog.shallow_copy(
            extra_dataset_patterns=self._extra_dataset_patterns
        )
```

before calling `_run()`, so we modify different object. But `runtime` pattern is stored in `CatalogConfigResolver`, so it remains the same between the runs.

In the new catalog, it will be as you expect.

ElenaKhaustova:
Please note that `shallow_copy` will be removed from the new catalog, but the mechanism to remove runtime patterns after the run will be helpful anyway to not affect consequent runs.

ankatiyar (approved):
Given that we are removing `shallow_copy` stuff in the future, this solution makes sense to me! Thanks @ElenaKhaustova 👍🏾

ElenaKhaustova:
Moved to draft now as want to double-check some cases that @noklam shared.

noklam:
For reference I shared my edge cases here: https://github.com/noklam/kedro-runner-bug-investigation/commits/main/

noklam (approved):

merelcht (approved):
LGTM!

ankatiyar (approved):
Thanks @ElenaKhaustova 👍🏾

Issue #3864, opened 2024-05-13
Backfill some old documentation versions with static assets


astrojuanlu:
## Description
In #3586 we introduced https://github.com/kedro-org/kedro-sphinx-theme across our 3 Sphinx subprojects.

This was done to revert the use of an S3 bucket to hold static assets for reusability #3016.

However, the moment that S3 bucket ceases to exist, several versions of the docs will immediately appear like this:

![Screenshot 2024-05-13 at 09-29-41 Welcome to Kedro’s award-winning documentation! — kedro 0 19 3 documentation](https://github.com/kedro-org/kedro/assets/316517/68dfba56-5457-402c-9a17-3e247a378c38)

Fixing this isn't straightforward: https://docs.readthedocs.io/en/stable/versions.html RTD versions are generated from git tags, so modifying old ones requires some work.

We need to understand what would be the impact of this breakage (what versions are affected), come up with a solution, and decide on what versions (if not all) we will apply it.

## Context
We don't know how many doc versions are affected, so it is good to figure that out as well. 

## Possible Implementation

One possible way of doing this is:

1. Pick the old version to fix, for example `0.19.3`
2. Create a branch that departs from that tag, for example named `0.19.3.x`
3. Make the _necessary changes_
5. Make a PEP 440 post-release https://peps.python.org/pep-0440/#post-releases, for example named `0.19.3.post1`
6. Hide the `0.19.3` version in the docs and show `0.19.3.post1` instead
7. Redirect `/en/0.19.3` to `/en/0.19.3.post1`

> Some projects use post-releases to address minor errors in a final release that do not affect the distributed software (for example, correcting an error in the release notes).

I haven't tested this ⚠️ whoever tries this approach should validate it with a personal project first.

Also, the "necessary changes" might be either backporting `kedro-sphinx-theme`, or even partially reverting #3292.

## Possible Alternatives

There might be other ways to do this, ideas welcome.

rashidakanchwala:
@astrojuanlu  - Should we have a similar ticket on Kedro-viz, or should we also assign a Kedro-viz engineer along with Framework to this ticket, so the work can be done across both in this sprint.

astrojuanlu:
Whatever works best 🙏🏼 I don't have a strong preference

SajidAlamQB:
> We don't know how many doc versions are affected, so it is good to figure that out as well.

[...]

Issue #3741, opened 2024-03-26
Improve SEO and maintenance of documentation versions


astrojuanlu:
In #2980 we discussed about the fact that too many Kedro versions appear in search results.

We fixed that in #3030 by manually controlling what versions did we want to be indexed.

This caused a number of issues though, most importantly #3710: we had been accidentally excluded our subprojects from our search results.

We fixed that in #3729 in a somewhat unsatisfactory fashion. In particular, there are concerns about consistency and maintainability https://github.com/kedro-org/kedro/pull/3729#discussion_r1539480799 (see also https://github.com/kedro-org/kedro/issues/2600#issuecomment-1772673087 about the problem of projects under `kedro-org/kedro-plugins` not having a `stable` version).

In addition, my mind has evolved a bit and I think we should only index 1 version in search engines: `stable`. There were concerns about users not understanding the flyout menu https://github.com/kedro-org/kedro/issues/2980#issuecomment-1705277829 and honestly the `latest` part is also quite confusing (#2823, https://github.com/readthedocs/readthedocs.org/issues/10674) but that's a whole separate discussion.

For now, the problems we want to solve are

- #2980 again (not reopening it, hence this issue) by allowing only 1 version, the most recent stable one, and
- The ongoing maintenance of `robots.txt`, ideally by not having to ever touch it again.

astrojuanlu:
And on the topic of the flyout, here's my thinking:

I think I have become numb to the whole `stable/latest` from RTD, but I think @stichbury is right this is not at all obvious.

Now, look at what happens when I change the default version to be `0.19.3` instead of the current `stable`:

1. User types `https://docs.kedro.org`
2. User gets redirected to `https://docs.kedro.org/en/0.19.3`
3. The flyout shows the number:

![Screenshot 2024-03-26 at 15-34-38 Welcome to Kedro’s award-winning documentation! — kedro 0 19 3 documentation](https://github.com/kedro-org/kedro/assets/316517/1d253d9c-f6bf-437d-a25c-770022015c34)

By having the number in the URL and also in the flyout _by default_, I think it's more obvious how the user should go and switch to their version of choice.

@stichbury in your opinion, do you think this would make our docs journey more palatable?

stichbury:
I think this is good, but doesn't it mean that you have to remember to increment the version number for `stable` in the control panel each time you make a release? If you don't it makes it hard to find the docs for that release (which incidentally are the latest stable 🤦 docs).

[...]

PR #4261, opened 2024-10-28
Manually create a sitemap.xml for docs SEO

DimedS:
## Description

We initially tried using autogenerated `robots.txt` and `sitemap.xml`. However, the issue with this approach is that it does not support sub-projects (Viz and Datasets).

This PR reverts to using a manual `robots.txt`, with a link to a manually created `sitemap.xml`. Both files are now located in the `extra_files` folder, and this path has been included in the `html_extra_path` configuration in `conf.py`.

If this approach proves successful, the next step will be to completely deindex older docs versions using the method proposed by RTD: [https://github.com/readthedocs/readthedocs.org/issues/10648#issuecomment-2485945574](https://github.com/readthedocs/readthedocs.org/issues/10648#issuecomment-2485945574).

## Checklist


docs/source/conf.py
```diff
@@ -223,6 +223,8 @@
 # We've decided to rely on RTD's default robots.txt, so we no longer need to specify a custom one,
 # see discussion at https://github.com/kedro-org/kedro/issues/3741
 # html_extra_path = [str(here / "robots.txt")]
+html_extra_path = [str(here / "sitemap.xml")]
```
astrojuanlu, line 226: 👀 So this was it...

DimedS: Unfortunately, I checked the build, and it didn’t help. Something is still overwriting the `sitemap.xml`, even with that line included.

DimedS:
After adding `html_extra_path = [str(here / "sitemap.xml")]` to `conf.py`, I observed the following post-build:

- The custom `sitemap.xml` is accessible at: [https://kedro--4261.org.readthedocs.build/en/4261/sitemap.xml](https://kedro--4261.org.readthedocs.build/en/4261/sitemap.xml)
- The autogenerated `sitemap.xml` is still located at: [https://kedro--4261.org.readthedocs.build/sitemap.xml](https://kedro--4261.org.readthedocs.build/sitemap.xml) and is likely the version that will be used at [https://docs.kedro.org/sitemap.xml](https://docs.kedro.org/sitemap.xml).

I also tried adding a custom `robots.txt`, but the autogenerated version still appears at [https://kedro--4261.org.readthedocs.build/robots.txt](https://kedro--4261.org.readthedocs.build/robots.txt). This makes it difficult to verify the effect in this build, and it’s possible that the custom `sitemap.xml` and `robots.txt` will only take effect at a later stage, potentially being used at [https://docs.kedro.org](https://docs.kedro.org).

docs/source/_templates/layout.html
```diff
@@ -0,0 +1,5 @@
+{% extends "!layout.html" %}
+{% block extrahead %}
+    {{ super() }}
+    <meta name="robots" content="noindex, nofollow">
```
astrojuanlu, line 4: We don't want to do this in all cases right? Only when the version is old

DimedS: yes, it's possible to modify it to handle that, I'm just trying to understand, is it in reality possible to rebuild our old versions, what do you think? And currently I'm also thinking maybe that `noindex` is not the right solution at all:
https://github.com/kedro-org/kedro/issues/3741#issuecomment-2490590143

astrojuanlu: > is it in reality possible to rebuild our old docs versions, what do you think?

We did it to address a security issue https://github.com/kedro-org/kedro/issues/3864 less than ideal, I know.

However, there's another way of doing it that doesn't require rebuilding: https://github.com/readthedocs/readthedocs.org/issues/10648#issuecomment-2485945574 we should try that first.

In any case, I think the priority for this PR should be to avoid blocking the subprojects.

> For the current release, I propose moving forward only with a manual update to robots.txt and sitemap.xml that will solve P0.

👍🏼

docs/source/conf.py
```diff
@@ -220,10 +220,11 @@
 
 # html_extra_path used to define a path to robots.txt which is used by webcrawlers
 # to ignore or allow certain links.
-# We've decided to rely on RTD's default robots.txt, so we no longer need to specify a custom one,
-# see discussion at https://github.com/kedro-org/kedro/issues/3741
-# html_extra_path = [str(here / "robots.txt")]
-html_extra_path = [str(here / "sitemap.xml")]
+# We've tried to rely on RTD's default `robots.txt`, but it doesn't support sub-projects (Viz and Datasets),
+# so we decided to use our own `robots.txt` again, with a link to the custom `sitemap.xml`.
+# See the discussion at https://github.com/kedro-org/kedro/issues/3741.
```
astrojuanlu, line 225: Please refer to https://github.com/readthedocs/readthedocs.org/issues/6841 too

astrojuanlu (approved):
💯

noklam (approved):

Issue #4250, opened 2024-10-22
ThreadRunner Dataset DatasetAlreadyExistsError: Dataset has already been registered


noklam:
## Description

Originated from https://github.com/kedro-org/kedro/pull/4210

## Context

Upon investigation, I found that this error seems to be related to dataset factory pattern only.

1. If dataset are registered dataset in catalog, I can run without any error from 0.18.12 (didn't go earlier since factory not introduced, to current release)
2. If dataset factory is used, it starts breaking since 0.18.12

The current conclusion is that this is not an error introduced recently. Though there seems to be partial fix previously but it doesn't works for my test case.

Related:
- #4007 
- https://github.com/kedro-org/kedro/issues/3739#issuecomment-2019190009, an user say that by pre-registering all the dataset it works.
- https://github.com/kedro-org/kedro/issues/4191

## Steps to Reproduce
Using a similar test written in https://github.com/kedro-org/kedro/pull/4210 from `benchmark_runner.py` for ThreadRunner.

This is the snippet that I use:
```python
# Write the benchmarking functions here.
# See "Writing benchmarks" in the asv docs for more information.

import time
from pathlib import Path

import yaml

from kedro.io.data_catalog import DataCatalog
from kedro.pipeline import node
from kedro.pipeline.modular_pipeline import pipeline
from kedro.runner import ParallelRunner, SequentialRunner, ThreadRunner


# Simulate an I/O-bound task
def io_bound_task(input_data):
    time.sleep(2)  # Simulate an I/O wait (e.g., reading from a file)
    output = input_data
    return output


# Simulate a compute-bound task (matrix multiplication)
def compute_bound_task(input_data) -> str:
    # Simulate heavy compute that are not using multicore (not pandas/numpy etc)
    ans = 1
    for i in range(1, 50000):
        ans = ans * i
    return "dummy"


def create_data_catalog():
    """
    Use dataset factory pattern to make sure the benchmark cover the slowest path.
    """
    catalog_conf = """

'output_{pattern}':
    type: pandas.CSVDataset
    filepath: benchmarks/data/'{pattern}.csv'

'numpy_{pattern}':
    type: pickle.PickleDataset
    filepath: benchmarks/data/'{pattern}.pkl'

'{catch_all}':
    type: pandas.CSVDataset
    filepath: benchmarks/data/data.csv


# dummy_1:
#     type: pandas.CSVDataset
#     filepath: benchmarks/data/data.csv

# dummy_8:
[...]

Issue #3935, opened 2024-06-06
[DataCatalog]: Spike - Lazy dataset loading


ElenaKhaustova:
## Description
Users are required to install all dependencies even for unused datasets, leading to unnecessary complexity and confusion.

We propose implementing a lazy dataset loading feature to allow users to load only the datasets they need without causing pipeline failures.

Relates to https://github.com/kedro-org/kedro/issues/2829

## Context

- "You need to install all dependencies even for unused datasets (in case you want to run pipeline partially or do not load some dataset when standalone catalog usage)."
- "We have a lot of data entries and different dependencies and when we just want to rerun an anaysis partially, we are frustrated because we need to install all the packages to just load one data source. Why would I need to install excel dependencies to instantiate the `DataCatalog` to load a `csv` which does not need Excel?"
- The error users get now in case of missing dependencies is unclear https://github.com/kedro-org/kedro/issues/3911
```
DatasetError: An exception occurred when parsing config for dataset 'companies':
No module named 'pandas'. Please see the documentation on how to install relevant dependencies for kedro_datasets.pandas.CSVDataset:
```

## Spike task
Investigate how to actually do the lazy loading: 1) the actual lazy loading of datasets, only import the datasets at the time we load the data 2) understand what part of the pipeline needs to be run, and only import what's required for that run.

ElenaKhaustova:
## Kedro Viz workflow

kedro_viz -> integrations -> kedro -> data_loader -> _load_data_helper (https://github.com/kedro-org/kedro-viz/blob/main/package/kedro_viz/integrations/kedro/data_loader.py#L58 ) - This creates all required Kedro objects like Session, Context, DataCatalog and Pipelines

**Specially for datasets we need DataCatalog object**

- kedro_viz -> integrations -> kedro -> data_loader -> _load_data_helper -> catalog = context.catalog (https://github.com/kedro-org/kedro-viz/blob/main/package/kedro_viz/integrations/kedro/data_loader.py#L101 )
[...]

PR #4262, opened 2024-10-28
Move warm-up from session to runner

ElenaKhaustova:
## Description
Solves https://github.com/kedro-org/kedro/issues/4250

Relates to https://github.com/kedro-org/kedro/issues/3935 - step 5 in the proposed solution

## Development notes
Moved warm-up from the session to the `AbstractRunner` before we call `_run()` and made this logic common for all runners. This simplified the logic as now it's common for all runners and all the patterns are resolved before the pipeline run.

We added a unit test with dataset patterns to check that `ThreadRunner` is not failing with a `Dataset 'name' has already been registered` error. In the test, we check that the dataset was registered at the warm-up, and we successfully passed to loading it, though we do not do the actual loading.

We tried to make the full test with actual data loading — https://github.com/kedro-org/kedro/commit/208a24b7d093c85a0e8e00b55747e5e9231c5f41. But we faced a problem where the first thread was trying to load data before it was created — https://github.com/kedro-org/kedro/actions/runs/11562442829/job/32183629244. This happened mostly at the CI for the latest Python versions only and was hard to reproduce locally. We tried some checks if the file exists before calling `run()` but it didn't help. So, we changed the test to the current one to exclude data creation.

## Checklist


kedro/runner/runner.py
```diff
@@ -112,6 +112,10 @@ def run(
             self._logger.info(
                 "Asynchronous mode is enabled for loading and saving data"
             )
+
```
noklam, line 115: Can we add a comment about this? I also notice this line is duplicated:
`registered_ds = [ds for ds in pipeline.datasets() if ds in catalog]`

Maybe we can refactor this into one loop so we only loop `pipeline.datasets()` once?

ElenaKhaustova: 1. Kept one loop and added a comment about warm-up
2. The line is duplicated on purpose, there's a comment explaining the difference but I also renamed the variable to avoid confusion

noklam (approved):

merelcht (approved):
Thanks for the detailed description on the PR! This makes all sense to me 👍 
Can you add a comment in the release notes for future reference?

kedro/runner/runner.py
idanov: `registered_ds_no_runtime_patterns` is a bit of a mouthful, could we come up with a shorter name? E.g. `warm_concrete_ds` (not necessarily the best option, but just an example)?

Tbh, I don't mind the two loops - they show the purpose better without too much of a slowdown, but the current refactor is also ok.

idanov (approved):

Issue #3932, opened 2024-06-05
[DataCatalog]: Spike - Catalog serialization and deserialization support


ElenaKhaustova:
## Description

1. Users admit the lack of persistency in the add workflow, as there is no built-in functionality to save modified catalogs.
2. Users express the need for an API to save and load catalogs after compilation or modification by converting catalogs to `YAML` format and back.
3. Users encounter difficulties loading pickled `DataCatalog` objects when the Kedro version changes when loading, leading to compatibility issues. They require a solution to serialize and deserialize the `DataCatalog` object without dependency on Kedro versions.

We propose to explore the feasibility of implementing `to_yaml()` and `from_yaml()` methods for the `DataCatalog` object to facilitate serialization and deserialization without dependency on Kedro versions.

## Context
User feedback:

- Add workflow is missing persistency, so you can not save modified catalog: "You have a catalog and then you start adding extra stuff to it, currently we just throw away those added things when they close a notebook."
- Catalog to `YAML` function is needed to save modified catalog: "People have always asked for it. Could I have a catalog to `YAML` function so that you could actually spit out the `YAML` files that are needed to do this again later on?"
- Competitors provide the functionality to compile catalog and showcase the result: "I would point to the DPC compile workflow. And actually, if you do DBT run it does DBT compile first and then runs the compiled outputs. Whereas in Kedro, you have your very concise complicated `YAML` and it will all that compilation happens at run time and there's no way for the user to see it."
- When pickling `DataCatalog` object they experience difficulties in loading it back if the kedro version is different: "Serialization is an issue because I often pickle a catalog (mostly as part of a mlflow model). Pickling the catalog is really something that leads to a lot of problems because if I don't have the exact same Kedro version when I want to load the catalog, if the object has any change inside - private method or attribute it will lead to error."

https://github.com/Galileo-Galilei/kedro-mlflow/blob/64b8e94e1dafa02d979e7753dab9b9dfd4d7341c/kedro_mlflow/mlflow/kedro_pipeline_model.py#L143

```
# pseudo code
pickle.dumps(catalog)
pickle.loads(catalog) # this will fail if I reload with a newer kedro version and any attributes (even private) has changed. This breaks much more often that we should expect. 
```

"It would be much more robust to be able to do this":
```
# pseudo code
catalog.serialize("path/catalog.yml") # name TBD: serialize? to_config? to_yaml? to_json? to_dict? 
[...]

PR #4270, opened 2024-10-30
[DataCatalog]: Lazy dataset loading

ElenaKhaustova:
## Description
Implementation of https://github.com/kedro-org/kedro/issues/3935#issuecomment-2433144639

The PR is done on top of https://github.com/kedro-org/kedro/pull/4262

## Checklist


noklam (approved):
Thanks for explaining, I think the solution make sense to only materialise the needed datasets.

kedro/io/kedro_data_catalog.py
```diff
@@ -210,17 +244,21 @@ def get(
         Returns:
             An instance of AbstractDataset.
         """
-        if key not in self._datasets:
+        if key not in self._datasets and key not in self._lazy_datasets:
             ds_config = self._config_resolver.resolve_pattern(key)
             if ds_config:
                 self._add_from_config(key, ds_config)
 
+        non_initialized_dataset = self._lazy_datasets.pop(key, None)
```
Galileo-Galilei, line 252: Am I correct that we also want this PR to pave the way for #3932 by storing the configuration in the catalog for lazy dataset? If so, should we really remove the dataset from ``_lazy_datasets`` once it is materialized? If we do this it prevents further access to the original raw configuration  (this can still be reassessed further, and I am not sure we want to tackle both problems in this PR, but it's worth thiniking it it will fits well in the broader roadmap)

ElenaKhaustova: We think that proper solution for https://github.com/kedro-org/kedro/issues/3932 will require storing configuration at the dataset level and `dataset.to_yaml()`, `dataset.from_yaml()` for the following reasons:

- We allow to initializing catalog from datasets or their configurations. If it's the first case currently we are not able to retrieve configuration from the dataset object and thus cannot serialize it.
- And the same problem appears if someone adds dataset to the catalog which is probably the main use case for catalog dump and load.

merelcht (approved):
Very clean solution! 🌟 The explanations also make perfect sense to me, thanks for adding that as context.

rashidakanchwala (approved):
This looks great! I’m hopeful that lazy loading datasets will also improve the initial load time in Kedro-Viz. Looking forward to seeing this!

RELEASE.md
```diff
@@ -2,11 +2,13 @@
 
 ## Major features and improvements
 * Implemented dict-like interface for `KedroDataCatalog`.
+* Implemented lazy dataset loading for `KedroDataCatalog`.
```
idanov, line 5:
```suggestion
* Implemented lazy dataset initialising for `KedroDataCatalog`.
```

kedro/io/kedro_data_catalog.py
```diff
@@ -398,7 +451,10 @@ def _get_dataset(
         return self.get_dataset(dataset_name, version, suggest)
 
     def add(
-        self, ds_name: str, dataset: AbstractDataset, replace: bool = False
+        self,
+        ds_name: str,
+        dataset: AbstractDataset | _LazyDataset,
```
idanov, line 456: Are we sure we'd like to expose `_LazyDataset` to the users? If yes, then maybe we need to make it a proper `LazyDataset` that can be instantiated even outside of the catalog. Otherwise, I'd prefer for us to hide it from this method signature.

idanov (approved):

kedro/io/kedro_data_catalog.py
ElenaKhaustova: We don't want to expose it and we added it here only until we use the old `add()` method for compatibility with the old catalog. I will go after the breaking change as setter will be used instead.

PR #4299, opened 2024-11-05
Fix `--version` option for Kedro command

### LINKED ISSUE #4296: `kedro --version` isn't working
## Description
Reported by @noklam, since adding lazy loading of Kedro subcommands, the `--version`/`-V` option isn't working. 

## Context
This bug is originating in Kedro 0.19.7 -> https://github.com/kedro-org/kedro/pull/3883

ankatiyar:
## Description

Fix #4296 

## Development notes

Copying the comment from https://github.com/kedro-org/kedro/issues/4296#issuecomment-2456616327
> I figured out what the problem is:
> 
> https://github.com/kedro-org/kedro/blob/a1fae5018f35243a5e49a54a9dd3223b2c4ea743/kedro/framework/cli/cli.py#L220
> 
> Due to the changes in lazy loading PR, I re-ordered the global commands list to consider
> 
> * first the commands loaded from plugins,
> * then `cli` which is the group with `info` and the `version_option` decorator
> * and then the `global_commands` group which contains the `new` and `starter` lazy commands.
> 
> So if any plugin with global commands (eg. Kedro-Viz) is installed in your env, the `--version` option doesn't work. It works when you uninstall Kedro viz. Which is why it must be working in the CI and for @DimedS
> 
> The solution is simply to re-order the command groups to `[cli, *load_entry_points("global"), global_commands]` but that would mean that users can't overwrite `kedro info` which I think is acceptable.

## Checklist

### DIFF
--- a/kedro/framework/cli/cli.py
+++ b/kedro/framework/cli/cli.py
@@ -217,7 +217,7 @@ class KedroCLI(CommandCollection):
         combines them with the built-in ones (eventually overriding the
         built-in ones if they are redefined by plugins).
         """
-        return [*load_entry_points("global"), cli, global_commands]
+        return [cli, *load_entry_points("global"), global_commands]
 
     @property
     def project_groups(self) -> Sequence[click.MultiCommand]:


noklam (approved):
Tested and it works fine.

> The solution is simply to re-order the command groups to [cli, *load_entry_points("global"), global_commands] but that would mean that users can't overwrite kedro info which I think is acceptable.

Could you explains a bit more why this is the case? It's unclear to me why re-ordering fix this

ankatiyar:
> Could you explains a bit more why this is the case? It's unclear to me why re-ordering fix this

I suspect that the merging of command groups that we do in custom `CommandCollection` is the issue, before the lazy loading `cli` group was ahead of the commands loaded from plugins but the over-writing commands from plugins replaced the core commands. Not a 100 percent sure myself without fully digging into how `version_option` in click works. 🤔

ElenaKhaustova (approved):
Thank you, @ankatiyar!

Issue #4329, opened 2024-11-13
[DataCatalog]: Catalog to config


ElenaKhaustova:
## Description
Implement `KedroDataCatalog.to_config()`method as a part of catalog serialization/deserialization feature https://github.com/kedro-org/kedro/issues/3932

## Context

**Requirements:**

- Catalog already has `from_config`, so `KedroDataCatalog.to_config()` have to output configuration further used with the existing `KedroDataCatalog.from_config()` method to load it. method https://github.com/kedro-org/kedro/blob/9464dc716c987ac0bcadba49aa97a4fa1ae18248/kedro/io/kedro_data_catalog.py#L268
- We want to solve this problem at the framework level and avoid existing datasets' modifications where possible.

### Solution description

We consider 3 different ways of loading datasets:

1. Lazy datasets loaded from the config — in this case, we store the dataset configuration at the catalog level; the dataset object is not instantiated.
2. Materialized datasets loaded from the config — we store the dataset configuration at the catalog level and use `dataset.from_config()` method to instantiate dataset which calls the underlying dataset constructor.
3. Materialized datasets added to the catalog — instantiated datasets' objects are passed to the catalog, dataset configuration is not stored at the catalog level.

**1** - can be solved at the catalog level
**2 and 3** require retrieving dataset configuration from instantiated dataset object

#### Solution for **2 and 3** avoiding existing datasets' modifications (as per requirements)

1. Use `AbstractDataset.__init_subclass__` which allows to change the behavior of subclasses from inside the `AbstractDataset`: https://docs.python.org/3/reference/datamodel.html#customizing-class-creation
2. Create a decorator for child init
3. Save the original child init
4. Replace original child init with decorated init calling post init where we save call args: https://docs.python.org/3/library/inspect.html#inspect.getcallargs
5. Save call args at the level of `AbstractDataset` in the `_init_args` field.
6. Implement `AbstractDataset.to_config()` to retrieve configuration from the instantiated dataset object based on the object's `_init_args`.

#### Implement `KedroDataCatalog.to_config`
Once **2 and 3** are solved, we can implement a common solution at the catalog level. For that, we need to consider cases when we work with lazy and materialized datasets and retrieve configuration from the catalog or using `AbstractDataset.to_config()`.

[...]

Issue #4327, opened 2024-11-13
Discrepancy between setting `save_version` via catalog constructor and when passing datasets


ElenaKhaustova:
## Description
In the catalog constructor https://github.com/kedro-org/kedro/blob/9464dc716c987ac0bcadba49aa97a4fa1ae18248/kedro/io/kedro_data_catalog.py#L65 and `from_config` https://github.com/kedro-org/kedro/blob/9464dc716c987ac0bcadba49aa97a4fa1ae18248/kedro/io/kedro_data_catalog.py#L273 method we require passing **one** save version for all datasets. We further use this save version to instantiate all datasets objects.

This constraint is bypassed when passing instantiated dataset objects to catalog constructor.

## Context

1. It's not clear whether this constraint is a requirement.
2. It's not clear what `save version` should we save and load back when serialization/deserialization https://github.com/kedro-org/kedro/issues/3932 

## Possible Implementation

1. Clarify the requirement of having single `load_version` per catalog.
2. Based on the first point either remove the requirement or apply it for instantiated datasets passed to the catalog.

PR #4347, opened 2024-11-22
Validate datasets versions

ElenaKhaustova:
## Description
Solves https://github.com/kedro-org/kedro/issues/4327

Merge before https://github.com/kedro-org/kedro/issues/4329

## Development notes

Added `_validate_versions` function to ensure all datasets in a catalog adhere to a versioning scheme - we allow single load version per dataset in the catalog and one save version for all datasets in the catalog. The function automatically updates the provided load versions based on the versions specified for the individual datasets. It also ensures all versioned datasets in the catalog share the same save version. If a conflict arises, a `VersionAlreadyExistsError` is raised.

Validation is applied to both `DataCatalog` and `KedroDataCatalog` when a catalog is created or the dataset is added.

## Checklist


kedro/io/core.py
```diff
@@ -955,3 +964,56 @@ def confirm(self, name: str) -> None:
...
+def _validate_versions(
+    datasets: dict[str, AbstractDataset] | None,
+    load_versions: dict[str, str],
+    save_version: str | None,
+) -> tuple[dict[str, str], str | None]:
+    """Validates and synchronizes dataset versions for loading and saving.
+
+    Insures consistency of dataset versions across a catalog, particularly
+    for versioned datasets. It updates load versions and validates that all
+    save versions are consistent.
```
merelcht, line 978:
```suggestion
    """Validates and synchronises dataset versions for loading and saving.

    Ensures consistency of dataset versions across a catalog, particularly
    for versioned datasets. It updates load versions and validates that all
    save versions are consistent.
```

kedro/io/core.py
```diff
@@ -955,3 +964,56 @@ def confirm(self, name: str) -> None:
...
+
+    Args:
+        datasets: A dictionary mapping dataset names to their instances.
+            if None, no validation occurs.
+        load_versions: A mapping between dataset names and versions
+            to load.
+        save_version: Version string to be used for ``save`` operations
+            by all datasets with enabled versioning.
```
merelcht, line 986:
```suggestion
        save_version: Version string to be used for ``save`` operations
            by all datasets with versioning enabled.
```

kedro/io/core.py
```diff
@@ -955,3 +964,56 @@ def confirm(self, name: str) -> None:
...
+        if isinstance(cur_ds, AbstractVersionedDataset) and cur_ds._version:
+            if cur_ds._version.load:
+                cur_load_versions[ds_name] = cur_ds._version.load
+            if cur_ds._version.save:
+                cur_save_version = cur_save_version or cur_ds._version.save
+                if cur_save_version != cur_ds._version.save:
+                    raise VersionAlreadyExistsError(
+                        f"Cannot add a dataset `{ds_name}` with `{cur_ds._version.save}` save version. "
+                        f"Save version set for the catalog is `{cur_save_version}`"
+                        f"All datasets in the catalog must have the same Save version."
```
merelcht, line 1016:
```suggestion
                        f"All datasets in the catalog must have the same save version."
```

merelcht (approved):
LGTM, and all makes sense 👍 

One thing I was wondering is if this could somehow break existing workflows that use the `DataCatalog`?

ElenaKhaustova:
> LGTM, and all makes sense 👍
> 
> One thing I was wondering is if this could somehow break existing workflows that use the `DataCatalog`?

That's a good question. Technically, we always required `DataCatalog` with versioned datasets, and we have a note in the docstrings that the catalog can have only one version. But we never validated it, and this requirement can be bypassed. So we consider this change a fix—that's why we applied it for both old and new catalogs. We also think this shouldn't be the common case, but even if some workflows break, that's because people used versioning not as expected.

The alternative was to apply these changes just to `KedroDataCatalog`.

merelcht:
> That's a good question. Technically, we always required `DataCatalog` with versioned datasets to have only one save version. But we never validated it, and this requirement can be bypassed. So we consider this change as a fix—that's why we applied it for both old and new catalogs. We also think this shouldn't be the common case, but even if some workflows break, that's because people used versioning not as expected.
> 
> The alternative is to apply these changes just to `KedroDataCatalog`.

Yes so technically speaking this is breaking for people who have been passing instantiated dataset objects to catalog constructor with different save versions. It's a very rare thing to do and I agree this is actually improving the expected behaviour. So let's just keep it like this, but then at least if someone does complain we know what's going on.

idanov (approved):
LGTM 👍

PR #4353, opened 2024-11-26
Refactor run methods more into abstract method

### LINKED ISSUE #4290: Abstract `_run` as much as possible
## Description
The `_run()` method implementation in the various runners is very similar. We can refactor this to be even more similar so we can move all common code into the `_run()` in the `AbstractRunner`.

## Context
This will make it easier for people to implement a custom runner, because the `_run()` method will have more in place to get started from.

merelcht:
## Description
Resolves #4290 

## Development notes
- Refactored and merged logic of the various `_run` method implementations into the abstract `_run` method. 
- Added abstract `_get_executor` method, implemented by each runner.
- Refactored logic to determine number of workers from `ParallelRunner` and `ThreadRunner` into shared `validate_max_workers` method.
- Changed `hook_manager` argument in runners to allow it to be `None`, which is needed for the `ParallelRunner`, because hook manager can't be serialised.
- Updated `TestSuggestResumeScenario` for `SequentialRunner`, because it's now using a `ThreadPoolExecutor`, the suggestions can vary per run. I've manually created a project with the same pipelines and verified that the suggestions (even if they vary) do always work. 
- Added `TestSuggestResumeScenario` tests for `ThreadRunner`. 

## Checklist


tests/runner/test_sequential_runner.py
```diff
@@ -271,9 +271,9 @@ class TestSuggestResumeScenario:
         [
             (["node1_A", "node1_B"], r"No nodes ran."),
             (["node2"], r"(node1_A,node1_B|node1_B,node1_A)"),
-            (["node3_A"], r"(node3_A,node3_B|node3_B,node3_A)"),
-            (["node4_A"], r"(node3_A,node3_B|node3_B,node3_A)"),
-            (["node3_A", "node4_A"], r"(node3_A,node3_B|node3_B,node3_A)"),
+            (["node3_A"], r"(node3_A,node3_B|node3_B,node3_A|node3_A)"),
+            (["node4_A"], r"(node3_A,node3_C|node3_C,node3_A|node3_A)"),
```
merelcht, line 275:
```suggestion
            (["node4_A"], r"(node3_A,node3_B|node3_B,node3_A|node3_A)"),
```

kedro/runner/sequential_runner.py
```diff
@@ -46,11 +46,16 @@ def __init__(
             is_async=is_async, extra_dataset_patterns=self._extra_dataset_patterns
         )
 
+    def _get_executor(self, max_workers: int) -> ThreadPoolExecutor:
```
noklam, line 49: Is this necessary to create a thread for SeqeuntialRunner?

kedro/runner/runner.py
```diff
@@ -443,3 +546,27 @@ def run_node(
     )
     node = task.execute()
     return node
+
+
+def validate_max_workers(max_workers: int | None) -> int:
```
noklam, line 551: Is there a reason why `_validate_catalog`  and `_validate_nodes` are private methods and this one is public?

kedro/runner/thread_runner.py
```diff
@@ -104,42 +100,9 @@ def _run(
...
-                        self._suggest_resume_scenario(pipeline, done_nodes, catalog)
-                        raise
-                    done_nodes.add(node)
-                    self._logger.info("Completed node: %s", node.name)
-                    self._logger.info(
-                        "Completed %d out of %d tasks", len(done_nodes), len(nodes)
-                    )
-
-                    self._release_datasets(node, catalog, load_counts, pipeline)
+        super()._run(
```
noklam, line 103: Is it still needed if it's just using the base method?

kedro/runner/runner.py
```diff
@@ -6,17 +6,31 @@
...
+import sys
 import warnings
 from abc import ABC, abstractmethod
-from collections import deque
+from collections import Counter, deque
+from concurrent.futures import (
+    FIRST_COMPLETED,
+    ProcessPoolExecutor,
+    ThreadPoolExecutor,
+    wait,
```
noklam, line 18: Would this import `multiprocessing` as a dependencies? I recalled in the past we have issues with `ShelveStore` because even importing the library cause issues on restricted environment like AWS Lambda.

kedro/runner/sequential_runner.py
merelcht: Strictly speaking it's not necessary, because it's just one thread. I did it to make it more consistent with the other runners and to make it possible to have the shared logic in the abstract `_run()` method:
```
        with self._get_executor(max_workers) as pool:
            while True:
```

kedro/runner/runner.py
```diff
@@ -166,25 +180,97 @@ def run_only_missing(
...
         return self.run(to_rerun, catalog, hook_manager)
 
+    @abstractmethod  # pragma: no cover
+    def _get_executor(
+        self, max_workers: int
+    ) -> ThreadPoolExecutor | ProcessPoolExecutor:
+        """Abstract method to provide the correct executor (e.g., ThreadPoolExecutor or ProcessPoolExecutor)."""
+        pass
+
     @abstractmethod  # pragma: no cover
```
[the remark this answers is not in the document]
merelcht: I'd say yes, because you can still overwrite this when creating a custom Runner. But now it's also easier to create a custom runner, because the _run() method has more in place to get started from.

kedro/runner/runner.py
merelcht: I did that because `_validate_catalog` and `_validate_nodes` need to be implemented on the Runner level and aren't relevant to all runners atm, while `validate_max_workers` is a helper method used in both `ThreadRunner` and `ParallelRunner`. I can change it if it doesn't make sense.

kedro/runner/thread_runner.py
merelcht: Yes, because `_run()` is still an abstract method.

kedro/runner/runner.py
idanov: It feels odd to have this one exposed, although I appreciate that importing a private one in a couple of places also feels odd. Could we make it private in a non-disruptive way? Maybe a class method (or maybe not, since it's used in the ParallelRunnerManager)?

kedro/runner/sequential_runner.py
idanov: Would that have any consequences for non thread-safe datasets?

kedro/runner/runner.py
```diff
@@ -166,25 +180,97 @@ def run_only_missing(
 
         return self.run(to_rerun, catalog, hook_manager)
 
+    @abstractmethod  # pragma: no cover
+    def _get_executor(
```
idanov, line 184: Could we make this less specific and just return `Executor`? Both executors have the same interface and it would be nice if we just program against the interface rather than the concretes. That applies to all other places where for type hints we could just set the type as `Executor` and not the concrete implementations.

https://docs.python.org/3/library/concurrent.futures.html#executor-objects

idanov (commented):
It's amazing how similar now all the runners start to look like, I love it!

kedro/runner/sequential_runner.py
merelcht: For `SequentialRunner` it will only use 1 thread, so it shouldn't have any consequences (although I'm sure there are edge cases). A quick search in our datasets doesn't show any mentions of thread-safety. Do you know of any specific dataset not being thread-safe?

kedro/runner/runner.py
merelcht: Done!

kedro/runner/runner.py
merelcht: Changed this to a private class method.

kedro/runner/thread_runner.py
```diff
@@ -85,14 +78,17 @@ def _get_required_workers_count(self, pipeline: Pipeline) -> int:
             else required_threads
         )
 
+    def _get_executor(self, max_workers: int) -> ThreadPoolExecutor:
```
idanov, line 81:
```suggestion
    def _get_executor(self, max_workers: int) -> Executor:
```

kedro/runner/sequential_runner.py
idanov: I don't have any dataset in mind, was just curious if we have considered this. If we don't have any indication that this doesn't make a difference in any way, we should just go ahead and deal with it only if it causes problems later.

idanov, line 49:
```suggestion
    def _get_executor(self, max_workers: int) -> Executor:
```

idanov (approved):
Awesome, this makes all the runners look almost identical!

kedro/runner/parallel_runner.py
```diff
@@ -189,14 +172,17 @@ def _get_required_workers_count(self, pipeline: Pipeline) -> int:
 
         return min(required_processes, self._max_workers)
 
+    def _get_executor(self, max_workers: int) -> ProcessPoolExecutor:
```
merelcht, line 175:
```suggestion
    def _get_executor(self, max_workers: int) -> Executor:
```

kedro/runner/runner.py
```diff
@@ -166,25 +175,95 @@ def run_only_missing(
...
+        node_dependencies = pipeline.node_dependencies
+        todo_nodes = set(node_dependencies.keys())
+        done_nodes: set[Node] = set()
+        futures = set()
+        done = None
+        max_workers = self._get_required_workers_count(pipeline)
+
+        with self._get_executor(max_workers) as pool:
+            while True:
+                ready = {n for n in todo_nodes if node_dependencies[n] <= done_nodes}
```
DimedS, line 218: I propose reducing the complexity of that line in a separate PR because this becomes more critical with the unification of runners. As I understand, the previous implementation of the SequentialRunner (most popular runner) simply executed an ordered list of nodes sequentially. However, the current approach performs a full loop over all rest nodes after each executed node, this results in an overall complexity greater than quadratic.

I believe we can achieve linear complexity by maintaining a count of unmet dependencies for each node. As nodes are completed, we decrement the counter for their dependents and mark a node as ready when its counter reaches zero.

DimedS (approved):
Thank you for the PR, @merelcht! Great job on unifying and simplifying the runner's code - it looks great to me! I’ve left one comment suggesting a potential complexity optimisation.

kedro/runner/runner.py
merelcht: That's a good point @DimedS. I'll create a new issue for this, so it's not forgotten.

merelcht: https://github.com/kedro-org/kedro/issues/4373

PR #4368, opened 2024-12-04
Change example test on default template

lrcouto:
## Description

Part of https://github.com/kedro-org/kedro/issues/3954

Changes the default template example test to match https://github.com/kedro-org/kedro-starters/pull/261

## Checklist


kedro/templates/project/{{ cookiecutter.repo_name }}/tests/test_run.py
```diff
@@ -1,42 +1,20 @@
...
 # functionality
-class TestProjectContext:
-    def test_project_path(self, project_context):
-        assert project_context.project_path == Path.cwd()
+
+class TestKedroRun:
+    def test_kedro_run(self):
+        bootstrap_project(Path.cwd())
+
+        with KedroSession.create(project_path=Path.cwd()) as session:
```
datajoely, line 19: If you do this, how do we test hooks?

kedro/templates/project/{{ cookiecutter.repo_name }}/tests/test_run.py
```diff
@@ -1,42 +1,20 @@
...
 # The tests below are here for the demonstration purpose
 # and should be replaced with the ones testing the project
 # functionality
-class TestProjectContext:
-    def test_project_path(self, project_context):
-        assert project_context.project_path == Path.cwd()
+
+class TestKedroRun:
+    def test_kedro_run(self):
+        bootstrap_project(Path.cwd())
```
DimedS, line 17: I think it would be helpful for users to explain what's happening here. Adding a link to the docs in the comments could provide more clarity:
[Testing a Kedro Project](https://docs.kedro.org/en/stable/tutorial/test_a_project.html).

It might also be a good idea to update the docs to include a more detailed explanation of this testing example.

DimedS (approved):
Thanks, @lrcouto! I like the change from Context to Session. I tried it manually, and it works well. I've left a few comments.

kedro/templates/project/{{ cookiecutter.repo_name }}/src/{{ cookiecutter.python_package }}/pipelines/example/pipeline.py
```diff
@@ -0,0 +1,10 @@
+from kedro.pipeline import Pipeline, node
+from .nodes import create_data, process_data
+
+def create_pipeline(**kwargs):
+    return Pipeline(
+        [
+            node(func=create_data, inputs=None, outputs="raw_data", name="create_data_node"),
+            node(func=process_data, inputs="raw_data", outputs="processed_data", name="process_data_node"),
+        ]
```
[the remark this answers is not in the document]
lrcouto: Yeah that's a question I was asking myself as well. Having no example nodes would cause the test to fail, but I'm not sure if that would be that undesirable, since it's supposed to be a code example. 

Another option would be mocking the response instead of using an actual node, but that would go against the "no mocking" idea that was discussed on the issue.

kedro/templates/project/{{ cookiecutter.repo_name }}/tests/test_run.py
lrcouto: We were discussing [here](https://github.com/kedro-org/kedro-starters/pull/261#pullrequestreview-2478629238) that the test should be focused on executing a run. Maybe we could have another example for hook functionality?

datajoely: So I wasn't part of that discussion - but I was part of the design of the initial test file design and this was the reason we use a context not session. This is simpler for most use-cases so we can just put a hook testing part in the docs, but also it would be nice to just pass hooks to the session.

lrcouto: That's good to know, thanks for the context! I can change the test to pass the hooks to the session.

merelcht (changes requested):
We shouldn't be adding example code in the default template. Instead, I'd say let's just update the tests to verify that there isn't a pipeline to run. We can add some docstrings in the test to explain what's happening.

docs/source/development/automated_testing.md
```diff
@@ -85,39 +85,28 @@ tests
...
+        with KedroSession.create(project_path=Path.cwd()) as session:
+            assert session.run() is not None
 ```
 
 This test is redundant, but it introduces a few of `pytest`'s core features and demonstrates the layout of a test file:
-- [Fixtures](https://docs.pytest.org/en/7.1.x/explanation/fixtures.html#about-fixtures) are used to define resources used in tests.
 - Tests are implemented in methods or functions beginning with `test_` and classes beginning with `Test`.
 - The `assert` statement is used to compare the result of the test with an expected value.
 
+Although this specific example does not utilize fixtures, they are an essential part of pytest for defining reusable resources across tests. See [Fixtures](https://docs.pytest.org/en/7.1.x/explanation/fixtures.html#about-fixtures)
```
merelcht, line 108:
```suggestion
Although this specific example does not utilise fixtures, they are an essential part of pytest for defining reusable resources across tests. See [Fixtures](https://docs.pytest.org/en/7.1.x/explanation/fixtures.html#about-fixtures)
```

merelcht, line 108:
```suggestion
Although this specific example does not utilize fixtures, they are an essential part of pytest for defining reusable resources across tests. See [the pytest docs about fixtures](https://docs.pytest.org/en/7.1.x/explanation/fixtures.html#about-fixtures) for instructions on how to use them.
```

merelcht (approved):
Minor comments on the docs, but otherwise this looks good to me! 👍

PR #4389, opened 2024-12-23
Remove checkout branch check when not an official starter

### LINKED ISSUE #4388: `kedro new --starter <local template>` doesn't work offline
## Description

Reported by user on Slack: https://kedro.hall.community/kedro-new-command-fails-due-to-offline-environment-v8JlcFqqmr0T

The Kedro new command pointing to the local copy of a starter template doesn't work without an internet connection. 
From a brief investigation, it seems like this regression was introduced in #3900 in Kedro 0.19.7 where the fallback mechanism was added. When there's no connection, the function `_get_latest_starters_version()` returns an empty string which throws the `Invalid Version` error in https://github.com/kedro-org/kedro/blob/f6ecca509cf15e62a576ff2ac1abe80dd1121efe/kedro/framework/cli/starters.py#L133

ankatiyar:
## Description

Fix #4388 

## Development notes

Removed the check that happens in `_select_checkout_branch_for_cookicutter()` which was causing failures when the user is offline for cases when `--starter` is path to a template rather than an official starter.

## Checklist

### DIFF
--- a/kedro/framework/cli/starters.py
+++ b/kedro/framework/cli/starters.py
@@ -345,7 +345,6 @@ def new(  # noqa: PLR0913
         checkout = _select_checkout_branch_for_cookiecutter(checkout)
     elif starter_alias is not None:
         template_path = starter_alias
-        checkout = _select_checkout_branch_for_cookiecutter(checkout)
     else:
         template_path = str(TEMPLATE_PATH)
 
--- a/kedro/framework/cli/starters.py
+++ b/kedro/framework/cli/starters.py
@@ -519,7 +519,7 @@ def _get_available_tags(template_path: str) -> list:
         # tags: ['/tags/version', '/tags/version^{}']
         # unique_tags: {'version'}
 
-    except git.GitCommandError:
+    except git.GitCommandError:  # pragma: no cover
         return []
     return sorted(unique_tags)
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -5,6 +5,7 @@
 
 ## Bug fixes and other changes
 * Added validation to ensure dataset versions consistency across catalog.
+* Fixed a bug in project creation when using a custom starter template offline.
 
 ## Breaking changes to the API
 ## Documentation changes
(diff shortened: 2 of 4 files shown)


kedro/framework/cli/starters.py
```diff
@@ -345,7 +345,6 @@ def new(  # noqa: PLR0913
         checkout = _select_checkout_branch_for_cookiecutter(checkout)
     elif starter_alias is not None:
         template_path = starter_alias
-        checkout = _select_checkout_branch_for_cookiecutter(checkout)
```
DimedS, line 348: Does this mean that checkout will only be possible with official starters? What if someone uses a template from Git and wants to use the checkout feature? It seems like this won't be possible after the removal.

ElenaKhaustova, line 348: Implementation of `_select_checkout_branch_for_cookiecutter` for reference:
```
def _select_checkout_branch_for_cookiecutter(checkout: str | None) -> str:
    if checkout:
        return checkout
    elif _kedro_version_equal_or_lower_to_starters(version):
        return version
    else:
        return "main"
```

I see where the problem is. We call `_kedro_version_equal_or_lower_to_starters`which requires a remote connection in case when `checkout` was not set. But if we remove `_select_checkout_branch_for_cookiecutter` then checkout will remain `None` or an empty string. Is that a correct behaviour? Shall we add another check here instead, like `checkout = checkout or "main"`?

P.S. This code is really confusing overall and it definitely needs refactoring sometime in future 🫣

ElenaKhaustova: I think one can still provide the `checkout` argument via command line arguments.

ankatiyar: Like Elena said, it's not going to be changed to "main" if it's not provided but if the user has provided one, it will remain unchanged.

DimedS: Makes sense, thanks for the clarification. I think it's fine to remove it in that case. We definitely don't need to check the version online in this scenario.

DimedS (approved):
Thanks, @ankatiyar !

kedro/framework/cli/starters.py
ElenaKhaustova: 👍

Issue #4340, opened 2024-11-19
Running a pipeline mutates the original catalog object


FlorianGD:
## Description
When a pipeline is run, a `shallow_copy` of a catalog is made with the `extra_dataset_pattern` of the runner, and the original object is mutated. It adds an `extra_dataset_patterns` that is a "catch all". This prevents for example running a pipeline twice with the same catalog, as the second one will not have any free outputs.

## Context
I have a pipeline that exists with 2 namespaces. I wanted to test it for both, and I parametrized the tests. The tests used the same catalog as a fixture. This fixture had the `module` scope so as not to re create it every time. When updating to kedro 0.19, the test fails for the second parameter, when my test pipeline does not have any free output. After investigating, I found that the pipeline does not have any free output because the second time the catalog has `catalog._extra_dataset_patterns={'{default}': {'type': 'MemoryDataset'}}` which matches all the datasets.

## Steps to Reproduce

To reproduce the symptoms (running the pipeline twice does not have any free output the second time):

```python
>>> from kedro.io import DataCatalog
>>> from kedro.io.memory_dataset import MemoryDataset
>>> from kedro.pipeline import node, pipeline
>>> from kedro.runner import SequentialRunner
>>> my_pipeline = pipeline([node(lambda x: x, inputs=["foo"], outputs="bar")])
>>> catalog = DataCatalog()
>>> catalog.add("foo", MemoryDataset())
>>> catalog.save("foo", 1)
>>> runner = SequentialRunner()
>>> vars(catalog._config_resolver)
{
    '_runtime_patterns': {},
    '_dataset_patterns': {},
    '_default_pattern': {},
    '_resolved_configs': {}
}
>>> runner.run(my_pipeline, catalog)
{'bar': 1}
>>> vars(catalog._config_resolver)
{
    '_runtime_patterns': {'{default}': {'type': 'MemoryDataset'}},
    '_dataset_patterns': {},
    '_default_pattern': {},
    '_resolved_configs': {}
}
>>> runner.run(my_pipeline, catalog)
{}
```
I think the problem comes from this behavior:
```python
>>> catalog = DataCatalog()
>>> runner = SequentialRunner()
>>> catalog.config_resolver.list_patterns()
[]
>>> catalog.shallow_copy(extra_dataset_patterns=runner._extra_dataset_patterns)
None
>>> catalog.config_resolver.list_patterns()
['{default}']
```

## Expected Result
I want that `bar` was kept as a free output when running the pipeline the second time

[...]

PR #4429, opened 2025-01-20
Remove runtime patterns after run

ElenaKhaustova:
## Description
Solves https://github.com/kedro-org/kedro/issues/4340

## Development notes
Added a method to remove runtime patters after the run so they live only within the run and do not affect other runs.

This solution originally was suggested in https://github.com/kedro-org/kedro/pull/4236 but was replaced with another one which solved the original problem but not catalog mutation.

## Checklist


kedro/io/catalog_config_resolver.py
```diff
@@ -332,3 +332,10 @@ def add_runtime_patterns(self, dataset_patterns: Patterns) -> None:
         """Add new runtime patterns and re-sort them."""
         self._runtime_patterns = {**self._runtime_patterns, **dataset_patterns}
         self._runtime_patterns = self._sort_patterns(self._runtime_patterns)
+
+    def remove_runtime_patterns(self, dataset_patterns: Patterns) -> None:
+        """Remove runtime patterns and re-sort them."""
+        for pattern_name in dataset_patterns:
+            if pattern_name in self._runtime_patterns:
+                del self._runtime_patterns[pattern_name]
+        self._runtime_patterns = self._sort_patterns(self._runtime_patterns)
```
merelcht, line 341: Which patterns will still be left here to sort? 🤔

tests/io/test_kedro_data_catalog.py
```diff
@@ -291,6 +291,15 @@ def test_get_dataset_matching_pattern(self, data_catalog):
         ds = data_catalog.get_dataset(match_pattern_ds)
         assert isinstance(ds, MemoryDataset)
 
+    def test_remove_runtime_pattern(self, data_catalog):
```
merelcht, line 294: Can you also update `test_run_twice_giving_same_result` which you added in the previous PR to fix this issue to verify the catalog isn't changed unexpectedly?

merelcht (commented):
> This solution originally was suggested in https://github.com/kedro-org/kedro/pull/4236 but was replaced with another one which solved the original problem but not catalog mutation.

Why was this solution replaced in the first PR? I couldn't quite figure it out from the thread there.

ElenaKhaustova:
> > This solution originally was suggested in #4236 but was replaced with another one which solved the original problem but not catalog mutation.
> 
> Why was this solution replaced in the first PR? I couldn't quite figure it out from the thread there.

There were actually two bugs and both of them were related to the way we process MemoryDataset's and how we define free outputs. The first reported by the user could be resolved by removing patterns but the second could not, so another solution was implemented to handle them both. But mutation remained as we didn't go with the patterns removing solution. 

Here is the comment about the implemented solution https://github.com/kedro-org/kedro/pull/4236#issuecomment-2422546681

The previous solution also covers the current bug, but we still think mutation looks confusing and should not happen, so in this PR, we remove patterns added at runtime.

kedro/io/catalog_config_resolver.py
ElenaKhaustova: Technically, we remove only those that passed to the method, and we don't care if they remove all, as it is controlled from the outside. So, in case something remains, we sort it.

Otherwise, we can remove all of them without accepting input for this method.

I preferred this, but both options look alright to me, so I don't mind updating if the current one looks confusing.

merelcht: No makes sense. I was just wondering if anything at all would be left.

tests/io/test_kedro_data_catalog.py
ElenaKhaustova: Done

merelcht (approved):
LGTM 👍

merelcht (approved):

lrcouto (approved):
Tested it here, works fine!

Issue #4436, opened 2025-01-21
`kedro catalog list` fails when using `KedroDataCatalog`


ElenaKhaustova:
## Description
`kedro catalog list` command fails with `KeyError` when using new `KedroDataCatalog`.

## Context
The issue appeared in the 0.19.10 version after lazy dataset initializing was merged, as we did not update dataset access after and only considered initialized datasets for `KedroDataCatalog`.

https://github.com/kedro-org/kedro/blob/fba7c53d7b594c6da30cc012e17a04f90ddc9c7d/kedro/framework/cli/catalog.py#L70

## Steps to Reproduce
1. Use kedro spaceflights-pandas project template
2. Update settings to use new `KedroDataCatalog`
3. Run `kedro catalog list`

## Expected Result
No errors occur

## Actual Result

<img width="943" alt="Image" src="https://github.com/user-attachments/assets/9b845286-8ea4-4164-87eb-bbc5551a83ba" />

## Your Environment

* Kedro version used (`pip show kedro` or `kedro -V`): kedro, version 0.19.10
* Python version used (`python -V`): Python 3.11.11

PR #4438, opened 2025-01-23
Fix `catalog._datasets` access for `KedroDataCatalog`

ElenaKhaustova:
## Description
Solves https://github.com/kedro-org/kedro/issues/4436

## Development notes
After we implemented lazy dataset loading for `KedroDataCatalog`, we differed internal attributes `_datasets` and `_lazy_datasets` and started tracking them separately. At the same time, for old `DataCatalog,` we extensively used the internal `_datasets` attribute across the codebase, which caused a compatibility issue as the `_datasets` attribute stored a different set of datasets for old (all datasets) and new catalogs (all materialized datasets).

This behaviour didn't affect `kedro run` as we did a warm-up before the run, so needed datasets were materialized and there was no discrepancy.

We fixed that at the `KedroDataCatalog` level and left TODOs for when we switch to the new catalog completely:

1. Added `__datasets` property for `KedroDataCatalog`, so we could still differ materialized and lazy datasets at the catalog level
2. Made `KedroDataCatalog.datatest` return both materialized and lazy datasets
3. Updated `__getattribute__` for `KedroDataCatalog`, so if someone calls `KedroDataCatalog._datatest` they get the same result as `KedroDataCatalog.datatest`. We could add just `_datasets` property but for that, we needed to modify `CatalogProtocol` which we preferred to avoid.
4. Added a test to check that when having materialized and lazy datasets both `KedroDataCatalog.datatest` and `KedroDataCatalog._datatest` return the same.

**Note:** These changes are temporary and needed for compatibility with the old catalog. They will be removed when migrating to the new catalog; corresponding TODOs were left.

## Checklist


kedro/io/kedro_data_catalog.py
```diff
@@ -505,7 +514,7 @@ def get_dataset(
             # Flag to turn on/off fuzzy-matching which can be time consuming and
             # slow down plugins like `kedro-viz`
             if suggest:
-                matches = difflib.get_close_matches(ds_name, self._datasets.keys())
+                matches = difflib.get_close_matches(ds_name, self.keys())
```
merelcht, line 517: Why did you replace `self._datasets.keys()` here with `self.keys()` instead of `self.__datasets.keys()`?

merelcht (approved):
Perhaps add a test to check that `kedro catalog list` now works? 

Otherwise LGTM 👍

kedro/io/kedro_data_catalog.py
ElenaKhaustova: `self.__datasets` do not include lazy loaded datasets but only materialized. `self.keys()` include both `self.__datasets` and `self._lazy_datasets`.

ElenaKhaustova:
@merelcht 

> Perhaps add a test to check that kedro catalog list now works?

I think we might need to tackle this for all commands together. Now all catalog CLI commands unit tests only run on the old catalog.

I tried it, but because of the way catalog CLI tests are organized, I couldn't do it in a reasonable time. I think we need to create a separate ticket to address that, so this one doesn't block the release in case it takes time.

merelcht:
> I tried it, but because of the way catalog CLI tests are organized, I couldn't do it in a reasonable time. I think we need to create a separate ticket to address that, so this one doesn't block the release in case it takes time.

Sounds good. Can you create the ticket?

ElenaKhaustova:
> Now all catalog CLI commands unit tests only run on the old catalog.

https://github.com/kedro-org/kedro/issues/4442

ElenaKhaustova:
> Sounds good. Can you create the ticket?

https://github.com/kedro-org/kedro/issues/4442

lrcouto (approved):
LGTM!

PR #4449, opened 2025-01-28
Implement catalog filter for `KedroDataCatalog`

ElenaKhaustova:
## Description
Implemented `KedroDataCatalog.filter()` to filter datasets by name and type.

Solves https://github.com/kedro-org/kedro/issues/3917

Reasoning: https://github.com/kedro-org/kedro/issues/3917#issuecomment-2612578566

## Development notes
Implemented on top of https://github.com/kedro-org/kedro/pull/4448  <-- please review it first.

## Checklist


merelcht (approved):
Already discussed on the call that this looks good! 👍 Left some minor comments.

kedro/io/kedro_data_catalog.py
```diff
@@ -745,3 +813,12 @@ def exists(self, name: str) -> bool:
         except DatasetNotFoundError:
             return False
         return dataset.exists()
+
+
+def _compile_pattern(regex: str, regex_flags: int | re.RegexFlag) -> re.Pattern:
```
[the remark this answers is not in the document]
ElenaKhaustova: Good point, will rename it

tests/io/test_kedro_data_catalog.py
```diff
@@ -175,6 +177,61 @@ def test_multi_catalog_list_bad_regex(self, multi_catalog):
...
+        )
+
+    def test_catalog_filter_bad_regex(self, multi_catalog):
+        """Test that bad regex is caught accordingly"""
+        escaped_regex = r"\(\("
+        pattern = f"Invalid regular expression provided: '{escaped_regex}'"
+        with pytest.raises(SyntaxError, match=pattern):
+            multi_catalog.filter(name_regex="((")
+        with pytest.raises(SyntaxError, match=pattern):
+            multi_catalog.filter(type_regex="((")
```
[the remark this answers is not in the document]
ElenaKhaustova: The first check is for `name_regex` the second is for `type_regex` 🙂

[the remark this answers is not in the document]
merelcht: ahhhh now I see it! 😂

ankatiyar (approved):
manually tested, looks good! 👍🏾

ElenaKhaustova:
After the conversation with @idanov, we concluded that this method doesn't bring much new functionality, so it makes sense to keep the old naming. However, we're still unsure about its necessity.

Our initial thinking was that the original `list()` method doesn't give any value since it allows filtering only by dataset names, which can now be easily done by applying a regex filter to `key()`. That’s why we suggested extending the functionality to filter by kind, as users might struggle to access dataset types given that now we have lazy and materialized datasets. So, there might be an additional value in “by kind” filtering.

Based on the above we still have two options:

- Remove `list()` method
- Keep `list()` method but extend it with by_type (which basically means just renaming `filter()` -> `list()` it in the current implementation)

Please share your feedback on these two points (naming and necessity) 🙏

datajoely:
> I'd vote for `Keep list() method but extend it with by_type (which basically means just renaming filter() -> list() it in the current implementation)`

The 3 things I don't like as a user today:

- Unable to filter by type
- Namespaces are weird (__ internal representation)
- Factories aren't visible until used

astrojuanlu:
> which can now be easily done by applying a regex filter to `keys()`

I don't follow, could you elaborate? How would the code would look like with and without this PR?

ElenaKhaustova:
> > which can now be easily done by applying a regex filter to `keys()`
> 
> I don't follow, could you elaborate? How would the code would look like with and without this PR?

```python
pattern = re.compile(regex, flags=regex_flags)
filtered_ds = [ds_name for ds_name in catalog.keys() if pattern.search(ds_name)] #  or iterate via values/items
```

The point is that since dataset names can be easily accessed, it's also easy to apply custom filtering on top, and we don't need a dedicated method for it. However, it's a bit more complicated with dataset types, so filtering by types could add some value.

kedro/io/kedro_data_catalog.py
```diff
@@ -551,11 +551,82 @@ def add(
             )
         self.__setitem__(ds_name, dataset)
 
+    def filter(
+        self,
+        name_regex: str | None = None,
+        name_regex_flags: int | re.RegexFlag = re.IGNORECASE,
+        type_regex: str | None = None,
```
marrrcin, line 558: When I've read the PR description about _filter by type_ I though that will be basically `instanceof` filtering, allowing to pass `Type` parameter, not string based class name, moreover a fully qualified one, which is more laborious to obtain for a plugin developer than just doing `data_catalog.filter(by_type=PluginSpecificDataset)`. WDYT?

astrojuanlu:
Thanks @ElenaKhaustova , that's what I imagined. In my view, given that our users aren't necessarily software engineers, our APIs shouldn't just be perfect [partitions](https://en.wikipedia.org/wiki/Partition_of_a_set) of the set of possible use cases, but allow for some convenience methods that, as @datajoely says, alleviate the burden a bit.

kedro/io/kedro_data_catalog.py
datajoely: Oh I hadn't even considered that, I thought it would be the string representation people do in the YAML config! To be honest, both would be useful

merelcht:
I still feel like `filter()` is a more descriptive name and makes it more obvious that you can provide filter arguments. Whereas with `list` I wouldn't really expect to be able to add a regex argument. 

So my vote would be: introduce `filter` now and remove `list` when we do 1.0.0

ElenaKhaustova:
> Thanks @ElenaKhaustova , that's what I imagined. In my view, given that our users aren't necessarily software engineers, our APIs shouldn't just be perfect [partitions](https://en.wikipedia.org/wiki/Partition_of_a_set) of the set of possible use cases, but allow for some convenience methods that, as @datajoely says, alleviate the burden a bit.

Could you please clarify whether you mean keeping the existing `list` implementation?

ElenaKhaustova:
From the feedback and discussion here, we're going to:

1. Keep `filter()` name
2. Extend the current `filter()` implementation by accepting actual types

RELEASE.md
```diff
@@ -1,6 +1,8 @@
 # Upcoming Release
 
 ## Major features and improvements
+* Implemented `KedroDataCatalog.filter()` to filter datasets by name and type.
```
merelcht, line 4:
```suggestion
* Added `KedroDataCatalog.filter()` to filter datasets by name and type.
```

kedro/io/kedro_data_catalog.py
```diff
@@ -551,11 +551,100 @@ def add(
...
+
+        Args:
+            name_regex: Optional regular expression to filter dataset names by name.
+            name_regex_flags: Optional regex flags for the name filter.
+                By default, IGNORECASE key is set.
+            type_regex: Optional regular expression to filter dataset names by their type.
+                The provided regex is matched against the full dataset type path, for example:
+                `kedro_datasets.pandas.parquet_dataset.ParquetDataset`.
+            type_regex_flags: Optional regex flags for the type filter.
+            by_type: Optional dataset types to filter by.
```
merelcht, line 576: I think it would be helpful to give an example of a `by_type` argument. And maybe clarify that this is an instance of type of check instead of regex match? And also clarify it can take a single type or list. I know people can also read that from the function signature, but better to be extra clear IMO.

merelcht (approved):
LGTM 👍 

I really like the addition of the `by_type` filter on top of the regex for type.

kedro/io/kedro_data_catalog.py
```diff
@@ -551,11 +551,100 @@ def add(
...
+            >>> from kedro_datasets.pandas import SQLQueryDataset
+            >>> from kedro.io import MemoryDataset
+            >>> catalog.filter(name_regex="data", by_type=[MemoryDataset, SQLQueryDataset])
+        """
+        filtered = self.keys()
+
+        # Apply name filter if specified
+        if name_regex:
+            pattern = _compile_regex_pattern(name_regex, name_regex_flags)
+            filtered = [ds_name for ds_name in filtered if pattern.search(ds_name)]
```
astrojuanlu, line 609: Instead of compiling the regex on behalf of the user, shouldn't we do `re.search(pattern, ds_name)`? That way we don't need to receive `*_regex_flags` and we don't need the compilation step either.

Use case: Imagine compiling the regex is very expensive. But the `.filter()` function is used in a hot loop. We would be compiling a regex many times unnecessarily. Instead, the user can `.compile` the regex outside of the function and pass it. And if they don't care, `re.search` will work with a normal string too.

Thoughts?

ElenaKhaustova: I agree that it will be easier for us from the implementation point of view. But I can hardly imagine such cases when compiling regex will be expensive, just because catalogs do not include millions of datasets, and their names mostly include alphabet characters plus some others.

The suggestion makes sense to me but I think we should do what's more convenient for users from the API point of view.

ElenaKhaustova: For me, this will not make much difference, so I probably agree to replace it with a compiled pattern.

astrojuanlu: It reduces the API surface, has potential (yet possibly minimal) performance benefits, simplifies the implementation, and doesn't remove any functionality (the user can still pass a string). So if there are no disagreements I'd say let's do it.

astrojuanlu (approved):
💯

ankatiyar (approved):

Issue #4486, opened 2025-02-17
`SequentialRunner` might fail with non thread-safe code


astrojuanlu:
## Description
Since Kedro 0.19.11, `SequentialRunner` was refactored and is known to occasionally fail or produce unexpected behavior with code that is not thread-safe. See https://github.com/kedro-org/kedro/pull/4353 for more details.

Existing issues related to this:
- https://github.com/Galileo-Galilei/kedro-mlflow/issues/624
- https://github.com/kedro-org/kedro-plugins/issues/1012

This is technically a breaking change, and we are trying to gauge user impact. If this is a problem for you, please leave a comment below.

astrojuanlu:
@merelcht I think we should treat this as a breaking change and try to address it or revert it ahead of the next release, since it's causing disruption that's often difficult to spot and holding users back from upgrading to 0.19.11.

astrojuanlu:
Another way I've seen this manifest is that `kedro run` doesn't immediately react to `SIGINT`. Randomly found it when trying a hook that did `time.sleep(10)`.

jakepenzak:
We noticed this issue for our team when running our pipeline in databricks, where it manifested as a failure of the spark session to be accessed in certain nodes/imported modules. Admittedly, I don't fully understand the exact mechanics behind this failure (outside of spark session not being fully thread safe?), but I brought it up in a [slack convo](https://kedro-org.slack.com/archives/C03RKP2LW64/p1740072376782789) and was able to link it to this specific issue.
 
Here is the excerpt for those not on slack: 

> Hello! I am running into issues with Kedro 0.19.11 release while running pipelines in databricks. Specifically, I am running into an error where an imported python module for a node is unable to find active SparkSession via SparkSession.getActiveSession(). Our pipeline is comprised entirely of Ibis.TableDataset datasets & I/O with pyspark backend. What is throwing me is that other nodes use the pyspark connection and are able to perform operations properly across the spark session, but fails on this single node when leveraging an imported module that it is unable to find the spark session. This issue is not present in Kedro 0.19.10. My best guess is that it has something to do with the updated code in kedro/runner/sequential_runner.py using ThreadPoolExecutor and possible scoping issues? Apologies on the somewhat scattered explanation, there is quite a bit I don't fully understand here, so appreciate any help or guidance. Lmk if I can provide any additional info as well.

PR #4502, opened 2025-02-19
Change runner logic to not create pool for sequential runner

merelcht:
## Description
Address #4486 

## Checklist

### DIFF
--- a/kedro/runner/runner.py
+++ b/kedro/runner/runner.py
@@ -189,8 +189,8 @@ class AbstractRunner(ABC):
         return self.run(to_rerun, catalog, hook_manager)
 
     @abstractmethod  # pragma: no cover
-    def _get_executor(self, max_workers: int) -> Executor:
-        """Abstract method to provide the correct executor (e.g., ThreadPoolExecutor or ProcessPoolExecutor)."""
+    def _get_executor(self, max_workers: int) -> Executor | None:
+        """Abstract method to provide the correct executor (e.g., ThreadPoolExecutor, ProcessPoolExecutor or None if running sequentially)."""
         pass
 
     @abstractmethod  # pragma: no cover
@@ -226,7 +226,30 @@ class AbstractRunner(ABC):
         done = None
         max_workers = self._get_required_workers_count(pipeline)
 
-        with self._get_executor(max_workers) as pool:
+        pool = self._get_executor(max_workers)
+        if pool is None:
+            for exec_index, node in enumerate(nodes):
+                try:
+                    Task(
+                        node=node,
+                        catalog=catalog,
+                        hook_manager=hook_manager,
+                        is_async=self._is_async,
+                        session_id=session_id,
+                    ).execute()
+                    done_nodes.add(node)
+                except Exception:
+                    self._suggest_resume_scenario(pipeline, done_nodes, catalog)
+                    raise
+                self._logger.info("Completed node: %s", node.name)
+                self._logger.info(
+                    "Completed %d out of %d tasks", len(done_nodes), len(nodes)
+                )
+                self._release_datasets(node, catalog, load_counts, pipeline)
+
+            return  # Exit early since everything runs sequentially
+
+        with pool as executor:
             while True:
                 ready = {n for n in todo_nodes if node_dependencies[n] <= done_nodes}
                 todo_nodes -= ready
@@ -238,9 +261,9 @@ class AbstractRunner(ABC):
                         is_async=self._is_async,
                         session_id=session_id,
                     )
-                    if isinstance(pool, ProcessPoolExecutor):
+                    if isinstance(executor, ProcessPoolExecutor):
                         task.parallel = True
-                    futures.add(pool.submit(task))
+                    futures.add(executor.submit(task))
                 if not futures:
                     if todo_nodes:
                         self._raise_runtime_error(todo_nodes, done_nodes, ready, done)
--- a/kedro/runner/sequential_runner.py
+++ b/kedro/runner/sequential_runner.py
@@ -5,10 +5,6 @@ of provided nodes.
 
 from __future__ import annotations
 
-from concurrent.futures import (
-    Executor,
-    ThreadPoolExecutor,
-)
 from typing import TYPE_CHECKING, Any
 
 from kedro.runner.runner import AbstractRunner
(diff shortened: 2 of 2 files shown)


kedro/runner/runner.py
```diff
@@ -226,7 +226,30 @@ def _run(
...
+                    raise
+                self._logger.info("Completed node: %s", node.name)
+                self._logger.info(
+                    "Completed %d out of %d tasks", len(done_nodes), len(nodes)
+                )
+                self._release_datasets(node, catalog, load_counts, pipeline)
+
+            return  # Exit early since everything runs sequentially
+
+        with pool as executor:
```
DimedS, line 252: Could you please clarify why it is initially called a pool but later renamed to an executor? Why not use executor from the beginning?

DimedS (approved):
Thanks, @merelcht! That’s a more robust approach. One small question from my side

kedro/runner/runner.py
merelcht: >  with pool as executor:

I can't name them both the same in the context manager, so it's a bit arbitrary and I could do it the other way round if that makes more sense. Or something like `with executor as executor_context:` WDYT?

DimedS: Oh, I see, I forgot that `as` is required. Thanks for the clarification! Let's leave it as it is.

kedro/runner/runner.py
```diff
@@ -226,7 +226,30 @@ def _run(
         done = None
         max_workers = self._get_required_workers_count(pipeline)
 
-        with self._get_executor(max_workers) as pool:
+        pool = self._get_executor(max_workers)
```
ElenaKhaustova, line 229: It looks good to me from a logical point of view.

But I would suggest moving this into `SequentialRunner._run`. Otherwise, we modify the behaviour of the base class based on what is inherited from it, which is not entirely correct from the implementation point of view and `AbstractRunner._run` becomes too long. I understand that it will require some duplication, but in the `SequentialRunner._run` method, we can add a note explaining why we keep the implementation like that. But adding it to `AbstractRunner._run` will overload it even more.

merelcht: I don't have a very strong opinion on this, but my counter argument is then wouldn't it be confusing that the thread and parallel logic is in the `AbstractRunner._run` method but sequential isn't?

merelcht:
I addressed @ElenaKhaustova 's comment about moving the sequential logic to the `SequentialRunner` in https://github.com/kedro-org/kedro/pull/4502/commits/d6a247cf848537bd5d779121defba3f6f892459e . I personally don't have a huge preference for doing it that way or keeping it inside the abstract runner. They both have pros and cons. My main con for having sequential runner override the run method is only that it might be confusing that abstract runner has the logic for thread an parallel execution but not sequential, but arguably it's a bit cleaner. 

@DimedS any thoughts?

kedro/runner/runner.py
```diff
@@ -226,38 +226,44 @@ def _run(
...
-                        self._suggest_resume_scenario(pipeline, done_nodes, catalog)
-                        raise
-                    done_nodes.add(node)
-                    self._logger.info("Completed node: %s", node.name)
-                    self._logger.info(
-                        "Completed %d out of %d tasks", len(done_nodes), len(nodes)
-                    )
-                    self._release_datasets(node, catalog, load_counts, pipeline)
+        pool = self._get_executor(max_workers)
+        if pool is not None:
```
astrojuanlu, line 230: Not sure if it's related to what @merelcht is saying, but for the record I was skimming this code and thought "what if `pool is None`?" (there's no `else` branch here). Just a comment from the peanut gallery.

merelcht: Yeah exactly! I think it might not be immediately obvious that that logic is now inside `SequentialRunner`.

kedro/runner/runner.py
DimedS: I agree with Elena that, from a Pythonic perspective, common logic should be placed in `_run()` within the Abstract class. If there are exceptions, we should override the common logic with specific behavior, which is how runners worked previously. However, I thought the goal of the previous PR, which Merel is currently modifying, was to centralise the runner's logic within the `Abstract` class. 

We had already decided that the `_run()` function in the abstract class would rely on `_get_executor()`, which would be implemented specifically in different subclasses. I don't see any issues with this approach. For me, the main question is how large and readable `AbstractRunner._run()` will be. As Merel pointed out, consolidating all the running logic in one place will be beneficial, which was also the intention of the previous PR.

DimedS:
> I addressed @ElenaKhaustova 's comment about moving the sequential logic to the `SequentialRunner` in [d6a247c](https://github.com/kedro-org/kedro/commit/d6a247cf848537bd5d779121defba3f6f892459e) . I personally don't have a huge preference for doing it that way or keeping it inside the abstract runner. They both have pros and cons. My main con for having sequential runner override the run method is only that it might be confusing that abstract runner has the logic for thread an parallel execution but not sequential, but arguably it's a bit cleaner.
> 
> @DimedS any thoughts?

I agree, I also think that, given the current state, it would be beneficial to keep all the running logic in the Abstract class. I have provided a full comment in the thread: https://github.com/kedro-org/kedro/pull/4502#discussion_r1970236047.

kedro/runner/runner.py
ElenaKhaustova: Yeah, I see the points and I would totally agree with them if the implementation remained the same (without) a separate branch for `SequentialRunner`.
My main points are:

1. Currently base class behaviour is controlled by child classes. I think we came to the current state because of the initial design and changes we make now. But if we designed it from scratch we would probably not have implemented it that way.
2. I would argue on readability too. First of all abstract `_run` is already quite long, second the implementation might be clear to us - who already have some context. But if you first look at it for example in order to understand how `ThreadRunner` works one would also need to look at the `SequentialRunner` and at the `is None` branch.

I would still prefer splitting the `SequentialRunner` logic. However, I don't want to block the PR, as the functionality is there anyway, so I'm leaving it up to the PR's author.

kedro/runner/runner.py
ElenaKhaustova: Here we could just raise an error saying that abstract `_run` requires executor set.

ElenaKhaustova (approved):
Approving with a comment left in the thread.

kedro/runner/runner.py
merelcht: @ElenaKhaustova I totally get what you're saying, but when it comes to readability this is also partly preference: for me the code actually became a bit harder to understand after moving out the sequential logic, because you then need to also navigate to the other runner classes to understand which one overwrite the abstract logic and which don't and how. Anyway, I'll merge it as is now and we might be able to address this better when doing the refactoring for decoupling the catalog logic.

Issue #4469, opened 2025-02-07
Deprecate `DataCatalog`


merelcht:
### Description
Add deprecation warnings for `DataCatalog`, so that users will be aware it's going to be removed in `1.0.0` and replaced by `KedroDataCatalog`.

PR #4510, opened 2025-02-21
Add `DataCatalog` deprecation warning

ElenaKhaustova:
## Description
Solves https://github.com/kedro-org/kedro/issues/4469

## Checklist


ankatiyar (approved):
LGTM!
It shows up twice now when you do `kedro run` but I suppose it's because of the deepcopy?

ElenaKhaustova:
> LGTM! It shows up twice now when you do `kedro run` but I suppose it's because of the deepcopy?

That's because we recreate catalog object to inject patterns for the old catalog: https://github.com/kedro-org/kedro/blob/9c70baeb895966c889ca9da59de7735e8e4c67d4/kedro/runner/runner.py#L114

astrojuanlu (commented):
I'm thinking: If a user is using 0.19.11 already, what would be the best way to make the code forward compatible?

Something like this?

```python
try:
    import KedroDataCatalog as DataCatalog
except ImportError:
    import DataCatalog
```

- If old Kedro, then old `DataCatalog`
- If 0.19.x Kedro, then `KedroDataCatalog`
- If Kedro >=1.x, then new `DataCatalog`

However, is `KedroDataCatalog` meant to  be a 100 % compatible drop-in replacement?

I understand that this is not as terrible as the `DataSet` migration, but still worth thinking how the `settings.py` etc should look like for folks that have an explicit `DATA_CATALOG_CLASS` set and so on

ElenaKhaustova:
> I'm thinking: If a user is using 0.19.11 already, what would be the best way to make the code forward compatible?
> 
> Something like this?
> 
> ```python
> try:
>     import KedroDataCatalog as DataCatalog
> except ImportError:
>     import DataCatalog
> ```
> 
> * If old Kedro, then old `DataCatalog`
> * If 0.19.x Kedro, then `KedroDataCatalog`
> * If Kedro >=1.x, then new `DataCatalog`
> 
> However, is `KedroDataCatalog` meant to be a 100 % compatible drop-in replacement?
> 
> I understand that this is not as terrible as the `DataSet` migration, but still worth thinking how the `settings.py` etc should look like for folks that have an explicit `DATA_CATALOG_CLASS` set and so on

They are compatible now but that won't be the case after replacement, because we will remove all the temporal API kept for compatibility. Even if we suggest something now for importing to be forward-compatible it will not help much because of the API changes.

kedro/io/data_catalog.py
```diff
@@ -160,6 +162,12 @@ def __init__(  # noqa: PLR0913
             >>>                   save_args={"index": False})
             >>> catalog = DataCatalog(datasets={'cars': cars})
         """
+        warnings.warn(
+            "`DataCatalog` has been deprecated and will be replaced by an improved alternative, `KedroDataCatalog`, in Kedro 1.0.0."
+            "After this change, the `DataCatalog` name will persist, but its functionality will align with `KedroDataCatalog`. "
+            "For more details, refer to the documentation: https://docs.kedro.org/en/stable/data/index.html#kedrodatacatalog-experimental-feature",
```
merelcht, line 168:
```suggestion
            "`DataCatalog` has been deprecated and will be replaced by `KedroDataCatalog` in Kedro 1.0.0."
            "After this change, the `DataCatalog` name will persist, but its functionality will align with `KedroDataCatalog`. "
            "For more details, refer to the documentation: https://docs.kedro.org/en/stable/data/index.html#kedrodatacatalog-experimental-feature",
```

While I agree it's an improvement it's up to the user to decide if it is 😉 this way the message is shorter as well.

merelcht (approved):

astrojuanlu:
> > I'm thinking: If a user is using 0.19.11 already, what would be the best way to make the code forward compatible?
> > Something like this?
> > ```python
> > try:
> >     import KedroDataCatalog as DataCatalog
> > except ImportError:
> >     import DataCatalog
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * If old Kedro, then old `DataCatalog`
> > * If 0.19.x Kedro, then `KedroDataCatalog`
> > * If Kedro >=1.x, then new `DataCatalog`
> > 
> > However, is `KedroDataCatalog` meant to be a 100 % compatible drop-in replacement?
> > I understand that this is not as terrible as the `DataSet` migration, but still worth thinking how the `settings.py` etc should look like for folks that have an explicit `DATA_CATALOG_CLASS` set and so on
> 
> They are compatible now but that won't be the case after replacement, because we will remove all the temporal API kept for compatibility. Even if we suggest something now for importing to be forward-compatible it will not help much because of the API changes.

Got it. Then maybe what we can do is, in Kedro 1.0.x, keep the `KedroDataCatalog`, but in a deprecated form. Something like

```
>>> from kedro.io import KedroDataCatalog
This class is only kept for compatibility purposes and it will be removed in Kedro 2.0,
please do `from kedro.io import DataCatalog` instead
```

That way, code that's written to work with `KedroDataCatalog` in 0.19 will continue to work in 1.0 without changes.

WDYT?

merelcht:
> Got it. Then maybe what we can do is, in Kedro 1.0.x, keep the `KedroDataCatalog`, but in a deprecated form. Something like
> 
> ```
> >>> from kedro.io import KedroDataCatalog
> This class is only kept for compatibility purposes and it will be removed in Kedro 2.0,
> please do `from kedro.io import DataCatalog` instead
> ```
> 
> That way, code that's written to work with `KedroDataCatalog` in 0.19 will continue to work in 1.0 without changes.
> 
> WDYT?

Isn't this a bit of a deprecation overkill? My guess is that only a tiny percentage of people are already using `KedroDataCatalog` and the change they have to do is minimal. Do we really want to keep a deprecation message in our code-base all the way through the 1.0.0 series for this?

astrojuanlu:
> Do we really want to keep a deprecation message in our code-base all the way through the 1.0.0 series for this?

It's a burden, I know... but otherwise it feels like we're punishing users that are up to date 😓 Compare:

- User A: Using Kedro 0.19.3, they have no idea about the `KedroDataCatalog` and use `DataCatalog` for everything. Later they upgrade directly to `1.0`, they might have to make some adjustments to their code because of the (expected) breaking changes but in the best case scenario everything continues to work.
- User B: Upgrading to every micro release. In Kedro 0.19.12 they see the deprecation warning, change their code to use `KedroDataCatalog` instead. Then they upgrade to Kedro 1.0 and their code stops working, they have to rename it to `DataCatalog` again. Alternatively, they can ignore the `KedroDataCatalog` hoping that everything continues to work.

My question is how do we communicate to User B what's the best way to make a smooth transition?

ElenaKhaustova:
@astrojuanlu, @merelcht 

I think the key problem here is that using `KedroDataCatalog` now does not guarantee breaking changes won't affect their code in 1.0.0. Since `KedroDataCatalog` include old `DataCatalog` APIs. So only those who moved to `KedroDataCatalog` **and** new API won't be affected which should be very small number if any exist. So keeping both names until 2.0.0 won't be helpful in practice.

I don't think we punish User B, since in the deprecation warning we explain that the old naming `DataCatalog` will remain. What might be missing in the explanation is that a few `KedroDataCatalog` methods related to the old catalog will be removed.

Possibly we could add deprecation warnings for specific methods too, suggesting replacement with a new API, though this might produce too much printing when running pipelines. The alternative is to extend the `DataCatalog` deprecation message with something like:

```
"`DataCatalog` has been deprecated and will be replaced by `KedroDataCatalog` in Kedro 1.0.0."
"After this change, the `DataCatalog` name will persist, but its functionality will align with `KedroDataCatalog`. "
"Some `KedroDataCatalog` APIs have been retained for compatibility with `DataCatalog`, including the `datasets` property and the `get_datasets`, `_get_datasets`, `add`,` list`, `add_feed_dict`, and `shallow_copy` methods. These will be removed or replaced with updated alternatives in Kedro 1.0.0."
"For more details, refer to the documentation: https://docs.kedro.org/en/stable/data/index.html#kedrodatacatalog-experimental-feature"
```

The suggestion to keep the `KedroDataCatalog` name looks good to me as well, but again, in 1.0.0, the main problem will be the API update, not the name.

astrojuanlu:
Okay, so if I understand correctly then there will be

- The 0.19 `DataCatalog` that everybody knows
- A short-lived `KedroDataCatalog` that won't exist in 1.0
- The 1.0 `DataCatalog` which will be incompatible with 0.19 `DataCatalog` in subtle ways

> I don't think we punish User B, since in the deprecation warning we explain that the old naming DataCatalog will remain.

The naming will remain, but the API will change. And with this plan there's a transition period in which there's not a very clear incentive to using the `KedroDataCatalog` because users will need to modify their code again when 1.0 is out. Won't this hinder our ability to detect problems with the 2.0 catalog early?

In the end, this is similar to a `from __future__ import` mechanism of sorts.

> Since KedroDataCatalog include old DataCatalog APIs. [...]
> Possibly we could add deprecation warnings for specific methods too, suggesting replacement with a new API, though this might produce too much printing when running pipelines.

And for this reason it's difficult at the moment to know what's old API and what is new without comparing the code side by side 😕 

Is it worth retaining the compatibility shims of `KedroDataCatalog`? Wouldn't it be easier if folks could already anticipate what the 1.0 `DataCatalog` will look like, instead of us removing functionality in the `develop` branch and then users discovering by surprise? Or maybe we could retain them, but mark them all (the old APIs in the new catalog) as deprecated?

I hope I'm not sounding like a pain in the neck but after looking at how much friction the uppercase-DataSet caused for users migrating to Kedro 0.19 and kedro-datasets, and given that we still have time, I'm hoping that we can plan plan this transition with an eye on the user experience.

ElenaKhaustova:
Realistically I think the only users who might be affected are plugin developers - those who utilize kedro components directly. And they all keep using the old catalog and waiting for the 1.0.0 release to transit to the new one.

The rest are using the catalog as part of the framework and to debug. The first requires setting `DATA_CATALOG_CLASS = KedroDataCatalog` in the `settings.py`, so replacing the class name won't be a big deal. As for debugging, they'll probably be happy to get rid of private `_get_dataset()` and replace `list()` with `filter()` as well as a few other methods.

If we keep maintaining the layer needed for old and new catalog compatibility, it will affect other components too, like `session`  and `runners` and prevent us from making changes there.

I see the concerns shared, but the scale doesn't look comparable with the uppercase-DataSet update, so maybe we should not overcomplicate the transition.

Another argument is if you keep maintaining both APIs people will never transition to a new one. They will only do that if it stops working.

merelcht:
> Okay, so if I understand correctly then there will be

> The 0.19 DataCatalog that everybody knows
> A short-lived KedroDataCatalog that won't exist in 1.0

My point is to **not** rename `KedroDataCatalog` to `DataCatalog` so that it won't have to be short-lived and it will be very clear to us when supporting users which catalog a user is dealing with. 

> The 1.0 DataCatalog which will be incompatible with 0.19 DataCatalog in subtle ways

And for this point, this also won't be the case if we just don't rename `KedroDataCatalog`. 

In any case, regardless of whether we rename or not, I would like to do some user validation with the `KedroDataCatalog` before `1.0.0` is released so we already know how it will be received and we still have a chance to change things in case users are unhappy.

astrojuanlu (approved):

astrojuanlu:
For future reference, we proposed an alternative plan here https://github.com/kedro-org/kedro/issues/4721

Issue #3982, opened 2024-07-02
Add remote (cloud) options for --conf-source 


OlegPodlipalin:
## Description

Currently, we deploy Kedro projects as part of our SaaS solutions within Docker containers, managed by a Prefect server on AWS ECS. These containers are configured to handle all possible configurations for various tenants. However, as we transition to an on-prem solution, we face a challenge where exposing all tenant configurations within a single container poses security and relevance issues since the client will have access to configurations that are not pertinent to them.

We're looking for a way to modularize configuration management without exposing unnecessary data. While Kedro supports cloud storage in the data catalog via the "protocol" option with the `fsspec` library, it does not support a similar mechanism for the configuration source. Enhancing this functionality would significantly streamline the deployment process for our on-prem solutions.

## Context

This change is crucial for us as it would allow for more secure and relevant configurations per tenant, without the risk of exposing data not intended for them. Implementing this feature would also align configuration management with how data sources are currently handled, making our systems more robust and flexible. This feature could benefit other users who require more granular control over configuration management in environments where host file systems are not a viable option.

## Possible Implementation

A potential implementation could involve extending the existing functionality of Kedro's OmegaConfLoader to recognize and handle various storage protocols using the `_parse_filepath` function from [kedro/io/core.py](https://github.com/kedro-org/kedro/blob/main/kedro/io/core.py) lines 690-728:
```python
def _parse_filepath(filepath: str) -> dict[str, str]:
    """Split filepath on protocol and path. Based on `fsspec.utils.infer_storage_options`.

    Args:
        filepath: Either local absolute file path or URL (s3://bucket/file.csv)

    Returns:
        Parsed filepath.
    """
    if (
        re.match(r"^[a-zA-Z]:[\\/]", filepath)
        or re.match(r"^[a-zA-Z0-9]+://", filepath) is None
    ):
        return {"protocol": "file", "path": filepath}

    parsed_path = urlsplit(filepath)
    protocol = parsed_path.scheme or "file"

    if protocol in HTTP_PROTOCOLS:
        return {"protocol": protocol, "path": filepath}

    path = parsed_path.path
    if protocol == "file":
        windows_path = re.match(r"^/([a-zA-Z])[:|]([\\/].*)$", path)
        if windows_path:
            path = ":".join(windows_path.groups())

    options = {"protocol": protocol, "path": path}

    if parsed_path.netloc and protocol in CLOUD_PROTOCOLS:
[...]

PR #4543, opened 2025-03-04
Add cloud protocol support for `--conf-source`

SajidAlamQB:
## Description

Related to: https://github.com/kedro-org/kedro/issues/3982

This PR adds fsspec cloud support into the `OmegaConfigLoader`, and now lets `--conf-source` argument to load configurations from remote locations (e.g., s3://, abfs://, gs://) just like Kedro does for data catalogs.

## Development notes

- Extended `_initialise_filesystem_and_protocol()` to parse the filepath for protocol and path (e.g., "s3", "abfs", etc.) then make an fsspec filesystem using the protocol.
- Moved the `_parse_filepath` function from `kedro.io.core` and moved it to `kedro.utils` to make it usable between different components. (re-exported them into `kedro.io.core` for backward compatibility)
- Added constants (`HTTP_PROTOCOLS`, `CLOUD_PROTOCOLS`) to `utils.py` for reuse.
- Updated documentation guide on how to use `--conf-source` with remote protocols.
- Added unit tests to cover new changes.

I've tested this manually using an s3 bucket see the run below.

Gif:

conf source

## Checklist


astrojuanlu (commented):
The docs look good!

docs/source/configuration/configuration_basics.md
```diff
@@ -168,6 +169,42 @@ For both the `tar.gz` and `zip` file, the following structure is expected:
...
+kedro run --conf-source=s3://my-bucket/configs/
+
+# Azure Blob Storage
+kedro run --conf-source=abfs://container@account/configs/
+
+# Google Cloud Storage
+kedro run --conf-source=gs://my-bucket/configs/
+```
+
+Authentication for remote storage access follows the same mechanisms as those used by the data catalog.
```
ankatiyar, line 195: How does this work? It would probably have to be environment variables right or some how pre-configured in the terminal? because I won't be able to say for the conf-source use this credential from `credentials.yml`?

SajidAlamQB: Good point I was thinking of the more underlying code but yes we do this through environment variables.

Maybe we include something like this in the docs:

For example, when using S3:
```
# Set credentials in environment
export AWS_ACCESS_KEY_ID=your_access_key
export AWS_SECRET_ACCESS_KEY=your_secret_key
kedro run --conf-source=s3://my-bucket/configs/
```

ankatiyar: Yeah would be better to clarify what the users need to do to set up credentials (even if it's just linking to the respective cloud platforms docs)

docs/source/configuration/configuration_basics.md
```diff
@@ -168,6 +169,64 @@ For both the `tar.gz` and `zip` file, the following structure is expected:
...
+kedro run --conf-source=s3://my-bucket/configs/
+
+# Azure Blob Storage
+kedro run --conf-source=abfs://container@account/configs/
+
+# Google Cloud Storage
+kedro run --conf-source=gs://my-bucket/configs/
+```
+
+#### Authentication for remote configuration
```
ElenaKhaustova, line 195: Maybe we should explicitly say that setting the credentials from `credentials.yml` will not work to prevent potential issues?

kedro/config/omegaconf_config.py
```diff
@@ -358,20 +385,34 @@ def load_and_merge_dir_config(
...
         ):
-            protocol = "zip"
+            return fsspec.filesystem(protocol="zip", fo=conf_source), "zip"
+
+        # Parse to check for protocol
+        options = _parse_filepath(conf_source)
+        protocol = options["protocol"]
+
+        # Create and return the appropriate filesystem
+        if protocol in HTTP_PROTOCOLS:
```
ElenaKhaustova, line 407: Is this expected that the current `if` and the following `elif` return the same? If so, does it make sense to join them into one?

SajidAlamQB: Good catch, I initially thought there would be some extra implementation between both but it turned out a lot simpler, we can join them into one.

kedro/config/omegaconf_config.py
```diff
@@ -145,9 +148,17 @@ def __init__(  # noqa: PLR0913
         # Register globals resolver
         self._register_globals_resolver()
 
+        # Store original conf_source
+        self._original_conf_source = conf_source
```
ankatiyar, line 152: Why is this necessary? It's not being used anywhere else after this.

ankatiyar (commented):
I have another question, does it work when it's a `.tar.gz` or `zip` file on the cloud?

SajidAlamQB:
> I have another question, does it work when it's a `.tar.gz` or `zip` file on the cloud?

No, the current implementation doesn't support this.

kedro/config/omegaconf_config.py
SajidAlamQB: Good catch this was left behind.

ankatiyar (approved):
LGTM!
I would add a line in the docs under the section for "how to use configuration from a compressed file" that it doesn't work with the cloud options

ElenaKhaustova (approved):
LGTM 👍

PR #4598, opened 2025-03-20
Add missing space in `DataCatalog` deprecation warning

yury-fedotov:
Screenshot 2025-03-20 at 2 41 49 PM

> I see that other lines have this space on the right end of each string, so doing the same.


yury-fedotov:
Strange, how does it make docs build fail 👀

SajidAlamQB (approved):
Good catch, thank you @yury-fedotov!

merelcht (approved):

Issue #4584, opened 2025-03-18
[DataCatalog2.0]: Remove `DataCatalog`


ElenaKhaustova:
## Description
Partially addresses https://github.com/kedro-org/kedro/issues/4476

We suggest starting work on the `develop` branch to remove the old `DataCatalog` to have a clean slate for adding further updates.

Tasks required:

- [x] Removing `DataCatalog`: https://github.com/kedro-org/kedro/pull/4607
- [x] Resolving `KedroDataCatalog` TODOs, updating `CatalogProtocol`: https://github.com/kedro-org/kedro/pull/4609
- [x] Updating corresponding unit tests: https://github.com/kedro-org/kedro/pull/4646
- [x] Marking old catalog related doc pages "Subject to update"

We suggest splitting this work into several PRs to simplify the reviewing process.

**NOTE:** since we decided to keep the current versioning, ignore TODOs with removing `version argument`. Like 
```python
# TODO: remove version from input arguments when breaking change
```

datajoely:
BIG

PR #4609, opened 2025-03-25
Resolving KedroDataCatalog TODOs and updating CatalogProtocol

ravi-kumar-pilla:
## Description

Partially resolves https://github.com/kedro-org/kedro/issues/4584

## Development notes

* This handles all the TODOs after removing DataCatalog
* Updated CatalogProtocol to sync with KedroDataCatalog
* This PR should be good for manual testing

**NOTE:** The PR is based out off `chore/remove-dc` for easier reviews

## Checklist


kedro/io/kedro_data_catalog.py
```diff
@@ -484,71 +470,6 @@ def _add_from_config(self, ds_name: str, ds_config: dict[str, Any]) -> None:
 
         self.add(ds_name, ds)
```
ElenaKhaustova, line 471: We need to replace it with `__setitem__`

ElenaKhaustova (changes requested):
Changes on the catalog side looks good to me but we also need to double check that this version is runnable - meaning we replaced all the occurrences of removed catalog methods within the codebase.

For example: https://github.com/kedro-org/kedro/blob/6837c9c4c743113de19298c3ad21229b69f795a6/kedro/framework/context/context.py#L242

The easiest way to do this is just test the current version with some toy project, maybe spaceflights-pandas. Note: you will also need to uninstall `kedro-telemetry` cause it's importing the old catalog.

ravi-kumar-pilla:
Hi @ElenaKhaustova ,

I tested the starter project `spaceflights-pandas` with current branch and it runs fine (we should uninstall kedro-telemetry and kedro-viz as both use old catalog). 

I haven't tested with runtime_params and other complex scenarios (dataset factories etc). Since `add_feed_dict` is removed, I added some alternative in `kedro/framework/context/context.py`. Also, left some `TODO: KDC` comments to clarify with you. Please review when you have some time. I will test other scenarios if this looks good.

Thank you

kedro/framework/context/context.py
```diff
@@ -238,47 +240,61 @@ def _get_catalog(
...
         )
 
-        feed_dict = self._get_feed_dict()
-        catalog.add_feed_dict(feed_dict)
+        # TODO: [KDC] Need to confirm if it is better to have add_raw_data
+        # similar to add_feed_dict ?
+        raw_data = self._get_raw_data()
+
+        for ds_name, ds_data in raw_data.items():
+            dataset = (
```
ElenaKhaustova, line 248: `catalog.__setter__` does these checks itself

```suggestion
  for ds_name, ds_data in raw_data.items():
      catalog[ds_name] = dataset
```

kedro/framework/context/context.py
```diff
@@ -238,47 +240,61 @@ def _get_catalog(
...
+                ds_data
+                if isinstance(ds_data, AbstractDataset)
+                else MemoryDataset(data=ds_data)
+            )
+            catalog[ds_name] = dataset
+
         _validate_transcoded_datasets(catalog)
+
+        # TODO: [KDC] Removed feed_dict from the args. Should we add raw_data
```
ElenaKhaustova, line 257: Good point since we removing any `feed_dict` names at the catalog side.

kedro/framework/context/context.py
```diff
@@ -238,47 +240,61 @@ def _get_catalog(
             save_version=save_version,
         )
 
-        feed_dict = self._get_feed_dict()
-        catalog.add_feed_dict(feed_dict)
+        # TODO: [KDC] Need to confirm if it is better to have add_raw_data
```
ElenaKhaustova, line 243: Maybe we can rename these to `_get_parameters` and `add_parameters` since that's what we actually do?

kedro/framework/hooks/specs.py
```diff
@@ -20,12 +20,11 @@ class DataCatalogSpecs:
     """Namespace that defines all specifications for a data catalog's lifecycle hooks."""
 
     @hook_spec
-    def after_catalog_created(  # noqa: PLR0913
+    def after_catalog_created(
         self,
         catalog: CatalogProtocol,
         conf_catalog: dict[str, Any],
         conf_creds: dict[str, Any],
-        feed_dict: dict[str, Any],
```
ElenaKhaustova, line 28: Here we can also add a `parameters` argument instead if moving to the new naming.

ElenaKhaustova (commented):
Tested on my side - all works as expected! 🚀

Left a few comments on the implementation and naming.

kedro/framework/context/context.py
ravi-kumar-pilla: Yes, but we named it as raw_data in kedro_data_catalog - https://github.com/kedro-org/kedro/pull/4609/files#diff-0a74699b4eabc57e8a4d40f84eef8bee8cb38952367edfe407f2e9d8c476c9aeR62

I can rename it to parameters but feed_dict was renamed to raw_data in kedro_data_catalog, so I replicated the same here. Let me know if this is not clear (for me parameters is more clear. Even in kedro_data_catalog, raw_data is not so clear)

ElenaKhaustova: The thing is that this is raw data for the catalog — the catalog doesn't know about the parameters and anything can be passed as raw data. But we add parameters at the level of the context, so here it makes sense to specify name.

My suggestion is to get rid of feed_dict everywhere as it's definitely confusing, and rename where we work with parameters to specify that and keep `raw_data` at the catalog level.

ravi-kumar-pilla: Make sense. Thank you

kedro/framework/context/context.py
```diff
@@ -238,47 +238,65 @@ def _get_catalog(
...
         )
         return catalog
 
-    def _get_feed_dict(self) -> dict[str, Any]:
-        """Get parameters and return the feed dictionary."""
+    def _add_parameters(
+        self, catalog: CatalogProtocol, parameters: dict[str, Any]
+    ) -> None:
+        """Add datasets to catalog using the data provided through `parameters`.
+        `parameters` is a dictionary where the keys represent dataset names and the values can either be raw data or
```
ElenaKhaustova, line 260: I'm not sure this is a valid description of the parameters, as we're always adding a dictionary.

ravi-kumar-pilla: So parameters are dictionaries with param_name as the dataset_name (f"params:{param_name}") and value can be raw_data or AbstractDataset instances. This was the same for feed_dict. What could be a valid description ?

ElenaKhaustova: I think the value for parameters cannot be `AbstractDataset` as parameters are the dictionary we get from the config loader.

ElenaKhaustova: You're mixing definitions at the catalog and context levels. At the catalog value, raw data can be anything, including `AbstractDataset `. But here, we just add parameters loaded from `.yaml`—they are simple Python objects.

kedro/io/kedro_data_catalog.py
```diff
@@ -227,24 +204,27 @@ def __setitem__(self, key: str, value: Any) -> None:
...
-            self._load_versions, self._save_version = _validate_versions(
+            self._load_versions, self._save_version = self._validate_versions(
                 {key: value}, self._load_versions, self._save_version
             )
-            self.__datasets[key] = value
+            self._datasets[key] = value
         elif isinstance(value, _LazyDataset):
             self._lazy_datasets[key] = value
         else:
             self._logger.info(f"Adding input data as a MemoryDataset - {key}")
```
ElenaKhaustova, line 217: @ravi-kumar-pilla, qq - do you think this logging message is informative? I find it a bit annoying, so I'm thinking about removing it. Wdyt?

ElenaKhaustova (approved):
Thank you @ravi-kumar-pilla!

It's so nice to see clean catalog ☺️

kedro/framework/context/context.py
```diff
@@ -238,47 +238,62 @@ def _get_catalog(
...
-            feed_dict=feed_dict,
+            parameters=parameters,
             save_version=save_version,
             load_versions=load_versions,
         )
         return catalog
 
-    def _get_feed_dict(self) -> dict[str, Any]:
-        """Get parameters and return the feed dictionary."""
+    def _add_parameters(
```
ElenaKhaustova, line 256: Just have noticed that this should be a function but not a class method

ravi-kumar-pilla: Should we add this to some utils inside kedro/framework ?

kedro/io/kedro_data_catalog.py
ravi-kumar-pilla: I feel it is informative if the user missed adding to catalog by mistake. At the same time it can be lot of info in case the project is big with a lot of MemoryDataset. Shall we reduce the level of logging to debug may be ?

ElenaKhaustova: Yeah, let's put `debug` level then. Thank you!

kedro/io/kedro_data_catalog.py
```diff
@@ -482,72 +468,7 @@ def _add_from_config(self, ds_name: str, ds_config: dict[str, Any]) -> None:
             self._save_version,
         )
 
-        self.add(ds_name, ds)
```
ElenaKhaustova, line 485: I think this one was removed by mistake. It should be replaced with  `self.__setitem__(ds_name, ds)` but not removed

kedro/framework/context/context.py
```diff
@@ -238,47 +238,53 @@ def _get_catalog(
...
-            feed_dict=feed_dict,
+            parameters=parameters,
             save_version=save_version,
             load_versions=load_versions,
         )
         return catalog
 
-    def _get_feed_dict(self) -> dict[str, Any]:
-        """Get parameters and return the feed dictionary."""
+    def _get_parameters(self) -> dict[str, Any]:
```
merelcht, line 259: 😍 finally no more "feed_dict" 🥳

kedro/framework/context/context.py
```diff
@@ -238,47 +238,53 @@ def _get_catalog(
...
+        """Returns a dictionary with data to be added in memory as `MemoryDataset`` instances.
+        Keys represent parameter names and the values are parameter values."""
         params = self.params
-        feed_dict = {"parameters": params}
+        params_dict = {"parameters": params}
 
-        def _add_param_to_feed_dict(param_name: str, param_value: Any) -> None:
-            """This recursively adds parameter paths to the `feed_dict`,
+        def _add_param_to_params_dict(param_name: str, param_value: Any) -> None:
+            """This recursively adds parameter paths to the `params_dict`,
```
merelcht, line 266: Can we add what "params_dict" exactly contains? I think that will make it easier to understand what this method does.

kedro/framework/hooks/specs.py
```diff
@@ -37,7 +37,7 @@ def after_catalog_created(  # noqa: PLR0913
             catalog: The catalog that was created.
             conf_catalog: The config from which the catalog was created.
             conf_creds: The credentials conf from which the catalog was created.
-            feed_dict: The feed_dict that was added to the catalog after creation.
+            parameters: The datasets that are added to the catalog after creation.
```
merelcht, line 40: Are these parameters actual datasets or is this data coming from `parameters.yml`? I'm wondering if it's datasets why we wouldn't call the argument datasets then?

kedro/framework/context/context.py
ElenaKhaustova: It contains what came from `self.config_loader["parameters"]`

kedro/framework/hooks/specs.py
ElenaKhaustova: They are not actual datasets, but they can be accessed as datasets via catalog. The dictionary with nested parameters comes from `self.config_loader["parameters"]`, which is passed separately to this hook.

ElenaKhaustova (commented):
Found out one more place to replace old catalog API: `_validate_transcoded_datasets` in the `context.py`. Now it should be simplified just to:

```python
for dataset_name in catalog:
    _transcode_split(dataset_name)
```

ravi-kumar-pilla:
> Found out one more place to replace old catalog API: `_validate_transcoded_datasets` in the `context.py`. Now it should be simplified just to:
> 
> ```python
> for dataset_name in catalog:
>     _transcode_split(dataset_name)
> ```

My bad. Let me search for APIs that are removed one more time 👍 . Thank you @ElenaKhaustova

kedro/framework/hooks/specs.py
ravi-kumar-pilla: Should I change the description to `The parameters that are added to the catalog after creation.` ? if it is confusing. I left it as datasets in the `catalog` context.

PR #4629, opened 2025-04-01
Ensure dependencies come in a consistent order in grouped_nodes_by_namespace

lrcouto:
## Description

Changes the `pipeline.grouped_nodes_by_namespace` function to ensure dependencies are always inserted in the result object in the same order, preventing the flakiness that is causing issues in https://github.com/kedro-org/kedro-plugins/pull/1031.

Two changes were applied. The first was using the `self.grouped_nodes` property of the pipeline class to ensure the nodes are topologically ordered. Then dependencies are stored in a list instead of a set, to ensure insertion order is preserved. 

## Checklist

### DIFF
--- a/kedro/pipeline/pipeline.py
+++ b/kedro/pipeline/pipeline.py
@@ -371,15 +371,10 @@ class Pipeline:
 
     @property
     def grouped_nodes_by_namespace(self) -> dict[str, dict[str, Any]]:
-        """Return a dictionary of the pipeline nodes grouped by top-level namespace with
-        information about the nodes, their type, and dependencies. The structure of the dictionary is:
-        {'node_name/namespace_name' : {'name': 'node_name/namespace_name','type': 'namespace' or 'node','nodes': [list of nodes],'dependencies': [list of dependencies]}}
-        This property is intended to be used by deployment plugins to group nodes by namespace.
-        """
+        """Return a dictionary of the pipeline nodes grouped by top-level namespace."""
         grouped_nodes: dict[str, dict[str, Any]] = defaultdict(dict)
-        toposorted_nodes = [node for group in self.grouped_nodes for node in group]
 
-        for node in toposorted_nodes:
+        for node in self.nodes:
             key = node.namespace.split(".")[0] if node.namespace else node.name
 
             if key not in grouped_nodes:
@@ -392,15 +387,17 @@ class Pipeline:
 
             grouped_nodes[key]["nodes"].append(node)
 
-            dependencies = []
+            # Ensure dependencies are not duplicated on the returned list
+            dependencies = grouped_nodes[key]["dependencies"]
+            unique_dependencies = set(dependencies)
+
             for parent in self.node_dependencies[node]:
                 parent_key = (
                     parent.namespace.split(".")[0] if parent.namespace else parent.name
                 )
-                if parent_key != key and parent_key not in dependencies:
+                if parent_key != key and parent_key not in unique_dependencies:
                     dependencies.append(parent_key)
-
-            grouped_nodes[key]["dependencies"].extend(dependencies)
+                    unique_dependencies.add(parent_key)
 
         return grouped_nodes
 
--- a/kedro/pipeline/pipeline.py
+++ b/kedro/pipeline/pipeline.py
@@ -371,7 +371,11 @@ class Pipeline:
 
     @property
     def grouped_nodes_by_namespace(self) -> dict[str, dict[str, Any]]:
-        """Return a dictionary of the pipeline nodes grouped by top-level namespace."""
+        """Return a dictionary of the pipeline nodes grouped by top-level namespace with
+        information about the nodes, their type, and dependencies. The structure of the dictionary is:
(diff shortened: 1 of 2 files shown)


kedro/pipeline/pipeline.py
```diff
@@ -377,27 +377,31 @@ def grouped_nodes_by_namespace(self) -> dict[str, dict[str, Any]]:
         This property is intended to be used by deployment plugins to group nodes by namespace.
         """
         grouped_nodes: dict[str, dict[str, Any]] = defaultdict(dict)
-        for node in self.nodes:
-            if node.namespace:
-                key = node.namespace.split(".")[0]  # only take top level namespace
-            else:
-                key = node.name
+        toposorted_nodes = [node for group in self.grouped_nodes for node in group]
```
DimedS, line 380: I think that line is not needed, self.nodes already returns toposorted order, right?

ankatiyar:
I'm not sure this is necessary since `Pipeline.nodes` is already sorted topologically
https://github.com/kedro-org/kedro/blob/32c72fe35632ea2b8c0182e8f20464e3489edb13/kedro/pipeline/pipeline.py#L339-L351

kedro/pipeline/pipeline.py
```diff
@@ -377,27 +377,31 @@ def grouped_nodes_by_namespace(self) -> dict[str, dict[str, Any]]:
...
-                else:
-                    dependencies.add(parent.name)
-            grouped_nodes[key]["dependencies"].update(dependencies)
+                parent_key = (
+                    parent.namespace.split(".")[0] if parent.namespace else parent.name
+                )
+                if parent_key != key and parent_key not in dependencies:
+                    dependencies.append(parent_key)
+
+            grouped_nodes[key]["dependencies"].extend(dependencies)
```
DimedS, line 403: it will be possible duplication here?

DimedS (commented):
Thanks for the PR, @lrcouto! I agree that we should eliminate the use of set to achieve deterministic behaviour. I've added a few questions in the comments.

kedro/pipeline/pipeline.py
lrcouto: True, I'll change that.

kedro/pipeline/pipeline.py
lrcouto: Probably not, but I wanted to guarantee it since lists, unlike sets, allow duplicated entries. Do you think it's redundant?

DimedS: From my understanding, if a namespace contains two nodes (e.g., node2 and node3), each dependent on the same node1, the current implementation will duplicate the dependency list within that namespace. This duplication occurs because each node (node2 and node3) separately adds the shared dependency with:

`grouped_nodes[key]["dependencies"].extend(dependencies)`

We should therefore ensure dependencies are deduplicated to avoid redundancy.

lrcouto:
Went back to using self.node since it's already toposorted and used a set to track redundant dependencies.

kedro/pipeline/pipeline.py
```diff
@@ -371,33 +371,34 @@ def grouped_nodes(self) -> list[list[Node]]:
...
-                    if parent_namespace != key:
-                        dependencies.add(parent_namespace)
-                else:
-                    dependencies.add(parent.name)
-            grouped_nodes[key]["dependencies"].update(dependencies)
+                parent_key = (
+                    parent.namespace.split(".")[0] if parent.namespace else parent.name
+                )
+                if parent_key != key and parent_key not in unique_dependencies:
+                    dependencies.append(parent_key)
```
DimedS, line 399: thanks for addressing comments, @lrcouto ! I'm trying to understand how `dependencies` will be appended to `grouped_nodes[key]["dependencies"]` ?

lrcouto: `dependencies = grouped_nodes[key]["dependencies"]` creates a reference to `grouped_nodes[key]["dependencies"`. IIRC the way Python handles it under the hood is that this kind of assignment does not create a new object in memory, it kinda works like a C++ pointer.

For example, if I have:

```py
a = {"key": []}
b = a["key"]
b.append("value")

print(a["key"])
```

It outputs `["value"]`

lrcouto: Or you can do like:

```py
a = {"key": []}
b = a["key"]
b.append("value")

print(id(a["key"]))
print(id(b))
```

And it shows both of them are pointing to the same memory address.

DimedS (approved):
Thanks for the PR, @lrcouto ! LGTM

kedro/pipeline/pipeline.py
```diff
@@ -371,33 +371,34 @@ def grouped_nodes(self) -> list[list[Node]]:
 
     @property
     def grouped_nodes_by_namespace(self) -> dict[str, dict[str, Any]]:
-        """Return a dictionary of the pipeline nodes grouped by top-level namespace with
-        information about the nodes, their type, and dependencies. The structure of the dictionary is:
```
ankatiyar, line 375: Any reason to update the docstring? I had it shorter in the beginning but made it longer based on PR feedback so might be good to leave it as it is?

ankatiyar (approved):
LGTM thanks @lrcouto

PR #4630, opened 2025-04-01
feat/error-message-for-DataSet

Huongg:
## Description
Fixes #4580 

## Development notes
Include an extra check to  prevents the use of dataset types with uppercase 'S' in 'Dataset', which is no longer supported as of Kedro 0.19.

## Checklist

### DIFF
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -567,6 +567,15 @@ def parse_dataset_definition(
         )
 
     dataset_type = config.pop(TYPE_KEY)
+
+    # This check prevents the use of dataset types with uppercase 'S' in 'Dataset',
+    # which is no longer supported as of Kedro 0.19.
+    if isinstance(dataset_type, str):
+        if dataset_type.endswith("Set"):
+            raise ValueError(
+                f"Since Kedro 0.19, 'Dataset' is spelled with a lowercase 's'. Got '{dataset_type}'."
+            )
+        
     class_obj = None
     if isinstance(dataset_type, str):
         if len(dataset_type.strip(".")) != len(dataset_type):
--- a/tests/io/test_core.py
+++ b/tests/io/test_core.py
@@ -344,6 +344,13 @@ class TestCoreFunctions:
         with pytest.raises(DatasetError, match=pattern):
             parse_dataset_definition({"type": dataset_name})
 
+    def test_parse_dataset_definition_invalid_uppercase_s_in_dataset(self):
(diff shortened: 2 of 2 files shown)


kedro/io/core.py
```diff
@@ -567,6 +567,15 @@ def parse_dataset_definition(
         )
 
     dataset_type = config.pop(TYPE_KEY)
+
+    # This check prevents the use of dataset types with uppercase 'S' in 'Dataset',
+    # which is no longer supported as of Kedro 0.19.
+    if isinstance(dataset_type, str):
+        if dataset_type.endswith("Set"):
+            raise ValueError(
```
ankatiyar, line 575: Maybe this should be a warning since it could be a custom dataset too?

yury-fedotov:
Thank you very much for implementing this!

I agree with Ankita's suggestion of showing a warning instead of raising an error, as we probably want to allow custom datasets to end with `"Set"`.

astrojuanlu:
Does this conflict or overlap with https://github.com/kedro-org/kedro/pull/4580 ?

Huongg:
> Does this conflict or overlap with #4580 ?

hey no i don't think so, I've checked with Rashida and she has not implemented this in her PR.

kedro/io/core.py
Huongg: hey after changing the code, I realised that we use the warning then everything continues running, but eventually, another error will arise because the DataSet is not found. In that case, why not display the error at this stage instead?

ankatiyar: yeah so if it the dataset doesn't exist in `kedro-datasets` or `kedro.io` then the proper error will show but if it's a custom dataset with the "DataSet" spelling, it should continue running and not error out.

kedro/io/core.py
```diff
@@ -567,6 +567,15 @@ def parse_dataset_definition(
...
+                f"Since Kedro 0.19, 'Dataset' is spelled with a lowercase 's'. Got '{dataset_type}'."
```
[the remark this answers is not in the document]
Huongg: ah got you, I've updated it to be a warning rather error. And as we said, the behaviour will be to throw warning first and eventually it will be error, so i updated the test to reflect this too. Let me know you think

kedro/io/core.py
Huongg: @ankatiyar hey got you, I've updated it to be a warning now and also reflects it in the test, to ensure it throws warning first, then eventually it will be error. Let me know what you think?

ankatiyar (approved):

Issue #4039, opened 2024-07-29
`IncrementalDataset` cannot be used with namespaces


gtauzin:
## Description
I have a namespace-based incremental dataset and wish to use the confirms attribute to trigger CHECKPOINT update further down my pipeline. However, based on discussions on Slack, it seems that incremental datasets are not meant to be used within namespaces and so `confirms` is not "namespaced" by design.

Following discussion with @noklam on Slack, it seems that my use case could justify having "namespaced" `confirms`.

## Context
I have many devices that regularly record event files and push it to a S3 bucket. I would like to run a preprocessing pipeline that is different for each device and that would for each of them:
1. Load all new files as dataframes, preprocess them and concatenate the preprocessed recorded event and save the results to another S3 bucket
2. Load all preprocessed recorded files computed so far and concatenate them

Then , I use the concatenation of all recorded preprocessed event seen so far for data science purposes.

The way I achieve this with Kedro is:
- For step 1, I use `IncrementalDataset` and the concatenated dataframe is saved using a versioned ParquetDataset
- For step 2, I use a `PartionedDataset` that is able to find all preprocessed recorded event computer so far (with `load_args` `withdirs` and `max_depth` set accordingly)

Those steps are done for each device, so I use namespace to reuse the same logic for all of them varying the S3 bucket path. I need the confirms to be at step 2 because only then I can consider new files to have been processed.

## Workaround
@noklam suggested to try putting the namespace in the argument, e.g. confirms=namespace.data, as a workaround and I can confirm this worked.

gtauzin:
I believe this is also hidding a bug. If the incremental dataset is namespaced and the confirms argument is not explicitely set as per the workaround, no checkpoint file is created. I would guess that this is because if confirms is not provided, it is set to the incremental dataset name without the namespace and this dataset does not actually exist.

astrojuanlu:
Thanks @gtauzin and sorry for the slow response. We will investigate the issue you mention first.

gtauzin:
@ElenaKhaustova Thanks for looking into this!

DimedS:
Thanks for raising the issue, @gtauzin!

From my point of view, the problem was that `confirms` wasn't being renamed when creating a namespaced pipeline using the `pipeline()` function - unlike `inputs`, `outputs`, and `parameters`, which are correctly handled. I’ve addressed this in [PR #4634](https://github.com/kedro-org/kedro/pull/4634) - feel free to review!

[...]

PR #4634, opened 2025-04-02
Apply namespace to `confirms` in modular pipelines to support namespaced IncrementalDatasets (#4039)

DimedS:
## Description

This PR fixes [#4039](https://github.com/kedro-org/kedro/issues/4039), which highlighted that the `confirms` attribute on `Node` objects does not automatically receive namespace prefixing when used in modular pipelines. As a result, `IncrementalDataset` with a namespace would fail to trigger expected checkpoint behaviour, unless a manual workaround (`confirms=f"{namespace}.dataset"`) was applied.

### What changed
In modular pipeline construction, we now apply `_process_dataset_names(...)` to the `confirms` field the same way it's applied to `inputs` and `outputs`. This ensures confirms are properly namespaced when a pipeline namespace is present:

```python
confirms=_process_dataset_names(node._confirms),
```

### Why it matters
This update ensures that `IncrementalDataset` and other use cases relying on `confirms` work seamlessly with namespaced pipelines. It also removes the need for the manual workaround proposed in the issue and aligns `confirms` behaviour with that of `inputs` and `outputs`.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -3,6 +3,7 @@
 ## Major features and improvements
 ## Bug fixes and other changes
 * Fixed bug where project creation workflow would use the `main` branch version of `kedro-starters` instead of the respective release version.
+* Fixed namespacing for `confirms` during pipeline creation to support `IncrementalDataset`.
 ## Breaking changes to the API
 ## Documentation changes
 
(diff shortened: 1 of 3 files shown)


gtauzin (approved):
Thank you @DimedS! Looks very straightforward :)

ankatiyar (approved):
Thanks @DimedS 🚀

merelcht (approved):
LGTM! ⭐ Can you add this to the release notes?

Issue #2805, opened 2023-07-17
`pipeline.pipeline` confuses IDE autocompletion


noklam:
## Description

We import `pipeline` in our starter and this confuse IDE like VSCode, as it thinks `pipeline` is a module instead of the `pipeline` function.

`from kedro.pipeline import Pipeline, node, pipeline`

Changing it to the following avoid the issue.

`from kedro.pipeline.modular_pipeline import pipeline`
![image](https://github.com/kedro-org/kedro/assets/18221871/6430a53f-118e-430d-bf7c-cb3092c8ddad)

The problem here is IDE confuse `kedro.pipeline.modular_pipeline` as `kedro.pipeline.pipeline` module.

## Context

Solution 1:
- Move the `pipeline` definition (Or what's the difference between pipeline and modular pipeline? Will be a breaking change if we move modules)

Solution 2:
- Change starter to import from `modular_pipeline` explicitly, should we keep the shortcut in this case?

Solution 3:
- find out ways that won't confuse IDE (VS Code in my case, user report this with PyCharm too)

## Actual Result

```
-- If you received an error, place it here.
```

```
-- Separate them if you have more than one.
```

## Your Environment

* Kedro version used (`pip show kedro` or `kedro -V`):
* Python version used (`python -V`):
* Operating system and version:

astrojuanlu:
or

```
from kedro import Pipeline

Pipeline.make(...)
```

😄 #712

(see also #2723 about "modular pipelines" vs just "pipelines")

noklam:
Discussed in backlog grooming. Will keep this as a low priority for the 0.19 release and see if we have time to fit this in.

If we end up don't want to break anything in 0.19, we should still update the starters to avoid these syntax highlight problem by importing from `kedro.pipeline.modular_pipeline` directly.

astrojuanlu:
There's more confusion: https://github.com/kedro-org/kedro-viz/issues/1522#issue-1885805761

Some users are creating the pipelines with the `Pipeline` class.

I think we should tackle this and #2723 at one.

astrojuanlu:
I just saw this code snippet in someone else's code:

```python
my_pipeline = Pipeline([
    Node(filter_func, "df_input", "df_filtered"),
    Node(actual_func, "df_filtered", "df_output"),
])
```

is there a reason why we are using the `pipeline` and `node` helpers instead of the `Pipeline` and `Node` class initializers directly?

rashidakanchwala:
[...]

PR #4675, opened 2025-04-15
[TSC VOTE] Update recommendation to use classes instead of methods to instantiate `Node` and `Pipeline` objects

merelcht:
## Description

>[!IMPORTANT]
> This PR needs TSC votes 

This PR is meant to demonstrate the proposal of changing our recommendation to use the `node()` and `pipeline()` methods into using the `Node` and `Pipeline` classes instead to instantiate a node and pipeline respectively. 

See more context and discussion on this: https://github.com/kedro-org/kedro/discussions/4594#discussioncomment-12752426

## Development notes
Update the starter template to demonstrate what the change will be across starters and documentation.

## Checklist


DimedS (approved):
Thanks, @merelcht! I agree - we should remove functions entirely from our recommendations and only include classes.

astrojuanlu (approved):
Thanks @merelcht for this PR! I do support this change. The nice thing is that old code will continue working post 1.0, so users won't be forced to migrate 👍🏼

Galileo-Galilei (approved):

lrcouto (approved):

deepyaman (approved):
+1

marrrcin (approved):

ankatiyar (approved):

datajoely (approved):

idanov (dismissed):
My vote is no, as explained in my comment here: https://github.com/kedro-org/kedro/discussions/4594#discussioncomment-12932978

deepyaman:
> My vote is no, as explained in my comment here: [#4594 (comment)](https://github.com/kedro-org/kedro/discussions/4594#discussioncomment-12932978)

I'm going to comment on your response here to avoid splitting the context:

> I don't think this is a good idea for multiple reasons:
> 
> - all existing code uses lowercase
> - all existing learning resources use lowercase
> - the lowercase is visually nicer to the eye and easier to type
> - all LLM is trained on lowercase
> - given all of the above, change in style will only introduce confusion for new users and LLM suggestions too
> - experience with previous useless renamings (i.e. the DataSet to Dataset) shows that there's zero benefits and lots of frustration with users (mentioned in multiple of interviews that the renaming of DataSet to Dataset caused a lot of frustration with users)

As @astrojuanlu wrote, "The nice thing is that old code will continue working post 1.0, so users won't be forced to migrate." It's a proposal to change the recommended way to create the class (i.e. call the constructor instead of calling a function that simply calls the constructor).

In terms of benefits, I think the big things are:

1. The relationship between the call to construct an object and the object becomes more clear (you don't have to wonder if `pipeline` or `node` is doing something special in the function other than creating the class... which it's not; the `__repr__` looks more akin to what you write).
2. You don't need to import `pipeline` and `Pipeline`—one for the function call, and one for the type hint. This is an annoying thing in the template.

> - in Python, essential and staple constructors like `dict`, `list`, `tuple` are all lowercase, which the `node` and `pipeline` emulate by showing how essential and staple constructors they are for a Kedro project

I don't think this is mirrored by significant projects (e.g. scikit-learn) outside stdlib. Even in stdlib, it's a bit confusing; why is `defaultdict` lowercase, but `NamedTuple` isn't?

> - `node` can be very easily reworked into a decorator, thus keeping the naming case between declarative vs decorated 

Yes, this has been in the back of my head, too! I think it would be great to enable `@node` and `@pipeline`, perhaps in a syntax similar to Prefect's `@task` and `@flow` (as opposed to the ~weird~less intuitive Metaflow way...). However, I think it's _better_ to distinguish the case for declarative vs. decorated here, which is why I feel recommending direct class construction for declarative makes sense.

> Before we go with this suggestion, can we get a few valid reasons why this is necessary? In general it will be a good practice to always discuss breaking changes (or preludes to breaking changes), especially major ones like this one, to be reasoned from the principle: **It's a no, until you prove that yes is worthwhile**. At this point, after nearly 7 years of history using lowercase, I think it's a bit too late to vote entirely based on preference (as none of the benefits listed are real benefits or explained concretely enough, so the only reason I see for this change is personal preferences).

Hopefully some of the comments above help clarify (1) the benefits and (2) that this wouldn't be breaking.

> PS: My argument mainly focuses on `Node`, while I acknowledge that for `Pipeline` you could see much less history or support for suggesting lowercase `pipeline` instead of `Pipeline`. While my `no` in the vote stays for any suggestions for `node`, I think I could be convinced that `Pipeline` should be the recommended version.

Agree the benefit is also greater for `Pipeline`, especially because of that second import just for type annotation of `create_pipeline()`, but think there is additional benefit to the docs recommendation around `Node` and keeping consistent (and avoiding the extra function call).

astrojuanlu:
> The argument for `Pipeline` vs `pipeline` is explained here https://github.com/kedro-org/kedro/issues/2805 long history of IDEs getting confused, several users commenting there. Further thoughts outlined a bit above this thread https://github.com/kedro-org/kedro/discussions/4594#discussioncomment-12753592
>
> The argument for `Node` vs `node` is a consistency one. Given that code will actually continue working, at least for now, I don't think this is equivalent to the `DataSet` renaming. That experience does inform how should we approach this though, to avoid repeating the same mistakes.

idanov:
I think I am not too opposed to the `pipeline` -> `Pipeline` change. Should we move on with the change for `Pipeline` first then, as it seems that we are mostly in agreement there? I still don't think the `node` one is necessary and would rather drop it from this proposal.

SajidAlamQB (approved):

merelcht:
> I think I am not too opposed to the `pipeline` -> `Pipeline` change. Should we move on with the change for `Pipeline` first then, as it seems that we are mostly in agreement there? I still don't think the `node` one is necessary and would rather drop it from this proposal.

My worry with this though is that it might make things even more confusing. Why is `node` with lowercase and `Pipeline` with upper-case? 

Thinking about this some more, I feel like the benefits don't really outweigh the downsides. It's a small change in recommendation, but we've had the lower-case for years and so any materials learners will find (including from GenAI bots) is going to mention the lowercase notation. Kedro already has a steep learning curve and additional things like this, while small on its own, might add to people having difficultly understanding when to use what. Many users won't get why we suddenly changed this recommendation when functionally nothing changes. 

I can't cast an "official" vote, since this is my PR, but at this point I vote against.

rashidakanchwala:
Hey team,

I get the pros of switching from pipeline() and node() to class-based Pipeline and Node

But I’m a bit confused about the trade-offs. During the hackathon, we saw Copilot/Claude/Cursor  always suggest the wrong usage of DataSet instead of Dataset. I’m worried the same could happen if we change long-standing lowercase functions—especially with more people relying on Copilot.

Maybe worth a quick tech design session to weigh up pros and cons?

astrojuanlu:
Do MCP servers solve that?

rashidakanchwala:
I don’t think MCP server fully solves this—at least not yet. Right now, it only supports Claude and Cursor, and users have to manually add it as a tool. So it doesn’t help in the typical flow where someone just opens VSCode or uses Copilot without extra setup or asks ChatGPT/Claude for help via chat interface.

Funny thing—when I tried generating a Kedro project using Cursor, it actually got everything right except that one thing: it used DataSet instead of Dataset. So yeah, even if switching to Pipeline and Node is the right direction long-term, I worry that AI-generated code will keep defaulting to the older patterns for a while—and users will have to keep correcting them.

That kind of friction might frustrate people and make Kedro feel harder to use.

 Maybe we wait a bit and revisit once tools like MCP are more widely adopted?

i know @DmitrySorokinQB also tried this.

astrojuanlu:
Let me summarize (hoping that we can reach a decision asynchronously). I think there's three separate things here.

(1) @idanov does not agree that the current status quo is bad:

> - the lowercase is visually nicer to the eye and easier to type
> - in Python, essential and staple constructors like dict, list, tuple are all lowercase, which the node and pipeline emulate by showing how essential and staple constructors they are for a Kedro project

It's been reported by users that having `Pipeline`, `pipeline`, `Node`, and `node` is confusing for both humans and IDEs https://github.com/kedro-org/kedro/issues/2805 I hope we are all mindful of the user perspective here.

(2) @idanov and @rashidakanchwala are worried about this situation being like the `DataSet` renaming:

> - experience with previous useless renamings (i.e. the DataSet to Dataset) shows that there's zero benefits and lots of frustration with users (mentioned in multiple of interviews that the renaming of DataSet to Dataset caused a lot of frustration with users) (@idanov)
> ---
> But I’m a bit confused about the trade-offs. During the hackathon, we saw Copilot/Claude/Cursor always suggest the wrong usage of DataSet instead of Dataset. I’m worried the same could happen if we change long-standing lowercase functions—especially with more people relying on Copilot. (@rashidakanchwala)

But @deepyaman original proposal, which @merelcht implemented here and requested a vote on, was to change the template _without_ breaking the original functions. In other words: old code will continue working. As such, I think the situations are not equivalent, and I would like us to apply the nuance that this topic deserves.

(3) @idanov, @rashidakanchwala, @merelcht are worried that old educational materials will instantly become outdated, which will add some confusion:

> - all existing learning resources use lowercase
> - all LLM is trained on lowercase
> - given all of the above, change in style will only introduce confusion for new users and LLM suggestions too

I think this is a valid concern and one that should be weighted against the cons.

My opinion is reflected by my vote, and it hasn't changed. I would like to insist that it's okay at this stage for TSC members that have already voted to reconsider their decision in case this information is new to them. Finally, If the TSC at large [votes](https://docs.kedro.org/en/latest/contribution/technical_steering_committee.html#voting-process) to reject the proposal, I will of course accept the decision. But I think we should follow our own processes, for consistency and transparency.

---

I'd like to add a reflection on the broader meaning of this seemingly small change: under what circumstances do we allow ourselves to evolve and change the Kedro APIs? Kedro 1.0 seemed like a good opportunity to address some (not all) of the design and usability debt we have, but I hear that people are nevertheless worried about backwards compatibility. Every change is potentially disruptive for someone's [workflow](https://xkcd.com/1172/), and we are implementing a bunch of those already: `DataCatalog` is being removed , we're evolving `KedroSession`, etc. I think we are missing a consistent approach for evolving the project in a way that allows us to adapt to changing requirements while avoiding excessive disruption.

ravi-kumar-pilla:
Hi Team,

I like the idea of using class based initializations (Pipeline() and Node()). Since this would be a breaking change if we delete the functions (which does the same work), it is good we think of backward compatibility and make it easy for users to upgrade. 

At the same time, I do have an opinion that v1.0.0 should be expected to bring long term changes (some breaking) focussed on stability. So I am happy even if we drop the functions all-together and recommend a stable way of creating objects.

Thank you

noklam (approved):
I agree with the change with pipeline -> Pipeline, with some comments on the implementation details.

Looking at the implementation, can we further simplify things? If this is signalling the Uppercase is the preferred choice which lowercase as a backward compatible way. Can we just use an alias?

```
# alias
node = Node
pipeline = Pipeline
```


I think if we keep it as alias, it's not too important if we pick the upper case or lower case and it would work with LLM nicely.

kedro/framework/cli/utils.py
```diff
@@ -19,6 +19,9 @@
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Sequence
+
+    from click import BaseCommand, MultiCommand
```
merelcht, line 23:
```suggestion
```

kedro/framework/cli/utils.py
```diff
@@ -19,6 +19,8 @@
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Sequence
+
+
```
merelcht, line 23:
```suggestion
```

PR #4560, opened 2025-03-11
added enhancing_kedro_documentation.md to integrations

Stephanieewelu:
Description
Add enhancing_kedro_documentation.md to integrations
Notes
Incorporating Ibis for SQL query execution into Kedro's workflow provides an efficient and scalable method for working with databases. With this contribution, users will gain an accessible and Pythonic way to handle SQL queries directly within their Kedro pipelines.

PR #4676, opened 2025-04-15
Feature/ibis integration docs

Stephanieewelu:
This contribution seeks to expand the Kedro documentation by integrating tutorials and working examples on how to run SQL queries with Ibis within Kedro projects. Ibis provides a high-level, Pythonic interface for SQL queries, allowing users to work with databases more intuitively while maintaining modular and scalable pipelines. By incorporating Ibis, Kedro users can efficiently execute SQL queries as part of their data workflows without needing to resort to hardcoded SQL or rely on less flexible methods.


docs/source/integrations/ibis.md
```diff
@@ -0,0 +1,407 @@
...
+# Install the appropriate backend connector
+# For PostgreSQL
+pip install 'ibis-framework[postgres]'
+# For MySQL
+pip install 'ibis-framework[mysql]'
+# For SQLite
+pip install 'ibis-framework[sqlite]'
+# For BigQuery
+pip install 'ibis-framework[bigquery]'
+```
```
astrojuanlu, line 33: I would rather be more prescriptive here and build all the example with DuckDB, which can be pip-installed and it's very easy to use (similar to what @deepyaman described in https://kedro.org/blog/building-scalable-data-pipelines-with-kedro-and-ibis)

docs/source/integrations/ibis.md
```diff
@@ -0,0 +1,407 @@
...
+```yaml
+# conf/base/db.yml
+database:
+  type: postgres  # or mysql, sqlite, bigquery, etc.
+  host: localhost
+  port: 5432
+  database: mydatabase
+  username: myuser
+  password: ${DB_PASSWORD}  # Use credentials from environment variables
+```
```
astrojuanlu, line 52: I think generally in Kedro we would put this under `conf/local/credentials.yml`

docs/source/integrations/ibis.md
```diff
@@ -0,0 +1,407 @@
...
+
+```python
+from <package_name>.hooks import IbisHooks
+
+hooks = [IbisHooks()]
+```
+
+## Creating an Ibis dataset for Kedro's DataCatalog
+
+To integrate Ibis with Kedro's DataCatalog, you can create a custom dataset that handles Ibis tables and query results. Here's an example implementation:
```
astrojuanlu, line 110: I think there's no need to create a custom dataset, one already exists! https://docs.kedro.org/projects/kedro-datasets/en/kedro-datasets-6.0.0/api/kedro_datasets.ibis.TableDataset.html

docs/source/integrations/ibis.md
```diff
@@ -0,0 +1,407 @@
...
+
+
+class IbisHooks:
+    @hook_impl
+    def after_context_created(self, context) -> None:
+        """Initialize an Ibis connection using the config
+        defined in project's conf folder.
+        """
+        # Load the database configuration
+        db_config = context.config_loader["database"]
```
astrojuanlu, line 70: If https://docs.kedro.org/projects/kedro-datasets/en/kedro-datasets-6.0.0/api/kedro_datasets.ibis.TableDataset.html is used, I think this hook is not needed?

docs/source/integrations/ibis.md
```diff
@@ -0,0 +1,407 @@
...
+    
+    # Select relevant columns
+    result = joined.select([
+        customers.id,
+        customers.name,
+        orders.order_date,
+        orders.amount
+    ])
+    
+    return result
```
astrojuanlu, line 244: These look mostly good to me 👍🏼 Could you maybe pass `ruff` over them? So that double quotes are used instead of single quotes etc

astrojuanlu (changes requested):
Hi @Stephanieewelu , good to see your second attempt! This is looking promising 💪🏼 

My recommendation is to remove `docs/source/integrations/enhancing_kedro_documentation.md`, since it looks like `docs/source/integrations/ibis.md` already contains a good chunk of useful content. That will address the RTD problem:

```
 sphinx.errors.SphinxWarning: /home/docs/checkouts/readthedocs.org/user_builds/kedro/checkouts/4676/docs/source/integrations/enhancing_kedro_documentation.md:document isn't included in any toctree 
```

(https://app.readthedocs.org/projects/kedro/builds/27863930/)

I made some high level comments after a first pass.

Stephanieewelu:
Thank you so much for your valuable support @astrojuanlu. I would utilize your guidance now and make necessary adjustments.

astrojuanlu:
Continues in #4560

Issue #4566, opened 2025-03-12
Can we add fallback behaviour for automatically pre-fixed namespace datasets?


ankatiyar:
## Description
Following discussion #4506 

Currently a namespaced pipeline also expects the inputs, outputs and parameters to be namespaced. Users can either:
- Add namespaced datasets/parameters to configuration
- Specify mappings of inputs and outputs in the `pipeline()` function

The goal for this ticket is to explore if this behaviour can be updated so:
- The input, output, parameter mapping can be automatic (eg. use the non-namespaced entry if the namespaced dataset is not available)
- This behaviour is clear and unambiguous 

## Context
https://github.com/kedro-org/kedro/issues/4506#issuecomment-2718681148

ankatiyar:
I've opened a PR to demo the solution for this #4700 
Would like to get opinions on the solution.

Proposal:
Add a `rename` (or `rename_datasets`) flag to the `pipeline()` method. 
When namespacing their pipelines, currently users have to either:
- namespace datasets in catalog 
- specify mapping for `inputs` `outputs` and `parameters` 

This will allow users to disable the renaming of the datasets when they are using the namespacing feature for deployment purposes.
The value of adding an argument over automatic renaming is so that this setting is deliberate and doesn't confuse the users. The automatic disabling would also add overhead for checking if the datasets are available in the catalog and will couple `Pipeline` and the `DataCatalog` class.

ankatiyar:
Closed by #4700

PR #4700, opened 2025-05-06
Let users disable datasets renaming for namespaced pipelines

ankatiyar:
## Description

Fix https://github.com/kedro-org/kedro/issues/4566 This PR is a prototype to demo the solution

## Development notes

- Add a `prefix_namespace` argument to `pipeline()` which defaults to `True` but can be set to `False` instead of automatic renaming of the datasets 

What it would look like for users:
```python
from kedro.pipeline import Pipeline, node, pipeline
from kedro_example.pipelines.data_science.nodes import train_model
from .nodes import create_model_input_table, preprocess_companies, preprocess_shuttles


    return pipeline(
        [
            node(
                func=preprocess_companies,
                inputs="companies",
                name="preprocess_companies_node",
            ),
            node(
                func=preprocess_shuttles,
                inputs="shuttles",
                outputs="preprocessed_shuttles",
                name="preprocess_shuttles_node",
            ),
            node(
                func=create_model_input_table,
                inputs=["preprocessed_shuttles", "preprocessed_companies", "reviews"],
                outputs="model_input_table",
                name="create_model_input_table_node",
            ),            
        ],
        namespace="data_processing",
        prefix_namespace=False,
    )
```

TODOs if we decide to go for this solution:

- [x]  docstrings
- [ ] Update docs 
- [x] Add unit tests
- [x] Release notes
- [x] target branch `main` or `develop` 

## Checklist


kedro/pipeline/pipeline.py
```diff
@@ -146,6 +146,7 @@ def __init__(  # noqa: PLR0913
         parameters: str | set[str] | dict[str, str] | None = None,
         tags: str | Iterable[str] | None = None,
         namespace: str | None = None,
+        rename: bool = True,
```
ElenaKhaustova, line 149: Just adding a comment not to forget to update docstrings with some clarifications regarding this parameter.

ElenaKhaustova (commented):
The suggested solution looks good to me!

I think one important thing not mentioned here is what will be the API for users to enable/disable this behaviour.

kedro/pipeline/pipeline.py
ankatiyar: i've updated the PR description with TODOs and the user facing API for more clarity for reviewers

ElenaKhaustova (approved):
Thank you for adding extra explanation!

For me `rename` flag seems a big confusing. Maybe we can consider some other names like `namespace_entries`/ `namespace_datasets`. Not a blocker as solution itself looks good.

DimedS (commented):
Thanks for the PR, @ankatiyar!

I fully support the solution - adding an option to avoid prefixing of **inputs, outputs, parameters and confirms** when creating a new pipeline with a namespace makes perfect sense.

A few concerns to consider in the current PR:

1. **Naming of the option ("rename")**  
   The current name does not seem clear enough. Imagine seeing `Pipeline(..., rename=False)` — it is not obvious what is not being renamed. Is it the pipeline itself, the nodes, the datasets?  
   Given that this is a niche option, the name should be more descriptive and self-explanatory. 

2. **Placement of the logic in `_rename()`**  
   Currently, the logic to conditionally skip renaming is embedded in the `_rename()` function. However, based on the name, this function should only be responsible for renaming. It is confusing to pass an option to a renaming function that effectively tells it *not* to rename.  
   A cleaner approach would be to make this decision at a higher level, such as in `_process_dataset_names()`. That function can decide:  
   - If renaming is needed → call `_rename()`  
   - If renaming is not needed → skip `_rename()` entirely.

merelcht:
I like the idea of adding this argument, but similar to what @ElenaKhaustova and @DimedS have mentioned, it's not clear to me what "rename" actually means. If `rename` is False, I would assume no renaming is needed, but it's the other way around 😅 Namespaces is already a difficult feature to understand, so I think we need to be more verbose with the name for this argument to make it clear what the purpose is.

> Does this go in main or develop?

While the way this is implemented is a non-breaking change, I'd suggest adding it to `1.0.0` to beef up the new feature set 🙂

astrojuanlu:
Instead of `rename`, what about the ability of setting a custom prefix? And then users could do `override_prefix=""` or similar ❓

ankatiyar:
I changed the name to `prefix_namespace` but I'm also not sure if it is unambiguous enough. Can I get some votes/ideas?
Some ideas/options:
- `rename`
- `rename_datasets`
- `prefix_namespace`
- `auto_prefix_namespace`
- `auto_prefix`
- `override_prefix`

DimedS:
> I changed the name to `prefix_namespace` but I'm also not sure if it is unambiguous enough. Can I get some votes/ideas? Some ideas/options:
> 
> * `rename`
> * `rename_datasets`
> * `prefix_namespace`
> * `auto_prefix_namespace`
> * `auto_prefix`
> * `override_prefix`
> 
> @merelcht @astrojuanlu @ElenaKhaustova @lrcouto @DimedS

thanks @ankatiyar , I like `prefix_datasets_with_namespace`, for me `prefix_namespace` sounds like we plan to prefix namespace itself

merelcht:
Initially I thought `prefix_namespace` is a logical name, but exploring this a bit more it seems like Dmitry's suggestion `prefix_datasets_with_namespace` is more accurate. 

After testing, the namespace is still applied to the node name itself, so when running viz the suggestions to run a node show something like `kedro run --to-nodes='data_science.evaluate_model([regressor;X_test;y_test]) -> None'`

And running `kedro registry describe` will show something like:
```
Nodes:
- data_science.preprocess_data (preprocess_data)
- preprocess_data (preprocess_data)
- preprocess_shuttles (preprocess_shuttles)
- create_model_input_table (create_model_input_table)
- data_science.split_data (split_data)
- data_science.train_model (train_model)
- data_science.evaluate_model (evaluate_model)
```

ankatiyar:
@merelcht The idea for this feature is to not have the users rename entries in the catalog with prefixes or have to define a mapping dictionary for the inputs/outputs. I think prefixing the name of the node is still valuable because we're still namespacing the pipeline and it should show on viz that way

ankatiyar:
@merelcht oh yeah, I agree! Will update the PR with the suggestion!

DimedS (approved):
great work, @ankatiyar !

ankatiyar:
PS will do docs in a separate PR it's too spread across

merelcht (approved):
Thanks @ankatiyar this looks great 👍 

Makes sense to do the docs separately. That way you can also make sure they are done with the new formatting etc. (cc: @Huongg )

Issue #4684, opened 2025-04-22
Consider using structured objects to return the output of node grouping API


ankatiyar:
## Description
Based on user feedback #4382 about the node grouping API added in #4427 with plugin developers
Consider replacing the output of `Pipeline.grouped_nodes_by_namespace` to a structured object like a [data class](https://docs.python.org/3/library/dataclasses.html) instead of a `dict` 

## Context
From the interview:
> However, I suggest replacing the default dictionary with a structured object (e.g. data class). This would improve type safety, reduce boilerplate code, and make it easier to enforce proper typing among developers. With the current dictionary-based structure, I have to handle None cases explicitly, which adds unnecessary complexity.

## Possible Implementation
- http://docs.python.org/3/library/dataclasses.html

DimedS:
I like that - it makes a lot of sense. We should also remember to update `kedro-airflow` accordingly.

DimedS:
closed with #4729

Issue #4685, opened 2025-04-22
Consider making the node grouping API more extensible


ankatiyar:
## Description
Based on user feedback https://github.com/kedro-org/kedro/issues/4382 about the node grouping API added in https://github.com/kedro-org/kedro/pull/4427 with plugin developers

The API `Pipeline.grouped_nodes_by_namespace` is a property which returns the nodes grouped only by namespace for plugins to use for deployment. Making this API re-usable so that it returns the same structured output whether the nodes are grouped or not will reduce the development effort on the plugin side.

## Context
> My only suggestion is to provide a helper method or unified API that returns the same structured output, whether nodes are grouped by namespace or not. This would keep the API consistent and save plugin developers from having to duplicate logic to convert plain nodes into the grouped structure. Otherwise, every plugin will end up copy-pasting the same code, which could easily be handled by the framework itself.

> That’s why I asked if the dictionary has the same structure as .nodes—if not, every plugin would need to duplicate code, increasing maintenance.
In the plugins I’ve worked on, we typically map deployment units as pure nodes and run them using kedro run with from_nodes and to_nodes. If the structure includes the nodes to execute, I can use it the same way. In the future, it might shift to generating --namespace runs, but for now, this simple pattern works well.

## Possible Implementation

- Make the API more general - it returns "deployment object" 
- Consider making this a method which accepts the grouping type: by namespace, no grouping, by memory datasets, by tags etc so it's extensible in the future

## Possible Alternatives

- Keep it as it is?

DimedS:
Thanks @ankatiyar, I completely agree - this will significantly simplify and clean up the code across many deployment plugins, and make it easier to add grouping functionality where it's not yet implemented.

datajoely:
I would love to group by the following attributes

- Tags
- Namespaces
- Regex on name
- Registered pipeline (appreciate this is more difficult, but we have evidence that this is what people do today along with tags)
- Catalog metadata (which would encapsulate Kedro viz layer)

PR #4708, opened 2025-05-08
Refactor grouped_nodes_by_namespace() to return structured GroupedNode list

DimedS:
## Description
First part of #4684 - modified `Pipeline.grouped_nodes_by_namespace()` to return a list of `GroupedNode` objects instead of a dictionary.

This improves the structure and type safety of the API, and makes it easier for deployment plugins and other consumers to work with node grouping results in a uniform way.

There was flakiness in the tests because
```
@property  
def node_dependencies(self) -> dict[Node, set[Node]]:  
```
is not deterministic due to the use of set.
To address this, I implemented additional sorting in `grouped_nodes_by_namespace().`

## Checklist


lrcouto:
I like the introduction of the `GroupedNode` class, I think it could be a good base for future extensions to the API as well, since it'd return a predictable format to plugins.

I have a question about

`for parent in sorted(self.node_dependencies[node], key=lambda n: n.name):`

Couldn't sorting by name introduce other issues when alphabetical order and execution order don't match?

kedro/pipeline/pipeline.py
```diff
@@ -542,40 +551,38 @@ def grouped_nodes(self) -> list[list[Node]]:
...
+            grouped_nodes_map[key].nodes.append(node.name)
 
-            # Ensure dependencies are not duplicated on the returned list
-            dependencies = grouped_nodes[key]["dependencies"]
+            # Ensure dependencies are not duplicated
+            dependencies = grouped_nodes_map[key].dependencies
             unique_dependencies = set(dependencies)
 
-            for parent in self.node_dependencies[node]:
+            for parent in sorted(self.node_dependencies[node], key=lambda n: n.name):
```
ankatiyar, line 577: Wondering if we can address the determinism in the `node_dependencies` property itself? And if sorting by name is the right choice?

DimedS: I’ve also thought about this, and it’s definitely possible. However, the function currently returns a `set`, which is non-deterministic. To support this change, we’d need to alter the return type - and I’m not sure how much that would break, especially across all the plugins and integrations. The impact could be significant, I don't think that's a good way to proceed with that.

DimedS:
> I have a question about
> 
> `for parent in sorted(self.node_dependencies[node], key=lambda n: n.name):`
> 
> Couldn't sorting by name introduce other issues when alphabetical order and execution order don't match?

In my opinion, if a node depends on other nodes that must be completed before it can run, the exact order in which those dependencies are executed doesn't matter - as long as all are finished beforehand. That's likely why a set was chosen to store dependencies for each node. In this case, we're introducing determinism purely to avoid issues in tests. The original execution order of the dependencies isn't important, since we can't predict which node will finish first - they are all equal in terms of dependency requirements.

lrcouto:
> > I have a question about
> > `for parent in sorted(self.node_dependencies[node], key=lambda n: n.name):`
> > Couldn't sorting by name introduce other issues when alphabetical order and execution order don't match?
> 
> In my opinion, if a node depends on other nodes that must be completed before it can run, the exact order in which those dependencies are executed doesn't matter - as long as all are finished beforehand. That's likely why a set was chosen to store dependencies for each node. In this case, we're introducing determinism purely to avoid issues in tests. The original execution order of the dependencies isn't important, since we can't predict which node will finish first - they are all equal in terms of dependency requirements.

That makes sense. In this case, I think the PR is good to go!

lrcouto (approved):

kedro/pipeline/pipeline.py
```diff
@@ -131,6 +132,14 @@ class ConfirmNotUniqueError(Exception):
     pass
 
 
+@dataclass
+class GroupedNode:
```
merelcht, line 136: Can you add some docstring here to explain what the class is for? I'm also wondering if it makes more sense for this to go in `node.py`.. and should it be `GroupedNode` or `GroupedNodes` plural?

kedro/pipeline/pipeline.py
```diff
@@ -542,40 +551,38 @@ def grouped_nodes(self) -> list[list[Node]]:
         return [list(group) for group in self._toposorted_groups]
 
     @property
-    def grouped_nodes_by_namespace(self) -> dict[str, dict[str, Any]]:
-        """Return a dictionary of the pipeline nodes grouped by top-level namespace with
-        information about the nodes, their type, and dependencies. The structure of the dictionary is:
-        {'node_name/namespace_name' : {'name': 'node_name/namespace_name','type': 'namespace' or 'node','nodes': [list of nodes],'dependencies': [list of dependencies]}}
-        This property is intended to be used by deployment plugins to group nodes by namespace.
+    def grouped_nodes_by_namespace(self) -> list[GroupedNode]:
```
merelcht, line 554: It might be confusing that the `grouped_nodes` property above doesn't return a `GroupedNode`. If this class is specifically for namespace grouping, maybe that should be reflected in the name.

merelcht (commented):
I like the concept of using a dataclass here, but I'm not entirely sure this is the right name for it. I've left some thoughts on the PR.

kedro/pipeline/pipeline.py
DimedS: In the next issue (https://github.com/kedro-org/kedro/issues/4685) we plan to make that API more generic, so name of the function will be changed - like grouped_nodes(with option --group_by)

kedro/pipeline/pipeline.py
DimedS: I agree. I've renamed it to GroupedNodes, moved it to node.py, and added a docstring.

DimedS:
Closing in favour of #4729, which correctly targets the develop branch for the upcoming major release. All previous commits  have been carried over.

Issue #3276, opened 2023-11-06
Rich add newline to long log message and it cannot be copy and paste directly into terminal


noklam:
## Description

```
    text = ",".join([f"node_{i}" for i in range(100)])
    logger.info(f"A very long log message - {text}")
```

I started with a `pandas-iris` starter and add this into one of the node. Copying and paste it into terminal produce this result. This  
conflicts with some feature such as resume pipeline (kedro will produce a log suggestion how to recover from a fail pipeline). This is common with medium size project.

![image](https://github.com/kedro-org/kedro/assets/18221871/a1e2b39c-f590-45fc-ab05-c76877da31e2)

## Actual Result

```
-- If you received an error, place it here.
```

```
-- Separate them if you have more than one.
```

## Your Environment

* Kedro version used (`pip show kedro` or `kedro -V`):
* Python version used (`python -V`):
* Operating system and version:

datajoely:
for this sort of error we could use [pyperclip](https://pypi.org/project/pyperclip/) which pandas includes to make `pd.DataFrame.to_clipboard()` possible...

noklam:
Nice! Would `rich` override this? It would be a great solution if works.

astrojuanlu:
I'm not sure it's possible to keep the rich logging while solving this issue. Probably we should push for making rich optional #2928. Voting to close this issue.

david-stanley-94:
Relevant code from `logging.py` for restarting a failed pipeline is:
```
            self._logger.warning(
                "There are %d nodes that have not run.\n"
                "You can resume the pipeline run from the nearest nodes with "
                "persisted inputs by adding the following "
                "argument to your previous command:\n%s",
                len(remaining_nodes),
                postfix,
            )
```
Can this not be altered to have a copy-pasteable form of the `postfix`, which contains the CLI element, printed to terminal using the `out` method from `rich` (https://rich.readthedocs.io/en/stable/console.html#low-level-output)? Either (a) as duplicate to follow the current logging call, or (b) as a replacement for removing the postfix from the current logging call?

astrojuanlu:
> I'm not sure it's possible to keep the rich logging while solving this issue. Probably we should push for making rich optional https://github.com/kedro-org/kedro/issues/2928. Voting to close this issue.

rich cannot be made fully optional, we have a transitive dependency with it.

david-stanley-94:
[...]

PR #4709, opened 2025-05-08
Make suggested resume commands copy-pasteable even when they extend over multiple lines

david-stanley-94:
## Description

- Adds a plain text log handler for log outputs that users benefit from being copy pasteable, such as suggested resume commands after pipeline failure
- Ensures such outputs are copy pasteable even for medium and large pipelines whose resume commands would span multiple lines
- The existing rich handler formats these with line breaks for text wrapping, padding spaces, etc, which means commands are not directly copy pasteable

## Development notes

- Add a plain text log handler in `logging.py`
- Configure the plain text handler in `framework/project/rich_logging.yml` (where the rich handler is configured)
- Add the plain text handler as a property to the kedro runner in `runner/runner.py` (where the rich handler is used)
- Use the plain text handler for outputting the `postfix` variable in the `_suggest_resume_scenario` method, in `runner/runner.py`. The text preceding the suggested command is still to use the rich handler.

On testing, gives the following output for arbitrary error raised in a node:

image

## Checklist

### DIFF
--- a/kedro/framework/project/rich_logging.yml
+++ b/kedro/framework/project/rich_logging.yml
@@ -9,10 +9,18 @@ handlers:
     # Advance options for customisation.
     # See https://docs.kedro.org/en/stable/logging/logging.html#project-side-logging-configuration
     # tracebacks_show_locals: False
+  
+  plain:
+    class: kedro.logging.PlainTextHandler
 
 loggers:
   kedro:
     level: INFO
+  
+  kedro.runner.plain:
+    level: INFO
+    handlers: [plain]
+    propagate: False
 
 root:
   handlers: [rich]
--- a/kedro/logging.py
+++ b/kedro/logging.py
@@ -13,6 +13,32 @@ import rich.traceback
 from kedro.utils import _is_databricks
 
 
+class PlainTextHandler(logging.StreamHandler):
+    """A simple handler that outputs log messages with minimal formatting.
+
+    This handler is designed for cases where plain text output without any fancy 
+    formatting, level indicators, timestamps or line wrapping is needed, such as
+    when providing shell command suggestions to users.
+    """
+
+    def __init__(self, stream=None):
+        super().__init__(stream)
+        # Create a formatter that only outputs the message without any additional info
+        self.formatter = logging.Formatter("%(message)s")
+
+    def emit(self, record: LogRecord) -> None:
+        # Skip level indicators entirely and just output the message
+        if record.levelno >= self.level:
+            try:
+                msg = self.formatter.format(record)
+                stream = self.stream
+                stream.write(msg)
+                stream.write(self.terminator)
+                self.flush()
+            except Exception:
+                self.handleError(record)
+
+
 class RichHandler(rich.logging.RichHandler):
     """Identical to rich's logging handler but with a few extra behaviours:
     * warnings issued by the `warnings` module are redirected to logging
--- a/kedro/runner/runner.py
+++ b/kedro/runner/runner.py
@@ -343,8 +343,9 @@ class AbstractRunner(ABC):
                 f"There are {len(remaining_nodes)} nodes that have not run.\n"
                 "You can resume the pipeline run from the nearest nodes with "
                 "persisted inputs by adding the following "
-                f"argument to your previous command:\n{postfix}"
+                f"argument to your previous command:"
             )
+            self._plain_logger.warning(postfix)
 
     @staticmethod
     def _release_datasets(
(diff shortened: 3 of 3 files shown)


astrojuanlu:
Thanks for this PR @david-stanley-94 !

A couple of questions:

1. I see you're focusing specifically on resume suggestions. Are there other situations that make the logs non copy-pasteable we should look at?
2. Is it feasible to achieve this without even touching the Kedro core? For example, by configuring a plain logger coming from a plugin or from a custom module. Would it work?

david-stanley-94:
> Thanks for this PR @david-stanley-94 !
> 
> A couple of questions:
> 
> 1. I see you're focusing specifically on resume suggestions. Are there other situations that make the logs non copy-pasteable we should look at?
> 2. Is it feasible to achieve this without even touching the Kedro core? For example, by configuring a plain logger coming from a plugin or from a custom module. Would it work?

1. Have searched the kedro repo for `_logger`, `logger.`, and can only find one other instance, in the `__init__.py` file of the `ipython` dir, where the plain text handler might be useful:

```python
def _resolve_project_path(
    path: str | None = None, local_namespace: dict[str, Any] | None = None
) -> Path:
    """
    Resolve the project path to use with reload_kedro, updating or adding it
    (in-place) to the local ipython Namespace (``local_namespace``) if necessary.

    Arguments:
        path: the path to use as a string object
        local_namespace: Namespace with local variables of the scope where the line
            magic is invoked in a dict.
    """
    if path:
        project_path = Path(path).expanduser().resolve()
    else:
        if (
            local_namespace
            and local_namespace.get("context")
            and hasattr(local_namespace["context"], "project_path")
        ):
            project_path = local_namespace["context"].project_path
        else:
            project_path = _find_kedro_project(Path.cwd())
        if project_path:
            logger.info(
                "Resolved project path as: %s.\nTo set a different path, run "
                "'%%reload_kedro <project_root>'",
                project_path,
            )
```

2. Good question! Let me try to look over the kedro plugins docs, etc, to better understand how that might work. Any additional guidance appreciated.

merelcht:
Is this a lot different from the suggestion in the docs https://docs.kedro.org/en/stable/logging/index.html#how-to-use-plain-console-logging ?

noklam:
I understand the problem here when `rich` breaks useful information into chunks and make it unsable. I wonder if there are way to solve this in a generic way. This PR target the resume suggestion, but the line break are problematic for things like file path where you can normally do `Cmd + Click`.

I see there are option for soft-wrapping but it doesn't seem to be available for `rich.pretty.install`.
https://rich.readthedocs.io/en/stable/console.html

merelcht:
After running both the `console` logger and your `plain` logger, I see the main difference is that with your logger you don't get stuff like `2025-05-14 16:36:35,637 - kedro.runner.sequential_runner - WARNING` in front of the log lines. 

I feel like this is kind of a personal preference thing and I'm not sure this actually adds significant value that merits this being added to the Kedro core. With the `console` logger you can already achieve the goal of copy-pasteable log lines.

david-stanley-94:
Yeah the aim is to solve issues akin to that described in #3276 , as opposed to more generally changing how logging is done. Ideally the logging formatting/behaviour would stay exactly as it is, just that long resume commands are copy pasteable (as implicitly intended and is the case for short ones) and long filepaths are correctly clickable (as is the case for short ones).

One use case would be running a large pipeline taking 30 mins to run, that breaks almost at the end. The user can copy paste out the resume command, and then fix the bug. Currently they will then have to spend some time (several minutes? I forget how long this has taken me before) fixing the broken command, or else rerun from the beginning (taking 30 mins). Ideally, they would just be able to copy paste straight back into the terminal without delay (a couple of seconds).

astrojuanlu:
I haven't said anything here because I haven't had time to properly review the changes but I think we should not underestimate the impact of Kedro logs not being easily copy-pasteable by default. It's something that bites me almost every week.

ravi-kumar-pilla:
Hi @david-stanley-94 ,

Thank you for the contribution. Apologies for the delayed response on this. 

Have you tried [Rich Text](https://rich.readthedocs.io/en/latest/text.html) and would it be helpful if we use it in areas where copy-pasting should be supported, instead of adding a new handler ?

I tried using rich Text for the example screenshot and it outputs similar to what you have with the new handler -

image

Let me know what you think of this. Thank you

david-stanley-94:
Thanks for the suggestion, @ravi-kumar-pilla! Would need to test if that works as an alternative.

I wanna say I tested this at the time and that it didn't, but tbh it's so long ago now I don't remember, and it seems I didn't document what things, other than a different logger, that I had tried (sorry about that).

merelcht:
Hi @david-stanley-94, it's been a while! Do you still want to work on this? Otherwise I suggest we close the PR.

david-stanley-94:
@merelcht Happy to close, don't have time to return to this right now sorry

merelcht:
No worries @david-stanley-94 ! Really appreciate your effort to improve Kedro 🙏

Issue #3575, opened 2024-01-30
Spike: proposal for better default node names


astrojuanlu:
## Description

When a node has no name, the default is to use a string representation for it. For example:

```
'split_data([model_input_table; params: model_options]) -> [X_train;X _test;y_train;y_test]'
```

This poses a series of problems:

- https://github.com/kedro-org/kedro-plugins/issues/397 "and node.name defaults to the signature"
- https://github.com/kedro-org/kedro/issues/1828 was already discussed in tech design, but no significant changes were made to node name defaults
- Users defaulting to `func.__name__` for a more convenient node name https://linen-slack.kedro.org/t/16107421/a-tip-i-wanted-to-share-you-can-name-a-node-using-the-dunder#0e36ce67-23d8-4638-9e79-113d66b2229d
- Surprising behavior when trying to filter nodes without names https://github.com/kedro-org/kedro/pull/3510#issuecomment-1905943315

## Context

I contend that `'split_data([model_input_table; params: model_options]) -> [X_train;X _test;y_train;y_test]'` is a bad name because:

- A human would never name their nodes like that
- It's more of a description or a `__repr__` rather than a "name"

Plus all the problems discussed above.

## Task
Come up with a proposal for a better default node name, taking into account:
- backwards compatibility 
- filtering by node should be user with the new default
- requirements outlined in discussion

## Possible Implementation

Use `func.__name__` as default node names.

deepyaman:
> Use `func.__name__` as default node names.

Don't node names need to be unique? Even if not ideal, the current approach guarantees uniqueness, while the latter very frequently won't.

astrojuanlu:
Correct: the current approach guarantees uniqueness under the current assumptions of "no two nodes can output the same dataset".

More strawman proposals:

- `f"{func.__name__}_{uuid4()}`
- `f"{func.__name__}_0` (and continues with `_1`, `_2` if node functions are reused)
- `f"{func.__name__}_{simplify(self.outputs())}` (for instance `split_data__X_train_X_test_y_train_y_test`)

More ideas?

noklam:
It may be obvious, but can I ask why node name need to be unique? I feel like they exists for different purposes, I would like to map them out and see if there are opportunities to combine them. There would be at least 2 different purpose, one is for internal working, the other for human-readable.

Currently there are different names.
- node.name (with namespace)
- node.short_name
- node._name
- node._unique_key (hashable)
- node._func_name

[...]

PR #4715, opened 2025-05-09
Proposal: Change default node name

merelcht:
## Description
Addresses https://github.com/kedro-org/kedro/issues/3575

This PR proposes changes to improve the readability and consistency of default node names while preserving the uniqueness requirement crucial for deployment tooling.

### 1. Consistency in name printing
Make the `kedro registry describe` command return the actual node `name` and not `node._name or node._func_name`. This makes it easier for users to check what the node name is that Kedro recognises for commands like `kedro run --nodes=...` https://github.com/kedro-org/kedro/blob/3ddf28ad28799b8561058d5276281168116a33c1/kedro/framework/cli/registry.py#L49

This change is not linked to the other 2 and I'd suggest doing this even if none of the other changes are accepted.

### 2. Use function name as default node name + **add on hash**
Make the default node name to be either the name the user has provided or the function name (as suggested in #3575 ) https://github.com/kedro-org/kedro/blob/3ddf28ad28799b8561058d5276281168116a33c1/kedro/pipeline/node.py#L282

**Updated:** after the feedback, I've updated the implementation to use the function_name for the base of the default node name and append a unique hash to that
https://github.com/kedro-org/kedro/blob/940fb0003c14ad8b5381a276ed4c231370463ed3/kedro/pipeline/node.py#L305-L312

#### 3. Reverted: Validate duplication by `unique_key` instead of node name
**Updated:** after the feedback, I decided to revert this back to the way this was since unique names are essential for deployment.

Make the `_validate_duplicate_nodes` logic to use the `unique_key` instead of the node name to check for duplication. 
https://github.com/kedro-org/kedro/blob/7e68c245233623ef32fecc6f4a40828c1e68a11f/kedro/pipeline/pipeline.py#L1288-L1293

Also update the error message to mention what duplication means for nodes: combination of function used, inputs and outputs:

https://github.com/kedro-org/kedro/blob/7e68c245233623ef32fecc6f4a40828c1e68a11f/kedro/pipeline/pipeline.py#L1312-L1315

This change can also be done even if we decide not the change the default node naming. I completely agree with Nok in https://github.com/kedro-org/kedro/issues/3575#issuecomment-1936693023

## Checklist


noklam (commented):
quick question without running the code yet.

What's the expected outcome when the same function is applied in two nodes? I expected to see this as a test case as this is the main issue with using the function name directly as node name. 

Does this also affect how `node` is printed in the Kedro run logs?

merelcht:
> quick question without running the code yet.
> 
> What's the expected outcome when the same function is applied in two nodes? I expected to see this as a test case as this is the main issue with using the function name directly as node name.

I have not updated the tests yet 😄 haha I wanted to see if people like this before doing that. Will update the description to mention that.

kedro/pipeline/node.py
```diff
@@ -281,7 +281,7 @@ def name(self) -> str:
         Returns:
             Node's name if provided or the name of its function.
         """
-        node_name = self._name or str(self)
+        node_name = self._name or self._func_name
```
astrojuanlu, line 284: IIUC this is one of the key bits, right? `name` is a calculated property, and instead of using `__str__` below, it's using `._func_name` in the absence of a given `._name`

kedro/pipeline/pipeline.py
```diff
@@ -613,7 +616,16 @@ def only_nodes(self, *node_names: str) -> Pipeline:
...
             )
 
-        nodes = [self._nodes_by_name[name] for name in node_names]
+        nodes = []
+        for name in node_names:
+            nodes_with_name = self._nodes_by_name[name]
+            if len(nodes_with_name) != 1:
+                raise ValueError(
+                    f"Multiple nodes found with name '{name}'. "
+                    "You can give your nodes a unique name using the last argument of 'node()'."
```
astrojuanlu, line 625: This is the part I don't quite understand. Won't this yield an error in cases where previously it was working? (For example, if the same function is used in 2 nodes). Shouldn't this use the `._unique_key`?

merelcht: This will only complain if the `only_node` filter is triggered, e.g. when running `kedro run --nodes=....` and passing a name that appears more than once. But when doing `kedro run` it will be fine as long as the _combination of arguments_  (which is the unique_key) passed to the node is unique. If we have two nodes:


```
 node(
            func=preprocess_data,
            inputs="companies",
        ),
 node(
            func=preprocess_data,
            inputs="companies_new",
        ),
```

Their names are the same,`preprocess_data`, but the unique keys are different. For node 1 (preprocess_data, companies, preprocessed_companies) and for node 2 (preprocess_data, companies_new, preprocessed_companies) and so the validation will pass.

merelcht: So actually the piece of code above is more a flag argument "ambiguity" check than a node duplication check.

kedro/pipeline/node.py
merelcht: This is the key bit for the second proposal in this PR as described in the PR description. This part makes it possible to use the function name to filter a run by a specific node.

kedro/framework/cli/registry.py
```diff
@@ -46,7 +46,7 @@ def describe_registered_pipeline(
     nodes = []
     for node in pipeline_obj.nodes:
         namespace = f"{node.namespace}." if node.namespace else ""
-        nodes.append(f"{namespace}{node._name or node._func_name} ({node._func_name})")
+        nodes.append(f"{namespace}{node.name} ({node._func_name})")
```
479BetaCoder, line 49: I like the proposals, but I have a question on the implementation here. if `node.name` defaults to `node._func_name` and also has a namespace appended in the [implementation](https://github.com/kedro-org/kedro/pull/4715/files#diff-26deb45c9dfb4d41c486319bdfd4e858f42aa728e2efbcbeb68496a0445f11adR278), why are we doing this here again ?

ravi-kumar-pilla, line 49: I like the proposals, but I have a question on the implementation here. if node.name defaults to node._func_name and also has a namespace appended in the [implementation](https://github.com/kedro-org/kedro/pull/4715/files#diff-26deb45c9dfb4d41c486319bdfd4e858f42aa728e2efbcbeb68496a0445f11adR278), why are we doing this here again ?

ravi-kumar-pilla: My bad, so this is an independent implementation even if (2) proposal does not go in ?

ravi-kumar-pilla:
Hi @merelcht , I like the proposals. Thank you

DimedS:
Thanks for the analysis and proposal, @merelcht!

I really like the idea of making node names clearer - the current situation with overly long default names, especially when users don't explicitly name the node, is not ideal.

I support **Proposal 2**, which suggests using the function name as the basis for the node name. However, we do need to ensure this approach avoids potential name duplication.

I'm not in favour of **Proposal 3**, which suggests allowing non-unique node names. From a deployment perspective, this would be problematic. Kedro deployment plugins typically rely on unique node names to execute converted pipelines node by node, e.g.:

`kedro run --node=node_name_1`

If node names are no longer unique, it would create significant inconsistencies and break downstream tooling. Therefore, I strongly recommend we maintain uniqueness in node names.

kedro/framework/cli/registry.py
merelcht: Yes indeed! The cli command logic is handled separately.

datajoely:
As a user, I can say I never explicitly name the nodes and rely on the defaults. I quiet like the idea of relying on the function name but don't really have a great solution for the duplicate issue, perhaps classpath as a fallback?

deepyaman:
Largely agree with @DimedS that deduplication is important, especially for deployment. At the same time (as one of the motivations for this issue), the current node names aren't suitable for deployment, either (also see https://github.com/gtauzin/kedro-dagster/pull/13 as a recent example).

An incrementing ID to deal with duplication is OK, but I wonder if "stability" of node names matters? E.g. I wouldn't want my node to be `my_node_2` today and `my_node_1` tomorrow. Of course, you can say users can explicitly name nodes, but--same as @datajoely--I honestly don't do it as a user.

This may be overcomplicating it, but I think:

- Current node name is a good `repr`
- `func.__name__` is a good default name (e.g. display in Viz)
- Nodes should have an ID; this could be `sha1(repr())` or something? (or the ID as a suffix)

The first two feel pretty aligned also with what @astrojuanlu said in the initial issue.  For the last bit, I think it would be nice to have such an ID, or you could even document a "standard" ID recommended for deployment where you need something unique--but need this to solve the common deployment issue.

(Apologies in advance if restated stuff people already said; skimmed things a bit quickly after a long day. Happy to reread this later and/or be corrected if I made any dumb comments.)

merelcht:
Thanks all for the thoughtful feedback!

Based on the comments, I’ll update the PR to:

- Use the function name as the base for default node names
- Preserve uniqueness by suffixing duplicates deterministically
- Keep the uniqueness requirement for node names (not relaxing it via unique_key) 
- Keep the `kedro registry describe` change

I’ll push a revised version shortly and would love your thoughts once it's updated.

rashidakanchwala:
I like the proposal. I’d suggest that in cases where duplicate node names are detected, we automatically append a hash only to those duplicates to ensure uniqueness to support deployment usecase. The hash could be based on `node_func(input → output)`, resulting in names like `process_data__abc123`, `process_data__yvj134`

merelcht:
> I like the proposal. I’d suggest that in cases where duplicate node names are detected, we automatically append a hash only to those duplicates to ensure uniqueness to support deployment usecase. The hash could be based on `node_func(input → output)`, resulting in names like `process_data__abc123`, `process_data__yvj134`

I've done pretty much what you describe here, but for all nodes and not just for the duplicate nodes. The name is added in the node class, so there we don't know what other nodes exist and detecting duplication would be quite tricky.

DimedS (approved):
thanks, @merelcht ! I support your last proposal

tests/pipeline/test_node.py
```diff
@@ -420,7 +420,7 @@ def test_invalid_name(self, bad_name):
     def test_namespaced(self):
         n = node(identity, ["in"], ["out"], namespace="namespace")
         assert str(n) == "identity([in]) -> [out]"
-        assert n.name == "namespace.identity([in]) -> [out]"
+        assert n.name == "namespace.identity__d8883cb5"
```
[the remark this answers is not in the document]
rashidakanchwala: @merelcht , i think this might be a good idea. We faced this issue in kedro-viz; when we made a few changes in the tests; all the hashes got updated and it was a pain to change them.

rashidakanchwala (approved):
Looks good, thanks @merelcht . left a minor comment on tests

astrojuanlu (approved):
I did QA like this:

```
In [1]: from kedro.pipeline import Pipeline

In [2]: from kedro.pipeline.node import Node

In [9]: def passthrough(in_):
   ...:     return in_
   ...: 

In [10]: p = Pipeline([Node(func=passthrough, inputs=["a"], outputs=["b"])])

In [11]: p.nodes[0].name
Out[11]: 'passthrough__c0a78389'
```

With the latest released version:

```
In [1]: from kedro.pipeline import Pipeline

In [2]: from kedro.pipeline.node import Node

In [3]: def passthrough(in_):
   ...:     return in_
   ...: 

In [4]: p = Pipeline([Node(func=passthrough, inputs=["a"], outputs=["b"])])

In [5]: p.nodes[0].name
Out[5]: 'passthrough([a]) -> [b]'
```

LGTM!

PR #4729, opened 2025-05-13
Refactor grouped_nodes_by_namespace() to return structured GroupedNode list

DimedS:
This is the updated version of #4708, now targeting the `develop` branch instead of `main`, since this is a breaking change intended for the upcoming major release.

## Description
First part of #4684 - modified `Pipeline.grouped_nodes_by_namespace()` to return a list of `GroupedNode` objects instead of a dictionary.

This improves the structure and type safety of the API, and makes it easier for deployment plugins and other consumers to work with node grouping results in a uniform way.

There was flakiness in the tests because
```
@property  
def node_dependencies(self) -> dict[Node, set[Node]]:  
```
is not deterministic due to the use of set.
To address this, I implemented additional sorting in `grouped_nodes_by_namespace().`

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -15,6 +15,7 @@
 * Renamed instances of `extra_params` and `_extra_params` to `runtime_params`.
 * Removed the `modular_pipeline` module and moved functionality to the `pipeline` module instead.
 * Renamed `ModularPipelineError` to `PipelineError`.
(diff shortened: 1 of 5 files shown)


lrcouto (approved):

merelcht (approved):
Thanks @DimedS, this looks good to me 👍 

Let's add this to the release notes as well, with a short mention that this is for deployment.

PR #4745, opened 2025-05-15
Extend the node grouping API

DimedS:
## Description
Solves #4685 

## Checklist


kedro/pipeline/node.py
```diff
@@ -31,7 +31,7 @@ class GroupedNodes:
     """
 
     name: str
-    type: str  # "namespace" or "node"
+    type: str  # "namespace" or "nodes"
```
ankatiyar, line 34: Just wondering if maybe this should be expanded too: namespace, single-node, node-group etc?

kedro/pipeline/pipeline.py
```diff
@@ -541,13 +541,52 @@ def grouped_nodes(self) -> list[list[Node]]:
 
         return [list(group) for group in self._toposorted_groups]
 
-    @property
-    def grouped_nodes_by_namespace(self) -> list[GroupedNodes]:
-        """Return a list of grouped nodes by top-level namespace with
-        information about the nodes, their type, and dependencies.
+    def grouped_nodes_custom(
```
ankatiyar, line 544: Suggestion on the name, could be `grouped_nodes_for_deployment`?

ankatiyar (commented):
overall looks good, some minor suggestions on names and one question about the memory grouping.

lrcouto:
Looking good! In general, I agree with Ankita's suggestions as well.

kedro/pipeline/node.py
DimedS: I considered that approach, but decided it would introduce unnecessary complexity. If someone wants to check whether there's only one node, they can simply use `len(nodes)==1` instead of introducing a new hardcoded type.

kedro/pipeline/pipeline.py
```diff
@@ -575,6 +613,66 @@ def grouped_nodes_by_namespace(self) -> list[GroupedNodes]:
 
         return list(grouped_nodes_map.values())
 
+    # ruff: noqa: PLR0912
+    def _group_by_memory(self, memory_datasets: set[str]) -> list[GroupedNodes]:
```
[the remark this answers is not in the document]
DimedS: I think we can check it - what do you think we should do if the name isn’t present in the current pipeline?

kedro/pipeline/pipeline.py
```diff
@@ -541,13 +541,52 @@ def grouped_nodes(self) -> list[list[Node]]:
...
+            return self._group_by_namespace()
+        if group_by.lower() == "memory":
+            if memory_datasets is None:
+                raise ValueError(
+                    "'memory_datasets' must be provided when grouping by memory."
+                )
+            return self._group_by_memory(memory_datasets)
+        raise ValueError(f"Unsupported group_by strategy: {group_by}")
+
+    def _group_by_none(self) -> list[GroupedNodes]:
```
merelcht, line 573: This confuses me.. why are we returning `GroupedNodes` when we don't want to group? I'm guessing it's to help with returning a standard format for deployment? 

Can you describe the use case in which this behaviour is triggered?

kedro/pipeline/pipeline.py
merelcht, line 617: I don't feel comfortable adding this to the core of Kedro. I remember we had a pretty big discussion when this behaviour was added to Kedro Airflow in https://github.com/kedro-org/kedro-plugins/pull/241. It feels like a rash decision to just add it here without talking about it again.

The main argument against grouping by memory was that namespaces should be the advised way to group things. However, at the time there was no out of the box way to do that. Now we've added that, I feel like the grouping by memory shouldn't be encouraged anymore. 

We can always add this post 1.0.0, since that's a non-breaking change, but I want to discuss it first.

merelcht (changes requested):
The generalisation looks good, but I have some questions and concerns about the current grouping methods added. I'd prefer to keep it to namespaces now and only work on extending later on if needed.

We've had many discussions on grouping in which we decided to recommend namespaces going forward, but this implementation seems to go against this again by offering other ways of grouping which we never agreed on as being recommended.

kedro/pipeline/pipeline.py
DimedS: Thanks, @merelcht — you're right. That unification was proposed during the plugin developers' interview.

You can check the relevant PR here: https://github.com/kedro-org/kedro-plugins/pull/1072/files.  
It significantly reduces plugin code complexity — instead of handling each case manually, you simply call the `grouped_nodes_custom()` API, passing a `group_by` option (e.g., `"namespace"`, `"memory"`, or `None`). The result is a unified list of nodes that can be easily transformed into another pipeline format, such as Airflow.

Previously, without this approach, each grouping case had to be handled separately within the plugin logic.

For example, in `kedro-airflow`, when the user runs `kedro airflow create` without grouping, the plugin calls:

```python
deployment_objects = grouped_nodes_custom(group_by=None)
```

This returns a stable structure — a flat list of nodes — which is then converted to an Airflow DAG.

If the user wants grouping (e.g., by memory or namespace), the same API is used with the corresponding group_by value. The structure of the output remains consistent, making further processing straightforward.

kedro/pipeline/pipeline.py
DimedS: I believe that grouping by memory is a distinct strategy, separate from the approaches we discussed when we agreed to recommend namespace-based node grouping for deployment. In our recent documentation update — where we described pipelines, tags, and namespaces — memory grouping is not mentioned:  
https://docs.kedro.org/en/stable/deployment/nodes_grouping.html

This is because memory dataset grouping solves a very specific problem: it allows you to automatically split your pipeline into executable units. This problem cannot be addressed using namespaces, tags, or pipelines alone, since those require manual annotation of nodes.

In my view, we should proceed with supporting the "group by memory" option now because:

1. It was proposed during user interviews with plugin developers (i.e., actual users of this API).
2. It addresses a different use case than namespace grouping — it's a separate solution for a separate problem.
3. It's already implemented in the `kedro-airflow` plugin, where it significantly reduces and simplifies the plugin code.
4. The memory-grouping algorithm is non-trivial. Replicating it in every plugin would be inefficient, but plugin developers still need to solve this issue somehow. By default, plugins treat each node as a separate container run, which doesn’t work for projects like the `spaceflights` starter out of the box. As a result, plugins end up introducing custom workarounds — for example:  
   https://kedro-vertexai.readthedocs.io/en/0.12.0/source/03_getting_started/01_quickstart.html#adjusting-data-catalog-to-be-compatible-with-vertex-ai

DimedS:
> The generalisation looks good, but I have some questions and concerns about the current grouping methods added. I'd prefer to keep it to namespaces now and only work on extending later on if needed.
> 
> We've had many discussions on grouping in which we decided to recommend namespaces going forward, but this implementation seems to go against this again by offering other ways of grouping which we never agreed on as being recommended.

Thanks for the review, @merelcht.

I've responded to each comment in detail. Overall, I believe we should proceed with supporting memory-based grouping for now. This approach is based on feedback from plugin developer interviews and addresses a specific need that isn’t covered by our current recommendation to use namespace-based grouping.

In my view, memory grouping doesn’t conflict with our overall strategy — it solves a different problem. 

Looking ahead, I suggest that after the 1.0.0 release, we revisit our grouping recommendations and consider formally supporting additional grouping strategies. There is clear demand for them, and plugin developers are already using these approaches — for example, grouping by tag has been added in one of our community deployment plugins, and there's strong support in the following discussion:  
https://github.com/kedro-org/kedro/issues/4685#issuecomment-2821107101

If you think it makes sense, we can also bring this group by memory question up in one of the upcoming Tech Design meetings for further discussion.

kedro/pipeline/pipeline.py
DimedS: Fixed

DimedS:
Thanks for your feedback @lrcouto , @ankatiyar , @merelcht !
All changes proposed in the [Tech Design discussion](https://github.com/kedro-org/kedro/issues/4685#issuecomment-2901320077) have been implemented.
Requesting a re-review.
Related changes in kedro-airflow for testing: https://github.com/kedro-org/kedro-plugins/pull/1072

merelcht (approved):
One minor comment, but otherwise LGTM! 👍 

Don't forget to add this to the release notes.

kedro/pipeline/pipeline.py
```diff
@@ -546,13 +546,43 @@ def grouped_nodes(self) -> list[list[Node]]:
 
         return [list(group) for group in self._toposorted_groups]
 
-    @property
-    def grouped_nodes_by_namespace(self) -> list[GroupedNodes]:
-        """Return a list of grouped nodes by top-level namespace with
-        information about the nodes, their type, and dependencies.
+    def group_nodes_custom(
```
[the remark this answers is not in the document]
DimedS: I actually agree. However, considering that we already have a property with a very similar name:

```
@property
def grouped_nodes(self) -> list[list[Node]]:

```
it could be confusing to distinguish between `grouped_nodes` and `group_nodes`. That said, I also agree that using a `custom` postfix does not add clarity in this case.

What do you think, @merelcht ?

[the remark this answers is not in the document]
merelcht: hmm yes good point, maybe we can name it `group_nodes_by` to make it clear you can pass a grouping method?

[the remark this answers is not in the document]
ankatiyar: Maybe `get_node_grouping()`. I also don't like the `custom` part but I realised there's already a `grouped_nodes` property.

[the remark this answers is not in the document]
DimedS: updated to `group_nodes_by`, thanks for the proposals! I changed release.md also

Issue #4742, opened 2025-05-15
[DataCatalog2.0]: Change pattern resolution default behaviour when accessing datasets


ElenaKhaustova:
## Description
After we implemented this https://github.com/kedro-org/kedro/issues/4475#issuecomment-2749010199 there's no difference between the catalog's runtime and non-runtime states now. Everything that is resolved is added to the dataset collection. This behaviour was reported as confusing during the internal testing phase.

Currently, `get()` and `get_type()` always return datasets and types because of the runtime pattern, which is reported as confusing behaviour. A few implementation considerations related to runners made us do such an implementation, but we concluded that this should not affect ordinary users.

We suggest:

1. Preserve a mechanism for `get()` method to fallback to the runtime pattern, but do not do it by default
2. Revise all the method's names for `catalog_config_resolver`, so it is clear how the resolution happens

## Context
 https://github.com/kedro-org/kedro/issues/4740

ElenaKhaustova:
Solved in https://github.com/kedro-org/kedro/pull/4749

PR #4749, opened 2025-05-19
Update pattern resolution behaviour

ElenaKhaustova:
## Description
Solves https://github.com/kedro-org/kedro/issues/4742

## Development notes
Changes made:

1. Added `fallback_to_runtime_pattern` flag to `get()` method which is disabled by default. It is set to `True` at runtime, but by default `get()` and `__getitem__` now only resolve dataset patterns (set via `catalog.yaml`) and do not fallback to runtime patterns.
2. Updated `get_type()` to not fall back to the runtime pattern. Now, it returns an empty string if the dataset is not in the catalog. It relies on the `__contains__` method when checking the dataset's existence in the catalog.
3. Updated other related methods not to fall back to runtime pattern by default.
4. Updated function names for `CatalogConfigResolver`, to explicitly point to what we resolve for, now we have:

- catalog_config_resolver.match_dataset_pattern() - checks if the dataset name matches the dataset pattern
- catalog_config_resolver.match_user_catch_all_pattern() - checks if dataset name matches the user defined catch all pattern
**NOTE: The above two patterns are now the part of the check for dataset existence in the `__contains__` method**

- catalog_config_resolver.match_runtime_pattern() - checks if dataset name matches the default runtime pattern

5. Updated some unit tests to check pattern resolution with `fallback_to_runtime_pattern` enabled/disabled

## Checklist

### DIFF
--- a/kedro/io/catalog_config_resolver.py
+++ b/kedro/io/catalog_config_resolver.py
@@ -80,15 +80,17 @@ class CatalogConfigResolver:
         self,
         config: dict[str, dict[str, Any]] | None = None,
         credentials: dict[str, dict[str, Any]] | None = None,
-        runtime_patterns: Patterns | None = None,
+        default_runtime_patterns: Patterns | None = None,
     ):
-        if runtime_patterns is None:
+        if default_runtime_patterns is None:
             self._logger.warning(
                 f"Since runtime patterns are not provided, setting"
                 f"the runtime pattern to default value: {DEFAULT_RUNTIME_PATTERN}"
             )
-        self._runtime_patterns = runtime_patterns or DEFAULT_RUNTIME_PATTERN
-        self._dataset_patterns, self._default_pattern = self._extract_patterns(
+        self._default_default_runtime_patterns = (
+            default_runtime_patterns or DEFAULT_RUNTIME_PATTERN
+        )
+        self._dataset_patterns, self._user_runtime_pattern = self._extract_patterns(
             config, credentials
         )
         self._resolved_configs = self._resolve_credentials(config, credentials)
@@ -201,8 +203,8 @@ class CatalogConfigResolver:
         """List al patterns available in the catalog."""
         return (
             list(self._dataset_patterns.keys())
-            + list(self._default_pattern.keys())
-            + list(self._runtime_patterns.keys())
+            + list(self._user_runtime_pattern.keys())
+            + list(self._default_default_runtime_patterns.keys())
         )
 
     @classmethod
@@ -214,10 +216,10 @@ class CatalogConfigResolver:
         matches = self._get_matches(self._dataset_patterns.keys(), ds_name)
         return next(matches, None)
 
-    def match_default_pattern(self, ds_name: str) -> str:
+    def match_runtime_pattern(self, ds_name: str) -> str:
         """Match a dataset name against default and runtime patterns in a dictionary."""
-        default_patters = set(self._default_pattern.keys())
-        default_patters.update(set(self._runtime_patterns.keys()))
+        default_patters = set(self._user_runtime_pattern.keys())
+        default_patters.update(set(self._default_default_runtime_patterns.keys()))
         matches = self._get_matches(default_patters, ds_name)
         # We assume runtime pattern always matches at the end
         return next(matches)
@@ -225,8 +227,8 @@ class CatalogConfigResolver:
     def _get_pattern_config(self, pattern: str) -> dict[str, Any]:
         return (
             self._dataset_patterns.get(pattern)
-            or self._default_pattern.get(pattern)
-            or self._runtime_patterns.get(pattern)
+            or self._user_runtime_pattern.get(pattern)
+            or self._default_default_runtime_patterns.get(pattern)
             or {}
         )
 
@@ -326,7 +328,7 @@ class CatalogConfigResolver:
         if ds_name not in self._resolved_configs:
             matched_pattern = self.match_dataset_pattern(
                 ds_name
-            ) or self.match_default_pattern(ds_name)
+            ) or self.match_runtime_pattern(ds_name)
             pattern_config = self._get_pattern_config(matched_pattern)
             ds_config = self._resolve_dataset_config(
                 ds_name, matched_pattern, copy.deepcopy(pattern_config)
@@ -334,7 +336,7 @@ class CatalogConfigResolver:
 
             if (
                 self._pattern_specificity(matched_pattern) == 0
-                and matched_pattern in self._default_pattern
+                and matched_pattern in self._user_runtime_pattern
             ):
                 self._logger.warning(
                     "Config from the dataset factory pattern '%s' in the catalog will be used to "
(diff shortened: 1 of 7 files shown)


kedro/io/data_catalog.py
```diff
@@ -230,22 +229,31 @@ def __len__(self) -> int:
...
+        Raises:
+            DatasetError: When the dataset in not in the internal collection, does not match
+                dataset pattern and fallback_to_runtime_pattern is set to False.
+
         Returns:
             An instance of AbstractDataset.
         """
-        if key not in self._datasets and key not in self._lazy_datasets:
+        if key not in self and not fallback_to_runtime_pattern:
+            raise DatasetNotFoundError(f"Dataset '{key}' not found in the catalog")
```
ravi-kumar-pilla, line 255: Hi @ElenaKhaustova , this looks better than before. What happens if the user defined a default pattern ? Since contains also does not account for this, it will throw a DatasetNotFoundError.
Shouldn't we just ignore runtime_patterns ?

kedro/io/catalog_config_resolver.py
```diff
@@ -209,13 +209,13 @@ def list_patterns(self) -> list[str]:
...
-    def match_pattern(self, ds_name: str) -> str | None:
-        """Match a dataset name against catalog patterns in a dictionary."""
+    def match_dataset_pattern(self, ds_name: str) -> str | None:
+        """Match a dataset name against catalog patterns."""
         matches = self._get_matches(self._dataset_patterns.keys(), ds_name)
         return next(matches, None)
 
     def match_default_pattern(self, ds_name: str) -> str:
-        """Match a dataset name against default patterns in a dictionary."""
+        """Match a dataset name against default and runtime patterns in a dictionary."""
```
ravi-kumar-pilla, line 218: [Nit] It would be nice to separate `match_default_pattern` and `match_runtime_pattern` too ?

ravi-kumar-pilla: The way I am thinking is `_default_pattern` is also defined by the user, so this should come under `_dataset_patterns`. Just because it is a catch_all pattern, may be we distinguish it with `_dataset_patterns`

ElenaKhaustova: Please see the above clarification.

We can split them, but the use case for applying them separately is not clear to me. Do you have any specific case in mind?

kedro/io/data_catalog.py
ravi-kumar-pilla: okay, thanks for explaining. My understanding was also the same. But I think catch_all pattern defined by the user is also a factory pattern at the catalog.yml level. If the user did define it, it should not be a `runtime_pattern` injected by Kedro. 

So, I feel may be we show a debug info or warning saying that user defined catch-all pattern is used when we match with the catch-all pattern. Still it should be part of `contains` check I guess (because it is a user defined one and we did not do anything).

ElenaKhaustova: I see your points, and I don't mind that. The only side effect I see with this change is that we will keep resolving everything as it was before with the runtime pattern. So when the user adds the catch-all pattern, it will start catching everything. If we're happy with this, I'll go ahead with the suggestion.

ElenaKhaustova: @ravi-kumar-pilla wdyt?

ravi-kumar-pilla: I think we should do that as it is not something we are doing but the user opted for. But I will wait for @merelcht to respond.

ElenaKhaustova: I updated the logic as suggested to try it out

kedro/io/catalog_config_resolver.py
```diff
@@ -201,32 +203,37 @@ def list_patterns(self) -> list[str]:
         """List al patterns available in the catalog."""
         return (
             list(self._dataset_patterns.keys())
-            + list(self._default_pattern.keys())
-            + list(self._runtime_patterns.keys())
+            + list(self._user_catch_all_pattern.keys())
+            + list(self._default_runtime_patterns.keys())
```
ravi-kumar-pilla, line 207: can there be multiple `catch_all_pattern` and `runtime_patterns` ?

ElenaKhaustova: Technically, there can be a few catch_all_patterns if users set them via catalog.yaml, but that's not the expected behaviour, I think. Currently, we only take the last catch_all_pattern - the rest will remain in dataset patterns, and the first match will be used. As for runtime patterns - the implementation can handle several - but at the class level we set just one. Probably the case when multiple can be applied is when users redefine catalog with the runtime patterns. But maybe we can reduce it to one and adjust the naming correspondingly as multiple patterns can be set via catalog.yaml anyway.

ravi-kumar-pilla: I think documentation should do and leave it to users if they want to specify multiple. So no code changes are needed

kedro/io/catalog_config_resolver.py
```diff
@@ -80,15 +80,17 @@ def __init__(
...
             self._logger.warning(
                 f"Since runtime patterns are not provided, setting"
                 f"the runtime pattern to default value: {DEFAULT_RUNTIME_PATTERN}"
             )
-        self._runtime_patterns = runtime_patterns or DEFAULT_RUNTIME_PATTERN
-        self._dataset_patterns, self._default_pattern = self._extract_patterns(
+        self._default_runtime_patterns = (
+            default_runtime_patterns or DEFAULT_RUNTIME_PATTERN
+        )
+        self._dataset_patterns, self._user_catch_all_pattern = self._extract_patterns(
```
ravi-kumar-pilla, line 93:
```suggestion
        self._dataset_patterns, self._user_catch_all_patterns = self._extract_patterns(
```

ravi-kumar-pilla:
Hi @ElenaKhaustova , I left some minor comments but the functionality works as expected. Awesome work !!

ravi-kumar-pilla (approved):

kedro/io/data_catalog.py
```diff
@@ -552,12 +564,11 @@ def filter(
 
     def get_type(self, ds_name: str) -> str:
         """Access dataset type without adding resolved dataset to the catalog."""
+        if ds_name not in self:
+            return ""
```
merelcht, line 568: Maybe this is something we should validate with users, IMO it also makes sense to return an error here saying the dataset doesn't exist/not found in the catalog.

merelcht (approved):
One minor comment, but otherwise the good looks good and works as expected 👍

kedro/io/catalog_config_resolver.py
ElenaKhaustova: Here we only use one, as explained before. We take dataset patterns, sort them and take the last catch all pattern as `user_catch_all_pattern`, so the name is not in plural.

kedro/io/data_catalog.py
ElenaKhaustova: Yeah, it was planned to be an internal method first. But then I realised it's also a good API for users as it allows getting type without resolving datasets. 

Updated it to return the `DatasetNotFoundError`

kedro/io/data_catalog.py
```diff
@@ -551,13 +563,17 @@ def filter(
         return filtered
 
     def get_type(self, ds_name: str) -> str:
-        """Access dataset type without adding resolved dataset to the catalog."""
+        """Access dataset type without adding resolved dataset to the catalog.
+
+        Raises:
+            DatasetError: When the dataset in not in the internal collection, does not match
```
ravi-kumar-pilla, line 569: DatasetNotFoundError

Issue #4635, opened 2025-04-02
Explore the “Run only missing” feature


merelcht:
## Introduction
Through the years a feature to skip certain nodes or run only nodes with missing data has been requested over and over again. It's not a trivial problem to solve and requires holding some state. Now we're preparing to release `1.0.0` it seems like a good time to revisit this request once again and decide whether it's worth adding it for this major release.

## Background
https://github.com/kedro-org/kedro/issues/2307
https://github.com/kedro-org/kedro/issues/2410
https://github.com/kedro-org/kedro/issues/221

## Problem
Re-running nodes which have:

- Persisted outputs
- No upstream dependencies which would cause their output to change

is an unnecessary expense. It might be a good idea to have a flag which would automatically skip running these nodes.

It is currently possible to achieve this by specifying nodes to run from, though this process is manual and potentially error-prone.

### What's in scope
The aim of this Spike is to propose a way to make it possible to only run nodes that must run because they haven't persisted data.

### What's not in scope
The actual implementation. After the spike we'll decide if it's worth implementing this for `1.0.0` and hold a TSC vote for approach.

ravi-kumar-pilla:
Hi Team,

After collecting some information, understanding the context and a brief discussion with @astrojuanlu, here are my thoughts -

**Skipping node execution earlier discussions:**

**1. Implementing a --run-only-missing CLI Flag**

* Flag Introduction: Introduce a new command-line interface (CLI) flag, e.g., --run-only-missing, to the kedro run command. 
* Node Evaluation: During pipeline execution, for each node, check if all its outputs are already present and up-to-date. If so, and if none of its inputs have changed, skip the node execution. 
* Integration: Modify the pipeline runner to incorporate this conditional execution logic based on the presence of the --run-only-missing flag. 

  **Pros:**
  * User Control: Provides users with explicit control over this behavior.
  * Backward Compatibility: Does not alter existing default behaviors; users opt-in to the new feature. 
  * Simplicity: Straightforward to implement and understand.

  **Cons:**
  * User Responsibility: Relies on users to remember to use the flag when appropriate. 
  * Limited Automation: Does not automatically detect and skip redundant computations without user intervention.

  Ivan’s opinion on the flag - https://github.com/kedro-org/kedro/issues/30#issuecomment-513215607

<hr>

**2. Custom Pipeline Runner with Caching Mechanism**

[...]

PR #4789, opened 2025-05-29
Add `--only-missing-outputs` CLI flag to skip nodes with existing outputs

SajidAlamQB:
## Description

Related to: https://github.com/kedro-org/kedro/issues/4635

This PR implements Approach 2 from the @ravi-kumar-pilla spike #4635, adding a new `--only-missing-outputs` CLI flag to `kedro run` that allows users to skip nodes whose outputs already exist in the catalog.

When this flag is enabled, Kedro will skip the execution of nodes if all their persistent outputs already exist in the catalog.

For this initial MVP, the feature focuses on the existence of persistent datasets and does not involve hashing of data or code for change detection.

## How it works

**Important:** This MVP implementation uses a **pipeline-level approach** rather than node-by-node checking:

1. First, it identifies all **final persistent outputs** that are missing from the entire pipeline
2. Then, it builds the minimal subgraph needed to produce those missing outputs using `pipeline.to_outputs()`
3. Finally, it runs **all nodes in that subgraph**, even if some intermediate outputs already exist

### Example:
If you have a pipeline: `A → B → C` where `C` is missing but `B` exists:
- The feature will run both `A → B → C` to ensure consistency
- It does NOT check if `B` exists and skip just node `A`

More granular node-level checking with change detection is planned for post-1.0.0 see spike, https://github.com/kedro-org/kedro/issues/4635.

## Development notes

1. **CLI**: Added `--only-missing-outputs` flag to the `kedro run` command.
2. **Session**: Modified `KedroSession.run()` to accept the new parameter and use the appropriate runner method.
3. **Runner**: Updated to include `only_missing_outputs: bool = False`.
4. Removed `run_only_missing` legacy code and its test.

If `only_missing_outputs` is True, the runner first calls a new private method: 

`_filter_pipeline_for_missing_outputs(pipeline, catalog)`: This method returns a potentially modified (reduced) pipeline.

This new method is responsible for identifying which nodes need to run. It iterates through all nodes in the original pipeline to find all unique persistent outputs that are currently missing from the catalog, using the `_is_persistent_and_missing()` helper.

If no persistent outputs are missing, it returns an empty Pipeline([]), effectively skipping all nodes.

If there are missing persistent outputs, it uses `pipeline.to_outputs(*missing_persistent_outputs)` to construct a new pipeline. This should nherently handle the exclusion of "wasteful" nodes.

`_is_persistent_and_missing`:

This new helper method determines if a given dataset name corresponds to a persistent dataset that does not currently exist.

It first attempts to instantiate the dataset using `catalog.get(output)`.
It correctly identifies ephemeral datasets (e.g., those with `_EPHEMERAL = True`, like `MemoryDataset`) and considers them as "not needing to be checked for existence" for this feature's purpose.

For persistent datasets, it uses `catalog.exists(output)` to check for current data.
It handles datasets defined explicitly in the catalog, those created by dataset factories, and those implicitly becoming `MemoryDataset` if undefined.

## Testing with Spaceflights

**Scenario: All Outputs Exist**

- Run `kedro run` once to generate everything.
- Run `kedro run --only-missing-outputs`.
- Expected: All nodes should be skipped. Timestamps on files like data/03_primary/model_input_table.parquet, data/06_models/regressor.pickle, and files in data/08_reporting/ should not change.

- Run `kedro run` once.
- Delete data/03_primary/model_input_table.parquet.
- Run `kedro run --only-missing-outputs`.
- Expected:
`create_model_input_table_node` must run.
`preprocess_companies_node` and `preprocess_shuttles_node` must run because their outputs are inputs to `create_model_input_table_node`.

- Temporarily edit `conf/base/catalog.yml` for regressor: change type: `pickle.PickleDataset` to type: `pandas.CSVDataset`.
- Delete data/06_models/regressor.pickle if it exists.
- Run `kedro run --only-missing-outputs`.
- Expected: The run should fail.

**Scenario: Interaction with `--to-nodes `**

- Run `kedro run once`.
- Run `kedro run --to-nodes split_data_node --only-missing-outputs`.
- Expected:
The `--only-missing-outputs` flag operates on the pipeline after other filters (like `--to-nodes`) have been applied.
It then looks for persistent outputs of that specific filtered pipeline that are missing.
Since all persistent outputs of the `--to-nodes split_data_node` pipeline already existed, nothing needed to be run.

**Scenario: Running a specific pipeline**

- Run `kedro run once`.
- Delete data/06_models/regressor.pickle.
- Run `kedro run --pipeline data_science --only-missing-outputs`.
- Expected:
`train_model_node` must run.
`split_data_node` must run (ephemeral outputs).

## Checklist


ankatiyar:
I have a question: when I tried to run this with spaceflights, it also runs the `split_data_node` because its outputs are `MemoryDatasets`. Is there a way to distinguish nodes where the outputs/inputs are meant to be memory datasets vs nodes with inputs/outputs "missing" i.e. not persisted?

kedro/runner/runner.py
```diff
@@ -155,6 +155,42 @@ def run(
...
+
+        """
+        missing = []
+        for node in pipeline.nodes:
+            for output in node.outputs:
+                if not catalog.exists(output):
+                    missing.append(output)
+                    break
+
+        to_run = pipeline.only_nodes_with_outputs(*missing)
```
ravi-kumar-pilla: Hi @SajidAlamQB , The code and tests looks great. I am wondering if it would be helpful to log the nodes that are skipped which would help for debugging. The log level can be info or debug (not particular on this)

SajidAlamQB:
> #4635

Good catch! When we check `catalog.exists(output)`, these datasets always return `False`.

So we should probably do something like this:

`MemoryDataset` → Always run these nodes since they're meant to be recreated
Persistent datasets that don't exist → These are truly "missing" and need to be generated
Persistent datasets that exist → Skip these nodes

I'll update the `run_incremental` method to check for ephemeral datasets.

ravi-kumar-pilla:
> > #4635
> 
> Good catch! When we check `catalog.exists(output)`, these datasets always return `False`.
> 
> So we should probably do something like this:
> 
> `MemoryDataset` → Always run these nodes since they're meant to be recreated Persistent datasets that don't exist → These are truly "missing" and need to be generated Persistent datasets that exist → Skip these nodes
> 
> I'll update the `run_incremental` method to check for ephemeral datasets.

Why should we need another check for MemoryDataset ? Is this for differentiating dataset type only ? What would be the benefit with the extra check since catalog.exists will anyway return False and we run the node

SajidAlamQB:
> Why should we need another check for MemoryDataset ? Is this for differentiating dataset type only ? What would be the benefit with the extra check since catalog.exists will anyway return False and we run the node

@ravi-kumar-pilla that makes a lot of sense actually. I was overthinking it. @ankatiyar  when you tested with spaceflights and saw `split_data_node` running, was that the behaviour you expected?

kedro/framework/session/session.py
```diff
@@ -396,9 +399,14 @@ def run(  # noqa: PLR0913
             run_params=record_data, pipeline=filtered_pipeline, catalog=catalog
         )
         try:
-            run_result = runner.run(
-                filtered_pipeline, catalog, hook_manager, session_id
-            )
+            if only_missing_outputs:
+                run_result = runner.run_incremental(
```
DimedS: thanks @SajidAlamQB , why it's not a one more step in `filtered_pipeline`, but additional run function?

ankatiyar:
> > Why should we need another check for MemoryDataset ? Is this for differentiating dataset type only ? What would be the benefit with the extra check since catalog.exists will anyway return False and we run the node
> 
> @ravi-kumar-pilla that makes a lot of sense actually. I was overthinking it. @ankatiyar when you tested with spaceflights and saw `split_data_node` running, was that the behaviour you expected?

Actually what I meant was, when `preprocessed_companies` is missing, there is no need to run `split_data_node` just because `X_train`,`y_train` etc is missing/are not in the catalog (and not persisted as opposed to in the catalog but not persisted). It might make sense to run it when `train_node`'s output is missing which needs the memory datasets to also run. My question was mainly, should we distinguish when the outputs are missing and needed for the node you actually intend to run vs just running it because it generates memory datasets? Might not be worth doing this distinction and keeping things simple on our end which just the `catalog.exists()` check

ankatiyar:
accidentally found out you can edit other people's comments on github lol

ravi-kumar-pilla:
> Actually what I meant was, when `preprocessed_companies` is missing, there is no need to run `split_data_node` just because `X_train`,`y_train` etc is missing/are not in the catalog (and not persisted as opposed to in the catalog but not persisted). It might make sense to run it when `train_node`'s output is missing which needs the memory datasets to also run. My question was mainly, should we distinguish when the outputs are missing and needed for the node you actually intend to run vs just running it because it generates memory datasets? Might not be worth doing this distinction and keeping things simple on our end which just the `catalog.exists()` check

If we need to distinguish the output type (persistence vs non-peristence and missing), it makes sense to do the check. As you said it might not be worth to distinguish atleast for this use case

SajidAlamQB:
>  My question was mainly, should we distinguish when the outputs are missing and needed for the node you actually intend to run vs just running it because it generates memory datasets? Might not be worth doing this distinction and keeping things simple on our end which just the `catalog.exists()` check.

I think the simple `catalog.exists(`) approach is probably the right call. We can always enhance it later if users find it's running too many unnecessary nodes.

merelcht:
Why do you need `run_incremental()`? Can't you use the existing `run_only_missing()` method?

SajidAlamQB:
> Why do you need `run_incremental()`? Can't you use the existing `run_only_missing()` method?

I think @DimedS's suggestion to move all the filtering to the session layer, so we don't need any new runner methods at all might be better. The runner interface can stay minimal and clean.

kedro/framework/session/session.py
ankatiyar: Pipeline doesn't have access to catalog so it can't check right?

ravi-kumar-pilla:
> Why do you need `run_incremental()`? Can't you use the existing `run_only_missing()` method?

Hi @merelcht , As discussed on the spike work, `run_only_missing` methods checks for the entire pipeline outputs but not at each node level. So we need a different method which loops through each node.

I do like @DimedS suggestion on filtering the pipeline even before we execute the run.

kedro/framework/session/session.py
```diff
@@ -361,6 +363,49 @@ def run(  # noqa: PLR0913
             node_namespaces=namespaces,
         )
 
+        # Store original node count for logging
```
ravi-kumar-pilla: Can we have the entire filtering thing in a separate method ?

ravi-kumar-pilla: It would be nice to have a method (similar to filter or may be modify filter) which can give you a pipeline which has only missing outputs. wdyt @SajidAlamQB @DimedS

SajidAlamQB: Sounds like it would fit nicely with this pattern. We could add a method to the Pipeline class, but pipeline would need the catalog, so we'd need be sure there aren't any circular imports.

ravi-kumar-pilla: make sense of not importing catalog there. lets leave it here and see what others think. Thank you

merelcht:
I'm not sure I'm a fan of adding this pipeline filtering logic in the session. It's mixing responsibilities of the classes. 

I get the issue with `run_only_missing` now, but since that's not actually used in the code, maybe we can just update that to the behaviour we need. It was added way back in 0.14 and the only mention in the docs (https://docs.kedro.org/en/stable/nodes_and_pipelines/slice_a_pipeline.html) was added in 2022. Users can only access that by directly creating a runner and calling the method on it, so it's not likely this is used a lot.

DimedS:
> I'm not sure I'm a fan of adding this pipeline filtering logic in the session. It's mixing responsibilities of the classes.

I agree with Merel that the session is probably not the right place to handle pipeline filtering logic - it makes more sense to move that logic to the runner. However, I don’t quite understand why it’s implemented as a separate method, `runner.run_only_missing()`, rather than as an option within the existing `runner.run()` method that performs additional filtering at the start.

SajidAlamQB:
> `A -> MemoryDataset -> B -> PersistentDataset ` and I run it with the `--only-missing-outputs` flag, am I right in thinking that node **A** will still be executed to produce a `MemoryDataset`, even though node **B** (which has a persistent output) will be skipped?

Thanks @DimedS good catch! You're right the current behavior would wastefully run Node A to produce MemoryDataset that won't be consumed by the skipped Node B.

I will update it to check `dataset._EPHEMERAL` and only include non-ephemeral datasets in the missing outputs check.

DimedS:
> > `A -> MemoryDataset -> B -> PersistentDataset ` and I run it with the `--only-missing-outputs` flag, am I right in thinking that node **A** will still be executed to produce a `MemoryDataset`, even though node **B** (which has a persistent output) will be skipped?
> 
> Thanks @DimedS good catch! You're right the current behavior would wastefully run Node A to produce MemoryDataset that won't be consumed by the skipped Node B.
> 
> I will update it to check `dataset._EPHEMERAL` and only include non-ephemeral datasets in the missing outputs check.

Sounds great, but it can be an opposite case, like:

`A -> MemoryDataset -> B -> PersistentDataset which is not currently exists`, 

in that case we need to execute node A, otherwise node B cannot be executed, right? Am I right that current code is not properly handle that case, it will execute only B?

ravi-kumar-pilla:
> Thanks, @SajidAlamQB - great job! The structure of the code looks good to me.
> 
> I have a question about the algorithm: If I have a pipeline like this:
> 
> `A -> MemoryDataset -> B -> PersistentDataset ` and I run it with the `--only-missing-outputs` flag, am I right in thinking that node **A** will still be executed to produce a `MemoryDataset`, even though node **B** (which has a persistent output) will be skipped?
> 
> If so, this behaviour seems questionable - executing node A just to generate an in-memory dataset that won't be used doesn't make much sense. Should we consider changing this behaviour?

Great question ! If we want to handle that, we might need an extra check on deciding whether to skip a node execution - Skip if (output of a node is persistent) or (output is not persistent  and nodes with input as current node's output have a persistent output). I am sure it will add complexity but I think it is worth investigating. Thank you @DimedS

rashidakanchwala:
Yes, I agree. I believe @ankatiyar’s initial point was along the same lines -  if the task is only creating a MemoryDataset that backs a PersistentDataset, it ideally shouldn’t run if the persistent dataset already exists, but it definitely should if it doesn’t. If we can add logic to handle this, that would be great!

SajidAlamQB:
Hi team the `--only-missing-outputs` feature is ready for review and meets all the changes we've discussed.

The core implementation:

- Skips nodes whose persistent outputs already exist, reducing unnecessary computation.
- Handles upstream dependencies correctly, ensures required nodes still run when needed.
- Avoids wasteful executions, implements the optimisation discussed by @DimedS and @ankatiyar.

Edge Cases:

For the scenario @DimedS raised: A -> MemoryDataset -> B -> PersistentDataset

- If PersistentDataset exists: Both nodes A and B are correctly skipped
- If PersistentDataset is missing: Both nodes A and B run as required

For @ankatiyar's concern about nodes creating MemoryDatasets that back PersistentDatasets:

- The implementation identifies when intermediate nodes are only producing outputs for skipped downstream nodes these "wasteful" nodes are properly removed from execution.

kedro/runner/runner.py
```diff
@@ -67,12 +67,14 @@ def __init__(
     def _logger(self) -> logging.Logger:
         return logging.getLogger(self.__module__)
 
-    def run(
+    def run(  # noqa: PLR0913
         self,
         pipeline: Pipeline,
         catalog: CatalogProtocol,
         hook_manager: PluginManager | None = None,
+        session_id: str | None = None,
```
ElenaKhaustova: Is that correct?

SajidAlamQB: good catch, that was left over from an earlier implementation. Cleaned up now.

kedro/runner/runner.py
```diff
@@ -155,6 +161,161 @@ def run(
...
+            )
+            return pipeline.filter(node_names=[])
+
+        # Now we need to ensure we include all upstream dependencies of nodes_to_run
+        # Get all upstream nodes required for nodes that need to run
+        required_nodes = set(nodes_to_run)
+        for node in nodes_to_run:
+            # Add all upstream dependencies
+            upstream_pipeline = pipeline.to_nodes(node.name)
+            required_nodes.update(upstream_pipeline.nodes)
```
DimedS: Thanks, @SajidAlamQB! I think the functionality is working as expected now. One concern I have is the algorithm’s complexity - due to the current implementation, it becomes quadratic. As I understand it, after the initial filtering, it performs an upstream traversal for each selected node, which results in looping over all nodes multiple times. This can be quite expensive for large pipelines. I believe it’s possible to reduce this to linear complexity, but we can consider optimising that in a follow-up.

DimedS (approved):
Thanks @SajidAlamQB for addressing review comments, have one more but we can address it later

rashidakanchwala:
hi i get this error. it works fine when i do `kedro run` but when i do `kedro run --only-missing-outputs`. I get this 



```
[06/16/25 15:28:00] INFO     Using '/Users/Rashida_Kanchwala/kedro-universe/kedro/kedro/framework/project/rich_logging.yml' as logging          __init__.py:270
                             configuration.                                                                                                                    
[06/16/25 15:28:01] INFO     Kedro project demo-project                                                                                          session.py:331
│ /Users/Rashida_Kanchwala/miniconda3/envs/ked19py311/bin/kedro:8 in <module>                      │
│ /Users/Rashida_Kanchwala/kedro-universe/kedro/kedro/framework/cli/cli.py:260 in main             │
│   257 │   """                                                                                    │
│   258 │   _init_plugins()                                                                        │
│   259 │   cli_collection = KedroCLI(project_path=find_kedro_project(Path.cwd()) or Path.cwd())   │
│ ❱ 260 │   cli_collection()                                                                       │
│   261                                                                                            │
│ /Users/Rashida_Kanchwala/miniconda3/envs/ked19py311/lib/python3.11/site-packages/click/core.py:1 │
│ 161 in __call__                                                                                  │
│ /Users/Rashida_Kanchwala/kedro-universe/kedro/kedro/framework/cli/cli.py:162 in main             │
│   159 │   │   )                                                                                  │
│   160 │   │                                                                                      │
│   161 │   │   try:                                                                               │
│ ❱ 162 │   │   │   super().main(                                                                  │
│   163 │   │   │   │   args=args,                                                                 │
│   164 │   │   │   │   prog_name=prog_name,                                                       │
│   165 │   │   │   │   complete_var=complete_var,                                                 │
│ /Users/Rashida_Kanchwala/miniconda3/envs/ked19py311/lib/python3.11/site-packages/click/core.py:1 │
│ 082 in main                                                                                      │
│ /Users/Rashida_Kanchwala/miniconda3/envs/ked19py311/lib/python3.11/site-packages/click/core.py:1 │
│ 697 in invoke                                                                                    │
│ /Users/Rashida_Kanchwala/miniconda3/envs/ked19py311/lib/python3.11/site-packages/click/core.py:1 │
│ 443 in invoke                                                                                    │
│ /Users/Rashida_Kanchwala/miniconda3/envs/ked19py311/lib/python3.11/site-packages/click/core.py:7 │
│ 88 in invoke                                                                                     │
│ /Users/Rashida_Kanchwala/kedro-universe/kedro/kedro/framework/cli/project.py:244 in run          │
│   241 │   with KedroSession.create(                                                              │
│   242 │   │   env=env, conf_source=conf_source, runtime_params=params                            │
│   243 │   ) as session:                                                                          │
│ ❱ 244 │   │   return session.run(                                                                │
│   245 │   │   │   tags=tuple_tags,                                                               │
│   246 │   │   │   runner=runner_obj(is_async=is_async),                                          │
│   247 │   │   │   node_names=tuple_node_names,                                                   │
│ /Users/Rashida_Kanchwala/kedro-universe/kedro/kedro/framework/session/session.py:402 in run      │
│   399 │   │   │   run_params=record_data, pipeline=filtered_pipeline, catalog=catalog            │
│   400 │   │   )                                                                                  │
│   401 │   │   try:                                                                               │
│ ❱ 402 │   │   │   run_result = runner.run(                                                       │
│   403 │   │   │   │   filtered_pipeline,                                                         │
│   404 │   │   │   │   catalog,                                                                   │
│   405 │   │   │   │   hook_manager,                                                              │
│ /Users/Rashida_Kanchwala/kedro-universe/kedro/kedro/runner/runner.py:99 in run                   │
│    96 │   │   """                                                                                │
│    97 │   │   # Apply missing outputs filtering if requested                                     │
│    98 │   │   if only_missing_outputs:                                                           │
│ ❱  99 │   │   │   pipeline = self._filter_pipeline_for_missing_outputs(pipeline, catalog)        │
│   100 │   │                                                                                      │
│   101 │   │   # Check which datasets used in the pipeline are in the catalog or match            │
│   102 │   │   # a pattern in the catalog, not including extra dataset patterns                   │
│ /Users/Rashida_Kanchwala/kedro-universe/kedro/kedro/runner/runner.py:187 in                      │
│ _filter_pipeline_for_missing_outputs                                                             │
│   184 │   │   │                                                                                  │
│   185 │   │   │   for output in node.outputs:                                                    │
│   186 │   │   │   │   if output in catalog:                                                      │
│ ❱ 187 │   │   │   │   │   dataset = catalog._datasets[output]                                    │
│   188 │   │   │   │   │   is_ephemeral = getattr(dataset, "_EPHEMERAL", False)                   │
│   189 │   │   │   │   │                                                                          │
│   190 │   │   │   │   │   # Only check existence for persistent datasets                         │
KeyError: 'ingestion.int_typed_companies'
```

ElenaKhaustova:
Also agree with @DimedS that the current implementation has O(N^2) complexity, though we only need to consider each node only once when traversing the graph (pipeline).

ankatiyar:
I've also come across some of the issues flagged above:
- I get a `KeyError` that @rashidakanchwala mentioned 
- The order of the nodes gets messed up I think, and some nodes are randomly skipped when I run the whole project 

Screenshot 2025-06-16 at 15 36 31

- I think it's also a bit tricky with dataset factory related datasets, would this work? 
- Also concerned about the complexity of this algorithm 🤔

kedro/runner/runner.py
```diff
@@ -155,6 +161,161 @@ def run(
...
+        # First, identify all nodes that need to run based on their outputs
+        nodes_to_run = set()
+        skipped_nodes = set()
+
+        for node in pipeline.nodes:
+            node_has_missing_persistent_output = False
+
+            for output in node.outputs:
+                if output in catalog:
+                    dataset = catalog._datasets[output]
```
ankatiyar: This should be `catalog._datasets.get(output)` This throws `KeyError` otherwise

SajidAlamQB:
> If we traversed (dfs or bfs) the graph starting from nodes that have free inputs only there wouldn't be any corner cases and we would not need to add extra nodes and then remove them. While traversal we only need to check if the node's outputs are persisted, if so we can remove it otherwise we can't.

I like this suggestion, hopefully this should resolve the complexity concerns.

rashidakanchwala:
hi i tested this with kedro-viz demo project and it doesn't work as expected. I did `kedro run` which ran my whole pipeline and then i did `kedro run --only-missing-outputs` and it 

```
INFO     Skipping 7 nodes with existing persistent outputs: feature_engineering.create_feature_importance__173f2884,          runner.py:316
ingestion.apply_types_to_shuttles, reporting.create_feature_importance_plot__e7a53e99,                                            
 reporting.create_matplotlib_chart__15c38707, reporting.get_top_shuttles_data__3cb3499e,                                           
reporting.make_cancel_policy_bar_chart__7c6d9304, reporting.make_price_histogram__e0da8d32                                        
INFO     Running 14 out of 21 nodes (skipped 7 nodes) 
```
ideally it should skip all of them right? because all have outputs and I haven't removed any.

SajidAlamQB:
> ideally it should skip all of them right? because all have outputs and I haven't removed any.

It should be fixed now.

kedro/runner/runner.py
```diff
@@ -92,14 +94,24 @@ def run(
...
+
         # Check which datasets used in the pipeline are in the catalog or match
         # a pattern in the catalog, not including extra dataset patterns
         # Run a warm-up to materialize all datasets in the catalog before run
         warmed_up_ds = []
         for ds in pipeline.datasets():
-            if ds in catalog:
-                warmed_up_ds.append(ds)
-                _ = catalog.get(ds)
+            try:
```
ElenaKhaustova: Why do we need this check here and below?

SajidAlamQB: Probably not needed anymore I will remove the blocks.

SajidAlamQB:
> ideally it should skip all of them right? because all have outputs and I haven't removed any.

This was specific issue with the custom dataset in demo-project, `ImageDataset`, was missing an `_exist` method which the new runner logic relied on. Adding it in should fix and skip all.

kedro/runner/runner.py
```diff
@@ -155,6 +161,222 @@ def run(
...
+                if input_name in node_by_output:
+                    add_node_with_dependencies(node_by_output[input_name])
+
+        # Start from nodes with missing outputs and add all their dependencies
+        for node in nodes_with_missing_outputs:
+            add_node_with_dependencies(node)
+
+        return required_nodes
+
+    def _log_pipeline_filtering(
```
ElenaKhaustova: Is it used anywhere?

kedro/runner/runner.py
```diff
@@ -155,6 +161,222 @@ def run(
...
+        try:
+            if output in catalog:
+                return not catalog.exists(output)
+            # Not in catalog and no factory - treat as missing
+            return True
+        except Exception:
+            # If we can't determine, assume it's missing to be safe
+            return True
+
+    def _find_required_nodes(
```
ElenaKhaustova: Is it used anywhere?

kedro/runner/runner.py
```diff
@@ -155,6 +161,222 @@ def run(
...
+            try:
+                return not catalog.exists(output)
+            except Exception as e:
+                self._logger.warning(
+                    f"Could not check existence of {output}: {e}. Assuming it's missing."
+                )
+                return True
+
+        # Dataset not explicitly in catalog - check factory patterns
+        try:
```
ElenaKhaustova: Why is this, and below is wrapped with `try except`?

kedro/runner/runner.py
```diff
@@ -155,6 +161,222 @@ def run(
...
+    def _filter_pipeline_for_missing_outputs(
+        self, pipeline: Pipeline, catalog: CatalogProtocol
+    ) -> Pipeline:
+        """Filter pipeline to only include nodes needed to produce missing persistent outputs."""
+        original_node_count = len(pipeline.nodes)
+
+        # First identify which persistent outputs are missing
+        missing_persistent_outputs = set()
+
+        for node in pipeline.nodes:
```
ElenaKhaustova, line 173: I'm still a bit concerned about this implementation. I suggest we do bfs, where we add a node to the queue only if its outputs are persisted (starting from the nodes with free inputs only) and finish the traversal when the queue is empty. Those added to the queue can be removed. The complexity will be O(nodes_cnt).

Or it should be some other traversal.

Alternatively, could you please explain your solution in short?

kedro/runner/runner.py
SajidAlamQB: Sorry leftover code, its been through many iterations.

kedro/runner/runner.py
SajidAlamQB: Hi @ElenaKhaustova, thanks for all the feedback.

Here’s the short explanation of the current approach:

Current solution works in two simple steps:

1. First we do one quick pass over all nodes to find the list of all persistent datasets that are missing. This gives us our list of target outputs.

2. Secondly we then pass this list of missing outputs to the Kedro method `pipeline.to_outputs()` it works backwards from the goals and returns the smallest possible sub-pipeline needed to create them.

imo this is better than a manual forward traversal (like BFS) because  as far as I understand the `to_outputs` automatically and correctly handles complex dependency cases, especially skipping "wasteful" intermediate nodes that only lead to outputs that already exist.

kedro/runner/runner.py
```diff
@@ -155,6 +161,122 @@ def run(
...
+            # Not in catalog and no factory - will be MemoryDataset (ephemeral)
+            return False
+
+        # Check if it would be ephemeral
+        ds = catalog.get(output)
+        is_ephemeral = ds is not None and hasattr(ds, "_EPHEMERAL") and ds._EPHEMERAL
+
+        return False if is_ephemeral else not catalog.exists(output)
+
+    def _is_output_missing(self, output: str, catalog: CatalogProtocol) -> bool:
```
ElenaKhaustova: Should this also be removed?

kedro/runner/runner.py
ElenaKhaustova: Thanks for explaining.

If that's how `pipeline.to_outputs()` works, let's stick to it.

kedro/runner/runner.py
SajidAlamQB: I was thinking about this, the new implementation pretty much supersedes the old `run_only_missing` method. 

But this will be a breaking change and we would need to be clearly state it. If we are okay with that we can remove it now. 

Users who use this method won't have time to migrate their code away from the old method before it is removed in a future release since we didn't mark as deprecated though. wdyt?

kedro/runner/runner.py
```diff
@@ -155,6 +161,122 @@ def run(
...
+        """Check if a specific output is missing.
+
+        Args:
+            output: The output dataset name.
+            catalog: The data catalogue to check.
+
+        Returns:
+            True if the output is missing or couldn't be checked.
+        """
+        dataset = catalog._datasets.get(output)
```
ElenaKhaustova: You're using some catalog deprecated API here, which will have to be rewritten when merging all together.

Could you change the target from `develop` to `feature-1.0.0` so you can test all the changes together?

noklam:
The PR is probably lagging behind so the diff is super big and I am not sure which file to review yet, can you give me some pointers?

> 
> How it works
> When --only-missing-outputs is specified:
> 
> The runner checks each node's outputs using catalog.exists()
> Nodes with all outputs already persisted are skipped
> Only nodes with at least one missing output are executed

Just looking at the PR description, I have questions about whether this is a desired behavior. From my point of view this feature will be useful in a few cases:
- Debug purpose
- Recover from a partial fail run, i.e. pipeline stopped at somewhere and I know where to restart
- Development, i.e. I update logic in node 3 so I expected anything run after that need to be re-run.
- Efficient run to produce the desired datasets

When mixing this with versioning/real use case, does checking `.exists()` sufficiently determined whether or not a node need to be re-run? The output can be produced from previous run which is invalid/stale, how would a user specify that?

SajidAlamQB:
> The PR is probably lagging behind so the diff is super big and I am not sure which file to review yet, can you give me some pointers?

Thanks @noklam to make the review easier, the core logic is concentrated in these files:

- `kedro/framework/cli/project.py`: This is where the `--only-missing-outputs` CLI flag is defined.

- `kedro/framework/session/session.py`: The run method here was updated to accept the `only_missing_outputs` boolean and pass it down to the runner.

- `kedro/runner/runner.py`: Main changes are here
       - The `run() `method has the main if `only_missing_outputs`: switch.
       -The new `_filter_pipeline_for_missing_outputs()` method contains the logic.
       - We are using `pipeline.to_outputs()` to handle the dependency resolution automatically.

kedro/runner/runner.py
SajidAlamQB: The new API can't distinguish between an undefined dataset and an ephemeral, in-memory dataset that hasn't been created yet.

Using `dataset = catalog._datasets.get(output)` still works and handles the three states of a dataset:

- Found in `_datasets`.
- Found via in catalog. 
- Not found by in catalog. (ephemeral)

kedro/framework/cli/project.py
```diff
@@ -207,6 +209,11 @@ def package(metadata: ProjectMetadata) -> None:
     help=PARAMS_ARG_HELP,
     callback=_split_params,
 )
+@click.option(
+    "--only-missing-outputs",
+    is_flag=True,
```
astrojuanlu, line 214: Wait, do we want to make this the default?

SajidAlamQB: I believe making it the default would be a dangerous breaking change as it would deviate significantly from what users have been expecting. We should keep it as an opt-in flag (`--only-missing-outputs`), so its a deliberate choice for the users.

kedro/runner/runner.py
SajidAlamQB: I created another PR that targets the `feature-1.0.0` branch: https://github.com/kedro-org/kedro/pull/4848.

kedro/framework/cli/project.py
astrojuanlu: Sorry, I read `is_flag=True` and thought it said `default=True` or something 🤦🏼 and I also saw lots of tests explicitly marking `only_missing_outputs=False` which confused me

SajidAlamQB:
These are the scenarios I've tested with the spaceflights tutorial:

## Testing with Spaceflights

**Scenario: All Outputs Exist**

- Run `kedro run` once to generate everything.
- Run `kedro run --only-missing-outputs`.
- Expected: All nodes should be skipped. Timestamps on files like data/03_primary/model_input_table.parquet, data/06_models/regressor.pickle, and files in data/08_reporting/ should not change.

- Run `kedro run` once.
- Delete data/03_primary/model_input_table.parquet.
- Run `kedro run --only-missing-outputs`.
- Expected:
`create_model_input_table_node` must run.
`preprocess_companies_node` and `preprocess_shuttles_node` must run because their outputs are inputs to `create_model_input_table_node`.

- Temporarily edit `conf/base/catalog.yml` for regressor: change type: `pickle.PickleDataset` to type: `pandas.CSVDataset`.
- Delete data/06_models/regressor.pickle if it exists.
- Run `kedro run --only-missing-outputs`.
- Expected: The run should fail.

**Scenario: Interaction with `--to-nodes `**

- Run `kedro run once`.
- Run `kedro run --to-nodes split_data_node --only-missing-outputs`.
- Expected:
The `--only-missing-outputs` flag operates on the pipeline after other filters (like `--to-nodes`) have been applied.
It then looks for persistent outputs of that specific filtered pipeline that are missing.
Since all persistent outputs of the `--to-nodes split_data_node` pipeline already existed, nothing needed to be run.

Scenario: Running a specific pipeline

- Run `kedro run once`.
- Delete data/06_models/regressor.pickle.
- Run `kedro run --pipeline data_science --only-missing-outputs`.
- Expected:
`train_model_node` must run.
`split_data_node` must run (ephemeral outputs).

tests/runner/test_sequential_runner.py
```diff
@@ -272,6 +272,402 @@ def test_confirms(self, mocker, test_pipeline, is_async):
...
+        # Mock exists for persistent dataset
+        catalog.exists = mocker.Mock(return_value=False)
+
+        runner = SequentialRunner()
+        runner.run(test_pipeline, catalog, only_missing_outputs=True)
+
+        # Should run because persistent output is missing
+        assert "Running all 1 nodes" in caplog.text
+
+    def test_is_output_missing_various_conditions(self, mocker):
```
merelcht: This test is very hard to read. Can't you break each of the separate test cases where there's a comment now into a test?

tests/runner/test_sequential_runner.py
```diff
@@ -272,6 +272,402 @@ def test_confirms(self, mocker, test_pipeline, is_async):
...
+        catalog["C"] = MemoryDataset()
+        catalog["D"] = LambdaDataset(load=lambda: None, save=lambda x: None)
+
+        runner = SequentialRunner()
+        runner.run(test_pipeline, catalog, only_missing_outputs=True)
+
+        # All 4 nodes should run since D is missing
+        assert "Running all 4 nodes" in caplog.text
+
+    def test_is_persistent_and_missing(self, mocker):
```
merelcht: Same here, ideally a test should only have 1 maybe 2 asserts, but not mocking logic followed by an assert each time.

kedro/runner/runner.py
SajidAlamQB: Removed.

ElenaKhaustova:
@SajidAlamQB should we close it in favour of https://github.com/kedro-org/kedro/pull/4848?

SajidAlamQB:
Closing this PR in favour of: https://github.com/kedro-org/kedro/pull/4848

Issue #3080, opened 2023-09-25
Validate the `.` dot notation in Kedro pipeline


noklam:
## Description

https://github.com/kedro-org/kedro-viz/issues/1522

User are confused with disconnected graph in kedro-viz
https://docs.kedro.org/en/stable/tutorial/add_another_pipeline.html

## Context

Internally `.` is reserved for namespace, but Kedro Framework never block this/validate this eagerly. So user can create something that looks like a namespace pipeline but isn't.

User should use the `namespace` argument instead of using `inputs=namespace.my_input`.

https://github.com/kedro-org/kedro/blob/3d41425c39ff061809f2febb87994d787f03bf4d/kedro/pipeline/modular_pipeline.py#L225-L226

Internally, `dot` is used for namespace. Noted that even `Node` have `_namespace` attribute but it is actually more about which pipeline it is belong to. It's ambiguous what's the namespace of a node because the input and output can come from different namespace.
## Possible Implementation

Validate this behavior and block it. We probably cannot validate the node because it is a valid internal representation. We can however validate the argument in `node` constructor if it is desired.

## Possible Alternatives

Validate this behavior and raise warning.

datajoely:
There is a wider point about how we internally store namespaces in the catalog with `__` and it's weird!

MehdiNV:
Leaving notes for the future:
Some validation is completed, but like Nok notes above, may not be usable as we prefix. Regardless, may be of interest in the future - check the highlighted parts below for how Kedro already attempts some validation:
![image](https://github.com/kedro-org/kedro/assets/35583819/1350e358-e1a1-451f-9df4-82bd225e8ac9)

astrojuanlu:
For now, let's add a `WARNING` to the logs when the user uses a `.` in an input or output specifier.

rashidakanchwala:
It will be on a develop branch. The decision has been made to not allow '.' notation in node names

merelcht:
Completed in https://github.com/kedro-org/kedro/pull/4687

ankatiyar:
Actually upon further digging: This only fails in case when node and pipeline level namespace are both applied
```
def create_pipeline(**kwargs) -> Pipeline:
    return pipeline(
        [
            node(
                func=preprocess_companies,
                inputs="companies",
                outputs="preprocessed_companies",
                name="preprocess_companies_node",
                namespace="preprocessing",
            ),
            node(
                func=preprocess_shuttles,
                inputs="shuttles",
[...]

PR #4797, opened 2025-06-02
Limit input/output name validation check to top level namespace

ankatiyar:
## Description

Related to https://github.com/kedro-org/kedro/issues/3080#issuecomment-2890850794

## Development notes

The error about `.` dots not being allowed in input/output names was falsely showing up when the node level namespace is assigned. This is because we were doing a `startswith` comparison with `node.namespace` which includes both `pipeline_level_namespace.node_level_namespace` but the inputs and outputs only get assigned the `node_level_namespace.dataset_name`. This PR changes the check to make the comparison only with the first level namespace.

## Checklist


ankatiyar:
Q for reviewers: Should this check be happening when `Pipeline` is created instead? At that point the namespace is not added and maybe that is the right place to check if the inputs/outputs contain dots?

tests/pipeline/test_node.py
```diff
@@ -541,10 +541,11 @@ def test_valid_multi_level_namespace(self):
...
-                namespace="namespace.othernamespace",
-            )
+        n = node(
+            func=self.dummy_function,
+            inputs="namespace.input_dataset",
+            outputs="namespace.output_dataset",
+            namespace="namespace.other",
+        )
+        assert n.inputs == ["namespace.input_dataset"]
+        assert n.outputs == ["namespace.output_dataset"]
```
DimedS, line 561: Thanks, @ankatiyar ! Could you please comment how this test is currently related to validation check above?

ankatiyar: When the top level namespace matches, it should not error. Previously it would raise an error

DimedS (changes requested):
Thanks for the proposal, @ankatiyar!

This validation is a bit unclear to me. I think it would be helpful to improve the docstring to better explain the purpose of the check, so that it's more understandable for users. Also, the current error message is quite vague — as a user, I’m not sure what action I’m supposed to take when I encounter it.

More importantly, I’m struggling to understand the rationale behind this validation. Why is it a problem if my node takes input from a dataset like `A.input1` that belongs to a different pipeline (with a different namespace), especially if my current node is also in a separate namespaced pipeline? Since they are distinct pipelines with different structures, I don’t see why such input should be disallowed.

lrcouto:
> Q for reviewers: Should this check be happening when `Pipeline` is created instead? At that point the namespace is not added and maybe that is the right place to check if the inputs/outputs contain dots?

In my head it makes sense to do it in the node, since it's a feature of the node. 🤔

tests/pipeline/test_node.py
merelcht: The test name and docstring should also be updated to reflect the expected behaviour. And maybe it also makes sense to add a test that does check error raising when the top-level is wrong. Or do we have that already?

merelcht:
@DimedS the problem is that sometimes users use the dot notation as part of the name of input/output, not meaning for it to represent a `namespace` see https://github.com/kedro-org/kedro/issues/3080

DimedS:
> @DimedS the problem is that sometimes users use the dot notation as part of the name of input/output, not meaning for it to represent a `namespace` see #3080

I see. Perhaps instead of implementing the current PR, we should consider changing this check to raise a `warning` rather than an `error`. The PR addresses some real cases involving namespaces, but it doesn't seem to cover others - such as when the input comes from a previously executed, differently namespaced pipeline.

ankatiyar:
I've updated the docstring and added a test

tests/pipeline/test_node.py
```diff
@@ -541,10 +541,23 @@ def test_valid_multi_level_namespace(self):
...
+
+    def test_mismatched_namespace(self):
+        """Test that mismatched namespaces in inputs and outputs raise a ValueError."""
+        with pytest.raises(
+            ValueError, match="Invalid dataset name 'namespace.input_dataset'"
```
merelcht: Shouldn't this throw for the `other_namespace` ?

ankatiyar: omg yes sorry fixed this

DimedS:
> > @DimedS the problem is that sometimes users use the dot notation as part of the name of input/output, not meaning for it to represent a `namespace` see #3080
> 
> I see. Perhaps instead of implementing the current PR, we should consider changing this check to raise a `warning` rather than an `error`. The PR addresses some real cases involving namespaces, but it doesn't seem to cover others - such as when the input comes from a previously executed, differently namespaced pipeline.

thanks, @ankatiyar , could you please comment on that? I propose to change from `error` to `warning` because this check sometimes can be triggered in normal working use cases.

ankatiyar:
> thanks, @ankatiyar , could you please comment on that? I propose to change from `error` to `warning` because this check sometimes can be triggered in normal working use cases.

@DimedS I see that happening too, the error would be breaking so if we're sticking to error we can relax it later. We wouldn't be able to make it an error post 1.0. Happy to change it to a warning if that makes more sense? @merelcht @lrcouto

tests/pipeline/test_node.py
rashidakanchwala: I am still a bit confused. the original test 

```
node(
                func=self.dummy_function,
                inputs="namespace.subnamespace.input_dataset",
                outputs="namespace.subnamespace.output_dataset",
                namespace="namespace.othernamespace",
            )
```

would now pass with the implementation. but this feels incorrect.

tests/pipeline/test_node.py
```diff
@@ -539,12 +539,25 @@ def test_valid_multi_level_namespace(self):
         assert n.inputs == ["namespace.subnamespace.input_dataset"]
         assert n.outputs == ["namespace.subnamespace.output_dataset"]
 
-    def test_invalid_multi_level_namespace(self):
-        """Test that mismatched multi-level namespaces raise a ValueError."""
-        with pytest.raises(ValueError, match="Invalid dataset name"):
+    def test_only_top_level_namespace_match(self):
```
rashidakanchwala: maybe the test name here can be  `test_dataset_namespace_matches_pipeline_namespace(self)`:

rashidakanchwala (approved):
LGTM. I think the names are a bit confusing, but after @ankatiyar explained that:
- dataset names are `pipeline_namespace.dataset_name`, and
- node namespaces are usually `pipeline_namespace.node_namespace`, and that's what it used to be matching against

Maybe this can be made clearer in the docstrings and test names. I'm approving this for now as the implementation makes sense.

tests/pipeline/test_node.py
ankatiyar:
```suggestion
    def test_dataset_namespace_matches_pipeline_namespace(self):
```

DimedS (approved):
thanks, great job, @ankatiyar !

Issue #4176, opened 2024-09-18
ParalleRunner hangs on Linux Server


Dekermanjian:
## Description
I have a pipeline that I would like to run using the ParallelRunner. When I run this pipeline on my local windows machine it works just fine. However, when I try running the exact same pipeline on a Linux server (Rocky Linux) it will just hang at the loading datasets stage.

* Kedro version used (`pip show kedro` or `kedro -V`): 0.19.8
* Python version used (`python -V`): 3.11.8
* Operating system and version: Rocky Linux version 8.10

noklam:
Can you provide some more context, if possible to share a simplified version of repository that we can try to reproduce locally.

Dekermanjian:
Yeah, of course. Let me try to put together something simple that will hang on the server and then I'll share the repo with you.

Dekermanjian:
@noklam Okay, I figured out why it is not working. I just don't understand why it doesn't work on Linux but it does on Windows. Here is a simple example: https://github.com/Dekermanjian/test-parallel-runner

The reason it is not working on the linux server is because I am loading a parquet file in my settings.py file. When I load that file in the simple example the ParallelRunner will hang at the loading dataset stage. If you comment that line out (line 6) then it will work. You can generate the data by running the notebook I created.

Dekermanjian:
Hey @noklam, were you able to reproduce the issue? I am wondering if this is just happening on my end based on how the Linux server I am using is set up.

ankatiyar:
Hey @Dekermanjian, I was trying to reproduce this but I lack the input data, would you be able to commit it to your test project repo or share it with us? (assuming that it is sanitised and shareable)

Dekermanjian:
@ankatiyar, yes, I can do that now. I am sorry, I forgot to adjust the .gitignore before pushing. Okay, I have now pushed the data to the repo. 

A couple of things that I noticed while further testing. I noticed that in `settings.py` if I read the parquet file with pandas then it does not hang. I can also read it with pandas and then turn it to a polars data frame and it will also not hang. So it seems the problem is reading the file using polars. This is most likely a polars problem and not a kedro problem, but it would be good for someone else to reproduce this so that others that may experience the same problem can refer to this issue for a quick fix.

Dekermanjian:
[...]

PR #4801, opened 2025-06-03
Allow option for user to select multiprocessing method on ParallelRunner

lrcouto:
## Description

In the issue https://github.com/kedro-org/kedro/issues/4176, one of our users reported pipeline execution "hanging" on Linux environments when ParallelRunner was used. This was because on Linux, Python defaults to fork, which can conflict with libraries like Polars that spawn internal threads. Windows and MacOS default to spawn, which avoids the issue.

This PR allows for the user to choose to use either "fork" or "spawn" through a `KEDRO_MP_CONTEXT` environment variable. If nothing is set, the system default will be used.

### Reproducing the error

To reproduce the error, I've converted one of the default datasets in a spaceflights project to a parquet file, and loaded it in the projects `settings.py` as such, to make it spawn internal threads.

```py
import polars as pl

_dummy_df = pl.read_parquet("data/01_raw/shuttles.parquet")

_dummy_shape = _dummy_df.shape
```

And I've also modified one of the already existing nodes to load the data into Polars data structures, to guarantee that the threads are initialized during the pipeline run and ensure that the multiprocessing issue actually happens:

```py
def create_model_input_table(
    shuttles: pd.DataFrame, companies: pd.DataFrame, reviews: pd.DataFrame
) -> pd.DataFrame:
    """Combines all data to create a model input table using Polars.

    Args:
        shuttles: Preprocessed data for shuttles.
        companies: Preprocessed data for companies.
        reviews: Raw data for reviews.
    Returns:
        Model input table.
    """

    shuttles_pl = pl.from_pandas(shuttles)
    companies_pl = pl.from_pandas(companies)
    reviews_pl = pl.from_pandas(reviews)

    rated_shuttles = shuttles_pl.join(reviews_pl, left_on="id", right_on="shuttle_id")
    rated_shuttles = rated_shuttles.drop("id")
    model_input_table = rated_shuttles.join(companies_pl, left_on="company_id", right_on="id")
    model_input_table = model_input_table.drop_nulls()

    return model_input_table.to_pandas()
```

This will make the run hang when using ParallelRunner. Tested on Ubuntu 24.04 LTS.

`export KEDRO_MP_CONTEXT=spawn` should allow it to run normally.

## Checklist

### DIFF
--- a/tests/runner/test_parallel_runner.py
+++ b/tests/runner/test_parallel_runner.py
@@ -440,6 +440,8 @@ class TestSuggestResumeScenario:
 
 class TestMultiprocessingGetExecutorContextSelection:
     def test_get_executor_default(self, mocker):
+        if sys.platform == "win32" and sys.version_info < (3, 11):
+            pytest.skip("fork context is not available on Windows")
         mocker.patch.dict(os.environ, {}, clear=True)
         runner = ParallelRunner()
         executor = runner._get_executor(2)
@@ -459,6 +461,8 @@ class TestMultiprocessingGetExecutorContextSelection:
         assert executor._mp_context.get_start_method() == context
 
     def test_get_executor_invalid_context(self, mocker):
+        if sys.platform == "win32" and sys.version_info < (3, 11):
+            pytest.skip("fork context is not available on Windows")
         mocker.patch.dict(os.environ, {"KEDRO_MP_CONTEXT": "invalid"})
         runner = ParallelRunner()
         executor = runner._get_executor(2)


kedro/runner/parallel_runner.py
```diff
@@ -173,7 +175,12 @@ def _get_required_workers_count(self, pipeline: Pipeline) -> int:
         return min(required_processes, self._max_workers)
 
     def _get_executor(self, max_workers: int) -> Executor:
-        return ProcessPoolExecutor(max_workers=max_workers)
+        context = os.environ.get("KEDRO_MP_CONTEXT")
```
astrojuanlu, line 178: Any reason to not allow a setting in ParallelRunner rather than an env var?

lrcouto: I chose an env var because it was the "minimally invasive" approach to changing the code while still not requiring much extra effort from the user. But it could be done in other ways too. What did you imagine?

My logic choosing this was something like:
- It requires little change on the code on our side. Making it as a new parameter or something to be fetched from the `params` of a project would require changing multiple functions. Making it be fetched from the environment requires changing only one.
- Users who are not having issues don't have to do anything or change anything on their projects.
- Environment variables are quick and easy to set up and don't require specific knowlege of Kedro or the Kedro project structure.
- It's easy to set an environment variable even if you're running it in an automated setting, like in test setups that create and run the project for example.

The downside that I see is that the user would need to know that the variable exists and can be used, which I did add to the documentation but could be not that obvious to find.

kedro/runner/parallel_runner.py
```diff
@@ -173,7 +175,12 @@ def _get_required_workers_count(self, pipeline: Pipeline) -> int:
...
+        if context:
```
ElenaKhaustova:
```suggestion
        if context and context not in {"fork", "spawn"}:
```

ElenaKhaustova (approved):
Left minor comment, otherwise, LGTM!

Issue #4683, opened 2025-04-20
Missing Pipeline Inputs Bug in 0.19.11 and 0.19.12


lordsoffallen:
## Description
Kedro version > 0.19.10 has issues with showing missing pipeline inputs.

## Steps to Reproduce
1. Create a kedro node with wrong parameter name
2. Try to run kedro 

## Expected Result
Show the error like the earlier versions

## Your Environment

* Kedro version used (`pip show kedro` or `kedro -V`): 0.19.11 and 0.19.12
* Python version used (`python -V`): 3.11
* Operating system and version: Ubuntu

lordsoffallen:
Error is too long, so I am adding it here:

## Actual Result
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /home/ftopal/Projects/text2shots/.venv/bin/kedro:8 in <module>                                   │
│                                                                                                  │
│ /home/ftopal/Projects/text2shots/.venv/lib/python3.11/site-packages/kedro/framework/cli/cli.py:2 │
│ 63 in main                                                                                       │
│                                                                                                  │
│   260 │   cli_collection = KedroCLI(                                                             │
│   261 │   │   project_path=_find_kedro_project(Path.cwd()) or Path.cwd()                         │
│   262 │   )                                                                                      │
│ ❱ 263 │   cli_collection()                                                                       │
│   264                                                                                            │
│                                                                                                  │
│ /home/ftopal/Projects/text2shots/.venv/lib/python3.11/site-packages/click/core.py:1161 in        │
│ __call__                                                                                         │
│                                                                                                  │
│ /home/ftopal/Projects/text2shots/.venv/lib/python3.11/site-packages/kedro/framework/cli/cli.py:1 │
│ 63 in main                                                                                       │
│                                                                                                  │
│   160 │   │   )                                                                                  │
│   161 │   │                                                                                      │
│   162 │   │   try:                                                                               │
│ ❱ 163 │   │   │   super().main(                                                                  │
│   164 │   │   │   │   args=args,                                                                 │
│   165 │   │   │   │   prog_name=prog_name,                                                       │
│   166 │   │   │   │   complete_var=complete_var,                                                 │
│                                                                                                  │
│ /home/ftopal/Projects/text2shots/.venv/lib/python3.11/site-packages/click/core.py:1082 in main   │
│                                                                                                  │
[...]

PR #4819, opened 2025-06-10
Remove self from datasets init to prevent recursion

DimedS:
## Description
Solves #4683

### DIFF
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -355,6 +355,7 @@ class AbstractDataset(abc.ABC, Generic[_DI, _DO]):
             init_func(self, *args, **kwargs)
             # Capture and save the arguments passed to the original __init__
             self._init_args = getcallargs(init_func, self, *args, **kwargs)
+            self._init_args.pop("self", None)  # removed to prevent recursion
 
         # Replace the subclass's __init__ with the new_init
         # A hook for subclasses to capture initialization arguments and save them


merelcht (approved):
Amazing!! Thanks so much @DimedS, I tried it out and it works well with both catalogs ⭐

merelcht:
Can you add it to the release notes as bug fix for `0.19.14`?

DimedS:
> Can you add it to the release notes as bug fix for `0.19.14`?

ok, I will try to also add one more test to prevent this case

ravi-kumar-pilla (approved):
Thank you @DimedS .

rashidakanchwala:
thanks @DimedS , let's merge this so we can do a 0.19.14 release today/tomorrow.

kedro/io/core.py
```diff
@@ -355,6 +355,7 @@ def new_init(self, *args, **kwargs) -> None:  # type: ignore[no-untyped-def]
             init_func(self, *args, **kwargs)
             # Capture and save the arguments passed to the original __init__
             self._init_args = getcallargs(init_func, self, *args, **kwargs)
+            self._init_args.pop("self", None)  # removed to prevent recursion
```
ElenaKhaustova, line 357: We probably need to remove the following line above:

```python
self._init_args.pop("self", None)  # type: ignore[attr-defined]
```

DimedS: thanks @ElenaKhaustova, removed

PR #4824, opened 2025-06-14
[Minor / Code quality] Improve operations with `RICH_INSTALLED` constant

yury-fedotov:
## Description
I was reading through the `ipython` module of Kedro and realised that usage of `RICH_INSTALLED` constant can be slightly improved in terms of readability and type checking.


kedro/ipython/__init__.py
```diff
@@ -49,7 +49,7 @@
 
 FunctionParameters = MappingProxyType
 
-RICH_INSTALLED = True if importlib.util.find_spec("rich") is not None else False
+RICH_INSTALLED: Final = importlib.util.find_spec("rich") is not None
```
yury-fedotov, line 52: - Adding `Final` so that `mypy` protects against re-assignments of value.
- `True if condition else False` is always redundant in Python, and can be simplified at least to `bool(condition)`, but in this case even to `condition` itself.

kedro/ipython/__init__.py
```diff
@@ -287,7 +287,7 @@ def _create_cell_with_text(text: str) -> None:
 
 def _print_cells(cells: list[str]) -> None:
     for cell in cells:
-        if RICH_INSTALLED is True:
+        if RICH_INSTALLED:
```
yury-fedotov, line 290: `is True` is redundant: `RICH_INSTALLED` is of `bool` type and can be just used as `if condition:`.

ElenaKhaustova (approved):
Thank you, @yury-fedotov 👍

PR #4825, opened 2025-06-14
Annotate `copy_mode` in `MemoryDataset` and `CachedDataset` as `Literal` instead of `str`

yury-fedotov:
## Description

`MemoryDataset` and `CachedDataset` operate with `copy_mode` argument which clearly has 3 options:

- "deepcopy"
- "copy"
- "assign"

This argument is annotated as `str`, however an appropriate type would be `Literal`.

Advantages:

- Signature readability improved
- `mypy` protects from calling those methods providing something else to this argument


datajoely:
Love this

SajidAlamQB (approved):
@yury-fedotov thank you!

kedro/io/core.py
```diff
@@ -1004,6 +1005,18 @@ def exists(self, name: str) -> bool:
         """Checks whether registered dataset exists by calling its `exists()` method."""
         ...
 
+    @staticmethod
```
ElenaKhaustova: I think this method shouldn't be here
See https://github.com/kedro-org/kedro/blob/c9391d68dbdc86d953d12bc5f4b2f3936d756375/kedro/io/core.py#L943

kedro/io/core.py
```diff
@@ -1004,6 +1005,18 @@ def exists(self, name: str) -> bool:
...
+    def _validate_versions(
+        datasets: dict[str, AbstractDataset] | None,
+        load_versions: dict[str, str],
+        save_version: str | None,
+    ) -> tuple[dict[str, str], str | None]:
+        """Validates and synchronises dataset versions for loading and saving."""
+        ...
+
+
+TCopyMode = Literal["deepcopy", "copy", "assign"]
```
ElenaKhaustova: Can we please move it to the top?

ElenaKhaustova (changes requested):

ElenaKhaustova (approved):
Thank you @yury-fedotov and @SajidAlamQB!

yury-fedotov:
Thank you @SajidAlamQB for finalizing it 🔥

Issue #4771, opened 2025-05-24
Deduplicate utility functions in tests: `identity`, `biconcat`, etc.


yury-fedotov:
## Context

Looking through tests, I realized that some utility functions are defined multiple times:

<img width="814" alt="Image" src="https://github.com/user-attachments/assets/66a1a191-5e6e-4a73-a028-8f5f3cb9f640" />

<img width="822" alt="Image" src="https://github.com/user-attachments/assets/46989856-bf17-4b9e-8a78-e24032e6bc9e" />

And a few more.

## Proposal

It might be helpful to deduplicate them, and add typing, to improve developer experience while adding / maintaining tests.

## Possible implementation

Single `testing_utils` module, where all those are defined, and imported to all test files which need them.

---

Would that help the dev team? I can implement this change if you think it would.

Anshumank399:
Hello @merelcht @yury-fedotov 
First time contributor for the repo, would like to take this issue up and work on it.

merelcht:
Completed in #4826

PR #4826, opened 2025-06-14
Deduplicate functions in the tests folder

Anshumank399:
## Description
This PR is to resolve the improvements suggested in the #4771 issue.

## Development notes
Created a new file `testing_utils.py`, which entails all the functions that are generic and might be used multiple times thorough out the test folder.


tests/testing_utils.py
```diff
@@ -0,0 +1,77 @@
...
+
+def exception_fn(*args: Any) -> None:
+    """Raise a test exception."""
+    raise Exception("test exception")
+
+
+def return_none(arg: Any) -> None:
+    """Return None."""
+    arg = None
+    return arg
```
yury-fedotov: `arg` variable is in local scope within function's body, so can't this be simplified to:

```suggestion
    return None
```

?

tests/testing_utils.py
```diff
@@ -0,0 +1,77 @@
...
+    """Return all arguments as a tuple."""
+    return args
+
+
+def first_arg(*args: Any) -> Any:
+    """Return the first argument."""
+    return args[0]
+
+
+def exception_fn(*args: Any) -> None:
```
yury-fedotov: `None` is incorrect output type here, because this function would never return anything - it always raises an exception.

Python has a special type for that: [`NoReturn`](https://docs.python.org/3/library/typing.html#typing.Never), that needs to be imported from `typing` module.

```suggestion
def exception_fn(*args: Any) -> NoReturn:
```

tests/testing_utils.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def biconcat(input1: str, input2: str) -> str:
+    """Concatenate two strings."""
+    return input1 + input2  # pragma: no cover
+
+
+def triconcat(input1: str, input2: str, input3: str) -> str:
+    """Concatenate three strings."""
+    return input1 + input2 + input3  # pragma: no cover
```
yury-fedotov:
```suggestion
    return f"{input1}{input2}{input3}"  # pragma: no cover
```

And same goes for `biconcat`.

tests/testing_utils.py
```diff
@@ -0,0 +1,77 @@
+"""Common utility functions for testing."""
+
+from typing import Any
+
+import pandas as pd
+
+
+def identity(input1: str) -> str:
```
yury-fedotov: Signature is currently limited to `str`, which I think is too narrow. This function doesn't do anything string-specific. Returning an input itself for any input can work with any type.

Better typing in this case would be this:

```suggestion
def identity(input1: T) -> T:
```

Where `T` needs to be defined above as:

```
from typing import TypeVar

T = TypeVar("T")
```

tests/framework/context/test_context.py
```diff
@@ -156,9 +156,7 @@ def mock_settings(mocker):
     return mocker.patch("kedro.framework.project.settings", mocked_settings)
 
 
-@pytest.fixture
-def dummy_dataframe():
-    return pd.DataFrame({"col1": [1, 2], "col2": [4, 5], "col3": [5, 6]})
+dummy_dataframe = pytest.fixture(dummy_dataframe)
```
DimedS: why we are not creating this fixture inside of the testing_utils.py?

DimedS (commented):
Thanks you, @Anshumank399 , it looks mostly great! 
have one question

tests/testing_utils.py
```diff
@@ -0,0 +1,78 @@
+"""Common utility functions for testing."""
```
ElenaKhaustova: It looks like we already have https://github.com/kedro-org/kedro/blob/main/tests/test_utils.py file, so we don't need to create a separate one.

tests/testing_utils.py
```diff
@@ -0,0 +1,78 @@
...
+    """Concatenate three strings."""
+    return f"{input1}{input2}{input3}"  # pragma: no cover
+
+
+def constant_output() -> str:
+    """Return a constant output string."""
+    return "output"  # pragma: no cover
+
+
+def dummy_dataframe():
```
ElenaKhaustova: We can make it a fixture and move to https://github.com/kedro-org/kedro/blob/main/tests/conftest.py

tests/testing_utils.py
```diff
@@ -0,0 +1,78 @@
...
+    """Return a dummy pandas DataFrame for testing."""
+    return pd.DataFrame({"col1": [1, 2], "col2": [4, 5], "col3": [5, 6]})
+
+
+def dummy_dataframe_simple():
```
ElenaKhaustova: We can make it a fixture and move to https://github.com/kedro-org/kedro/blob/main/tests/conftest.py

ElenaKhaustova (commented):
Nice idea, @Anshumank399!

Made a few suggestion, the rest looks good 👍

Huongg:
Hi @Anshumank399, just checking in to see if you have capacity to address the remaining minor comments, and resolve the merge conflicts. We'd love to get this merged before our release later this week if possible.

tests/runner/conftest.py
```diff
@@ -6,45 +6,31 @@
...
-
-
-def sink(arg):
-    pass
-
-
-def fan_in(*args):
-    return args
+def source():
+    return "stuff"
```
merelcht: This one is also in the `test_utils`, so can be removed and imported instead.

PR #4847, opened 2025-06-17
Fix tests runners

DimedS:
## Description
Fix runners tests for 1.0 release


ElenaKhaustova (approved):
Tested locally, all passes!

PR #4848, opened 2025-06-17
Pipeline-based `--only-missing-outputs` into feature-1.0.0

SajidAlamQB:
## Description

This PR introduces a new CLI flag, `--only-missing-outputs`, to the kedro run command. When this flag is enabled, Kedro will skip the execution of nodes if all their persistent outputs already exist in the catalog.

This significantly speed up development and re-runs of pipelines, especially when only parts of the pipeline need to be updated or when resuming after a partial run.

For this initial MVP, the feature focuses on the existence of persistent datasets and does not involve hashing of data or code for change detection.

## How it works

**Important:** This MVP implementation uses a **pipeline-level approach** rather than node-by-node checking:

1. First, it identifies all **final persistent outputs** that are missing from the entire pipeline
2. Then, it builds the minimal subgraph needed to produce those missing outputs using `pipeline.to_outputs()`
3. Finally, it runs **all nodes in that subgraph**, even if some intermediate outputs already exist

### Example:
If you have a pipeline: `A → B → C` where `C` is missing but `B` exists:
- The feature will run both `A → B → C` to ensure consistency
- It does NOT check if `B` exists and skip just node `A`

More granular node-level checking approach can be found here, https://github.com/kedro-org/kedro/pull/4859. 

Change detection with hashing etc., is planned for post-1.0.0 see spike, https://github.com/kedro-org/kedro/issues/4635.

## Development notes

1. **CLI**: Added `--only-missing-outputs` flag to the `kedro run` command.
2. **Session**: Modified `KedroSession.run()` to accept the new parameter and use the appropriate runner method.
3. **Runner**: Updated to include `only_missing_outputs: bool = False`.

If `only_missing_outputs` is True, the runner first calls a new private method: 

`_filter_pipeline_for_missing_outputs(pipeline, catalog)`: This method returns a potentially modified (reduced) pipeline.

This new method is responsible for identifying which nodes need to run. It iterates through all nodes in the original pipeline to find all unique persistent outputs that are currently missing from the catalog, using the `_is_persistent_and_missing()` helper.

If no persistent outputs are missing, it returns an empty Pipeline([]), effectively skipping all nodes.

If there are missing persistent outputs, it uses `pipeline.to_outputs(*missing_persistent_outputs)` to construct a new pipeline. This should nherently handle the exclusion of "wasteful" nodes.

`_is_persistent_and_missing`:

This new helper method determines if a given dataset name corresponds to a persistent dataset that does not currently exist.

It first attempts to instantiate the dataset using `catalog.get(output)`.
It correctly identifies ephemeral datasets (e.g., those with `_EPHEMERAL = True`, like `MemoryDataset`) and considers them as "not needing to be checked for existence" for this feature's purpose.

For persistent datasets, it uses `catalog.exists(output)` to check for current data.
It handles datasets defined explicitly in the catalog, those created by dataset factories, and those implicitly becoming `MemoryDataset` if undefined.

## Testing with Spaceflights

**Scenario: All Outputs Exist**

- Run `kedro run` once to generate everything.
- Run `kedro run --only-missing-outputs`.
- Expected: All nodes should be skipped. Timestamps on files like data/03_primary/model_input_table.parquet, data/06_models/regressor.pickle, and files in data/08_reporting/ should not change.

- Run `kedro run` once.
- Delete data/03_primary/model_input_table.parquet.
- Run `kedro run --only-missing-outputs`.
- Expected:
`create_model_input_table_node` must run.
`preprocess_companies_node` and `preprocess_shuttles_node` must run because their outputs are inputs to `create_model_input_table_node`.

- Temporarily edit `conf/base/catalog.yml` for regressor: change type: `pickle.PickleDataset` to type: `pandas.CSVDataset`.
- Delete data/06_models/regressor.pickle if it exists.
- Run `kedro run --only-missing-outputs`.
- Expected: The run should fail.

**Scenario: Interaction with `--to-nodes `**

- Run `kedro run once`.
- Run `kedro run --to-nodes split_data_node --only-missing-outputs`.
- Expected:
The `--only-missing-outputs` flag operates on the pipeline after other filters (like `--to-nodes`) have been applied.
It then looks for persistent outputs of that specific filtered pipeline that are missing.
Since all persistent outputs of the `--to-nodes split_data_node` pipeline already existed, nothing needed to be run.

**Scenario: Running a specific pipeline**

- Run `kedro run once`.
- Delete data/06_models/regressor.pickle.
- Run `kedro run --pipeline data_science --only-missing-outputs`.
- Expected:
`train_model_node` must run.
`split_data_node` must run (ephemeral outputs).

## Checklist

### DIFF
--- a/tests/runner/test_only_missing_outputs.py
+++ b/tests/runner/test_only_missing_outputs.py
@@ -5,7 +5,6 @@ import pytest
 from kedro.io import (
     AbstractDataset,
     DataCatalog,
-    LambdaDataset,
     MemoryDataset,
     SharedMemoryDataCatalog,
 )
@@ -77,13 +76,7 @@ def create_persistent_dataset(request):
     """Factory fixture to create appropriate persistent dataset based on runner type"""
 
     def _create_dataset(exists_result=False):
-        runner_class = request.node.callspec.params.get("runner_class")
-        if runner_class == ParallelRunner:
-            # Use serializable dataset for ParallelRunner
-            return DummyDataset(exists_result=exists_result)
-        else:
-            # Use LambdaDataset for other runners
-            return LambdaDataset(load=lambda: None, save=lambda x: None)
+        return DummyDataset(exists_result=exists_result)
 
     return _create_dataset
 


tests/runner/test_sequential_runner.py
```diff
@@ -271,6 +271,432 @@ def test_confirms(self, mocker, test_pipeline, is_async):
         fake_dataset_instance.confirm.assert_called_once_with()
 
 
+class TestOnlyMissingOutputs:
```
ElenaKhaustova: Should these tests be run with all the runners?

kedro/runner/runner.py
```diff
@@ -110,45 +116,87 @@ def run(
...
+            True if the output is persistent and missing.
         """
-        free_outputs = pipeline.outputs() - set(catalog.keys())
-        missing = {ds for ds in catalog if not catalog.exists(ds)}
-        to_build = free_outputs | missing
-        to_rerun = pipeline.only_nodes_with_outputs(*to_build) + pipeline.from_inputs(
-            *to_build
-        )
+        # First check if it's explicitly in the catalog
+        dataset = catalog._datasets.get(output)
```
ElenaKhaustova: We should use public API instead

kedro/runner/runner.py
```diff
@@ -110,45 +116,87 @@ def run(
...
+        if dataset is not None:
+            # It's explicitly in the catalog
+            is_ephemeral = getattr(dataset, "_EPHEMERAL", False)
+            if is_ephemeral:
+                return False
+
+            # It's persistent - check existence
+            return not catalog.exists(output)
+
+        if output not in catalog:
```
ElenaKhaustova, line 179: This should be checked first, then get dataset from catalog

kedro/runner/runner.py
```diff
@@ -110,45 +116,87 @@ def run(
...
-        free_outputs = pipeline.outputs() - set(catalog.keys())
-        missing = {ds for ds in catalog if not catalog.exists(ds)}
-        to_build = free_outputs | missing
-        to_rerun = pipeline.only_nodes_with_outputs(*to_build) + pipeline.from_inputs(
-            *to_build
-        )
+        # First check if it's explicitly in the catalog
+        dataset = catalog._datasets.get(output)
+
+        if dataset is not None:
```
ElenaKhaustova: Returned dataset cannot be `None`

ElenaKhaustova (commented):
Thank you, @SajidAlamQB!

The implementation looks cleaner now. 

Did we test it with `ThreadRunner` and `ParallelRunner`?

SajidAlamQB:
> Did we test it with `ThreadRunner` and `ParallelRunner`?

Yes, they both worked.

tests/runner/test_sequential_runner.py
SajidAlamQB: Done.

lrcouto (approved):
Tested locally and works as expected.

kedro/runner/runner.py
```diff
@@ -110,45 +116,80 @@ def run(
...
-    def run_only_missing(
-        self,
-        pipeline: Pipeline,
-        catalog: CatalogProtocol | SharedMemoryCatalogProtocol,
-        hook_manager: PluginManager | None = None,
-    ) -> dict[str, Any]:
-        """Run only the missing outputs from the ``Pipeline`` using the
-        datasets provided by ``catalog``, and save results back to the
-        same objects.
+    def _filter_pipeline_for_missing_outputs(
```
ElenaKhaustova, line 119: @merelcht, are we sure about the decision to make this feature available only via CLI? I probably missed some context, but I want to double-check that's what we want to do.

ElenaKhaustova (commented):
I've tested it manually and it works not as I expected.

What I did:
1. Used spaceflights-pandas
2. Run full pipeline as usual 
3. Removed regressor saved after the first run
4. Run with `--only-missing-outputs` flag

My expectation:
The execution starts from the `split_data_node` since everything above is persisted, so `preprocess_companies_node`, `preprocess_shuttles_node` and `create_model_input_table_node` are skipped.

Actual output (it runs everything up to the node that outputs `regressor`):
```bash
(kedro-py3.11-test-env) ➜  only-missing kedro run --only-missing-outputs
[06/18/25 18:14:23] INFO     Using 'conf/logging.yml' as logging configuration. You can change this by setting the KEDRO_LOGGING_CONFIG         __init__.py:270
                             environment variable accordingly.
                    INFO     Kedro project only-missing                                                                                          session.py:332
[06/18/25 18:14:25] INFO     Skipping 2 nodes with existing persistent outputs: compare_passenger_capacity_exp__bbf8efcd, evaluate_model_node     runner.py:151
                    INFO     Running 5 out of 7 nodes (skipped 2 nodes)                                                                           runner.py:155
                    INFO     Using synchronous mode for loading and saving data. Use the --async flag for potential performance gains.  sequential_runner.py:59
                             https://docs.kedro.org/en/stable/nodes_and_pipelines/run_a_pipeline.html#load-and-save-asynchronously
                    INFO     Loading data from companies (CSVDataset)...                                                                   data_catalog.py:1050
                    INFO     Running node: preprocess_companies_node: preprocess_companies() ->                                                     node.py:400
                    INFO     Saving data to preprocessed_companies (ParquetDataset)...                                                     data_catalog.py:1010
[06/18/25 18:14:26] INFO     Completed node: preprocess_companies_node                                                                            runner.py:247
                    INFO     Completed 1 out of 5 tasks                                                                                           runner.py:248
                    INFO     Loading data from shuttles (ExcelDataset)...                                                                  data_catalog.py:1050
[06/18/25 18:14:28] INFO     Running node: preprocess_shuttles_node: preprocess_shuttles() ->                                                       node.py:400
                    INFO     Saving data to preprocessed_shuttles (ParquetDataset)...                                                      data_catalog.py:1010
                    INFO     Completed node: preprocess_shuttles_node                                                                             runner.py:247
                    INFO     Completed 2 out of 5 tasks                                                                                           runner.py:248
                    INFO     Loading data from preprocessed_shuttles (ParquetDataset)...                                                   data_catalog.py:1050
                    INFO     Loading data from preprocessed_companies (ParquetDataset)...                                                  data_catalog.py:1050
                    INFO     Loading data from reviews (CSVDataset)...                                                                     data_catalog.py:1050
                    INFO     Running node: create_model_input_table_node: create_model_input_table() ->                                             node.py:400
                    INFO     Saving data to model_input_table (ParquetDataset)...                                                          data_catalog.py:1010
                    INFO     Completed node: create_model_input_table_node                                                                        runner.py:247
                    INFO     Completed 3 out of 5 tasks                                                                                           runner.py:248
                    INFO     Loading data from model_input_table (ParquetDataset)...                                                       data_catalog.py:1050
                    INFO     Loading data from params:model_options (MemoryDataset)...                                                     data_catalog.py:1050
                    INFO     Running node: split_data_node: split_data() -> [X_train;X_test;y_train;y_test]                                         node.py:400
                    INFO     Saving data to X_train (MemoryDataset)...                                                                     data_catalog.py:1010
                    INFO     Saving data to X_test (MemoryDataset)...                                                                      data_catalog.py:1010
                    INFO     Saving data to y_train (MemoryDataset)...                                                                     data_catalog.py:1010
                    INFO     Saving data to y_test (MemoryDataset)...                                                                      data_catalog.py:1010
                    INFO     Completed node: split_data_node                                                                                      runner.py:247
                    INFO     Completed 4 out of 5 tasks                                                                                           runner.py:248
                    INFO     Loading data from X_train (MemoryDataset)...                                                                  data_catalog.py:1050
                    INFO     Loading data from y_train (MemoryDataset)...                                                                  data_catalog.py:1050
                    INFO     Running node: train_model_node: train_model([X_train;y_train]) ->                                                      node.py:400
                    INFO     Saving data to regressor (PickleDataset)...                                                                   data_catalog.py:1010
                    INFO     Completed node: train_model_node                                                                                     runner.py:247
                    INFO     Completed 5 out of 5 tasks                                                                                           runner.py:248
                    INFO     Pipeline execution completed successfully in 2.5 sec.
```

So I think that `pipeline.to_outputs(*missing_persistent_outputs)` does not reduce the pipeline to minimum.

kedro/runner/runner.py
ElenaKhaustova: Previously this was a public method for runner

SajidAlamQB:
> I've tested it manually and it works not as I expected.
> Actual output (it runs everything up to the node that outputs `regressor`):
> So I think that `pipeline.to_outputs(*missing_persistent_outputs)` does not reduce the pipeline to minimum.

Hey @ElenaKhaustova, thanks for testing! This behaviour is actually the intended for MVP implementation.

The current implementation does **pipeline-level checking** rather than **node-level checking**:

1. It identifies that `regressor` is missing
2. Uses `pipeline.to_outputs("regressor")` to build the complete pipeline needed to produce `regressor`
3. This includes ALL upstream nodes, even if their intermediate outputs already exists

What you expected (skipping nodes whose outputs already exist) would require node-level checking, which is more complex and is to be done post-1.0.0.

This was discussed in the [spike issue](https://github.com/kedro-org/kedro/issues/4635#issuecomment-2880346189) where the team decided:
- **For 1.0.0 MVP**: Simple pipeline-level approach (what's implemented here)
- **Post 1.0.0**: More sophisticated node-level checking with proper change detection (hashing, etc.)

So while `preprocessed_companies`, `preprocessed_shuttles`, and `model_input_table` exist, the MVP still runs those nodes because they're part of the pipeline to produce the missing `regressor`.

For now, this feature is most useful when you have multiple independent pipeline branches, it will skip entire branches whose final outputs exist.

kedro/runner/runner.py
SajidAlamQB: The previous `run_only_missing()` method on the runner was removed because it was never actually used in the codebase (only in tests) and it was added back in Kedro 0.14 but never documented properly (only mentioned once in the docs).

The new implementation is integrated directly into the `run()` method via the `only_missing_outputs` parameter, which can be:

- Used via CLI: `kedro run --only-missing-outputs`
- Used programmatically: `runner.run(pipeline, catalog, only_missing_outputs=True)`

So the functionality is still accessible programmatically it's just part of the main `run()` method now rather than a separate method. The `_filter_pipeline_for_missing_outputs` is now just a helper method.

ElenaKhaustova:
> Hey @ElenaKhaustova, thanks for testing! This behaviour is actually the intended for MVP implementation.
> 
> The current implementation does **pipeline-level checking** rather than **node-level checking**:
> 
> 1. It identifies that `regressor` is missing
> 2. Uses `pipeline.to_outputs("regressor")` to build the complete pipeline needed to produce `regressor`
> 3. This includes ALL upstream nodes, even if their intermediate outputs already exists
> 
> What you expected (skipping nodes whose outputs already exist) would require node-level checking, which is more complex and is to be done post-1.0.0.
> 
> This was discussed in the [spike issue](https://github.com/kedro-org/kedro/issues/4635#issuecomment-2880346189) where the team decided:
> 
> * **For 1.0.0 MVP**: Simple pipeline-level approach (what's implemented here)
> * **Post 1.0.0**: More sophisticated node-level checking with proper change detection (hashing, etc.)
> 
> So while `preprocessed_companies`, `preprocessed_shuttles`, and `model_input_table` exist, the MVP still runs those nodes because they're part of the pipeline to produce the missing `regressor`.
> 
> For now, this feature is most useful when you have multiple independent pipeline branches, it will skip entire branches whose final outputs exist.

Thanks for explaining, I see your points.

My understanding was that:
1. For 1.0.0 MVP, we assume that all persisted datasets do not need to be recreated. In the above example, we do exactly this with the datasets created by nodes following the node that created the regressor (missing). So, with the current implementation, we're treating nodes preceding the regressor and after it in different ways. I find it a bit confusing, especially given the name of the flag `--only-missing-outputs`. We can achieve the behaviour when we actually run only missing and still keep the implementation simple. That requires just one graph traversal, as mentioned before.
2. For post 1.0.0, we will apply a more complex approach with hashing, where we identify whether the node logic has changed and whether we need to recreate existing persisted datasets.

If my understanding is not correct, I do not want to block the PR, and I am happy to approve it. Need more opinions from the team.

kedro/runner/runner.py
ElenaKhaustova: You're right, my bet

SajidAlamQB:
> 1. For 1.0.0 MVP, we assume that all persisted datasets do not need to be recreated. In the above example, we do exactly this with the datasets created by nodes following the node that created the regressor (missing). So, with the current implementation, we're treating nodes preceding the regressor and after it in different ways. I find it a bit confusing, especially given the name of the flag `--only-missing-outputs`. We can achieve the behaviour when we actually run only missing and still keep the implementation simple. That requires just one graph traversal, as mentioned before.

Hi @ElenaKhaustova, I think understand what your expectation was for the `--only-missing-outputs` behaviour. This PR (the pipeline-level approach) does treat upstream and downstream nodes differently relative to an missing output like regressor.

In your example, even though `model_input_table.parquet` exists, the current implementation would still re-run `create_model_input_table_node` because it uses `pipeline.to_outputs()` which builds the complete subgraph needed for the target without checking intermediate persistence.

To address this and provide the team with both options, I have an alternative PR: #4859 (Node-based `--only-missing-outputs`). This new PR implements the logic using a reverse topological graph traversal. For each node, it checks:

- If it produces a missing final pipeline output.
- If any downstream node that must run needs an output from this node (and that output is ephemeral or persistent-but-missing).
- If the node itself has any of its own persistent outputs missing.

**Key differences:**

**OPTION 1**: PR #4848 (Current): Simpler pipeline-level approach, runs all upstream nodes.
**OPTION 2**: PR #4859 (Alternative): Granular node-level approach, skips nodes with existing persistent outputs.

The main difference between the PRs is the filtering logic found in the `runner.py` otherwise they are almost the same barring tests.

Would love the team's input on which approach better for the 1.0.0 MVP! @merelcht @ravi-kumar-pilla

merelcht:
I haven't looked at the code yet, but started first testing the feature. I'm quite confused about the behaviour. I compared this to the [old "run_only_missing" functionality](https://github.com/kedro-org/kedro/blob/main/kedro/runner/runner.py#L160) that was added way back in 0.14.0 and that makes more sense to me. 

> It correctly identifies ephemeral datasets (e.g., those with _EPHEMERAL = True, like MemoryDataset) and considers them as "not needing to be checked for existence" for this feature's purpose.

I don't think the above assumption is correct and I find it rather confusing that `MemoryDataset` are considered "existing". They are ephemeral so I would expect nodes that have ephemeral output to always re-run, _unless_ there is a persisted output further down the line. For example, the spaceflights `data_science` pipeline output is "None" (it prints logging message), so I'd expect this node to always run if this is the end of my pipeline. [This also corresponds with the proposal on the original issue: ](https://github.com/kedro-org/kedro/issues/4635#issuecomment-2880346189)

> We will work on a proposal for an MVP which checks for if dataset is missing i.e., a node should not run if its output is persistent and exists. If possible come up with a non-breaking way to introduce it in Kedro v1.0.0.

I'd suggest we get back to the drawing board to clarify the requirements for this feature.

Update: quickly tested https://github.com/kedro-org/kedro/pull/4859, which is more in line with my expectations, but the final node in `data_science` is still not re-run.

SajidAlamQB:
> Update: quickly tested #4859, which is more in line with my expectations, but the final node in `data_science` is still not re-run.

Hi @merelcht, Thanks for the detailed feedback and testing. I had some different interpretations for the MVP, especially around ephemeral datasets and nodes with no persistent outputs. Thanks for clearing that up.

I've now updated the implementation in #4859 (Node-based --only-missing-outputs) with algorithm that uses a reverse topological sort and specific rules for each node. I believe this new version in #4859 aligns much more closely with the behaviour we expected:

- Nodes with no outputs will run if they are part of the active execution path.

- Nodes producing ephemeral outputs will run if their output is needed by a downstream node that is set to run. They are not skipped just because a further downstream persistent output exists (unless that makes their consumer skippable).

- If an upstream node A produces a persistent output `data_A` that exists, and a downstream node B (which needs `data_A`) must run (maybe because B's own output is missing), then A will be skipped, and B will load the existing `data_A`.

Since #4859 reflects what we wanted the MVP to be I propose we close this PR, #4848 and focus on #4859 (Node-based --only-missing-outputs).

Would you be able to take another look at the behaviour with the latest changes in PR #4859 when you have some time. Thanks.

SajidAlamQB:
Closing in favour of: https://github.com/kedro-org/kedro/pull/4859

PR #4859, opened 2025-06-18
Node-based `--only-missing-outputs` for granular skipping

SajidAlamQB:
## Description

This PR brings in the `--only-missing-outputs` CLI flag for `kedro run`.

The goal here is to skip a node if its outputs are already there and aren't needed, making re-runs faster.

**How it works:**

When `--only-missing-outputs` is used:

1.  We first figure out the "parent-child" relationships between nodes.
2.  Then, we go through the pipeline nodes in **reverse topological order** (from the end to start).
3.  For each node, we decide if it *must* run based on these main rules:
    *   **Node has no outputs?** This ensures side-effect nodes run if their part of the pipeline is active.
    *   **Node has its own missing persistent output?** Yes, it runs. This ensures any broken persistent node gets fixed, even for intermediate datasets.
    *   **Is an output of this node (ephemeral or persistent-and-missing) needed by a downstream "child" node that we've *already decided must run*?** Yes, then this "parent" node also runs to supply that child.

This MVP focuses on the *existence* of persistent datasets. More advanced change detection (like hashing to see if data or code is "stale") is a separate discussion for post-1.0.0 (see Spike: #4635).

**Development Notes:**

*   **CLI:** Uses the `--only-missing-outputs` flag.
*   **Session:** `KedroSession.run()` passes the `only_missing_outputs` flag to the runner.
*   **Runner Logic (`AbstractRunner`):**
    *   The main `run()` method calls `_filter_pipeline_for_missing_outputs()` if the flag is true.
    *   `_filter_pipeline_for_missing_outputs()`: handles the node-based filtering.
        *   Calls `_build_node_children_map()` to get dependencies.
        *   Calls `_determine_nodes_to_run()` which iterates in reverse topological order.
        *   `_determine_nodes_to_run()` uses `_should_node_run()` for each node.
    *   `_should_node_run()`: Goes through three rules mentioned above, calling helpers like:
        *   `_has_missing_persistent_outputs()`
        *   `_outputs_needed_by_children()`

## Testing Spaceflights

**Scenario: All Persistent Outputs Exist**

*   Run `kedro run` once to generate everything.
*   Run `kedro run --only-missing-outputs`.
*   **Expected:**
    *   Nodes producing only **persistent outputs** that all exist will be **skipped**. 

**Scenario: Delete `regressor.pickle`**

*   Run `kedro run` once.
*   Delete `data/06_models/regressor.pickle`.
*   Run `kedro run --only-missing-outputs`.
*   **Expected:**
    *   `train_model_node`: **Must run** (its persistent output `regressor` is missing).
    *   `split_data_node`: **Must run** (its ephemeral outputs `X_train`, `y_train` are needed by `train_model_node` which is running).
    *   `evaluate_model_node`: **Must run** (it has no outputs, so it's always a candidate; also, its input `regressor` is being recomputed).

*   Run `kedro run` once.
*   Delete `data/03_primary/model_input_table.parquet`.
*   Run `kedro run --only-missing-outputs`.
*   **Expected:**
    *   `create_model_input_table_node`: **Must run** (its persistent output `model_input_table` is missing).
    *   `split_data_node`: **Must run** 
    *   `evaluate_model_node`: **Must run** 

*   Temporarily edit `conf/base/catalog.yml` for `regressor`: change `type: pickle.PickleDataset` to `type: pandas.CSVDataset`.
*   Delete `data/06_models/regressor.pickle` if it exists.
*   Run `kedro run --only-missing-outputs`.
*   **Expected:** it **should fail** with an error.

**Scenario: Interaction with `--to-nodes split_data_node`**

*   Run `kedro run` once (all outputs exist, including `model_input_table.parquet`).
*   Run `kedro run --to-nodes split_data_node --only-missing-outputs`.
*   **Expected:**
    *  All nodes in this sub-pipeline are **SKIPPED**.

**Scenario: Running a specific pipeline (`--pipeline data_science`)**

*   Run `kedro run` once.
*   Delete `data/06_models/regressor.pickle`.
*   Run `kedro run --pipeline data_science --only-missing-outputs`.
*   **Expected:**
    *   All three nodes (`split_data_node`, `train_model_node`, `evaluate_model_node`) in the `data_science` pipeline will run.** `model_input_table.parquet` (input to the `data_science` pipeline) will be loaded as it exists.

## Checklist


rashidakanchwala:
I haven't looked at this code-wise. But I tested all the different scenarios on `demo-project` and it's working as expected :)

kedro/runner/runner.py
```diff
@@ -105,45 +111,193 @@ def run(
...
+        output_to_node, node_consumers = self._build_dependency_maps(pipeline)
+
+        # Determine which nodes need to run
+        nodes_to_run: set[Node] = set()
+        for node in sorted_nodes:
+            run_this_node = False
+
+            # Always run nodes with no outputs
+            if self._should_run_for_no_outputs(node):
+                run_this_node = True
```
DimedS: Why do we always run nodes that have no outputs?

DimedS (approved):
Great job, @SajidAlamQB! I really like the structure of the code - extending the run function with an additional pipeline filtering step is clean and intuitive. The algorithm itself is also now well designed, requiring just a single pipeline traversal with what appears to be linear complexity. I just have one small question.

kedro/runner/runner.py
ravi-kumar-pilla: +1 ... are these leaf nodes of a pipeline and if there is no further dependency why should we run these ?

ravi-kumar-pilla:
Awesome work @SajidAlamQB .... The code looks way better and the algorithm seems to cover corner cases. I tested the scenarios mentioned here and will need some more time for testing few more cases. But overall the code looks great apart from the minor [comment](https://github.com/kedro-org/kedro/pull/4859#discussion_r2160557890).

Also while testing (using spaceflights-pandas starter) for -

**Scenario: Delete model_input_table.parquet**

**Actual** - `train_model_node` , `preprocess_companies_node` and `preprocess_shuttles_node` are skipped which differs from expected. Why would these nodes run ? I think it should be `create_model_input_table_node` that must run as it rightly does

```bash
        INFO     Skipping 6 nodes with existing outputs:                          runner.py:213
                             compare_passenger_capacity_exp__822c8b63,                                     
                             compare_passenger_capacity_go__1e477111,                                      
                             create_confusion_matrix__2789b6a9, preprocess_companies_node,                 
                             preprocess_shuttles_node, train_model_node
```

- This is same as `kedro run`. Is the idea behind the test case to show the error handling is still the same ?

Thank you

kedro/runner/runner.py
merelcht: This probably comes from my comment on: https://github.com/kedro-org/kedro/pull/4848#issuecomment-2988158294

Arguably, it doesn't make sense to run nodes with `None` output, but when I was testing with spaceflights it seemed odd to me. Happy to change this "requirement" if others don't think it makes sense.

SajidAlamQB:
>  think it should be `create_model_input_table_node` that must run as it rightly does

Thanks @ravi-kumar-pilla for reviewing and testing. You're correct this scenario is now fixed in description.

> **Scenario: Dataset with Faulty Catalog Definition**
> 
> * This is same as `kedro run`. Is the idea behind the test case to show the error handling is still the same ?
> 

Yes, to essentially show that we are "failing fast" and haven't changed fundamental error handling.

kedro/runner/runner.py
SajidAlamQB: As Merel said, I don't have a strong preference myself. 

What does the team think is the more intuitive?

Example:

Should `evaluate_model_node` run if `regressor.pickle` was loaded from an existing file (and `train_model_node` was skipped)?

Or should it only run if `regressor.pickle` was actually re-generated by `train_model_node` in this specific `--only-missing-outputs` run?

Happy to adjust the logic.

docs/getting-started/commands_reference.md
```diff
@@ -118,6 +118,7 @@ the names of relevant nodes, datasets, envs, etc. in your project.
...
 | `kedro run --conf-source=<path_to_compressed file>`                 | Only possible when using the [``OmegaConfigLoader``](../configure/configuration_basics.md#omegaconfigloader). Specify a compressed config file in `zip` or `tar` format.                                                                            |
 | `kedro run --params=<param_key1>=<value1>,<param_key2>=<value2>`    | Does a parametrised run with `{"param_key1": "value1", "param_key2": 2}`. These will take precedence over parameters defined in the `conf` directory. Additionally, dot (`.`) syntax can be used to address nested keys like `parent.child:value` |
+| `kedro run --only-missing-outputs`                                  | Run only the nodes required to produce missing persistent outputs. If a node's persistent outputs already exist, it and its upstream dependencies (if not needed for other missing outputs) will be skipped.                                      |
```
merelcht:
```suggestion
| `kedro run --only-missing-outputs`                                  | Run only the nodes required to produce missing persistent outputs. If a node's persistent outputs already exist, the node and its upstream dependencies (if not needed for other missing outputs) will be skipped.                                      |
```

kedro/runner/runner.py
```diff
@@ -105,45 +111,193 @@ def run(
 
         return run_output
 
-    def run_only_missing(
```
merelcht, line 108: Happy to see this removed. Since it was a public method, its removal should be explicitly mentioned in the release notes.

kedro/runner/runner.py
```diff
@@ -105,45 +111,193 @@ def run(
...
+            self._logger.info(
+                f"Skipping {skipped_count} nodes with existing outputs: "
+                f"{', '.join(sorted(skipped_names))}"
+            )
+
+        self._logger.info(
+            f"Running {final_node_count} out of {original_node_count} nodes"
+        )
+
+    def _filter_pipeline_for_missing_outputs(
```
merelcht: Nit: can you order the methods in the order in which they are called? `_filter_pipeline_for_missing_outputs` is the top level called from `run` but now it's all the way at the bottom, making it harder to read the code.

kedro/runner/runner.py
```diff
@@ -105,45 +111,193 @@ def run(
...
+    ) -> Pipeline:
+        """Filter pipeline using reverse topological order."""
+        original_node_count = len(pipeline.nodes)
+
+        # Get nodes in reverse topological order
+        sorted_nodes = list(pipeline.nodes)
+        sorted_nodes.reverse()
+
+        # Build dependency mappings
+        output_to_node, node_consumers = self._build_dependency_maps(pipeline)
```
merelcht: `output_to_node` isn't used so you can change it to `_`:

```suggestion
        _, node_consumers = self._build_dependency_maps(pipeline)
```

kedro/runner/runner.py
```diff
@@ -105,45 +111,193 @@ def run(
 
         return run_output
 
-    def run_only_missing(
+    def _build_dependency_maps(
+        self, pipeline: Pipeline
+    ) -> tuple[dict[str, Node], dict[Node, set[Node]]]:
+        """Build mappings for efficient node dependency lookups.
```
merelcht: Can we use the parent/child terminology instead of "consumer" "producer" "mappings"? The parent/child terminology is what is used throughout the codebase.

kedro/runner/runner.py
```diff
@@ -105,45 +111,193 @@ def run(
...
+                output_to_node[output] = node
+
+        # Build consumer relationships
+        for node in pipeline.nodes:
+            for inp in node.inputs:
+                if inp in output_to_node:
+                    producer = output_to_node[inp]
+                    node_consumers[producer].add(node)
+
+        return output_to_node, node_consumers
```
merelcht: I noticed below in `_filter_pipeline_for_missing_outputs` that `output_to_node` isn't used, why is it returned?

kedro/runner/runner.py
```diff
@@ -105,45 +111,193 @@ def run(
...
+        output_to_node: dict[str, Node] = {}
+        node_consumers: dict[Node, set[Node]] = {node: set() for node in pipeline.nodes}
+
+        for node in pipeline.nodes:
+            for output in node.outputs:
+                output_to_node[output] = node
+
+        # Build consumer relationships
+        for node in pipeline.nodes:
+            for inp in node.inputs:
```
merelcht: We always aim to have more explicit argument names rather than abbreviations in the codebase.

```suggestion
            for input in node.inputs:
```

kedro/runner/runner.py
```diff
@@ -105,45 +111,193 @@ def run(
...
 
         Args:
-            pipeline: The ``Pipeline`` to run.
-            catalog: An implemented instance of ``CatalogProtocol`` or ``SharedMemoryCatalogProtocol`` from which to fetch data.
-            hook_manager: The ``PluginManager`` to activate hooks.
-        Raises:
-            ValueError: Raised when ``Pipeline`` inputs cannot be
-                satisfied.
+            output: The output dataset name.
+            catalog: The data catalogue to check.
```
merelcht:
```suggestion
            catalog: The data catalog to check.
```

kedro/runner/runner.py
```diff
@@ -105,45 +111,193 @@ def run(
...
+
+        dataset = catalog.get(output)
+        if getattr(dataset, "_EPHEMERAL", False):
+            # Ephemeral output - must be generated
+            return True
+
+        # Persistent dataset - check if it exists
+        return not catalog.exists(output)
+
+    def _should_run_for_downstream_consumer(
```
merelcht: This method is almost unreadable because of all the if/else/for loop nesting. Can you rewrite that to make it easier to follow?

kedro/runner/runner.py
```diff
@@ -105,45 +111,193 @@ def run(
...
+        """Check if node should run because it has missing persistent outputs."""
+        for output in node.outputs:
+            if self._is_persistent_and_missing(output, catalog):
+                self._logger.debug(
+                    f"Node '{node.name}' must run: has missing output '{output}'"
+                )
+                return True
+        return False
+
+    def _is_output_needed_by_consumer(
```
merelcht: This code is pretty much the opposite of the persistent check `_is_persistent_and_missing`, can we consolidate these methods? Or at least use similar naming? It confused me that this name is so different when the logic is so similar.

merelcht (commented):
I've done an initial review and left comments. I find the code quite hard to follow. I don't think the methods are broken up in the most logical way. 

I'm guessing you've used copilot on this code, because some of the terminology isn't very common in our codebase. Let's make sure to make code consistent with the existing base in terms of naming formats and terminology.

kedro/runner/runner.py
ravi-kumar-pilla: I do not fully understand (the use case) why someone needs to have a node which does not provide an output...since we consider our nodes to be pure python functions which do not alter inputs and we have hooks to do any validation steps or logging info.

However if there needs to be some kind of logging or any form of validation that needs to be done via nodes, the node has to run.  

For the initial version to increase flag adoption, I think it is fine to go ahead with running the nodes with `None` as output (considering the users are doing some important logging and we do not want them to change that just to use the flag).

Thank you

SajidAlamQB:
> I've done an initial review and left comments. I find the code quite hard to follow. I don't think the methods are broken up in the most logical way.

Hi @merelcht, I've updated the ordering of the runner code to flow more easily and make it more readable and broken down some of large nested code into smaller methods and adjusted language that has gone through an llm to align with our vocab this should be easier to read now. Let me know if you have any other thoughts.

kedro/runner/runner.py
```diff
@@ -105,45 +111,211 @@ def run(
...
+        return nodes_to_run
 
+    def _should_node_run(
+        self,
+        node: Node,
+        catalog: CatalogProtocol | SharedMemoryCatalogProtocol,
+        nodes_to_run: set[Node],
+        node_children: dict[Node, set[Node]],
+    ) -> bool:
+        """Check if a node should run based on various criteria."""
```
merelcht: Can you add explicitly in the doc string what these criteria are? Better to be a bit verbose.

kedro/runner/runner.py
```diff
@@ -105,45 +111,211 @@ def run(
...
+            return True
+
+        # Run if node's outputs are needed by children that will run
+        if self._outputs_needed_by_children(node, catalog, nodes_to_run, node_children):
+            return True
+
+        return False
+
+    def _has_missing_persistent_outputs(
+        self, node: Node, catalog: CatalogProtocol | SharedMemoryCatalogProtocol
```
merelcht: Just noticed that a lot of these methods take `self` as first argument because they're inside the runner class, but are not actually runner related. It makes more sense to move these outside the class. They can still be in the file but just outside the class.

kedro/runner/runner.py
```diff
@@ -105,45 +111,211 @@ def run(
...
+                if self._is_dataset_ephemeral_or_missing(dataset, catalog):
+                    self._logger.debug(
+                        f"Node '{node.name}' must run: produces '{dataset}' "
+                        f"needed by running child '{child.name}'"
+                    )
+                    return True
+
+        return False
+
+    def _is_dataset_missing(
```
merelcht: It's kind of confusing that this method and the one below have very similar and overlapping logic, but are still different. Can we maybe rename this one to `_is_persistent_dataset_missing`?

merelcht (commented):
Thanks for addressing my comments @SajidAlamQB ! I've left a couple more, most important is about the placement of the methods.

kedro/runner/runner.py
```diff
@@ -690,3 +548,156 @@ def _find_initial_node_group(pipeline: Pipeline, nodes: Iterable[Node]) -> list[
...
+                logger.debug(
+                    f"Node '{node.name}' must run: produces '{dataset}' "
+                    f"needed by running child '{child.name}'"
+                )
+                return True
+
+    return False
+
+
+def is_persistent_dataset_missing(
```
merelcht: Any reason why this is a public method and all the others private?

merelcht (commented):
I think this is nearly there! The code is looking much cleaner and I find it easy to follow now 🙂 

One final suggestion I have is to add an e2e test for this feature in the `run.feature` file.

PR #4877, opened 2025-06-20
Release Kedro 1.0.0rc1

ElenaKhaustova:
## Description
🚀

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -6,7 +6,7 @@
 ## Documentation changes
 ## Community contributions
 
-# [PRE-RELEASE] 1.0.0rc1
+# Release 1.0.0rc1
 
 ## Major features and improvements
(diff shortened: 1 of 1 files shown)


rashidakanchwala:
When I created this GitHub workflow, I structured the release notes with headings like this — https://github.com/rashidakanchwala/kedro/commit/6b77168686327c777a399469b3ba542f70832b4e.

I think we do this because otherwise it might not work properly. See here - https://github.com/kedro-org/kedro/blob/main/.github/workflows/check-release.yml#L79

DimedS (approved):
🍾

ankatiyar:
Yeah I think @rashidakanchwala is right, iirc the release notes heading needs to be "Release xxx" so that the automatic notes are generated properly

merelcht (approved):
🥳🍾🎈

PR #4896, opened 2025-06-25
Update kedro version to align with RC

ElenaKhaustova:
## Description
We need to update Kedro version to align with RC: https://github.com/kedro-org/kedro/pull/4877

Currently, as @ankatiyar reported, if one has the `main` version installed but in their project the `kedro_init_version` is set to `1.0.0rc1`, it doesn’t work when `kedro run`.

```console
 ValueError: Your Kedro project version 1.0.0rc1 does not match Kedro package version 0.19.14 you are running. Make sure to update your project 
template. See https://github.com/kedro-org/kedro/blob/main/RELEASE.md for how to migrate your Kedro project.
```

Suggested steps:
1. **Disable the release workflow**
2. Merge this PR
3. Enable the release workflow

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -44,6 +44,6 @@ authors:
 - family-names: Brugman
   given-names: Simon
 title: Kedro
-version: 0.19.14
-date-released: 2025-06-17
+version: 1.0.0rc1
+date-released: 2025-06-20
 url: https://github.com/kedro-org/kedro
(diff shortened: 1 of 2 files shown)


SajidAlamQB (approved):

merelcht (approved):

ankatiyar:
Side note: I was thinking one shouldn't have to update their project's `kedro_init_version` to make it work with the higher version of Kedro, as it is the "init_version". As long as the project is compliant, it should work? Does this warrant a ticket? I know it only affects projects when there's breaking versions so might not be worth changing the behaviour. WDYT?

Issue #4922, opened 2025-06-30
Rename `dataset.to_config()` method


ElenaKhaustova:
## Description
`catalog.to_config()` method supports capturing datasets added dynamically to the catalog. Internally, it uses `dataset.to_config()`, which we made a public API as well.  The last outputs the dataset configuration saved while dataset `__init__`, so it does not support dynamic dataset changes.

![Image](https://github.com/user-attachments/assets/8ee15510-554e-4b4f-8926-eb69c362543e)

Since `dataset.to_config()` is a public API we suggest renaming it to better reflect its purpose.

astrojuanlu:
So, `ds._load_args` works, but `ds.to_config()` won't reflect it, is that it?

Rather than renaming the method, should we think of making `to_config()` reflect these dynamic changes? Or make datasets immutable?

(Not for 1.0 necessarily, but I am confused by this behavior)

ElenaKhaustova:
> ds._load_args works, but ds.to_config() won't reflect it, is that it?

Only stores dataset `__init__` arguments, so it won't reflect any change unless replacing dataset.

This was done that way for two reasons:

1. there was a strong push not to allow dynamic ds modifications, so this behaviour is not expected in general and we only allow replacement
2. we didn’t want to modify each kedro dataset, so at the level of the abstract dataset on the frameworks side we only capture dataset init args, that’s why any modifications are not reflected

> Rather than renaming the method, should we think of making to_config() reflect these dynamic changes?

Probably yes, if we decide that point 1 is not what we expect anymore, and not for 1.0 due to the time shortage.

> Or make datasets immutable?

One can't change datasets properties via catalog. But the problem now is that we made this method public for datasets.

I agree that the behaviour is confusing. To mitigate this for 1.0, we can make the `ds.to_config()` method private or rename it to reflect its actual functionality.

Galileo-Galilei:
I didn't had in mind this was made on purpose, but I think this misses the point of this feature request. In my understanding (and selfishly, for kedro-mlflow), I clearly want to serialize the catalog and have an idempotent pipeline, i.e. serialize / deserialize the catalog has the exact same behaviour. I don't see use case for which I want to get the original version of the Dataset.

If we don't have better solutions, I'd prefer make it private and revisit after 1.0 (maybe that's the expected behaviour, but I don't think I would use it as is personally, there's too much risk to have a deferred bug only found at reading time, potentially days/weeks/months after)

PR #4932, opened 2025-07-01
Rename `AbstractDataset.to_config()` method

ElenaKhaustova:
## Description
Solves https://github.com/kedro-org/kedro/issues/4922

## Development notes
- Renamed to `_init_config()` to clearly signal it's for internal use.
- Updated docstring to emphasize that this captures initial configuration only, not dynamic state. 
- Preserved all existing logic.

## Checklist

### DIFF
--- a/kedro/framework/context/catalog_mixins.py
+++ b/kedro/framework/context/catalog_mixins.py
@@ -173,7 +173,7 @@ class CatalogCommandsMixin:
                 continue
 
             unresolved_config, _ = self.config_resolver._unresolve_credentials(
-                ds_name, ds.to_config()
+                ds_name, ds._init_config()
             )
             explicit_datasets[ds_name] = unresolved_config
 
--- a/kedro/io/core.py
+++ b/kedro/io/core.py
@@ -204,10 +204,12 @@ class AbstractDataset(abc.ABC, Generic[_DI, _DO]):
             ) from err
         return dataset
 
-    def to_config(self) -> dict[str, Any]:
-        """Converts the dataset instance into a dictionary-based configuration for
-        serialization. Ensures that any subclass-specific details are handled, with
-        additional logic for versioning and caching implemented for `CachedDataset`.
+    def _init_config(self) -> dict[str, Any]:
+        """Internal method to capture the dataset's initial configuration
+        as provided during instantiation.
+
+        This configuration reflects only the arguments supplied at `__init__` time
+        and does not account for any runtime or dynamic changes to the dataset.
 
         Adds a key for the dataset's type using its module and class name and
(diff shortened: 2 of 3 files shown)


DimedS (commented):
Thanks, @ElenaKhaustova.

This part seems a bit confusing to me. If `to_config()` on the dataset class level is misleading because it returns only the initial constructor parameters, then I agree it’s a good idea to consider renaming it.

However, I had a look at the `DataCatalog.to_config()` method, and it seems that it ultimately just uses the same `_init_config()` for each dataset.

Could you clarify how the initial config from the dataset level is transformed into the final result returned by the catalog-level `to_config()`? Specifically, why can't we define a more accurate `to_config()` at the dataset level that fully reflects its current state and reuse it directly in the catalog?

ElenaKhaustova:
@DimedS please see the explanation at the original ticket: https://github.com/kedro-org/kedro/issues/4922#issuecomment-3023572048

ElenaKhaustova:
@DimedS, the short answer is because we do not allow dataset modifications, only replacement

DimedS:
> @DimedS, the short answer is because we do not allow dataset modifications, only replacement

I'm trying to understand how DataCatalog.to_config() works well, if it's just based on _init_config() that always returns initial state. As I understood it's because on the level of dataCatalog in case of dataset modification we are replacing Dataset objects with a new one - actual, and forget about previous object. So catalog to_config() method will always return actual config because it will be based on _init versions of last objects, right? And each dataset object is immutable so can store only initial state, but because this is not clear for all of our users, we are deciding to clarify it with renaming, correct? I'm trying to understand what kind of problem they can face with current name to_config(), they can have an outdated Dataset object, and try to use to_config() on it with intent to receive actual state? But in that case their problem is not a method name - they need to use to a new object, right?

ElenaKhaustova:
@DimedS yep, your understanding of how it currently works is correct

> I'm trying to understand what kind of problem they can face with current name to_config()

The problem is that one cannot modify datasets via catalog, so `catalog.to_config()` works as expected. But one can manually create a dataset and modify its properties (see the screenshot in the original issue with the example), then `dataset.to_config()` looks confusing because it does not reflect the dynamic changes for the dataset.

Since we didn't mean this API `dataset.to_config()` to be used externally, we've renamed it to reflect its actual purpose and made it private. Additionally, it will simplify adding proper `dataset.to_config()` method in future if we decide so as it won't cause a breaking change.

ElenaKhaustova:
@DimedS 

> So it seems that Datasets technically mutable, but immutable only in case of proper usage with DataCatalog?

yep, that's right

Issue #4901, opened 2025-06-25
Confusing error when datasets/parameters aren't defined in catalog


merelcht:
## Description
When a pipeline has a namespace, but the namespace hasn't been added to the datasets in the catalog the error message is quite confusing.

## Steps to Reproduce

1. Create a kedro project with the example
2. Add a namespace "dp" to the data_processing pipeline
3. Run `kedro run`

## Expected Result
The error you get with 0.19.14
```
ValueError: Pipeline input(s) {'dp.shuttles', 'dp.companies', 'dp.reviews'} not found in the DataCatalog
```

## Actual Result
The error you get with the release candidate

```
DatasetError: Data for MemoryDataset has not been saved yet.
```

## Your Environment

* Kedro version used (pip show kedro or kedro -V): 1.0.0-rc1
* Python version used (python -V): 3.13

merelcht:
You get this same error when trying to access parameters that don't exist. E.g. delete the `parameters_data_science.yml` file from the example spaceflights project and do a `kedro run`. Where previously in 0.19.14 you'd see:
```
ValueError: Pipeline input(s) {'params:model_options'} not found in the DataCatalog
```

merelcht:
Fixed in #4941

PR #4941, opened 2025-07-03
Bring back old warm-up logic & error

merelcht:
## Description
Addresses https://github.com/kedro-org/kedro/issues/4901

## Development notes
I've put back the logic that exists in `0.19.14` and older that checks if any pipeline inputs are missing from the catalog.
This logic was removed as part of: https://github.com/kedro-org/kedro/pull/4592.

## Checklist


kedro/runner/runner.py
```diff
@@ -86,10 +86,24 @@ def run(
         if only_missing_outputs:
             pipeline = self._filter_pipeline_for_missing_outputs(pipeline, catalog)
 
+        # Check which datasets used in the pipeline are in the catalog or match
```
ElenaKhaustova, line 89: Thank you, @merelcht! I think it makes sense to bring back this check to provide a more informative error message in case the pipeline input is not in the catalog.

There are currently at least two cases when it will not work properly (we will still get `DatasetError: Data for MemoryDataset has not been saved yet.`):

1. If the user sets custom `catch-all` pattern, because then `ds in catalog` will always return `True`
2. If memory dataset is explicitly declared in `catalog.yml` - because we only check if the dataset is in the catalog

If we modify this check a bit, it should cover all the cases. For example, we can check inputs for existence. Cause we know that they definitely should exist in order to make the pipeline run.

```python
existing_ds = []
for ds in pipeline.datasets():
    _ = catalog.get(ds, fallback_to_runtime_pattern=True)
    if catalog.exists(ds):
        existing_ds.append(ds)

# Check if there are any input datasets that aren't in the catalog, 
# don't match a pattern in the catalog or do not exist
unsatisfied = pipeline.inputs() - set(existing_ds)

if unsatisfied:
    raise ValueError(
        f"Pipeline input(s) {unsatisfied} not found in the {catalog.__class__.__name__} or do not exist"
    )
```

ElenaKhaustova (approved):
LGTM 👍

ankatiyar (approved):
LGTM

Issue #4943, opened 2025-07-04
Exponential slowdown in `_validate_namespaces` for large layered pipelines


g199209:
### Summary
`kedro.pipeline.pipeline.Pipeline._validate_namespaces` performs an un-optimised DFS that becomes prohibitively slow for moderately large, layered pipelines.  
In a real-world project the command `kedro registry list` appears to “hang”; `py-spy` shows the process spending all its time recursing inside `_validate_namespaces`.

### Environment
* Kedro version: 0.19.3  
* Python version: 3.12.11  
* OS: Linux 4.18.0-477

### Minimal reproduction
The essential structure can be reproduced with a synthetic pipeline that has  

* **N** parallel “worker” nodes per layer, each consuming the previous layer’s single dataset, and  
* a “merge” node that takes the *N* outputs and produces the next layer’s single dataset.  

```
                 (previous-layer output)
                          │
        ┌─────┬─────┬ … ┬─────┐
        │     │     │   │     │
        ▼     ▼     ▼   ▼     ▼
    [worker₁] …               [workerₙ]
        │     │     │   │     │
        └─────┴─────┴ … ┴─────┘
                          │
                          ▼
                     [merge node]
                          │
                          ▼
                 (current-layer output)
```

```python
from kedro.pipeline import node, Pipeline

def dummy(x):  # no-op
    return x

def build_layered_pipeline(N=9, M=8):
    layers = []
    prev = "input_0"
    for layer in range(M):
        fanout = [f"w{layer}_{i}" for i in range(N)]
        workers = [
            node(dummy, prev, o, name=f"worker_{layer}_{i}")  # namespace intentionally None
            for i, o in enumerate(fanout)
        ]
        merge = node(dummy, fanout, f"input_{layer+1}",
                     name=f"merge_{layer}")
        layers.extend(workers + [merge])
        prev = f"input_{layer+1}"
    return Pipeline(layers)

pipeline = build_layered_pipeline()
pipeline.describe()   # or any CLI command that touches registry
```

Running this on a high-performance server takes tens of minutes (not just a few) and the stack trace is dominated by:

```
dfs (kedro/pipeline/pipeline.py:328)
...
```

### Why it slows down
For the structure above the number of distinct DFS paths equals `N^M`.  
With the realistic values from our project (`N=9`, `M=8`) this is **43 046 721** paths.  
`_validate_namespaces`:

1. Recursively explores the full path space without memoisation.
[...]

PR #4945, opened 2025-07-06
Optimise the namespace validation

idanov:
## Description

After adding validation for namespaces in https://github.com/kedro-org/kedro/pull/4603, Kedro was becoming prohibitively slow for any pipelines that had even slightly more complex pipelines that had many paths. The namespace validation was based on going through all paths, which could be exponentially many. This nicely explained issue has an example when the namespace validation just hangs https://github.com/kedro-org/kedro/issues/4943

I've added the example from the issue as a test.

## Development notes

I've reimplemented the algorithm to use the topologically sorted nodes and keep track of the visited pipeline at each node level. This way we go from `O(P)` down to `O(E + V)`, where `P` is the number of paths, `E` is the number of edges and `V` is the number of nodes. This is a very significant improvement, as `P` can be exponential in the number of nodes. When a violation of the namespace interruption occurs, then we use a `BFS` algorithm to find the shortest path starting from a node that exited the violated namespace. Bear in mind that there might be multiple paths that violate that, but only one will be reported (unlike in the previous algorithm), but that should be sufficient for debugging. 

## Checklist

### DIFF
--- a/kedro/pipeline/pipeline.py
+++ b/kedro/pipeline/pipeline.py
@@ -9,7 +9,7 @@ from __future__ import annotations
 import copy
 import difflib
 import json
-from collections import Counter, defaultdict
+from collections import Counter, defaultdict, deque
 from functools import cached_property
 from graphlib import CycleError, TopologicalSorter
 from itertools import chain
@@ -290,54 +290,69 @@ class Pipeline:
 
         self._toposorted_nodes: list[Node] = []
         self._toposorted_groups: list[list[Node]] = []
-        self._validate_namespaces(self.node_dependencies)
+        self._validate_namespaces()
 
-    def _validate_namespaces(self, node_parents: dict[Node, set[Node]]) -> None:
-        from warnings import warn
-
-        visited: dict[str, int] = dict()
-        path: list[str] = []
+    @cached_property
+    def _node_children(self) -> dict[Node, set[Node]]:
         node_children = defaultdict(set)
-        for child, parents in node_parents.items():
+        for child, parents in self.node_dependencies.items():
             for parent in parents:
                 node_children[parent].add(child)
+        return node_children
 
-        def dfs(n: Node, last_namespace: str) -> None:
-            curr_namespace = n.namespace or ""
-            if curr_namespace and curr_namespace in visited:
-                warn(
-                    f"Namespace '{curr_namespace}' is interrupted by nodes {path[visited[curr_namespace]:]} and thus invalid.",
-                    UserWarning,
-                )
+    def _shortest_path(self, start: Node, end: Node) -> list[Node]:
+        queue = deque([(start, [start])])
+        visited = set([start])
 
-            # If the current namespace is different from the last namespace and isn't a child namespace,
-            # mark the last namespace and all unrelated parent namespaces as visited to detect potential future interruptions
-            backtracked: dict[str, int] = dict()
-            if (
-                last_namespace
-                and curr_namespace != last_namespace
-                and not curr_namespace.startswith(last_namespace + ".")
-            ):
-                parts = last_namespace.split(".")
-                prefix = ""
-                for p in parts:
-                    prefix += p
-                    if not curr_namespace.startswith(prefix):
-                        backtracked[prefix] = len(path)
-                    prefix += "."
+        while queue:
+            node, path = queue.popleft()
+            if node == end:
+                return path
+            for child in self._node_children[node]:
+                if child not in visited:
+                    visited.add(child)
+                    queue.append((child, [*path, child]))
+        return []
 
-                visited.update(backtracked)
+    def _validate_namespaces(self) -> None:
+        from warnings import warn
 
-            path.append(n.name)
-            for child in node_children[n]:
-                dfs(child, n.namespace or "")
-            path.pop()
-            for key in backtracked.keys():
-                visited.pop(key, None)
+        node_parents: dict[Node, set[Node]] = self.node_dependencies
 
-        start = (n for n in node_parents if not node_parents[n])
-        for n in start:
-            dfs(n, "")
+        # TODO: See if this can become a persistent trie-like datastructure
+        seen_namespaces: dict[Node, dict[str, set[Node]]] = defaultdict(dict)
+        for node in self.nodes:
+            visited: dict[str, set[Node]] = defaultdict(set)
+            for parent in node_parents[node]:
+                last_namespace = parent.namespace or ""
+                curr_namespace = node.namespace or ""
+                # If curr_namespace was visited in paths leading to parent
+                if curr_namespace and curr_namespace in seen_namespaces[parent]:
+                    start = next(iter(seen_namespaces[parent][curr_namespace]))
+                    path = [n.name for n in self._shortest_path(start, node)]
+                    warn(
+                        f"Namespace '{curr_namespace}' is interrupted by nodes {path[:-1]} and thus invalid.",
+                        UserWarning,
+                    )
+                # All visited namespaces for the current node get updated with the parent's visited namespaces
+                for ns, nodes in seen_namespaces[parent].items():
+                    visited[ns].update(nodes)
+                # If the current namespace is different from the last namespace and isn't a child namespace,
+                # mark the last namespace and all unrelated parent namespaces as visited to detect potential future interruptions
+                if (
+                    last_namespace
+                    and curr_namespace != last_namespace
+                    and not curr_namespace.startswith(last_namespace + ".")
+                ):
+                    parts = last_namespace.split(".")
+                    prefix = ""
+                    for p in parts:
+                        prefix += p
+                        if not curr_namespace.startswith(prefix):
+                            visited[prefix].add(node)
+                        prefix += "."
+            # Now we have created the visited namespaces for the current node
+            seen_namespaces[node] = visited
 
     def __repr__(self) -> str:  # pragma: no cover
         """Pipeline ([node1, ..., node10 ...], name='pipeline_name')"""
--- a/kedro/pipeline/pipeline.py
+++ b/kedro/pipeline/pipeline.py
@@ -312,14 +312,14 @@ class Pipeline:
                 if child not in visited:
                     visited.add(child)
                     queue.append((child, [*path, child]))
-        return []
+        return []  # pragma: no cover
 
     def _validate_namespaces(self) -> None:
         from warnings import warn
 
         node_parents: dict[Node, set[Node]] = self.node_dependencies
 
-        # TODO: See if this can become a persistent trie-like datastructure
+        # Here we keep all visited namespaces for each node
         seen_namespaces: dict[Node, dict[str, set[Node]]] = defaultdict(dict)
         for node in self.nodes:
             visited: dict[str, set[Node]] = defaultdict(set)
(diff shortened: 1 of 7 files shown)


kedro/pipeline/pipeline.py
```diff
@@ -290,54 +290,69 @@ def __init__(  # noqa: PLR0913
...
+            if node == end:
+                return path
+            for child in self._node_children[node]:
+                if child not in visited:
+                    visited.add(child)
+                    queue.append((child, [*path, child]))
+        return []  # pragma: no cover
+
+    def _validate_namespaces(self) -> None:
+        from warnings import warn
```
merelcht: Why is this import here and not at the top of the file?

kedro/pipeline/pipeline.py
```diff
@@ -290,54 +290,69 @@ def __init__(  # noqa: PLR0913
...
+    @cached_property
+    def _node_children(self) -> dict[Node, set[Node]]:
         node_children = defaultdict(set)
-        for child, parents in node_parents.items():
+        for child, parents in self.node_dependencies.items():
             for parent in parents:
                 node_children[parent].add(child)
+        return node_children
+
+    def _shortest_path(self, start: Node, end: Node) -> list[Node]:
```
merelcht, line 305: Can you add a short comment on what this does, just for future reference?

tests/pipeline/test_pipeline_namespaces.py
```diff
@@ -45,6 +66,14 @@ def test_pipeline_with_continuous_namespace():
         warnings.simplefilter("error")  # Convert warnings to exceptions
         Pipeline(nodes)
 
+    def test_big_pipeline():
+        # Create a big pipeline with many layers of parallel paths
+
+        # No warning should be raised
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")  # Convert warnings to exceptions
+            create_a_big_pipeline()
```
merelcht: Is this indented on purpose?

kedro/pipeline/pipeline.py
```diff
@@ -290,54 +290,69 @@ def __init__(  # noqa: PLR0913
...
-            dfs(n, "")
+        # Here we keep all visited namespaces for each node
+        seen_namespaces: dict[Node, dict[str, set[Node]]] = defaultdict(dict)
+        for node in self.nodes:
+            visited: dict[str, set[Node]] = defaultdict(set)
+            for parent in node_parents[node]:
+                last_namespace = parent.namespace or ""
+                curr_namespace = node.namespace or ""
+                # If curr_namespace was visited in paths leading to parent
+                if curr_namespace and curr_namespace in seen_namespaces[parent]:
```
merelcht: Can you assign `seen_namespaces[parent]` to a variable for readability? I think it will help with especially the line below.

kedro/pipeline/pipeline.py
```diff
@@ -290,54 +290,69 @@ def __init__(  # noqa: PLR0913
...
-                dfs(child, n.namespace or "")
-            path.pop()
-            for key in backtracked.keys():
-                visited.pop(key, None)
-
-        start = (n for n in node_parents if not node_parents[n])
-        for n in start:
-            dfs(n, "")
+        # Here we keep all visited namespaces for each node
+        seen_namespaces: dict[Node, dict[str, set[Node]]] = defaultdict(dict)
```
DimedS, line 326: It would be great if the purpose of that dictionary could be described more clearly. It's not immediately obvious what role the `set[Node]` plays. 

As I understand it, this will be a set of nodes that are seen first during traversal after the current namespace ends.

kedro/pipeline/pipeline.py
```diff
@@ -290,54 +290,69 @@ def __init__(  # noqa: PLR0913
...
+                if curr_namespace and curr_namespace in seen_namespaces[parent]:
+                    start = next(iter(seen_namespaces[parent][curr_namespace]))
+                    path = [n.name for n in self._shortest_path(start, node)]
+                    warn(
+                        f"Namespace '{curr_namespace}' is interrupted by nodes {path[:-1]} and thus invalid.",
+                        UserWarning,
+                    )
+                # All visited namespaces for the current node get updated with the parent's visited namespaces
+                for ns, nodes in seen_namespaces[parent].items():
+                    visited[ns].update(nodes)
```
DimedS, line 350: I believe the complexity of these lines is quadratic in the worst-case scenario. For example, in a linear chain like Node1 → Node2 → Node3 → ..., where each node has a separate namespace, the merging of namespace data accumulates linearly at each step, resulting in quadratic growth overall.

ankatiyar:
I'm getting two warnings when I try it out with interrupted namespaces:
```
[07/09/25 14:24:04] WARNING  /Users/ankita_katiyar/kedro/kedro/kedro/pipeline/pipeline.py warnings.py:110
                             :341: UserWarning: Namespace 'data_science' is interrupted                  
                             by nodes ['create_model_input_table_node'] and thus invalid.                
                               warn(                                                                     
                                                                                                         
                    WARNING  /Users/ankita_katiyar/kedro/kedro/kedro/pipeline/pipeline.py warnings.py:110
                             :341: UserWarning: Namespace 'data_science' is interrupted                  
                             by nodes ['create_model_input_table_node',                                  
                             'data_science.split_data_node'] and thus invalid.                           
                               warn(  
```
In my spaceflights project the `preprocess_companies` is namespaced `data_science` as well as the whole `data_science` pipeline has the same namespace. So the first warning is correct, but the second one isn't.

The algorithm largely makes sense!
I'm also surprised by how many times it gets triggered because it's in the `Pipeline` constructor which gets called several times, the warnings only show up once though.

kedro/pipeline/pipeline.py
idanov: That's how it was in the previous code, can pull it out to the top.

kedro/pipeline/pipeline.py
idanov: I think it's pretty clear from the name, but sure - happy to add a comment.

kedro/pipeline/pipeline.py
idanov: That's a good point, will add it and it might make it more readable.

kedro/pipeline/pipeline.py
idanov: Good point, will add a comment!

kedro/pipeline/pipeline.py
idanov: Not sure if that's the case, because these are all namespaces that were already visited. So technically, yes - in the worst case if all is added. That's why I was considering a persistent data structure, but in practice I doubt it will be such a bad time, e.g. 1000 nodes in one line, each in completely different namespaces (very unlikely scenario), that will result it about half a million iterations, which is definitely under 1 second execution time. Future improvements can try eliminating this, but honestly, the worst case is very artificially achieved. The reality will be more like $O(n * log n)$ as the depth of the pipeline will tend to approximate $log n$, even more so that the number of namespaces is going to be an order of magnitude smaller than the number of nodes.

tests/pipeline/test_pipeline_namespaces.py
idanov: Yes, the point of the test is to capture the warnings and convert them into errors instead, so the test will fail if any warnings are generated.

kedro/pipeline/pipeline.py
idanov: Now when I think about it, adding only one node per namespace should be sufficient, as we only use the first one when reporting the path in case of interruption. Let's leave these comments in the PR for posterity in case we see slowdowns here in the future, I don't want to apply this optimisation because I don't want to risk correctness (as I haven't thought out very well if this might have implication on the correctness of the solution at the end).

idanov:
> I'm getting two warnings when I try it out with interrupted namespaces:
> 
> ```
> [07/09/25 14:24:04] WARNING  /Users/ankita_katiyar/kedro/kedro/kedro/pipeline/pipeline.py warnings.py:110
>                              :341: UserWarning: Namespace 'data_science' is interrupted                  
>                              by nodes ['create_model_input_table_node'] and thus invalid.                
>                                warn(                                                                     
>                                                                                                          
>                     WARNING  /Users/ankita_katiyar/kedro/kedro/kedro/pipeline/pipeline.py warnings.py:110
>                              :341: UserWarning: Namespace 'data_science' is interrupted                  
>                              by nodes ['create_model_input_table_node',                                  
>                              'data_science.split_data_node'] and thus invalid.                           
>                                warn(  
> ```
> 
> In my spaceflights project the `preprocess_companies` is namespaced `data_science` as well as the whole `data_science` pipeline has the same namespace. So the first warning is correct, but the second one isn't.
> 
> The algorithm largely makes sense! I'm also surprised by how many times it gets triggered because it's in the `Pipeline` constructor which gets called several times, the warnings only show up once though.

Interesting, could you please share your pipeline, as it is quite difficult for me to reproduce your problem without your pipeline...

ankatiyar:
This is what the pipelines look like in a spaceflights project
`data_processing`
```python
    return pipeline(
        [
            node(
                func=preprocess_companies,
                inputs="companies",
                name="preprocess_companies_node",
                namespace="data_science",
            ),
            node(
                func=preprocess_shuttles,
                inputs="shuttles",
                outputs="preprocessed_shuttles",
                name="preprocess_shuttles_node",
             ),
            node(
                func=create_model_input_table,
                inputs=["preprocessed_shuttles", "preprocessed_companies", "reviews"],
                outputs="model_input_table",
                name="create_model_input_table_node",           ),            
        ],
    )
```


`data_science`
```python
    return pipeline(
        [
            
            node(
                func=split_data,
                inputs=["model_input_table", "params:model_options"],
                outputs=["X_train", "X_test", "y_train", "y_test"],
                name="split_data_node",
            ),
            node(
                func=train_model,
                inputs=["X_train", "y_train"],
                outputs="regressor",
                name="train_model_node",
            ),
            node(
                func=evaluate_model,
                inputs=["regressor", "X_test", "y_test"],
                outputs=None,
                name="evaluate_model_node",
            ),
        ],
        namespace="data_science",
        inputs={"model_input_table": "model_input_table"},
        parameters={"model_options": "params:model_options"},
    )

File: kedro/pipeline/pipeline.py [hunk 7]
```diff
@@ -290,54 +291,71 @@ def __init__(  # noqa: PLR0913
 
         self._toposorted_nodes: list[Node] = []
         self._toposorted_groups: list[list[Node]] = []
-        self._validate_namespaces(node_parents)
+        self._validate_namespaces()
```
g199209: @idanov As I mentioned in #4943, I think we can skip this validation if there's no node namespace at all? i.e.:

```python
if any(n.namespace for n in self._nodes):
    self._validate_namespaces()
```

tests/pipeline/test_pipeline_namespaces.py
merelcht: I mean the whole test is indented and so not run as part of the test suite.

kedro/pipeline/pipeline.py
idanov: That won't make a huge difference, but I guess we can add it.

idanov:
I have fixed the issue discovered by @ankatiyar and changed the behaviour to report only the first interrupted namespace level (as the previous behaviour was potentially producing too many warnings). I've corrected one of the tests and added a new property to the `Node` class, as this is a property that @rashidakanchwala might find useful in Kedro Viz (as there's currently already an implementation of this called `exlode_namespaces`).

merelcht (approved):
Tested this against @ankatiyar's example, and it's working as expected 👍 Thanks @idanov ! ⭐

ankatiyar (approved):
Manually tested with different cases, works perfectly! 💯

Issue #4909, opened 2025-06-26
Warning about dot in dataset name contradicts dataset factory docs


merelcht:
## Description
In the docs about dataset factories we use the dot notation in examples: https://docs.kedro.org/en/1.0.0rc1/catalog-data/kedro_dataset_factories/#how-to-generalise-datasets-of-the-same-type-in-different-layers

But now in the release candidate you see: 
```
UserWarning: Dataset name '01_raw.shuttles' contains '.' characters, which is not recommended as   warnings.py:110
                             the dot notation is reserved for automatic namespacing in Kedro. Consider using a different naming convention.
```
when you use the dot. 

I think this is confusing and we should update our docs to give examples that don't produce warnings.

DimedS:
Thanks, @merelcht. I think these are a bit tricky questions.

In the given example, the warning won’t be triggered if two namespaced pipelines (or nodes) are associated with the same namespaced datasets. So in my view, the use of a dot (`"."`) in the example actually reflects our recommendation on how to properly structure pipelines using namespaces. Technically, we could replace the dot with an underscore (`"_"`), but I’m not sure it would add any real benefit.

My main concern is with the warning itself - it’s not ideal, as it can be triggered even in valid scenarios. For example, the warning may appear when one namespaced pipeline takes the output of another as input, which is a perfectly valid situation. That’s why in [#4797](https://github.com/kedro-org/kedro/pull/4797), we changed it from an error to a warning - to simply inform users that dots should be used only in namespaces.

Would be glad to hear any additional opinions, @ankatiyar

ankatiyar:
I'm surprised because the warning should only be triggered when the dataset names in the pipeline inputs or outputs contain dots, i.e.
```
Pipeline(
  ...
inputs = "namespace.x",
)
```
Because, the namespace should automatically be prefixed by Kedro. But it is expected to define namespaced datasets in the `catalog.yml` like this anyway, with or without dataset factories.

merelcht:
> I'm surprised because the warning should only be triggered when the dataset names in the pipeline inputs or outputs contain dots, i.e.
> 
> ```
> Pipeline(
>   ...
> inputs = "namespace.x",
> )
> ```
> 
> Because, the namespace should automatically be prefixed by Kedro. But it is expected to define namespaced datasets in the `catalog.yml` like this anyway, with or without dataset factories.

Well yes the issue is that when you use the dataset factory in the catalog, e.g. like this for spaceflights:

```yml
[...]

Issue #4223, opened 2024-10-11
Placeholders for catalog.yml not working properly when value contains underscore


Acomar0:
## Description
Placeholders for catalog.yml not working properly when value contains underscore

## Context
I am encountering some issues regarding the use of placeholders for the data catalog and  I was hoping you can shed some light on this .
I have the following pipeline:
```
load_date = settings.LOAD_DATE_COMPARISON.get("current")
previous_load_date = settings.LOAD_DATE_COMPARISON.get("previous")

def create_pipeline(**kwargs) -> Pipeline:


    format_data_quality = pipeline(
                [   node(
                        func= compare_id,
                        inputs=[f"maestro_indicadores_{load_date}",
                                f"maestro_indicadores_{previous_load_date}"],
                        outputs=f"compare_id_{load_date}_{previous_load_date}",
                        name="compare_id_node",
                        tags = "compare_id"
    ),]
    )
    return format_data_quality
```
With the corresponding catalog entry for the output:
```
compare_id_{load_date}_{previous_load_date}:
  type: json.JSONDataset
  filepath: reports/{load_date}/id_comparison/id_comparison_{load_date}_{previous_load_date}.json
```
The issue here is that whenever the value of load date is something like 2024_07_01, it will generate a path like:
reports/**2024**/id_comparison/id_comparison_ 2024_07_01_2024_05_01.json

**Note that the first placeholder is not being substituted with the intended value, while the others are.**
This will only happen when the value of load_date contains underscores, not happening with dots or hyphens.

* Kedro version used : 0.19.8
* Python version used : 3.11.4
* Operating system and version: Win11 Pro 10.0.22.631

noklam:
Thanks for creating the issue @Acomar0, we will look into it soon.

merelcht:
Hi @Acomar0, apologies for the late response. 

I've had time to dig into this issue and it's caused by how [the `parse` library](https://github.com/r1chardj0n3s/parse) we use works. This pattern `compare_id_{load_date}_{previous_load_date}` is mapped to `compare_id_ 2024_07_01_2024_05_01` correctly, but lower down it has parsed `load_date` as `2024` and `previous_load_date` as `07_01_2024_05_01`. It's looking for the smallest unit that matches the regex pattern and in this case for `load_date` and `previous_load_date` separated by `_` is `2024` and `07_01_2024_05_01` separated by `_`. You can fix this by not using `_` as separator between the dates, but another symbol like `#`.

PR #4976, opened 2025-07-16
Change dot to underscore to avoid misundestanding

DimedS:
## Description
Solves #4909

### DIFF
--- a/docs/catalog-data/kedro_dataset_factories.md
+++ b/docs/catalog-data/kedro_dataset_factories.md
@@ -288,21 +288,21 @@ You can use multiple placeholders in the same pattern. For example, consider the
 entries share `type`, `file_format` and `save_args`:
 
 ```yaml
-processing.factory_data:
+processing_factory_data:
   type: spark.SparkDataset
   filepath: data/processing/factory_data.parquet
   file_format: parquet
   save_args:
     mode: overwrite
 
-processing.process_data:
+processing_process_data:
   type: spark.SparkDataset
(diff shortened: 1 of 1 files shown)


docs/catalog-data/kedro_dataset_factories.md
```diff
@@ -288,21 +288,21 @@ You can use multiple placeholders in the same pattern. For example, consider the
 entries share `type`, `file_format` and `save_args`:
 
 ```yaml
-processing.factory_data:
+processing_factory_data:
```
ankatiyar: Oh this might be problematic because `preprocessing_factory_data` can be split multiple ways. i.e. `preprocessing` and `factory_data` or `preprocessing_factory` and `data`. which probably not work with the `parse` library we use for dataset factories. Can we try a different separator?

ankatiyar:
Is this the only example using the `.` dot notation?

DimedS:
> Is this the only example using the `.` dot notation?

previous and next examples are with namespaces, I don't think they should be fixed because if you want to use namespace you should use dot.

docs/catalog-data/kedro_dataset_factories.md
merelcht: Yes excellent point!! See https://github.com/kedro-org/kedro/issues/4223

DimedS: changed to "-", hope it works)

ankatiyar (approved):

merelcht (approved):

PR #5043, opened 2025-08-06
Use tomllib, if available, else fall back to tomli

deepyaman:
## Description
`tomli` is more standards-compliant than `toml`, and it's including in the standard library as `tomllib` since Python 3.11.

`toml` has not had a release since October 31, 2020.

## Development notes
[Built the compatibility layer, as documented.](https://github.com/hukkin/tomli?tab=readme-ov-file#building-a-tomlitomllib-compatibility-layer)

Unfortunately, `kedro-telemetry` relies on `kedro` to provide the `toml` dependency, so that needs to be updated as well.

## Checklist

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -131,6 +131,7 @@ def _install_project_requirements(context):
     )
     install_reqs = [req for req in install_reqs if "{" not in req and "#" not in req]
     install_reqs.append("kedro-datasets[pandas-csvdataset]")
-    install_reqs.append("toml")  # TODO(deepyaman): Migrate `kedro-telemetry` or add `toml` to the plugin's dependencies
+    # TODO(deepyaman): Migrate `kedro-telemetry` or add `toml` to plugin's dependencies.
+    install_reqs.append("toml")
     call([context.pip, "install", *install_reqs], env=context.env)
     return context
--- a/features/environment.py
+++ b/features/environment.py
@@ -119,18 +119,7 @@ def _setup_minimal_env(context):
             ],
             env=context.env,
         )
-        call(
-            [
-                context.python,
-                "-m",
-                "pip",
-                "install",
-                "-e",
-                ".",
-                "toml>=0.10.0",  # TODO(deepyaman): Migrate `kedro-telemetry` off `toml`
-            ],
-            env=context.env,
-        )
+        call([context.python, "-m", "pip", "install", "-e", "."], env=context.env)
         return context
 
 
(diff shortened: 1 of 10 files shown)


pyproject.toml
```diff
@@ -30,7 +30,8 @@ dependencies = [
     "pluggy>=1.0",
     "PyYAML>=4.2,<7.0",
     "rich>=12.0,<15.0",
-    "toml>=0.10.0",
+    "tomli>=1.1.0;python_version<'3.11'",  # tomllib is part of the standard library in Python 3.11+
+    "tomli-w",
```
merelcht, line 34: Why do you need the separate `tomli-w` dependency?

merelcht (commented):
The changes make sense, just had a question on why two dependencies are needed.

pyproject.toml
deepyaman: `tomli` (and the `tomllib` included in the Python standard library from 3.11) are read-only, and `tomli-w` is the recommended writing library (same author).

merelcht: Got it! Shame they didn't put it in the same library 🥲

merelcht (approved):
Thanks @deepyaman !

features/environment.py
```diff
@@ -119,7 +119,18 @@ def _setup_minimal_env(context):
...
-        call([context.python, "-m", "pip", "install", "-e", "."], env=context.env)
+        call(
+            [
+                context.python,
+                "-m",
+                "pip",
+                "install",
+                "-e",
+                ".",
+                "toml>=0.10.0",  # TODO(deepyaman): Migrate `kedro-telemetry` off `toml`
```
ankatiyar: This still needed after https://github.com/kedro-org/kedro-plugins/pull/1155 is merged?

deepyaman: It shouldn't be necessary once `kedro-telemetry` is released?

I do think `kedro-telemetry` should be released first regardless. I'm not sure if there's a specific consideration around using a newer version of Kedro (that uses `tomli` or `tomllib` instead of `toml`) with an older version of `kedro-telemetry` (that depended on the implicit `toml` dependency coming from Kedro), but I think this could just be documented/it would be pretty obvious to install `toml` (or upgrade `kedro-telemetry`).

features/environment.py
```diff
@@ -119,7 +119,18 @@ def _setup_minimal_env(context):
...
+            ],
+            env=context.env,
+        )
```
deepyaman:
```suggestion
        call([context.python, "-m", "pip", "install", "-e", "."], env=context.env)
```

features/environment.py
deepyaman: @ankatiyar I was able to remove it with the `kedro-telemetry` 0.6.4 release!

Issue #4818, opened 2025-06-10
Annoying DeprecationWarning


g199209:
DeprecationWarning: 'kedro run' flag '--namespace' is deprecated and will be replaced with '--namespaces' from Kedro 1.0.0 to allow running multiple namespaces.

This warning is stupid:

1. There's no `--namespaces` flag now, I can only use `--namespace`. I think we should support both new and old flags in transitional period, and show deprecation warning. But there's no new flag at all now!

2. I only use `kedro run` in command line, no `--namespace` or `--ns` flag at all, but still see this warning. I havn't find the reason why now.

cramirez98:
Im having the same issue and im not using namespaces (just a plain kedro run --nodes=node).

yury-fedotov:
+1, I find the "DataCatalog will be replaced by KedroDataCatalog" warning a bit noisy in the CLI output too, given that I don't use DataCatalog object explicitly in the Python API at all.

ravi-kumar-pilla:
Hi, You can definitely reduce some noise using `export PYTHONWARNINGS="ignore::DeprecationWarning:kedro"`  in your run environment or you can also adjust the log level in logging.yml.

However, since this is coming from Kedro core, it is somewhat unavoidable at the moment. 

@ankatiyar are you aware of any better alternatives to reduce notifications mentioned above ? Thank you

ankatiyar:
@ravi-kumar-pilla not off the top of my head apart from what you said!

@g199209 @cramirez98 @yury-fedotov The 1.0 release will be out soon, but the 1.0 pre-release is already out that you could start using :) https://pypi.org/project/kedro/1.0.0rc1/

ravi-kumar-pilla:
Closing the issue as the workaround to get rid of the warnings is to upgrade Kedro version to 1.0.0rc1.

g199209:
> [@ravi-kumar-pilla](https://github.com/ravi-kumar-pilla) not off the top of my head apart from what you said!
> 
> [@g199209](https://github.com/g199209) [@cramirez98](https://github.com/cramirez98) [@yury-fedotov](https://github.com/yury-fedotov) The 1.0 release will be out soon, but the 1.0 pre-release is already out that you could start using :) https://pypi.org/project/kedro/1.0.0rc1/

@ankatiyar Upgrade to 1.0.0rc1 doesn't solve this problem, still have this warning

merelcht:
[...]

PR #4728, opened 2025-05-13
Amend catalog deprecation warning

ElenaKhaustova:
## Description
Solves https://github.com/kedro-org/kedro/issues/4721

## Development notes

1. Now both catalogs have deprecation warnings
2. Users utilizing `KedroDataCatalog` will see the warning all the time
3. Users utilizing `DataCatalog` within framework will not see the warning - it is suppressed
4. Users utilizing `DataCatalog` directly via importing to their codebase and initializing it - will see the warning

## Developer Certificate of Origin
We need all contributions to comply with the [Developer Certificate of Origin (DCO)](https://developercertificate.org/). All commits must be signed off by including a `Signed-off-by` line in the commit message. [See our wiki for guidance](https://github.com/kedro-org/kedro/wiki/Guidelines-for-contributing-developers/).

If your PR is blocked due to unsigned commits, then you must follow the instructions under "Rebase the branch" on the GitHub Checks page for your PR. This will retroactively add the sign-off to all unsigned commits and allow the DCO check to pass.

## Checklist

- [ ] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [ ] Signed off each commit with a [Developer Certificate of Origin (DCO)](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository)
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [ ] Added tests to cover my changes
- [ ] Checked if this change will affect Kedro-Viz, and if so, communicated that with the Viz team

PR #5062, opened 2025-08-25
Release/0.19.15

DimedS:
## Description

This PR introduces the **0.19.15 release** of Kedro.  
Although Kedro 1.0.0 has already been released, this patch release addresses two issues for users still on the 0.19 series:

- https://github.com/kedro-org/kedro/issues/4818  
- https://github.com/kedro-org/kedro/pull/4945  

---

### Branching strategy

Two dedicated branches were created for this backport release:

- [`base-0.19.14`](https://github.com/kedro-org/kedro/tree/base-0.19.14): created from the `0.19.14` tag.  
- [`release/0.19.15`](https://github.com/kedro-org/kedro/tree/release/0.19.15): contains the changes for 0.19.15.  
  - Removed the deprecation warning.  
  - Cherry-picked commits for namespace validation.  

This PR merges `release/0.19.15` into `base-0.19.14`.

---

### Workflow changes

- **`.github/workflows/check-release.yml`** was updated to trigger the release workflow when this PR is merged into `base-0.19.14`.  
- **`.github/workflows/release-starters.yml`** was modified so that it does **not** trigger a `kedro-starters` release automatically. The `kedro-starters` release will be triggered manually later.

---

### Additional CI/CD modifications

To ensure all checks pass, the following adjustments were made:

- Removed links to the `main` branch in unit tests referencing `kedro-starters`.  
- Used a fresh virtual environment in some e2e tests.  
- Updated the Makefile for `pip-compile`.  
- Fixed multiple documentation links to ensure successful RTD builds.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -44,6 +44,6 @@ authors:
 - family-names: Brugman
   given-names: Simon
 title: Kedro
-version: 0.19.14
-date-released: 2025-06-17
+version: 0.19.15
+date-released: 2025-09-01
 url: https://github.com/kedro-org/kedro
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,12 +2,16 @@
 
 ## Major features and improvements
 ## Bug fixes and other changes
-* Improved namespace validation efficiency to prevent significant slowdowns when creating large pipelines
-
 ## Breaking changes to the API
 ## Upcoming deprecations for Kedro 1.0.0
 ## Documentation changes
 ## Community contributions
+
+# Release 0.19.15
+
+## Bug fixes and other changes
+* Improved namespace validation efficiency to prevent significant slowdowns when creating large pipelines
+* Made deprecation warning appear only when the deprecated `--namespace` flag is used
 
 # Release 0.19.14
 
--- a/docs/source/development/commands_reference.md
+++ b/docs/source/development/commands_reference.md
@@ -109,7 +109,7 @@ Returns output similar to the following, depending on the version of Kedro used 
 | |/ / _ \/ _` | '__/ _ \
 |   <  __/ (_| | | | (_) |
 |_|\_\___|\__,_|_|  \___/
-v0.19.14
+v0.19.15
 
 Kedro is a Python framework for
 creating reproducible, maintainable
--- a/kedro/__init__.py
+++ b/kedro/__init__.py
@@ -13,7 +13,7 @@ if TYPE_CHECKING:
 import sys
 import warnings
 
-__version__ = "0.19.14"
+__version__ = "0.19.15"
 
 
 class KedroDeprecationWarning(DeprecationWarning):


ankatiyar:
I'd also suggest renaming the base branch to indicate that it'll hold the tag for 0.19.15 - this will avoid confusion in the future

ElenaKhaustova:
> This mostly looks good, but now there's no deprecation warning at all for the namespaces flag. We should still have one, but it should only show when a users uses the flag and not all the time they do `kedro run`.

I did a similar thing for `DataCatalog`: https://github.com/kedro-org/kedro/pull/4728

You can keep the warning, but add logic to suppress it if the flag is not used.

DimedS:
Thank you for the review and comments, @merelcht, @ankatiyar, and @ElenaKhaustova!
I’ve addressed all your suggestions and agree with the points raised.

I decided to go with a simpler approach for notifications - they’ll now be triggered whenever the `--namespace` option is provided.

Requesting a final re-review 🙏

Also, I noticed that the docs didn’t build, but no error was shown in the logs. Has anyone seen this before or might know what’s causing it?

kedro/framework/cli/cli.py
```diff
@@ -225,7 +225,7 @@ def project_groups(self) -> Sequence[click.MultiCommand]:
         project and the plugins, then combines them with the built-in ones.
         Built-in commands can be overridden by plugins, which can be
         overridden by a custom project cli.py.
-        See https://docs.kedro.org/en/stable/extend_kedro/common_use_cases.html#use-case-3-how-to-add-or-modify-cli-commands
+        See https://docs.kedro.org/en/0.19.14/extend_kedro/common_use_cases.html#use-case-3-how-to-add-or-modify-cli-commands
```
ankatiyar, line 228: All these docs link point to 0.19.14 - is there a reason?

ankatiyar: Oh maybe because stable will keep pointing to 1.0.0 and 0.19.15 isnt published so builds fail?

.github/workflows/check-release.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - base-0.19.15
```
ankatiyar, line 7: Worth double checking if the release will be triggered still because this might need to be added to the `main` version of the release

ankatiyar (approved):
Thanks @DimedS

.github/workflows/check-release.yml
DimedS: could you comment more? I think it should work

kedro/framework/cli/cli.py
DimedS: I couldn’t use stable because the redirection we recently implemented causes the link check to fail. I also couldn’t use 0.19.15 since it hasn’t been released yet, and 1.0.0 belongs to the next major version. So I decided to proceed with this option.

.github/workflows/check-release.yml
ankatiyar: When I was implementing nightly checks on the `develop` branch, the scheduled tasks would use the workflow from `main` branch. But I suppose if `all-checks` is working then this should too!

ElenaKhaustova (approved):
Thank you, @DimedS!

merelcht (approved):
The deprecation warning works well now 👍

merelcht:
@DimedS Regarding the failed docs build, did you try to re-run it already? Might just be a glitch.

Issue #2465, opened 2023-03-24
Should OmegaConfigLoader Support Cross-Environment Interpolation? Investigation Required


noklam:
# Context:
Parent #2458 An user report that cross-environment interpolation is allowed for `OmegaConfigLoader`.

# Description
I think it shouldn't be allowed as this can become confusing with the use of `globals`, user then have to check many files to confirm the source of parameters - and it's hard to reason about the resolution order.

# Possible outcome
1. This is fine, we should enable
2. This isn't something we want, figure out what's need to be done to prevent this

              @merelcht: ok but why would you return a `dict` in `load_and_merge_dir_config` instead of a DictConfig object? I think that if we could keep the config data format internally a `DictConfig` object and only cast it to a dict when returning from `__getitem__` (it does then resolve variable interpolation etc). Oh and we should use `OmegaConf.to_container()` to cast it because it then nicely resolve variables etc. (unless you set `resolve=False`).

About interpolation: do we want to allow interpolation *across* envs? E.g. 
``` 
--- 
# conf/base/parameters.yaml
server:
  host: localhost
  port: 80

--- 
# conf/local/parameters.yaml
client:
  url: http://${server.host}:${server.port}/
  server_port: ${server.port}
``` 
Because that's we currently (want to) do (I think)

_Originally posted by @MatthiasRoels in https://github.com/kedro-org/kedro/issues/2458#issuecomment-1481780423_

merelcht:
We need to verify what the current behaviour is. I would expect that cross-environment interpolation doesn't work. It also doesn't seem like desirable behaviour as it would be hard to reason about how and where the values of the templates come from.

MatthiasRoels:
I agree with the fact that it could be hard to reason about where values come from. But on the other hand, allowing cross-env interpolation would mimic what the `globals` currently do in the `TemplatedConfigLoader` class right? Only in this version, you would need no `globals` pattern! One drawback of this off course is that you can only use variable interpolation within the same pattern (which differs from the `globals` pattern in behaviour in `TemplatedConfigLoader`).

noklam:
I have done some testing and cross-env interpolation does not work currently @MatthiasRoels (expected as we never intend to allow it)

You are correct that with `TemplatedConfigLoader` the global scope is for all pattern, but this would only affect `parameters`.

All of options has some overlap area and we need to think what's the best way. One thing we don't want to see is that people start abusing global everywhere.

[...]

PR #5066, opened 2025-08-26
Add OmegaConfigLoader._get_unresolved_config method which return dictconfig instead of fully resolved configuration

### LINKED ISSUE #2973: Make ConfigLoader return DictConfig instead of dict only
## Description

As a plugin developer, I have several use cases for which I want to have control over configuration at runtime. With the new ``OmegaConfigLoader``, configuration is available in ``context.config_loader["catalog"]``, which returns a dictionary with all key already resolved. 

I'd like to access the "raw" configuration (with interpolated variables not resolved yet) and not the fully resolved ones. I could reload the yaml file, but kedro performs hidden operations to merge environments in [``load_and_merge_dir_config``](https://github.com/kedro-org/kedro/blob/e2bb4cd5cac9e25248046f2be2015f734b8d8d37/kedro/config/omegaconf_config.py#L218-L304) that I don't want to reproduce because  it is hard to maintain and error prone. 

## Context

 One use case I have  is to identify "globals" variables use for the specific pipeline to log them in mlflow (I don't want to log all the "globals.yml" file to avoid polluting my mlflow with globals unrelated to my pipeline). In order to do this, I'd like to check in the "raw" configuration which key will be resolved with the globals resolver 

``OmegaConf`` stores configuration in a [special object called ``DictConfig``](https://omegaconf.readthedocs.io/en/2.3_branch/usage.html#id36). This object has several helpers methods, particularly to help to traverse the tree structure, to check if a key is interpolated... I'd like to access this object directly to benefits from : 
- all the kedro's logic from [``load_and_merge_dir_config``](https://github.com/kedro-org/kedro/blob/e2bb4cd5cac9e25248046f2be2015f734b8d8d37/kedro/config/omegaconf_config.py#L218-L304)
- all the OmegaConf helpers to handle this configuration

## Possible Implementation

The ``DictConfig`` object is propagated all way long in ``load_and_merge_dir_config``, and it is converted back to a dictionary [in the return statement with the ``to_container``](https://github.com/kedro-org/kedro/blob/e2bb4cd5cac9e25248046f2be2015f734b8d8d37/kedro/config/omegaconf_config.py#L299). It is converted [on the fly in the ``__getitem__`` method](https://github.com/kedro-org/kedro/blob/e2bb4cd5cac9e25248046f2be2015f734b8d8d37/kedro/config/omegaconf_config.py#L143). 

I think a potential implementation could look like: 
- Add a ``load_dict_config`` (name TBD) methods to ``OmegaConfigLoader`` that basically implements the current ``__getitem__`` method without converting config back to a container
- simplify ``__getitem__`` to only perform the conversion, e.g. something like: 

```python
def __getitem__(self, key):
    OmegaConf.to_container(self.load_dict_config(key), resolve=True)
```

Is this something you might consider of interest, or ever already considered ?

Galileo-Galilei:
## Description

The goal is to **make the underlying DictConfig** returned by omegaconf accessible to developers as described in #2973 (see rational in the issue: it opens a bunch of use cases). The idea is to have a way to access the unresolved config directly from the catalog. With OmegaConfigLoader, you can do ``config["catalog"]`` or ``config.get("catalog")`` to access the resolved catalog configuration.

This PR adds a ``config._get_unresolved_config("catalog")`` methods which returns the omegaconf ``DictConfig`` object. 

## Development notes

This is almost only refactoring to postpone the resolution of config as late as possible (basically, keep the DictConfig object until __getitem__ is called). 

Steps: 
- ``__getitem__`` is only a wrapper of two methods: 
    - ``_get_unresolved_config`` which get the ``DictConfig`` object. It itself calls the refactored ``_load_and_merge_dir_config``which is almost identical to the existing ``load_and_merge_dir_config``, except that the conversion to a dict is not done at the end of the function
    - _convert_config_to_dict which handles the conversion through ``to_container`` calls
I have updated ``OmegaConfigLoader.load_and_merge_dir_config`` to split it into 2 methods: 
- ``_load_and_merge_dir_config_as_dictconfig`` which contains almost all the logic of the original method but returns a DictConfig object (not resolved to a container yet)
- ``load_and_merge_dir_config`` which now call ``_load_and_merge_dir_config_as_dictconfig`` and ``Omegaconf.to_container`` to convert back this ``DictConfig`` object to a python ``dict``. 

#### Comments: 

- ``OmegaConfigLoader.load_and_merge_dir_config`` is no longer used, but it is kept for retrocompatiblity. it now call 
``_load_and_merge_dir_config`` and ``_convert_config_to_dict`` to avoid code duplication, but this pure refactoring. It should be removed in future version, but since it is public API I left it there for now.  
- I did not modify any of the existing logic, hence, all tests pass except 1 (see below) and the public API remains untouched. 
- I did not modify the public API because it should be a much wider change, but I think the namings  should be reworded and the ``_get_unresolved_config```should ultimately be part of the public API. This changes is a first step to give a way to developers to open new use cases which are currently blocked and measure adoption while keeping the cost for the engineering team quite low. This should ultimately be reassessed whe polishing the public API surface for kedro 2.0. 

#### :warning: To be discussed

The change is not fully backward compatible (and I don't see how I can make it fully backward compatible) because the postponing the resolution of the config after merging the different config (from the same folder, then from different environments - base and the other one chosen) leads to differents results, cf the following example: 

```yaml 
# conf/base/parameters.yml

param1: "${param2}
param2: "base"

# conf/local/parameters.yml
param2: "local"

```
leads to 
- ``{param1: "base", param2: "local"}`` with the current version of code (first resolved, then merged)
- ``{param1: "local", param2: "local"}`` with the proposed change in this PR (first merged, then resolved)

My intuition is that the proposed changed should the expected behaviour from a user point of view, but I was very surprised that there exist a test which suggest that the current behaviour is intended, see: 

https://github.com/kedro-org/kedro/blob/a578398dda3af9a35f5bc20312b34d5b2a05914c/tests/config/test_omegaconf_config.py#L736-L743

No matter what I think , this would be technically a breaking change and cannot be implemented as is. 

Can someone explain the context and *why* we choose this order of resolution ? Maybe @merelcht, @ankatiyar or @astrojuanlu who did most of the work on this topic IIRC? 

If someone has a clever idea to make this work I'd love to make this change merged. 

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -175,3 +175,6 @@ kedro/datasets/*
 # Mise
 .mise.toml
 mise.toml
+
+# uv
+uv.lock
--- a/kedro/config/omegaconf_config.py
+++ b/kedro/config/omegaconf_config.py
@@ -173,10 +173,11 @@ class OmegaConfigLoader(AbstractConfigLoader):
             self._globals = value
         super().__setitem__(key, value)
 
-    def _getitem__(self, key: str) -> dict[str, Any]:
-        return self._get_dict_config(key)
+    def __getitem__(self, key: str) -> dict[str, Any]:
+        config = self._getitem_dict_config(key)
+        return self._convert_config_to_dict(config, key)
 
-    def _get_dict_config(self, key: str) -> dict[str, Any]:  # noqa: PLR0912
+    def _getitem_dict_config(self, key: str) -> dict[str, Any]:  # noqa: PLR0912
         """Get configuration files by key, load and merge them, and
         return them in the form of a config dictionary.
 
@@ -288,7 +289,7 @@ class OmegaConfigLoader(AbstractConfigLoader):
         return KeysView(self.config_patterns)
 
     @typing.no_type_check
-    def _load_and_merge_dir_config_as_dictconfig(
+    def _load_and_merge_dir_config(
         self,
         conf_path: str,
         patterns: Iterable[str],
@@ -375,6 +376,14 @@ class OmegaConfigLoader(AbstractConfigLoader):
 
         return OmegaConf.merge(*aggregate_config)
 
+    def _convert_config_to_dict(self, config: DictConfig, key: str) -> DictConfig:
+        if key == "parameters":
+            # Merge with runtime parameters only for "parameters"
+            return OmegaConf.to_container(config, resolve=True)
+
+        config_container = OmegaConf.to_container(config, resolve=True)
+        return {k: v for k, v in config_container.items() if not k.startswith("_")}
+
     def load_and_merge_dir_config(
         self,
         conf_path: str,
@@ -382,7 +391,9 @@ class OmegaConfigLoader(AbstractConfigLoader):
         key: str,
         processed_files: set,
         read_environment_variables: bool | None = False,
-    ) -> dict[str, Any]:
+    ) -> dict[
+        str, Any
+    ]:  # TODO: remove this method in a further API polishing, it is kept for retrocompatibility
         """Load and merge configuration files from a directory.
 
         Args:
@@ -397,20 +408,11 @@ class OmegaConfigLoader(AbstractConfigLoader):
         """
         # Implementation goes here
 
-        merged_omegaconf_config = self._load_and_merge_dir_config_as_dictconfig(
+        merged_omegaconf_config = self._load_and_merge_dir_config(
             conf_path, patterns, key, processed_files, read_environment_variables
         )
 
-        if key == "parameters":
-            # Merge with runtime parameters only for "parameters"
-            return OmegaConf.to_container(merged_omegaconf_config, resolve=True)
-
-        merged_config_container = OmegaConf.to_container(
-            merged_omegaconf_config, resolve=True
-        )
-        return {
-            k: v for k, v in merged_config_container.items() if not k.startswith("_")
-        }
+        return self._convert_config_to_dict(merged_omegaconf_config, key)
 
     @staticmethod
     def _initialise_filesystem_and_protocol(
--- a/kedro/config/omegaconf_config.py
+++ b/kedro/config/omegaconf_config.py
@@ -263,7 +263,6 @@ class OmegaConfigLoader(AbstractConfigLoader):
                 )
             else:
                 raise exc
-
         resulting_config = self._merge_configs(config, env_config, key, env_path)
 
         if not processed_files and key != "globals":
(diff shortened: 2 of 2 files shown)


Galileo-Galilei:
Should I add a note in the ``RELEASE.md`` and in the documentation for a private function?

noklam:
> Should I add a note in the RELEASE.md and in the documentation for a private function?

I don't think we document private function in release note. But I get the intention of this is making use the new function in plugins right? In that case, what's preventing to make this available as a public API (new, non-breaking API?).

Galileo-Galilei:
Yes, it is for sure intended to be public in the end. I was reluctant to make it public immediately because it is a strong commitment with the "no breaking change" policy and I thought we could make it public later, after I could gather feedback from a couple of plugin developers. I can make it public immediately if all think this is the right move.

ankatiyar:
re: the resolution and merging behaviour, I think that was intentional to not allow for confusing and unintentional variable interpolation across environments. Globals (`globals:` resolver) allows for cross-environment interpolation otherwise resolution is meant to be per environment. While it was a convention that we created, it might not be ideal to change it now _after_ the breaking release.

merelcht:
Thanks for the PR @Galileo-Galilei, this looks like the correct approach to me. Since it will be quite a while before we can now remove public API I agree with adding the methods as private now and convert them to public after some feedback from plugin developers 👍 

On the topic of the resolution & merging order, it's been a while and I don't remember all of the decision making. I did find https://github.com/kedro-org/kedro/issues/2465 which describes we don't want to resolve between environments to make things easier to follow. I think it's a good point, but it's also a good time to re-assess this. In any case, if we decide to change the behaviour we need to think of how to do that without a breaking change.

ravi-kumar-pilla:
Hi @Galileo-Galilei ,

I agree with Merel on introducing the `_get_unresolved_config` as a private method for now. 

On the resolution order i.e., **resolve then merge** as we have now vs **merge then resolve**,   though I am not aware of the entire decision making process, since we have this in the [docs](https://docs.kedro.org/en/stable/api/config/kedro.config.OmegaConfigLoader/) -

```
When the same key appears in any two config files located in different (sub)directories, the last processed config path takes precedence and overrides this key and any sub-keys.
```

I feel this should apply to how resolutions work. i.e., `merge then resolve` considering the entire config is treated as one file at the end because we are allowing overrides and clubbing multiple environment config files.

Thank you

Galileo-Galilei:
Hi @ravi-kumar-pilla , thanks for providing this doc  that suggests the "merge then resolve" should likely be the correct behaviour. In general I think it's a bad idea to change the default of the underlying library (omegaconf here ) because it needs extra documentation which is kind of a _state of failure_ as pointed out by juan. Moreover this is very likely this subtlety has been noticed only by very few users, and they would have likely be fine with the other way.

Regarding the method, since we now have experimental decorator this is exactly the kind of situation where it should be used right? So the method can be public and experimental?

If everyone agrees, I can take over it not this weekend but the next one and submit a final PR with duplicated methods all along the way.

Galileo-Galilei:
Sorry I am currrently overwhelmed at work and can't take over this PR for now. I'll try to do it but can't give a timeline right now. If anything comes up during KedroServiceSession redesign, feel free to take over / do something parallel.

PR #5118, opened 2025-09-23
Fix some typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

Note that it fixes `pattens` in a signature (and body) of an internal function


merelcht:
Hi @yarikoptic, thanks for this suggestion. We have "Vale" in place too to check spelling for markdown files. Before adding in another tool, I think we should check if there's any overlap and if not make sure they check things in similar ways to keep spelling consistency across python and markdown files.

kedro/io/catalog_config_resolver.py
```diff
@@ -289,9 +289,9 @@ def _validate_pattern_config(cls, ds_name: str, ds_config: dict[str, Any]) -> No
         CatalogConfigResolver._validate_pattern_config(pattern, config)
         # No error
         pattern = "{namespace}.int_{name}"
-        config = {"filepath": "{nam}.csv"}
+        config = {"filepath": "{name}.csv"}
         CatalogConfigResolver._validate_pattern_config(pattern, config)
-        # DatasetError: Incorrect dataset configuration provided. Keys used in the configuration {'{nam}'} should present in
```
ankatiyar, line 294: This seems to be an intentional misspelling

yarikoptic: gotcha - fixed

yarikoptic:
@merelcht note that codespell does not actually "spell" -- it just fixes typos it knows about. So there should be less false positives. But also it checks all and any text file(s) it finds, so not language/format aware/specific at all. Altogether could be considered complimentary. Your call

deepyaman:
> Hi @yarikoptic, thanks for this suggestion. We have "Vale" in place too to check spelling for markdown files. Before adding in another tool, I think we should check if there's any overlap and if not make sure they check things in similar ways to keep spelling consistency across python and markdown files.

Overall, Vale is more of a prose style linter, while Codespell is a minimalist spell checker. That said, Vale does have a spell checker, too. But why is it off for Markdown? 🤔 https://github.com/kedro-org/kedro/blob/491c822f87dda65ff21456213d5dc5e6cd1a5eb4/.vale.ini#L7

The other problem with Vale IMO is that it gives so many suggestions that people just ignore them Looking at the logs on https://github.com/kedro-org/kedro/actions/runs/16328163642/job/46123623172, there are even some ERROR severity things flagged (not introduced by the PR).

I think it's fine to use Vale for spelling if it works, but it's worth understanding why it's not flagging some of these obvious issues.

ankatiyar:
@deepyaman The spell check is on, it just uses the Kedro config https://github.com/kedro-org/kedro/blob/main/.github/styles/Kedro/Spellings.yml instead of the default Vale one.

ankatiyar:
We already have Vale as a spell checker for docs, but since we added it, it only ran on the new additions which is why I think some of these typos were left out. I suppose `codespell` also checks for typos in the code? I'd be in favour of accepting the corrections for the typos but leaving out `codespell` for now. It seems promising but we already have a tool for spell checking, it might be overkill?

merelcht:
I agree with @ankatiyar ! Let's include the fixes, but leave `codespell` out for now. If that's okay with you @yarikoptic, we can clean up the PR and get it merged.

PR #5142, opened 2025-10-01
[Draft] Add docs on working with Pandera for data validation

rashidakanchwala:

## Checklist

### DIFF
--- a/docs/integrations-and-plugins/pandera.md
+++ b/docs/integrations-and-plugins/pandera.md
@@ -1,0 +1,694 @@
+# How to add Pandera data validation to your Kedro workflow
+
+[Pandera](https://pandera.readthedocs.io/) is an open-source data validation library for pandas DataFrames that enables schema validation with expressive and flexible checks. It helps ensure data quality by validating types, ranges, business rules, and statistical properties of your datasets.
+
+Adding Pandera to a Kedro project enables you to catch data quality issues early in your pipeline. For example, you can validate that CSV columns have the correct types, enforce business rules (like "price must be positive"), and automatically coerce string data to appropriate types. When collaborating with others on a Kedro project, Pandera schemas serve as documentation of your data contracts.
+
+## Prerequisites
+
+You will need the following:
+
+- A working Kedro project. The examples in this document assume the `spaceflights-pandas` starter. If you're unfamiliar with the Spaceflights project, check out [our tutorial](../tutorials/spaceflights_tutorial.md).
+- Pandera installed into the project.
+
+To set yourself up, create a new Kedro project:
+
+```bash
+kedro new --starter=spaceflights-pandas --name spaceflights-pandera
+```
+
+Now navigate to your project directory and add Pandera with pandas support to your `requirements.txt`:
+
+```bash
+cd spaceflights-pandera
+echo 'pandera[pandas]>=0.18.0' >> requirements.txt
+```
+
+Install the project dependencies:
+
+```bash
+uv pip install -r requirements.txt
+```
+
+```{note}
+Although Pandera works best when working with tabular data pipelines, you can validate any pandas DataFrame in your Kedro workflow, not just ML/AI pipelines.
+```
+
+## Simple use cases
+
+This section explains how you can use Pandera to validate your Kedro datasets in a straightforward way.
+
+### Basic validation of Kedro datasets using Pandera schemas
+
+The simplest way to add validation is to define schemas for your datasets and validate them using Kedro hooks.
+
+First, create a directory structure for your schemas:
+
+```bash
+mkdir -p src/spaceflight_pandera/schemas
+touch src/spaceflight_pandera/schemas/__init__.py
+```
+
+Create `src/spaceflights_pandera/schemas/raw.py` with validation schemas for your raw datasets:
+
+```python
+"""Pandera schemas for raw data validation."""
+from pandera import Column, DataFrameSchema, Check
+
+companies_schema = DataFrameSchema(
+    columns={
+        "id": Column(
+            int,
+            nullable=False,
+            unique=True,
+            coerce=True,  # Automatically convert "123" -> 123
+        ),
+        "total_fleet_count": Column(
+            float,
+            nullable=True,
+            checks=Check.greater_than_or_equal_to(0),
+            coerce=True,
+        ),
+        "iata_approved": Column(
+            str,
+            nullable=True,
+            checks=Check.isin(["t", "f"]),
+        ),
+    },
+    strict=False,  # Allow additional columns not in schema
+    coerce=True,   # Enable automatic type conversion
+    name="companies_raw",
+)
+
+shuttles_schema = DataFrameSchema(
+    columns={
+        "id": Column(int, nullable=False, unique=True, coerce=True),
+        "passenger_capacity": Column(
+            int,
+            nullable=False,
+            checks=Check.greater_than_or_equal_to(1),
+            coerce=True,
+        ),
+        "engines": Column(
+            float,  # Use float to handle nullable integers (NaN compatibility)
+            nullable=True,
+            checks=Check.greater_than_or_equal_to(0),
+            coerce=True,
+        ),
+    },
+    strict=False,
+    coerce=True,
+    name="shuttles_raw",
+)
+
+reviews_schema = DataFrameSchema(
+    columns={
+        "shuttle_id": Column(int, nullable=False, unique=True, coerce=True),
+        "review_scores_rating": Column(
+            float,
+            nullable=True,
+            checks=Check.in_range(0, 100),
+            coerce=True,
+        ),
+        "number_of_reviews": Column(
+            int,
+            nullable=True,
+            checks=Check.greater_than_or_equal_to(0),
+            coerce=True,
+        ),
+    },
+    strict=False,
+    coerce=True,
+    name="reviews_raw",
+)
+```
+
+```{tip}
+**Alternative schema style:**  
+Instead of defining schemas as `DataFrameSchema` objects, Pandera also supports a Pydantic-like API with `DataFrameModel` classes and type annotations.  
+This can give you IDE/type-checking support. See the [Typed Schemas guide](https://pandera.readthedocs.io/en/stable/schema_models.html).
+```
+
+
+```{important}
+Use `coerce=True` to automatically convert types during validation. This is essential for CSV data where columns are initially loaded as strings.
+
+For nullable integer columns, use `float` type instead of `int` because pandas represents nullable integers as float64 (since NaN is a float value).
+```
+
+
+
+Next, create a hook to validate datasets. Create `src/spaceflights_pandera/hooks/validation.py`:
+
+```python
+"""Kedro hooks for Pandera validation integration."""
+import logging
+from typing import Any, Dict
+
+import pandas as pd
+from kedro.framework.hooks import hook_impl
+from kedro.io import DataCatalog
+from pandera.errors import SchemaError, SchemaErrors
+
+from ..schemas.raw import companies_schema, shuttles_schema, reviews_schema
+
+logger = logging.getLogger(__name__)
+
+
+class PanderaValidationHooks:
+    """Kedro hooks that integrate Pandera validation."""
+
+    def __init__(self):
+        # Map catalog dataset names to their validation schemas
+        self.raw_schemas = {
+            "companies": companies_schema,
+            "shuttles": shuttles_schema,
+            "reviews": reviews_schema,
+        }
+
+    @hook_impl
+    def before_node_run(
+        self,
+        node,
+        catalog: DataCatalog,
+        inputs: Dict[str, Any],
+        is_async: bool,
+    ) -> Dict[str, Any]:
+        """Validate and coerce input data before node execution."""
+        validated_inputs = {}
+
+        for input_name, data in inputs.items():
+            schema = self.raw_schemas.get(input_name)
+
+            if schema is None or not isinstance(data, pd.DataFrame):
+                validated_inputs[input_name] = data
+                continue
+
+            logger.info(f"Validating '{input_name}' for node '{node.name}'")
+
+            try:
+                validated_data = schema.validate(data, lazy=True)
+                logger.info(f"✓ Validation passed for '{input_name}'")
+                validated_inputs[input_name] = validated_data
+            except (SchemaError, SchemaErrors) as e:
+                logger.error("Validation failed for '%s': %s", input_name, e)
+                raise
+
+        return validated_inputs
+```
+
+Finally, register the hook in `src/spaceflights_pandera/settings.py`:
+
+```python
+"""Project settings."""
+from spaceflights_pandera.hooks.validation import PanderaValidationHooks
+
+HOOKS = (PanderaValidationHooks(),)
+```
+
+From this point, when you execute `kedro run` you will see the validation logs:
+
+```bash
+[10/01/25 11:10:13] INFO     Validating 'companies' for node 'preprocess_companies_node'
+                    INFO     ✓ Validation passed for 'companies'
+                    INFO     Validating 'shuttles' for node 'preprocess_shuttles_node'
+                    INFO     ✓ Validation passed for 'shuttles'
+                    INFO     Validating 'reviews' for node 'create_model_input_table_node'
+                    INFO     ✓ Validation passed for 'reviews'
+```
+
+**What happens when validation fails?**  
+Let’s make one schema rule deliberately too strict so you can see Pandera in action.
+
+In `src/spaceflights_pandera/schemas/raw.py`, change the `total_fleet_count` column check:
+
+```diff
+- checks=Check.greater_than_or_equal_to(0),
++ checks=Check.greater_than(100),  # intentionally too strict
+```
+
+Now run the pipeline again:
+
+```bash
+kedro run
+```
+
+You should see Pandera raise a SchemaErrors exception, reporting exactly which rows failed:
+
+```bash
+SchemaErrors: Column 'total_fleet_count' failed check 'greater_than_or_equal_to(100)'
+failure cases: [1.0, 2.0, 5.0, ...]
+```
+
+Because the hook uses lazy=True, Pandera collects all errors at once, making it easy to spot every problem.
+
+Finally, reset the schema back to the correct check (greater_than_or_equal_to(0)) before continuing with the rest of the tutorial.
+
+
+### Controlling validation behavior with environment variables
+
+You can control validation behavior without modifying code by using environment variables.
+
+Create configuration helpers in `src/spaceflights_pandera/schemas/__init__.py`:
+
+```python
+"""Configuration helpers for validation behavior."""
+import os
+
+
+def is_validation_enabled() -> bool:
+    """Control validation via VALIDATION_ENABLED environment variable.
+
+    Set to 'false' to disable all validation (useful for quick debugging).
+    Default: true
+    """
+    return os.getenv("VALIDATION_ENABLED", "true").lower() == "true"
+
+
+def should_fail_on_error() -> bool:
+    """Control whether validation errors stop the pipeline.
+
+    - true: Pipeline stops on validation errors (recommended for CI/CD)
+    - false: Log errors but continue (useful for data exploration)
+
+    Default: true
+    """
+    return os.getenv("VALIDATION_FAIL_ON_ERROR", "true").lower() == "true"
+
+
+def get_lazy_flag() -> bool:
+    """Control Pandera's lazy validation mode.
+
+    - true: Collect ALL validation errors before failing (better error visibility)
+    - false: Fail on first validation error (faster for CI/CD)
+
+    Set via VALIDATION_LAZY environment variable.
+    Default: true (collect all errors)
+    """
+    return os.getenv("VALIDATION_LAZY", "true").lower() == "true"
+```
+
+Update your hook to use these configuration helpers:
+
+```diff
+
+ from pandera.errors import SchemaError, SchemaErrors
++from ..schemas import is_validation_enabled, should_fail_on_error, get_lazy_flag
+ 
+ class PanderaValidationHooks:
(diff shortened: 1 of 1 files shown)


docs/integrations-and-plugins/pandera.md
```diff
@@ -0,0 +1,687 @@
+# How to add Pandera data validation to your Kedro workflow
+
+[Pandera](https://pandera.readthedocs.io/) is an open-source data validation library for pandas DataFrames that enables schema validation with expressive and flexible checks. It helps ensure data quality by validating types, ranges, business rules, and statistical properties of your datasets.
+
+Adding Pandera to a Kedro project enables you to catch data quality issues early in your pipeline. For example, you can validate that CSV columns have the correct types, enforce business rules (like "price must be positive"), and automatically coerce string data to appropriate types. When collaborating with others on a Kedro project, Pandera schemas serve as documentation of your data contracts.
```
noklam: I would against promoting this pattern while this is possible. The I/O are usually define in DataCatalog, if the logic start leaking to hooks/data validation this makes it harder to debug. It also cause potentially problem, i.e. if data validation is not enable, the pipeline will be incorrect because of not casting the type.

docs/integrations-and-plugins/pandera.md
```diff
@@ -0,0 +1,687 @@
...
+from kedro.framework.hooks import hook_impl
+from kedro.io import DataCatalog
+from pandera.errors import SchemaError, SchemaErrors
+
+from ..schemas.raw import companies_schema, shuttles_schema, reviews_schema
+
+logger = logging.getLogger(__name__)
+
+
+class PanderaValidationHooks:
```
noklam:
```suggestion
class PanderaValidationHook:
```

docs/integrations-and-plugins/pandera.md
rashidakanchwala: you mean the “coerce string data” part right?. That makes sense. I’ll also remove the example I share later on along with the notes. You’re right; pandera should focus on validation and raising errors, not casting or changing the data! thanks

noklam: exactly

docs/integrations-and-plugins/pandera.md
```diff
@@ -0,0 +1,687 @@
+# How to add Pandera data validation to your Kedro workflow
```
deepyaman:
```suggestion
# How to validate data using Pandera in your Kedro workflow
```
Maybe? I feel like it emphasizes validating data as the end goal more than adding a Pandera integration, but could go either way.

docs/integrations-and-plugins/pandera.md
```diff
@@ -0,0 +1,687 @@
...
+
+[Pandera](https://pandera.readthedocs.io/) is an open-source data validation library for pandas DataFrames that enables schema validation with expressive and flexible checks. It helps ensure data quality by validating types, ranges, business rules, and statistical properties of your datasets.
```
deepyaman: Not just pandas.

docs/integrations-and-plugins/pandera.md
```diff
@@ -0,0 +1,687 @@
...
+
+Install the project dependencies:
+
+```bash
+uv pip install -r requirements.txt
+```
+
+
+!!! note
+    Although Pandera works best when working with tabular data pipelines, you can validate any pandas DataFrame in your Kedro workflow, not just ML/AI pipelines.
```
deepyaman:
```suggestion
```
I don't understand the purpose of this note? What are you trying to communicate?

docs/integrations-and-plugins/pandera.md
```diff
@@ -0,0 +1,687 @@
...
+
+## Simple use cases
+
+This section explains how you can use Pandera to validate your Kedro datasets in a straightforward way.
```
deepyaman:
```suggestion
This section explains how you can use Pandera to validate your Kedro datasets.
```
Too wordy?

docs/integrations-and-plugins/pandera.md
```diff
@@ -0,0 +1,687 @@
...
+- A working Kedro project. The examples in this document assume the `spaceflights-pandas` starter. If you're unfamiliar with the Spaceflights project, check out [our tutorial](../tutorials/spaceflights_tutorial.md).
+- Pandera installed into the project.
+
+To set yourself up, create a new Kedro project:
+
+```bash
+kedro new --starter=spaceflights-pandas --name spaceflights-pandera
+```
+
+Navigate to your project directory and add Pandera with pandas support to your requirements.txt:
```
deepyaman:
```suggestion
Navigate to your project directory and add Pandera with pandas support to your `requirements.txt`:
```

or

```suggestion
Navigate to your project directory and add Pandera with pandas support to your requirements file:
```

docs/integrations-and-plugins/pandera.md
```diff
@@ -0,0 +1,687 @@
...
+        inputs: Dict[str, Any],
+        is_async: bool,
+    ) -> Dict[str, Any]:
+        """Validate and coerce input data before node execution."""
+        validated_inputs = {}
+
+        for input_name, data in inputs.items():
+            schema = self.raw_schemas.get(input_name)
+
+            if schema is None or not isinstance(data, pd.DataFrame):
```
deepyaman: Why is the `isinstance()` check required?

docs/integrations-and-plugins/pandera.md
```diff
@@ -0,0 +1,687 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class PanderaValidationHooks:
+    """Kedro hooks that integrate Pandera validation."""
+
+    def __init__(self):
+        # Map catalog dataset names to their validation schemas
+        self.raw_schemas = {
```
deepyaman, line 141: Feels very weird that you're defining your schemas in the hook; seems like it breaks the software engineering principles that Kedro advocates for. I know you mention somewhere below that you could store schemas in YAML, but maybe even for this example it makes sense to put a pointer to the schema in YAML or some other config?

docs/integrations-and-plugins/pandera.md
```diff
@@ -0,0 +1,687 @@
...
+                    INFO     ✓ Validation passed for 'companies'
+                    INFO     Validating 'shuttles' for node 'preprocess_shuttles_node'
+                    INFO     ✓ Validation passed for 'shuttles'
+                    INFO     Validating 'reviews' for node 'create_model_input_table_node'
+                    INFO     ✓ Validation passed for 'reviews'
+```
+
+### What happens when validation fails?
+
+Let's make one schema rule deliberately too strict so you can see Pandera in action.
```
deepyaman:
```suggestion
Let's make one schema rule deliberately fail so you can see Pandera in action.
```
Nit: I think "too strict" would be more like if it was the wrong type but still usable; arbitrarily high value sounds more like deliberately failing. :)

docs/integrations-and-plugins/pandera.md
```diff
@@ -0,0 +1,687 @@
...
+
+```console
+SchemaErrors: Column 'total_fleet_count' failed check 'greater_than_or_equal_to(100)'
+failure cases: [1.0, 2.0, 5.0, ...]
+```
+
+Because the hook uses `lazy=True`, Pandera collects all errors at once, making it easy to spot every problem.
+
+!!! note
+    Reset the schema back to the correct check (`greater_than_or_equal_to(0)`) before continuing with the rest of the tutorial.
```
deepyaman:
```suggestion
Reset the schema back to the correct check (`greater_than_or_equal_to(0)`) before continuing with the rest of the tutorial.
```

Nit: I don't know if this needs to be a callout.

docs/integrations-and-plugins/pandera.md
```diff
@@ -0,0 +1,687 @@
...
+# Log errors but don't stop pipeline (for data exploration)
+export VALIDATION_FAIL_ON_ERROR=false
+kedro run
+
+# Fail on first error (for faster CI/CD)
+export VALIDATION_LAZY=false
+kedro run
+```
+
+### Validating intermediate datasets
```
deepyaman: Does it add additional value to go through this?

docs/integrations-and-plugins/pandera.md
```diff
@@ -0,0 +1,687 @@
...
+failure cases: [1.0, 2.0, 5.0, ...]
+```
+
+Because the hook uses `lazy=True`, Pandera collects all errors at once, making it easy to spot every problem.
+
+!!! note
+    Reset the schema back to the correct check (`greater_than_or_equal_to(0)`) before continuing with the rest of the tutorial.
+
+
+### Controlling validation behavior with environment variables
```
deepyaman, line 319: This seems like a good, more advanced tip.

If we're covering things like this, I think it's also worth covering caching validation results (i.e. if a dataset is an input to 10 nodes, it's only validated once).

deepyaman: Also, looking below, is there a reason why this doesn't also fall under the advanced usage section?

docs/integrations-and-plugins/pandera.md
deepyaman: I'm OK with it either way; I feel like use of "Hooks" as the class name suffix is pretty common.

docs/integrations-and-plugins/pandera.md
```diff
@@ -0,0 +1,687 @@
...
+    logger.info(f"✓ Validation passed for '{input_name}'")
+    validated_inputs[input_name] = validated_data
+except SchemaError as e:
+    self._handle_validation_error(input_name, e)
+    validated_inputs[input_name] = data
+```
+
+### DataFrame-level checks for business rules
+
+So far, our intermediate schema only checked column types and uniqueness.
```
deepyaman: But there was use of built-in `Check`s before that was more than just column types and uniqueness. Not sure what is "business rules;" maybe this is more custom checks?

That said, that's also covered in Pandera documentation, so not sure if need to cover all the types of checks unless the integration is different.

docs/integrations-and-plugins/pandera.md
```diff
@@ -0,0 +1,687 @@
...
+```
+
+
+!!! tip
+    **Distribution checks and hypothesis testing**
+    Beyond simple rules, Pandera supports **statistical hypothesis tests** (e.g. two-sample tests) to validate whether two datasets come from the same distribution.
+    This is particularly useful for detecting **data drift** between training and serving environments.
+    See [Hypothesis Testing](https://pandera.readthedocs.io/en/stable/hypothesis.html).
+
+### Driving validation with Kedro parameters
```
deepyaman: Interesting use case, didn't think about it before

docs/integrations-and-plugins/pandera.md
```diff
@@ -0,0 +1,687 @@
...
+```
+
+Now you can adjust validation dynamically at runtime:
+
+```bash
+# Override via Kedro CLI
+kedro run --params="validation.rating_mean_min=45.0,validation.rating_mean_max=90.0"
+```
+
+### Writing tests for your schemas
```
deepyaman, line 547: Nit: Wonder if Kedro documentation should be responsible for this. I guess, on the one hand, you could say it's promoting better practices overall.

docs/integrations-and-plugins/pandera.md
```diff
@@ -0,0 +1,687 @@
...
+
+
+## Further reading
+
+* [Pandera documentation](https://pandera.readthedocs.io/)
+* [Kedro hooks documentation](https://docs.kedro.org/en/stable/hooks/introduction.html)
+* [Lazy validation](https://pandera.readthedocs.io/en/stable/lazy_validation.html)
+* [Type coercion](https://pandera.readthedocs.io/en/stable/dtype_validation.html)
+* [Custom checks](https://pandera.readthedocs.io/en/stable/checks.html)
+* [Multi-backend support](https://pandera.readthedocs.io/en/stable/ecosystem.html) - Pandera also validates data in Dask, Polars, Modin, and PySpark
```
deepyaman, line 623: Ibis also.

docs/integrations-and-plugins/pandera.md
```diff
@@ -0,0 +1,687 @@
...
+
+For a declarative approach to validation using catalog metadata, see the community-maintained [kedro-pandera plugin](https://github.com/Galileo-Galilei/kedro-pandera).
```
deepyaman: It's not really maintained right now AFAIK. 😅

docs/integrations-and-plugins/pandera.md
rashidakanchwala: The schemas are not defined in the hook; they are imported 

`from ..schemas.raw import companies_schema, shuttles_schema, reviews_schema`

in the hooks we are just mapping the schema to the catalog. We could have a function where we can map this in the data catalog just like how kedro-pandera does. but i thought that make this example a bit complicated.

docs/integrations-and-plugins/pandera.md
rashidakanchwala: This use case shows business-rule validation — logic involving multiple columns based on domain knowledge. You are right it doesn’t introduce new Pandera functionality. If others feel it overlaps with existing Pandera docs, I’m happy to remove it.

docs/integrations-and-plugins/pandera.md
rashidakanchwala: This was an AI-added bit that I was debating myself if it was actually required! Removed it now ! thanks

docs/integrations-and-plugins/pandera.md
rashidakanchwala: I will leave this for now; if others feel the same. i can remove it

docs/integrations-and-plugins/pandera.md
rashidakanchwala: I am removing this part for now; it's not been working correctly :\

docs/integrations-and-plugins/pandera.md
```diff
@@ -0,0 +1,570 @@
...
+```
+
+!!! tip
+    **Alternative schema style:**
+    Instead of defining schemas as `DataFrameSchema` objects, Pandera also supports a Pydantic-like API with `DataFrameModel` classes and type annotations.
+    This can give you IDE/type-checking support. See the [Typed Schemas guide](https://pandera.readthedocs.io/en/stable/schema_models.html).
+
+
+!!! info
+    For nullable integer columns, use `float` type instead of `int` because pandas represents nullable integers as float64 (since NaN is a float value).
```
noklam: is there a reference about this? I am not sure if this is true.

What does this mean then when you use nullable integer in the example?
```
"number_of_reviews": Column(
            int,
            nullable=True,
     
```

rashidakanchwala: Thanks for pointing this! you're  right, pandas represents nullable integers as `Int64`. Let me remove the note.

docs/integrations-and-plugins/pandera.md
lvijnck: I agree, we've struggled quite a bit with the validation logic. Either keep the validation logic close to the dataset, or keep it close to the node logic I'd say.

As part of our project, we went for the latter:

1. A small wrapper on top of Pandera to work both with Pandas and Spark
2. Decorators to validate input/output of functions

Some code:

Wrapper:

```python
"""
Temp. wrapper module to enable data quality checks defined independently of the underlying
library used to represent the data. Currently supports Spark and Pandas dataframes.

NOTE: Should be removed as soon as Pandera supports this.
"""

import inspect
import typing
from dataclasses import dataclass
from functools import wraps
from typing import Dict, List, Optional

import pandas as pd
import pandera.pandas as pa
import pandera.pyspark as pas
import pyspark.sql as ps
from pandera.decorators import _handle_schema_error


@dataclass
class Column:
    """Data class to represent a class agnostic Pandera Column."""

    type_: type
    checks: Optional[List] = None
    nullable: bool = True


@dataclass
class DataFrameSchema:
    """Data class to represent a class agnostic Pandera DataFrameSchema."""

    columns: Dict[str, Column]
    unique: Optional[List] = None
    strict: bool = False

    def build_for_type(self, type_) -> typing.Union[pas.DataFrameSchema, pa.DataFrameSchema]:
        # Build pandas version
        if type_ is pd.DataFrame:
            return pa.DataFrameSchema(
                columns={
                    name: pa.Column(col.type_, checks=col.checks, nullable=col.nullable)
                    for name, col in self.columns.items()
                },
                unique=self.unique,
                strict=self.strict,
            )

        # Build pyspark version
        if type_ is ps.DataFrame:
            return pas.DataFrameSchema(
                columns={
                    name: pas.Column(col.type_, checks=col.checks, nullable=col.nullable)
                    for name, col in self.columns.items()
                },
                unique=self.unique,
                strict=self.strict,
            )

        raise TypeError()


def check_input(schema: DataFrameSchema, df_name: str):
    """Decorator to validate output schema of decorated function.

    Args:
        schema: Schema to validate
        df_name (optional): name of output arg to validate
    """

    def decorator(func):
        @wraps(func)
        def wrapper(*args, **kwargs):
            # Get type hints for the function
            type_hints = typing.get_type_hints(func)

            # Bind args and kwargs to parameter names
            sig = inspect.signature(func)
            bound = sig.bind(*args, **kwargs)
            bound.apply_defaults()

            value = bound.arguments.get(df_name)
            arg_type = type_hints.get(df_name)

            df_schema = schema.build_for_type(arg_type)

            if value is not None:
                try:
                    # breakpoint()
                    df_schema.validate(value, lazy=False)
                except pa.errors.SchemaError as e:
                    _handle_schema_error("check_input", func, df_schema, value, e)

            output = func(*args, **kwargs)

            return output

        return wrapper

    return decorator


def check_output(
    schema: DataFrameSchema, df_name: Optional[str] = None, raise_df_undefined: bool = True, pass_columns: bool = False
):
    """Decorator to validate output schema of decorated function.

    Args:
        schema: Schema to validate
        df_name (optional): name of output arg to validate
        raise_df_undefined: if set, ensures error is thrown if `df_name` is not defined
        pass_cols (optional): boolean indicating whether cols should be passed into callable
    """

    def decorator(func):
        @wraps(func)
        def wrapper(*args, **kwargs):
            if not (type_ := typing.get_type_hints(func).get("return")):
                raise TypeError("No output typehint specified!")

            if df_name:
                if not typing.get_origin(typing.Dict) == dict:
                    raise TypeError("Specified df_name arg, but function output typehint is not dict.")

                type_ = typing.get_args(type_)[1]

            df_schema = schema.build_for_type(type_)

            if pass_columns:
                output = func(*args, **kwargs, cols=list(schema.columns.keys()))
            else:
                output = func(*args, **kwargs)

            if df_name is not None:
                df = output.get(df_name)

                if df is None and raise_df_undefined:
                    raise RuntimeError(f"df {df_name} not found!")
            else:
                df = output

            if df is not None:
                try:
                    df_schema.validate(df, lazy=False)
                except pa.errors.SchemaError as e:
                    _handle_schema_error("check_output", func, df_schema, df, e)

            return output

        return wrapper

    return decorator

```

A node file

```python
@check_output(
    schema=DataFrameSchema(
        columns={
            "source": Column(T.StringType(), nullable=False),
            "target": Column(T.StringType(), nullable=False),
            "source_embedding": Column(T.ArrayType(T.FloatType()), nullable=False),
            "target_embedding": Column(T.ArrayType(T.FloatType()), nullable=False),
        },
        unique=["source", "target"],
    )
)
def attach_embeddings(
    pairs_df: ps.DataFrame,
    nodes: ps.DataFrame,
) -> ps.DataFrame:

    # logic here ...
```

lrcouto:
Tested the tutorial itself and it's easy to follow. I could implement all of the steps with no issues.

After trying both the GX and Pandera versions of this, I kind of agree that keeping the validation logic close to a hook instead of a node or dataset feels a little off. It works just fine, but doesn't feel "Kedro-ish", if that makes sense.

DimedS:
Thank you, @rashidakanchwala , for the detailed write-up - I think this approach works well overall and explains Pandera integration.
That said, I personally don’t like using **hooks** for data validation. A few reasons:  

- Hooks add hidden control flow; validation becomes implicit and harder to trace or debug.  
- They don’t work well with **lazy datasets** - validation would either be skipped or force premature data loading.  
- Conceptually, validation feels much closer to the **DataCatalog**, since it’s about enforcing data contracts at I/O boundaries rather than orchestration logic.  

In my view, it would be ideal if we could express validation declaratively in the catalog itself, for example:

```yaml
companies:
  type: pandas.CSVDataset
  filepath: data/01_raw/companies.csv
  validator: spaceflights_pandera.schemas.companies_schema
  validate_on: load  # or "save"
```

Then the DataCatalog would automatically trigger validation when the dataset is actually loaded or saved.  
As I understand it, implementing this would just require a slightly modified `DataCatalog`, which should be feasible since we allow such extensions from 1.0.0.  

It could even be extended with something like a `mode:` flag, where users could request validation of *all project inputs* before any nodes start - effectively an eager validation mode for extra safety before pipeline execution.

So I mean, I’m not against adding the approach you proposed to our docs, but I think it should be treated as a short-term solution. We should continue exploring how to make this simpler through the DataCatalog - for example, by explaining how to build a custom catalog or by figuring out which plugin architecture would fit us best.

It would be great to hear @ElenaKhaustova thoughts on this topic as well.

rashidakanchwala:
thanks both for your comments. @lrcouto , and @DimedS :) -- this is very useful in thinking about our overall data validation approach

docs/integrations-and-plugins/pandera.md
deepyaman: @lvijnck Is there any reason you can't just use the Pandera-provided `check_types` function decorator?

Cc @rashidakanchwala with the Pydantic integration leveraging type hints, maybe this is a way to go?

docs/integrations-and-plugins/pandera.md
```diff
@@ -0,0 +1,626 @@
...
+
+
+Install the project dependencies:
+
+```bash
+uv pip install -r requirements.txt
+```
+
+
+## Use cases
```
ravi-kumar-pilla, line 34:
```suggestion
## Integration Options
```

PR #5148, opened 2025-10-03
Improve load node#4169

### LINKED ISSUE #4169: Improve `%load_node` for node that doesn't have persisted dataset
## Description

Originated from https://github.com/kedro-org/kedro/issues/4158

> %load_node works great but it doesn't work on MemoryDataset, I don't save every node and it's not easy to figure out which nodes I need to run again to produce the data

## Context

- This is much more powerful because not every node has persisted data and this limited the usage of the feature.
- This enable much more powerful slicing feature in `kedro-viz`, which is currently limited because we do not want the slicing generate a command that Kedro does not know how to run.
- Same as above, if we are able to support more flexible slicing, we may end up expanding the API of `kedro run` as well since we can support different combinations of slicing

## Possible Implementation

Check out the `suggest_resume_scanerio` in `SequentialRunner`:
https://github.com/kedro-org/kedro/blob/91765e3e70f8a4ba7ef3931a7855aed0ec97eb82/kedro/runner/runner.py#L200-L210

This roughly has the logic to resume pipeline but it's hidden in a private API, we need to surface this for more generic usage.

khushboo-katariaa11:
### Description
- Improve `%load_node` to work when node inputs are not persisted (e.g., `MemoryDataset`).
- When inputs are missing, the generated cell now invokes `session.run(to_outputs=[...], only_missing_outputs=True)` to materialize them, then loads inputs and calls the node function.
- Addresses and closes #4169; unblocks more powerful slicing/debugging flows (e.g., VS Code and Kedro-Viz integration).

### Development notes
- Edited `kedro/ipython/__init__.py`:
  - `_format_node_inputs_text()` now:
    - Collects required input dataset names.
    - Detects missing inputs via `catalog.list()`/`catalog.exists()`.
    - If any missing, runs `session.run(to_outputs=missing, only_missing_outputs=True)` to build them.
    - Loads inputs with `catalog.load(...)` and proceeds with the node call.
- Updated tests in `tests/ipython/test_ipython.py` to reflect the new generated inputs cell content.
- Behavior is backward compatible for normal usage where `%reload_kedro` has provided `session` and `catalog`.
- No public API surface changes; scope is the generated code for `%load_node`.

### Developer Certificate of Origin
We need all contributions to comply with the Developer Certificate of Origin (DCO). All commits must be signed off by including a `Signed-off-by` line in the commit message. See the linked guidance.

### Checklist
- [x] Signed off each commit with a DCO sign-off
- [ ] Opened this PR as a Draft if WIP
- [x] Updated the documentation to reflect the code changes (note: add a line in the `%load_node` docs that missing inputs are auto-built)
- [ ] Added a description of this change in the RELEASE.md file
- [x] Checked if this change will affect Kedro-Viz, and if so, communicated with the Viz team (this enables more flexible slicing)


noklam:
Instead of automatically run logic on behalf of the user, I wonder if there can be a less intrusive way to do this:
1. Add a flag, i.e. `--rerun-missing`
2. Identify the missing nodes and print the missing data when `--rerun-missing` is not applied.
3. Print/Show hints how user can "persisted" the intermediate data so they do not need to rerun this next time if they prefer (what changes need to be made in the catalog / what datasets are missing that required re-computed?)

If the flag is applied already - then there is no need to show this as we assume user know what they are doing.

@SajidAlamQB Is the `run_only_missing` feature useful here or is it not available with `session`?

merelcht:
Hi @khushboo-katariaa11, it's been a while since you opened this PR. Are you still interested in working on it? 🙂

PR #5208, opened 2025-11-11
Make Kedro core lightweight by moving cookiecutter, gitpython, rich, and build out of core dependencies

rashidakanchwala:
## Description

**NOTE - Please don’t look too closely at the code... this was a quick experiment written with AI code assist. The main thing to review is the flow, particularly how project creation dependencies are invoked ephemerally in a non-breaking way**.

This update makes Kedro’s core installation lighter by removing dependencies that are only needed during project creation. 

Cookiecutter, gitpython, rich, and build are no longer part of Kedro’s default dependencies. They’re now grouped under a new optional extra called [create] and installed only when needed.

When users run `kedro new` or` kedro pipeline create` without these extras installed, Kedro automatically falls back to using uvx to temporarily install the required tools for that command.

All other Kedro commands (like kedro run or kedro viz) continue to work as before — with no new dependencies added to the runtime environment.

## Development Notes

- Removes cookiecutter, gitpython, rich, and build from Kedro’s core dependencies.
- Adds a new optional extra:
  [project.optional-dependencies]
  create = ["cookiecutter", "gitpython", "rich", "build"]
- Adds a fallback so `kedro new` and `kedro pipeline create` automatically run via uvx if the create extra isn’t installed.
- Keeps Kedro’s runtime environment smaller and faster to install.

Also, In this PR, I’ve shown how rich could be installed ephemerally, but to truly make it optional we’d need some changes to the logging setup (to handle missing imports gracefully). technically, that would let us remove rich from core dependencies and move it under an optional extra like kedro[rich].

## QA Notes

Before testing, install this dev build: `pip install -e .`

  ### 1. Project creation

```kedro new --starter=<starter>```
 Should automatically use uvx to install the create dependencies and create the project successfully.

 ### 2. Immediately after project creation

Open the generated project’s conf/logging.yml.

Comment out the rich handler section and replace the root handlers with console.

It should look like this:
```
# rich:
#   class: kedro.logging.RichHandler
#   rich_tracebacks: True
#   # Advanced options for customisation.
#   # See https://docs.kedro.org/en/stable/logging/logging.html#project-side-logging-configuration
#   # tracebacks_show_locals: False

root:
  handlers: [console, info_file_handler]
```

### 3. Pipeline creation

`kedro pipeline create <name>`
Should also work through the same uvx fallback if cookiecutter, gitpython, rich, or build are not installed locally.

### 4. Run the project

`kedro run`
Logs should display normally in the console using the console handler.

## Checklist

### DIFF
--- a/kedro/framework/cli/_ephemeral.py
+++ b/kedro/framework/cli/_ephemeral.py
@@ -1,0 +1,92 @@
+"""Helper for ephemeral re-execution when optional tools are missing."""
+
+from __future__ import annotations
+
+import functools
+import importlib.util
+import os
+import shutil
+import subprocess
+import sys
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+
+def _have_all(modules: list[str]) -> bool:
+    """Check if all required modules are importable."""
+    return all(importlib.util.find_spec(m) for m in modules)
+
+
+def ensure_or_ephemeral(required_modules: list[str], argv: list[str]) -> None:
+    """If any required module is missing, re-exec the CLI via uvx ephemerally.
+
+    Args:
+        required_modules: List of module names to check (e.g., ["cookiecutter", "git"])
+        argv: Command-line arguments (sys.argv)
+
+    Raises:
+        SystemExit: If deps are missing and fallback is not possible/allowed
+    """
+    if _have_all(required_modules):
+        return  # All tools available, proceed normally
+
+    # Missing tools - check if ephemeral fallback is allowed
+    if os.environ.get("KEDRO_NO_UVX_FALLBACK"):
+        miss = [m for m in required_modules if not importlib.util.find_spec(m)]
+        raise SystemExit(
+            "Missing tools: " + ", ".join(miss) + "\n"
+            "Install locally, e.g.:\n"
+            f"  pip install {' '.join(miss)}\n"
+            "Or unset KEDRO_NO_UVX_FALLBACK to allow ephemeral fallback."
+        )
+
+    # Check if uvx is available
+    if not shutil.which("uvx"):
+        raise SystemExit(
+            "Missing tools and no uvx found.\n"
+            "Install tools locally OR install uv:\n"
+            "  pip install uv"
+        )
+
+    # Prevent infinite loop if uvx re-exec fails
+    if os.environ.get("KEDRO_UVX_REEXEC") == "1":
+        raise SystemExit("uvx re-exec loop detected; aborting.")
+
+    # Re-execute via uvx
+    env = {**os.environ, "KEDRO_UVX_REEXEC": "1"}
+    cmd = ["uvx", "--from", "kedro[create]", "kedro", *argv[1:]]
+    print(
+        "🔧 Missing tools — running ephemerally via uvx:\n  ",
+        " ".join(cmd),
+        file=sys.stderr,
+    )
+    raise SystemExit(subprocess.call(cmd, env=env))
+
+
+def ephemeral_if_missing(modules: list[str]) -> Callable:
+    """Decorator for Click commands that need extra tooling.
+
+    Args:
+        modules: List of module names required (e.g., ["cookiecutter", "git"])
+
+    Returns:
+        Decorator function
+
+    Example:
+        @cli.command()
+        @ephemeral_if_missing(["cookiecutter", "git"])
+        def new(...):
+            ...
+    """
+
+    def deco(func: Callable) -> Callable:
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):  # type: ignore[no-untyped-def]
+            ensure_or_ephemeral(modules, sys.argv)
+            return func(*args, **kwargs)
+
+        return wrapper
+
+    return deco
--- a/kedro/framework/cli/pipeline.py
+++ b/kedro/framework/cli/pipeline.py
@@ -11,6 +11,7 @@ from typing import TYPE_CHECKING, Any, NamedTuple
 import click
 
 import kedro
+from kedro.framework.cli._ephemeral import ephemeral_if_missing
 from kedro.framework.cli.utils import (
     KedroCliError,
     _clean_pycache,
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by datajoely** [comment]
Elegant, simple, ship it!


Other projects like Marimo have gone all in on uv and left pip behind. I'd argue rich should stay in here, just because it complicates things a little... but moving cookiecutter to ephemeral is a no brainer. In the past the click versions in cookiecutter and Kedro conflicted, this is the right way to do it

**Comment by noklam** [comment]
See
- https://github.com/kedro-org/kedro/issues/3967#issuecomment-2191004303
- https://github.com/kedro-org/kedro/issues/3967#issuecomment-2486889490

I am happy to move this out of the core dependency, `uvx` is nice but it's not something new it used to be possible with `pipx` already. I recalled the main argument was that this is still considered as a major breaking change. I think the key here is whether the team want to accept the breaking change.

File: kedro/framework/cli/_ephemeral.py [hunk 1]
```diff
@@ -0,0 +1,92 @@
...
+            "  pip install uv"
+        )
+
+    # Prevent infinite loop if uvx re-exec fails
+    if os.environ.get("KEDRO_UVX_REEXEC") == "1":
+        raise SystemExit("uvx re-exec loop detected; aborting.")
+
+    # Re-execute via uvx
+    env = {**os.environ, "KEDRO_UVX_REEXEC": "1"}
+    cmd = ["uvx", "--from", "kedro[create]", "kedro", *argv[1:]]
```
astrojuanlu, line 59: This seems quite cursed, but I didn't manage to break it 🗿

What's the purpose? To retain backwards compatibility?

rashidakanchwala: haha yes, the code will be cleaned up and done properly later ... this was just quick experimentation 🙂

and yep, the idea is to keep things non-breaking for users. So when they install Kedro, it only brings in the runtime dependencies. When they run `kedro new` or `kedro pipeline create`, it’ll call uvx to install the project-creation deps ephemerally. That way, the flow stays the same for users but keeps their env dependency light!

astrojuanlu: Understood, thank you.

In case it helps, I think there could be other ways to keep the current behavior in a non-breaking way. For example:

- Make a new `kedro-core` package that has minimal dependencies
- Make `kedro` depend on `kedro-core` plus cookiecutter, gitpython, rich, build
- So, upon `pip install kedro` everything stays the same, but `pip install kedro-core` gives the minimal functionality

rashidakanchwala: yup, that also makes sense. 

But I also wanted to ask do you see any cons of doing this style of approach. I was thinking if we do it like this then we still have one kedro package to maintain and have other dependencies as extras

astrojuanlu: I've seen both approaches in the ecosystem. ordeq clearly favours sub-packages. Dask, on the other hand, follows the extras approach proposed here https://docs.dask.org/en/stable/install.html:

image

One could argue that the extras approach is more maintainer-friendly (since no extra machinery or release is needed), and the `-core` package approach is more user friendly (since `pip install kedro` will continue working as expected).

...So, following @/tiangolo 's [advice on how he built FastAPI](https://youtu.be/mwvmfl8nN_U?si=FHwFL2XTqTZJTYrV&t=597), I always optimize for the user 🙂 hence I think the `kedro-core` approach is superior.

tiangolo

DimedS:
I think it looks promising, if we can implement this in 1.* series without breaking change, this will solve main issue @sbrugman highlighted in Ordeq tech design #5178

kedro/framework/cli/_ephemeral.py
deepyaman: > One could argue that the extras approach is more maintainer-friendly (since no extra machinery or release is needed), and the `-core` package approach is more user friendly (since `pip install kedro` will continue working as expected).
> 
> ...So, following @/tiangolo 's [advice on how he built FastAPI](https://youtu.be/mwvmfl8nN_U?si=FHwFL2XTqTZJTYrV&t=597), I always optimize for the user 🙂 hence I think the `kedro-core` approach is superior.

I support this.

deepyaman:
> Understood, thank you.
> 
> In case it helps, I think there could be other ways to keep the current behavior in a non-breaking way. For example:
> 
> * Make a new `kedro-core` package that has minimal dependencies
> * Make `kedro` depend on `kedro-core` plus cookiecutter, gitpython, rich, build
> * So, upon `pip install kedro` everything stays the same, but `pip install kedro-core` gives the minimal functionality

+1 to the alternative proposed by @astrojuanlu 

To be explicit, my understanding is that a project generated with `kedro new` will then depend on `kedro-core` (and not `kedro`), but the CLI commands will come from the `kedro-core` package.

The one weird bit is that you have to recommend `uvx kedro pipeline create`, right, since that depends on Cookiecutter? But if a user _really_ wanted to avoid that, they could always `pip install kedro` and accept the heavier dependencies. Or I guess you can detect if `uvx` or `pipx` is installed when the user invokes the CLI command, and if so redirects to the `uvx` or `pipx` version, something like what you did.

> When users run `kedro new` or `kedro pipeline create` without these extras installed, Kedro automatically falls back to using uvx to temporarily install the required tools for that command.

rashidakanchwala:
Thanks @deepyaman and @astrojuanlu.

So the ideal flow here would probably be:

Users who want a dependency-light install can do `pip install kedro-core` and for commands like `kedro new` or `kedro pipeline create`, we can recommend using uvx or pipx to run them with dependencies installed ephermally. 

Also if someone installs `kedro-core` and then runs `kedro new`, we could just raise a clear message like:

“This command requires extra dependencies. You can install them ephemerally via uvx/pipx, or run `pip install kedro` to include all creation dependencies <list deps>.”

Most people so far seem a bit unsure about this PR pattern, where Kedro automatically detects missing libraries and calls uvx through a subprocess. It’s probably better to let users trigger this themselves rather than having Kedro do it automatically.

astrojuanlu:
Notice that `pip install kedro-core` is not the same as `pip install kedro[core]`, could you clarify your message @rashidakanchwala ?

Issue #2690, opened 2023-06-15
Document more advanced hook use cases


astrojuanlu:
## Description
Hooks are stateful objects, which enables users to, for example, store the context in the `after_context_created` hook and use it later in a hook that doesn't receive it:

```python
class SomeHook:

@hook_spec
def after_context_created(self, context):
   ...
   self.my_azure_config = xxx

@hook_spec
def before_pipeline_create(self, pipeline):
   do_something_about_my_pipeline(pipeline, self.my_azure_config)
```

(code sample by @noklam)

We should better document this.

## Context
Follow up from gh-506 and other discussions, see for example https://www.linen.dev/s/kedro/t/12112601/does-anyone-know-if-there-is-a-reason-why-we-could-not-pass-#4e36a67f-36d3-4354-a3d5-4347a59ef28f

This is very useful when migrating projects from older versions of Kedro, customize pipeline execution, and more.

noklam:
I thought I have created a ticket for this but I didnt. Thanks for creating this!

noklam:
We need to explain WHY user need to do this, and examples of how to do it.

noklam:
https://github.com/Galileo-Galilei/kedro-mlflow/blob/845ad919c9dbd020e948e8adc2e0f9064de1ef68/kedro_mlflow/framework/hooks/mlflow_hook.py#L50-L63 is a good example. This get asked in the intermediate training, so maybe we can create an example that showcase this.

astrojuanlu:
Just today I wanted to apply such an "advanced" hook use case: storing the catalog and then injecting datasets on the fly. However, it doesn't work:

```python
class MissingDatasetHooks:
    @hook_impl
    def after_catalog_created(self, catalog: DataCatalog):
        self._catalog = catalog

    @hook_impl
    def before_dataset_loaded(self, dataset_name):
        dataset = self._catalog._get_dataset(dataset_name)
        try:
            dataset.load()
        except DataSetError:
            # Create EmptyDataset on the fly
            logger.warning("Attempted to load dataset %s which doesn't exist yet, injecting it", dataset_name)
            missing_dataset = MissingDataSet(dataset=dataset)
            self._catalog.add(dataset_name, missing_dataset, replace=True)
```

the `self._catalog` that gets saved receives the `.add(..., replace=True)` correctly, but the `catalog.load` that comes immediately after the `before_dataset_loaded` hook still has the old dataset:

[...]

PR #5222, opened 2025-11-24
docs: Add documentation for stateful hooks pattern (Issue #2690)

vivekkumarrathour:
## Description

This PR addresses #2690 by documenting how hooks are stateful objects and can be used to store context for use across different hook methods.

Added a new section "Using stateful hooks to store context" to `docs/extend/hooks/examples.md` that:
- Explains the concept of stateful hooks
- Provides a code example showing how to store context in `after_context_created` and use it in `before_pipeline_run`
- Lists the benefits of using this pattern
- Explains common use cases (migrating projects, customizing execution, cross-cutting concerns)

## Development notes

This is a documentation-only change that adds 38 lines to the hooks examples file. No code changes were made. The example follows the same format as existing hook examples in the file.


docs/extend/hooks/examples.md
```diff
@@ -387,3 +387,41 @@ Node input overwrites implemented in `before_node_run` affect only a specific no
 
 
 Once you have implemented a new Hook you must register it as described in the [Hooks documentation](./introduction.md#registering-the-hook-implementation-with-kedro), and then run Kedro.
+
+## Using stateful hooks to store context
```
ravi-kumar-pilla, line 391: Hi @vivekkumarrathour , Thank you for contributing. I appreciate the pattern documented here, however I feel this to be a common practice when you create a class and encapsulate hook implementations. 

It would be nice to have a use-case which adds additional functionality to your kedro project (like the ones mentioned above) through hooks and may be use the pattern described here in doing so. 

wdyt @ElenaKhaustova @rashidakanchwala 

Thank you

rashidakanchwala: I like this but I think this section should be on this page- https://docs.kedro.org/en/1.1.1/extend/hooks/common_use_cases/ 
The hooks example are proper use cases examples. So it doesn't fit here

merelcht:
Thanks for the contribution @vivekkumarrathour ! I agree with @rashidakanchwala that this fits better on the page with common use cases: https://docs.kedro.org/en/1.1.1/extend/hooks/common_use_cases/.

Could you also have a look at the `Vale` comments? These give guidance on our best practices for language use in our docs. You can read more about that in our docs styling guide: https://github.com/kedro-org/kedro/wiki/Kedro-documentation-style-guide

PR #5260, opened 2025-12-08
Fix broken doc links

ravi-kumar-pilla:
## Description

Resolves https://github.com/kedro-org/kedro/issues/5261

## Development notes

- Fixed broken reference links in codebase
- Added a linkchecker using lychee in `lint.yml`

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -69,7 +69,7 @@ test = [
     "pytest-xdist[psutil]>=2.2.1,<4.0",
     "pytest>=7.2,<10.0",
     "requests_mock",
-    "s3fs>=2021.4,<2025.11",
+    "s3fs>=2021.4,<2025.13",
     # mypy related dependencies
     "pandas-stubs",
     "types-PyYAML",


ravi-kumar-pilla:
Hi Team,

We have a partial linkchecker implementation here - https://github.com/kedro-org/kedro/blob/main/Makefile#L40
which we are not using anywhere (or I could not find it). Should we remove this ? Now that we have introduced a github [job](https://github.com/kedro-org/kedro/pull/5260/files#diff-107e910e9f2ebfb9a741fa10b2aa7100cc1fc4f5f3aca2dfe78b905cbd73c0d2R43) for linkcheck within codebase or should we keep it to run locally ? If we need to run locally, we need to install `lychee` in local env. I can clean it up if we want to remove this. 

cc: @merelcht @ElenaKhaustova 

Thank you

merelcht:
We also have the MkDocs linkcheck job (docs-linkcheck.yml) - what does that check then if it's not catching these errors? Only internal references? Maybe @lrcouto knows? 

If these checks test separate things, it might still be good to add them in the same job.

ElenaKhaustova:
> We also have the MkDocs linkcheck job (docs-linkcheck.yml) - what does that check then if it's not catching these errors? Only internal references? Maybe @lrcouto knows?
> 
> If these checks test separate things, it might still be good to add them in the same job.

I think it only checks internal links, as inside the job, it only tries to build documents in strict mode. I wonder if it is supposed to be extended by other checks as well. Otherwise, the job name is confusing.

Edit

This is what it actually does: `# this checks: mkdocs.yml is valid, all listed pages exist, plugins are correctly configured, no broken references in nav or Markdown links (internal), broken links and images (internal, not external)`

ElenaKhaustova:
> Hi Team,
> 
> We have a partial linkchecker implementation here - https://github.com/kedro-org/kedro/blob/main/Makefile#L40 which we are not using anywhere (or I could not find it). Should we remove this ? Now that we have introduced a github [job](https://github.com/kedro-org/kedro/pull/5260/files#diff-107e910e9f2ebfb9a741fa10b2aa7100cc1fc4f5f3aca2dfe78b905cbd73c0d2R43) for linkcheck within codebase or should we keep it to run locally ? If we need to run locally, we need to install `lychee` in local env. I can clean it up if we want to remove this.

Are we sure it replaces`mkdocs build --strict` fully? If yes, we probably should remove.

lrcouto:
> We also have the MkDocs linkcheck job (docs-linkcheck.yml) - what does that check then if it's not catching these errors? Only internal references? Maybe @lrcouto knows?
> 
> If these checks test separate things, it might still be good to add them in the same job.

For linkchecking external links, you can use the linkcheck command on the Makefile, that also runs lychee. We previously decided to not put it on CI because there's a lot of things like websites that return 403 when lychee tries to check them and thus count as errors. Maybe it needs a specific job or configuration for this, but it returns a lot of "false positives".

Though I'm seeing now that the link that was used as an example is an internal one? That should've been caught by the strict build.

ravi-kumar-pilla:
> We also have the MkDocs linkcheck job (docs-linkcheck.yml) - what does that check then if it's not catching these errors? Only internal references?

@merelcht .. I think the `docs-linkcheck.yml` can validate broken links in the markdown files and not the python code, so it is called docs-linkcheck :) . 

If you have something like [this](https://github.com/kedro-org/kedro/pull/5260/files#diff-cc4e916905ba553d257700ff851973772c0d77018d7673a2c4b22f309d8c3e46R152), I don't think `mkdocs build` can help with it. We need a custom linkchecker that scans python code for link references and flag incase of a 404 or use lychee.

> Are we sure it replaces mkdocs build --strict fully? If yes, we probably should remove.

@ElenaKhaustova .. No it does not replace `mkdocs build`. I was asking if we still need the [make command](https://github.com/kedro-org/kedro/blob/main/Makefile#L40) which we are not using anywhere apart from may be running locally as @lrcouto pointed out.

> We previously decided to not put it on CI because there's a lot of things like websites that return 403 when lychee tries to check them and thus count as errors.

@lrcouto ... make sense. I too encountered this but tried to add few links to [ignore](https://github.com/kedro-org/kedro/pull/5260/files#diff-fafe8d160920b6bd97b3c5235607969496ac0550eb4eeb9833393711fc210f18) . I tested locally and it works fine. For now the CI is working as expected, but yes it does check for external links and it is hard to find and ignore every link. The good thing of having this in our CI is we can avoid this situation next time though it means we will have some flaky tests, which should be easy to fix.

I would recommend we add lychee to our workflow, either in this ticket or may be I will create a separate one. If we want to have this in the current PR, let me know if it is fine to have this within lint.yml. Thank you

ElenaKhaustova:
@ravi-kumar-pilla, thanks for explaining!

QQ - Does it work fine if we refer to the page in the docs that is being added in the same PR?

> let me know if it is fine to have this within lint.yml

I think this is fine, since it only checks links in Python files

> I would recommend we add lychee to our workflow

We can add it as a pre-commit hook, probably in a separate PR

ElenaKhaustova:
I would keep make command for local testing as well

ravi-kumar-pilla:
> Does it work fine if we refer to the page in the docs that is being added in the same PR?

Good question. I think it will fail as I the link request is made to the hosted site and not the current build. I will try to run lychee against the current `mkdocs build` site. Let me update the test. Thank you

ravi-kumar-pilla:
Hi @ElenaKhaustova , 

I updated the check and it now runs against the current docs build - https://github.com/kedro-org/kedro/pull/5260/files#diff-9508f368888660e5d5076fb211974e43e12c9a0595c366d5eb425e3df5534418R48

Let me know if you have further questions. Thank you

merelcht (approved):
Thanks @ravi-kumar-pilla ! ⭐

PR #5290, opened 2025-12-18
Implement list_versions for versioned datasets

mvhensbergen:
## Description

We want to be able to programmatically inspect the different versions of datasets generated by batch training for versioned datasets. It seems there is no API for this and the workaround is to do directory listing of the dataset path directory to figure this out outside of Kedro.

The AbstractDataset class already has functionality to traverse this directory (and also pick the latest version) so it makes sense to offer the listing of versions of a dataset as a method.

Sample use case:
```
from kedro.io import Version
from kedro_datasets.json import (
    JSONDataset,
)

filename = "mydata.json"

# Init dataset for inspection
ds = JSONDataset(filepath=filename)

# Get available versions
versions = ds.list_versions(full_path=False)
print(versions)

# Select a specific version
v = versions[2]

# Reinitialize dataset with that version
v = Version(load=v,save=None)
ds = JSONDataset(filepath=filename, version=v)

# Load
data = ds.load()
print(data)
```

## Development notes

I have moved the path traversal logic to a separate function for reuse and have added a public method 'list_versions' which either returns the full paths of all the different versions (default) or only the versions if `full_path=False`.

Some other musings which you may guide me on:
1. I am aware that the `list_versions` functionality might not be compatible with the philosophy of that a Dataset instance should always have a certain version. At the same time, I believe that the Dataset class is the right place for the implementation of this feature as it has all the code to do this already. What is your opinion on this?
2. In that light, would the Dataset class be more generic if it is always versionless, and that the particular version can be specified by the `load` method? In that way the DataCatalog (which also can be supplied with version parameters) would explicitly call `load` with that version, yet the Dataset instance itself is more generic.
3. I have made the directory traversal uncached now, is this correct?

## Checklist

- [v ] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [v ] Signed off each commit with a [Developer Certificate of Origin (DCO)](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository)
- [ v] Opened this PR as a 'Draft Pull Request' if it is work-in-progress

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,7 +68,7 @@ test = [
     "pytest-xdist[psutil]>=2.2.1,<4.0",
     "pytest>=7.2,<10.0",
     "requests_mock",
-    "s3fs>=2021.4,<2026.2",
+    "s3fs>=2021.4,<2026.5",
     # mypy related dependencies
     "pandas-stubs",
     "types-PyYAML",


mvhensbergen:
I'd like to wait for the documentation until I am certain that the idea of adding the list_versions API is approved. :-)

DimedS:
thanks for your contribution, @mvhensbergen ! We will try to review it soon!

kedro/io/core.py
```diff
@@ -752,6 +748,18 @@ def _fetch_latest_load_version(self) -> str:
                 f"due to insufficient permission. Exception: {exc}"
             )
             raise VersionNotFoundError(message) from exc
+
+        return [path for path in version_paths if self._exists_function(path)]
```
DimedS, line 754: Are we duplicating logic around line 764 when computing `most_recent`?  
It looks like `_get_versions()` already filters existing paths, so the additional `exists` check here may be redundant.

kedro/io/core.py
```diff
@@ -810,6 +818,14 @@ def _get_save_path(self) -> PurePosixPath:
     def _get_versioned_path(self, version: str) -> PurePosixPath:
         return self._filepath / version / self._filepath.name
 
+    def list_versions(self, full_path: bool = True) -> list:
+        version_paths = self._get_versions()
+
+        if full_path:
+            return version_paths
+
+        return [path.split(os.sep)[-2] for path in version_paths]  # [-2] is version dir
```
DimedS: It would be good to add an end-to-end test covering this behaviour, as this line may cause issues on Windows due to path separator differences.

SajidAlamQB:
Thanks @mvhensbergen for this contribution! The `list_versions` looks good. As @DimedS noted there are a few things to address before we can merge. 

Since `_get_versions()` now filters paths with `_exists_function`, the check in `_fetch_latest_load_version` is redundant.

Also maybe adding a test for the empty versions case too.

kedro/io/core.py
ElenaKhaustova: Agree with @DimedS, also we should probably move it to a separate function, like `_get_version_from_path` and then test it properly.

ElenaKhaustova:
```suggestion
        rreturn [PurePath(path).parent.name for path in version_paths]
```

mvhensbergen:
Hi all,

thank you for your suggestions. I have made a few changes in the last commit. Unfortunately, due to some (temporary) changes in work/life balance I have less time to complete it all completely.

I did:
- emoved spurious .exists check
- make separate function for get version as suggested in the review

I didn't yet:
- change appropriate documentation
- think how to implement this for other versioned sets.

It would be great if you could help a bit in finalizing this PR with me!

kedro/io/core.py
```diff
@@ -752,9 +748,19 @@ def _fetch_latest_load_version(self) -> str:
...
             raise VersionNotFoundError(message) from exc
-        most_recent = next(
-            (path for path in version_paths if self._exists_function(path)), None
-        )
+
+        return [path for path in version_paths if self._exists_function(path)]
+
+    # 'key' is set to prevent cache key overlapping for load and save:
+    # https://cachetools.readthedocs.io/en/stable/#cachetools.cachedmethod
+    @cachedmethod(cache=attrgetter("_version_cache"), key=partial(hashkey, "load"))
```
DimedS: we removed cachetools as I remember?

DimedS (approved):
Thanks @mvhensbergen and @rashidakanchwala , LGTM!

rashidakanchwala:
thanks @mvhensbergen for your contribution. I will merge this PR now :)

PR #5320, opened 2026-01-09
Fix starters tests after Spark starter changes

DimedS:
After the recent updates to the Spark starter, selecting the PySpark tool (`6` / `pyspark`) no longer adds `spark.yml` and `hooks.py` to the generated project.

This PR updates the expected file counts in the starters tests so that PySpark selection no longer contributes additional files.


ElenaKhaustova (approved):
Thank you!

PR #5328, opened 2026-01-12
Add support for user-injectable node preview functions

### LINKED ISSUE #5317: Implement user-injectable node preview function
## Description

Enable Kedro nodes to expose a user-defined preview function (similar to `dataset.preview()`), to support visualizing complex objects such as compiled LangGraph graphs or ML pipelines.

## Context

Currently, only datasets support previews. For projects that wrap complex agentic or orchestration frameworks (e.g., LangGraph, LangChain), it’s valuable to visualize intermediate compiled objects.

See an example of such project here: https://github.com/kedro-org/kedro-academy/tree/main/kedro-agentic-workflows
KEP for the proposal: https://github.com/kedro-org/kedro/discussions/5309

## Possible Implementation
- Create an experimental feature that provides extensible preview mechanism for a kedro node while maintaining backward compatibility.
- Add an optional `preview_fn` parameter to `kedro.pipeline.node()` and expose a `node.preview()` API.

```python
def visualize_langgraph(graph):
    return graph.get_graph().draw_mermaid()

node(
    func=compile_langgraph,
    inputs=["config"],
    outputs="compiled_graph",
    name="compile_langgraph",
    preview_fn=visualize_langgraph,
)
```

**Benefits:**

- Richer interactivity in Kedro-Viz without the need to support cycles in the graph.
- Extensible preview mechanism for any arbitrary node output.
- Enables visual inspection of agentic workflows (e.g., LangGraph).

ravi-kumar-pilla:
## Description

Resolves #5317 

## Development notes

The preview feature allows nodes to define a `preview_fn` that generates visualizations of node execution results. Here's what was implemented:

- Added `PreviewPayload` as a data class which represents the structured output from preview functions
- Added an optional `preview_fn` parameter when creating nodes that accepts a callable and returns a PreviewPayload.
- Warns the user with `KedroExperimentalWarning` when someone passes `preview_fn` to node initialization.
- Added `node.preview()` method which executes the preview function if available. It also validates that the return value is a `PreviewPayload` instance.
- Updated tests

**Example Usage:**

```py
def my_preview_fn():
    return PreviewPayload(kind="json", content='{"result": "data"}')

node(func, inputs="data", outputs="result", preview_fn=my_preview_fn)
```

More info in the KEP: https://github.com/kedro-org/kedro/discussions/5309

## Checklist

### DIFF
--- a/kedro/pipeline/llm_context.py
+++ b/kedro/pipeline/llm_context.py
@@ -14,7 +14,7 @@ and loaded by Kedro before the node runs. Tools are instantiated at execution
 time and automatically assigned readable names based on the returned objects.
 """
 
-from collections.abc import Callable, Sequence
+from collections.abc import Callable, Iterable, Sequence
 from dataclasses import dataclass, field
 from typing import Any, NamedTuple, TypeVar
 
@@ -143,7 +143,7 @@ class LLMContextNode(Node):
     ```
     """
 
-    def __init__(
+    def __init__(  # noqa: PLR0913
         self,
         *,
         outputs: str,
@@ -151,6 +151,9 @@ class LLMContextNode(Node):
         prompts: list[str],
         tools: list[_ToolConfig] | None = None,
         name: str | None = None,
+        tags: str | Iterable[str] | None = None,
+        confirms: str | list[str] | None = None,
+        namespace: str | None = None,
     ):
         """Create an LLMContextNode.
 
@@ -160,6 +163,10 @@ class LLMContextNode(Node):
             prompts: List of dataset names containing prompt content.
             tools: Optional list of tool configurations created via `tool(...)`.
             name: Optional node name; also used as the logical context identifier.
+            tags: Optional set of tags to be applied to the node.
+            confirms: Optional name or the list of the names of the datasets
+                that should be confirmed.
+            namespace: Optional node namespace.
         """
         inputs = {"llm": llm}
 
@@ -201,18 +208,27 @@ class LLMContextNode(Node):
 
         # call the Node constructor with the func, inputs, outputs, name
         super().__init__(
-            func=construct_context, inputs=inputs, outputs=outputs, name=name
+            func=construct_context,
+            inputs=inputs,
+            outputs=outputs,
+            name=name,
+            tags=tags,
+            confirms=confirms,
+            namespace=namespace,
         )
 
 
 @experimental
-def llm_context_node(
+def llm_context_node(  # noqa: PLR0913
     *,
     outputs: str,
     llm: str,
     prompts: list[str],
     tools: list[_ToolConfig] | None = None,
     name: str | None = None,
+    tags: str | Iterable[str] | None = None,
+    confirms: str | list[str] | None = None,
+    namespace: str | None = None,
 ) -> Node:
     """
     !!! warning "Experimental"
@@ -229,7 +245,12 @@ def llm_context_node(
         prompts: List of dataset names containing prompt content.
         tools: Optional list of tool configurations created via `tool(...)`.
             Each tool declares the Kedro inputs required to construct it.
-        name: Optional name for the node and for the created context.
+        name: Optional node name; also used as the logical context identifier.
+        tags: Optional set of tags to be applied to the node.
+        confirms: Optional name or the list of the names of the datasets
+            that should be confirmed.
+        namespace: Optional node namespace.
+
 
     Returns:
         A Kedro Node that loads all declared datasets, instantiates tools,
@@ -256,4 +277,7 @@ def llm_context_node(
         prompts=prompts,
         tools=tools,
         name=name,
+        tags=tags,
+        confirms=confirms,
+        namespace=namespace,
     )
--- a/tests/pipeline/test_llm_context.py
+++ b/tests/pipeline/test_llm_context.py
@@ -2,6 +2,7 @@ import pytest
 
 from kedro.pipeline.llm_context import (
     LLMContext,
+    LLMContextNode,
     _get_tool_name,
     _normalize_outputs,
     llm_context_node,
@@ -185,3 +186,39 @@ class DummyClass:
 def test_get_tool_name_all_branches(obj, expected_name):
     """_get_tool_name should derive a stable, human-friendly tool name."""
     assert _get_tool_name(obj) == expected_name
+
+
+def test_llm_context_node_sets_tags():
+    """LLMContextNode should propagate tags to the underlying Node."""
+    node_obj = LLMContextNode(
+        outputs="out",
+        llm="llm",
+        prompts=[],
+        tags=["llm", "experimental"],
+    )
+
+    assert node_obj.tags == {"llm", "experimental"}
+
+
+def test_llm_context_node_sets_confirms():
+    """LLMContextNode should propagate confirms to the underlying Node."""
+    node_obj = LLMContextNode(
+        outputs="out",
+        llm="llm",
+        prompts=[],
+        confirms=["llm", "prompt"],
+    )
+
+    assert node_obj.confirms == ["llm", "prompt"]
+
+
+def test_llm_context_node_sets_namespace():
+    """LLMContextNode should propagate namespace to the underlying Node."""
+    node_obj = LLMContextNode(
+        outputs="out",
+        llm="llm",
+        prompts=[],
+        namespace="llm_nodes",
+    )
+
+    assert node_obj.namespace == "llm_nodes"


ravi-kumar-pilla:
Hi @ElenaKhaustova and @merelcht ,

I have added the implementation proposed in the KEP. Please review and if the code placement looks good, I will add docs, release notes to followup. Once this feature is released, we can plan on supporting mermaid to start with on Kedro-Viz.  

Thank you

ravi-kumar-pilla:
Hi @ElenaKhaustova ,

Since this is experimental, I haven't decided on the content data-type yet, but I would like to get your thoughts on the below stricter contract, since `content: str` is not ideal in all the scenarios for the renderer.

**1. Having separate contracts and encode the contract as a discriminated union -**

```py
from typing import Literal, Union, Any
from pydantic import BaseModel

class TextPreview(BaseModel):
    kind: Literal["text"]
    content: str

class MermaidPreview(BaseModel):
    kind: Literal["mermaid"]
    content: str

class JsonPreview(BaseModel):
    kind: Literal["json"]
    content: dict | list

class TablePreview(BaseModel):
    kind: Literal["table"]
    content: list[dict[str, Any]]

class PlotlyPreview(BaseModel):
    kind: Literal["plotly"]
    content: dict  # Plotly JSON

class ImagePreview(BaseModel):
    kind: Literal["image"]
    content: str  # url or base64
    meta: dict[str, Any] | None = None

class CustomPreview(BaseModel):
    kind: Literal["custom"]
    content: dict
    meta: dict[str, Any]

PreviewPayload = Union[
    TextPreview,
    MermaidPreview,
    JsonPreview,
    TablePreview,
    PlotlyPreview,
    ImagePreview,
    CustomPreview,
]
```

**2. Document the contract and have a validation in preview method:**

```py
@dataclass
class PreviewPayload:
    """
    Contract:
    - kind="text"     → content: str
    - kind="mermaid"  → content: str (Mermaid DSL)
    - kind="json"     → content: dict | list
    - kind="table"    → content: list[dict]
    - kind="plotly"   → content: dict (Plotly JSON schema)
    - kind="image"    → content: str (URL or base64), meta["mime"] optional
    - kind="custom"   → content: dict, meta["renderer"] required
    """
    kind: Literal["mermaid", "json", "text", "image", "plotly", "table", "custom"]
    content: Any
    meta: dict[str, Any] | None = None
```

```py
def validate_preview(payload: PreviewPayload) -> None:
    if payload.kind in {"text", "mermaid"} and not isinstance(payload.content, str):
        raise TypeError(f"{payload.kind} content must be str")

    if payload.kind == "json" and not isinstance(payload.content, (dict, list)):
        raise TypeError("json content must be dict or list")

    if payload.kind == "table":
        if not (
            isinstance(payload.content, list)
            and all(isinstance(row, dict) for row in payload.content)
        ):
            raise TypeError("table content must be list[dict]")

    if payload.kind == "plotly" and not isinstance(payload.content, dict):
        raise TypeError("plotly content must be dict")

    if payload.kind == "custom":
        if not isinstance(payload.content, dict):
            raise TypeError("custom content must be dict")
        if not payload.meta or "renderer" not in payload.meta:
            raise ValueError("custom requires meta['renderer']")

```

Thank you

kedro/__init__.py
```diff
@@ -24,9 +24,14 @@ class KedroPythonVersionWarning(UserWarning):
     """Custom class for warnings about incompatibilities with Python versions."""
 
 
+class KedroExperimentalWarning(UserWarning):
```
ElenaKhaustova: Can we reuse this one?

https://github.com/kedro-org/kedro/blob/d762f6c7dac544601d65538f0a6a8c1210230805/kedro/utils.py#L171

kedro/pipeline/node.py
```diff
@@ -170,6 +193,21 @@ def __init__(  # noqa: PLR0913
...
         self._confirms = confirms
 
+        if preview_fn:
+            if not callable(preview_fn):
+                raise ValueError(
+                    _node_error_message(
+                        f"preview_fn must be a function, not '{type(preview_fn).__name__}'."
+                    )
+                )
+            warn(
```
ElenaKhaustova: You may need to do something extra to make sure it only warn once

For example: https://github.com/kedro-org/kedro/blob/d762f6c7dac544601d65538f0a6a8c1210230805/kedro/utils.py#L228

kedro/pipeline/node.py
```diff
@@ -389,6 +428,32 @@ def confirms(self) -> list[str]:
         """
         return _to_list(self._confirms)
 
+    def preview(self, *args: Any, **kwargs: Any) -> PreviewPayload | None:
```
ElenaKhaustova: For simplicity, we agreed not to provide datasets as arguments in this case. However, simple serializable objects like str, list, dict, etc. should be possible. Can you please add some examples of what it will look like?

ElenaKhaustova:
@ravi-kumar-pilla, some thoughts about the contracts:

1. The list itself looks good and should be more than enough for now
2. Shall we add `kind`, so the discrimination is explicit?

```python
from typing import Annotated, Union
from pydantic import Field

PreviewPayload = Annotated[
    Union[
        TextPreview,
        MermaidPreview,
        JsonPreview,
        TablePreview,
        PlotlyPreview,
        ImagePreview,
        CustomPreview,
    ],
    Field(discriminator="kind"),
]
```
3. If we're using Pydantic, why do we need to validate manually with `validate_preview`?

kedro/pipeline/node.py
ravi-kumar-pilla: good catch 🗡️

kedro/pipeline/node.py
ravi-kumar-pilla: Yes, I think there is no way the user can pass args from node definition but if they want to access `preview` method of node directly and would like to pass args, they can do so. I added few examples now to make it clear. But let me know if this is confusing. Thank you

ravi-kumar-pilla:
> The list itself looks good and should be more than enough for now

Yes the issue is not about `kind` values but with `content`. The data-type of `content` is `str` which is not optimal for all values supported in `kind`. I do not want `content` to have `Any` data-type as it will be loosely typed.

> 3. If we're using Pydantic, why do we need to validate manually with `validate_preview`?

Yes, I mentioned 2 variations [here](https://github.com/kedro-org/kedro/pull/5328#issuecomment-3745261409)

1. With pydantic classes.  Issue I see here is kedro does not have pydantic dependency (atleast to my knowledge), so I am not sure if I can introduce it now.

2. The other variation is without pydantic classes, just having a docstring and then adding a validate_preview.

Thank you

kedro/pipeline/node.py
```diff
@@ -38,12 +41,29 @@ class GroupedNodes:
...
+    function should return to enable visualization of node execution results in
+    various formats.
+
+    Args:
+        kind: The type of preview content.
+        content: The actual preview content as a string.
+        meta: Optional metadata dictionary. Defaults to None.
+    """
+
+    kind: Literal["mermaid", "json", "text", "image", "plotly", "table", "custom"]
```
merelcht: Nit: but I'd probably name this `type` instead of `kind`. `type` is the more standard name to use.

kedro/pipeline/node.py
ElenaKhaustova: Yeah, it is a bit.

Perhaps we should include an explicit note in the docs and docstrings that this implementation is static, not dynamic. So the data cannot be passed dynamically.

ElenaKhaustova:
I do prefer validation with pydantic, rather than manual. But agree with Merel's point that we probably should not introduce extra dependencies with an experimental feature.

In order to move forward we can go with dataclass + manual validation for now. Keep it lightweight and easy to remove or replace with Pydantic later. The downside is that the content has to be typed as `Any`.

kedro/pipeline/node.py
ElenaKhaustova: `kind` is often used for data contracts, and it helps avoid conflicts with Python’s built-in type.

kedro/pipeline/node.py
ravi-kumar-pilla: I think for the initial release, lets remove it.

kedro/pipeline/node.py
ravi-kumar-pilla: @merelcht I thought about this initially and I agree with Elena about `conflicts with Python’s built-in type`. That was the reason I kept `kind` and I find this being used in many examples.

ravi-kumar-pilla:
Hi @merelcht and @ElenaKhaustova,

Thank you for the suggestions. Using pydantic would simplify the validation process, but since this is an experimental feature and I do not want to make pydantic a hard dependency for using an experimental feature, I went for dataclasses.

I added a [preview_contract](https://github.com/kedro-org/kedro/pull/5328/changes#diff-773a53a596e525c6604b9ae8b2a9b6fb5b8ba1cdf6e02a64bf615fd8982930ad) and this might be a bit too much for an experimental feature but I feel it serves -
1. A strict contract and a stable api for designing renderers in the frontend
2. Makes it easy to convert this into a long-term feature from experimental
3. Validations can be improved specific to each preview type. Easily extendable design.
4. Avoids errors with serialization and avoids using `content: Any`

Let me know your thoughts on this. 

**NOTE:** I will update tests if the new contract looks good.

Thank you

kedro/pipeline/node.py
merelcht: Okay! I'm happy with that.

kedro/pipeline/preview_contract.py
```diff
@@ -0,0 +1,303 @@
+from __future__ import annotations
```
ElenaKhaustova, line 1: Maybe we should move it out of `kedro.pipeline` to not overload it

kedro/pipeline/preview_contract.py
```diff
@@ -0,0 +1,303 @@
...
+
+@dataclass(frozen=True)
+class JsonPreview:
+    kind: Literal["json"]
+    content: JSONValue
+    meta: Meta | None = None
+
+    def __post_init__(self) -> None:
+        assert_json_value(self.content, "$.content")
+        if self.meta is not None:
```
ElenaKhaustova: Instead of repeating meta validation everywhere, can we introduce a small shared base helper?

```python
def _validate_meta(meta: Meta | None) -> None:
    if meta is not None:
        assert_json_value(meta, "$.meta")
```

kedro/pipeline/preview_contract.py
```diff
@@ -0,0 +1,303 @@
...
+    | JsonPreview
+    | TablePreview
+    | PlotlyPreview
+    | ImagePreview
+    | CustomPreview
+)
+
+
+# JSON serialization helpers
+def _dataclass_to_json_dict(payload: Any) -> JSONObject:
```
ElenaKhaustova: In almost every preview’s `__post_init__` we already validate json, and then we validate it here during serialisation. Can we simplify it and just do it once?

kedro/pipeline/preview_contract.py
```diff
@@ -0,0 +1,303 @@
...
+    Any,
+    Literal,
+    TypeAlias,
+    Union,
+    cast,
+)
+
+# JSON-safe type system
+JSONScalar: TypeAlias = str | int | float | bool | None
+JSONValue: TypeAlias = Union[JSONScalar, "JSONObject", "JSONArray"]
```
ElenaKhaustova, line 14: Maybe we can simplify it and accept something like

```python
JSONLike = dict[str, Any] | list[Any] | str | int | float | bool | None
```

And then just rely on `json.dumps()` to fail if something slips through. This then removes recursive traversal below.

kedro/pipeline/preview_contract.py
```diff
@@ -0,0 +1,303 @@
...
+            assert_json_value(self.meta, "$.meta")
+
+    def to_dict(self) -> JSONObject:
+        return _dataclass_to_json_dict(self)
+
+
+@dataclass(frozen=True)
+class ImagePreview:
+    kind: Literal["image"]
+    content: ImageContent
```
ElenaKhaustova: Leave it for you to decide, but maybe we can simplify image content too.

```python
@dataclass
class ImagePreview:
    kind: Literal["image"]
    content: str  # url or base64
    meta: {"mime": "..."} optional
```

Which is what you already had originally. The current design is cleaner architecturally, but arguably too strict for previews.

kedro/pipeline/preview_contract.py
```diff
@@ -0,0 +1,303 @@
...
+
+    rows: list[dict[str, JSONValue]]
+    columns: list[ColumnDef] | None = None
+    row_id_key: str | None = None
+
+    def __post_init__(self) -> None:
+        if not isinstance(self.rows, list):
+            raise TypeError("TableContent.rows must be a list")
+
+        for i, row in enumerate(self.rows):
```
ElenaKhaustova: Do we need column definitions, row IDs, and strict JSON validation for a preview, or can this evolve later? This is well designed but also heavy.

kedro/pipeline/preview_contract.py
ravi-kumar-pilla: Can you recommend any other place to put this ? I am not sure if any other folder fits a node preview contract ?

kedro/pipeline/preview_contract.py
ravi-kumar-pilla: I will simplify the `assert_json_value` to use `json.dumps` instead of the recursive logic. But I will keep the static type checking to avoid using `Any` at construction time and for static type checkers

ravi-kumar-pilla:
Hi @ElenaKhaustova , 

As discussed, I kept the contract strict for designing stable renderers. I addressed most of your comments, and few nit ones are left. I am working on `kedro-viz` renderers to consume and display the preview payload. 

**Next Steps:**
1. Once the PR on kedro-viz is ready, I will test this end-end. But I think this PR should be good to go if approved.
2. I will add docs in a separate PR

Thank you

merelcht:
Agree with @ElenaKhaustova's comment on this feeling heavy, but I also think that's because we don't have a lot of similar functionality in Kedro where validation is so important. One suggestion I have is to make a `BasePreview` class the other classes can inherit from to reduce duplication. Something like:

```python
@dataclass(frozen=True)
class BasePreview(ABC):
    """Base class for all preview types with shared functionality."""

    meta: Meta | None = None

    def __post_init__(self) -> None:
        """Validate meta field if provided."""
        if self.meta is not None:
            assert_json_value(self.meta, "$.meta")

    def to_dict(self) -> JSONObject:
        """Convert preview to JSON-serializable dict."""
        if not is_dataclass(self) or isinstance(self, type):
            raise TypeError(f"Not JSON-serializable: {type(self).__name__}")
        return asdict(self)
```

And then your classes become a lot simpler like:

```python
@dataclass(frozen=True)
class MermaidPreview(BasePreview):
    content: str
    kind: Literal["mermaid"] = field(default="mermaid", init=False)

    def __post_init__(self) -> None:
        super().__post_init__()
        _validate_type(self.content, str, "MermaidPreview.content")
```

Posted the whole snippet here:
[preview_contract.py](https://github.com/user-attachments/files/24667449/preview_contract.py)

kedro/pipeline/preview_contract.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+@dataclass(frozen=True)
+class TablePreview:
+    content: list[dict[str, JSONValue]]
+    meta: Meta | None = None
+    kind: Literal["table"] = field(default="table", init=False)
+
+    def __post_init__(self) -> None:
+        if not isinstance(self.content, list):
```
ElenaKhaustova: Nit: Can we move this to a separate function, so `TablePreview.__post_init__ `becomes two lines.

```python
def _validate_table_rows(rows: list[dict[str, JSONValue]]) -> None:
    for i, row in enumerate(rows):
        if not isinstance(row, dict):
            raise TypeError(f"Table content row[{i}] must be dict")
        if not all(isinstance(k, str) for k in row.keys()):
            raise TypeError(f"Table content row[{i}] keys must be str")
```

kedro/pipeline/preview_contract.py
```diff
@@ -0,0 +1,181 @@
...
+    def __post_init__(self) -> None:
+        if not isinstance(self.content, str):
+            raise TypeError("TextPreview.content must be str")
+        _validate_meta(self.meta)
+
+    def to_dict(self) -> JSONObject:
+        return _dataclass_to_json_dict(self)
+
+
+@dataclass(frozen=True)
```
ElenaKhaustova, line 93: Shall we combine `TextPreview` and `MermaidPreview` base class to reduce near-duplicate code?

Or, as suggested by @merelcht try to create a common base class for all.

kedro/pipeline/preview_contract.py
```diff
@@ -0,0 +1,181 @@
...
+class TextPreview:
+    content: str
+    meta: Meta | None = None
+    kind: Literal["text"] = field(default="text", init=False)
+
+    def __post_init__(self) -> None:
+        if not isinstance(self.content, str):
+            raise TypeError("TextPreview.content must be str")
+        _validate_meta(self.meta)
```
ElenaKhaustova: I was thinking that right now, every preview validates its meta in `__post_init__`. This happens as soon as the object is created, even if nobody ever serializes it.

Shall we, instead of validating meta in every `__post_init__ ` validate it once in `_dataclass_to_json_dict()` before sending to the frontend?

Cause in parallel, we're working on shortening startup time, too.

kedro/pipeline/preview_contract.py
ElenaKhaustova: Maybe to `kedro/`?

Do we expect them to be used anywhere else but pipeline?

ravi-kumar-pilla:
> Posted the whole snippet here: [preview_contract.py](https://github.com/user-attachments/files/24667449/preview_contract.py)

Thank you @merelcht for the suggestions. I made the changes. 

>  JSONObject is already dict[str, JSONValue], so we only need JSON validation

I modified the above instances and added back the runtime validation in post_init as the datatype mentioned in dataclass is for static checking and it does not validate anything at runtime (unless we use pydantic). I kept rest of the changes as is.

Thank you

kedro/pipeline/preview_contract.py
ravi-kumar-pilla: Yes I created a common base class as per Merel's suggestion. Since we need to inform the kind to frontend, we cannot club both but we can also create another TextBasePreview (which might be again too much). 

So I am keeping them separate but also added BasePreview for all.

kedro/pipeline/preview_contract.py
ravi-kumar-pilla: I cannot think of any other case but since this is related to node previews, I placed it closer to node.py. @merelcht any suggestions on the file placement ?

ElenaKhaustova: Let's leave it then.

I don't like it being there, but I also don't want to create a separate folder in the structure.

kedro/pipeline/preview_contract.py
```diff
@@ -0,0 +1,161 @@
...
+    try:
+        json.dumps(data)
+    except (TypeError, ValueError) as e:
+        raise TypeError(
+            f"{path}: value is not JSON-serializable, got {type(data).__name__}"
+        ) from e
+
+
+@dataclass(frozen=True)
+class BasePreview(ABC):
```
ElenaKhaustova: Nit: `BasePreview` doesn’t currently define any abstract methods, so inheriting from ABC isn’t strictly necessary. So it could be dropped for simplicity.

kedro/pipeline/preview_contract.py
merelcht: Yes let's leave it for now. I don't have a good alternative in mind either.

PR #5340, opened 2026-01-22
Do not load all pipelines when only one is run.

lrcouto:
## Description

A proposed solution for https://github.com/kedro-org/kedro/issues/5276

This will pass an optional parameter with a pipeline's name to `find_pipelines()`, through `register_pipelines()`, which will then try import only the requested pipeline instead of all of the ones available in the project.

```py
    if requested_pipelines is not None and requested_pipelines:
        for pipeline_name in requested_pipelines:
            pipeline_module_name = f"{PACKAGE_NAME}.pipelines.{pipeline_name}"
            try:
                pipeline_module = importlib.import_module(pipeline_module_name)
```

This does require changes on the `register_pipeline()` function to work - the new `pipeline` parameter needs to be passed. However, inspecting the registry function before attempting to call it to check if the parameter is present, and only passing this pipeline name if it is, should ensure that this is backwards compatible. It should not break any existing projects, they'll continue working as they did before.

## Developer Certificate of Origin
We need all contributions to comply with the [Developer Certificate of Origin (DCO)](https://developercertificate.org/). All commits must be signed off by including a `Signed-off-by` line in the commit message. [See our wiki for guidance](https://github.com/kedro-org/kedro/wiki/Guidelines-for-contributing-developers/).

If your PR is blocked due to unsigned commits, then you must follow the instructions under "Rebase the branch" on the GitHub Checks page for your PR. This will retroactively add the sign-off to all unsigned commits and allow the DCO check to pass.

## Checklist

- [ ] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [ ] Signed off each commit with a [Developer Certificate of Origin (DCO)](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository)
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [ ] Added tests to cover my changes
- [ ] Checked if this change will affect Kedro-Viz, and if so, communicated that with the Viz team

PR #5296, opened 2025-12-27
add support for running multiple pipelines via --pipelines

Mohmn:
## Description 
fixes #4950

## Developer Certificate of Origin
We need all contributions to comply with the [Developer Certificate of Origin (DCO)](https://developercertificate.org/). All commits must be signed off by including a `Signed-off-by` line in the commit message. [See our wiki for guidance](https://github.com/kedro-org/kedro/wiki/Guidelines-for-contributing-developers/).

If your PR is blocked due to unsigned commits, then you must follow the instructions under "Rebase the branch" on the GitHub Checks page for your PR. This will retroactively add the sign-off to all unsigned commits and allow the DCO check to pass.

## Checklist

- [x] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [ ] Signed off each commit with a [Developer Certificate of Origin (DCO)](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository)
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [x] Added tests to cover my changes
- [ ] Checked if this change will affect Kedro-Viz, and if so, communicated that with the Viz team

PR #5346, opened 2026-01-31
Prototype bottleneck pipelines alternate option

### LINKED ISSUE #5276: Investigate and propose solution to performance bottleneck in KedroSession
Follow up to https://github.com/kedro-org/kedro/issues/5247

During the spike, we profiled a performance bottleneck in KedroSession, described in detail here: https://github.com/kedro-org/kedro/issues/5247#issuecomment-3644337136

KedroContext and the `load_context()` method we also investigated during this spike and also load all of the context into memory before the session run starts. But it's less likely that it'd cause significant slowness on a typical use case. We decided to focus on the KedroSession initially.

KedroSession retrieves pipelines through the global pipelines object from `kedro.framework.project`, which is an instance of `_ProjectPipelines`. This object is designed as a lazily loaded, dict-like interface that defers pipeline loading until first access. However, the current lazy-loading implementation eagerly loads all pipelines on first access to any pipeline key.

---

The current behavior is the following:

- `_ProjectPipelines` starts empty and uninitialized.
- `bootstrap_project()` → `configure_project()` sets `_pipelines_module` but does not load data.
- When `KedroSession.run()` accesses `pipelines[name]`, the `__getitem__` call is wrapped by `_load_data_wrapper`.
- _load_data() imports the pipelines registry module and calls register_pipelines().
- `register_pipelines()`, by default, calls `find_pipelines()`, which constructs all pipelines and returns a full dictionary.
- The entire pipelines dictionary is loaded into memory before the requested pipeline is returned.

This means that even when a single pipeline is requested, all pipelines are instantiated first.

---

The objective of this issue is to go deeper into the investigation into how pipelines are loaded during KedroSession initialization and execution, identify where unnecessary overhead is introduced (particularly when only a single pipeline is requested), and evaluate whether the current lazy-loading design achieves its intended performance benefits.

The expected outcome is a concrete, well-scoped proposal for improving performance, potentially by enabling more granular or truly lazy pipeline loading, along with an assessment of trade-offs.

Related tickets - #2879

lrcouto:
## Description

Full context: https://github.com/kedro-org/kedro/issues/5354

Alternate (and maybe slightly cleaner) solution to https://github.com/kedro-org/kedro/pull/5340.

The introduction of the option to run multiple pipelines with the `--pipeline`  flag in the CLI (https://github.com/kedro-org/kedro/pull/5296) also caused the pipeline registry to be accessed multiple times in a single session. This led to some issues due to the `_load_data()` function in `_ProjectPipelines` not having been designed to have "partial loading" states. The way it is working right now, in the main branch, also causes all pipelines to be loaded for each specific pipelines requested.

This solution aims to prevent that by passing the full comma-separated string with all the requested pipelines through the registry, and letting the `find_pipelines()` function handle parsing it and loading the specific pipelines requested by the user. This way the registry would be accessed only once, and the pipelines would be loaded only once, and only the ones that the user specifically requested.

## Checklist


lrcouto:
Closed in favor of https://github.com/kedro-org/kedro/pull/5401

PR #5352, opened 2026-02-02
Remove `cachetools` dependency

rashidakanchwala:
## Description

`Cachetools` 7 introduced several breaking changes that made `@cachedmethod` significantly stricter, including converting it into a descriptor and requiring instances to have a mutable `__dict__`. 
These assumptions do not align well with Kedro’s `ParallelRunner`, which relies on pickling and unpickling dataset instances across processes, and this resulted in runtime failures when upgrading to `cachetools` 7. Although pinning the dependency avoided the immediate issue, further investigation showed that `cachetools` is not well-suited to this use case. Kedro only caches two values per dataset instance (the resolved load and save versions) and does not require eviction policies, TTLs, or multi-key caching.

As a long-term solution, this change removes the `cachetools` dependency and replaces it with lightweight manual caching. This simplifies the implementation, removes an unnecessary dependency, and avoids future breakages in parallel execution.

## Development notes
- Replaced `@cachedmethod` usage with manual per-instance caching for resolved load ~~and save~~ versions.
- ~~Added explicit thread-safety using a per-instance lock to ensure correct behavior under `ThreadRunner`~~
- ~~Preserved multiprocessing compatibility by excluding the lock from pickling via `__getstate__` and `__setstate__`, ensuring ParallelRunner continues to work correctly.~~
- `Cachetools` is still currently used in `kedro-datasets` for `PartitionedDataset` and `IncrementalDataset`. Another change to remove this dependency there is tracked in the related PR: https://github.com/kedro-org/kedro-plugins/pull/1297. The test suite for this PR has been executed  tempoerarily against the branch referenced in that PR.

## Checklist

### DIFF
--- a/tests/io/test_core.py
+++ b/tests/io/test_core.py
@@ -621,30 +621,6 @@ class TestAbstractVersionedDataset:
         assert my_versioned_dataset._cached_load_version is None
         assert my_versioned_dataset._cached_save_version is None
 
-    def test_fetch_latest_save_version_caching(self, my_versioned_dataset, mocker):
-        # Mock generate_timestamp to return predictable values
-        mock_generate = mocker.patch(
-            "kedro.io.core.generate_timestamp",
-            side_effect=["2024-01-01T00.00.00.000Z", "2024-01-02T00.00.00.000Z"],
-        )
-
-        # First call should generate and cache
-        result1 = my_versioned_dataset._fetch_latest_save_version()
-        assert result1 == "2024-01-01T00.00.00.000Z"
-        assert my_versioned_dataset._cached_save_version == "2024-01-01T00.00.00.000Z"
-        mock_generate.assert_called_once()
-
-        # Second call should return cached value without calling generate_timestamp again
-        result2 = my_versioned_dataset._fetch_latest_save_version()
-        assert result2 == "2024-01-01T00.00.00.000Z"
-        mock_generate.assert_called_once()  # Still only called once
-
-        # Clear cache and call again - should generate new timestamp
-        my_versioned_dataset._clear_version_cache()
-        result3 = my_versioned_dataset._fetch_latest_save_version()
-        assert result3 == "2024-01-02T00.00.00.000Z"
-        assert mock_generate.call_count == 2
-
     def test_fetch_latest_load_version_success(self, my_versioned_dataset, mocker):
         mock_get_versioned_path = mocker.patch(
             "kedro.io.core.AbstractVersionedDataset._get_versioned_path",


kedro/io/core.py
```diff
@@ -870,7 +890,33 @@ def exists(self) -> bool:
 
     def _release(self) -> None:
         super()._release()
-        self._version_cache.clear()
+        # Clear version cache
+        with self._version_cache_lock:
+            self._cached_load_version = None
+            self._cached_save_version = None
```
DimedS: It seems to make sense to wrap this logic in a helper like `self._clear_version_cache()`

kedro/io/core.py
```diff
@@ -870,7 +890,33 @@ def exists(self) -> bool:
...
+
+    def __getstate__(self) -> dict[str, Any]:
```
DimedS: This part isn’t particularly elegant, but I’m not sure there’s a clean way to avoid the two methods here.

rashidakanchwala:
> Additionally, could you clarify why we need self._cached_save_version at all? In my understanding, we only need this value once - at the moment we save the dataset (and we already clear the cache before saving). At that point we determine a new save destination, generate the version, save to it, and that’s it. The next save should naturally go to a new version.

@DimedS thanks, that makes sense. I initially followed the same logic we had with cachetools, since I wanted to avoid changing behavior as part of removing the dependency.

I agree this could potentially be simplified. I’ll wait to see if @merelcht  and @elena-khaustova-qb  have any other thoughts as well, and then make that change.

kedro/io/core.py
rashidakanchwala: Agreed, I don’t like it either. I added `__getstate__`/`__setstate__` so the per-instance lock doesn’t get pickled, which is required for Python multiprocessing (ParallelRunner). I couldn’t find a cleaner way to keep the lock for threads while staying pickle-safe.

Also, we had to handle a similar pickling issue with the previous cachetools approach too and this ends up being a simpler version of that workaround.

ElenaKhaustova (commented):
Thank you, @rashidakanchwala!

The overall direction of removing cachetools and replacing it with manual per-instance caching looks good and addresses the compatibility issues introduced in cachetools 7.

One point I’d like to revisit is the introduction of `RLock`. The previous implementation using `@cachedmethod + Cache` was not thread-safe and did not provide atomicity guarantees under `ThreadRunner` (duplicate computation and last-write-wins behavior were possible and accepted). From that perspective, adding a lock changes semantics rather than preserving existing behavior.

From a `ParallelRunner` perspective, this also doesn’t introduce new concerns: even with shared-memory datasets, dataset instances are not shared across processes. Each worker receives its own unpickled instance, so version caching has always been process-local. This was true with `cachetools` and remains true with the manual cache, with or without a lock. A per-instance lock therefore doesn’t affect multiprocessing correctness either way.

Given that, if the goal of this PR is to remove `cachetools` while keeping behavior equivalent, I think we can safely omit the lock for now and keep the change focused. Introducing explicit thread-safety is a reasonable improvement, but it’s not related to the `cachetools` removal and could be addressed in a follow-up PR with a clear motivation.

This would help avoid bundling two behavior changes (dependency removal + stronger concurrency guarantees) into a single change.

rashidakanchwala:
@ElenaKhaustova You’re right ... thanks for pointing this out. I had assumed cachetools implicitly handled threading, but that’s incorrect: cachedmethod only provides thread-safety when a lock is explicitly passed, which we weren’t doing before.

This does simplify things. I’ve updated the code accordingly to keep the behavior aligned with the previous implementation... much simpler now 🙂

rashidakanchwala:
@DimedS I initially removed it based on your point, but added it back as discussed. I’m not entirely sure about all the call paths, but the intent is to avoid accidentally generating two slightly different timestamps during a save

ElenaKhaustova (approved):
Thank you, @rashidakanchwala, LGTM!

I agree that we should not remove the saved version; we should only reimplement the behaviour equivalent to the one with cache, but without it.

PR #5356, opened 2026-02-05
Add warnings when Node returns value but outputs=None

### LINKED ISSUE #5232: Pipeline/node validation is not catching discrepancy in node output.
## Description

Initially pointed out here: https://github.com/kedro-org/kedro/issues/5226
Discrepancy originally found on kedro-starters fixed here: https://github.com/kedro-org/kedro-starters/pull/299

Kedro node validation did not raise any kind of warning when a node output was set to `None` in `pipeline.py` but returned a dict on the node function itself.

Suggestions:
- Test if other similar discrepancies are caught (e.g. does an error occur when pipeline definition outputs something but node returns `None`?).
- Add validation for these cases.

rashidakanchwala:
## Description

Resolves #5232

## Development notes

This PR adds validation for the `outputs=None` case in node._outputs_to_dictionary().
If a node function returns a value while `outputs=None`, Kedro now emits a clear, actionable warning explaining that the return value will be ignored and how to fix the node definition.

I also checked for other missing node-level validations. The inverse case, where outputs are defined but the function returns None is already handled at the dataset layer. Attempting to save None to a dataset raises a DatasetError, with existing coverage in the sequential and parallel runner tests. Since this behavior is already validated downstream, no additional node-level validation was added.

## Checklist

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -33,8 +33,6 @@ jobs:
           activate-environment: true
       - name: Install dependencies
         run: |
-          # TEMP: For passing tests, will be removed before merge
-          uv pip install "git+https://github.com/kedro-org/kedro-plugins.git@fix/remove-cachetools-deps#subdirectory=kedro-datasets"
           make install-test-requirements
           uv pip install pip
           make install-pre-commit
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -30,8 +30,6 @@ jobs:
           activate-environment: true
       - name: Install dependencies
         run: |
-            # TEMP: For passing tests, will be removed before merge
-            uv pip install "git+https://github.com/kedro-org/kedro-plugins.git@fix/remove-cachetools-deps#subdirectory=kedro-datasets"
             make install-test-requirements
             make install-pre-commit
       - name: pip freeze
--- a/tests/io/test_data_catalog.py
+++ b/tests/io/test_data_catalog.py
@@ -625,29 +625,31 @@ class TestDataCatalog:
                     ds_name="bad", ds_config="not_dict"
                 )
 
-        def test_confirm(self, tmp_path, caplog, mocker):
-            """Confirm the dataset"""
-            with caplog.at_level(logging.INFO):
-                mock_confirm = mocker.patch(
-                    "kedro_datasets.partitions.IncrementalDataset.confirm"
-                )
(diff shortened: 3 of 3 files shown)


kedro/pipeline/node.py
```diff
@@ -663,6 +663,15 @@ def _from_list() -> dict:
...
 
         if self._outputs is None:
+            if outputs is not None:
+                node_name = self._name or self._func_name
+                warnings.warn(
+                    f"Node '{node_name}' returned a value of type "
+                    f"'{type(outputs).__name__}', but outputs=None. "
+                    f"This value will not be included in the pipeline outputs. "
+                    f"To keep it, specify outputs in the node definition.",
+                    UserWarning,
```
ravi-kumar-pilla, line 674: > but outputs=None.

This part of the message is a bit hard to follow. [Nit] suggestion -

```py
(
    f"Node '{node_name}' returned a value of type "
    f"'{type(outputs).__name__}', but the node is defined with outputs=None. "
    "As a result, this value will be discarded and not included in the "
    "pipeline outputs. To persist the result, specify one or more outputs "
    "in the node definition."
)

```

or

```py
(
    f"Node '{node_name}' returned a value of type "
    f"'{type(outputs).__name__}', but the node is defined with outputs=None. "
    "Kedro ignores return values for nodes without declared outputs. "
    "If this return value is intentional, declare outputs in the node "
    "definition to include it in the pipeline."
)

```
Thank you

lrcouto (approved):
Tested it locally and it works. I also like Ravi's second suggestion for the warning message.

kedro/pipeline/node.py
ravi-kumar-pilla: Thank you for updating the message @rashidakanchwala . One design question would be, is this warning required for every node that has this issue or can we stick to showing the warning only once like we did for [previews](https://github.com/kedro-org/kedro/pull/5328/changes#diff-26deb45c9dfb4d41c486319bdfd4e858f42aa728e2efbcbeb68496a0445f11adR189) and other cases earlier ?

rashidakanchwala: Warning once makes sense for preview_fn since it’s a global feature-level notice.
This case is node-specific: if multiple nodes are incorrect, warning only once could mean the user fixes the first node but misses the second or third. So I think we should keep emitting the warning for every affected node. If it turns out to be too noisy and users raise concerns, we can always revisit and refine the behaviour later.

Issue #5315, opened 2026-01-07
Documentation for Parameter Validation framework


ravi-kumar-pilla:
## Description

Publish user-facing documentation + examples for parameter validation: how to opt in via type hints, supported typing targets (Pydantic/dataclasses), runtime params behavior etc.

### Documentation page (if applicable)

https://docs.kedro.org/en/stable/configure/parameters/

## Context

Kedro [docs](https://docs.kedro.org/en/stable/configure/parameters/) already cover where parameters live (conf/.../parameters*.yml) and how runtime parameters are supplied and merged. The new [feature](https://github.com/kedro-org/kedro/discussions/5234) changes the user experience by enabling typed objects and fail-fast validation when accessing params. This needs to be documented.

SajidAlamQB:
Completed in: https://github.com/kedro-org/kedro/pull/5467

Issue #5314, opened 2026-01-07
Tests for parameter validation framework


ravi-kumar-pilla:
## Description

- Add comprehensive unit tests validating correctness, compatibility, and error handling for the new parameter validation framework.
- Add/Update benchmark tests 

## Context

The [KEP](https://github.com/kedro-org/kedro/discussions/5234) design introduces multiple components (ParameterValidator, TypeExtractor, ModelFactory, SourceFilters, exceptions) and a new behavior at `context.params` access time, including caching which needs testing. 

## Possible Implementation

- Add unit tests for each component following our usual unit test patterns
- Update existing benchmark tests to include benchmarking parameter validation. Look at the performance metrics gathered during [spike](https://github.com/kedro-org/kedro/issues/5161) work and adapt to asv benchmarking.

## Possible Alternatives

NA

SajidAlamQB:
Unit tests are complete they were included alongside each implementation (PRs #5403, #5419, #5421, #5422, #5423). All at 100% coverage.

Still outstanding: benchmark tests using ASV to measure parameter validation performance impact on context.params access time. This should cover the overhead of TypeExtractor pipeline walking and `instantiate_model` for various numbers of typed parameters.

ravi-kumar-pilla:
@SajidAlamQB any plans of doing this in coming sprints ?

Issue #5313, opened 2026-01-07
Implement parameter validation framework


ravi-kumar-pilla:
## Description

Implement native parameter validation + typed instantiation in Kedro core, following the consolidated validation framework design. The core deliverable is that accessing context.params performs fail-fast validation and (when applicable) returns typed objects (e.g., Pydantic models / dataclasses) rather than raw dicts.

## Context

The [KEP](https://github.com/kedro-org/kedro/discussions/5234) proposes a validation framework integrated at the context level (`KedroContext.params` as the entry point), orchestrated by `ParameterValidator`, which:

1. Merges config parameters with runtime params
2. Discovers required types from pipeline node signatures via TypeExtractor + source filters
3. Instantiates typed objects via ModelFactory
4. Caches results for performance. 

## Possible Implementation

**1. Context integration**

  - Implement `_get_validated_params()` (or equivalent) and wire it to `KedroContext.params` as described in the [KEP](https://github.com/kedro-org/kedro/discussions/5234)
  - Add a cache keyed by environment + runtime params hash (or a safe equivalent) so repeated `context.params` calls are cheap.

**2. Type discovery**

  - Implement `TypeExtractor` to walk the active pipeline(s) and inspect node function signatures + annotations.
  - Use a `SourceFilter` abstraction; in this ticket, focus on the ParameterFilter path (datasets can be added in future).

**3. Model instantiation**

  - Implement ModelFactory to support: 
    - Pydantic models when installed (optional dependency).
    - Python dataclasses.
    - Fallback behavior: if no type requirement exists, return raw values untouched.
  - Provide clear errors via custom exceptions (e.g., a specialized ValidationError that includes param path + expected type + source node info). 

Draft implementation - https://github.com/kedro-org/kedro/tree/test/validate_no_hook/kedro/framework/validation

## Possible Alternatives

NA

SajidAlamQB:
## Parameter Validation Framework: Implementation Plan

Following discussion with @merelcht and @ravi-kumar-pilla, the original draft PR (#5371) is being broken down into smaller, incremental PRs each containing a single component with its tests.

### Architecture

```mermaid
graph TD
    A[kedro run] --> B[KedroSession]
    B --> C[KedroContext]
    C --> D[_get_validated_params]
    D --> E[ParameterValidator]
    D --> F["config_loader['parameters']"]
    F --> G["OmegaConf.merge (runtime params)"]
    E --> H[TypeExtractor]
[...]

Issue #5161, opened 2025-10-08
Spike: Explore implementation options for native Pydantic parameter validation


DimedS:
🔗 Related proposal: [Add native support for Pydantic parameters to Kedro (#5110)](https://github.com/kedro-org/kedro/issues/5110)

---

## 🎯 Objective
Investigate the technical feasibility and implementation options for introducing **optional native Pydantic-based parameter validation** in Kedro.  
The goal is to enable early (fail-fast) validation of parameters and provide type-safe integration with future features such as JSON Schema export, UI auto-generation in Kedro-Viz, and MCP/LLM interfaces.

---

## 🧱 Scope of this spike
1. **Explore implementation approaches:**
   - Core integration (within `KedroSession` or `KedroContext`).
   - Plugin or hook-based validation (e.g., `after_context_created`).
   - Hybrid approach leveraging `typing` introspection on node inputs.

2. **Prototype behaviour:**
   - Load YAML parameters.
   - Detect node functions with Pydantic `BaseModel` annotations.
   - Validate and instantiate Pydantic objects before pipeline execution.
   - Handle validation errors gracefully with clear error messages.

3. **Performance benchmarking:**
   - Measure session creation time impact for small and large parameter sets.
   - Compare baseline vs validation-enabled startup.

4. **Dependency evaluation:**
   - Assess whether Pydantic should be a core dependency or an optional extra (`kedro[pydantic]`).

---

## ⚙️ Deliverables
- Short design note comparing 2–3 viable implementation strategies.
- Basic proof of concept demonstrating validation at session creation.
- Benchmark results and initial performance estimates.
- Recommendation on:
  - Integration location (core vs plugin).
  - Default vs optional dependency.
  - Potential follow-up KEP for formal adoption.

---

## 🕒 Estimated effort
~1 week (engineer spike)

---

## 🧭 Next steps after spike
- Prepare a **Kedro Enhancement Proposal (KEP)** summarising findings.  
- Present results at TSC for prioritisation and roadmap inclusion.

datajoely:
Great write up @DimedS

ravi-kumar-pilla:
Thank you for the write-up and the info in the initial issue. I went through the discussions and the Tech design. Started off with a plugin approach which seems a bit straightforward. Tried implementing a draft [here](https://github.com/kedro-org/kedro-plugins/tree/test/validator/kedro-validator) which works well. Please give it a try @datajoely 

**Next Steps:**
[...]

PR #5371, opened 2026-02-11
[Reference] Implement parameter validation framework

SajidAlamQB:
# Description

Implements the parameter validation framework as proposed in the [KEP](https://github.com/kedro-org/kedro/discussions/5234) and tracked in #5313.

Accessing `context.params` now performs fail-fast validation and returns typed objects (Pydantic models / dataclasses) when node function signatures include type hints for parameter inputs. Projects without typed parameter hints are completely unaffected fully backward compatible.

Related: #5110 (parent), #5161 (spike), #5314 (tests), #5315 (docs)

### Fixes applied during integration

Building on Ravi's draft implementation, the following issues were found and fixed while integrating into core and running stress tests:

1. **Nested parameter paths for typed models** — `_get_parameters` builds catalog entries like `params:model_options.test_size` by recursing into dicts. When `model_options` is a Pydantic model or dataclass instead of a dict, the recursion broke because `isinstance(value, dict)` returns `False` for these types. Added handling to call `model_dump()` for Pydantic and `asdict()` for dataclasses so nested paths resolve correctly.

2. **Pipeline loading crashes in test environments** — `extract_types_from_pipelines` called `dict(pipelines)` outside the try/except block. This triggers lazy loading of the project's pipeline registry. In test fixtures using fake package names, this throws `ModuleNotFoundError`. Moved `dict(pipelines)` inside the try/except so it returns empty requirements gracefully.

3. **Missing parameters treated as hard errors** — If the type extractor found a type hint in any pipeline but the parameter didn't exist in YAML, it raised a `ValidationError`. Changed to skip with a debug log — the user might not be running that pipeline, or the parameter might be intentionally absent.

4. **Shallow merge breaking nested CLI overrides** — `validate_raw_params` took `config_params` and `runtime_params` separately and merged with `{**config_params, **runtime_params}`. Running `--params="model_options.test_size=0.3"` wiped out sibling fields (`random_state`, `features`) because the shallow merge replaced the entire `model_options` dict. Traced through the config loader — `config_loader["parameters"]` already deep-merges runtime params via `OmegaConf.merge` in `load_and_merge_dir_config`. Simplified `validate_raw_params` to accept a single `raw_params` dict directly from the config loader. Added fallback to `self._runtime_params` when config is missing entirely.

5. **Escaped newlines in error messages** — Validation errors used literal `\\n` instead of actual newlines, producing unreadable single-line error output.

6. **KedroContext field initialisation cleanup** — Changed `_parameter_validator` from lazy `None` init to `field(init=False, factory=ParameterValidator)` — the attrs-idiomatic approach. `_validated_params_cache` intentionally remains `None` as a sentinel to distinguish "not yet computed" from "computed with empty result".

### How to test

Stress test project with expanded scenarios: https://github.com/SajidAlamQB/kedro-parameter-validation-test

Clone the repo and run the test matrix from the [integration testing comment](https://github.com/kedro-org/kedro/pull/5371#issuecomment-3915450657) below.

### Known limitations / follow-ups

- **Type extractor walks all pipelines** — `extract_types_from_pipelines` inspects every registered pipeline regardless of which one is being run. Nothing breaks, but it's unnecessary work. Scoping extraction to the target pipeline could be a follow-up optimisation.
- **Shared model instances** — Multiple nodes referencing the same typed parameter get the same object instance. Pydantic models are immutable so this is safe. Dataclasses are mutable, so a node could theoretically mutate a shared instance. Worth considering for a follow-up.
- **Conflicting type requirements across pipelines** — If two pipelines define different type hints for the same parameter key, last one wins silently. Could warn or error in a future iteration.

## Development notes

Based on the draft implementation from the `test/validate_no_hook` branch:

- `KedroContext.params` calls `_get_validated_params()` with caching (simple cache, sufficient because `KedroSession` enforces one run per session)
- `TypeExtractor` inspects pipeline node signatures using pluggable `SourceFilter` abstraction (focused on `ParameterSourceFilter` for this ticket)
- `ModelFactory` supports Pydantic models (optional), Python dataclasses, and falls back to raw values for builtin types
- `ValidationError` and `ModelInstantiationError` with detailed error messages
- Validation is opt-in by design per TSC tech design consensus invisible to projects that don't define typed models

## Checklist


SajidAlamQB:
# Integration Testing & Fixes

### Stress test project

I've put together an expanded spaceflights project for testing the validation framework across different scenarios: 

https://github.com/SajidAlamQB/kedro-parameter-validation-test

Clone it and run the test matrix below against the validation branch. Let me know if there are any other scenarios I should be covering.

### Fixes applied during integration

**1. Nested parameter paths for typed models**

`_get_parameters` builds catalog entries like `params:model_options.test_size` by recursing into dicts. When `model_options` is a Pydantic model or dataclass instead of a dict, the recursion broke because `isinstance(value, dict)` returns `False` for these types. Added handling to call `model_dump()` for Pydantic and `asdict()` for dataclasses so nested paths resolve correctly.

**2. Pipeline loading crashes in test environments**

`extract_types_from_pipelines` called `dict(pipelines)` outside the try/except block. This triggers lazy loading of the project's pipeline registry. In test fixtures using fake package names, this throws `ModuleNotFoundError`. Moved `dict(pipelines)` inside the try/except so it returns empty requirements gracefully. Resolved CI failures.

**3. Missing parameters treated as hard errors**

If the type extractor found a type hint in any pipeline but the parameter didn't exist in YAML, it raised a `ValidationError`. This is too aggressive the user might not be running that pipeline, or the parameter might be intentionally absent. Changed to skip with a debug log.

**4. Shallow merge breaking nested CLI overrides**

`validate_raw_params` took `config_params` and `runtime_params` separately and merged with `{**config_params, **runtime_params}`. Running `--params="model_options.test_size=0.3"` produced `runtime_params = {"model_options": {"test_size": 0.3}}`, and the shallow merge replaced the entire `model_options` dict, wiping out `random_state` and `features`.

Traced through the config loader `config_loader["parameters"]` already deep-merges runtime params via `OmegaConf.merge` in `load_and_merge_dir_config`. The second merge was redundant and destructive. Simplified `validate_raw_params` to accept a single `raw_params` dict directly from the config loader.

**5. Escaped newlines in error messages**

Validation errors used `\\n` instead of actual newlines, producing unreadable single-line error output.

**6. KedroContext field initialisation cleanup**

Changed `_parameter_validator` from lazy `None` init to `field(init=False, factory=ParameterValidator)` `_validated_params_cache` intentionally remains `None` to distinguish "not yet computed" from "computed with empty result".

### Test setup

The stress test project includes:

**Models:** `ModelOptions` (Pydantic, flat), `ReportingConfig` (Pydantic, nested sub-model with `Field(ge=1, le=10)` constraint), `EvaluationConfig` (dataclass)

**Parameters:** 20+ untyped top-level params in `parameters.yml` to verify they're ignored by validation

**Pipelines:** `data_processing` (completely untyped nodes), `data_science` (mixed typed and untyped nodes)

### Test matrix

| # | Scenario | Command | Result |
|---|----------|---------|--------|
| 1 | Full pipeline run | `kedro run` | ✅ Pass |
| 2 | Untyped-only pipeline | `kedro run --pipeline data_processing` | ✅ Pass |
| 3 | Typed pipeline only | `kedro run --pipeline data_science` | ✅ Pass |
| 4 | CLI nested override on typed field | `kedro run --params="model_options.test_size=0.3"` | ✅ Pass — test_size=0.3, sibling fields preserved |
| 5 | Invalid type on Pydantic field | `kedro run --params="model_options.test_size=banana"` | ✅ Fail-fast before node execution |
| 6 | Pydantic Field constraint violation | `kedro run --params="reporting_config.precision=99"` | ✅ Fail-fast — "should be less than or equal to 10" |
| 7 | CLI override on dataclass field | `kedro run --params="evaluation_config.threshold=0.9"` | ✅ Pass — threshold=0.9 |
| 8 | Override untyped param | `kedro run --params="batch_size=128"` | ✅ Pass — validation ignores it |
| 9 | Deep nested override | `kedro run --params="reporting_config.notification.email=new@test.com"` | ✅ Pass |

All fail-fast scenarios errored during catalog creation before any node execution, with clear Pydantic error messages.

kedro/framework/context/context.py
```diff
@@ -205,21 +211,41 @@ def catalog(self) -> CatalogProtocol:
...
+        """
+        # Return cached result if available
+        if self._validated_params_cache is not None:
+            return self._validated_params_cache
+
+        # Get raw parameters note: config loader includes runtime params
+        try:
+            raw_params = self.config_loader["parameters"]
+        except MissingConfigException as exc:
+            warn(f"Parameters not found in your Kedro project config.\n{exc!s}")
```
ravi-kumar-pilla, line 226: I did not quite understand this exception. Is this if config is not found or parameters are not found ? May be this is not even a new change introduced in this PR but would like to understand

kedro/framework/context/context.py
```diff
@@ -205,21 +211,41 @@ def catalog(self) -> CatalogProtocol:
...
 
+    def _get_validated_params(self) -> dict[str, Any]:
+        """Get validated parameters with caching support.
+
+        Returns:
+            Validated and transformed parameters with model instantiation.
+        """
+        # Return cached result if available
+        if self._validated_params_cache is not None:
+            return self._validated_params_cache
```
ravi-kumar-pilla, line 221: Do we have a test or may be did you see if this cache is really used. I am sure for multiple session runs it wont be used but within a session run, do we know if its worth having this ?

kedro/framework/context/context.py
```diff
@@ -295,8 +321,20 @@ def _add_param_to_params_dict(param_name: str, param_value: Any) -> None:
...
             key = f"params:{param_name}"
             params_dict[key] = param_value
+
+            # Convert typed objects to dicts for nested path registration
+            nested_dict = None
             if isinstance(param_value, dict):
-                for key, val in param_value.items():
+                nested_dict = param_value
+            elif self._is_pydantic_model(param_value):
+                nested_dict = param_value.model_dump()
```
ravi-kumar-pilla: If I understand this correctly, we are trying to register nested paths of a parameter value by first converting them into dicts and then registering. But this would not preserve the pydantic type. 

Lets say if we have a nested model - `params:a.b` in which `a` is a pydantic model and `b` is also a pydantic model. Now we expect someone using these models in a kedro Node to access it like `a.b` which should give not a dict but an instance of `b` pydantic class. 

Does this work if we do a `model_dump` here and change it to dicts ?

kedro/framework/validation/validators/parameter_validator.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+class ParameterValidator:
+    """Parameter validator that combines type extraction and model instantiation."""
+
+    def __init__(self) -> None:
+        """Initialize parameter validator."""
+        self.type_extractor = TypeExtractor(ParameterSourceFilter())
+        self.model_factory = ModelFactory()
+        self.logger = logging.getLogger(__name__)
```
ravi-kumar-pilla: In kedro, I think we register the logger at the top and not as a class instance. I dont see anything wrong with this but if we want to stick to the pattern, may be we can do the same

kedro/validation/parameter_validator.py
```diff
@@ -0,0 +1,125 @@
...
+            except Exception as exc:
+                error_msg = f"Parameter '{param_key}': {exc}"
+                validation_errors.append(error_msg)
+                self.logger.error(error_msg)
+
+        # Log success summary
+        if instantiated_count > 0:
+            self.logger.info(
+                f"Successfully instantiated {instantiated_count} parameter models"
+            )
```
ravi-kumar-pilla, line 83: I used these logs during testing in the draft PR. I am not sure if it really helps in logging this at info level. May be we can get this down to debug level logging or remove it based on what the team feels like

kedro/framework/validation/model_factory.py
```diff
@@ -0,0 +1,104 @@
...
+
+from .exceptions import ModelInstantiationError
+
+
+class ModelFactory:
+    """Handles instantiation of various model types like Pydantic models and dataclasses."""
+
+    def __init__(self) -> None:
+        """Initialize the model factory."""
+        self.logger = logging.getLogger(__name__)
```
ravi-kumar-pilla: Same as above if we want logger to be not an instance variable

kedro/framework/validation/model_factory.py
```diff
@@ -0,0 +1,104 @@
...
+            Instantiated dataclass
+        """
+        if isinstance(raw_value, dict):
+            return model_type(**raw_value)
+        else:
+            # If raw_value is not a dict, try to instantiate with empty dict
+            # This will likely fail but provides a clear error message
+            return model_type()
+
+    def _is_pydantic_model(self, model_type: type) -> bool:
```
ravi-kumar-pilla: I see we used this in 2 places now. May be we can move this to utils or some shared location. Unless we want to change the behavior mentioned here - https://github.com/kedro-org/kedro/pull/5371/changes#r2825189074

kedro/framework/validation/README.md
```diff
@@ -0,0 +1,558 @@
+# Kedro Validation Framework
```
ravi-kumar-pilla: should we remove this and handle any docs in a separate PR which addresses - https://github.com/kedro-org/kedro/issues/5315

kedro/framework/validation/source_filters.py
```diff
@@ -0,0 +1,77 @@
...
+    def extract_key(self, source_name: str) -> str:
+        """Extract parameter key from params:key format."""
+        return source_name.split(":", 1)[1]
+
+    def get_log_message(self, key: str, type_name: str) -> str:
+        """Generate parameter-specific log message."""
+        return f"Found parameter requirement: {key} -> {type_name}"
+
+
+class DatasetSourceFilter(SourceFilter):
```
ravi-kumar-pilla: Lets remove this class for now. Since we are only validating Parameters, either we can completely remove the class SourceFilter or keep it for easy extendable design. I personally like this way of doing. But if the team feels it being over-engineered, we can think about extending this when we have DatasetValidation.

kedro/framework/validation/type_extractor.py
```diff
@@ -0,0 +1,200 @@
...
+class TypeExtractor:
+    """Extracts type requirements from various sources like pipeline nodes."""
+
+    def __init__(self, source_filter: SourceFilter) -> None:
+        """Initialize the type extractor.
+
+        Args:
+            source_filters: List of source filters to use for extraction.
+        """
+        self.logger = logging.getLogger(__name__)
```
ravi-kumar-pilla: same comment about logger

kedro/validation/type_extractor.py
```diff
@@ -0,0 +1,200 @@
...
+    ) -> dict[str, str]:
+        """Build mapping from dataset names to argument names.
+
+        Args:
+            node: Node object
+            signature: Function signature
+
+        Returns:
+            Dictionary mapping dataset names to argument names
+        """
```
ravi-kumar-pilla, line 141: I could follow these functions as I have some idea on these. But it would be worth adding examples to each function which would help enrich our docs and also help in PR reviews. Especially the type_extractor where we have a bunch of helper functions.

I would also recommend adding examples in doc strings across the PR. May be 1-2 should be great

kedro/validation/type_extractor.py
```diff
@@ -0,0 +1,200 @@
...
+        dataset_to_arg: dict[str, str] = {}
+        node_inputs = getattr(node, "inputs", None)
+
+        if isinstance(node_inputs, dict):
```
ravi-kumar-pilla, line 145: how does this handle nested dicts ?

kedro/framework/validation/type_extractor.py
```diff
@@ -0,0 +1,200 @@
...
+        """
+        type_requirements = {}
+
+        for node in getattr(pipeline, "nodes", []):
+            node_type_requirements = self.extract_types_from_node(node)
+            type_requirements.update(node_type_requirements)
+
+        return type_requirements
+
+    def extract_types_from_node(self, node: Node) -> dict[str, type]:
```
ravi-kumar-pilla: do we want this to be public or should we make it private ? For now I do not think of a usecase where someone wants to directly use this. Can we make it private ?

ravi-kumar-pilla:
Hi @SajidAlamQB ,

Great work and I like the description provided in the PR. I left some comments, please have a look.

**Regarding known limitations:**

> 1. Type extractor walks all pipelines — `extract_types_from_pipelines` inspects every registered pipeline regardless of which one is being run. Nothing breaks, but it's unnecessary work. Scoping extraction to the target pipeline could be a follow-up optimisation. 

This is good and we can have it in a separate PR

> 2. Shared model instances — Multiple nodes referencing the same typed parameter get the same object instance. Pydantic models are immutable so this is safe. Dataclasses are mutable, so a node could theoretically mutate a shared instance. Worth considering for a follow-up. 

I think we internally do a deepcopy for [MemoryDataset](https://github.com/kedro-org/kedro/blob/main/kedro/io/memory_dataset.py#L64). When I tested it there was no issue, might be I used a data type that does a deepcopy. But this is an issue and I feel we should test this and apply a fix in this PR.

> 3. Conflicting type requirements across pipelines — If two pipelines define different type hints for the same parameter key, last one wins silently. Could warn or error in a future iteration.

I would like to hear from the team what the behavior should be. I would suggest we warn it and document about this behavior.

Once the above comments and the suggestions on limitations are tackled, I think this PR is good to be opened for review.

Thank you

kedro/framework/validation/model_factory.py
SajidAlamQB: Moved to utils

kedro/framework/context/context.py
SajidAlamQB: This is existing Kedro behaviour when there's no `parameters.yml` in any config directory, `config_loader["parameters"]` raises `MissingConfigException`. The handler warns the user and falls back to runtime params from --params CLI flag so the pipeline can still run without a config file.

kedro/framework/context/context.py
SajidAlamQB: Yes the cache is used within a single run. `context.catalog` is a property that calls `_get_catalog()` → `_get_parameters()` → `self.params` → `_get_validated_params()`. Since catalog has no caching of its own, every access rebuilds the catalog and re-triggers param loading the validation walks all pipeline nodes and instantiates models so caching prevents that on repeated access. 

There's an existing test `test_params` that checks this path the cache kicks in on the second access through `_get_parameters.`

kedro/framework/context/context.py
SajidAlamQB: Good catch. Fixed instead of `model_dump()` which flattens everything to dicts we now iterate over `model_fields` with `getattr()` to preserve nested Pydantic types.

kedro/validation/type_extractor.py
SajidAlamQB: Node inputs are always flat either a list/tuple of dataset names or a dict mapping dataset names to argument names. Nesting comes from dot notation in parameter names like `params:model_options.test_size`, which is handled by `resolve_nested_path`, not this function.

SajidAlamQB:
Thanks @ravi-kumar-pilla for the detailed review, I've addressed all the comments. Here's a summary:

**Shared model instances:**

I checked and `MemoryDataset` already handles this safely it defaults to `deepcopy` for any type that isn't pandas/numpy/ibis, so Pydantic models and dataclasses both get deepcopied on both `save()` and `load()`. Each node receives its own independent instance no shared mutation risk. I'll add a test to verify this explicitly.

**Type extractor walks all pipelines:** Agreed lets do a follow-up PR.

**Conflicting types:** Added a warning when conflicting types are detected across pipelines. Last pipeline's type wins. Happy to discuss stricter behaviour with the team.

**Other changes in this push:**
- Moved loggers to module-level across all files
- Moved `_is_pydantic_model` to shared `validation/utils.py`
- Removed `DatasetSourceFilter` (Will work investigating this in following sprint)
- Removed `README.md` (We will handle docs in another PR)
- Renamed `extract_types_from_node` to private
- Fixed `model_dump()` → shallow field access to preserve nested Pydantic sub-model types
- Downgraded log to debug level
- Added docstring examples across `type_extractor.py`, `model_factory.py`, and `source_filters.py`

kedro/framework/validation/utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+    class MyModel(BaseModel):
+        x: int = 1
+
+
+    is_pydantic_model(MyModel(x=1))  # True
+    is_pydantic_model({"x": 1})  # False
+    ```
+    """
```
merelcht: This feels like an unnecessary amount of docstring for such a simple function. Same for the one below.

kedro/framework/validation/exceptions.py
```diff
@@ -0,0 +1,47 @@
+"""Custom exceptions for validation framework."""
+
+from __future__ import annotations
+
+
+class ValidationError(Exception):
```
merelcht: Is it necessary to have such extensive exception classes? We usually keep them very simple like: 

```python
class MissingConfigException(Exception):
    """Raised when no configuration files can be found within a config path"""

    pass
```

For config loader and the ones for catalog are similar too.

kedro/framework/validation/model_factory.py
```diff
@@ -0,0 +1,110 @@
...
+
+    class ModelOptions(BaseModel):
+        test_size: float
+        random_state: int
+
+
+    factory = ModelFactory()
+    raw = {"test_size": 0.2, "random_state": 3}
+    instance = factory.instantiate("model_options", raw, ModelOptions)
+    # instance is ModelOptions(test_size=0.2, random_state=3)
```
merelcht: Is the expectation that users import these classes directly in their project? If not, then I don't think it's necessary to have docstrings with examples everywhere.

kedro/framework/validation/model_factory.py
```diff
@@ -0,0 +1,110 @@
...
+        ```
+        """
+        try:
+            if is_pydantic_class(model_type):
+                return self._instantiate_pydantic(raw_value, model_type)
+            elif dataclasses.is_dataclass(model_type):
+                return self._instantiate_dataclass(raw_value, model_type)
+            else:
+                logger.debug(
+                    "Source '%s' type %s is not a supported model type, using raw value",
```
merelcht: What is meant with `raw value`? Can someone understand this logging message as is in the stacktrace?

merelcht (commented):
Thanks @SajidAlamQB for the description and testing instructions. As mentioned on Slack I think it would be great if we can break this PR up a bit and actually add tests alongside implementation to aid with review. 

I'm leaving some high level comments already, but will need more time to understand the changes. 

One observation is that I think the full `validation` directory with files should go at a higher level: `kedro/validation` instead of inside framework. This logic has a clear role which I think warrants it being it's own category. It also makes the code base easier to understand and see at a glance what features exist.

PR #5381, opened 2026-02-17
Docs: Restructure navigation to Diátaxis model (IA review)

rashidakanchwala:
This is a **draft PR for review only**. This PR is not intended to be merged in its current form.

It restructures the Kedro documentation navigation to align with Diátaxis principles: https://diataxis.fr/

- Tutorial
- How-To
- Explanation
- Reference

The goal is to validate whether this information architecture makes sense before investing in deeper content changes.

🔗 Preview the restructured docs here:  
https://kedro--5381.org.readthedocs.build/en/5381/

---

## What Changed

- Reorganised navigation into Diátaxis-aligned sections
- Re-categorised existing pages accordingly
- No intentional content rewrites

---

## Review Focus (Structure First)

Please review primarily from an information architecture perspective:

- Does the overall structure make sense?
- Are sections grouped appropriately?
- Is categorisation clear?
- Are there major structural concerns?
- Should we adopt Diátaxis as the long-term documentation model?

Content-level feedback is not the main focus,  
but if you notice anything clearly odd or misplaced, feel free to flag it.

### DIFF
--- a/docs/stylesheets/globals.css
+++ b/docs/stylesheets/globals.css
@@ -134,12 +134,13 @@
     .md-nav__item--section > .md-nav__link[for] {
         color: var(--text-primary-color);
         align-items: center;
-        margin-left: 20px;
     }
 
     /* Apply border-radius to scrollbar container */
     .md-sidebar__scrollwrap {
       border-radius: 12px;
+      overflow-y: auto;
+      max-height: calc(100vh - 7.2rem);
     }
 
     .md-sidebar__inner {
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -300,8 +300,8 @@ nav:
       - FAQs: getting-started/faq.md
       - Tutorials:
           - Kedro Spaceflights tutorial:
-              - Spaceflight tutorial: tutorials/spaceflights_tutorial.md
               - Spaceflight tutorial template: tutorials/tutorial_template.md
+              - Spaceflight tutorial: tutorials/spaceflights_tutorial.md
               - Set up data: tutorials/set_up_data.md
               - Create a pipeline: tutorials/create_a_pipeline.md
               - Add another pipeline: tutorials/add_another_pipeline.md
@@ -316,6 +316,7 @@ nav:
           - Create a minimal Kedro project: create/minimal_kedro_project.md
           - Customise a new project: create/customise_project.md
           - Kedro tools: create/new_project_tools.md
+          - Migration guide: about/migration.md
 
       - Pipelines:
           - Run a pipeline: build/run_a_pipeline.md
@@ -362,6 +363,7 @@ nav:
           - Concepts: getting-started/kedro_concepts.md
           - Kedro architecture: getting-started/architecture_overview.md
           - Kedro starters: create/starters.md
+          - Experimental APIs: about/experimental.md
 
       - Pipelines:
           - Nodes: build/nodes.md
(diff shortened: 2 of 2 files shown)


merelcht:
The structure is great and I like the sections projects, pipelines, data, configuration, deployment and extension. One observation is that the `Integration` section is actually mostly how-to content as well.. I think it makes sense to have it as a separate top level section, but just wondering if this makes sense to people.

- Explanations: configuration basics -> the second half of this docs is actually how-tos so we should probably split the page and put the how-tos in the right section
- Explanations: deployment -> I'd suggest renaming the "how to group your nodes effectively" to just "Node grouping" since this is mostly explanation
- Reference: makes sense to me that CLI and API docs are here, but I'm not sure about the migration guide, telemetry explanation and experimental features.. Maybe these should go to About?
- About: I saw Ali's comment on Slack about this page and the Contributing one. It does indeed feel like we should add some content here. 
- Kedro-Viz & Kedro-Datasets: feels like these sections don't actually follow the Diataxis model. Is the intention to eventually incorporate these in the other top-level sections or keep them separate like this?

AliceCima10:
I’m leaving a few additional thoughts here:
- Under the API section on the Reference page, I see three additional subsections at the end: Migration Guide, Kedro Telemetry, and Experimental APIs. I think Experimental APIs makes sense to keep under Reference. However, I’m wondering whether the Migration Guide would be better placed under the How-to Guides section, and whether Kedro Telemetry might added to the About section or integration? 
- Regarding Merel's point on Kedro-Viz and Datasets — I think it makes sense to keep them as separate tabs. One thing could be to apply Diátaxis principles within the Kedro-Viz docs themselves, wdyt? Right now they mix how-to guides, explanations, etc...

ElenaKhaustova:
Thank you, @rashidakanchwala. I like the structure and the proposal to improve it!

Here are a few comments / suggestions / questions.

1. A lot of our pages (for example, https://kedro--5381.org.readthedocs.build/en/5381/catalog-data/kedro_dataset_factories/#how-to-generalise-datasets-of-the-same-type, which is now under `Explanations`) currently mix how-to with explanations, so we plan further restructuring of them later on, right?
2. Then I also noticed that some pages currently appear under several sections, for example, https://kedro--5381.org.readthedocs.build/en/5381/build/llm_context_node/ is under `Explanations` and `How to`. It's fine if we aim to decide later where it belongs and how to split it.
3. Diátaxis also assumes a `tutorials` section, which totally makes sense for us as well. Do we plan to embed it as a separate tab? Currently, we keep tutorials under `Get Started`.
4. Do we plan to keep the `Get Started` tab at all and use it as the main page?
5. Shall we follow the same structure under `How to` and `Explanations`? Currently `How to` tab has `Development section`, but `Explanations` does not. I think it will be easier to navigate if they follow the same underlying structure.
6. It feels that `Kedro Viz` and `Kedro Datasets` should be displayed separately from the docs structure. Meaning that they are bigger pieces and not part of the kedro core. Currently, it makes an impression that this is another tab for kedro docs.
Screenshot 2026-02-25 at 17 52 39

---

I agree on this - keeping them separately makes sense to me as well as updating Kedro-Viz docs structure later on. Dataset's docs should be fine as is, as they're rendered from the docstrings.

> Regarding Merel's point on Kedro-Viz and Datasets — I think it makes sense to keep them as separate tabs. One thing could be to apply Diátaxis principles within the Kedro-Viz docs themselves, wdyt? Right now, they mix how-to guides, explanations

rashidakanchwala:
Closing this ticket for now; I will refer to the comments in this ticket for the next work.

Issue #5354, opened 2026-02-04
KedroSession and Pipeline Loading: Design & Performance Considerations


lrcouto:
The goal of this discussion is to examine how `KedroSession` and `_ProjectPipelines` currently handle loading, storing, and retrieving pipelines, how this design impacts performance, and what changes might be worth considering in future updates to `KedroSession`.

### Current design

Today, this is the flow that occurs when a pipeline run is triggered, either via the CLI or programmatically.

```mermaid
graph TD
    %% Nodes
    Start(["KedroSession.run()"])
    
    subgraph Initialization ["Pipeline Resolution"]
        GetItem["_ProjectPipelines.\__getitem\__"]
        LoadData["_load_data()"]
        Register["register_pipelines()"]
        Find["find_pipelines()"]
        SetFlag["Set is_data_loaded = True"]
    end

    subgraph Execution ["Run Pipelines"]
        Run["Pipeline Execution"]
        End(["Session Ends"])
    end


    Start -->|"Calls pipeline[name]"| GetItem
    GetItem --> LoadData
    LoadData --> Register
    Register --> Find
    Find --> Register
    Register --> LoadData
    LoadData --> SetFlag
    
 
    SetFlag -->|"Loaded"| Run
    Run --> End
```

In more concrete terms:
- The session is created and [`KedroSession.run()`](https://github.com/kedro-org/kedro/blob/37ec20e7db2a214d03b37ca1e83656eaabe4718e/kedro/framework/session/session.py#L282) is invoked.
- The session accesses the `pipeline` object, which is an instance of [`_ProjectPipelines`](https://github.com/kedro-org/kedro/blob/37ec20e7db2a214d03b37ca1e83656eaabe4718e/kedro/framework/project/__init__.py#L171).
- [`_load_data()`](https://github.com/kedro-org/kedro/blob/37ec20e7db2a214d03b37ca1e83656eaabe4718e/kedro/framework/project/__init__.py#L199) calls the [`register_pipelines()`](https://github.com/kedro-org/kedro/blob/37ec20e7db2a214d03b37ca1e83656eaabe4718e/kedro/templates/project/%7B%7B%20cookiecutter.repo_name%20%7D%7D/src/%7B%7B%20cookiecutter.python_package%20%7D%7D/pipeline_registry.py#L8) function, located in the Kedro project.
[...]

PR #5401, opened 2026-02-24
Pass list of pipelines to run to find_pipelines

lrcouto:
## Description
Full context: https://github.com/kedro-org/kedro/issues/5354

This PR comes from the previous discussion on the choice to load every pipeline present in a project into `_ProjectPipelines` regardless of if they're gonna be run or not. Solutions for this issue were proposed (https://github.com/kedro-org/kedro/pull/5340, https://github.com/kedro-org/kedro/pull/5346), but because of the current redesign work on `KedroSession`, we've decided that making such extensive changes on the way the session runs would not be ideal.

For this PR, a `pipelines_to_find` parameter is made available on the `find_pipelines()` function, so a list of strings containing the names of the pipelines to be run can be passed on the pipeline registry in the project itself. For example,

```py
def register_pipelines() -> dict[str, Pipeline]:
    """Register the project's pipelines.

    Returns:
        A mapping from pipeline names to ``Pipeline`` objects.
    """
    pipelines = find_pipelines(pipelines_to_find=["data_science", "reporting"])
    pipelines["__default__"] = sum(pipelines.values())
    return pipelines
```

## Checklist

### DIFF
--- a/kedro/framework/project/__init__.py
+++ b/kedro/framework/project/__init__.py
@@ -402,9 +402,9 @@ def find_pipelines(  # noqa: PLR0912, PLR0915
     https://docs.kedro.org/en/stable/build/pipeline_registry/
 
     Args:
+        raise_errors: If ``True``, raise an error upon failed discovery.
         pipelines_to_find: Optional list of pipeline names to load selectively.
             If ``None`` or contains ``"__default__"``, all pipelines are loaded.
-        raise_errors: If ``True``, raise an error upon failed discovery.
 
     Returns:
         A generated mapping from pipeline names to ``Pipeline`` objects.


kedro/framework/project/__init__.py
```diff
@@ -368,7 +368,9 @@ def _create_pipeline(pipeline_module: types.ModuleType) -> Pipeline | None:
     return obj
 
 
-def find_pipelines(raise_errors: bool = False) -> dict[str, Pipeline]:  # noqa: PLR0912
+def find_pipelines(  # noqa: PLR0912, PLR0915
+    pipelines_to_find: list[str] | None = None, raise_errors: bool = False
+) -> dict[str, Pipeline]:
```
merelcht, line 392: Changing the order of arguments is a breaking change, so `pipelines_to_find` will need to go last.

kedro/framework/project/__init__.py
```diff
@@ -398,8 +402,20 @@ def find_pipelines(raise_errors: bool = False) -> dict[str, Pipeline]:  # noqa:
             ``Pipeline`` object, or if the module import fails up front.
             If ``raise_errors`` is ``True``, see Raises section instead.
     """
+    # Safety check: If PACKAGE_NAME is None, we can't discover pipelines
+    if PACKAGE_NAME is None:
+        if pipelines_to_find is None or "__default__" in pipelines_to_find:
+            return {"__default__": pipeline([])}
+        return {}
```
merelcht: Is this needed for the change? I actually think it might be better to throw an exception here. If `PACKAGE_NAME` is None, this might be a sign of a bigger issue and should be flagged.

kedro/framework/project/__init__.py
```diff
@@ -398,8 +402,20 @@ def find_pipelines(raise_errors: bool = False) -> dict[str, Pipeline]:  # noqa:
...
+
     pipeline_obj = None
 
+    # Determine if specific pipelines were requested
+    load_all = pipelines_to_find is None or "__default__" in pipelines_to_find
+    requested_pipelines: set[str] | None = None
+    if not load_all and pipelines_to_find is not None:
+        requested_pipelines = set(pipelines_to_find)
```
merelcht: I don't think it's necessary to have the `pipelines_to_find is not None` check again, because having `load_all` already implies this:

```suggestion
load_all = pipelines_to_find is None or "__default__" in pipelines_to_find
requested_pipelines = None if load_all else set(pipelines_to_find)
```

kedro/framework/project/__init__.py
```diff
@@ -398,8 +402,20 @@ def find_pipelines(raise_errors: bool = False) -> dict[str, Pipeline]:  # noqa:
...
+
     # Handle the simplified project structure found in several starters.
```
merelcht: I was thinking that we probably only need to call this part of the logic if requesting all pipelines. This simplified structure is referring to a project where you only have one pipeline and so no `pipelines` directory. So in that case we can move the `if load_all:` to here.

kedro/framework/project/__init__.py
```diff
@@ -420,15 +436,55 @@ def find_pipelines(raise_errors: bool = False) -> dict[str, Pipeline]:  # noqa:
...
+            pipeline_obj = _create_pipeline(pipeline_module)
+            if pipeline_obj is not None:
+                pipelines_dict[pipeline_name] = pipeline_obj
+            elif raise_errors:
+                raise KeyError(f"Pipeline '{pipeline_name}' not found")
+
+        return pipelines_dict
+
+    # Load all pipelines (when no specific pipelines requested)
     for pipeline_dir in pipelines_package.iterdir():
```
merelcht: We can simplify this logic: right now there's duplication in the pipeline loading logic:

```python
pipeline_module_name = f"{PACKAGE_NAME}.pipelines.{pipeline_name}"
            try:
                pipeline_module = importlib.import_module(pipeline_module_name)
            except Exception as exc:
                if raise_errors:
                    raise ImportError(
                        f"An error occurred while importing the "
                        f"'{pipeline_module_name}' module."
                    ) from exc

                warnings.warn(
                    IMPORT_ERROR_MESSAGE.format(
                        module=pipeline_module_name, tb_exc=traceback.format_exc()
                    )
                )
                continue

            pipeline_obj = _create_pipeline(pipeline_module)
            if pipeline_obj is not None:
                pipelines_dict[pipeline_name] = pipeline_obj
            elif raise_errors:
                raise KeyError(f"Pipeline '{pipeline_name}' not found")

        return pipelines_dict
```

That code snippet now appears twice. 

Instead, you can just check the negative case: the case when the `pipeline_name` is not in the `requested_pipelines` and just `continue`. This would go on line 498.

```python
if requested_pipelines is not None and pipeline_name not in requested_pipelines:
       continue
```

tests/framework/project/test_pipeline_discovery.py
```diff
@@ -301,3 +301,137 @@ def create_pipeline(**kwargs) -> Pipeline:
...
+    "mock_package_name_with_pipelines",
+    [{"pipeline_one", "pipeline_two", "pipeline_three"}],
+    indirect=True,
+)
+def test_find_pipelines_selective_load(mock_package_name_with_pipelines):
+    configure_project(mock_package_name_with_pipelines)
+    pipelines = find_pipelines(pipelines_to_find=["pipeline_one"])
+    assert set(pipelines) == {"pipeline_one"}
+    assert "__default__" not in pipelines
+    assert sum(pipelines.values()).outputs() == {"pipeline_one"}
```
merelcht, line 316: Why do you need to take the `sum` here?

kedro/framework/project/__init__.py
lrcouto: Went with `requested_pipelines: set[str] | None = None if load_all else set(pipelines_to_find)  # type: ignore[arg-type]` otherwise it throws an [arg-type] linter error. I'll fix the rest and then I'll see if I can un-ignore it.

kedro/framework/project/__init__.py
lrcouto: This works and removes one of the `importlib.import_module()` calls. I just had to add a check at the end of the function to handle pipelines that were requested but do not exist on disk. I don't think there's any places before `find_pipelines()` where this kind of verification is done.

tests/framework/project/test_pipeline_discovery.py
lrcouto: To check if the whole pipeline is in the dict, not just the correct key. I think this is why they are being used anyway, they were there on the older tests 🤔

kedro/framework/project/__init__.py
```diff
@@ -398,37 +404,57 @@ def find_pipelines(raise_errors: bool = False) -> dict[str, Pipeline]:  # noqa:
...
+            if requested_pipelines is not None:
+                missing_str = ", ".join(sorted(requested_pipelines))
+                error_msg = f"Pipeline(s) not found: {missing_str}"
+                if raise_errors:
+                    raise KeyError(error_msg) from exc
+                warnings.warn(error_msg)
+                return {}
             return pipelines_dict
 
+    seen: set[str] = set()
```
merelcht, line 476: What does `seen` mean and why do we need to keep track of it?

lrcouto: This is to cover the case of when a pipeline is requested, but does not exist, for example if you tried to run a pipeline with the wrong name. This error handling was happening previously on that other loop that contained duplicate code.

The for loops gets pipelines with `pipelines_package.iterdir()` and tries to import them. But this will only go through modules that actually exist in the project. At the end I'm comparing these pipelines the function has already "seen" with those that were requested, and if there's any difference, it means the user requested a pipeline that does not exist.

merelcht: Got it!

ankatiyar:
For clarification, seems like for now, we're just letting the users narrow down which pipelines to load in their `pipeline_registry.py` but if someone runs `kedro run --pipeline=data_science` Kedro/Kedro session will still load all the pipelines (or just the ones specified by the user) and we'll deal with selective loading on our end later once the design of the new session is finalised?

lrcouto:
> For clarification, seems like for now, we're just letting the users narrow down which pipelines to load in their `pipeline_registry.py` but if someone runs `kedro run --pipeline=data_science` Kedro/Kedro session will still load all the pipelines (or just the ones specified by the user) and we'll deal with selective loading on our end later once the design of the new session is finalised?

Yes, exactly. The idea @merelcht and I were discussing was to have this pipeline selection be a parameter in the ProjectPipelines global object, which could be then set through the CLI, programmatically, etc But since the session is still being worked on, we thought it'd be better to just do the find_pipelines() part of it for now. This will allow users to manually set it on their registry if they need to, but will not affect anything or require any changes otherwise.

kedro/framework/project/__init__.py
```diff
@@ -400,12 +402,16 @@ def find_pipelines(raise_errors: bool = False) -> dict[str, Pipeline]:  # noqa:
     https://docs.kedro.org/en/stable/build/pipeline_registry/
 
     Args:
+        pipelines_to_find: Optional list of pipeline names to load selectively.
```
ankatiyar, line 406: Order in the docstring also needs to be updated I think

lrcouto: Done

PR #5403, opened 2026-02-25
Add validation framework exceptions and pydantic utilities

SajidAlamQB:
## Description
First in a series of PRs implementing the parameter validation framework (#5313). This PR adds the exceptions and utility functions that all subsequent validation components depend on.

See #5313 for the full implementation plan and architecture diagram.

## PR chain:

→ PR 1: Exceptions + utils (this PR)
PR 2: Source filters + tests
PR 3: Model factory + tests
PR 4: Type extractor + tests
PR 5: Parameter validator + tests
PR 6: Context integration + tests

## Developer Certificate of Origin
We need all contributions to comply with the [Developer Certificate of Origin (DCO)](https://developercertificate.org/). All commits must be signed off by including a `Signed-off-by` line in the commit message. [See our wiki for guidance](https://github.com/kedro-org/kedro/wiki/Guidelines-for-contributing-developers/).

If your PR is blocked due to unsigned commits, then you must follow the instructions under "Rebase the branch" on the GitHub Checks page for your PR. This will retroactively add the sign-off to all unsigned commits and allow the DCO check to pass.

## Checklist

- [x] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [x] Signed off each commit with a [Developer Certificate of Origin (DCO)](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository)
- [x] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [ ] Added tests to cover my changes
- [ ] Checked if this change will affect Kedro-Viz, and if so, communicated that with the Viz team

PR #5404, opened 2026-02-25
Add validation source filters

SajidAlamQB:
## Description

Second in a series of PRs implementing the parameter validation framework (#5313). Adds the `SourceFilter` abstraction and `ParameterSourceFilter` implementation used by the type extractor to identify which node inputs are parameters.

Depends on #5403 (exceptions + utils).

## PR chain:

PR 1: Exceptions + utils
→ PR 2: Source filters + tests (this PR)
PR 3: Model factory + tests
PR 4: Type extractor + tests

## Checklist


kedro/validation/source_filters.py
```diff
@@ -0,0 +1,37 @@
+"""Source filters for type extraction from different source types."""
+
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+
+
+class SourceFilter(ABC):
```
ravi-kumar-pilla, line 8: This is a clean implementation but may be an overkill considering we are only having parameters for now. So we can either get rid of this class and just do should_process -> extract_key inline where it is needed. I would still recommend having a separation like this.

kedro/validation/source_filters.py
```diff
@@ -0,0 +1,37 @@
...
+    """Abstract base class for source-specific filtering and key extraction."""
+
+    @abstractmethod
+    def should_process(self, source_name: str) -> bool:
+        """Determine if this filter should process the given source name."""
```
ravi-kumar-pilla, line 13: May be it is better to use `pass` or `...` (ellipses) as a placeholder. In `runner` we use pass as placeholder for abstract methods

kedro/validation/source_filters.py
rashidakanchwala: That makes sense. We can maybe keep this simple for now and introduce the abstraction once we add dataset validation.

merelcht:
Can you explain a bit more what the source filter is for and why it's a class? It seems a bit overkill for fairly little functionality. In the overall implementation (https://github.com/kedro-org/kedro/pull/5371) I can see it's used by the `TypeExtractor`, but my initial reaction is that it seems a bit too layered to have the sourcefilter wrapped in typeextractor which is in the init of the `ParameterValidation`. Is this a pattern more broadly used for validation frameworks?

kedro/validation/source_filters.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def should_process(self, source_name: str) -> bool:
+        """Check if source is a parameter source."""
+        return isinstance(source_name, str) and source_name.startswith("params:")
+
+    def extract_key(self, source_name: str) -> str:
+        """Extract parameter key from ``params:key`` format."""
+        return source_name.split(":", 1)[1]
+
+    def get_log_message(self, key: str, type_name: str) -> str:
```
merelcht, line 35: Formatting messages doesn't really seem to belong to filtering.

merelcht (commented):
I'm a bit hesitant to introduce this abstraction at this stage. It feel like it's solving a problem that doesn't exist yet. Instead the filtering could be reduced to a simple function inside the ParameterValidator with:
```
if not ds_name.startswith("params:"):
    continue
source_key = ds_name.split(":", 1)[1]
```

SajidAlamQB:
> Can you explain a bit more what the source filter is for and why it's a class? It seems a bit overkill for fairly little functionality. In the overall implementation (#5371) I can see it's used by the `TypeExtractor`, but my initial reaction is that it seems a bit too layered to have the sourcefilter wrapped in typeextractor which is in the init of the `ParameterValidation`. Is this a pattern more broadly used for validation frameworks?

Agreed removing the abstraction I'll close this PR and inline the filtering logic directly into the type extractor. We can introduce the abstraction later if dataset validation needs it

SajidAlamQB:
Closing this as there is consensus that the `SourceFilter` abstraction is premature. The filtering logic will be inlined directly into the type extractor in PR 4. We can introduce the abstraction later if dataset validation needs it 👍

PR #5407, opened 2026-02-26
Specify dataset name on dataset load failure

mvhensbergen:
## Description

Calling `load()` on a MemoryDataset that has not been saved yet yields the following error:

```
DatasetError: Data for MemoryDataset has not been saved yet.
```

While technically correct, the message is not very actionable because it doesn't say which MemoryDataset was called.

The suggested change (though possibly not the best solution) will raise the following error:

```
DatasetError: <dataset_name>: Data for MemoryDataset has not been saved yet.
``` 

## Development notes

I have implemented this by `try/catching` the load function on the Dataset and reraising a DatasetError with the name of the Dataset appended.

Some thoughts:
1. Is this the most generic place to do this or have I implemented a point solution?
2. Would it make sense to catch all exceptions instead of just DatasetError?

The items below I have not (yet) done as I would like to know first if this change is acceptable or if there is a better way to do it :-) 

## Checklist

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -186,9 +186,9 @@
         "filename": "tests/io/test_data_catalog.py",
         "hashed_secret": "15dd2c9ccec914f1470b4dccb45789844e49cf70",
         "is_verified": false,
-        "line_number": 584
+        "line_number": 585
       }
     ]
   },
-  "generated_at": "2026-03-24T10:54:45Z"
+  "generated_at": "2026-03-30T16:15:32Z"
 }


kedro/io/data_catalog.py
```diff
@@ -1052,7 +1052,12 @@ def load(self, ds_name: str, version: str | None = None) -> Any:
             extra={"markup": True},
         )
 
-        return dataset.load()
+        try:
+            ds_contents = dataset.load()
+        except DatasetError as e:
+            raise DatasetError(f"{ds_name}: {e}")
```
SajidAlamQB:
```suggestion
            raise DatasetError(f"{ds_name}: {e}") from e
```

SajidAlamQB (commented):
Good find thank you @mvhensbergen. A few things to fix before its ready.

Can we add some unit tests also update changelog.

The `DataCatalog.load()` is the right place. Also catching broadly and re-raising as `DatasetError` with the name is the right approach.

kedro/io/data_catalog.py
ankatiyar: +1

ankatiyar:
Hi @mvhensbergen thanks for the contribution, I think it does make sense to add the dataset name. 
Would you mind fixing the signoffs on the commits following these instructions: https://github.com/kedro-org/kedro/pull/5407/checks?check_run_id=64965631152 This will make the CI pass!

mvhensbergen:
Dear Kedro, sorry for the silence last couple of days but I hope to be able to work on it this week :-)

kedro/io/data_catalog.py
ravi-kumar-pilla, line 1029: can we update the docstring as we now raise `DatasetError` as well ?

RELEASE.md
```diff
@@ -2,6 +2,7 @@
 ## Major features and improvements
 * Added optional parameter validation that uses type hints of parameter inputs to auto-validate and instantiate Pydantic models/dataclasses with no impact on untyped parameters.
 ## Bug fixes and other changes
+* The `DatasetError` exception that is raised after calling `.load()` on a dataset now includes the name of the dataset for easier debugging.
```
ravi-kumar-pilla: [Nit] `DataCatalog.load()` now raises a `DatasetError` that includes the dataset name for easier debugging.

ravi-kumar-pilla:
Hi @mvhensbergen ,

Thank you for the contribution. Some minor comments to fix but otherwise looks good to me.

ankatiyar:
@ravi-kumar-pilla I've incorporated your suggestions :)

kedro/io/data_catalog.py
```diff
@@ -1052,7 +1053,12 @@ def load(self, ds_name: str, version: str | None = None) -> Any:
             extra={"markup": True},
         )
 
-        return dataset.load()
+        try:
```
ElenaKhaustova, line 1056: It seems like we don't need `ds_contents` variable here

```suggestion
try:
    return dataset.load()
except DatasetError as e:
    raise DatasetError(f"{ds_name}: {e}") from e
```

tests/io/test_data_catalog.py
```diff
@@ -96,6 +96,15 @@ def test_load_error(self, data_catalog):
         with pytest.raises(DatasetError, match=pattern):
             data_catalog.load("test")
 
+    def test_load_without_save_error(self, data_catalog):
```
ElenaKhaustova: The name of the test sounds a bit ambiguous, shall we rename it to something like `test_load_dataset_includes_name_in_error`?

tests/io/test_data_catalog.py
```diff
@@ -96,6 +96,15 @@ def test_load_error(self, data_catalog):
...
+        """Check the error when attempting to load a dataset
+        which was not saved to yet"""
+        name = "access_without_load"
+        catalog = DataCatalog(datasets={name: MemoryDataset()})
+        pattern = rf"{name}: Data for MemoryDataset has not been saved yet"
+        with pytest.raises(DatasetError, match=pattern):
+            catalog.load(name)
+
```
ElenaKhaustova, line 108: We should also update `test_load_error` test to verify the new name-prefixed format.

ankatiyar:
@ElenaKhaustova would be nice to do it for `save` as well, if it's too many changes we can do in a separate PR too?

ElenaKhaustova:
> @ElenaKhaustova would be nice to do it for `save` as well, if it's too many changes we can do in a separate PR too?

Yes, I was thinking about a separate PR as well. I'll open it as a follow-up.

Issue #5375, opened 2026-02-12
Add `--checkout` flag to `kedro new` for blank project with/without tools


ankatiyar:
## Description

Follow up for https://github.com/kedro-org/kedro-starters/issues/302
We want to add an option for selecting checkout branch similar to how it works with `--starters` but also for `kedro new` where the blank template is used
## Context

Context in https://github.com/kedro-org/kedro-starters/issues/302

## Possible Implementation

- Based on quick prototyping it seemed like just removing the error here would work but to be double checked by whoever is implementing it: https://github.com/kedro-org/kedro/blob/37ec20e7db2a214d03b37ca1e83656eaabe4718e/kedro/framework/cli/starters.py#L143
- Add tests

PR #5415, opened 2026-03-03
Allow checkout flag for default template projects

lrcouto:
## Description
https://github.com/kedro-org/kedro/issues/5375

The CLI `--checkout` flag can now be used on a new Kedro project from the default template, without a starter.

## Checklist

### DIFF
--- a/kedro/validation/__init__.py
+++ b/kedro/validation/__init__.py
@@ -1,0 +1,11 @@
+"""Kedro validation framework."""
+
+from .exceptions import ModelInstantiationError, ParameterValidationError
+from .utils import is_pydantic_class, is_pydantic_model
+
+__all__ = [
+    "ModelInstantiationError",
+    "ParameterValidationError",
+    "is_pydantic_class",
+    "is_pydantic_model",
+]
--- a/kedro/validation/exceptions.py
+++ b/kedro/validation/exceptions.py
@@ -1,0 +1,15 @@
+"""Custom exceptions for the validation framework."""
+
+from __future__ import annotations
+
+
+class ParameterValidationError(Exception):
+    """Raised when parameter validation fails."""
+
+    pass
+
+
+class ModelInstantiationError(ParameterValidationError):
+    """Raised when a typed model (Pydantic/dataclass) fails to instantiate from raw parameters."""
+
+    pass
--- a/kedro/validation/utils.py
+++ b/kedro/validation/utils.py
@@ -1,0 +1,41 @@
+"""Shared utilities for the validation framework."""
+
+from __future__ import annotations
+
+import dataclasses
+import inspect
+from typing import Any
+
+
+def is_pydantic_model(value: Any) -> bool:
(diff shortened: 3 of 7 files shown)


features/new.feature
```diff
@@ -5,6 +5,11 @@ Feature: New Kedro project
     When I run a non-interactive kedro new without starter
     Then the expected project directories and files should be created
 
+  Scenario: Create a new kedro project with --checkout flag and no starter
```
ankatiyar: This test (and the associated step) is probably not needed. We need the `--checkout` for the specific case of testing where tool is `pyspark` and/or `example` is yes and since those are pulled from `kedro-starters`, that's where the e2e tests are and should be. 

For when example pipeline is no and/or any of the first 5 tools are selected, we create a project with the current branch (afaik)

lrcouto: No problem, I just wanted to make sure everything was being tested properly. I'll remove it.

tests/framework/cli/starters/test_new_from_cli_flags.py
```diff
@@ -302,3 +294,78 @@ def test_starter_flag_with_example_flag(self, fake_kedro_cli):
             "Cannot use the --starter flag with the --example and/or --tools flag."
             in result.output
         )
+
+
+@pytest.mark.usefixtures("chdir_to_tmp")
```
ElenaKhaustova, line 299: I think tests are missing `assert result.exit_code == 0`, so if the CLI command fails (e.g., exit code 1 or 2 due to some validation error or exception), `mock_cookiecutter` might never be called at all.

tests/framework/cli/starters/test_new_from_cli_flags.py
```diff
@@ -7,7 +7,10 @@
 import pytest
 from click.testing import CliRunner
 
-from kedro.framework.cli.starters import _convert_tool_short_names_to_numbers
+from kedro.framework.cli.starters import (
```
ElenaKhaustova, line 10: **This comment relates to another unchanged line.**

When `--checkout` is used without `--starter` and without selecting PySpark or example pipeline (i.e., the plain blank template path), the checkout value gets passed through to cookiecutter but has no effect.

This means `kedro new --checkout some-branch --tools none --example no --name my-project` silently accepts `--checkout` but does nothing with it. This could confuse users who expect it to have an effect.

Shall we consider either:

- Raising a warning when `--checkout` is provided but the resolved template is a local path (no PySpark, no example, no starter).
- Or documenting in the help text that `--checkout` only applies when a remote template is involved (starters, PySpark tool, or example pipeline).

ElenaKhaustova, line 10: **This comment relates to another unchanged line.**

Shall we also update `CHECKOUT_ARG_HELP` text to reflect broader usage?

lrcouto: I think both of those make sense to be added, actually.

PR #5419, opened 2026-03-04
Add validation model factory

SajidAlamQB:
## Description

Third in a series of PRs implementing the parameter validation framework (#5313). Adds ModelFactory which handles instantiation of Pydantic models and dataclasses from raw config values.

Depends on #https://github.com/kedro-org/kedro/pull/5404 (source filters).

## PR chain:

PR 1: Exceptions + utils
PR 2: Source filters + tests
→ PR 3: Model factory + tests (this PR)
PR 4: Type extractor + tests

## Checklist

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -17,12 +17,14 @@
 * Added a warning when a node returns a value but is defined with `outputs=None`, clarifying that the return value is ignored.
 * Added `preserve_logging` flag to `configure_project()` to prevent runtime-added logging handlers from being overwritten when `configure_project()` is called after custom handlers have been attached (e.g. in a long-running server process such as FastAPI).
 * Added util method `find_config_file()` to handle different config file extensions (.yml, .yaml)
+* Added reusable suggestion functionality for mistyped pipeline names using `kedro run`
 
 ## Documentation changes
 ## Community contributions
 * [aziq](https://github.com/aziq)
 * [zhubaobao2024](https://github.com/zhubaobao2024)
 * [Camille Coeurjoly](https://github.com/Camille1992)
+* [sinanpl](https://github.com/sinanpl)
 
 # Release 1.2.0
 ## Major features and improvements
--- a/kedro/framework/cli/utils.py
+++ b/kedro/framework/cli/utils.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import difflib
 import importlib
 import logging
 import shlex
@@ -25,6 +24,8 @@ from typing import IO, Any
 
 import click
 from omegaconf import OmegaConf
+
+from kedro.utils import get_close_matches
 
 CONTEXT_SETTINGS = {"help_option_names": ["-h", "--help"]}
 MAX_SUGGESTIONS = 3
@@ -91,7 +92,7 @@ def forward_command(
 def _suggest_cli_command(
     original_command_name: str, existing_command_names: Iterable[str]
 ) -> str:
-    matches = difflib.get_close_matches(
+    matches = get_close_matches(
         original_command_name, existing_command_names, MAX_SUGGESTIONS, CUTOFF
     )
 
--- a/kedro/framework/session/session.py
+++ b/kedro/framework/session/session.py
@@ -8,6 +8,7 @@ import logging.config
 import os
 import subprocess
 import sys
+import textwrap
 import traceback
 from copy import deepcopy
 from pathlib import Path
@@ -28,7 +29,7 @@ from kedro.io.core import generate_timestamp
 from kedro.io.data_catalog import SharedMemoryDataCatalog
 from kedro.pipeline.pipeline import Pipeline
 from kedro.runner import AbstractRunner, ParallelRunner, SequentialRunner
-from kedro.utils import find_kedro_project
+from kedro.utils import find_kedro_project, get_close_matches
 
 if TYPE_CHECKING:
     from collections.abc import Iterable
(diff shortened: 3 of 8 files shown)


tests/validation/test_model_factory.py
```diff
@@ -0,0 +1,60 @@
...
+        assert result is raw
+
+    def test_instantiate_unsupported_type_str(self, model_factory):
+        raw = "hello"
+        result = model_factory.instantiate("name", raw, str)
+        assert result is raw
+
+    def test_error_message_includes_context(self, model_factory):
+        raw = {"wrong_field": "value"}
+        with pytest.raises(ModelInstantiationError, match="SampleDataclass"):
```
merelcht: Can we test a bit more from the error message not just `SampleDataclass`? This will also make it easier to understand what's expected to happen here.

kedro/validation/model_factory.py
```diff
@@ -0,0 +1,64 @@
...
+
+        except Exception as exc:
+            raise ModelInstantiationError(
+                f"Failed to instantiate {model_type.__name__} for "
+                f"source '{source_key}': {exc}"
+            ) from exc
+
+    def _instantiate_pydantic(self, raw_value: Any, model_type: type) -> Any:
+        """Handle Pydantic model instantiation."""
+        return model_type.model_validate(raw_value)  # type: ignore[attr-defined]
```
ravi-kumar-pilla: We should explicitly say we support this validation from pydantic v2 as model_validate is available in only v2. @SajidAlamQB may be in the docstrings somewhere or in our docs

kedro/validation/model_factory.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def _instantiate_dataclass(self, raw_value: Any, model_type: type) -> Any:
+        """Handle dataclass instantiation."""
+        if isinstance(raw_value, dict):
+            return model_type(**raw_value)
+        else:
+            return model_type()
```
ravi-kumar-pilla: what happens if raw_value is not dict. This actually drops the value if not dict. Should we handle it in a different way ?

kedro/validation/model_factory.py
SajidAlamQB: Added docstring 👍

kedro/validation/model_factory.py
```diff
@@ -0,0 +1,69 @@
...
+import logging
+from typing import Any
+
+from .exceptions import ModelInstantiationError
+from .utils import is_pydantic_class
+
+logger = logging.getLogger(__name__)
+
+
+class ModelFactory:
```
merelcht: Does this actually need to be a class? There's no `__init__` here and so we're essentially just constructing a model factory to then call the methods. Why not just call the methods directly?

tests/validation/test_model_factory.py
SajidAlamQB: Updated the error message test to be more verbose

kedro/validation/model_factory.py
SajidAlamQB: Addressed its now just module based

kedro/validation/model_factory.py
```diff
@@ -0,0 +1,70 @@
...
+            f"source '{source_key}': {exc}"
+        ) from exc
+
+
+def _instantiate_pydantic(raw_value: Any, model_type: type) -> Any:
+    """Handle Pydantic model instantiation.
+
+    Note: Requires Pydantic v2+ (uses `model_validate`).
+    """
+    return model_type.model_validate(raw_value)  # type: ignore[attr-defined]
```
merelcht, line 60: Do we need to do any error handling here in case this call fails?

kedro/validation/model_factory.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def instantiate_model(source_key: str, raw_value: Any, model_type: type) -> Any:
+    """Instantiate a typed model from a raw value.
+
+    Supports Pydantic models and dataclasses. Returns the raw value
+    unchanged for unsupported types.
+
+    Args:
+        source_key: The source key being processed.
```
merelcht: It's not immediately clear to me what `source_key` means 🤔 I get what it does a bit more from looking at the tests, but wondering if there's a term we can use that's easier to understand.

tests/validation/test_model_factory.py
```diff
@@ -0,0 +1,59 @@
...
+    def test_instantiate_pydantic_model(self):
+        raw = {"test_size": 0.2, "random_state": 3}
+        result = instantiate_model("model_options", raw, SamplePydanticModel)
+        assert isinstance(result, SamplePydanticModel)
+        assert result.test_size == 0.2
+        assert result.random_state == 3
+
+    def test_instantiate_pydantic_invalid_data(self):
+        raw = {"test_size": "banana", "random_state": 3}
+        with pytest.raises(ModelInstantiationError):
```
merelcht: Can we check the error messages here to make the tests easier to read? (I'm guessing here it's failing because "banana" is a str and not float?

tests/validation/test_model_factory.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def test_instantiate_unsupported_type_returns_raw(self):
+        raw = 42
+        result = instantiate_model("threshold", raw, int)
+        assert result is raw
+
+    def test_instantiate_unsupported_type_str(self):
+        raw = "hello"
+        result = instantiate_model("name", raw, str)
+        assert result is raw
```
merelcht: Can we check here that the debugging message is emitted?

kedro/validation/model_factory.py
SajidAlamQB: Maybe `param_key` is more intuitive

kedro/validation/model_factory.py
SajidAlamQB: Yes It's already handled. The try/except in `instantiate_model` catches any exception from `_instantiate_pydantic` and wraps it in `ModelInstantiationError`.

tests/validation/test_model_factory.py
```diff
@@ -0,0 +1,74 @@
...
+        result = instantiate_model("model_options", raw, SamplePydanticModel)
+        assert isinstance(result, SamplePydanticModel)
+        assert result.test_size == 0.2
+        assert result.random_state == 3
+
+    def test_instantiate_pydantic_invalid_data(self):
+        raw = {"test_size": "banana", "random_state": 3}
+        with pytest.raises(
+            ModelInstantiationError,
+            match="Failed to instantiate SamplePydanticModel for parameter 'model_options'",
```
merelcht: Can we add more details about what exactly went wrong with the instantiation? Or does that show up in the {exc} part of the exception? I'm thinking if `model_options` is a really large dict, it might take forever for someone to debug which field has the wrong type.

SajidAlamQB: The {exc} in the error message already includes Pydantic full output. I updated the test to show this.

PR #5421, opened 2026-03-04
Add validation type extractor

SajidAlamQB:
## Description

Fourth in a series of PRs implementing the parameter validation framework (#5313). Adds TypeExtractor which walks pipeline node signatures to discover typed parameter requirements.

Depends on #https://github.com/kedro-org/kedro/pull/5419 (model factory)

## PR chain:

PR 1: Exceptions + utils
PR 2: Source filters + tests
PR 3: Model factory + tests
→ PR 4: Type extractor + tests (this PR)

## Checklist

### DIFF
--- a/tests/validation/test_type_extractor.py
+++ b/tests/validation/test_type_extractor.py
@@ -99,22 +99,26 @@ class TestExtractTypesFromPipelines:
         assert result["eval_config"] == SampleDataclass
 
     def test_conflicting_types_warns(self, type_extractor, caplog):
-        """When two pipelines define different types for the same key, warn.
+        """When two pipelines define different types for the same key, warn."""
 
-        Uses MagicMock because get_type_hints cannot resolve types
-        defined inside a function scope.
-        """
-        node_a = MagicMock()
-        node_a.func = lambda opts: None
-        node_a.func.__annotations__ = {"opts": _TypeA}
-        node_a.func.__name__ = "func_a"
-        node_a.inputs = ["params:shared_config"]
+        def func_a(opts: _TypeA) -> None:
+            pass
 
-        node_b = MagicMock()
-        node_b.func = lambda opts: None
-        node_b.func.__annotations__ = {"opts": _TypeB}
-        node_b.func.__name__ = "func_b"
-        node_b.inputs = ["params:shared_config"]
+        def func_b(opts: _TypeB) -> None:
+            pass
+
+        node_a = kedro_node(
+            func=func_a,
+            inputs="params:shared_config",
+            outputs="output_a",
+            name="node_a",
+        )
+        node_b = kedro_node(
+            func=func_b,
+            inputs="params:shared_config",
+            outputs="output_b",
+            name="node_b",
+        )
 
         pipeline_a = MagicMock()
         pipeline_a.nodes = [node_a]


kedro/validation/model_factory.py
```diff
@@ -46,6 +46,8 @@ def instantiate(self, source_key: str, raw_value: Any, model_type: type) -> Any:
                 )
                 return raw_value
 
+        except ModelInstantiationError:
```
ravi-kumar-pilla: Is this intended ?

kedro/validation/type_extractor.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def __init__(self, source_filter: SourceFilter) -> None:
+        """Initialize the type extractor.
+
+        Args:
+            source_filter: Source filter to determine which node inputs to process.
+        """
+        self.source_filter = source_filter
+
+    def extract_types_from_pipelines(self) -> dict[str, type]:
```
ravi-kumar-pilla, line 21: I remember we do validation across all pipelines but did that change ? If someone runs specific pipeline, do we still validate pipelines that are not running in the session ? If so, we should try to get that filter either in this series of PRs or before we release this feature. This will effect teams working on multiple pipelines in tandem.

kedro/validation/type_extractor.py
```diff
@@ -0,0 +1,197 @@
...
+            dataset_to_arg.update(node_inputs)
+        elif isinstance(node_inputs, list | tuple):
+            sig_param_names = list(signature.parameters.keys())
+            for idx, ds in enumerate(node_inputs):
+                if idx < len(sig_param_names):
+                    dataset_to_arg[ds] = sig_param_names[idx]
+
+        return dataset_to_arg
+
+    def resolve_nested_path(self, data: dict, path: str) -> Any:
```
ravi-kumar-pilla: This feels like a utility method. Can be moved to utils ?

kedro/validation/type_extractor.py
```diff
@@ -0,0 +1,197 @@
...
+        for key in keys:
+            if isinstance(value, dict) and key in value:
+                value = value[key]
+            else:
+                return None
+
+        return value
+
+    def set_nested_value(self, data: dict, path: str, value: Any) -> None:
+        """Set a value at a dot-separated path, creating intermediate dicts as needed.
```
ravi-kumar-pilla: This feels like a utility method. Can be moved to utils ?

kedro/validation/model_factory.py
SajidAlamQB: Yes itsintentional `_instantiate_dataclass` raises `ModelInstantiationError` directly so we re-raise it to avoid double wrapping in the generic except Exception handler

kedro/validation/type_extractor.py
SajidAlamQB: Fair this is a limitation currently if we get this series of PRs in I will create a follow-up. For now it doesn't break anything.

kedro/validation/__init__.py
```diff
@@ -2,13 +2,23 @@
...
     "ParameterValidationError",
-    "instantiate_model",
+    "TypeExtractor",
     "get_typed_fields",
+    "instantiate_model",
     "is_pydantic_class",
     "is_pydantic_model",
+    "resolve_nested_path",
+    "set_nested_value",
 ]
```
ravi-kumar-pilla, line 12: Do we need helper methods in __all__ ?

__all__ = [
    "ModelInstantiationError",
    "ParameterValidationError",
    "TypeExtractor",
    "instantiate_model",
    "is_pydantic_class",
    "is_pydantic_model",
]

OR

__all__ = [
    "ModelInstantiationError",
    "ParameterValidationError",
    "TypeExtractor",
    "instantiate_model",
]

is enough ?

SajidAlamQB: Good point trimmed it now.

kedro/validation/type_extractor.py
```diff
@@ -0,0 +1,145 @@
...
+if TYPE_CHECKING:
+    from kedro.pipeline import Pipeline
+    from kedro.pipeline.node import Node
+
+logger = logging.getLogger(__name__)
+
+_PARAMS_PREFIX = "params:"
+
+
+class TypeExtractor:
```
merelcht, line 18: Does it make sense to call this `ParameterTypeExtractor` to make it extra clear this is for parameters only?

kedro/validation/type_extractor.py
```diff
@@ -0,0 +1,145 @@
...
+        Walks all pipelines (except ``__default__``) and inspects node
+        function signatures for type-annotated ``params:`` inputs.
+
+        Returns:
+            Dictionary mapping parameter keys to their expected types.
+        """
+        try:
+            from kedro.framework.project import pipelines
+
+            pipeline_dict = dict(pipelines)
```
merelcht, line 33: Isn't `pipelines` a dict already? I'm wondering if this casting and error handling is needed. My guess is that if pipelines can't be imported there's a bigger issue and the project doesn't run at all, or am I missing something?

kedro/validation/utils.py
```diff
@@ -39,3 +41,49 @@ def get_typed_fields(value: Any) -> dict[str, Any] | None:
...
+    for key in keys:
+        if isinstance(value, dict) and key in value:
+            value = value[key]
+        else:
+            return None
+
+    return value
+
+
+def set_nested_value(data: dict, path: str, value: Any) -> None:
```
merelcht: Are these specific to parameters? Might again be good to add that in the function name to clarify.

tests/validation/test_type_extractor.py
```diff
@@ -0,0 +1,253 @@
...
+        pipeline = MagicMock()
+        pipeline.nodes = [node]
+
+        with patch(
+            "kedro.framework.project.pipelines",
+            {"__default__": pipeline, "data_science": pipeline},
+        ):
+            result = type_extractor.extract_types_from_pipelines()
+
+        assert "eval_config" in result
```
merelcht, line 75: It's not immediately clear to me how this tests that the default pipeline was skipped.

tests/validation/test_type_extractor.py
```diff
@@ -0,0 +1,253 @@
...
+            create=True,
+        ):
+            result = type_extractor.extract_types_from_pipelines()
+        assert result == {}
+
+    def test_skips_default_pipeline(self, type_extractor):
+        def my_func(options: SampleDataclass) -> None:
+            pass
+
+        node = MagicMock()
```
merelcht: Can't we use the actual `Node` class?

kedro/validation/type_extractor.py
ravi-kumar-pilla: During initial design, I kept this as we might extend it beyond parameters

kedro/validation/utils.py
ravi-kumar-pilla: Sajid can add on this but for the parameter validation yes, this is for parameters only. I suggested to add this in utils as it seems like a utility function which can be extended to resolve any nested path. 

If you feel we should rename this to only parameters path, we can move this inside type extractor

kedro/validation/type_extractor.py
ravi-kumar-pilla: I think pipelines is a lazy-dict and we try to materialize it by doing dict(pipelines). We do similar thing on Kedro-Viz too. 

> I'm wondering if this casting

I do not think this is just for casting.

kedro/validation/type_extractor.py
merelcht: But wouldn't we then create a different TypeExtractor class for that? It looks like all functions in the class are specific for parameters.

kedro/validation/utils.py
merelcht: ah right fair point. In that case I would maybe just rename it to `resolve_nested_dict_path` and `set_nested_dict_value` to make it clear this is for dictionary objects and not e.g. file path resolution.

kedro/validation/type_extractor.py
ravi-kumar-pilla: Initial design had source filter injected within this if I remember correctly. Based on the source filter key, we can modify the functions within this class. Now that it is removed, this class feels like only for parameters. 

I would still say, we can have this as TypeExtractor and once we want dataset type extraction, we can have functions within this class to extract based on the source key (dataset or parameters). 

The other way would be to make a BaseTypeExtractor and then have ParameterTypeExtractor, DatasetTypeExtractor etc. But this for me seems overkill and since the focus is TypeExtraction (irrespective of source key), having functions specific to source key and calling them within TypeExtractor seems like a good choice.

Happy to hear any suggestions. 

Thank you

kedro/validation/type_extractor.py
```diff
@@ -0,0 +1,145 @@
...
+                        logger.warning(
+                            "Conflicting type requirements for parameter '%s': "
+                            "%s (existing) vs %s (from pipeline '%s'). "
+                            "Using %s.",
+                            key,
+                            existing_type.__name__,
+                            new_type.__name__,
+                            pipeline_name,
+                            new_type.__name__,
+                        )
```
merelcht, line 62: One thing that I forgot to mention before is that I'm wondering if this should actually throw an error instead of silently warn. If a user has given the same parameter a different type across pipelines it seems like a mistake that should be flagged during validation. What do you think?

kedro/validation/type_extractor.py
merelcht: That makes sense, thanks Ravi. I'm thinking that maybe it makes sense to make this class stateful (have an `__init__`) which can control the input needed for validation e.g. catalog when we add data validation. And perhaps `pipelines` should also be a class argument and instead of being initialised inside `extract_types_from_pipelines` it's inserted at the validator call level. 

I'm okay with merging this as is and revisiting in the final PR of this series when all is tied together, because that will make it clearer.

SajidAlamQB: Agree with Ravi's reasoning, lets keep as `TypeExtractor` for now happy to revisit in PR6 when everything is tied together as you suggested.

kedro/validation/type_extractor.py
SajidAlamQB: The pipelines is a lazy `_ProjectPipelines` object so as Ravi said the dict materialises it.

kedro/validation/utils.py
SajidAlamQB: Makes sense will do

kedro/validation/type_extractor.py
SajidAlamQB: Im leaning towards keeping it as a warning for now as common pattern is pipelines sharing parameters with compatible but technically different types (e.g. a base class vs subclass). 

Erroring would block the run entirely. But happy to adjust if the team prefers stricter.

tests/validation/test_type_extractor.py
SajidAlamQB: Now uses separate pipelines with distinct params and asserts `default_only` is not in the result

kedro/validation/type_extractor.py
ravi-kumar-pilla: I think a warning should be good for now considering we are not blocking the `kedro run`. Also validation layer should validate the type provided by the user and user providing multiple types may be intentional given the value out of the parameters might be the same for different pipeline runs even if the type is different.

ravi-kumar-pilla: This can be a code error or unintentional from the user but this is not a concern of the validation layer. So warning must help user inform of the possible mistake they made. Let me know what you think of this. Thank you

merelcht: This sounds reasonable to me. Let's make sure to document that clearly too so people know what to expect. If they don't like it we can always change the behaviour based on feedback.

tests/validation/test_type_extractor.py
```diff
@@ -0,0 +1,275 @@
...
+        node_a.func = lambda opts: None
+        node_a.func.__annotations__ = {"opts": _TypeA}
+        node_a.func.__name__ = "func_a"
+        node_a.inputs = ["params:shared_config"]
+
+        node_b = MagicMock()
+        node_b.func = lambda opts: None
+        node_b.func.__annotations__ = {"opts": _TypeB}
+        node_b.func.__name__ = "func_b"
+        node_b.inputs = ["params:shared_config"]
```
merelcht: can these be real nodes too like in the other tests?

PR #5422, opened 2026-03-04
Add validation parameter validator

SajidAlamQB:
## Description

Fifth in a series of PRs implementing the parameter validation framework (#5313). Adds ParameterValidator which orchestrates type extraction and model instantiation into a single validate_raw_params entry point.

Depends on #https://github.com/kedro-org/kedro/pull/5421 (type extractor).

## PR chain:

PR 1: Exceptions + utils
PR 2: Source filters + tests
PR 3: Model factory + tests
PR 4: Type extractor + tests
→ PR 5: Parameter validator + tests (this PR)

## Checklist

### DIFF
--- a/docs/integrations-and-plugins/deltalake_versioning.md
+++ b/docs/integrations-and-plugins/deltalake_versioning.md
@@ -12,7 +12,7 @@ In this example, you will use the `spaceflights-pandas` starter project which ha
 kedro new --starter spaceflights-pandas
 ```
 
-Kedro offers various connectors in the `kedro-datasets` package to interact with Delta tables: [`pandas.DeltaTableDataset`](https://docs.kedro.org/projects/kedro-datasets/en/stable/api/kedro_datasets/pandas.DeltaTableDataset/), [`spark.DeltaTableDataset`](https://docs.kedro.org/projects/kedro-datasets/en/stable/api/kedro_datasets/spark.DeltaTableDataset/), [`spark.SparkDataset`](https://docs.kedro.org/projects/kedro-datasets/en/stable/api/kedro_datasets/spark.SparkDataset/), [`databricks.ManagedTableDataset`](https://docs.kedro.org/projects/kedro-datasets/en/stable/api/kedro_datasets/databricks.ManagedTableDataset/), and [`ibis.FileDataset`](https://docs.kedro.org/projects/kedro-datasets/en/stable/api/kedro_datasets/ibis.FileDataset/) support the delta table format. In this tutorial, we will use the `pandas.DeltaTableDataset` connector to interact with Delta tables using Pandas DataFrames. To install `kedro-datasets` along with dependencies required for Delta Lake, add the following line to your `requirements.txt`:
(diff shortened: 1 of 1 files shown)


kedro/validation/parameter_validator.py
```diff
@@ -0,0 +1,113 @@
...
+    def __init__(self) -> None:
+        """Initialize parameter validator."""
+        self.type_extractor = TypeExtractor(ParameterSourceFilter())
+        self.model_factory = ModelFactory()
+
+    def get_pipeline_requirements(self) -> dict[str, type]:
+        """Get all parameter type requirements from available pipelines."""
+        return self.type_extractor.extract_types_from_pipelines()
+
+    def apply_validation(self, raw_params: dict, requirements: dict) -> dict:
```
ravi-kumar-pilla: can we make this private method ?

kedro/validation/parameter_validator.py
```diff
@@ -0,0 +1,113 @@
...
+
+class ParameterValidator:
+    """Orchestrates type extraction and model instantiation for parameters."""
+
+    def __init__(self) -> None:
+        """Initialize parameter validator."""
+        self.type_extractor = TypeExtractor(ParameterSourceFilter())
+        self.model_factory = ModelFactory()
+
+    def get_pipeline_requirements(self) -> dict[str, type]:
```
ravi-kumar-pilla: can we make this private method ?

kedro/validation/parameter_validator.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class ParameterValidator:
+    """Orchestrates type extraction and model instantiation for parameters."""
+
+    def __init__(self) -> None:
+        """Initialize parameter validator."""
+        self.type_extractor = TypeExtractor()
+
+    def _get_pipeline_requirements(self) -> dict[str, type]:
```
merelcht: This method seems a bit unnecessary. You can just call `self.type_extractor.extract_types_from_pipelines()` directly below in `validate_raw_params`

tests/validation/test_parameter_validator.py
```diff
@@ -0,0 +1,131 @@
...
+from kedro.validation.exceptions import ParameterValidationError
+
+from .conftest import SampleDataclass, SamplePydanticModel
+
+
+class TestParameterValidator:
+    def test_validate_raw_params_no_requirements(self, parameter_validator):
+        with patch.object(
+            parameter_validator, "_get_pipeline_requirements", return_value={}
+        ):
```
merelcht, line 21: I'm still thinking about making `TypeExtractor` having an init and having `pipelines` as a parameter. That way we can also simplify tests by not needing to patch things or a full project, but instead just pass a mock pipeline. I think the key question though is at what point it makes sense to load all pipelines and only doing that for validation if the user wants to validate their parameters.

ravi-kumar-pilla: > I think the key question though is at what point it makes sense to load all pipelines

Right now this is done when someone access parameters via context.params. This will trigger everything in the validation layer.

> I'm still thinking about making TypeExtractor having an init and having pipelines as a parameter

I agree on this. I think @SajidAlamQB created a follow up ticket somewhat related to this - https://github.com/kedro-org/kedro/issues/5443

kedro/validation/parameter_validator.py
ravi-kumar-pilla: +1

tests/validation/test_parameter_validator.py
SajidAlamQB: Yes aligns with what you mentioned on PR4 about making the class stateful. Happy to address that in PR6 (context integration) where we can wire up the pipeline loading at the right point in the flow. That way we can see the full picture before deciding where pipelines should be loaded.

merelcht: That sounds good!

Issue #5425, opened 2026-03-05
`KedroServiceSession` v1: Stateless session


ankatiyar:

DimedS:
Thanks, @ankatiyar! I like the proposal to start with a mostly stateless Kedro session and then introduce opportunities to load certain components into the session and reuse them across different runs.

A few things I’d like to mention:

1. **Pipeline loading during session initialisation**

   It may make sense to load pipelines during session initialisation, since as I understand they are not expected to change between runs. If pipelines are loaded when the session is created, `session.run()` should be faster even for the first run.

2. **Alignment with the Kedro API layer**

   We should also consider how this interacts with the Kedro API layer. When introducing the new `KedroSession`, the API layer code should be adjusted from the prototype implemented in #5370.

   Since we are already close to implementing the new session design, it might be better to ship the API layer **based on the new session**, to avoid introducing breaking changes later.

   This would likely mean that the API layer should also be released as an **experimental feature**, since it will depend on the experimental `KedroSession`. Wdyt @rashidakanchwala , @merelcht ?

rashidakanchwala:
> We should also consider how this interacts with the Kedro API layer. When introducing the new KedroSession, the API layer code should be adjusted from the prototype implemented in https://github.com/kedro-org/kedro/pull/5370.

The original plan was to have the API layer first support the existing KedroSession, and then add support for the new session additively, mainly because both pieces were evolving at the same time.

That said, since KedroServiceSession v1 is much more lightweight than initially planned, I’m comfortable with the API layer supporting both the original session and the new session (althought let's break into multiple smaller PRs) 

> Since we are already close to implementing the new session design, it might be better to ship the API layer based on the new session, to avoid introducing breaking changes later.

Given that we intend to support both sessions and the existing one isn’t being deprecated at this stage, this shouldn’t introduce breaking changes.

ankatiyar:
Thanks @DimedS, I agree, we should front-load pipelines, this will make the runs individually faster.
WRT API Layer (cc @rashidakanchwala), I think it could be based on `KedroServiceSession` from the get-go just so we're not writing and re-writing the code for this. Both these features can co-evolve together experimentally at first. From the user-perspective it doesn't make a difference if we use old session or new session since the functionality more or less remains the same w/ v1 of the `KedroServiceSession` atleast. 🤔

merelcht:
Thanks @ankatiyar! Sharing my thoughts on the points you mentioned in the description:

[...]

PR #5441, opened 2026-03-10
`KedroServiceSession` v1

### LINKED ISSUE #5427: Implement `KedroServiceSession` v1

ankatiyar:
## Description

PROTOTYPE FOR https://github.com/kedro-org/kedro/issues/5425

## 🚀 Prototype: KedroServiceSession (stateless multi-run session)

This PR introduces a first prototype of a new session abstraction and a stateless multi-run session implementation.

#### 2. Add `KedroServiceSession` (prototype)
- New session type designed for **multiple runs per session**
- Supports **per-run `runtime_params` injection**
- Uses a **stateless approach**:
  - creates a fresh `KedroContext` and `DataCatalog` for each run
~~- Preloads pipelines once at session creation for faster execution~~ (will be done in Subsequent PR/issue)

#### 3. Design decisions (v1)
- No session store (`_store`) for now
- ~~`session_id`~~ + `run_id` used for versioning (only timestamp)
- API kept close to existing `KedroSession` to minimise friction

### 📌 Motivation

Current `KedroSession` enforces a 1:1 mapping between session and run.  
This prototype explores a session model better suited for:

- long-lived processes (e.g. services, APIs)
- repeated pipeline execution
- dynamic runtime parameter injection

### ⚠️ Notes

- Coverage for `KedroServiceSession` is temporarily excluded (testing to be added in a branched off PR)
- Some aspects (naming, versioning strategy, store support) are still under discussion

---

Closes #5427

### DIFF
--- a/docs/overrides/welcome.html
+++ b/docs/overrides/welcome.html
@@ -206,7 +206,7 @@
                     <img src="https://img.shields.io/badge/license-Apache%202.0-blue.svg" alt="License is Apache 2.0">
                 </a>
                 <a href="https://pypi.org/project/kedro/" target="_blank" rel="noopener">
-                    <img src="https://img.shields.io/badge/3.9%20%7C%203.10%20%7C%203.11%20%7C%203.12%20%7C%203.13-blue.svg" alt="Python version 3.9, 3.10, 3.11, 3.12, 3.13">
+                    <img src="https://img.shields.io/badge/3.10%20%7C%203.11%20%7C%203.12%20%7C%203.13-blue.svg" alt="Python version 3.10, 3.11, 3.12, 3.13">
                 </a>
                 <a href="https://pypi.org/project/kedro/" target="_blank" rel="noopener">
                     <img src="https://badge.fury.io/py/kedro.svg" alt="PyPI package version">
--- a/docs/stylesheets/mermaid.css
+++ b/docs/stylesheets/mermaid.css
@@ -1,5 +1,5 @@
 /* Mermaid light mode - Kedro branding */
-[data-md-color-scheme="default"] {
+[data-md-color-scheme="light"] {
   --md-mermaid-node-bg-color: #FFF7E2;
   --md-mermaid-node-fg-color: #FFC900;
   --md-mermaid-label-bg-color: #FFF7E2;
@@ -8,7 +8,7 @@
 }
 
 /* Mermaid dark mode - Kedro branding */
-[data-md-color-scheme="slate"] {
+[data-md-color-scheme="dark"] {
   --md-mermaid-node-bg-color: #636363;
   --md-mermaid-node-fg-color: #FFC900;
   --md-mermaid-label-bg-color: #636363;
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -8,13 +8,13 @@ theme:
   custom_dir: docs/overrides/
   palette:
     # Palette toggle for light mode
-    - scheme: default
+    - scheme: light
       toggle:
         icon: material/white-balance-sunny
         name: Switch to dark mode
 
     # Palette toggle for dark mode
-    - scheme: slate
+    - scheme: dark
       toggle:
         icon: material/white-balance-sunny
         name: Switch to light mode
@@ -287,7 +287,7 @@ extra_css:
   - stylesheets/themes.css
   - stylesheets/footer.css
   - stylesheets/welcome.css
-   - stylesheets/mermaid.css
+  - stylesheets/mermaid.css
 
 extra_javascript:
   - javascript/readthedocs.js
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,7 +77,7 @@ test = [
     "types-requests",
 ]
 docs = [
-    "mkdocs>=1.6.1",
+    "mkdocs>=1.6.1,<2.0",
     "mkdocs-material>=9.6.11",
     "mkdocs-material-extensions>=1.3.1",
     "mkdocs-mermaid2-plugin>=1.2.1",


kedro/framework/session/service_session.py
```diff
@@ -0,0 +1,257 @@
...
+        )
+        try:
+            run_result = runner.run(
+                filtered_pipeline,
+                catalog,
+                hook_manager,
+                run_id=run_id,
+                only_missing_outputs=only_missing_outputs,
+            )
+            self._run_called = True
```
ElenaKhaustova: Is this a leftover from `KedroSession` where it enforces single-run?

kedro/framework/session/service_session.py
```diff
@@ -0,0 +1,257 @@
...
+        _register_hooks_entry_points(hook_manager, settings.DISABLE_HOOKS_FOR_PLUGINS)
+        self._hook_manager = hook_manager
+        self.pipelines = pipelines
+        self.env = env
+        self._conf_source = conf_source or str(
+            self._project_path / settings.CONF_SOURCE
+        )
+
+    @classmethod
+    def create(
```
ElenaKhaustova, line 83: This will cause an interchangeability problem - the two sessions expect runtime_params at different points in the lifecycle, and the CLI can only call one way.

The CLI calls `settings.SESSION_CLASS.create(env=env, conf_source=conf_source, runtime_params=params)`. Since `create()` here doesn't accept runtime_params or save_on_close, swapping to `KedroServiceSession` via `SESSION_CLASS` in settings will fail.

The same problem is with the `run()` method - pipeline_name vs pipeline_names.

kedro/framework/session/service_session.py
```diff
@@ -0,0 +1,257 @@
...
+        env: str | None = None,
+        conf_source: Path | str | None = None,
+    ) -> KedroServiceSession:
+        """Create a new instance of the session."""
+        validate_settings()
+
+        session = cls(
+            project_path=project_path,
+            session_id=session_id
+            or generate_timestamp(),  ## Should this be uuid instead?
```
ElenaKhaustova: For a long-running service session, the creation timestamp is not very meaningful - a session could be created at startup and live for hours. What we should care about is when each run happened. IMO, a session needs a stable unique identifier, not a point-in-time marker.

This points toward flipping the current approach:

- session_id -> UUID (stable identity for a long-lived session)
- run_id -> timestamp (tells you when the run happened, naturally sortable)

This also matters for the `save_version`. Right now this produces `<timestamp>_<uuid>` as a directory name for versioned datasets. If we flip it, it becomes `<uuid>_<timestamp>`, which is still long but at least the timestamp part (the bit users actually care about when browsing the filesystem) is at the end and visible without truncation.

rashidakanchwala: Could we use uuid7 here? I mentioned it in my [comment](https://github.com/kedro-org/kedro/issues/5425#issuecomment-4045558372)

kedro/framework/session/service_session.py
ankatiyar: I think we might have to adjust the code where CLI `kedro run` is called (also `ipython`) to pass `runtime_params` to `run` function instead of `create()` when `KedroServiceSession` is used. 🤔 

re: `pipeline_names` - we're going to deprecate `pipeline_name` in favour of `pipelines_names` anyway so I didn't include it

kedro/framework/session/service_session.py
```diff
@@ -0,0 +1,256 @@
...
+        """An instance of the config loader."""
+        config_loader_class = settings.CONFIG_LOADER_CLASS
+        return config_loader_class(  # type: ignore[no-any-return]
+            conf_source=self._conf_source,
+            env=self.env,
+            runtime_params=runtime_params,
+            **settings.CONFIG_LOADER_ARGS,
+        )
+
+    # TODO: Maybe the merging of runtime params and config params should be done here instead of in the context?
```
ankatiyar: @ElenaKhaustova @merelcht @DimedS 
I would like your opinion here as well, does it make sense for the `KedroContext` to be updated to just be a dataclass and do the merging of params + runtime_params in the session?

kedro/framework/session/service_session.py
ElenaKhaustova: I think there are two ways to interpret the UUIDv7 suggestion, and both have trade-offs:

### Option A: UUIDv7 replaces the entire `save_version`

Each run just gets a single UUIDv7 as its version:
```bash
data/model/019e3f5a-7b2c-7d4e-8f6a-9b0c1d2e3f4a/
data/model/019e3f5b-8c3d-7e5f-9a7b-0c1d2e3f4a5b/
data/model/019e3f5a-9d4e-7f6a-8b9c-1d2e3f4a5b6c/
```

These are time-ordered and unique, but you've lost any trace of which session each run belongs to. If you have multiple sessions running concurrently (e.g. two API clients), there's no way to tell from the filesystem which outputs came from which session. You'd have to dig through logs or hook data to reconstruct that.

### Option B: UUIDv7 as `run_id`, kept inside `session_id + "_" + run_id`

We keep the composite format, but use UUIDv7 for the run part:
```bash
data/model/a1b2c3d4-e5f6-7890-abcd-ef1234567890_019e3f5a-7b2c-7d4e-8f6a-9b0c1d2e3f4a/
data/model/a1b2c3d4-e5f6-7890-abcd-ef1234567890_019e3f5b-8c3d-7e5f-9a7b-0c1d2e3f4a5b/
```

Session grouping is preserved, but the path is 73+ characters of hex and dashes. Not practical for anyone browsing their data directories.

### Compare with `<uuid_session>_<timestamp_run>`

You get session grouping (shared prefix), chronological ordering (timestamp), and the human-readable part tells you exactly when the run happened. The session UUID provides the uniqueness, so the `run_id` doesn't need to be collision-resistant on its own — it just needs to be distinct within a session, which timestamps are in any realistic scenario.

```bash
data/model/a1b2c3d4-...2026-03-19T14.30.22.123Z/
data/model/a1b2c3d4-...2026-03-19T14.35.10.456Z/
```

kedro/framework/session/service_session.py
ElenaKhaustova: I think making the context a full dataclass may break the `CONTEXT_CLASS` extension point, the `after_context_created` hook, and the IPython `context.catalog`/`context.params` workflow.

Should we consider making the config loader the single source of truth for parameter resolution, so the context doesn't need its own copy, and we just stop passing `runtime_params` to the context constructor altogether?

kedro/framework/session/service_session.py
ankatiyar: I was thinking: `session_id` = uuid and `run_id` = timestamp and we use `session_id` + `_` + `run_id` as the versioning format. 

Do you think it makes sense to add another folder level in the versioned dataset so that the path looks like: `data/06_models/session_id/run_id/regressor.pkl`

ElenaKhaustova: > Do you think it makes sense to add another folder level in the versioned dataset so that the path looks like: data/06_models/session_id/run_id/regressor.pkl

Yeah, I think it will improve usability

rashidakanchwala: > Session grouping is preserved, but the path is 73+ characters of hex and dashes. Not practical for anyone browsing their data directories.

Good point, `<uuid_session>_<timestamp_run>` makes sense

I don't like further nesting, I prefer concatenation even tho it will get a bit longer

kedro/framework/session/service_session.py
merelcht: This is a very good point though. We need to ensure that it's possible to swap session types easily. It's fine if that doesn't happen yet in this first iteration, but let's not forget.

kedro/framework/session/service_session.py
```diff
@@ -0,0 +1,249 @@
...
+        self._logger.info("Run ID: %s", run_id)
+        save_version = run_id
+
+        context = self.load_context(runtime_params)
+        pipeline_names = pipeline_names or ["__default__"]
+        combined_pipeline = Pipeline([])
+        for name in pipeline_names:
+            try:
+                combined_pipeline += pipelines[name]
+            except KeyError as exc:
```
merelcht, line 168: In `KedroSession` a contributor recently added a suggestion message when the pipeline is not found. Shall we add that here too?

kedro/framework/session/session.py
```diff
@@ -75,14 +76,6 @@ def _jsonify_cli_context(ctx: click.core.Context) -> dict[str, Any]:
     }
 
 
-class KedroSessionError(Exception):
```
merelcht, line 78: I think technically speaking moving the error class is a breaking change for users who import it directly into their code. Does it need to be moved?

kedro/framework/session/service_session.py
ankatiyar: I'll work on that in a separate branch so this PR is not too heavy

kedro/framework/session/session.py
ankatiyar: Because it's being imported in kedro/framework/session/session.py, I believe you can still do from kedro.framework.session.session import KedroSessionError

kedro/framework/session/abstract_session.py
```diff
@@ -35,3 +35,11 @@ def __enter__(self) -> AbstractSession:
 
     def __exit__(self, _exc_type: Any, _exc_value: Any, _tb: Any) -> None:
         self.close()
+
+
+class KedroSessionError(Exception):
+    """``KedroSessionError`` raised by ``KedroSession``
+    in the case that multiple runs are attempted in one session.
```
merelcht: I guess this docstring isn't entirely accurate anymore now this error is also used for the service session.

merelcht (approved):
One minor thing: it might be a good idea to add in the doc string of `service_session.py` that this is a feature that is under active development and might change between versions as long as there's no docs for this.

ankatiyar:
@merelcht I added the note and updated the docstrings. I was planning to add minimal docs to go along with this, do you think we need to hold off on that incase there are changes to the API?

merelcht:
> @merelcht I added the note and updated the docstrings. I was planning to add minimal docs to go along with this, do you think we need to hold off on that incase there are changes to the API?

Minimal docs sound good to me and we can add the note about this class still changing there too.

kedro/framework/session/service_session.py
```diff
@@ -0,0 +1,247 @@
...
+            "load_versions": load_versions,
+            "runtime_params": runtime_params or {},
+            "pipeline_names": pipeline_names,
+            "namespaces": namespaces,
+            "runner": getattr(runner, "__name__", str(runner)),
+            "only_missing_outputs": only_missing_outputs,
+        }
+
+        # Run the runner
+        hook_manager = self._hook_manager
```
rashidakanchwala, line 233: Hi Ankita ... quick question: I’m trying to understand how hooks behave in a multi-run session, especially when they’re defined at the session level.

Do these hooks share any data across runs? If there is shared state, how does that impact subsequent runs?

tests/framework/session/conftest.py
```diff
@@ -23,14 +30,18 @@
...
 if TYPE_CHECKING:
-    from pathlib import Path
-
-    from kedro.framework.context.context import KedroContext
     from kedro.io import DataCatalog
 
 logger = logging.getLogger(__name__)
 
 MOCK_PACKAGE_NAME = "fake_package"
+_FAKE_PROJECT_NAME = "fake_project"
```
rashidakanchwala: nit - for consistency
```suggestion
MOCK_PROJECT_NAME = "fake_project"
```

tests/framework/session/conftest.py
```diff
@@ -374,9 +386,161 @@ def mock_session(mocker, mock_settings, mock_package_name, tmp_path):
...
     # KedroSession eagerly validates that a project's settings.py is correct by
     # importing it. settings.py does not actually exists as part of this test suite
     # since we are testing session in isolation, so the validation is patched.
     mocker.patch("kedro.framework.session.session.validate_settings")
+    mocker.patch("kedro.framework.session.service_session.validate_settings")
+
+
+@pytest.fixture
+def fake_project(tmp_path, mock_package_name):
+    fake_project_dir = Path(tmp_path) / "fake_project"
```
rashidakanchwala:
```suggestion
    fake_project_dir = Path(tmp_path) / _FAKE_PROJECT_NAME
```

kedro/framework/session/service_session.py
ankatiyar: As far as I understand, this initialises the hooks that come from `hooks.py` and settings. This is just the plugin manager and this shouldn't vary between the runs. For the hook execution, for example for `before_pipeline_run`, `after_pipeline_run` - these we run within the `run()` method with the run related config passed on to it. Here we create a copy, I've kept it similar to `KedroSession`. I dont think the hooks implementation shares any states.

kedro/framework/session/service_session.py
```diff
@@ -0,0 +1,247 @@
...
+        try:
+            run_result = runner.run(
+                filtered_pipeline,
+                catalog,
+                hook_manager,
+                run_id=run_id,
+                only_missing_outputs=only_missing_outputs,
+            )
+        except Exception as error:
+            hook_manager.hook.on_pipeline_error(
```
rashidakanchwala, line 246: Do we have a test for this?

tests/framework/session/test_service_session.py
```diff
@@ -0,0 +1,532 @@
...
+            )
+
+            session.run(runner=mock_runner, pipeline_names=_FAKE_PIPELINE_NAME)
+            second_run_id = (
+                session._hook_manager.hook.before_pipeline_run.call_args.kwargs[
+                    "run_params"
+                ]["run_id"]
+            )
+
+        assert first_run_id != second_run_id
```
rashidakanchwala, line 558: Can we also assert that the session id remains the same

tests/framework/session/test_service_session.py
```diff
@@ -0,0 +1,532 @@
...
+
+    def test_load_context(self, fake_project):
+        session = KedroServiceSession.create(project_path=fake_project)
+        context = session.load_context()
+        config_loader = session._get_config_loader()
+        assert isinstance(context, KedroContext)
+        assert isinstance(config_loader, OmegaConfigLoader)
+        assert context.config_loader == config_loader
+
+    def test_load_context_with_runtime_params(self, fake_project):
```
rashidakanchwala, line 64: Should we do a test to show session multiple runs  with multiple runtime parameters as well?

tests/framework/session/conftest.py
ankatiyar: Made the suggested changes

kedro/framework/session/service_session.py
```diff
@@ -0,0 +1,247 @@
...
+            conf_source=conf_source,
+            env=env,
+        )
+        return session
+
+    @property
+    def _logger(self) -> logging.Logger:
+        return logging.getLogger(__name__)
+
+    def __enter__(self) -> KedroServiceSession:
```
ElenaKhaustova: It looks like we redefine identical implementations of `__enter__` and `__exit__` which `AbstractSession` already provides

tests/framework/session/test_service_session.py
```diff
@@ -0,0 +1,562 @@
...
+            },
+        )
+        mock_runner.__name__ = "SequentialRunner"
+
+        with KedroServiceSession.create(
+            project_path=fake_project, session_id="fake_id"
+        ) as session:
+            session.run(
+                runner=mock_runner,
+                pipeline_names=_FAKE_PIPELINE_NAME,
```
ElenaKhaustova: Shouldn't it be `pipeline_names=[_FAKE_PIPELINE_NAME]` as pipeline_names expects `list[str] | None`?

kedro/framework/session/service_session.py
```diff
@@ -0,0 +1,247 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+    from kedro.config import AbstractConfigLoader
+    from kedro.framework.context import KedroContext
+
+
+class KedroServiceSession(AbstractSession):
+    """
+    KedroServiceSession offers a session implementation that allows for multiple runs and data injection.
```
ElenaKhaustova: Shall we provide a usage example here as we do for `KedroSession`?

tests/framework/session/test_service_session.py
```diff
@@ -0,0 +1,562 @@
...
+        outside_dir = tmp_path / "outside"
+        outside_dir.mkdir()
+        pyproject_path = tmp_path / "pyproject.toml"
+        if pyproject_path.exists():
+            pyproject_path.unlink()
+
+        # Change the current working directory to the outside directory
+        monkeypatch.chdir(outside_dir)
+
+        # Create a session and set run called to True - no need to run a full session, we can verify the logging message
```
ElenaKhaustova: I think this might be copied from then old test

PR #5442, opened 2026-03-10
Fix version path traversal

ravi-kumar-pilla:
## Description

Added checks to validate safe versions

## Development notes

- Added a check for filtering unsafe versions
- Raises respective error messages incase of an unsafe version
- Updated tests 

## Checklist

### DIFF
--- a/kedro/framework/cli/utils.py
+++ b/kedro/framework/cli/utils.py
@@ -507,7 +507,8 @@ def _split_load_versions(ctx: click.Context, param: Any, value: str) -> dict[str
         if _is_unsafe_version(version):
             raise KedroCliError(
                 f"Version string '{version}' is not allowed. "
-                "Version strings must not be absolute paths or contain '..' components."
+                "Version strings must be a single non-empty path component with no "
+                "path separators ('/' or '\\') and must not be '.' or '..'."
             )
         load_versions_dict[load_version_list[0]] = version
 
--- a/kedro/utils.py
+++ b/kedro/utils.py
@@ -10,7 +10,7 @@ import re
 import warnings
 from collections.abc import Callable, Iterable
 from functools import wraps
-from pathlib import Path, PurePosixPath, PureWindowsPath
+from pathlib import Path
 from typing import Any
 from urllib.parse import urlsplit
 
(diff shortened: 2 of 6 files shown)


kedro/utils.py
```diff
@@ -311,3 +311,19 @@ def find_config_file(
...
+
+def _is_unsafe_version(version: str) -> bool:
+    """Return True if the version string contains path traversal or absolute path components.
+
+    Checks against both POSIX and Windows path semantics to ensure platform-independent
+    protection against directory traversal attacks.
+    """
+    posix = PurePosixPath(version)
+    windows = PureWindowsPath(version)
+    return (
```
ElenaKhaustova: The issue description says that `path separators used as directory components` should be rejected. Currently, versions like `foo/bar` pass validation and create unexpected subdirectory structures. Also, `"." ` resolves to the parent path, which can cause confusing behaviour.

Shall we cover these cases as well?

- Empty strings `""`
- Forward slashes `"foo/bar"`
- Backslashes `"foo\bar"`
- Single dot `"."` which resolves `filepath/./name` to `filepath/name`

kedro/framework/cli/utils.py
```diff
@@ -503,7 +503,13 @@ def _split_load_versions(ctx: click.Context, param: Any, value: str) -> dict[str
                 f"'dataset_name:YYYY-MM-DDThh.mm.ss.sssZ',"
                 f"found {load_version} instead"
             )
-        load_versions_dict[load_version_list[0]] = load_version_list[1]
+        version = load_version_list[1]
+        if _is_unsafe_version(version):
+            raise KedroCliError(
+                f"Version string '{version}' is not allowed. "
```
ElenaKhaustova, line 509: Shall we also update error message text to mention path separators, since that should also be rejected (see comment below)?

ElenaKhaustova (commented):
Thank you, @ravi-kumar-pilla!

It looks mostly good, but I think we might be missing some cases mentioned in the original issue description (see the comment).

Could you please double-check that, and if so, extend the missing edge-case tests as well?

tests/framework/cli/test_cli.py
```diff
@@ -1020,6 +1031,30 @@ def test_fail_split_load_versions(self, fake_project_cli, fake_metadata):
...
+    )
+    def test_fail_split_load_versions_path_traversal(
+        self, fake_project_cli, fake_metadata, unsafe_version
+    ):
+        result = CliRunner().invoke(
+            fake_project_cli,
+            ["run", "--load-versions", unsafe_version],
+            obj=fake_metadata,
+        )
+        assert result.exit_code
```
ElenaKhaustova, line 1055:
```suggestion
        assert result.exit_code == 2
```

ravi-kumar-pilla: Hi @ElenaKhaustova , thanks for the review. The exit_code is 1 and not 2 similar to - https://github.com/kedro-org/kedro/blob/main/tests/framework/cli/test_cli.py#L1014

ElenaKhaustova: Good point, you're right.

PR #5449, opened 2026-03-12
Add Node and Pipeline snapshots for the inspection layer

ravi-kumar-pilla:
## Description

Resolves #5429 

**NOTE:** This PR is based on https://github.com/kedro-org/kedro/pull/5447

## Development notes

- Updated kedro/inspection/models.py
- Updated kedro/inspection/snapshot.py
- Added `tests/inspection/test_node_pipeline_snapshot.py`

## Developer Certificate of Origin
We need all contributions to comply with the [Developer Certificate of Origin (DCO)](https://developercertificate.org/). All commits must be signed off by including a `Signed-off-by` line in the commit message. [See our wiki for guidance](https://github.com/kedro-org/kedro/wiki/Guidelines-for-contributing-developers/).

If your PR is blocked due to unsigned commits, then you must follow the instructions under "Rebase the branch" on the GitHub Checks page for your PR. This will retroactively add the sign-off to all unsigned commits and allow the DCO check to pass.

## Checklist

- [x] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [x] Signed off each commit with a [Developer Certificate of Origin (DCO)](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository)
- [x] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [x] Added tests to cover my changes
- [x] Checked if this change will affect Kedro-Viz, and if so, communicated that with the Viz team

PR #5447, opened 2026-03-12
Inspection API implementation

ravi-kumar-pilla:
## Description

Resolves #5428 

## Development notes

**Added:**

- kedro/inspection/__init__.py
- kedro/inspection/models.py (ProjectMetadataSnapshot only)
- kedro/inspection/snapshot.py (_build_project_metadata_snapshot only)
- tests/inspection/__init__.py
- tests/inspection/test_metadata_snapshot.py

## Developer Certificate of Origin
We need all contributions to comply with the [Developer Certificate of Origin (DCO)](https://developercertificate.org/). All commits must be signed off by including a `Signed-off-by` line in the commit message. [See our wiki for guidance](https://github.com/kedro-org/kedro/wiki/Guidelines-for-contributing-developers/).

If your PR is blocked due to unsigned commits, then you must follow the instructions under "Rebase the branch" on the GitHub Checks page for your PR. This will retroactively add the sign-off to all unsigned commits and allow the DCO check to pass.

## Checklist

- [x] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [x] Signed off each commit with a [Developer Certificate of Origin (DCO)](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository)
- [x] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [x] Added tests to cover my changes
- [x] Checked if this change will affect Kedro-Viz, and if so, communicated that with the Viz team

PR #5458, opened 2026-03-19
Add Project snapshot for the inspection layer

### LINKED ISSUE #5431: `ProjectSnapshot` - model, orchestrator, tests
### Description

Implement `ProjectSnapshot` and `_build_project_snapshot()` — the full orchestrator that calls `bootstrap_project()` and assembles all snapshot components into a single result. This ticket also owns `_resolve_factory_patterns`, the cross-cutting step that needs both catalog and pipeline data and belongs in neither individual builder.

### Context

`_build_catalog_snapshot` and `_build_pipeline_snapshots` are each pure — they know nothing about each other. `_resolve_factory_patterns` is the explicit composition step that takes both outputs and adds factory-matched dataset entries referenced in pipeline node I/O but absent from `catalog.yml`. The orchestrator calls all steps in sequence and owns the composition.

### Scope

**`kedro/inspection/models.py`** — add:
```python
@dataclass
class ProjectSnapshot:
    metadata: ProjectMetadataSnapshot
    pipelines: list[PipelineSnapshot]
    datasets: dict[str, DatasetSnapshot]
    parameter_keys: list[str] = field(default_factory=list)
```

**`kedro/inspection/snapshot.py`** — add:

Private helper:
```python
def _resolve_factory_patterns(
    datasets: dict[str, DatasetSnapshot],
    pipeline_snapshots: list[PipelineSnapshot],
) -> dict[str, DatasetSnapshot]:
    """Return a new datasets dict enriched with factory-matched and MemoryDataset entries
    referenced in pipeline node I/O but absent from catalog.yml.
    Does not mutate the input dict.
    """
```

Orchestrator:
```python
def _build_project_snapshot(
    project_path: Path | str,
    env: str = "local",
) -> ProjectSnapshot:
    """Build a complete ProjectSnapshot.
    Calls bootstrap_project(), then assembles each component independently.
    _resolve_factory_patterns runs as an explicit step after both catalog and pipeline
    data are available.
    """
    metadata_raw = bootstrap_project(project_path)
    metadata = _build_project_metadata_snapshot(metadata_raw)
    pipeline_snapshots = _build_pipeline_snapshots()
    datasets, parameter_keys = _build_catalog_snapshot(project_path, env)
    datasets = _resolve_factory_patterns(datasets, pipeline_snapshots)
    return ProjectSnapshot(
        metadata=metadata,
        pipelines=pipeline_snapshots,
        datasets=datasets,
        parameter_keys=parameter_keys,
    )
```

**`tests/inspection/test_project_snapshot.py`**:
- `ProjectSnapshot`: construction, all fields, `dataclasses.asdict()` output
- `_resolve_factory_patterns`: factory-matched and MemoryDataset entries added; existing catalog entries not duplicated; returns new dict without mutating input
- `_build_project_snapshot`: all components assembled correctly
- `parameter_keys` contains names from `parameters.yml`, never values
- `json.dumps(dataclasses.asdict(snapshot))` produces valid JSON

### Deliverable

- Updated `kedro/inspection/models.py`
- Updated `kedro/inspection/snapshot.py`
- `tests/inspection/test_project_snapshot.py`

ravi-kumar-pilla:
## Description

Resolves #5431 

**NOTE:**
1. This PR is based out of https://github.com/kedro-org/kedro/pull/5447, https://github.com/kedro-org/kedro/pull/5449, https://github.com/kedro-org/kedro/pull/5451 . So for additional context have a look at the other PRs

## Development notes

* Added helper utility for assembling Kedro inspection snapshots.
* Updated `kedro/inspection/models.py`
* Updated `kedro/inspection/snapshot.py`
* Added `tests/inspection/test_helper.py`
* Added `tests/inspection/test_project_snapshot.py`

## Checklist

### DIFF
--- a/kedro/inspection/helper.py
+++ b/kedro/inspection/helper.py
@@ -72,8 +72,7 @@ def _resolve_factory_patterns(
         pipelines: List of pipeline snapshots whose node inputs/outputs are inspected.
 
     Returns:
-        A new mapping that contains all original entries plus any newly resolved
-        concrete dataset entries derived from factory patterns.
+        A new mapping containing only concrete dataset entries
     """
     pattern_configs = {
         name: config
@@ -92,7 +91,12 @@ def _resolve_factory_patterns(
             pipeline_ds_names.update(node_snapshot.inputs)
             pipeline_ds_names.update(node_snapshot.outputs)
 
-    result = dict(datasets)
+    # Start from non-pattern entries
+    result = {
+        name: snap
+        for name, snap in datasets.items()
+        if not CatalogConfigResolver.is_pattern(name)
+    }
 
     for ds_name in pipeline_ds_names:
         if ds_name in result:
--- a/tests/inspection/test_helper.py
+++ b/tests/inspection/test_helper.py
@@ -149,6 +149,8 @@ class TestResolveFactoryPatterns:
         # 'companies' direct entry should not be overwritten
         assert result["companies"].type == "pandas.CSVDataset"
         assert result["companies"].filepath is None
+        # factory pattern key is stripped
+        assert "{namespace}.{name}" not in result
 
     def test_factory_pattern_resolved_for_pipeline_input(self):
         """Pipeline input matching a factory pattern gets a concrete entry."""
@@ -199,8 +201,8 @@ class TestResolveFactoryPatterns:
         assert "data_science.processed" in result
         assert result["data_science.processed"].filepath == "data/processed.csv"
 
-    def test_original_pattern_entry_preserved(self):
-        """The factory pattern entry itself is still present in the result."""
+    def test_pattern_entry_removed_from_result(self):
+        """Factory pattern keys are stripped; only concrete resolved entries remain."""
         catalog_config = {
             "{namespace}.{name}": {"type": "pandas.CSVDataset"},
         }
@@ -211,7 +213,8 @@ class TestResolveFactoryPatterns:
         }
         pipelines = [_make_pipeline_snapshot([_make_node_snapshot(["ns.ds"], ["out"])])]
         result = _resolve_factory_patterns(catalog_config, datasets, pipelines)
-        assert "{namespace}.{name}" in result
+        assert "{namespace}.{name}" not in result
+        assert "ns.ds" in result
 
     def test_no_match_dataset_not_added(self):
         """A pipeline dataset that does not match any pattern is not added."""
@@ -229,6 +232,7 @@ class TestResolveFactoryPatterns:
         ]
         result = _resolve_factory_patterns(catalog_config, datasets, pipelines)
         assert "plain_name" not in result
+        assert "{namespace}.{name}" not in result
 
     def test_pattern_without_filepath_gives_none_filepath(self):
         """Resolved entries have None filepath when the pattern has no filepath."""
@@ -244,8 +248,8 @@ class TestResolveFactoryPatterns:
         result = _resolve_factory_patterns(catalog_config, datasets, pipelines)
         assert result["ns.ds"].filepath is None
 
-    def test_empty_pipelines_returns_unchanged(self):
-        """No pipelines means no enrichment needed."""
+    def test_empty_pipelines_strips_patterns(self):
+        """No pipelines means no concrete datasets are resolved; patterns are still stripped."""
         catalog_config = {
             "{namespace}.{name}": {"type": "pandas.CSVDataset"},
         }
@@ -255,7 +259,7 @@ class TestResolveFactoryPatterns:
             ),
         }
         result = _resolve_factory_patterns(catalog_config, datasets, [])
-        assert result == datasets
+        assert result == {}
 
     def test_more_specific_pattern_wins_regardless_of_catalog_order(self):
         """Pattern with higher specificity is chosen even when defined after a generic one."""
@@ -294,6 +298,9 @@ class TestResolveFactoryPatterns:
             result["data_science.int_customers"].filepath
             == "data/int_customers.parquet"
         )
+        # both pattern keys are stripped
+        assert "{namespace}.{name}" not in result
+        assert "{namespace}.int_{name}" not in result
 
     def test_multiple_nodes_across_pipelines(self):
         """Dataset names from multiple pipelines and nodes are all considered."""
@@ -313,3 +320,4 @@ class TestResolveFactoryPatterns:
         assert "p1.ds_a" in result
         assert "p1.ds_b" in result
         assert "p2.ds_c" in result
+        assert "{ns}.{name}" not in result


kedro/inspection/snapshot.py
```diff
@@ -100,3 +114,41 @@ def _build_pipeline_snapshots() -> list[PipelineSnapshot]:
...
+    project_path = Path(project_path)
+
+    metadata = bootstrap_project(project_path)
+    config_loader = _make_config_loader(project_path, env=env)
+
+    metadata_snapshot = _build_project_metadata_snapshot(metadata)
+    pipeline_snapshots = _build_pipeline_snapshots(dict(pipelines))
+    dataset_snapshots = _build_dataset_snapshots(config_loader)
+
+    # [TODO: Need to confirm with the team if we should resolve it or
```
rashidakanchwala: Agree, this should be resolved. It would be odd if someone asks for a project snapshot and gets back raw factory patterns instead of concrete dataset names.

kedro/inspection/helper.py
```diff
@@ -0,0 +1,129 @@
...
+        Returns an empty list when no parameters configuration is found.
+    """
+    try:
+        params: dict = config_loader["parameters"]
+    except (KeyError, MissingConfigException):
+        return []
+    return sorted(params.keys())
+
+
+def _resolve_factory_patterns(
```
ElenaKhaustova, line 60: This looks like a partial reimplementation of `CatalogConfigResolver.match_pattern` / `resolve_dataset_pattern`. It uses `parse(pattern, ds_name)`, interpolates filepaths via `format_map(parsed.named)`, and sorts by specificity — all things `CatalogConfigResolver` already does.

This duplication means two places to update if the pattern matching rules evolve (e.g., typed placeholders, regex patterns, catch-all handling). We tried to avoid that when redesigning `DataCatalog`.

We should consider whether the resolver can delegate to `CatalogConfigResolver` methods instead, or at least extract the shared logic into a reusable utility. `CatalogConfigResolver` already has everything needed — pattern sorting, matching, and config resolution, and it's lightweight (no dataset instantiation, no I/O). Also, if something is missing, we can expose an API or extend it, but it should be done through `CatalogConfigResolver` rather than duplicating logic.

kedro/inspection/snapshot.py
ElenaKhaustova: Yes, we should resolve. The snapshot exists to give consumers a complete structural view of the project. Without resolution, nodes reference datasets that don't appear in the datasets dict, and every plugin has to reimplement pattern matching independently — the exact problem the KEP aims to fix.

kedro/inspection/snapshot.py
```diff
@@ -81,9 +89,15 @@ def _node_to_snapshot(node: Node) -> NodeSnapshot:
     )
 
 
-def _build_pipeline_snapshots() -> list[PipelineSnapshot]:
+def _build_pipeline_snapshots(
+    pipelines: dict[str, Any],
```
ElenaKhaustova: Here we accept a `pipelines` parameter, which shadows the module-level `from kedro.framework.project import pipelines` import still present at the top of `snapshot.py`. Shall we consider renaming the parameter to something like `pipeline_dict` or `pipeline_registry` to avoid confusion — especially since the module-level `pipelines` is still used in `_build_project_snapshot` as `dict(pipelines)`.

tests/inspection/test_dataset_snapshot.py
```diff
@@ -28,6 +28,13 @@ def catalog_config():
     }
 
 
+@pytest.fixture
```
ElenaKhaustova: This on is added but never used

tests/inspection/test_project_snapshot.py
```diff
@@ -0,0 +1,243 @@
...
+    def test_datasets_populated(self):
+        result = _build_project_snapshot(self.project_path)
+        assert result.datasets is self.dataset_snapshots
+
+    def test_parameters_populated(self):
+        result = _build_project_snapshot(self.project_path)
+        assert result.parameters == ["model_options", "test_size"]
+
+    def test_bootstrap_called_with_project_path(self, mocker):
+        mock_bootstrap = mocker.patch(
```
ElenaKhaustova: Do we need to re-patch `bootstrap_project` here if `_patch_orchestrator` already patches it with the same target and identical return value? It looks like the reason is to capture a mock reference for the assertion — could we store the mock on self in the autouse fixture instead (e.g. `self.mock_bootstrap = mocker.patch(...)`) and assert on it directly?

kedro/inspection/helper.py
ravi-kumar-pilla: Yes, I explored that but we have some issues -

1. CatalogConfigResolver.__init__ calls self._resolve_credentials(config, credentials) immediately. The inspection layer ideally should not touch credentials. Our approach never reaches credentials at all.

2. For any dataset name that doesn't match a user pattern, resolve_pattern() matches the default runtime pattern and returns {"type": "kedro.io.MemoryDataset"}. That would silently create spurious DatasetSnapshot entries for every transient intermediate or parameter dataset in the pipeline - polluting the snapshot with false catalog entries.

3. resolve_pattern() returns a complete dataset config with every key (load_args, save_args, versioned, nested credentials, etc.). DatasetSnapshot only stores type and filepath. Using a heavy resolver to extract two fields is wasteful and introduces complexity that has nothing to do with inspection.

> at least extract the shared logic into a reusable utility

I will have a look at this again, but I was hesitant to change anything in CatalogConfigResolver as I wanted team's opinions on whether we need resolution of patterns. Please suggest if you see any better way to handle this. Thank you

ElenaKhaustova: The core logic here is that sorting patterns by specificity, matching them with parse(), and interpolating config values — already exists as classmethods on CatalogConfigResolver: `_sort_patterns`, `_get_matches`, and `_resolve_dataset_config`. They're pure utilities (no instance needed, no credentials, no side effects) and would cover everything here without reimplementation.

The only blocker is that they're currently private. But we can discuss whether it's fine, or shall we expose them as soon as we come up with the concrete APIs needed.

Out of the head (haven't tested), we can at least do something like this:

```python
def _resolve_factory_patterns(
    catalog_config: dict[str, Any],
    datasets: dict[str, DatasetSnapshot],
    pipelines: list[PipelineSnapshot],
) -> dict[str, DatasetSnapshot]:
    """Enrich the datasets mapping with concrete entries derived from factory patterns.
    """
    pattern_configs = {
        name: config
        for name, config in catalog_config.items()
        if CatalogConfigResolver.is_pattern(name)
    }
    if not pattern_configs:
        return datasets

    sorted_patterns = CatalogConfigResolver._sort_patterns(pattern_configs)

    pipeline_ds_names: set[str] = set()
    for pipeline_snapshot in pipelines:
        for node_snapshot in pipeline_snapshot.nodes:
            pipeline_ds_names.update(node_snapshot.inputs)
            pipeline_ds_names.update(node_snapshot.outputs)

    result = dict(datasets)

    for ds_name in pipeline_ds_names:
        if ds_name in result:
            continue

        matched = next(
            CatalogConfigResolver._get_matches(sorted_patterns, ds_name), None
        )
        if matched is None:
            continue

        resolved_config = CatalogConfigResolver._resolve_dataset_config(
            ds_name, matched, copy.deepcopy(sorted_patterns[matched])
        )

        result[ds_name] = DatasetSnapshot(
            name=ds_name,
            type=resolved_config.get("type", ""),
            filepath=resolved_config.get("filepath"),
        )

    return result
```

kedro/inspection/helper.py
```diff
@@ -0,0 +1,117 @@
...
+
+    Returns:
+        Sorted list of top-level parameter key strings.
+        Returns an empty list when no parameters configuration is found.
+    """
+    try:
+        params: dict = config_loader["parameters"]
+    except (KeyError, MissingConfigException):
+        return []
+    return sorted(params.keys())
```
SajidAlamQB, line 57: In the issue you mention that the nested keys are recursively flattened and sorted (e.g. params:model_options.test_size) but the code only returns top-level keys is that intentional?

tests/inspection/test_project_snapshot.py
```diff
@@ -0,0 +1,246 @@
...
+        result = _build_project_snapshot(self.project_path)
+        assert result.pipelines is self.pipeline_snapshots
+
+    def test_datasets_populated(self):
+        result = _build_project_snapshot(self.project_path)
+        assert result.datasets is self.dataset_snapshots
+
+    def test_parameters_populated(self):
+        result = _build_project_snapshot(self.project_path)
+        assert result.parameters == ["model_options", "test_size"]
```
SajidAlamQB, line 176: Maybe a test that calls `_build_project_snapshot(path)` without env and asserts that `_make_config_loader` is called with `env=None`?

kedro/inspection/helper.py
ravi-kumar-pilla: Good catch. I was in two minds -

1. Recursive flattening as initially thought but while implementing, I wanted to keep it simple as nested levels can go deep, so not sure how well to represent only keys
2. The current approach keeps it simple and we do have deep nesting that is already captured implicitly via NodeSnapshot.inputs for params that are actually used

But I would love to hear other thoughts on this.

@rashidakanchwala @ElenaKhaustova

rashidakanchwala: I checked how kedro-viz handles this -- it doesn't recursively flatten either, just top-level keys. I say let's go with option 2. We can revisit later if needed.

tests/inspection/test_helper.py
```diff
@@ -0,0 +1,315 @@
...
+        pipelines = [
+            _make_pipeline_snapshot(
+                [_make_node_snapshot(["raw"], ["data_science.processed"])]
+            )
+        ]
+        result = _resolve_factory_patterns(catalog_config, datasets, pipelines)
+        assert "data_science.processed" in result
+        assert result["data_science.processed"].filepath == "data/processed.csv"
+
+    def test_original_pattern_entry_preserved(self):
```
rashidakanchwala: Ravi, why are we preserving the factory pattern ?

kedro/inspection/helper.py
```diff
@@ -0,0 +1,117 @@
...
+        )
+        if matched is None:
+            continue
+
+        resolved_config = CatalogConfigResolver._resolve_dataset_config(
+            ds_name, matched, copy.deepcopy(sorted_patterns[matched])
+        )
+
+        result[ds_name] = DatasetSnapshot(
+            name=ds_name,
```
rashidakanchwala: nit : this is repeated twice, maybe we create a classmethod](nit: the DatasetSnapshot(name=..., type=config.get("type", ""), filepath=config.get("filepath")) construction is repeated in both snapshot.py and helper.py. Maybe add a classmethod on DatasetSnapshot to keep it DRY.)

kedro/inspection/helper.py
ElenaKhaustova: I agree with option 2 — keep top-level and revisit later if needed.

tests/inspection/test_helper.py
ravi-kumar-pilla: Hi @rashidakanchwala ,

Good one. Similar to the parameter keys, I was not sure if we should remove the factory pattern completely. 

**Keeping the factory patterns -**
1. This keeps `_resolve_factory_patterns` add-only which is simpler
2. It preserves the factory patterns mentioned in the catalog even when there is no dataset resolution giving users the impression of factory pattern usage
3. Since factory patterns are resolved at runtime, this is a catalog snapshot (which covers patterns as-is), removing unresolved patterns might not give full picture of the datasets.

**Remove all patterns -**
1. If the `datasets` entry i.e., DatasetSnapshot is only viewed as the ones used in pipelines, removing factory pattern make sense
2. Since factory patterns are templates and they are not usable datasets, having both patterns and resolved datasets might be misleading.
3. `_resolve_factory_patterns` will not be just add-only. It will now remove all patterns irrespective of their resolution

Let me know your thoughts @rashidakanchwala @ElenaKhaustova @SajidAlamQB

ravi-kumar-pilla: Based on our conversation @rashidakanchwala , I will remove all the patterns from snapshot. We can add patterns back if need be in future. Thank you

PR #5459, opened 2026-03-19
Add public API for inspection layer

### LINKED ISSUE #5432: Public Python API - `get_project_snapshot()` and integration tests
### Description

Expose the public-facing Python API: `get_project_snapshot()` in `kedro/inspection/__init__.py` with `__all__` exports and optional `pipeline_id` filtering. Validate the full end-to-end flow with integration tests against the  starter project.

### Context

This ticket delivers the API surface that plugin authors and platform teams call directly. 

### Scope

**`kedro/inspection/__init__.py`**:
```python
def get_project_snapshot(
    project_path: str | Path,
    env: str | None = None,
    pipeline_id: str | None = None,
) -> ProjectSnapshot:
    """Primary entry point. Returns a full ProjectSnapshot of the project.
    If pipeline_id is set, returns a snapshot scoped to that pipeline only.
    """

def _filter_pipeline(snapshot: ProjectSnapshot, pipeline_id: str) -> ProjectSnapshot:
    """Return a new snapshot containing only the requested pipeline and its datasets."""

__all__ = [
    "get_project_snapshot",
    "ProjectSnapshot",
    "PipelineSnapshot",
    "NodeSnapshot",
    "DatasetSnapshot",
    "ProjectMetadataSnapshot",
]
```

**`tests/inspection/test_integration.py`** (against spaceflights-pandas starter):
- `get_project_snapshot()` returns a `ProjectSnapshot`
- `snapshot.metadata.project_name` and `package_name` are correct
- At least one pipeline is present with nodes
- `snapshot.datasets` contains expected catalog entries with correct `type` and `filepath`
- `snapshot.parameter_keys` contains expected keys without values
- `pipeline_id` filtering returns only the requested pipeline
- Datasets in the filtered snapshot only include those referenced by the requested pipeline's nodes
- `json.dumps(dataclasses.asdict(snapshot))` produces valid JSON

### Deliverable

- `kedro/inspection/__init__.py` (final public API)
- `tests/inspection/test_integration.py`

ravi-kumar-pilla:
## Description

Resolves #5432 

## Development notes

- Add `get_project_snapshot` to `__init__.py`
- Add public api tests
- Add e2e tests

## Checklist

### DIFF
--- a/features/inspection.feature
+++ b/features/inspection.feature
@@ -4,23 +4,8 @@ Feature: Project Inspection API
     Given I have prepared a config file
     And I have run a non-interactive kedro new with starter "default"
 
-  Scenario: get_project_snapshot returns a valid snapshot for a default project
-    When I call get_project_snapshot on the project
-    Then the snapshot is a valid ProjectSnapshot
-
-  Scenario: The snapshot captures correct project metadata
-    When I call get_project_snapshot on the project
-    Then the snapshot metadata project_name is "project-dummy"
-    And the snapshot metadata package_name is "project_dummy"
-
-  Scenario: The snapshot captures registered pipelines
+  Scenario: A developer inspects a project to understand its structure without running it
     When I call get_project_snapshot on the project
     Then the snapshot pipelines include "__default__", "data_engineering", "data_science", "data_processing"
-
-  Scenario: The snapshot captures catalog datasets
-    When I call get_project_snapshot on the project
-    Then the snapshot datasets include "example_iris_data", "example_model", "example_predictions"
-
-  Scenario: The snapshot captures parameter keys
-    When I call get_project_snapshot on the project
-    Then the snapshot parameters include "example_learning_rate", "example_num_train_iter", "example_test_data_ratio"
+    And the snapshot datasets include "example_iris_data", "example_model", "example_predictions"
+    And the snapshot parameters include "example_learning_rate", "example_num_train_iter", "example_test_data_ratio"
--- a/features/steps/cli_steps.py
+++ b/features/steps/cli_steps.py
@@ -22,6 +22,7 @@ else:
 import kedro
 from features.steps import util
 from features.steps.sh_run import ChildTerminatingPopen, check_run, run
+from kedro.inspection import get_project_snapshot
 
 OK_EXIT_CODE = 0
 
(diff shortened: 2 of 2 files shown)


kedro/inspection/__init__.py
```diff
@@ -1 +1,38 @@
...
+    Args:
+        project_path: Path to the project root directory (the directory that
+            contains ``pyproject.toml``).
+        env: Optional run environment override (e.g. ``"staging"``). When
+            ``None``, the project's default run environment is used.
+
+    Returns:
+        A fully populated ``ProjectSnapshot``.
+    """
+    return _build_project_snapshot(project_path, env=env)
```
merelcht: Let's be consistent in using named parameters or not.
```suggestion
    return _build_project_snapshot(project_path=project_path, env=env)
```

tests/inspection/test_integration.py
```diff
@@ -0,0 +1,31 @@
...
+
+from kedro.inspection import get_project_snapshot
+from kedro.inspection.models import ProjectSnapshot
+
+
+class TestGetProjectSnapshot:
+    def test_get_project_snapshot_is_exported(self):
+        from kedro.inspection import __all__
+
+        assert "get_project_snapshot" in __all__
```
merelcht: This test seems overkill, it's basically checking the python import system.

ravi-kumar-pilla:
Hi @rashidakanchwala , @merelcht 

Thank you for reviewing. I addressed your comments. Please review and let me know if there are any issues.

features/steps/cli_steps.py
```diff
@@ -775,3 +775,46 @@ def delete_project_file(context, filepath):
         file_to_delete.unlink()
     elif file_to_delete.is_dir():
         shutil.rmtree(file_to_delete)
+
+
+@when("I call get_project_snapshot on the project")
+def call_get_project_snapshot(context):
+    """Call get_project_snapshot on the test project and store the result in context."""
+    from kedro.inspection import get_project_snapshot
```
merelcht: Why the imports inside the method and not just at the top of the file?

features/inspection.feature
```diff
@@ -0,0 +1,26 @@
+Feature: Project Inspection API
```
merelcht, line 1: This looks like a lot of e2e tests to basically only check that the snapshot is correct. The behave tests are quite slow, so I would prefer if we can optimise the testing here. We already have unit tests for granular details. E2e tests are meant to follow BDD (https://en.wikipedia.org/wiki/Behavior-driven_development) so the scenarios should be based on what a user is trying to achieve.

merelcht (changes requested):
Can you make these e2e tests closer to BDD please. Now they're basically unit tests written in behave.

features/inspection.feature
ravi-kumar-pilla: I too thought of this. Does combining all the scenarios make sense here like below ? @merelcht 

Scenario: A developer inspects a project to understand its structure without running it
    When I call get_project_snapshot on the project
    Then the snapshot pipelines include "__default__", "data_engineering", "data_science", "data_processing"
    And the snapshot datasets include "example_iris_data", "example_model", "example_predictions"
    And the snapshot parameters include "example_learning_rate", "example_num_train_iter", "example_test_data_ratio"


Thank you

features/steps/cli_steps.py
ravi-kumar-pilla: No specific reason. I kept it locally as it is only required for the cli step. Let me move it to module level

features/inspection.feature
merelcht: Yes, here it's less about testing all the details separately (which should be covered in unit tests) and more about as a user I need to achieve ... so when I request a snapshot I expect ..., so combining it makes sense to me.

PR #5479, opened 2026-04-01
Fix: AttributeError on UnionType in parameter validation type extractor

SajidAlamQB:
## Description

Reported by a [user on Slack](https://kedro-org.slack.com/archives/C03RKP2LW64/p1775037481550659) after the 1.3.0 release any node with a union or builtin type hint on a `params: input` would crash during type extraction.

Once this is in we can do a quick patch release 1.3.1.

## Development notes

* Fixed `AttributeError: 'types.UnionType' object has no attribute '__name__'` when node functions have non-Pydantic/dataclass type hints (e.g. `list[str] | None`, `int`, `dict`) on `params:` inputs
* The type extractor now skips types it cannot validate, filtering early with `is_pydantic_class()` and `dataclasses.is_dataclass()` checks
* Added `getattr` safety net in the conflicting types warning for edge cases

## Checklist

### DIFF
--- a/kedro/validation/type_extractor.py
+++ b/kedro/validation/type_extractor.py
@@ -141,7 +141,8 @@ class TypeExtractor:
                 ) and expected_type not in self._warned_union_types:
                     self._warned_union_types.add(expected_type)
                     type_names = " | ".join(
-                        getattr(a, "__name__", str(a)) for a in get_args(expected_type)
+                        getattr(type_arg, "__name__", str(type_arg))
+                        for type_arg in get_args(expected_type)
                     )
                     logger.warning(
                         "Union type hint (%s) is not supported for validation. "


kedro/validation/type_extractor.py
```diff
@@ -52,10 +55,10 @@ def extract_types_from_pipelines(self) -> dict[str, type]:
                             "%s (existing) vs %s (from pipeline '%s'). "
                             "Using %s.",
                             key,
-                            existing_type.__name__,
-                            new_type.__name__,
+                            getattr(existing_type, "__name__", str(existing_type)),
```
ElenaKhaustova, line 73: Shall we do the same for `logger.debug()` below?

ElenaKhaustova, line 73: There's no test exercising the getattr fallback with a type that actually lacks `__name__`. Shall we add it, since it's part of the fix?

kedro/validation/type_extractor.py
```diff
@@ -109,6 +112,13 @@ def _extract_types_from_node(self, node: Node) -> dict[str, type]:
             if not ds_name.startswith(_PARAMS_PREFIX):
                 continue
 
+            # Only record types we can actually validate (Pydantic models or dataclasses)
```
ElenaKhaustova, line 133: One thing to consider: get_type_hints resolves `Optional[MyModel]` into `MyModel | None`, which is a `UnionType` — not a class. So when the new guard checks `is_pydantic_class` and `is_dataclass`, both return False, and the parameter is silently skipped. Before this PR it would crash, now it's ignored — strictly better, but the user probably still expects validation to happen since the inner type is a valid Pydantic model. Might be worth unwrapping `X | None` and validate against `X`.

SajidAlamQB:
> Also, shall we do a patch release 1.3.1 after the fix? cc @merelcht

Hey Elena thanks for the quick review, I've also spoken with Merel she has approved it as patch release.

kedro/validation/type_extractor.py
SajidAlamQB: Really good point on `Optional[MyModel]`, I will add unwrapping for `X | None` unions where `X` is a single Pydantic model or dataclass. Multi-type unions are still skipped.

SajidAlamQB:
@ElenaKhaustova addressed your comments:

- Applied getattr fallback to `logger.debug` too
- Added `test_conflicting_types_getattr_fallback` using a `UnionType (no __name__)`
- Added `_unwrap_optional() `: `Optional[MyModel] / MyModel | None` now unwraps and validates against the inner type, with tests.

kedro/validation/type_extractor.py
ravi-kumar-pilla: I know this is a limitation but skipping multi-type unions cause some issues when it comes to skipping. May be we need to handle this via a log statement (occurs only once). In below scenario, IDE still supports `split_options.target` but since we skip it, its treated like dict which is a bit confusing

```py
def split_data(data: pd.DataFrame, split_options: SplitOptions | OtherOptions) -> Tuple:
    """Splits data into features and targets training and test sets.

    Args:
        data: Data containing features and target.
        parameters: Parameters defined in parameters.yml.
    Returns:
        Split data.
    """
    target_variable = split_options.target
    independent_variables = [x for x in data.columns if x != target_variable]
    test_size = split_options.test_size
    random_state = split_options.random_state
    logger.info(
        f"Splitting data for the following independent variables "
        f"{independent_variables} against the target of '{target_variable}' "
        f"with a test sized of {test_size} and a random state of "
        f"'{random_state}'"
    )

    X = data[independent_variables]
    y = data[target_variable]
    )

    return X_train, X_test, y_train, y_test

```

SajidAlamQB: Good catch @ravi-kumar-pilla I'll add a warning log when a multi-type union on a `params:` input is skipped.

kedro/validation/type_extractor.py
```diff
@@ -109,12 +127,35 @@ def _extract_types_from_node(self, node: Node) -> dict[str, type]:
...
+                    get_origin(expected_type) is Union
+                    or isinstance(expected_type, types.UnionType)
+                ) and param_key not in self._warned_union_keys:
+                    self._warned_union_keys.add(param_key)
+                    logger.warning(
+                        "Parameter '%s' has a union type hint (%s) which is not "
+                        "supported for validation. The parameter will be passed "
+                        "as a raw dictionary.",
+                        param_key,
+                        expected_type,
```
ravi-kumar-pilla: I am testing it with demo_project and I get multiple warnings -

@SajidAlamQB 

```bash
 WARNING  Parameter 'split_options_2' has a union type hint (demo_project.pipelines.modelling.nodes.SplitOptions | demo_project.pipelines.modelling.nodes.OtherOptions | None) which is not    type_extractor.py:144
                             supported for validation. The parameter will be passed as a raw dictionary.                                                                                                                               
                    WARNING  Parameter 'train_evaluation.linear_regression.split_options_2' has a union type hint (demo_project.pipelines.modelling.nodes.SplitOptions |                                          type_extractor.py:144
                             demo_project.pipelines.modelling.nodes.OtherOptions | None) which is not supported for validation. The parameter will be passed as a raw dictionary.                                                      
                    WARNING  Parameter 'train_evaluation.random_forest.split_options_2' has a union type hint (demo_project.pipelines.modelling.nodes.SplitOptions |                                              type_extractor.py:144
                             demo_project.pipelines.modelling.nodes.OtherOptions | None) which is not supported for validation. The parameter will be passed as a raw dictionary. 
```

I have in my modelling.yml -
```yaml
split_options_2:
  test_size: 0.2
  random_state: 3
  target: 'price'
```

And I used it in 2 places -
- modelling/nodes.py - train_model method
- modelling/nodes.py - split_data method

Also the type hint for example `demo_project.pipelines.modelling.nodes.SplitOptions |   demo_project.pipelines.modelling.nodes.OtherOptions | None` can become too big. wdyt ?

SajidAlamQB: Hey Ravi addressed in the latest push warnings now deduplicate with `_warned_union_types` set (fires once per union type, not per parameter key), and type names use `__name__` instead of full qualified paths tested it on my spaceflights project lmk if it all looks good on your end.

kedro/validation/type_extractor.py
```diff
@@ -135,18 +135,19 @@ def _extract_types_from_node(self, node: Node) -> dict[str, type]:
...
-                param_key = ds_name.split(":", 1)[1]
                 if (
                     get_origin(expected_type) is Union
                     or isinstance(expected_type, types.UnionType)
-                ) and param_key not in self._warned_union_keys:
-                    self._warned_union_keys.add(param_key)
+                ) and expected_type not in self._warned_union_types:
+                    self._warned_union_types.add(expected_type)
+                    type_names = " | ".join(
+                        getattr(a, "__name__", str(a)) for a in get_args(expected_type)
```
ravi-kumar-pilla: getattr(type_arg, "__name__", str(type_arg)) for type_arg in get_args(expected_type)

ravi-kumar-pilla (approved):
LGTM and works well. Thank you @SajidAlamQB

PR #5482, opened 2026-04-02
Handle varying session API for `kedro run` and `kedro ipython` 

### LINKED ISSUE #5481: Make `KedroSession` and `KedroServiceSession` interchangable with `SESSION_CLASS`
## Description
Due to some API differences, we'll have to adjust `kedro run` command and `kedro ipython` commands to have some conditional handling based on `SESSION_CLASS` defined in `settings.py`

ankatiyar:
## Description

This is based off of #5441

Fix #5481 

There is some differences in the API for `KedroSession` and `KedroServiceSession` so this needs to be handled when the CLI commands are using the session

## Development notes

- Before we were showing deprecation warning for `pipeline_names` within session but now I have added a deprecation warning when `--pipeline` is used and the CLI interface then sends it to session through `pipeline_names` (instead of `pipeline_name`) so that the warning is only shown once
- Adjusted session args into `create_kwargs` and `run_kwargs` where based on `settings.SESSION_CLASS` it will add `runtime_params` to the appropriate place
- Question for reviewers: What do we do in the longer term to make it so that the sessions are interchangable without special conditions. Should `create()` have a `runtime_params` also anyway? In CLI and ipython settings it makes sense that the `runtime_params` are injected at creation instead of per run.

## Checklist


rashidakanchwala:
Here are my thoughts on this 

`KedroServiceSession` will eventually replace `KedroSession`. In the long term, we will have only one session class, which removes the need for the current if/else logic.

Runtime parameters are called “runtime parameters” because they are specific to each run. For that reason, they should be injected at run time, not at session creation.

For IPython use cases:
- If users want the same parameters throughout the session, they should use parameters.yml.
- If users want different parameters for each run, they should pass them explicitly when running: `session.run(runtime_params=xyz)`.

kedro/framework/cli/project.py
```diff
@@ -263,28 +266,39 @@ def run(  # noqa: PLR0913
...
+        "from_nodes": from_nodes,
+        "to_nodes": to_nodes,
+        "from_inputs": from_inputs,
+        "to_outputs": to_outputs,
+        "load_versions": load_versions,
+        "pipeline_names": pipelines_to_run,
+        "namespaces": namespaces,
+        "only_missing_outputs": only_missing_outputs,
+    }
+    if settings.SESSION_CLASS == KedroSession:
```
rashidakanchwala, line 297: Can we add some code comments around this being temporary and the reason we have this.

tests/framework/cli/test_cli.py
```diff
@@ -1127,6 +1117,27 @@ def test_run_with_non_existent_conf_source(self, fake_project_cli, fake_metadata
         )
         assert expected_output in result.output
 
+    def test_run_with_service_session(self, fake_project_cli, fake_metadata, mocker):
```
rashidakanchwala, line 1120: For this test, can we also assert that `runtime_params` is called during run and not during create for KedroServiceSession

kedro/framework/cli/project.py
ElenaKhaustova, line 297: Should it be `issubclass` instead of `==`?

So this works too:
```python
class MySession(KedroSession):
    pass

SESSION_CLASS = MySession
```

kedro/ipython/__init__.py
```diff
@@ -125,14 +127,20 @@ def reload_kedro(
...
     configure_project(metadata.package_name)
-
-    session = settings.SESSION_CLASS.create(
-        project_path,
-        env=env,
-        runtime_params=runtime_params,
-        conf_source=conf_source,
-    )
-    context = session.load_context()
+    is_kedrosession = settings.SESSION_CLASS == KedroSession
```
ElenaKhaustova, line 130: Same here with `==`

kedro/framework/cli/project.py
```diff
@@ -263,28 +266,39 @@ def run(  # noqa: PLR0913
...
+        create_kwargs["runtime_params"] = params
```
ElenaKhaustova, line 298: Answering your question

> What do we do in the longer term to make it so that the sessions are interchangable without special conditions. Should create() have a runtime_params also anyway? 

I think we should make `KedroServiceSession.create()` also accept `runtime_params` as a default for runs to get rid of `if/else` entirely. And `KedroServiceSession` still supports per-run overrides.

kedro/framework/cli/project.py
ElenaKhaustova: If we keep branching

kedro/framework/cli/project.py
rashidakanchwala: I think we should be careful about this. When we were designing `KedroServiceSession`, we were trying to establish what belongs at session scope (expensive things that need to be loaded once, such as heavy artifacts and models) and what belongs at run scope (`runtime_data` and `runtime_params`).

If we allow `runtime_params` to be defined at session level and overridden at run level, it becomes convulated. Consider this scenario: a user has parameters defined in `parameters.yml` and starts a long-running session. They define some `runtime_params` at session level and forget about it. For some runs they pass `runtime_params` explicitly; for others they don't, assuming the run will fall back to `parameters.yml`. But it doesn't  and their results are inconsistent because of the session-level `runtime_params` they'd forgotten they set.

ankatiyar:
@rashidakanchwala @ElenaKhaustova 
I'm also leaning towards adding `runtime_params` in both `create()` and `run()`. This way we maintain compatibility with both the sessions, making them easily replaceable using `SESSION_CLASS` for all purposes and with no conditions (cleaner code for `ipython` and `kedro run` cases). 

From usability perspective, I think it'll add flexibility. Right now all the resolution is happening in `run()` for catalog and context but when we add pre-loading of things, it'll help to pass some `runtime_params` during creation of the session so that the catalog can be resolved (when `runtime_params:` resolver is used in the catalog) but also allow for injecting params to override them when they are used as parameters. It seems unlikely in the programmatic usage of `KedroServiceSession` that someone starts a session and injects runtime_parameters but forgets what they've set. 🤔

DimedS:
Thanks all -  great discussion. I think the key is to separate **API compatibility** from **semantic ownership**.

My preference would be:

### Long Term
- Keep `runtime_params` conceptually **run-scoped**, not session-scoped.
- This preserves a clear mental model and avoids hidden or “sticky” configuration that could lead to inconsistent results.

### Short Term
- Retain the current conditional handling to support both `KedroSession` and `KedroServiceSession`.
- Clearly document in the code that this is a **temporary solution**.
- Use `issubclass()` instead of `==` to support custom session implementations.

tests/framework/cli/test_cli.py
```diff
@@ -1127,6 +1117,26 @@ def test_run_with_non_existent_conf_source(self, fake_project_cli, fake_metadata
...
+        )
+        result = CliRunner().invoke(
+            fake_project_cli,
+            ["run", "--params", "foo=bar"],
+            obj=fake_metadata,
+        )
+        assert result.exit_code == 0
+        mock_create.assert_called_once()
+        call_kwargs = mock_create.call_args[1]
+        print(call_kwargs)
```
ElenaKhaustova, line 1137: Shall we remove this and add the missing assertion that `runtime_params` is passed to `run()`?

ElenaKhaustova (approved):
👍

ElenaKhaustova:
> What do we do in the longer term to make it so that the sessions are interchangable without special conditions. Should create() have a runtime_params also anyway?

My preference is still to avoid the branching and have both sessions accept runtime_params on create(). A few reasons:

1. There's no clear timeline for removing `KedroSession`, and the new session doesn't yet cover all cases the old one does (e.g., parallel execution). Both will coexist for a while.
2. Every new entry point (API layer, future CLI commands, plugins) will need the same issubclass check. It's two places today but will keep growing.
3. Default runtime params with per-run overrides is a standard pattern. It keeps the CLI and IPython code clean with no branching needed.

I understand the argument for keeping runtime params strictly run-scoped, and I won't block on this. But I wanted to put this perspective on record since it affects how much conditional logic we carry going forward.

rashidakanchwala (approved):
Approving this for now but let's have a proper PS discussion on pros/cons as both arguments are valid.

ankatiyar:
I'll merge this for now and discuss this when we also talk about pre-loading catalog where having `runtime_params` in `create()` also will be quite useful!

Issue #5443, opened 2026-03-10
Scope `type_extractor` to target pipeline instead of all registered pipelines for Validation


SajidAlamQB:
Currently `TypeExtractor.extract_types_from_pipelines` walks all registered pipelines regardless of which one is being run. This means if a user runs `kedro run --pipeline=data_science`, type requirements from other pipelines (e.g. data_engineering) are still extracted and validated.

Nothing breaks but it's unnecessary work and could cause confusing warnings about conflicting types from pipelines that aren't being executed.

Proposed change: Accept the target pipeline name (or pipeline object) and only extract types from that pipeline.

Related comment: https://github.com/kedro-org/kedro/pull/5421#discussion_r2902936537

PR #5494, opened 2026-04-13
Target pipeline for `type_extractor`

SajidAlamQB:
## Description

Related to: https://github.com/kedro-org/kedro/issues/5443

## Checklist


kedro/framework/context/context.py
```diff
@@ -178,7 +178,7 @@ class KedroContext:
     _runtime_params: dict[str, Any] | None = field(
         init=True, default=None, converter=deepcopy
     )
-    _validated_params_cache: dict[str, Any] | None = None
+    _validated_params_cache: dict[str | None, dict[str, Any]] | None = None
```
ravi-kumar-pilla, line 181: can the cache key be None ?

kedro/framework/context/context.py
```diff
@@ -213,14 +220,16 @@ def _get_validated_params(self) -> dict[str, Any]:
 
             pipeline_dict = dict(project_pipelines)
             validator = ParameterValidator(pipeline_dict)
```
ravi-kumar-pilla, line 222: I think it would be better to pass a single pipeline here (at line 222) to the validator when pipeline name is provided instead of the entire dictionary. This will remove passing pipeline_name in following functions like validator.validate_raw_params etc.

Also having pipeline_name as None is a bit odd. We need to think an alternate way but having something default might interfere with any user defined pipeline name. 

> Currently TypeExtractor.extract_types_from_pipelines walks all registered pipelines regardless of which one is being run. This means if a user runs kedro run --pipeline=data_science, type requirements from other pipelines (e.g. data_engineering) are still extracted and validated.

How does the scoped validation works when lets say there are pydantic models in data engineering pipeline but someone runs kedro run --pipeline=data_science ? Does this cause any error for data engineering pipeline as the models are not yet instantiated due to the scoped instantiation we do here ?

kedro/validation/type_extractor.py
```diff
@@ -44,21 +44,42 @@ def __init__(self, pipelines: dict[str, Pipeline]) -> None:
...
+                registered pipelines.
         """
-        all_type_requirements: dict[str, type] = {}
+        if pipeline_name is not None:
+            if pipeline_name not in self._pipelines:
+                raise KeyError(
+                    f"Pipeline '{pipeline_name}' not found. "
+                    f"Available pipelines: {sorted(self._pipelines.keys())}"
+                )
+            target = {pipeline_name: self._pipelines[pipeline_name]}
```
rashidakanchwala:
```suggestion
            pipelines_to_inspect = {pipeline_name: self._pipelines[pipeline_name]}
```

kedro/framework/context/context.py
```diff
@@ -286,10 +304,19 @@ def _get_catalog(
...
+
+        Args:
+            pipeline_name: Optional pipeline name to scope parameter
+                validation to. When ``None``, uses the ``params`` property
+                (validates all pipelines).
+        """
+        if pipeline_name is not None:
+            params = self._get_validated_params(pipeline_name=pipeline_name)
+        else:
+            params = self.params
```
rashidakanchwala:
```suggestion
      params = self._get_validated_params(pipeline_name=pipeline_name)
      params_dict: dict[str, Any] = {"parameters": params}
```

kedro/framework/session/session.py
```diff
@@ -407,10 +407,18 @@ def run(  # noqa: PLR0913
             else settings.DATA_CATALOG_CLASS
         )
 
+        # Scope parameter validation to the target pipeline when a single
+        # named pipeline is being run.  For the default pipeline or multiple
+        # pipelines, fall back to validating all pipelines.
+        validation_pipeline_name = (
+            names[0] if len(names) == 1 and names[0] != "__default__" else None
```
rashidakanchwala: nit - i was a bit confused where names came from; then I went and read up the codebase and read the logic. I was wondering if we can make this more clear call it `pipeline_names_to_run`  above or similar. This is ofcourse related to old code

kedro/framework/context/context.py
SajidAlamQB: Yes it can be None

kedro/framework/context/context.py
SajidAlamQB: This makes a lot of sense I've added this now the `pipeline_name` parameter is gone from both `ParameterValidator.validate_raw_params` and `TypeExtractor.extract_types_from_pipelines`. The filtering now happens once in `KedroContext._get_validated_params`.

SajidAlamQB: If a user later runs `kedro run --pipeline=data_engineering`, that's a fresh `KedroSession` with its own catalog, so the data_engineering Pydantic models get instantiated then.

kedro/framework/session/session.py
SajidAlamQB: Good catch but I think outside this PR's scope I'd rather keep this PR focused on the type-extractor scoping and address the rename in a separate follow-up so the diff stays clear. Happy to open a tiny follow up PR for it after this one merges.

ravi-kumar-pilla:
Review generated with the `review-kedro-pr` skill

---

## Overview

| Severity    | Count |
| ----------- | ----- |
| Critical    | 0     |
| Suggestions | 3     |

---

## Critical (must fix before merge)

No issues found.

---

## Suggestions (consider improving)

### 1. `kedro/framework/session/service_session.py` (~L206–L210)

**`KedroServiceSession.run` does not pass `pipeline_name` into `_get_catalog`.**

`KedroSession.run` scopes parameter / type-hint validation when a single non-default pipeline is run. `KedroServiceSession.run` uses the same `pipeline_names` pattern (`pipeline_names = pipeline_names or ["__default__"]`, then combines pipelines) but still calls `context._get_catalog(...)` without `pipeline_name`, so validation stays unscoped (all pipelines). Users of the service session can still see the same cross-pipeline noise this PR removes for `KedroSession`.

**Suggestion:** Apply the same `validation_pipeline_name` logic as in `kedro/framework/session/session.py` and pass `pipeline_name=validation_pipeline_name` into `_get_catalog` for parity.

---

### 2. GitHub PR description (checklist)

The PR template **checklist still shows unchecked** items for documentation, `RELEASE.md`, and tests, while the branch **does** include a `RELEASE.md` entry and new/updated tests.

**Suggestion:** Tick the relevant boxes and add a short “how this was tested” note under Development notes so reviewers see compliance at a glance.

---

### 3. `kedro/framework/context/context.py` — `params` property docstring

**`KedroContext.params` still calls `_get_validated_params()` with no pipeline**, so it reflects validation across registered pipelines in the unscoped sense. The catalog built during `kedro run --pipeline=…` uses scoped validation via `_get_catalog(..., pipeline_name=…)`.

**Suggestion:** Add a brief docstring line clarifying that `params` is not automatically tied to the current run’s target pipeline; scoping applies when the session builds the catalog with a specific pipeline. Reduces confusion for anyone comparing `context.params` to catalog parameters.

---

## Notes

### General review

- Scoping is pushed to callers via the pipeline mapping; the validation cache is keyed by `str | None`.
- For `pipeline_name is None`, `__default__` is omitted from the dict passed to the validator because other registered pipelines already cover its nodes.
- `KedroSession` threads `validation_pipeline_name` into `_get_catalog` / `_get_parameters`.
- Tests cover cache keys, unknown pipeline `ValueError`, and scoped `ParameterValidator` / `TypeExtractor` behavior.
- `.secrets.baseline` changes look like normal detect-secrets baseline maintenance.

### PR checklist (changelog / docs)

- `RELEASE.md` is updated under “Bug fixes and other changes”; no `docs/` change in the diff — acceptable if the RELEASE bullet is enough; optional follow-up is extra user-facing documentation.

### API / design

- Optional `pipeline_name` on `_get_catalog` keeps backward compatibility for overrides and hooks.
- `TypeExtractor` docstrings match the “caller filters pipelines” contract.

### Existing GitHub threads (do not re-open unless you disagree)

- **`None` as cache key:** Confirmed valid in discussion.
- **Rename `names` → e.g. `pipeline_names_to_run`:** Author deferred to a small follow-up PR; reasonable to keep this PR focused.

---

## Verdict

No blocking issues from this review. The highest-value follow-up is **mirroring scoped `_get_catalog` in `KedroServiceSession`** so both session entry points behave the same for single-pipeline runs.

RELEASE.md
```diff
@@ -10,6 +10,7 @@ NOTE: This session implementation is under active development and may occasional
 * Improved `%load_node` to include same-module helper dependencies via AST extraction, with explicit fallback warnings when extraction degrades to function-only source loading.
 
 ## Bug fixes and other changes
+* Scoped parameter type-hint validation to the pipeline being run. When `kedro run --pipeline=<name>` runs a single named pipeline, only that pipeline's node signatures are inspected for type hints, avoiding unnecessary work and spurious "conflicting type" warnings from pipelines that aren't executing.
```
ravi-kumar-pilla: May be we can move this to improvements section above ?

kedro/framework/session/session.py
ravi-kumar-pilla: Will it be an issue to validate multiple pipelines ? Why are we validating all the pipelines if lets say 2 pipelines are sent ? Though the ticket says target pipeline I feel we should validate params only for the requested pipelines everytime and validate all pipeline params for default pipeline. wdyt @rashidakanchwala

kedro/framework/session/session.py
```diff
@@ -407,10 +407,18 @@ def run(  # noqa: PLR0913
...
+        )
+
         catalog = context._get_catalog(
             catalog_class=catalog_class,
             save_version=save_version,
             load_versions=load_versions,
+            pipeline_name=validation_pipeline_name,
```
ravi-kumar-pilla: I did not think this through but pipeline_name sitting within `_get_catalog` seems out of place to me

ravi-kumar-pilla: I would rather prefer having an internal prop like `_pipelines_to_validate` in context, which we can set, rather than polluting `_get_catalog`. I feel linking CLI click args to internals is getting complex. 

I will like to get others opinion on this @rashidakanchwala @ankatiyar

ravi-kumar-pilla:
Hi @SajidAlamQB ,

Initially used - `review-kedro-pr` and there are only few suggestions but I found we are modifying few unrelated functions for validation. I suggested a different approach above which seems clean to me. Please take a look. We can connect tomorrow if something is not clear. Apologies for the delayed PR review. Thank you

SajidAlamQB:
Reworked based on @ravi-kumar-pilla's `_pipelines_to_validate` suggestion. Cache is back to a single slot, `pipeline_name` no longer goes through the validator chain, and `KedroServiceSession.run` gets parity for free. Ready for re-review.

kedro/framework/context/context.py
```diff
@@ -211,8 +222,31 @@ def _get_validated_params(self) -> dict[str, Any]:
...
+                    for name, pipe in project_pipelines.items()
+                    if name != "__default__"
+                }
+            else:
+                unknown = [
+                    name
+                    for name in self._pipelines_to_validate
+                    if name not in project_pipelines
+                ]
+                if unknown:
```
ravi-kumar-pilla: This should be a check for session.run and not the concern of param validation. If the pipeline passed via -p is not available, the availability check should be made in the code related to session.run. This is a good to have guard here but I feel this check is not directly related to validation of params

kedro/framework/session/session.py
ravi-kumar-pilla: cc: @SajidAlamQB

ravi-kumar-pilla:
Hi @SajidAlamQB ,

The code looks great. Thank you for addressing all the review comments. Few **Nit** comments before I test e2e tomorrow -
1. The docstrings or comments in the code are good but if possible can you please make them compact ?

2. **review-pr skill suggestion:**
`kedro/framework/context/context.py` — params property docstring (~L254–262)
`context.params` still calls `_get_validated_params()` without the session having set `_pipelines_to_validate`, so it validates all registered pipelines (except `__default__`). That is fine for hooks/notebooks, but it differs from catalog parameters built during session.run().

**Suggestion:** Add one line to the params docstring noting that scoping applies when the session sets `_pipelines_to_validate` before building the catalog, not when params is read in isolation.

Thank you

rashidakanchwala (commented):
## Kedro Security Scan
> Generated with `security-scan`.

### Overview
- **Mode:** PR
- **Scope:** Kedro framework
- **Target:** PR #5494 (`feat/type_extractor-target-pipeline`)
- **Files scanned:** `.secrets.baseline`, `RELEASE.md`, `kedro/framework/context/context.py`, `kedro/framework/session/service_session.py`, `kedro/framework/session/session.py`, `kedro/validation/type_extractor.py`
- **Rulesets:** `p/security-audit`, `p/secrets`, `p/python`, local `kedro-security-patterns.yml`
- **Findings reviewed:** 2 (both WARNING; both on lines NOT changed by this PR)
- **Highest Semgrep severities:** WARNING

### Classification summary
- **candidate_kedro_vulnerability:** 0
- **project_developer_responsibility:** 0
- **deployment_or_environment_issue:** 0
- **false_positive_or_informational:** 2
- **needs_manual_review:** 0

### Findings
- `kedro/framework/context/context.py:53` — `false_positive_or_informational` — `kedro-taint-string-param-to-path` (WARNING) — The flagged line is inside `_is_relative_path()`, which IS the sanitizer (it returns `False` for absolute paths via `PurePosixPath.is_absolute()`); also pre-existing code, not touched by this PR. — No action; suppress if it recurs.
- `kedro/framework/context/context.py:355` — `false_positive_or_informational` — `python-logger-credential-disclosure` (WARNING) — A `debug` log emitted only when the credentials config file is **missing**; the `%s` interpolates the `MissingConfigException` message (also a "not found" string), no actual credential value is logged; pre-existing code, not touched by this PR. — No action; suppress if it recurs.

### Manual review checks (per `reference.md` §7)
- **New third-party config-dict consumers:** none introduced; new `_pipelines_to_validate: list[str]` only feeds an internal Kedro `ParameterValidator` and a dict-key lookup. Clean.
- **New env-var → execution paths:** no new `os.environ`/`os.getenv` calls in the diff. Clean.
- **Path construction from external strings:** no new path construction in the diff; new CLI pipeline-name strings are only used as keys against a developer-authored pipeline registry. Clean.
- **`subprocess` / `eval` / `exec` / `pickle.loads` / `yaml.load` / dynamic import:** no new occurrences in the diff. Clean.

### Conclusion
No plausible Kedro vulnerabilities found. Both Semgrep findings are on lines unchanged by this PR and are clearly safe in context. PR #5494 introduces no new security-sensitive code paths.

rashidakanchwala (commented):
## Kedro Security Scan
> Generated with `security-scan`.

### Overview
- **Mode:** PR
- **Scope:** Kedro framework
- **Target:** PR #5494 (`feat/type_extractor-target-pipeline`)
- **Files scanned:** `.secrets.baseline`, `RELEASE.md`, `kedro/framework/context/context.py`, `kedro/framework/session/service_session.py`, `kedro/framework/session/session.py`, `kedro/validation/type_extractor.py` (`.agents/`, `tests/`, `docs/`, `features/` excluded)
- **Findings reviewed:** 2
- **Highest Semgrep severities:** WARNING

### Classification summary
- **candidate_kedro_vulnerability:** 0
- **needs_manual_review:** 0
- **project_developer_responsibility:** 0
- **deployment_or_environment_issue:** 0
- **false_positive_or_informational:** 2

### Findings
None actionable. 2 false positives on pre-existing lines.

### Manual review checks
All clean (config-dict consumers, env-var paths, path construction, subprocess/eval/exec/pickle/yaml/dynamic import).

### Conclusion
- No plausible Kedro vulnerabilities found in this PR.

ravi-kumar-pilla (approved):
Works well. Looks like CI is failing, otherwise the functionality works fine 👍

kedro/framework/session/session.py
```diff
@@ -407,6 +407,11 @@ def run(  # noqa: PLR0913
             else settings.DATA_CATALOG_CLASS
         )
 
+        # Scope validation to the requested pipelines; `__default__` falls
+        # back to validating every registered pipeline.
+        if "__default__" not in names:
```
ankatiyar, line 412: Minor: this is flagged by the agent review

If anyone reads `context.params` using the `after_context_created` hook, the `context._validate_params_cache` gets warmed up with params from all pipelines and doesn't scope the params validation just to the pipeline being run. This might be okay, just flagging.

RELEASE.md
```diff
@@ -8,8 +8,10 @@ NOTE: This session implementation is under active development and may occasional
 * Added a new CLI command `kedro server start` to run the server.
 * Added inspection API to get project snapshot.
 * Improved `%load_node` to include same-module helper dependencies via AST extraction, with explicit fallback warnings when extraction degrades to function-only source loading.
+* Scoped parameter type-hint validation to the pipeline being run. When `kedro run --pipeline=<name>` runs a single named pipeline, only that pipeline's node signatures are inspected for type hints, avoiding unnecessary work and spurious "conflicting type" warnings from pipelines that aren't executing.
```
ankatiyar: I think we should mention `--pipelines` instead of `--pipeline` as that is meant to be deprecated

kedro/framework/context/context.py
```diff
@@ -211,8 +217,20 @@ def _get_validated_params(self) -> dict[str, Any]:
             from kedro.framework.project import pipelines as project_pipelines
             from kedro.validation.parameter_validator import ParameterValidator
 
-            pipeline_dict = dict(project_pipelines)
-            validator = ParameterValidator(pipeline_dict)
+            if self._pipelines_to_validate is None:
+                # Skip `__default__` as its nodes are already covered by the others.
```
ankatiyar: This is not strictly true because Kedro by default creates the `__default__` pipeline as a sum of all pipelines but users can set a different default using `pipeline_registry`

ankatiyar (commented):
I'm not super familiar with the validation workflow but I'm wondering if it makes sense to send a dictionary of pipelines to the `context._pipelines_to_validate` instead of the pipeline names and importing `pipelines` there. We could directly send `filtered_pipelines` to context instead of loading them from names again?

kedro/framework/context/context.py
SajidAlamQB: Good catch, dropped the skip entirely.

kedro/framework/session/session.py
SajidAlamQB: Fixed by tracking the cached scope alongside the cache. If a hook reads `context.params` before the session sets `_pipelines_to_validate`, the next scoped read invalidates and recomputes. New test added for this too.

SajidAlamQB:
> I'm not super familiar with the validation workflow but I'm wondering if it makes sense to send a dictionary of pipelines to the `context._pipelines_to_validate` instead of the pipeline names and importing `pipelines` there. We could directly send `filtered_pipelines` to context instead of loading them from names again?

Will leave as names for now to keep the context loosely coupled to pipeline objects (the `dict(project_pipelines)` lookup in the unscoped case still needs the import either way). Happy to revisit if you feel strongly.

ankatiyar (approved):
@SajidAlamQB No strong feelings. That makes sense, if I understand correctly, validation should work even when not specifically through `session.run()` I'm guessing so you'll need to import the pipelines in context anyway. Is that right? 
LGTM

SajidAlamQB:
> @SajidAlamQB No strong feelings. That makes sense, if I understand correctly, validation should work even when not specifically through `session.run()` I'm guessing so you'll need to import the pipelines in context anyway. Is that right? LGTM

Yep exactly the `context.params` is reachable outside `session.run()`, so the context needs the registry import for the unscoped fallback either way. Passing a dict would only save the lookup in the scoped path, not the import.

merelcht (approved):
Thanks @SajidAlamQB , LGTM 👍

PR #5507, opened 2026-04-19
fix: prevent false --checkout warning when using official starters

### LINKED ISSUE #5503: `--checkout` warning when creating a new project even when it's not used
## Description
I was creating a new project with `kedro new --starter spaceflights-pandas` and this warning shows up about `--checkout` having no effect with default project template even when i'm not using the `--checkout` flag

## Steps to Reproduce
`kedro new --starter spaceflights-pandas`

Image

## Expected Result
The warning should only show when `--checkout` is used

avasis-ai:
## Summary
Running `kedro new --starter spaceflights-pandas` (without `--checkout`) incorrectly shows a warning: "The --checkout flag has no effect when using the default template..."

This happens because when an official starter alias is used, `checkout` gets auto-populated with the current version string via `_select_checkout_branch_for_cookiecutter()`. The warning then fires because `checkout` is truthy, even though the user never passed `--checkout`.

## Changes
Added a check to only show the warning when `template_path` equals the default `TEMPLATE_PATH`, meaning the user is actually using the default template without a starter.

Closes #5503

Assisted-by: GLM 5.1

### DIFF
--- a/kedro/framework/cli/starters.py
+++ b/kedro/framework/cli/starters.py
@@ -826,7 +826,7 @@ def _make_cookiecutter_args_and_fetch_template(
     else:
         # Use the default template path for non PySpark or example options:
         starter_path = template_path
-        if checkout:
+        if checkout and template_path == str(TEMPLATE_PATH):
             warnings.warn(
                 "The --checkout flag has no effect when using the default template "
                 "without one of the following flags:"


lrcouto:
Already addressed here: https://github.com/kedro-org/kedro/pull/5506

avasis-ai:
Closing in favor of #5506 by @lrcouto which addresses the same issue. Thanks!

PR #4244, opened 2024-10-20
Add support for Oracle Cloud Infrastructure (OCI)

antonacio:
## Description

Kedro currently does not support reading and writing from and to Oracle Cloud Infrastructure (OCI) Object Storage filesystem, because the "oci://" protocol is not included in the list of supported cloud protocols.

## Development notes

Similarly to Azure Data Lake Storage Gen2 URIs, [OCI's Object Storage filesystem (ocifs)](https://ocifs.readthedocs.io/en/latest/) uses the @ syntax for I/O operations: "oci://bucket@namespace/path_to_file". 

Therefore, to make Kedro datasets work with OCIFS, I added "oci" to the `CLOUD_PROTOCOLS` list and included the same file path processing step done for Azure Data Lake Storage Gen2 URIs.

## Developer Certificate of Origin
We need all contributions to comply with the [Developer Certificate of Origin (DCO)](https://developercertificate.org/). All commits must be signed off by including a `Signed-off-by` line in the commit message. [See our wiki for guidance](https://github.com/kedro-org/kedro/wiki/Guidelines-for-contributing-developers/).

If your PR is blocked due to unsigned commits, then you must follow the instructions under "Rebase the branch" on the GitHub Checks page for your PR. This will retroactively add the sign-off to all unsigned commits and allow the DCO check to pass.

## Checklist

- [x] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [x] Signed off each commit with a [Developer Certificate of Origin (DCO)](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository)
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [x] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [ ] Added tests to cover my changes
- [ ] Checked if this change will affect Kedro-Viz, and if so, communicated that with the Viz team

PR #5512, opened 2026-04-21
fix(runner): detect streaming nodes via generator functions

### LINKED ISSUE #5412: Runner unexpectedly iterates over custom objects (mne.Epochs) yielding numpy.ndarray before saving to DataCatalog
## Description

When returning an object that implements `__iter__` (specifically `mne.Epochs` from the MNE-Python library) from a node, Kedro unexpectedly iterates over or unpacks the object before passing it to the custom dataset's `save` method. Instead of the `DataCatalog` receiving the original `mne.Epochs` object, it receives a `numpy.ndarray` (which is what `mne.Epochs` yields when iterated over).

Notably, this behavior persists even when explicitly returning a dictionary from the node and defining the node's outputs as a dictionary mapping to bypass standard tuple unpacking.

## Context

I am building a Kedro pipeline to process neurophysiological data, relying heavily on standard MNE-Python data types (specifically `Raw` and `Epochs`). While my pipeline successfully passes and saves the continuous `Raw` objects without any issues, I am running into a big blocker with the `Epochs` objects.

Somewhere between my node successfully returning the `mne.Epochs` objects and the custom dataset's `save` method being triggered, the data structure is being unexpectedly altered. Because `mne.Epochs` objects are iterables that yield NumPy arrays, it appears Kedro is inadvertently unpacking or iterating over the object during the handoff. As a result, the `save` method receives a raw `numpy.ndarray` instead of the expected `mne.Epochs` object, which crashes the pipeline.

This same data flow works flawlessly when executed directly in a Jupyter Notebook, so the issue seems isolated to how the Kedro runner handles iterables during I/O operations.

## Steps to Reproduce

The issue can be reproduced by running a pipeline that saves an mne.Epochs object from a node using the custom dataset class shown below.

*Note: The issue occurs regardless of whether the node returns a tuple/list or an explicit dictionary (`{"dataset_key": epochs_object}`).*

## Expected Result

The Kedro runner should pass the exact, unmodified `mne.Epochs` object returned by the Python function (node) directly into the `data` argument of the custom dataset's `save(self, data)` method.

## Actual Result

The Kedro runner appears to iterate over or unpack the object. The `save` method receives a `numpy.ndarray` (the first item yielded by the iterator) instead of the `mne.Epochs` object, causing an `AssertionError` (or a structural failure if type checking is disabled).

```sh
(lxml_ops) vcali@vcali:/Dev/Kedro/lxml-ops$ kedro run --from-nodes "extract_epochs_node"
[03/01/26 15:44:54] INFO     Using                                                                                                             __init__.py:269
                             '/Dev/Kedro/lxml-ops/.pixi/envs/default/lib/python3.13/site-packages/kedro/framework/project/rich_loggin                
                             g.yml' as logging configuration.                                                                                                 
[03/01/26 15:44:54] INFO     Kedro project lxml_ops                                                                                             session.py:335
[03/01/26 15:44:55] INFO     Kedro is sending anonymous usage data with the sole purpose of improving the product. No personal data or IP        plugin.py:242
                             addresses are stored on our side. To opt out, set the `KEDRO_DISABLE_TELEMETRY` or `DO_NOT_TRACK` environment                    
                             variables, or create a `.telemetry` file in the current working directory with the contents `consent: false`. To                 
                             hide this message, explicitly grant or deny consent. Read more at https://docs.kedro.org/en/stable/about/telemetry/              
[03/01/26 15:44:56] WARNING  Workflow tracking is disabled during partial pipeline runs (executed using --from-nodes, --to-nodes, --tags,     run_hooks.py:135
                             --pipeline, and more). `.viz/kedro_pipeline_events.json` will be created only during a full kedro run. See issue                 
                             https://github.com/kedro-org/kedro-viz/issues/2443 for more details.                                                             
                    INFO     Using synchronous mode for loading and saving data. Use the --async flag for potential performance gains. sequential_runner.py:59
                             https://docs.kedro.org/en/stable/build/run_a_pipeline/#load-and-save-asynchronously                                              
                    INFO     Loading data from resampled_psg (MNERawFIFDataset)...                                                        data_catalog.py:1048
Opening raw data file /Dev/Kedro/lxml-ops/data/02_intermediate/sub-1_resampled_psg-raw.fif.gz...
Isotrak not found
    Range : 0 ... 2745899 =      0.000 ... 27458.990 secs
Ready.
[03/01/26 15:44:57] INFO     Loading data from resampled_headband (MNERawFIFDataset)...                                                   data_catalog.py:1048
Opening raw data file /Dev/Kedro/lxml-ops/data/02_intermediate/sub-1_resampled_headband-raw.fif.gz...
Isotrak not found
    Range : 0 ... 2745899 =      0.000 ... 27458.990 secs
Ready.
                    INFO     Loading data from params:epoch_duration (MemoryDataset)...                                                   data_catalog.py:1048
                    INFO     Running node: extract_epochs_node: extract_epochs() ->                                                                node.py:531
Used Annotations descriptions: [np.str_('N1'), np.str_('N2'), np.str_('N3'), np.str_('REM'), np.str_('Wake')]
Used Annotations descriptions: [np.str_('N1'), np.str_('N2'), np.str_('N3'), np.str_('REM'), np.str_('Wake')]
Not setting metadata
914 matching events found
No baseline correction applied
0 projection items activated
Loading data for 914 events and 3001 original time points ...
0 bad epochs dropped
Not setting metadata
914 matching events found
No baseline correction applied
0 projection items activated
Loading data for 914 events and 3001 original time points ...
0 bad epochs dropped

[DEBUG] epochs_psg type before return: <class 'mne.epochs.Epochs'>

[DEBUG] epochs_hb type before return: <class 'mne.epochs.Epochs'>

[03/01/26 15:48:32] INFO     Saving data to epoched_psg (MneEpochFifDataset)...                                                           data_catalog.py:1008
                    WARNING  No nodes ran. Repeat the previous command to attempt a new run.                                                     runner.py:335
│ /Dev/Kedro/lxml-ops/.pixi/envs/default/lib/python3.13/site-packages/kedro/io/core.py:30 │
│ 3 in save                                                                                        │
│    300 │   │   │                                                                                 │
│    301 │   │   │   try:                                                                          │
│    302 │   │   │   │   self._logger.debug("Saving %s", str(self))                                │
│ ❱  303 │   │   │   │   save_func(self, data)                                                     │
│    304 │   │   │   except (DatasetError, FileNotFoundError, NotADirectoryError):                 │
│    305 │   │   │   │   raise                                                                     │
│    306 │   │   │   except Exception as exc:                                                      │
│ /Dev/Kedro/lxml-ops/src/lxml_ops/datasets.py:126 in save                                │
│   123 │   def save(self, data: mne.BaseEpochs) -> None:                                          │
│   124 │   │   """Saves the epoched MNE object."""                                                │
│   125 │   │   self._filepath.parent.mkdir(parents=True, exist_ok=True)                           │
│ ❱ 126 │   │   assert isinstance(data, mne.BaseEpochs), f"Expected mne.BaseEpochs, but got {typ   │
│   127 │   │   data.save(self._filepath, **self._save_args)                                       │
│   128 │                                                                                          │
│   129 │   def _exists(self) -> bool:                                                             │
AssertionError: Expected mne.BaseEpochs, but got <class 'numpy.ndarray'>

The above exception was the direct cause of the following exception:

│ /Dev/Kedro/lxml-ops/.pixi/envs/default/bin/kedro:10 in <module>                         │
│ /Dev/Kedro/lxml-ops/.pixi/envs/default/lib/python3.13/site-packages/kedro/framework/cli │
│ /cli.py:260 in main                                                                              │
│   257 │   """                                                                                    │
│   258 │   _init_plugins()                                                                        │
│   259 │   cli_collection = KedroCLI(project_path=find_kedro_project(Path.cwd()) or Path.cwd())   │
│ ❱ 260 │   cli_collection()                                                                       │
│   261                                                                                            │
│ /Dev/Kedro/lxml-ops/.pixi/envs/default/lib/python3.13/site-packages/click/core.py:1485  │
│ in __call__                                                                                      │
│ /Dev/Kedro/lxml-ops/.pixi/envs/default/lib/python3.13/site-packages/kedro/framework/cli │
│ /cli.py:162 in main                                                                              │
│   159 │   │   )                                                                                  │
│   160 │   │                                                                                      │
│   161 │   │   try:                                                                               │
│ ❱ 162 │   │   │   super().main(                                                                  │
│   163 │   │   │   │   args=args,                                                                 │
│   164 │   │   │   │   prog_name=prog_name,                                                       │
│   165 │   │   │   │   complete_var=complete_var,                                                 │
│ /Dev/Kedro/lxml-ops/.pixi/envs/default/lib/python3.13/site-packages/click/core.py:1406  │
│ in main                                                                                          │
│ /Dev/Kedro/lxml-ops/.pixi/envs/default/lib/python3.13/site-packages/click/core.py:1873  │
│ in invoke                                                                                        │
│ /Dev/Kedro/lxml-ops/.pixi/envs/default/lib/python3.13/site-packages/click/core.py:1269  │
│ in invoke                                                                                        │
│ /Dev/Kedro/lxml-ops/.pixi/envs/default/lib/python3.13/site-packages/click/core.py:824   │
│ in invoke                                                                                        │
│ /Dev/Kedro/lxml-ops/.pixi/envs/default/lib/python3.13/site-packages/kedro/framework/cli │
│ /project.py:277 in run                                                                           │
│   274 │   with KedroSession.create(                                                              │
│   275 │   │   env=env, conf_source=conf_source, runtime_params=params                            │
│   276 │   ) as session:                                                                          │
│ ❱ 277 │   │   return session.run(                                                                │
│   278 │   │   │   tags=tuple_tags,                                                               │
│   279 │   │   │   runner=runner_obj(is_async=is_async),                                          │
│   280 │   │   │   node_names=tuple_node_names,                                                   │
│ /Dev/Kedro/lxml-ops/.pixi/envs/default/lib/python3.13/site-packages/kedro/framework/ses │
│ sion/session.py:421 in run                                                                       │
│   418 │   │   │   run_params=record_data, pipeline=filtered_pipeline, catalog=catalog            │
│   419 │   │   )                                                                                  │
│   420 │   │   try:                                                                               │
│ ❱ 421 │   │   │   run_result = runner.run(                                                       │
│   422 │   │   │   │   filtered_pipeline,                                                         │
│   423 │   │   │   │   catalog,                                                                   │
│   424 │   │   │   │   hook_manager,                                                              │
│ /Dev/Kedro/lxml-ops/.pixi/envs/default/lib/python3.13/site-packages/kedro/runner/runner │
│ .py:115 in run                                                                                   │
│   112 │   │   │   )                                                                              │
│   113 │   │                                                                                      │
│   114 │   │   start_time = perf_counter()                                                        │
│ ❱ 115 │   │   self._run(pipeline, catalog, hook_or_null_manager, run_id)  # type: ignore[arg-t   │
│   116 │   │   end_time = perf_counter()                                                          │
│   117 │   │   run_duration = end_time - start_time                                               │
│   118                                                                                            │
│ /Dev/Kedro/lxml-ops/.pixi/envs/default/lib/python3.13/site-packages/kedro/runner/sequen │
│ tial_runner.py:63 in _run                                                                        │
│   60 │   │   │   │   "Using synchronous mode for loading and saving data. Use the --async fla    │
│   61 │   │   │   │   "for potential performance gains. https://docs.kedro.org/en/stable/build    │
│   62 │   │   │   )                                                                               │
│ ❱ 63 │   │   super()._run(                                                                       │
│   64 │   │   │   pipeline=pipeline,                                                              │
│   65 │   │   │   catalog=catalog,                                                                │
│   66 │   │   │   hook_manager=hook_manager,                                                      │
│ /Dev/Kedro/lxml-ops/.pixi/envs/default/lib/python3.13/site-packages/kedro/runner/runner │
│ .py:240 in _run                                                                                  │
│   237 │   │   │   │   │   │   hook_manager=hook_manager,                                         │
│   238 │   │   │   │   │   │   is_async=self._is_async,                                           │
│   239 │   │   │   │   │   │   run_id=run_id,                                                     │
│ ❱ 240 │   │   │   │   │   ).execute()                                                            │
│   241 │   │   │   │   │   done_nodes.add(node)                                                   │
│   242 │   │   │   │   except Exception:                                                          │
│   243 │   │   │   │   │   self._suggest_resume_scenario(pipeline, done_nodes, catalog)           │
│ /Dev/Kedro/lxml-ops/.pixi/envs/default/lib/python3.13/site-packages/kedro/runner/task.p │
│ y:88 in execute                                                                                  │
│    85 │   │   │   │   self.run_id,                                                               │
│    86 │   │   │   )                                                                              │
│    87 │   │   else:                                                                              │
│ ❱  88 │   │   │   node = self._run_node_sequential(                                              │
│    89 │   │   │   │   self.node,                                                                 │
│    90 │   │   │   │   self.catalog,                                                              │
│    91 │   │   │   │   self.hook_manager,  # type: ignore[arg-type]                               │
│ /Dev/Kedro/lxml-ops/.pixi/envs/default/lib/python3.13/site-packages/kedro/runner/task.p │
│ y:186 in _run_node_sequential                                                                    │
│   183 │   │   │   hook_manager.hook.before_dataset_saved(                                        │
│   184 │   │   │   │   dataset_name=name, data=data, node=node                                    │
│   185 │   │   │   )                                                                              │
│ ❱ 186 │   │   │   catalog.save(name, data)                                                       │
│   187 │   │   │   hook_manager.hook.after_dataset_saved(                                         │
│   188 │   │   │   │   dataset_name=name, data=data, node=node                                    │
│   189 │   │   │   )                                                                              │
│ /Dev/Kedro/lxml-ops/.pixi/envs/default/lib/python3.13/site-packages/kedro/io/data_catal │
│ og.py:1015 in save                                                                               │
│   1012 │   │   │   extra={"markup": True},                                                       │
│   1013 │   │   )                                                                                 │
│   1014 │   │                                                                                     │
│ ❱ 1015 │   │   dataset.save(data)                                                                │
│   1016 │                                                                                         │
│   1017 │   def load(self, ds_name: str, version: str | None = None) -> Any:                      │
│   1018 │   │   """Loads a registered dataset.                                                    │
│ /Dev/Kedro/lxml-ops/.pixi/envs/default/lib/python3.13/site-packages/kedro/io/core.py:30 │
│ 8 in save                                                                                        │
│    305 │   │   │   │   raise                                                                     │
│    306 │   │   │   except Exception as exc:                                                      │
│    307 │   │   │   │   message = f"Failed while saving data to dataset {self!s}.\n{exc!s}"       │
│ ❱  308 │   │   │   │   raise DatasetError(message) from exc                                      │
│    309 │   │                                                                                     │
│    310 │   │   save.__annotations__["data"] = save_func.__annotations__.get("data", Any)         │
│    311 │   │   save.__annotations__["return"] = save_func.__annotations__.get("return")          │
DatasetError: Failed while saving data to dataset 
lxml_ops.datasets.MneEpochFifDataset(filepath='/Dev/Kedro/lxml-ops/data/03_primary/sub-1_psg-epo.fif.gz').
Expected mne.BaseEpochs, but got <class 'numpy.ndarray'>

```

### Catalog
```yaml
"epoched_headband":
  type: lxml_ops.datasets.MneEpochFifDataset
  filepath: "data/03_primary/sub-1_headband-epo.fif.gz"

"epoched_psg":
  type: lxml_ops.datasets.MneEpochFifDataset
  filepath: "data/03_primary/sub-1_psg-epo.fif.gz"

```

### Custom Dataset 
```python
class MneEpochFifDataset(AbstractDataset):
    """
    A custom Kedro dataset for saving and loading epoched MNE objects
    in the Neuromag -epo.fif format.
    """
    def __init__(
        self,
        filepath: str | Path,
        load_args: dict[str, Any] | None = None,
        save_args: dict[str, Any] | None  = None
    ):
        self._filepath = Path(filepath).resolve()
        self._load_args = load_args or {}
        self._save_args = save_args or {}

        if "overwrite" not in self._save_args:
            self._save_args["overwrite"] = True

    def load(self) -> mne.BaseEpochs:
        """Loads the epoched MNE object."""
        return mne.read_epochs(self._filepath, **self._load_args)

    def save(self, data: mne.BaseEpochs) -> None:
        """Saves the epoched MNE object."""
        self._filepath.parent.mkdir(parents=True, exist_ok=True)
        assert isinstance(data, mne.BaseEpochs), f"Expected mne.BaseEpochs, but got {type(data)}"
        data.save(self._filepath, **self._save_args)

    def _exists(self) -> bool:
        """Checks whether the Epochs FIF file exists."""
        return self._filepath.exists()

    def _describe(self) -> dict[str, Any]:
        """Provides metadata for the Kedro catalog."""
        return {"filepath": str(self._filepath)}

```

### Pipeline
```python
    return Pipeline(
        [
            Node(
                func=align_and_annotate,
                inputs=["psg_raw_data", "headband_raw_data", "psg_events_data"],
                outputs=["aligned_annotated_psg", "aligned_annotated_headband"],
                name="align_and_annotate_node",
            ),
            Node(
                func=resample_signals,
                inputs=[
                    "aligned_annotated_psg",
                    "aligned_annotated_headband",
                    "params:target_sfreq",
                ],
                outputs=["resampled_psg", "resampled_headband"],
                name="resample_signals_node",
            ),
            Node(
                func=extract_epochs,
                inputs=["resampled_psg", "resampled_headband", "params:epoch_duration"],
                outputs={
                    "epoched_psg": "epoched_psg",
                    "epoched_headband": "epoched_headband",
                },
                name="extract_epochs_node",
            ),
        ]
    )

```

## Your Environment

* Kedro version: 1.2.0
* Python version: 3.13.0
* Operating system and version: WSL2 Ubuntu 24.04.3 LTS
* MNE-Python version: 1.11.0

BhavayChopra:
## Description

Fixes #5412.

Non-async runners used `all(isinstance(d, Iterator) for d in outputs.values())` to decide whether a node's outputs should be streamed chunk-by-chunk through `interleave`. Because `collections.abc.Iterator` is a structural ABC, iterator-like domain objects such as `mne.Epochs` satisfied the check and were consumed before `catalog.save` received them.

The streaming decision is now based on the node-level contract instead:

```python
if outputs and inspect.isgeneratorfunction(node.func):
    ...
```

That keeps chunked saving restricted to generator-function nodes while passing ordinary iterable business objects to `catalog.save` unchanged. It also avoids depending on the concrete iterator type produced by `more_itertools.unzip()`, which differs across supported `more_itertools` versions.

## Development notes

- Rebased onto latest `upstream/main`.
- Replaced output-value `Iterator` detection with `inspect.isgeneratorfunction(node.func)`.
- Kept the `outputs and ...` guard so nodes with no outputs do not enter the streaming path vacuously.
- Moved the annotation-only `Iterable` import into `TYPE_CHECKING`.
- Updated `RELEASE.md` to cover all non-async runners (`SequentialRunner`, `ThreadRunner`, and `ParallelRunner`) and to credit the PR author.

### Tests

Added focused coverage in `tests/runner/test_run_generator_node.py`:

- Existing generator-node tests still assert chunked saving for single, tuple, list, and dict generator outputs.
- A compatibility test monkeypatches `kedro.pipeline.node.unzip` to return `map` objects, matching older supported `more_itertools` behaviour, and verifies multi-output generator nodes still stream chunk-by-chunk.
- `TestNonGeneratorIterables` verifies iterator-like but non-generator node outputs are passed to `catalog.save` unchanged.
- A real `numpy.ndarray` pass-through test pins a common iterable object case.

Local verification:

- `uv run pytest tests/runner/test_run_generator_node.py tests/runner/test_task.py --no-cov -q` -> 18 passed
- `uv run pytest tests/runner/ --no-cov -q --ignore=tests/runner/test_parallel_runner.py -k "not ParallelRunner"` -> 220 passed, 23 deselected
- `uv run pytest tests/runner/test_parallel_runner.py --no-cov -q` -> 69 passed
- `uvx ruff format --check tests/runner/test_run_generator_node.py kedro/runner/task.py` -> passed
- `uvx ruff check tests/runner/test_run_generator_node.py` -> passed

Thanks to @victor-cali for the clear reproduction in the original issue.

Signed off.

## Checklist

- [x] Checked if this change will affect Kedro-Viz — no impact expected (runner-internal predicate change).


PragnyaKhandelwal:
Hi @BhavayChopra and team! I saw this PR, and since I work heavily with mne-python, I wanted to help test the fix. I pulled this branch locally, set up a test script using `MemoryDataset`, and passed `EpochsArray` objects through a `SequentialRunner` node. I can confirm the runner now correctly preserves the `mne.BaseEpochs` objects by identity and no longer incorrectly iterates them into a `numpy.ndarray`. Great fix, Bhavay! Let me know if the team needs any other MNE-specific edge cases tested.

kedro/runner/task.py
```diff
@@ -166,8 +167,13 @@ def _run_node_sequential(
...
         items: Iterable = outputs.items()
-        # if all outputs are iterators, then the node is a generator node
-        if all(isinstance(d, Iterator) for d in outputs.values()):
+        # If all outputs are generator objects, treat this as a generator node
+        # and interleave the chunk streams into the catalog. We deliberately
+        # check for generators (not the broader ``Iterator`` ABC) so that
+        # iterable business objects -- e.g. ``mne.Epochs``, ``numpy.ndarray``,
+        # ``torch.utils.data.IterableDataset`` -- are passed through to
+        # ``catalog.save`` unchanged. See kedro-org/kedro#5412.
+        if outputs and all(inspect.isgenerator(d) for d in outputs.values()):
```
ElenaKhaustova: **Critical (regression, please address):** I think this predicate works in your local environment but silently corrupts data on older `more_itertools` versions that Kedro still supports.

**What's happening:** by the time we reach this line, multi-output generator-function nodes have already been routed through `Node._outputs_to_dictionary`, which calls `more_itertools.unzip(...)`. The return type of `unzip` is **not stable across versions**:

- `more_itertools >= 11.0.0` → returns generators (uses `iter_suppress`, which is itself a generator function) → `inspect.isgenerator(d) is True` → streaming path taken ✅
- `more_itertools 8.14.0 – 10.6.0` → returns `tuple(map(...) for ...)` → `inspect.isgenerator(<map>) is False` → streaming path **skipped**, and an unconsumed `map` object is passed to `catalog.save(name, <map object>)` ❌

Kedro declares `more_itertools>=8.14.0` in `pyproject.toml`, so a fairly large slice of users hits the broken path. I reproduced this end-to-end against the PR head + `more_itertools==8.14.0`:

```
more_itertools 8.14.0
left.save call_count: 1 (should be 5 for streaming)
  -> arg passed: <map object at 0x10706f2b0>
```

The pipeline reports "completed successfully" — there's no error, just silent corruption — which is arguably worse than the bug you set out to fix.

**Suggested fix** — detect the *node*, not the *output type*. This is the same check `task.py:60` already uses for the async-rejection path, and the same check `kedro/pipeline/node.py::_outputs_to_dictionary` uses to decide whether to peek at output values:

```python
if outputs and inspect.isgeneratorfunction(node.func):
```

I tried this locally on `more_itertools==8.14.0` and the full `tests/runner/test_run_generator_node.py` suite (all 8 tests, including your new `TestNonGeneratorIterables`) passes. It also passes on the latest `more_itertools`. As a bonus, the rationale comment can collapse to something like:

```python
# Stream chunks only for true generator-function nodes, so iterable
# business objects (e.g. mne.Epochs, numpy.ndarray) are passed to
# catalog.save unchanged. See #5412.
```

The other option is to raise the floor to `more_itertools>=11.0.0` in `pyproject.toml` and add a release note for the dependency bump, but the `isgeneratorfunction(node.func)` swap is cleaner and avoids touching deps.

RELEASE.md
```diff
@@ -6,8 +6,11 @@ NOTE: This session implementation is under active development and may occasional
 * Added inspection API to get project snapshot.
 
 ## Bug fixes and other changes
+* Fixed `SequentialRunner` iterating over iterable-but-non-generator node outputs (e.g. `mne.Epochs`, `numpy.ndarray`, `torch.utils.data.IterableDataset`) before saving them. Streaming behaviour is now restricted to actual generator objects, so custom business objects are passed to `catalog.save` unchanged.
```
ElenaKhaustova: **Suggestion:** the same buggy code path is shared by `ThreadRunner` and (when not async) `ParallelRunner` — they all call `_run_node_sequential` from `Task.execute()`. I confirmed this with a quick reproducer on `main`:

```
THREAD RUNNER left.save calls: [call('left_chunk_0'), call('left_chunk_1'), call('left_chunk_2')]
```

Could you reword the bullet so it doesn't undersell the scope, e.g.:

> Fixed all non-async runners (`SequentialRunner`, `ThreadRunner`, `ParallelRunner`) iterating over iterable-but-non-generator node outputs (e.g. `mne.Epochs`, `numpy.ndarray`, `torch.utils.data.IterableDataset`) before saving them. Streaming behaviour is now restricted to actual generator-function nodes, so custom business objects are passed to `catalog.save` unchanged.

RELEASE.md
```diff
@@ -6,8 +6,11 @@ NOTE: This session implementation is under active development and may occasional
...
+
 ## Documentation changes
 ## Community contributions
+* [victor-cali](https://github.com/victor-cali)
```
ElenaKhaustova: **Suggestion:** by convention the `## Community contributions` list credits the PR author, not the issue reporter (e.g. recently `[Pedro Antonacio](https://github.com/antonacio)` for #4244 by `antonacio`, `[Adam Kells](https://github.com/adamkells)` for the PR by `adamkells`). `victor-cali` is already credited via `Fixes #5412` and the PR description. Suggest:

```diff
-* [victor-cali](https://github.com/victor-cali)
+* [BhavayChopra](https://github.com/BhavayChopra)
```

Maintainers use this section to write release announcements, so getting the author here is what makes that work.

tests/runner/test_run_generator_node.py
```diff
@@ -100,3 +103,118 @@ def test_generator_node_dict(self, mocker, catalog):
         assert left.save.call_args_list == expected_left
         assert 10 == right.save.call_count
         assert right.save.call_args_list == expected_right
+
+
+class _IterableNotGenerator:
```
ElenaKhaustova, line 150: **Suggestion (nice-to-have):** the hand-rolled `_IterableNotGenerator` fixture cleanly proves the predicate change, but a single test with a real `numpy.ndarray` would pin the *actual* repro from #5412 and make the regression test resilient to future "the fixture still passes if I revert to `Iterator`" mistakes. `numpy` is already a transitive test dep (via `pandas` in `tests/conftest.py`), so no new dependency would be needed:

```python
def test_numpy_array_outputs_are_passed_through(mocker, catalog):
    import numpy as np
    arr_a, arr_b = np.array([1, 2, 3]), np.array([4, 5, 6])
    # ... same pattern as test_multiple_iterable_outputs_are_passed_through ...
    assert saved_a is arr_a and saved_b is arr_b
```

Totally optional — the existing tests already cover the contract.

ElenaKhaustova (changes requested):
## Kedro PR Review

Hi @BhavayChopra — thanks for picking this up, and for the unusually thorough write-up. Your diagnosis of #5412 is spot-on: the `Iterator` ABC predicate was structurally wrong (the existing comment "generator node" and the actual check disagreed), and the new `TestNonGeneratorIterables` class is a really nice addition.

I reproduced the original bug locally first to confirm, and everything you wrote about the root cause checks out. Two things came up in review that I'd like to discuss before merge — one of them is a subtle regression I think the PR will introduce that isn't visible in CI; details inline.

### Overview
- **Critical:** 1 (silent regression on supported `more_itertools` versions)
- **Suggestions:** 3

### Critical (please address before merge)
- [x] `kedro/runner/task.py:176` — `inspect.isgenerator(d)` regresses multi-output generator nodes on `more_itertools < 11.0.0` (Kedro's declared floor is `8.14.0`). Suggested fix: switch to `inspect.isgeneratorfunction(node.func)`.

### Suggestions (optional)
- [x] `RELEASE.md:9` — bug also affects `ThreadRunner` and `ParallelRunner` (all non-async runners flow through `_run_node_sequential`); please reword.
- [x] `RELEASE.md:13` — community-contribution credit by convention goes to the PR author (`BhavayChopra`), not the issue reporter (`victor-cali`).
- [x] `tests/runner/test_run_generator_node.py:108` — consider adding one regression test using a real `numpy.ndarray` (already a transitive test dep) so the fixture can't be silently re-broken.

### Notes
- **General:** Direction and intent are correct, diff is small and focused, new tests are a clear improvement, DCO signed.
- **API usage / docs:** No public API or docs impact.
- **Mergeability:** GitHub reports `CONFLICTING` against `main`; you'll need a quick rebase.
- **CI:** Green, but CI doesn't pin against the lower bound of `more_itertools`, which is why the regression below isn't visible there.

Thanks again — once the predicate change is in, this is a really clean fix.

BhavayChopra:
Thanks Elena, this was a great catch. I agree with your read: the runner should not depend on whatever concrete iterator type `more_itertools.unzip()` happens to return.

I updated the branch to use `inspect.isgeneratorfunction(node.func)` instead, which feels like the right level of abstraction here. Streaming is now decided from the node function itself, not from the transformed output values.

I also added a regression test that mimics older `more_itertools` returning `map` objects, plus the numpy pass-through test you suggested. While touching this, I rebased on latest `main`, updated the release note to mention all non-async runners, and changed the community credit to the PR author.

Local checks I ran:

- `uv run pytest tests/runner/test_run_generator_node.py tests/runner/test_task.py --no-cov -q` passed, 18 tests
- `uv run pytest tests/runner/ --no-cov -q --ignore=tests/runner/test_parallel_runner.py -k "not ParallelRunner"` passed, 220 tests
- `uv run pytest tests/runner/test_parallel_runner.py --no-cov -q` passed, 69 tests
- Ruff format and check passed on the changed Python files

I also ran the full suite locally. It got to 2117 passed and 8 skipped, with 2 failures in unrelated session logging tests. I checked those against a clean `upstream/main` worktree and reproduced the failure there too, so they do not appear to come from this PR.

ElenaKhaustova:
Thank you @BhavayChopra, great work 🚀

ElenaKhaustova (approved):
Just realised that I forgot to approve it 🫣

RELEASE.md
```diff
@@ -15,14 +15,15 @@ NOTE: This session implementation is under active development and may occasional
 * Removed outdated `TRANSCODING_SEPARATOR` alias from `kedro.pipeline.pipeline`.
 * Added an optional `pydantic` dependency extra, allowing users to enable Pydantic support with `pip install "kedro[pydantic]"`.
 * Fixed parameter validation for namespaced pipelines.
+* Fixed non-async runners (`SequentialRunner`, `ThreadRunner`, and `ParallelRunner`) iterating over iterable-but-non-generator node outputs (e.g. `mne.Epochs`) before saving them. Streaming behaviour is now restricted to generator-function nodes, so custom business objects are passed to `catalog.save` unchanged.
```
SajidAlamQB:
```suggestion
* Fixed runners iterating over iterable-but-non-generator node outputs (e.g. `mne.Epochs`) before saving them. Streaming behaviour is now restricted to generator-function nodes, so custom business objects are passed to `catalog.save` unchanged.
```

SajidAlamQB (approved):
LGTM, thank you @BhavayChopra!

PR #5501, opened 2026-04-15
HTTP server and related CLI command

ankatiyar:
## Description

Resolve https://github.com/kedro-org/kedro/issues/5446

This PR introduces:
- CLI command in `kedro/framework/cli/server.py`: `kedro server start`
- Server API layer in `kedro/server` module

## Development notes

Some things to note (and would like opinions on the design of):
- The server API layer is designed solely for `KedroServiceSession`. The use of `SESSION_CLASS` in `settings.py` will have no effect. 
- I've changed the original prototype CLI command from `kedro server http start` to `kedro server start` (it was originally two layered because of `kedro server mcp start`) - does it make sense to keep it two layered and see if mcp plugin commands can plug into it?
- Since `project_path`, `env` and `conf_source` go into session creation, I've added them as arguments (env and conf-source, not project_path) for `kedro server start`. These are set as env variables (along with project path) and read by the server API module. This is to not load the module when commands are loaded (lazy loading) to keep things optional.
- `KedroServiceSession.create()` is called only at the first time run request comes through, then is assigned to the `app.state` --> this is so when being used for inspection, a session is not created simply when the FastAPI app is spun up. 

## Developer Certificate of Origin
We need all contributions to comply with the [Developer Certificate of Origin (DCO)](https://developercertificate.org/). All commits must be signed off by including a `Signed-off-by` line in the commit message. [See our wiki for guidance](https://github.com/kedro-org/kedro/wiki/Guidelines-for-contributing-developers/).

If your PR is blocked due to unsigned commits, then you must follow the instructions under "Rebase the branch" on the GitHub Checks page for your PR. This will retroactively add the sign-off to all unsigned commits and allow the DCO check to pass.

## Checklist

- [ ] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [ ] Signed off each commit with a [Developer Certificate of Origin (DCO)](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository)
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [ ] Added tests to cover my changes
- [ ] Checked if this change will affect Kedro-Viz, and if so, communicated that with the Viz team

Issue #5446, opened 2026-03-11
Implement HTTP server for pipeline execution in Kedro core


DimedS:
## Summary

Implement a minimal HTTP server in Kedro core that allows triggering Kedro pipelines via a REST API.

This builds on top of the shared execution runner introduced in the #5445  and provides an HTTP interface that mirrors the behaviour of the `kedro run` CLI command.

The goal of this feature is to provide a simple, programmatic way to trigger pipelines from external systems such as:

- orchestration tools (Airflow, Prefect, Dagster)
- CI/CD pipelines
- internal dashboards
- automation scripts

The HTTP server should remain **minimal and additive**, aligned with the direction agreed in the KEP discussion #5387 .

---

## Context

Currently Kedro pipelines can be executed via:

- the CLI (`kedro run`)
- programmatic Python code using `KedroSession`

However, many production environments require triggering pipelines via **HTTP interfaces**.

The prototype implementation in #5370 introduced an HTTP server using FastAPI.

After the KEP discussion in #5387, the agreed direction is:

- implement a **minimal HTTP server in Kedro core**
- keep the scope intentionally small
- leave orchestration, scheduling, concurrency, and authentication to external systems

This ticket implements the HTTP interface on top of the shared runner introduced in the previous issue.

---

## Related issues

- Parent initiative: #5262 Kedro Service for Universal Deployment
- API layer issue: #5263 Build API Layer for Pipeline Execution
- Shared runner issue: (previous ticket)
- Prototype implementation: #5370 Add Kedro http and mcp servers
- KEP decision: #5387 KEP-4 Built-in HTTP & MCP Server
- Related session work: #5425 KedroServiceSession v1: Stateless session

---

## Scope

Implement a minimal HTTP server in Kedro core using FastAPI.

Proposed structure:

```
kedro/server/
 ├─ http_server.py
 ├─ runner.py
 └─ models.py
```

The HTTP server should:

- expose a small REST API for pipeline execution
- validate requests using the shared Pydantic models
- call the shared execution runner
- return a structured response

The HTTP layer should only handle **transport concerns** and must not contain pipeline execution logic.

---

## CLI command

The server should be accessible via the Kedro CLI:

```
kedro server http start
```

Possible options may include:

```
--host
--port
--reload
--debug
```

---

### Health endpoint

```
GET /health
```

Returns a simple health check response confirming the server is running.

Example response:

```
{
  "status": "ok"
}
```

---

### Pipeline execution endpoint

```
[...]

PR #5520, opened 2026-04-28
Part 1: Scaffolding for HTTP server + health check API

ankatiyar:
## Description

**First** Part of #5446 

The complete code is in #5501 for testing
This PR just adds the FastAPI HTTP server layer with one end point for health check (GET)

## Development notes

- I have left the `env` and `conf_source` arguments in the `create_http_server()` function and it is set as the `app.state`. This will be used to create the session for execution of pipelines.

## Checklist


kedro/server/http_server.py
```diff
@@ -0,0 +1,93 @@
...
+        title="Kedro Server",
+        description="HTTP API for running Kedro pipelines",
+        version=kedro_version,
+        lifespan=lifespan,
+    )
+    app.state.default_env = resolved_env
+    app.state.default_conf_source = resolved_conf_source
+    app.state.project_path = (
+        Path(project_path).resolve() if project_path is not None else get_project_path()
+    )
```
ankatiyar: will be using this for creation of session when I add post endpoint in a subsequent PR

kedro/server/http_server.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @app.get("/health", response_model=HealthResponse, tags=["health"])
+    async def health() -> HealthResponse:
+        """Health check endpoint.
+
+        Returns server status and Kedro version information.
+        """
+        project_path = app.state.project_path
+        return HealthResponse(
+            status="healthy",
```
merelcht, line 90: So this always returns "healthy" or is there a case where it returns something else? What if e.g. the server is not responding?

kedro/server/utils.py
```diff
@@ -0,0 +1,54 @@
...
+    Raises:
+        ServerSettingsError: If KEDRO_PROJECT_PATH is not set.
+    """
+    project_path = os.environ.get(KEDRO_PROJECT_PATH_ENV)
+
+    if not project_path:
+        raise ServerSettingsError(
+            f"Environment variable '{KEDRO_PROJECT_PATH_ENV}' is not set. "
+            "The Kedro server must be started using 'kedro server start' command "
+            "from within a Kedro project directory."
```
merelcht: can a user also manually set the env variable or can you only resolve this by running the CLI command?

pyproject.toml
```diff
@@ -98,7 +99,11 @@ jupyter = [
 benchmark = [
     "asv"
 ]
-all = [ "kedro[test,docs,jupyter,benchmark]" ]
+server = [
```
merelcht, line 101: Do you need pydantic as well?

merelcht (commented):
Generally looks good, some small questions from my side.

kedro/server/utils.py
ankatiyar: Yes, the user can also set this env variable. This is set by the CLI command when used. Otherwise users can:
- programatically set it using argument in `create_http_server()` and that will be used first if available
- It calls this function `get_project_path()` which reads the env variable which could be set by the CLI command or the user

kedro/server/http_server.py
ankatiyar: I was assuming when the server is unhealthy it just sends no response or timeout. When the project path doesn't exist, we can send "unhealthy" - there is an error in the `get_project_path()` bit when the path doesnt exist or is not set.
Looking into this and will get back with what happens to server when there's these errors.

kedro/server/__init__.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import annotations
+
+from typing import Any
+
+__all__ = ["create_http_server"]
+
+
+def create_http_server() -> Any:
```
ravi-kumar-pilla:
```suggestion
def create_http_server(**kwargs: Any) -> Any:
```

kedro/server/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+    """Create the HTTP server application (lazy import)."""
+    from kedro.server.http_server import create_http_server as _create
+
+    return _create()
```
ravi-kumar-pilla:
```suggestion
    return _create(**kwargs)
```

ravi-kumar-pilla: The create_http_server accepts parameters but the one here does not. Why is it so ?

kedro/server/http_server.py
```diff
@@ -0,0 +1,93 @@
+"""FastAPI HTTP server factory and route definitions for Kedro."""
+
+from __future__ import annotations
+
+import logging
+import os
+from collections.abc import AsyncGenerator
```
ravi-kumar-pilla: If `AsyncGenerator` is only for type checking, we can remove this import ?

kedro/server/http_server.py
```diff
@@ -0,0 +1,93 @@
...
+def create_http_server(
+    project_path: str | Path | None = None,
+    env: str | None = None,
+    conf_source: str | None = None,
+) -> FastAPI:
+    """Create and configure the FastAPI HTTP server.
+
+    Programmatic values passed to this factory take precedence over
+    environment-variable defaults set by the CLI.
+
```
ravi-kumar-pilla: Google-style docstring on create_http_server - Add an Args: section for project_path, env, and conf_source so API docs stay consistent with the rest of Kedro.

kedro/server/http_server.py
```diff
@@ -0,0 +1,93 @@
...
+    resolved_conf_source = (
+        conf_source
+        if conf_source is not None
+        else os.environ.get(KEDRO_SERVER_CONF_SOURCE)
+    )
+
+    app = FastAPI(
+        title="Kedro Server",
+        description="HTTP API for running Kedro pipelines",
+        version=kedro_version,
```
ravi-kumar-pilla, line 75: [Nit] should we have a server api version or continue using the kedro release versions ?

kedro/server/models.py
```diff
@@ -0,0 +1,20 @@
...
+
+from typing import Literal
+
+from pydantic import BaseModel, Field
+
+
+class HealthResponse(BaseModel):
+    """Response model for health check endpoint."""
+
+    status: Literal["healthy", "unhealthy"] = Field(
```
ravi-kumar-pilla, line 13:
```suggestion
    status: Literal["healthy"] = Field(
```

ravi-kumar-pilla: I am not sure where unhealthy will be used. This should be the server status whether it can take requests. So I feel unhealthy does not make sense since server is returning the HealthResponse

kedro/server/__init__.py
ankatiyar: Added kwargs to both

kedro/server/http_server.py
ankatiyar: Yep, removed

kedro/server/http_server.py
ankatiyar: Added

kedro/server/http_server.py
ankatiyar: It would be tricky to maintain version of the server API since it's a core functionality, I would prefer to keep this same as the kedro version 🤔

kedro/server/http_server.py
ankatiyar: I checked in various cases when the path does not exist or bootstrap is unsuccessful, we error out in the startup process. If the server is up and capable of responding, it is healthy (within the functionality we offer so far). Can keep this endpoint minimal for now (this might change when we add additional endpoints or functionality in the future)

kedro/server/models.py
ankatiyar: I think it's worth keeping this still, our health endpoint is minimal but incase users extend the server API for their use-cases, it would offer the option of returning "unhealthy" status without having to re-define the model. WDYT?

rashidakanchwala (approved):
LGTM. thanks

kedro/server/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+from typing import Any
+
+__all__ = ["create_http_server"]
+
+
+def create_http_server(**kwargs: Any) -> Any:
+    """Create the HTTP server application (lazy import)."""
+    from kedro.server.http_server import create_http_server as _create
+
+    return _create(**kwargs)
```
rashidakanchwala:
```suggestion
  def create_http_server(**kwargs: Any) -> Any:
      """Create the HTTP server application (lazy import)."""
      try:
          from kedro.server.http_server import create_http_server as _create
      except ModuleNotFoundError as exc:
          if exc.name == "fastapi":
              raise ImportError(
                  "The Kedro HTTP server requires optional dependencies. "
                  "Install them with `pip install 'kedro[server]'`."
              ) from exc
          raise

      return _create(**kwargs)
```

kedro/server/http_server.py
```diff
@@ -0,0 +1,96 @@
...
+def create_http_server(
+    project_path: str | Path | None = None,
+    env: str | None = None,
+    conf_source: str | None = None,
+) -> FastAPI:
+    """Create and configure the FastAPI HTTP server.
+
+    Programmatic values passed to this factory take precedence over
+    environment-variable defaults set by the CLI.
+    Arguments:
```
ElenaKhaustova: **Critical:** Use `Args:` instead of `Arguments:` — Google-style is the convention across `kedro/` (e.g. `kedro/utils.py`, `kedro/io/catalog_config_resolver.py`). `Arguments:` isn't recognised as a parameter section by Sphinx napoleon or mkdocstrings/Griffe by default, so the rendered API docs will fold these lines into one paragraph.

Also add a blank line between line 55 (`...set by the CLI.`) and the section header — Google style requires a blank line before each section header for it to be parsed correctly.

kedro/server/utils.py
```diff
@@ -0,0 +1,54 @@
...
+from pathlib import Path
+
+# Environment variable name for project path
+KEDRO_PROJECT_PATH_ENV = "KEDRO_PROJECT_PATH"
+KEDRO_SERVER_ENV = "KEDRO_SERVER_ENV"
+KEDRO_SERVER_CONF_SOURCE = "KEDRO_SERVER_CONF_SOURCE"
+
+# Default server settings
+DEFAULT_HOST = "127.0.0.1"
+DEFAULT_HTTP_PORT = 8000
```
ElenaKhaustova: **Suggestion:** `DEFAULT_HOST` and `DEFAULT_HTTP_PORT` aren't referenced anywhere in this PR — they only become relevant once the `kedro server http start` CLI lands (per the acceptance criteria in #5446). Consider deferring them to that PR (so they land alongside their consumer), or at minimum adding a brief comment noting the intended consumer so future readers don't think they're an unused public API.

kedro/server/http_server.py
```diff
@@ -0,0 +1,96 @@
...
+    )
+
+    app = FastAPI(
+        title="Kedro Server",
+        description="HTTP API for running Kedro pipelines",
+        version=kedro_version,
+        lifespan=lifespan,
+    )
+    app.state.default_env = resolved_env
+    app.state.default_conf_source = resolved_conf_source
```
ElenaKhaustova, line 79: **Suggestion:** `app.state.default_env` and `app.state.default_conf_source` are written here but never read anywhere in this PR (only `app.state.project_path` is read by `health()` on line 89). The PR description notes these are for the session-creation work in Part 2 — fair enough, but two consequences:

1. There's no test that the resolution precedence (`argument > env var > None`) actually works, so a regression could ship undetected. A focused unit test against `app.state` would be cheap, e.g.:
   ```python
   def test_create_http_server_resolves_env_from_argument_over_env_var(monkeypatch, mocker, tmp_path):
       monkeypatch.setenv("KEDRO_SERVER_ENV", "from_env_var")
       mocker.patch("kedro.server.http_server.get_project_path", return_value=tmp_path)
       mocker.patch("kedro.server.http_server.bootstrap_project")
       app = create_http_server(env="from_argument")
       assert app.state.default_env == "from_argument"
       assert app.state.default_conf_source is None
   ```
2. Alternatively, defer accepting `env` / `conf_source` until Part 2 to keep this PR strictly minimal (which would match the "scaffolding + health check" framing in the title).

kedro/server/utils.py
```diff
@@ -0,0 +1,54 @@
...
+class ServerSettingsError(Exception):
+    """Raised when server settings are invalid or missing."""
+
+    pass
+
+
+def get_project_path() -> Path:
+    """Resolve the Kedro project path from environment variable.
+
+    The server must be started via `kedro server start` which sets
```
ElenaKhaustova: **Suggestion:** Per your reply to merelcht above, the project path can also be set manually via the env var or by passing `project_path=` to `create_http_server()`. The docstring (line 27-29) and the error message (line 41-43) currently only mention the CLI path ("must be started using `kedro server start` command"), which understates the contract. Could we broaden the wording to reflect all three paths so the public-facing message matches the actual behaviour?

Also, a small naming note: issue #5446 spells the CLI as `kedro server http start` (with `http`), so worth aligning here to avoid a churn when the CLI lands.

kedro/server/__init__.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import annotations
```
ElenaKhaustova, line 6: **Suggestion:** Add a module-level docstring at the top of `kedro/server/__init__.py`. It's the public entry point for the new HTTP server feature, and every other public `kedro/*/__init__.py` has one (e.g. `kedro/io/__init__.py`, `kedro/runner/__init__.py`, `kedro/inspection/__init__.py`). Good place to note that this is optional and requires `pip install "kedro[server]"`.

Something like:
```python
"""``kedro.server`` provides an optional HTTP server for triggering Kedro pipelines.

Requires the `server` extra: ``pip install "kedro[server]"``.
"""
```

kedro/server/http_server.py
```diff
@@ -0,0 +1,96 @@
...
+    env: str | None = None,
+    conf_source: str | None = None,
+) -> FastAPI:
+    """Create and configure the FastAPI HTTP server.
+
+    Programmatic values passed to this factory take precedence over
+    environment-variable defaults set by the CLI.
+    Arguments:
+        project_path: Optional path to the Kedro project to serve. If not provided,
+            it will be resolved from environment variables or the current working directory.
```
ElenaKhaustova: **Suggestion:** The docstring says `project_path` falls back to "environment variables or the current working directory", but `get_project_path()` (in `kedro/server/utils.py`) only reads `KEDRO_PROJECT_PATH` and raises `ServerSettingsError` if it's missing — there's no CWD fallback. Tighten to match the actual behaviour, e.g.:
```
project_path: Path to the Kedro project to serve. If not provided,
    falls back to the ``KEDRO_PROJECT_PATH`` environment variable.
```

kedro/server/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+
+from typing import Any
+
+__all__ = ["create_http_server"]
+
+
+def create_http_server(**kwargs: Any) -> Any:
```
ElenaKhaustova: **Suggestion:** `(**kwargs: Any) -> Any` is the public entry point but discards the explicit `(project_path, env, conf_source) -> FastAPI` signature, so IDEs / mypy / pyright can't help users (e.g. catch `projct_path=...` typos or wrong types). Consider mirroring the real signature with a `TYPE_CHECKING`-guarded `FastAPI` import so fastapi stays lazily imported but users get type info:

```python
from __future__ import annotations
from pathlib import Path
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from fastapi import FastAPI

__all__ = ["create_http_server"]


def create_http_server(
    project_path: str | Path | None = None,
    env: str | None = None,
    conf_source: str | None = None,
) -> FastAPI:
    """Create the HTTP server application (lazy import)."""
    from kedro.server.http_server import create_http_server as _create
    return _create(project_path=project_path, env=env, conf_source=conf_source)
```

kedro/server/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+    """Create the HTTP server application (lazy import)."""
+    from kedro.server.http_server import create_http_server as _create
```
ElenaKhaustova: **Suggestion:** When base `kedro` is installed without the `server` extra, calling `create_http_server()` raises a bare `ImportError: No module named 'fastapi'` from this line. Wrap the lazy import and re-raise with an actionable message — there's a precedent for this kind of guidance at `kedro/io/core.py:567-576`:

```python
def create_http_server(**kwargs):
    try:
        from kedro.server.http_server import create_http_server as _create
    except ImportError as exc:
        raise ImportError(
            "FastAPI is not installed. Install the server extra with "
            '`pip install "kedro[server]"` to use the HTTP server.'
        ) from exc
    return _create(**kwargs)
```

tests/server/test_http_server.py
```diff
@@ -0,0 +1,47 @@
...
+
+        assert response.status_code == 200
+        assert response.json()["status"] == "healthy"
+        assert response.json()["project_path"] == str(project_path)
+        mock_bootstrap_project.assert_called_once_with(project_path)
+
+    def test_create_http_server_raises_on_bad_project_path(self, mocker, tmp_path):
+        mocker.patch(
+            "kedro.server.http_server.get_project_path",
+            side_effect=RuntimeError("cannot resolve project"),
```
ElenaKhaustova: **Suggestion:** `get_project_path()` actually raises `ServerSettingsError` (see `kedro/server/utils.py:40,49`), not `RuntimeError`. Mocking with `RuntimeError` lets the test pass but doesn't reflect production behaviour, so a regression where `create_http_server` accidentally swallows `ServerSettingsError` wouldn't be caught here. Use the real exception type:

```python
from kedro.server.utils import ServerSettingsError
...
side_effect=ServerSettingsError("cannot resolve project"),
...
with pytest.raises(ServerSettingsError, match="cannot resolve project"):
```

Minor: the `tmp_path` fixture in the signature on line 40 is unused and can be removed.

tests/server/test_utils.py
```diff
@@ -0,0 +1,25 @@
...
+
+def test_get_project_path_raises_when_environment_variable_missing(monkeypatch):
+    monkeypatch.delenv("KEDRO_PROJECT_PATH", raising=False)
+
+    with pytest.raises(ServerSettingsError, match="is not set"):
+        get_project_path()
+
+
+def test_get_project_path_raises_when_path_does_not_exist(monkeypatch):
+    monkeypatch.setenv("KEDRO_PROJECT_PATH", "/tmp/kedro-path-that-does-not-exist")
```
ElenaKhaustova: **Suggestion:** `/tmp/kedro-path-that-does-not-exist` is non-hermetic — the path could exist on a developer's machine, which would silently flip the test to asserting the wrong branch. Prefer `tmp_path / "nonexistent"` so the path is guaranteed unique and absent:

```python
def test_get_project_path_raises_when_path_does_not_exist(tmp_path, monkeypatch):
    missing = tmp_path / "nonexistent"
    monkeypatch.setenv(KEDRO_PROJECT_PATH_ENV, str(missing))

    with pytest.raises(ServerSettingsError, match="does not exist"):
        get_project_path()
```

Also consider using the `KEDRO_PROJECT_PATH_ENV` constant from `kedro/server/utils.py` (here and on lines 9, 15) instead of the literal string `"KEDRO_PROJECT_PATH"`.

ElenaKhaustova (commented):
## Kedro PR Review

### Overview
- **Critical:** 1 | **Suggestions:** 9

- [ ] `kedro/server/http_server.py:56` — Docstring header must be `Args:` (Google style); also missing blank line before section.

- [ ] `kedro/server/utils.py:15` — `DEFAULT_HOST` / `DEFAULT_HTTP_PORT` unused; defer to CLI PR or comment intended consumer.
- [ ] `kedro/server/http_server.py:78` — `default_env` / `default_conf_source` written but unread; add a resolution-precedence test or defer these arguments to Part 2.
- [ ] `kedro/server/utils.py:27` — Broaden CLI/env-var wording to match the real contract (env var or programmatic argument); align CLI name with #5446 (`kedro server http start`).
- [ ] `kedro/server/__init__.py:1` — Add module docstring noting `pip install "kedro[server]"` (every other public `kedro/*/__init__.py` has one).
- [ ] `kedro/server/http_server.py:58` — Docstring promises a CWD fallback that doesn't exist.
- [ ] `kedro/server/__init__.py:8` — Lazy wrapper signature loses typing; mirror real signature with a `TYPE_CHECKING`-guarded `FastAPI` import.
- [ ] `kedro/server/__init__.py:10` — Friendly `ImportError` message when `kedro[server]` isn't installed (precedent: `kedro/io/core.py:567`).
- [ ] `tests/server/test_http_server.py:43` — Mock `ServerSettingsError`, not `RuntimeError`; drop unused `tmp_path`.
- [ ] `tests/server/test_utils.py:22` — Use `tmp_path / "nonexistent"` and the `KEDRO_PROJECT_PATH_ENV` constant.

### Notes
- General review: PR is correctly scoped as Part 1 of #5446. Module layout matches the design (`http_server.py`, `models.py`); `utils.py` is a sensible addition for env-var handling but contains preemptive code (host/port constants and `app.state` attrs) not exercised in this PR.
- PR checklist: confirmed with author that the `RELEASE.md` entry will land in the final PR of the series.
- API usage: no misuse of Kedro abstractions. `bootstrap_project` invoked correctly in lifespan; constants centralised in `utils.py`.
- Existing open threads (`pydantic` extra, `Literal["healthy"]`, env-var/CLI flow) — leaving those discussions to resolve in their existing threads; nothing rises to Critical there.

kedro/server/models.py
ravi-kumar-pilla: Not particular on this but in general a server responding to a request means it is reachable and functioning (may be partial unless there is some dependency issues). We can always extend if we get a use-case.

ankatiyar:
- I have updated `get_project_path()` to `_resolve_project_path(path)` which will do validation for user provided path and raise errors if the path doesn't exist.
- Updated code and tests (where relevant) wrt the feedback

kedro/server/__init__.py
ankatiyar: Added what @rashidakanchwala suggested but i'm not able to test it in a non complicated manner at the moment so i've excluded from coverage but will look into this after

kedro/server/http_server.py
ankatiyar: added this

kedro/server/utils.py
```diff
@@ -0,0 +1,54 @@
...
+
+    pass
+
+
+def _resolve_project_path(project_path: str | Path | None = None) -> Path:
+    """Resolve the Kedro project path from environment variable or function argument.
+
+        The project path is expected to be set in the environment variable `KEDRO_PROJECT_PATH`,
+        if not provided programmatically. This function validates that the path exists and is a directory.
+    Args:
```
ElenaKhaustova, line 26: **Critical:** This new docstring still has Google-style formatting problems that will break API docs rendering:

- Lines 23-24 (the description paragraph) are indented with **8 spaces** instead of 4 — they're over-indented by one level. Sphinx napoleon and mkdocstrings/Griffe will treat them as part of an arg description (or as a code block), not as the function's description paragraph.
- There's no blank line between line 24 and line 25 (`Args:`). Google-style section headers need a preceding blank line to be recognised.

Suggested fix:
```suggestion
    """Resolve the Kedro project path from environment variable or function argument.

    The project path is expected to be set in the environment variable `KEDRO_PROJECT_PATH`,
    if not provided programmatically. This function validates that the path exists and is a directory.

    Args:
```

kedro/server/http_server.py
```diff
@@ -0,0 +1,94 @@
...
+def create_http_server(
+    project_path: str | Path | None = None,
+    env: str | None = None,
+    conf_source: str | None = None,
+) -> FastAPI:
+    """Create and configure the FastAPI HTTP server.
+
+    Programmatic values passed to this factory take precedence over
+    environment-variable defaults set by the CLI.
+    Args:
```
ElenaKhaustova, line 57: **Critical:** The `Arguments:` → `Args:` rename was applied (thanks!), but the **blank-line** part of the original finding is still open. Google-style section headers need a leading blank line, otherwise they're glued to the preceding paragraph and won't render as a section.

Two spots to fix in this docstring:
- Between line 55 (`...defaults set by the CLI.`) and line 56 (`Args:`)
- Between line 60 (last `conf_source` line) and line 61 (`Returns:`)

Suggested fix for the whole docstring body:
```suggestion
    """Create and configure the FastAPI HTTP server.

    Programmatic values passed to this factory take precedence over
    environment-variable defaults set by the CLI.

    Args:
        project_path: Optional path to the Kedro project to serve. If not provided,
            it will be resolved from KEDRO_PROJECT_PATH environment variable.
        env: Optional Kedro environment to use. Overrides the KEDRO_SERVER_ENV environment variable if provided.
        conf_source: Optional configuration source to use. Overrides the KEDRO_SERVER_CONF_SOURCE environment variable if provided.

    Returns:
        Configured FastAPI application instance.
    """
```

kedro/server/utils.py
```diff
@@ -0,0 +1,54 @@
...
+    raw_path = (
+        project_path
+        if project_path is not None
+        else os.environ.get(KEDRO_PROJECT_PATH_ENV)
+    )
+
+    if not raw_path:
+        raise ServerSettingsError(
+            f"Environment variable '{KEDRO_PROJECT_PATH_ENV}' is not set or provided as arguments. "
+            "The Kedro server must be started using from within a Kedro project directory."
```
ElenaKhaustova: **Suggestion:** Two small wording issues introduced when the `'kedro server start' command` text was removed:

- Line 43 has a dangling **"using"** — leftover from the original phrasing (`"using 'kedro server start' command"`). Should read *"must be started from within a Kedro project directory."*
- Line 42: the function takes a single `project_path` argument, so *"or passed as an argument"* (singular) reads better than *"or provided as arguments"* (plural).

```suggestion
            f"Environment variable '{KEDRO_PROJECT_PATH_ENV}' is not set or passed as an argument. "
            "The Kedro server must be started from within a Kedro project directory."
```

kedro/server/utils.py
```diff
@@ -0,0 +1,54 @@
...
+        The project path is expected to be set in the environment variable `KEDRO_PROJECT_PATH`,
+        if not provided programmatically. This function validates that the path exists and is a directory.
+    Args:
+        project_path: Optional path to the Kedro project. If not provided, it will be resolved from the `KEDRO_PROJECT_PATH` environment variable.
+
+    Returns:
+        Path to the Kedro project root.
+
+    Raises:
+        ServerSettingsError: If KEDRO_PROJECT_PATH is not set.
```
ElenaKhaustova: **Suggestion:** The `Raises:` clause is incomplete. The function also raises `ServerSettingsError` when:
- the resolved path doesn't exist (line 48-52), and
- now that `project_path` can be passed as an argument too, the "is not set" condition really means "neither env var nor argument is provided".

```suggestion
    Raises:
        ServerSettingsError: If neither ``KEDRO_PROJECT_PATH`` nor ``project_path`` is
            provided, or if the resolved path does not exist.
```

tests/server/test_utils.py
```diff
@@ -0,0 +1,47 @@
...
+
+def test_get_project_path_raises_when_environment_variable_missing(monkeypatch):
+    monkeypatch.delenv(KEDRO_PROJECT_PATH_ENV, raising=False)
+
+    with pytest.raises(ServerSettingsError, match="is not set"):
+        _resolve_project_path()
+
+
+def test_get_project_path_raises_when_path_does_not_exist(monkeypatch):
+    monkeypatch.setenv(KEDRO_PROJECT_PATH_ENV, "/tmp/nonexistent_kedro_path_xyz")
```
ElenaKhaustova, line 39: **Suggestion:** Adopting the `KEDRO_PROJECT_PATH_ENV` constant addressed half of the original feedback — thanks. The remaining half is the hardcoded `/tmp/nonexistent_kedro_path_xyz` path, which is non-hermetic (a developer could happen to have that directory on their machine, in which case the test would silently flip to asserting the wrong branch). The same literal also appears on line 47, so it's worth fixing both.

Use the `tmp_path` fixture so the path is guaranteed unique and absent:

```python
def test_resolve_project_path_raises_when_path_does_not_exist(tmp_path, monkeypatch):
    missing = tmp_path / "nonexistent"
    monkeypatch.setenv(KEDRO_PROJECT_PATH_ENV, str(missing))
    with pytest.raises(ServerSettingsError, match="does not exist"):
        _resolve_project_path()


def test_resolve_project_path_raises_when_argument_path_does_not_exist(tmp_path):
    missing = tmp_path / "nonexistent"
    with pytest.raises(ServerSettingsError, match="does not exist"):
        _resolve_project_path(missing)
```

Also a tiny consistency nit: with the function renamed to `_resolve_project_path`, the test names in this file still start with `test_get_project_path_*`. Worth a sweep rename to `test_resolve_project_path_*`.

ElenaKhaustova (approved):
Thank you!

merelcht (approved):
LGTM! 👍

Issue #5445, opened 2026-03-11
Implement shared server execution runner and models in Kedro core


DimedS:
## Summary

Introduce a shared execution runner and request/response models in Kedro core that act as the common pipeline execution layer for server-based interfaces.

This is the first follow-up ticket after the [Tech Design discussion](https://github.com/kedro-org/kedro/issues/5263#issuecomment-3890423581) and the #5387  decision on built-in server support. The goal is to extract and stabilise the common execution logic currently prototyped in #5370, so it can be reused by:

- the HTTP server implemented in Kedro core
- the MCP server implemented in the `kedro-mcp` plugin

The runner should mirror the behaviour of `kedro run` by using `KedroSession.create()` and `session.run(...)`.

---

## Context

The prototype in #5370 includes both HTTP and MCP server implementations.

After the KEP discussion in #5387, the agreed direction is:

- ship a **minimal HTTP server in Kedro core**
- keep **MCP experimentation in `kedro-mcp`**
- keep Phase 1 **minimal and additive**

To support this architecture cleanly, we should introduce a shared execution layer in core that is independent of HTTP or MCP transports.

---

## Related issues

- Parent initiative: #5262 Kedro Service for Universal Deployment
- API layer issue: #5263 Build API Layer for Pipeline Execution
- Prototype implementation: #5370 Add Kedro http and mcp servers
- KEP decision: #5387 KEP-4 Built-in HTTP & MCP Server
- Related session work: #5425 KedroServiceSession v1: Stateless session

---

## Scope

Implement a shared runner module together with shared Pydantic models in Kedro core.

Proposed structure:

```
kedro/server/
 ├─ runner.py
 └─ models.py
```

### Runner

The runner should provide a transport-agnostic execution function, for example:

```python
execute_pipeline(request: RunRequest) -> RunResponse
```

The function should:

1. accept a validated request model  
2. create a new `KedroSession`  
3. call `session.run(...)`  
4. return a structured response model  

The runner must **not depend on HTTP frameworks or MCP libraries**.

### Pydantic models

Introduce shared Pydantic models that define the execution contract.

For example:

```python
class RunRequest(BaseModel):
    ...
```

```python
class RunResponse(BaseModel):
    ...
```

These models should represent execution parameters aligned with the options supported by the **`kedro run` CLI command**.

The goal is to maintain **parity with CLI execution**, so that the same parameters that can be passed via `kedro run` can also be represented in the request model.

The models should be reusable by:

[...]

PR #5370, opened 2026-02-11
Add Kedro http and mcp servers

DimedS:
Closes #5263 

## Summary

This PR introduces two built-in server interfaces for Kedro under a unified `kedro server` CLI namespace:

1. **HTTP API** (`kedro server http start`) — A REST API for triggering pipeline execution from orchestrators, dashboards, and CI/CD systems.
2. **MCP server** (`kedro server mcp start`) — A [Model Context Protocol](https://modelcontextprotocol.io/) server that exposes Kedro pipelines as tools discoverable by AI agents (Claude, Copilot, etc.).

Both are **optional, zero-cost additions** — they introduce no new hard dependencies and do not affect existing CLI or library usage.

### Why an HTTP API?

Kedro pipelines are currently triggered exclusively via the CLI (`kedro run`) or programmatic Python calls. Many production and integration scenarios require HTTP-based execution:

- **Orchestration platforms** (Airflow, Prefect, Dagster) triggering runs via REST
- **CI/CD pipelines** invoking Kedro as a service
- **Internal tooling & dashboards** that need to kick off and monitor runs
- **Microservice architectures** where Kedro acts as a data processing service

### Why an MCP server?

The [Model Context Protocol](https://modelcontextprotocol.io/) is an open standard that lets AI agents discover and invoke tools over a structured JSON-RPC interface. Adding MCP support means:

- **AI agents** (Claude Desktop, GitHub Copilot, VS Code, Claude Code) can discover pipelines, inspect their structure, and execute them without custom integration code
- **Agentic workflows** can use Kedro pipelines as building blocks — an agent can list pipelines, read node details, then run the appropriate one
- **Zero glue code** — No need to write wrapper scripts; the MCP server exposes the same `KedroSession.run()` path as the CLI

### Why in core vs. a plugin?

1. **Shared execution engine** — Both server interfaces call the exact same `KedroSession.create() → session.run()` path as the CLI. Keeping them in core ensures they stay in sync with CLI changes automatically.
2. **First-class citizen** — Server support should feel like a native Kedro capability, not an afterthought.
3. **Discoverability** — `kedro server http start` / `kedro server mcp start` is easier to find than `pip install kedro-server && kedro server start`.
4. **Zero cost if unused** — FastAPI, Uvicorn, and the MCP SDK are optional extras. Users who don't need them pay nothing.

### Design Principle: "One execution engine, multiple callers"

```
kedro run --pipeline=X             →  KedroSession.create() → session.run()
[...]

PR #5521, opened 2026-04-28
Part 2: Add run endpoint to HTTP Server

ankatiyar:
## Description

Part of #5446 

This is built on top of #5520 and will be added to `main` after that

This PR adds the run endpoint for executing pipelines and related data models to the server module.

## Development notes

Some things to note (and would like opinions on the design of):
- The server API layer is designed solely for `KedroServiceSession`. The use of `SESSION_CLASS` in `settings.py` will have no effect. 
- Since `project_path`, `env` and `conf_source` go into session creation, I've added them as arguments (env and conf-source, not project_path) for `kedro server start`. These are set as env variables (along with project path) and read by the server API module. 
- `KedroServiceSession.create()` is called only at the first time run request comes through, then is assigned to the `app.state` --> this is so when being used for inspection, a session is not created simply when the FastAPI app is spun up. 

**UPDATE**
- Based on feedback, have removed the extra `PipelineExecutionError` and `PipelineExecutionResult` dataclasses and keep it only to `RunRequest` and `RunResponse` (also `ErrorDetails` which is a part of the response.
- Made the run request endpoint synchronous, use a lock for session creation (as suggested by @ElenaKhaustova)
- FastAPI runs sync handlers in a thread-pool executor, so two simultaneous `POST /run` requests will execute concurrently. The `session_lock` guards only session creation, not session execution. (Flagged by co-pilot). I don't want to complicate this implementation too much for now with concurrency till we go back to `KedroServiceSession` and update that.
- Added warning when `SESSION_CLASS` is not `KedroServiceSession`
- Made docstring etc changes suggested

## Checklist


kedro/server/models.py
```diff
@@ -2,11 +2,126 @@
...
+    message: str
+    traceback: list[str] | None = None
+
+
+@dataclass
+class PipelineExecutionResult:
+    """Result of a pipeline execution attempt."""
+
+    run_id: str
+    status: Literal["success", "failure"]  # "success" | "failed"
```
merelcht: Nit: comment seems a bit unnecessary

kedro/server/models.py
```diff
@@ -2,11 +2,126 @@
...
+    duration_ms: float
+    error: PipelineExecutionError | None = None
+
+    def to_dict(self) -> dict[str, Any]:
```
merelcht: Why is this needed?

kedro/server/models.py
```diff
@@ -2,11 +2,126 @@
...
+        default=None,
+        description="Construct the pipeline using only nodes which have this tag attached.",
+    )
+    load_versions: dict[str, str] | None = Field(
+        default=None,
+        description="Specify a particular dataset version (timestamp) for loading.",
+    )
+    pipelines: list[str] | None = Field(
+        default=None,
+        description="List of registered pipeline names to run. Cannot be used together with 'pipeline'.",
```
merelcht: Looks like this is an outdated description, since `pipeline` isn't an option.

kedro/server/models.py
```diff
@@ -2,11 +2,126 @@
...
+    )
+    namespaces: list[str] | None = Field(
+        default=None,
+        description="Run only nodes within these namespaces.",
+    )
+    params: dict[str, Any] | None = Field(
+        default=None,
+        description="Extra parameters to pass to the context initialiser.",
```
merelcht: What's the context initialiser?

tests/server/test_http_server.py
```diff
@@ -19,6 +25,63 @@ def test_lazy_import_wrapper(self, mocker):
...
+            tags=["train", "daily"],
+            node_names=["node_a"],
+            from_nodes=["node_a"],
+            to_nodes=["node_b"],
+            from_inputs=["raw"],
+            to_outputs=["model"],
+            load_versions={"cars": "2025-01-01T00.00.00.000Z"},
+            namespaces=["ns"],
+            only_missing_outputs=True,
+            params={"alpha": 1},
```
merelcht: Do you have to provide all of these values? If not, might be worth having a test that's just passing all mandatory arguments.

tests/server/test_http_server.py
```diff
@@ -45,3 +108,221 @@ def test_create_http_server_raises_on_bad_project_path(self, mocker, tmp_path):
...
+        assert payload["duration_ms"] == 12.34
+        assert payload["error"] == {
+            "type": "ValueError",
+            "message": "bad run",
+            "traceback": ["Traceback line"],
+        }
+
+
+class TestModels:
+    def test_pipeline_execution_result_to_dict_omits_error_when_none(self):
```
merelcht: I was a bit confused by the test name and content, it seems that here you don't get an error because the result is success and not because it's none?

tests/server/test_http_server.py
```diff
@@ -45,3 +108,221 @@ def test_create_http_server_raises_on_bad_project_path(self, mocker, tmp_path):
...
+            duration_ms=7.1,
+            error=PipelineExecutionError(
+                type="RuntimeError", message="failed", traceback=["line 1", "line 2"]
+            ),
+        )
+
+        as_dict = result.to_dict()
+
+        assert as_dict["error"]["traceback"] == ["line 1", "line 2"]
+        with pytest.raises(RuntimeError, match="cannot resolve project"):
```
ravi-kumar-pilla: what is this for ?

kedro/server/http_server.py
```diff
@@ -90,4 +101,160 @@ async def health() -> HealthResponse:
...
+            status=result.status,
+            duration_ms=result.duration_ms,
+            error=error_detail,
+        )
+        return res
+
     return app
+
+
+def execute_pipeline(  # noqa: PLR0913
```
ravi-kumar-pilla: [Nit] Should this be a private method ?

kedro/server/http_server.py
```diff
@@ -90,4 +101,160 @@ async def health() -> HealthResponse:
...
+    from_inputs: list[str] | None = None,
+    to_outputs: list[str] | None = None,
+    load_versions: dict[str, str] | None = None,
+    namespaces: list[str] | None = None,
+    only_missing_outputs: bool = False,
+) -> PipelineExecutionResult:
+    """Execute a Kedro pipeline and return a structured result.
+
+    This is the shared execution core used by both the HTTP and MCP
+    server interfaces.  It creates a fresh ``KedroServiceSession``, resolves
```
ravi-kumar-pilla: I think session is already created at the endpoint, we can remove `It creates a fresh ``KedroServiceSession``` ?

kedro/server/http_server.py
```diff
@@ -90,4 +101,160 @@ async def health() -> HealthResponse:
...
+    the runner, calls ``session.run()``, and returns a
+    :class:`PipelineExecutionResult`.
```
ravi-kumar-pilla: Not sure how docstrings will be rendered in mkdocs with :class: .... In Kedro we have not used `:class: ` pattern before or I might be wrong

kedro/server/models.py
```diff
@@ -2,11 +2,126 @@
...
+
+class RunResponse(BaseModel):
+    """Response model for pipeline execution."""
+
+    run_id: str = Field(description="Unique identifier for this pipeline run.")
+    status: Literal["success", "failure"] = Field(description="Run status.")
+    duration_ms: float = Field(description="Total execution time in milliseconds.")
+    error: ErrorDetail | None = Field(
+        default=None,
+        description="Error details if status is 'failed'.",
```
ravi-kumar-pilla:
```suggestion
        description="Error details if status is 'failure'.",
```

ravi-kumar-pilla: [Nit] I would prefer run status to be `success` and `failed` instead

kedro/server/http_server.py
```diff
@@ -90,4 +101,160 @@ async def health() -> HealthResponse:
...
+            exc_info=True,
+        )
+
+        error = PipelineExecutionError(
+            type=type(exc).__qualname__,
+            message=str(exc),
+            traceback=traceback.format_tb(exc.__traceback__),
+        )
+
+        return PipelineExecutionResult(
```
ravi-kumar-pilla: So even if the pipeline run fails, the endpoint returns a HTTP status code of 200 but with a different PipelineExecutionResult ? Should we consider returning a non 200 status code on failure ?

kedro/server/http_server.py
ankatiyar: Updated

kedro/server/http_server.py
ankatiyar: Removed this

kedro/server/http_server.py
ankatiyar: Yeah, right now the assumption is the HTTP request succeeded so the result is in the body. This is from the original design by Dmitry - would like opinions from the reviewers on if we should map some pipeline failures to HTTPExceptions?

tests/server/test_http_server.py
ankatiyar: removed

kedro/server/http_server.py
```diff
@@ -4,26 +4,37 @@
...
     KEDRO_SERVER_ENV,
     _resolve_project_path,
 )
+from kedro.utils import load_obj
 
 if TYPE_CHECKING:
     from collections.abc import AsyncGenerator
     from pathlib import Path
 
+    from .abstract_session import AbstractSession
```
ElenaKhaustova: **Critical:** This relative import resolves to `kedro/server/abstract_session.py`, which doesn't exist — the class actually lives in `kedro/framework/session/abstract_session.py`. Mypy is silenced by `ignore_missing_imports = true`, so it doesn't fail loudly, but the annotation on `_execute_pipeline(session: AbstractSession, ...)` silently degrades to `Any`, and pyright/IDEs will flag the file as unresolved.

Suggested fix:
```suggestion
    from kedro.framework.session.abstract_session import AbstractSession
```

kedro/server/http_server.py
```diff
@@ -92,4 +103,160 @@ async def health() -> HealthResponse:
             project_path=str(project_path),
         )
 
+    @app.post("/run", response_model=RunResponse, tags=["pipeline"])
+    async def run_pipeline(request: RunRequest) -> RunResponse:
```
ElenaKhaustova: **Suggestion:** `run_pipeline` is `async def` but `_execute_pipeline` (and the underlying `session.run` → runner → dataset I/O) is fully synchronous and blocking. FastAPI runs `async def` handlers directly on the event loop, so the issue is that a long pipeline run blocks every other request — including `/health`. Operators reaching for `/health` to check liveness during a run will see the server appear dead.

Two ways to fix:
- Make the handler synchronous: `def run_pipeline(...)` — FastAPI auto-runs it in a threadpool.
- Or keep `async def` and wrap the blocking call: `result = await fastapi.concurrency.run_in_threadpool(_execute_pipeline, session=app.state.session, ...)`.

The first option is simpler and matches Phase 1's "synchronous, single-worker" intent.

kedro/server/http_server.py
```diff
@@ -92,4 +103,160 @@ async def health() -> HealthResponse:
...
+        This endpoint mirrors the behavior of `kedro run` CLI command.
+        First run request creates a new KedroServiceSession, and subsequent requests reuse it.
+        Args:
+            request: Pipeline execution parameters.
+
+        Returns:
+            RunResponse with run_id, status, duration, and optional error details.
+        """
+
+        # create a session and assign to app state if not already created
```
ElenaKhaustova: **Suggestion (nit):** The comment narrates the code without adding intent — the next line literally reads `if not hasattr(app.state, "session"): app.state.session = ...`. Same flavour Merel flagged on `models.py`. Drop it.

kedro/server/http_server.py
```diff
@@ -92,4 +103,160 @@ async def health() -> HealthResponse:
...
+        if not hasattr(app.state, "session"):
```
ElenaKhaustova: **Suggestion:** This `hasattr` check isn't concurrency-safe. Two simultaneous initial `POST /run` requests can both pass it before either assignment, so two `KedroServiceSession`s get created — each runs `validate_settings()`, builds its own hook manager, and registers entry-point hooks; one then gets orphaned. Phase 1 is single-worker / sequential per design, so this is a corner case rather than a daily-driver bug — but the lock is two lines and removes the ambiguity entirely.

Suggested approach: initialize a lock once (e.g. in `lifespan` or `create_http_server`) and guard creation:

```python
async with app.state.session_lock:
    if not hasattr(app.state, "session"):
        app.state.session = KedroServiceSession.create(
            project_path=app.state.project_path,
            env=app.state.default_env,
            conf_source=app.state.default_conf_source,
        )
```

ElenaKhaustova (commented):
## Kedro PR Review

### Overview
- Critical: 1 | Suggestions: 11

- [ ] `kedro/server/http_server.py:37` — TYPE_CHECKING import resolves to a non-existent module.

- [x] `kedro/server/http_server.py:120` — Lazy session creation can race on simultaneous initial requests.
- [x] `kedro/server/http_server.py:107` — `async def` handler runs blocking work on the event loop.
- [x] `kedro/server/http_server.py:121` — `SESSION_CLASS` from `settings.py` is silently ignored.
- [x] `kedro/server/http_server.py:125` — Session attached to `app.state` is never closed on shutdown.
- [x] `kedro/server/http_server.py:119` — Redundant inline comment.
- [x] `kedro/server/http_server.py:182` — Docstring claims "shared by HTTP and MCP" but lives inside `http_server.py`; also says `KedroServiceSession` while the param is `AbstractSession`.
- [x] `kedro/server/http_server.py:188` — Stale references to a non-existent `*pipeline*` parameter.
- [x] `kedro/server/http_server.py:210` — `run_id: str | None` is dead-branch typing; `generate_timestamp()` always returns `str`.
- [x] `kedro/server/models.py:30` — Per-request `env` / `conf_source` design question — recommendation inline.
- [x] `kedro/server/models.py:57` — Consider `Literal[...]` for `runner` to document allowed values in OpenAPI.
- [x] `tests/server/test_http_server.py:545` — Endpoint test misplaced inside `TestExecutePipeline` and duplicates earlier coverage.

### Notes
- Release-notes and user-facing docs are deferred to the final PR in this series, per the author.
- Out of scope: integration tests covering the `/run` endpoint end-to-end (#5446 AC) — reasonable to land alongside the CLI/server-startup PR.

kedro/server/http_server.py
```diff
@@ -92,4 +103,160 @@ async def health() -> HealthResponse:
...
+            app.state.session = KedroServiceSession.create(
```
ElenaKhaustova: **Suggestion:** The server hardcodes `KedroServiceSession`, which silently ignores `settings.SESSION_CLASS`. The issue is that existing projects subclassing the session class (e.g. for telemetry / custom hooks) will get the default here while their `kedro run` invocations honour the override — same `settings.py`, two different runtime classes, no signal.

The PR description acknowledges this trade-off; that's fine for Phase 1, but please make it visible:
- Log a `WARNING` at server start if `settings.SESSION_CLASS` is non-default ("HTTP server hardcodes `KedroServiceSession`; the configured `SESSION_CLASS` will be ignored").
- Call this out clearly in the docs that land in the final PR of the series.

Even better (if cheap): honour `settings.SESSION_CLASS` if it is a subclass of `AbstractSession`, falling back to `KedroServiceSession`.

kedro/server/http_server.py
```diff
@@ -92,4 +103,160 @@ async def health() -> HealthResponse:
...
+                project_path=app.state.project_path,
+                env=app.state.default_env,
+                conf_source=app.state.default_conf_source,
+            )
```
ElenaKhaustova: **Suggestion (related to the session creation above):** The session attached to `app.state.session` here is never closed on shutdown. `KedroServiceSession.close()` exists and is part of the abstract contract — even if it currently just logs, the contract is "if you create one, close it", and other implementations may rely on it.

Suggested cleanup in the existing `lifespan` function:

```python
yield
session = getattr(app.state, "session", None)
if session is not None:
    session.close()
logger.info("Kedro server shutting down")
```

kedro/server/http_server.py
```diff
@@ -92,4 +103,160 @@ async def health() -> HealthResponse:
...
+    from_inputs: list[str] | None = None,
+    to_outputs: list[str] | None = None,
+    load_versions: dict[str, str] | None = None,
+    namespaces: list[str] | None = None,
+    only_missing_outputs: bool = False,
+) -> PipelineExecutionResult:
+    """Execute a Kedro pipeline and return a structured result.
+
+    This is the shared execution core used by both the HTTP and MCP
+    server interfaces.  It receives a `KedroServiceSession` as input , resolves
```
ElenaKhaustova: **Suggestion:** Two issues in this docstring:

1. It claims `_execute_pipeline` is "the shared execution core used by both the HTTP and MCP server interfaces", but the function lives inside `http_server.py` (which imports `fastapi`). Anything that wants to "share" it transitively pulls in FastAPI — so the docstring's claim isn't actually delivered. The original spec in #5445 / #5446 placed this in a separate transport-agnostic `kedro/server/runner.py` for exactly this reason. Given KEP-4 puts MCP in a plugin, the only consumer in core is HTTP — so either move this to `kedro/server/runner.py` (matches #5446's proposed file structure) or trim the docstring claim.

2. The parameter is annotated `AbstractSession`, not `KedroServiceSession`, so the prose is incorrect. There's also a typo `"input ,"` (space before comma).

kedro/server/http_server.py
```diff
@@ -92,4 +103,160 @@ async def health() -> HealthResponse:
...
+    the runner, calls ``session.run()``, and returns a
+    `PipelineExecutionResult`.
+
+    Args:
+        session: Kedro service session to use for pipeline execution.
+        pipeline_names: List of pipeline names (mutually exclusive with
```
ElenaKhaustova: **Suggestion:** This text references a `pipeline` parameter that doesn't exist on `_execute_pipeline` (only `pipeline_names`) — leftover from the prototype #5370 which had both forms. Suggested rewrite:

```
pipeline_names: Names of registered pipelines to run. If None,
    `KedroServiceSession.run()` defaults to ``__default__``.
```

kedro/server/http_server.py
```diff
@@ -92,4 +103,160 @@ async def health() -> HealthResponse:
...
+        load_versions: Pin specific dataset versions (name → timestamp).
+        namespaces: Run only nodes within these namespaces.
+        only_missing_outputs: Skip nodes whose outputs already exist.
+
+    Returns:
+        A `PipelineExecutionResult` with run id, status, duration,
+        and optional error details.
+    """
+    start_time = time.perf_counter()
+    run_id: str | None = generate_timestamp()
```
ElenaKhaustova: **Suggestion:** Typing `run_id` as `str | None` here is misleading — `generate_timestamp()` (in `kedro/io/core.py`) is annotated `-> str` and always returns a string, so the `None` branch is unreachable. As a result, the `run_id or "unknown"` fallbacks at lines 236, 246, 258 are dead branches that just hide intent.

```suggestion
    run_id: str = generate_timestamp()
```

…and drop every `run_id or "unknown"` downstream — just pass `run_id`.

kedro/server/models.py
```diff
@@ -2,11 +2,115 @@
...
+class PipelineExecutionResult:
+    """Result of a pipeline execution attempt."""
+
+    run_id: str
+    status: Literal["success", "failure"]
+    duration_ms: float
+    error: PipelineExecutionError | None = None
+
+
+class RunRequest(BaseModel):
```
ElenaKhaustova, line 10: **Design question:** `RunRequest` doesn't expose `env` / `conf_source` — they're resolved at `kedro server start` and reused for every request, so a single server process is locked to one `(env, conf_source)`. That's a behavioural change vs. `kedro run`, where `--env` is per-invocation. I think this is the right call for Phase 1, but it's worth being explicit so the constraint doesn't bite users.

Recommendation: keep it process-scoped because

- It matches how the HTTP server is deployed (one server per environment, like any other production service).
- It aligns with `KedroServiceSession`'s long-lived intent (#5425): create once, reuse. Caching by `(env, conf_source)` would mean N hook managers in one process — not something plugin authors expect.
- The example `{"env": "base", ...}` in #5446 predates #5425 and is illustrative, not normative.

If you go this route, please make the constraint visible:

1. Log the resolved `env` and `conf_source` at startup.
2. Update `kedro server start --env` / `--conf-source` help text to say it's process-scoped.
3. Mention server-scoping in the `/run` / `RunRequest` description so it shows up in `/docs`.
4. Set `model_config = ConfigDict(extra="forbid")` on `RunRequest` so a request with `{"env": "..."}` gets a 422 instead of a silent ignore.

If maintainers want CLI parity later, optional `env` / `conf_source` fields can be added forward-compatibly without breaking anyone.

kedro/server/models.py
```diff
@@ -2,11 +2,115 @@
...
+    )
+    to_nodes: list[str] | None = Field(
+        default=None,
+        description="A list of node names which should be used as an end point.",
+    )
+    node_names: list[str] | None = Field(
+        default=None,
+        description="Run only nodes with specified names.",
+    )
+    runner: str | None = Field(
```
ElenaKhaustova, line 38: **Suggestion:** As a free-form `str`, OpenAPI gives clients no hint about valid values — `/docs` and any generated client just say "string", even though there are exactly three accepted values. The list is small and stable:

```python
runner: Literal["SequentialRunner", "ParallelRunner", "ThreadRunner"] | None = Field(
    default=None,
    description="Runner to use.",
)
```

If you want to allow custom runners loaded via dotted import paths (since `_execute_pipeline` uses `load_obj`), keep `str | None` but add "or any importable subclass of `AbstractRunner`" to the description. Either way — pick one shape and document it.

tests/server/test_http_server.py
```diff
@@ -50,13 +87,515 @@ def test_health_endpoint_with_lifespan(self, mocker, tmp_path):
...
+        )
+
+        result = _execute_pipeline(session=mock_session, node_names=["node1", "node2"])
+
+        assert result.status == "success"
+        call_kwargs = mock_session.run.call_args[1]
+        assert call_kwargs["node_names"] == ("node1", "node2")
+        assert isinstance(call_kwargs["node_names"], tuple)
+
+    def test_run_endpoint_passes_request_parameters(self, mocker, tmp_path):
```
ElenaKhaustova: **Suggestion (nit):** This test is in the wrong class — it lives inside `TestExecutePipeline` but actually drives `/run` via `TestClient`, so it tests the endpoint, not the pure `_execute_pipeline` function. It also overlaps with `TestRunEndpoint.test_run_endpoint_passes_parameters_to_execute_pipeline` (same scenario, fewer parameters).

Suggested cleanup: move it into `TestRunEndpoint` and either replace the existing test with this fuller one, or merge them via `@pytest.mark.parametrize` on the request payload.

kedro/server/models.py
ankatiyar: I kept the `str | None` but added the suggested change to the description

kedro/server/models.py
ankatiyar: I added it to the `RunRequest` description and it will be documented - is the logging super necessary? 
The add it in the description for the CLI as well in the next PR. 
Added the forbid extra bit.

kedro/server/http_server.py
ankatiyar: Did the first option

kedro/server/http_server.py
ankatiyar: Added a warning. The discrepancy in the arguments of `create()` vs `run()` would make it annoying to cater to both.

kedro/server/http_server.py
ankatiyar: Added a close

kedro/server/http_server.py
ankatiyar: Removed this entirely, using `KedroServiceSession` in the `_execute_pipeline()` type checking

kedro/server/http_server.py
ankatiyar: Made the endpoint synchronous so using `threading.Lock()`

kedro/server/http_server.py
ankatiyar: Updated docstring and annotated parameter `KedroServiceSession`

kedro/server/models.py
ElenaKhaustova: Nope, feel free to skip it

kedro/server/http_server.py
```diff
@@ -92,4 +110,100 @@ async def health() -> HealthResponse:
             project_path=str(project_path),
         )
 
+    @app.post("/run", response_model=RunResponse, tags=["pipeline"])
+    def run_pipeline(request: RunRequest) -> RunResponse:
+        """Execute a Kedro pipeline.
```
ElenaKhaustova, line 115: **Suggestion (nit):** Google-style docstrings need a blank line after the one-line summary, and another blank line before the `Args:` block — without them, mkdocstrings / Sphinx-google render the whole pre-`Args:` text as one paragraph. The `_execute_pipeline` docstring just below has this right; this one's the only outlier.

```suggestion
        """Execute a Kedro pipeline.

        This endpoint accepts a `RunRequest` with parameters for pipeline execution,
        runs the pipeline using the session, and returns a `RunResponse` with the result.
        First run request creates a new `KedroServiceSession`, and subsequent requests reuse it.

        Args:
            request: Pipeline execution parameters.

        Returns:
            `RunResponse` with run_id, status, duration, and optional error details.
        """
```

ElenaKhaustova (approved):
Thank you for your hard work 💪

Left one nit comment, the rest looks good to go.

ravi-kumar-pilla (approved):
Awesome work !! @ankatiyar thank you

rashidakanchwala (approved):
LGTM. thanks @ankatiyar

PR #5522, opened 2026-04-28
Part 3: Add CLI command for HTTP Server

ankatiyar:
## Description

Third PR for  https://github.com/kedro-org/kedro/issues/5446

The complete code is in https://github.com/kedro-org/kedro/pull/5501 for testing

This PR adds the CLI command for starting the http server

## Development notes

Some things to note (and would like opinions on the design of):

- I've changed the original prototype CLI command from kedro server http start to kedro server start (it was originally two layered because of kedro server mcp start) - does it make sense to keep it two layered and see if mcp plugin commands can plug into it?
- Since project_path, env and conf_source go into session creation, I've added them as arguments (env and conf-source, not project_path) for kedro server start. These are set as env variables (along with project path) and read by the server API module. This is to not load the module when commands are loaded (lazy loading) to keep things optional.

## Checklist


kedro/framework/cli/server.py
```diff
@@ -0,0 +1,115 @@
...
+    click.echo()
+
+    # Start uvicorn programmatically
+    uvicorn.run(
+        "kedro.server.http_server:create_http_server",
+        factory=True,
+        host=host,
+        port=port,
+        reload=reload,
+        log_level="info",
```
ravi-kumar-pilla, line 114:
```suggestion
        log_level="info",
        reload_dirs=[str(project_path)],  # watch the Kedro project root
```

kedro/framework/cli/server.py
```diff
@@ -0,0 +1,115 @@
...
+    reload: bool,
+    env: str | None,
+    conf_source: str | None,
+) -> None:
+    """Start the Kedro HTTP server.
+
+    The server exposes pipeline execution via HTTP endpoints, allowing
+    external systems to trigger Kedro pipelines programmatically.
+
+    \b
```
ravi-kumar-pilla, line 71: are the break tags intentional ?

ravi-kumar-pilla (approved):
LGTM. I would like to test end-end once everything is merged but the code looks good. Thank you @ankatiyar

kedro/framework/cli/server.py
```diff
@@ -0,0 +1,115 @@
...
+    default=None,
+    help="Kedro configuration environment for server sessions.",
+)
+@click.option(
+    "--conf-source",
+    callback=validate_conf_source,
+    help="Path of a directory where project configuration is stored.",
+)
+@click.pass_obj
+def http_start(  # noqa: PLR0913
```
merelcht: Maybe rename this to `start_server` instead to be consistent with the CLI command the user runs.

kedro/framework/cli/server.py
```diff
@@ -0,0 +1,115 @@
...
+      POST /run     - Execute a pipeline
+
+    \b
+    Examples:
+      kedro server start
+      kedro server start --host 0.0.0.0 --port 8080
+      kedro server start --reload
+    """
+    try:
+        import uvicorn
```
merelcht, line 84: We check for `uvicorn` here but not `fastapi`, shouldn't both be checked?

kedro/framework/cli/server.py
```diff
@@ -0,0 +1,115 @@
...
+    project_path = metadata.project_path
+
+    # Set environment variables for the server process
+    os.environ[KEDRO_PROJECT_PATH_ENV] = str(project_path)
+
+    if env:
+        os.environ[KEDRO_SERVER_ENV] = env
+
+    if conf_source:
+        os.environ[KEDRO_SERVER_CONF_SOURCE] = conf_source
```
merelcht, line 105: I think the review agent you applied on the first PR also flagged this, but when `--env` or `--conf-source` are omitted, their env vars are simply not set, but if they were set by a previous invocation in the same process, those stale values silently override the user's intent.

kedro/framework/cli/server.py
merelcht: This new command should also be added to https://docs.kedro.org/en/1.3.1.post1/getting-started/commands_reference/

kedro/framework/cli/server.py
ankatiyar: I added this!

kedro/framework/cli/server.py
ankatiyar: Turns out the line breaks are intentional, otherwise it doesn't render properly in docs

kedro/framework/cli/server.py
ankatiyar: Done!

kedro/framework/cli/server.py
ankatiyar: I updated the error message to also say `pydantic` - though it is a dependency for `fastapi` anyway.
Originally hadn't done it because the command only needs `uvicorn` to run but `fastapi` is imported in `create_http_server` but it makes sense to fail now if the package is not installed.
Other commands will run fine without either packages and imports don't happen till the command is called. The only thing is now `fastapi` is imported without cause here.

kedro/framework/cli/server.py
ankatiyar: I added some code to delete the env variables before adding fresh ones, if `env` or `conf-source` is not provided when the command is called and the env variables exist, they're deleted

kedro/framework/cli/server.py
ankatiyar: This is added automatically by mkdocs

kedro/framework/cli/server.py
```diff
@@ -0,0 +1,120 @@
...
+      kedro server start --reload
+    """
+    try:
+        import fastapi  # noqa: F401
+        import uvicorn
+
+    except ImportError as exc:
+        raise KedroCliError(
+            "Kedro HTTP server requires 'fastapi', 'pydantic', and 'uvicorn' packages. "
+            "Install them with: pip install kedro[server]"
```
ElenaKhaustova: **Suggestion:** Quote `kedro[server]` in the install hint so the command is portable across shells. In zsh and fish, unquoted brackets are interpreted as a glob pattern and `pip install kedro[server]` fails with `no matches found: kedro[server]`. The companion message in `kedro/server/__init__.py` already uses the quoted form — they should match:

```suggestion
            "Kedro HTTP server requires 'fastapi', 'pydantic', and 'uvicorn' packages. "
            "Install them with: pip install 'kedro[server]'"
```

ElenaKhaustova (commented):
## Kedro PR Review

### Overview
- **Critical:** 0 | **Suggestions:** 2 | **Heads-ups:** 1

Overall the design is sound — the `factory=True` + env-var hand-off is the correct pattern to survive `--reload` (kwargs wouldn't carry across uvicorn's reload subprocesses), the lazy `fastapi`/`uvicorn` imports keep the optional extra truly optional, and the explicit "unset before set" of `KEDRO_SERVER_ENV` / `KEDRO_SERVER_CONF_SOURCE` correctly prevents stale values from leaking between invocations in the same shell. RELEASE.md is updated and tests cover the happy path, missing-deps, and stale-env-var cases.

### Suggestions
- [ ] `kedro/framework/cli/server.py:L89` — quote `kedro[server]` in the install hint; unquoted brackets are interpreted as a glob in zsh/fish.
- [ ] `kedro/framework/cli/server.py:L107-L108` — the two `Starting…` echoes plus uvicorn's own log produce three near-identical lines; consider collapsing.

### Heads-up
- [ ] `kedro/framework/cli/server.py` — flagging the `kedro server start` vs `kedro server http start` namespacing question explicitly so it doesn't get lost (you raised it in the PR description).

### Notes
- **General review:** Single-purpose PR, scoped well. Tests are meaningful and cover the behavior introduced here. Docstrings present and use Click `\b` for proper rendering.
- **PR checklist:** RELEASE.md entry added under the right section. Docs page at `docs/getting-started/commands_reference.md` uses `mkdocs-click` to autogenerate the project commands reference, so the new `server` group is picked up automatically — confirmed.
- **API usage:** Uses standard Kedro CLI primitives (`CONTEXT_SETTINGS`, `KedroCliError`, `validate_conf_source`, `@click.pass_obj` for `ProjectMetadata`, registered via the existing `LazyGroup` in `cli.py`). The existing comments from @merelcht (rename `server_start` → `start_server`, `fastapi` import check, stale env-var bug, docs ref autogen) and @ravi-kumar-pilla (`reload_dirs`, line breaks in the docstring) have all been addressed or replied to and aren't repeated here.

kedro/framework/cli/server.py
```diff
@@ -0,0 +1,120 @@
...
+
+    os.environ[KEDRO_PROJECT_PATH_ENV] = str(project_path)
+    if env:
+        os.environ[KEDRO_SERVER_ENV] = env
+
+    if conf_source:
+        os.environ[KEDRO_SERVER_CONF_SOURCE] = conf_source
+
+    click.echo(f"Starting Kedro HTTP server for project at: {project_path}")
+    click.echo(f"Starting server at: http://{host}:{port}")
```
ElenaKhaustova: **Suggestion:** These two lines both start with "Starting" and uvicorn itself logs a similar `Uvicorn running on http://…` right after, so the user ends up reading three near-identical messages. Consider collapsing into one informative line, e.g.:

```suggestion
    click.echo(f"Starting Kedro HTTP server at http://{host}:{port} (project: {project_path})")
    click.echo("Press CTRL+C to stop")
```

Non-blocking — purely cosmetic.

kedro/framework/cli/server.py
```diff
@@ -0,0 +1,120 @@
...
+    DEFAULT_HOST,
+    DEFAULT_HTTP_PORT,
+    KEDRO_PROJECT_PATH_ENV,
+    KEDRO_SERVER_CONF_SOURCE,
+    KEDRO_SERVER_ENV,
+)
+
+
+@click.group(name="server")
+def server_cli() -> None:
```
ElenaKhaustova, line 21: **Heads-up (no action needed if the answer is "no MCP"):** You raised this in the PR description and I want to flag it explicitly so it doesn't get lost. Shipping `kedro server start` as a flat command is fine *if* we don't plan to add other server protocols (e.g. an MCP server) under the same namespace. If we do, reorganising later into `kedro server http start` / `kedro server mcp start` would be a breaking change for anyone who scripted the flat form — the two-layer form costs nothing now and avoids deprecation pain later. Worth a maintainer call before merging; happy either way, just don't want this decided by default.

ankatiyar:
@ElenaKhaustova 
I've actually removed the click logging we were doing from the command - The http server module already logs the `project_path` at startup and `uvicorn` logs the host, port and "press Ctrl+C to quit" bit. We don't need to print it at all.

Briefly discussed with @rashidakanchwala about the mcp server command - It's not on the immediate agenda so I've kept the CLI command to be simple and one level. I feel like we still have some options if we were to build the mcp server command (eg Keep `kedro server start` default to HTTP but add `kedro server start --http` or `kedro server start --mcp`).

ElenaKhaustova (approved):
👍

RELEASE.md
```diff
@@ -4,6 +4,7 @@
 * Add official support for Python 3.14.
 * Added the first iteration of the `KedroServiceSession`, a new session implementation that allows for multiple runs and data injection.
 NOTE: This session implementation is under active development and may occasionally contain bugs or breaking changes. We encourage users to try it out and share their feedback with us.
+* Added a new HTTP Server layer which uses the `KedroServiceSession` to execute pipelines from HTTP requests, and exposes a new CLI command `kedro server start` to run the server.
```
merelcht: Maybe make this two separate bullets to make it clear they are essentially two features?

merelcht (approved):
Thanks @ankatiyar! This looks good 👍

PR #5542, opened 2026-05-06
Fix parameter validation for optional None

PragnyaKhandelwal:
## Description
Fix parameter validation for optional typed params so explicit `None` values are preserved and do not trigger model instantiation/validation errors. This aligns behavior with `Optional[Model]` / `Model | None` annotations.

### Changes:
- **Preserved Optional types in TypeExtractor** — Store the full `Optional[X]` type (not unwrapped) so `_apply_validation` can determine if `None` is allowed
- **Introduced _MISSING sentinel in resolve_nested_dict_path** — Return `_MISSING` for missing keys instead of `None`, allowing callers to distinguish "key absent" from "key present with None value"
- **Implemented 3-step validation logic in _apply_validation**:
  1. Skip missing keys (raw_value is `_MISSING`)
  2. For `None` values: allow only if type is `Optional`, else raise `ParameterValidationError`
  3. Unwrap `Optional` wrapper before instantiating non-`None` values
- **Added integration test** for `Optional[Model] = None` through `validate_raw_params()` for full extraction → validation path coverage

### Tests:
- **91 validation tests pass** 
- **100% code coverage** across all validation modules 
- Test coverage includes:
  - `Optional[Model] = None` → validates without error
  - `Model = None` (non-optional) → raises `ParameterValidationError`
  - Missing parameters → silently skipped
  - Valid non-None values → instantiated and validated

### API Changes:
- `resolve_nested_dict_path()` now returns `_MISSING` (internal sentinel) instead of `None` for missing keys. This is a behavior change for direct callers. See RELEASE.md for details.

All commits are signed off with DCO.

## Checklist


PragnyaKhandelwal:
Hi @ravi-kumar-pilla, @rashidakanchwala ....CI has been re-run for this branch but one check requires manual approval. Could you please approve the pending check so merge-gatekeeper can complete? Thanks!

ravi-kumar-pilla:
Hi @PragnyaKhandelwal ,

Thank you for the contribution. There is a parallel work on Parameter validation which modifies few things here. We would like to take this up once the ongoing [PR](https://github.com/kedro-org/kedro/pull/5494) is merged.

ravi-kumar-pilla:
Hi @PragnyaKhandelwal , 

In the meantime, 

1. Please make sure the file changes correspond to the fix (as I see some unrelated changes).
2. Add more description on why this change is important
3. Add quick testing guide for reviewers to test this change

Thank you

PragnyaKhandelwal:
Thanks for the feedback @ravi-kumar-pilla!
I've cleaned up the branch. It now only includes the core logic changes and relevant tests (the "Files Changed" count is down to the 6 relevant files).

**Why this matters:**
Currently, passing an explicit `None` to an `Optional` parameter triggers a validation crash. This fix ensures Kedro respects Python’s native type hints, which is critical for disabling features in ML pipelines.

**Testing Guide:**
1. `pip install -e .`
2. Run the new validation tests: `pytest tests/validation/`
3. Verify that a node receiving `Optional[Model] = None` no longer raises an error.

I've added the release note and signed the DCO. Ready for the workflows whenever you have a moment to approve!

tests/validation/test_parameter_validator.py
```diff
@@ -65,6 +66,27 @@ def test_missing_param_skips(self, parameter_validator):
         result = parameter_validator._apply_validation(raw, requirements)
         assert result == {"other_key": "value"}
 
+    def test_none_param_raises_when_not_optional(self, parameter_validator):
+        raw = {"model_options": None}
+        requirements = {"model_options": SamplePydanticModel}
```
ravi-kumar-pilla:
```py
requirements = {
"model_options": ParamRequirement(expected_type=SamplePydanticModel,
 allows_none=False,
 )}
```

kedro/validation/parameter_validator.py
```diff
@@ -46,17 +46,28 @@ def _apply_validation(self, raw_params: dict, requirements: dict) -> dict:
...
+            else:
+                expected_type = requirement
+                allows_none = False
             try:
-                raw_value = resolve_nested_dict_path(raw_params, param_key)
+                raw_value = resolve_nested_dict_path(
+                    raw_params, param_key, default=MISSING
+                )
 
-                if raw_value is None:
```
ravi-kumar-pilla, line 53: Hi @PragnyaKhandelwal , Thanks for your patience. I am trying to understand the edge case through your unit tests.

The case where -
raw_params = {"model_options": None}
requirements = {"model_options": SamplePydanticModel}

This will be caught at line 53 - `if raw_value is None:` and the validation will be skipped. This is due to the new [changes](https://github.com/kedro-org/kedro/pull/5542/changes#diff-4a028221853f9274c60bf5a5c72ac2429e2be4b504869731cd0816c900757c9eR59) pushed to `main` to accomodate `namespaces`

ravi-kumar-pilla:
Hi @PragnyaKhandelwal ,

I had a detailed look at the implementation and it mostly looks good. However there are some interlinked issues with the `Optional` typing. I will document my approach here and you can see if this makes sense. I am taking `parameters_data_science.yml` as a reference point -

```yaml
model_options:
  test_size: 0.2
  random_state: 3
  test_fil: null
  features:
    - engines
    - passenger_capacity
    - crew
    - d_check_complete
    - moon_clearance_complete
    - iata_approved
    - company_rating
    - review_scores_rating
```

```python
from pydantic import BaseModel, Field

class ModelOptions(BaseModel):
    test_size: float = Field(ge=0.1, le=0.5)
    random_state: int = Field(ge=0)
    features: list[str] = Field(min_length=1)
    test_fil: Optional[int]
```


## Summary

Three interconnected bugs cause `Optional[ModelOptions]` and non-optional `ModelOptions` to behave identically when a parameter value is `None`. 

- A parameter typed as `Optional[ModelOptions]` should accept `None` without error. (what you are trying to solve for I believe)
- A parameter typed as `ModelOptions` should raise `ParameterValidationError` when `None` is supplied.

**Neither behaviour works correctly today.**

## Bug 1 — `TypeExtractor` eagerly strips `Optional`

### Location
`kedro/validation/type_extractor.py` → `_extract_types_from_node`

### What happens today

`Optional[ModelOptions]` (i.e. `ModelOptions | None`) is unwrapped to `ModelOptions` *before* being stored in the requirements dict. All optionality information is thrown away.

```python
# BEFORE — line 143 (original)
# Unwrap Optional[X] / X | None so we validate against the inner type
expected_type = _unwrap_optional(expected_type)     # ← destructive: loses Optional

# ...
all_requirements[param_key] = expected_type         # stores ModelOptions, not Optional[ModelOptions]
```

Because the stored value is always the inner type, `_apply_validation` can never know whether the original annotation was `ModelOptions` or `Optional[ModelOptions]`.

### Fix

Use `_unwrap_optional` only to check whether the inner type is validatable, but store the **original** type (which may include `Optional`).

```python
# AFTER
# Unwrap Optional[X] / X | None to check if the inner type is validatable,
# but store the original type so _apply_validation knows if None is allowed.
inner_type = _unwrap_optional(expected_type)

if not (is_pydantic_class(inner_type) or dataclasses.is_dataclass(inner_type)):
    ...
    continue

all_requirements[param_key] = expected_type         # preserves Optional[ModelOptions]
```

---

## Bug 2 — `resolve_nested_dict_path` uses `None` as both "not found" and "found with value `None`"

### Location
`kedro/validation/utils.py` → `resolve_nested_dict_path`

### What happens today

The function returns `None` in two completely different situations:

| Situation | Example input | Return value |
|---|---|---|
| Key is absent from the dict | `{}`, looking up `"model_options"` | `None` |
| Key is present, value is `None` | `{"model_options": None}`, looking up `"model_options"` | `None` |

```python
# BEFORE — two distinct cases both return None
def resolve_nested_dict_path(data: dict, path: str) -> Any:
    if path in data:
        return data[path]       # returns None when value is None   ← ambiguous

    if "." not in path:
        return data.get(path)   # returns None when key is missing  ← ambiguous

    ...
    for key in keys:
        if isinstance(value, dict) and key in value:
            value = value[key]
        else:
            return None         # returns None when key is missing  ← ambiguous
    return value
```

Any caller that checks `if raw_value is None` cannot distinguish between the two cases.

### Fix

Introduce a `_MISSING` sentinel and return it whenever a key is absent. (You have done this in your implementation)

```python
# AFTER
_MISSING = object()

def resolve_nested_dict_path(data: dict, path: str) -> Any:
    """...
    Returns ``_MISSING`` if the key is not found under either strategy,
    so callers can distinguish "key absent" from "key present with None value".
    """
    if path in data:
        return data[path]           # returns None only when value IS None

    if "." not in path:
        return _MISSING if path not in data else data[path]

    keys = path.split(".")
    value = data
    for key in keys:
        if isinstance(value, dict) and key in value:
            value = value[key]
        else:
            return _MISSING         # key absent → sentinel, not None
    return value
```

---

## Bug 3 — `_apply_validation` cannot distinguish optional-`None` from required-`None`

### Location
`kedro/validation/parameter_validator.py` → `_apply_validation`

### What happens today

The validator uses `if raw_value is None: continue` for **every** `None` value regardless of whether the type is `Optional`. Because of Bug 2, this branch is also reached when the key is simply absent from the config. Both cases silently skip validation.

```python
# BEFORE
raw_value = resolve_nested_dict_path(raw_params, param_key)

if raw_value is None:           # fires for BOTH "key missing" AND "value is None"
    logger.debug("...skipping...")
    continue                    # always skips — never raises an error

validated_instance = instantiate_model(param_key, raw_value, expected_type)
```

**Consequence:** given `raw = {"model_options": None}` and `requirements = {"model_options": SamplePydanticModel}`, the validator silently skips validation and returns `{"model_options": None}` instead of raising
`ParameterValidationError`. The test below therefore **fails** (no exception is raised):

```python
def test_none_param_raises_when_not_optional(self, parameter_validator):
    raw = {"model_options": None}
    requirements = {"model_options": SamplePydanticModel}
    with pytest.raises(ParameterValidationError, match="Parameter validation failed"):
        parameter_validator._apply_validation(raw, requirements)
```
(The above unit test is passing CI, may be due to kedro installation from the released version and not the main branch)

### Fix

Use `_MISSING` to separate "key absent" from "key present with `None` value`, and use `_is_optional` (a new helper) to differentiate the two `None` cases.

```python
# AFTER
raw_value = resolve_nested_dict_path(raw_params, param_key)

# 1. Key not in config at all → skip silently (no requirement to fulfil)
if raw_value is _MISSING:
    logger.debug("Parameter '%s' not found in config, skipping validation", param_key)
    continue

# 2. Key present, value is None
if raw_value is None:
    if _is_optional(expected_type):   # Optional[ModelOptions] → None is valid
        continue
    # ModelOptions (non-optional) → None is an error
    error_msg = f"Parameter '{param_key}': value is None but type is not Optional"
    validation_errors.append(error_msg)
    logger.error(error_msg)
    continue

# 3. Non-None value → unwrap Optional wrapper before instantiating
inner_type = _unwrap_optional(expected_type)
validated_instance = instantiate_model(param_key, raw_value, inner_type)
```

The two new helpers live in `kedro/validation/utils.py`:

```python
def _is_optional(tp: type) -> bool:
    """Return True if tp is Optional[X] (i.e. X | None)."""
    if get_origin(tp) is Union or isinstance(tp, types.UnionType):
        return type(None) in get_args(tp)
    return False

def _unwrap_optional(tp: type) -> type:
    """Unwrap Optional[X] / X | None to X."""
    if get_origin(tp) is Union or isinstance(tp, types.UnionType):
        args = [a for a in get_args(tp) if a is not type(None)]
        if len(args) == 1:
            return args[0]
    return tp
```

`_unwrap_optional` was previously defined in `type_extractor.py`; it should be moved to
`utils.py` so both `type_extractor.py` and `parameter_validator.py` can share it.

---

I have made a draft implementation of the changes here - https://github.com/kedro-org/kedro/pull/5553

Please have a look at this. Thank you

PragnyaKhandelwal:
Thank you @ravi-kumar-pilla for putting together that detailed draft. I’ve aligned this PR with the approach you outlined: the _MISSING sentinel is in place, `_is_optional` and `_unwrap_optional` live in `utils.py`, and the validation flow now distinguishes missing values from explicit None. I also added the missing end-to-end regression test through `validate_raw_params()` for `Optional[SamplePydanticModel] = None`, alongside the existing unit coverage for the required-but-None case.

kedro/validation/type_extractor.py
```diff
@@ -31,17 +32,26 @@ def _type_name(tp: type) -> str:
     return getattr(tp, "__name__", str(tp))
 
 
-def _unwrap_optional(tp: type) -> type:
+@dataclasses.dataclass(frozen=True)
+class ParamRequirement:
```
ravi-kumar-pilla: I don't think we need this and I see the changes in my draft implementation are not  implemented here as you mentioned in the [comment](https://github.com/kedro-org/kedro/pull/5542#issuecomment-4433573330). Can you please have a look ?

ravi-kumar-pilla (changes requested):
Please follow the [draft](https://github.com/kedro-org/kedro/pull/5553) implementation and make the required changes. Thank you

PragnyaKhandelwal:
Hi @ravi-kumar-pilla, apologies for the confusion with my previous push! I have completely removed the `ParamRequirement` dataclass and aligned the logic strictly with your draft. `expected_type` is now stored as-is by the extractor, and `_is_optional` handles the `None` checks directly in the validator.

ravi-kumar-pilla:
Hi @PragnyaKhandelwal ,

Thank you for making changes, the code looks great to me. We need to address CI failures -
**1. Units tests need 100% coverage -**

Missing loc:
```
kedro/validation/utils.py                        61      2    97%   43, 50
```

**2. Lint errors need to be addressed - `make install-test-requirements` -> `make lint` and see what the issue is.**

I will do final round of e2e tests tomorrow. Also do not forget to add your name in `Release.md` file under `Community contributions` section. Cheers !

Thank you

ravi-kumar-pilla:
Additionally below are some suggestions from our reviewer agent skill worth making minor changes and addressing them -

- [ ] **PR description** — The body still lists ParamRequirement, tuple return from _unwrap_optional, etc., which the branch no longer does (per the May 13 author reply). Updating the description avoids confusing reviewers and future readers.

- [ ]  **tests/validation/test_model_factory.py** — The match= regex was relaxed from requiring the parameter name in the message to a shorter pattern. Suggestion: Confirm that is intentional (e.g. message shape changed upstream); if not, tighten the assertion again so regressions in error context stay visible.

- [ ]  **resolve_nested_dict_path return type** — It is not re-exported in kedro.validation.__all__, but it remains importable from kedro.validation.utils. Returning _MISSING instead of None for a missing key is a behavior change for direct callers. The docstring update in the diff helps; if you expect external use, a one-line note in RELEASE.md would make the contract explicit.

Thank you

PragnyaKhandelwal:
Hi @ravi-kumar-pilla,
I've just pushed the final round of fixes addressing all CI failures and review suggestions.
Ready for your e2e tests! 🚀

RELEASE.md
```diff
@@ -11,6 +11,8 @@ NOTE: This session implementation is under active development and may occasional
 * Added an HTTP endpoint `/snapshot` for accessing project snapshot.
 
 ## Bug fixes and other changes
+* Fixed parameter validation for optional typed params so explicit `None` values are preserved and do not trigger validation errors.
+* Updated `resolve_nested_dict_path()` in `kedro.validation.utils` to return `_MISSING` sentinel instead of `None` for missing keys, distinguishing "key absent" from "key present with None value". This is a behavior change for direct callers of the function.
```
ravi-kumar-pilla: > resolve_nested_dict_path return type — It is not re-exported in kedro.validation.all, but it remains importable from kedro.validation.utils. Returning _MISSING instead of None for a missing key is a behavior change for direct callers. The docstring update in the diff helps; if you expect external use, a one-line note in RELEASE.md would make the contract explicit.

Regarding this, I do not expect an external use, so I think we can remove this note from `Release.md`

ravi-kumar-pilla (approved):
This works well. Thank you so much for addressing all the comments. Appreciate your contribution @PragnyaKhandelwal

ElenaKhaustova (approved):
LGTM 👍

kedro/validation/utils.py
```diff
@@ -50,14 +72,15 @@ def resolve_nested_dict_path(data: dict, path: str) -> Any:
...
+    Returns ``_MISSING`` if the key is not found under either strategy,
+    so callers can distinguish "key absent" from "key present with None value".
     """
     # Flat key takes priority (handles namespace params, e.g. "demo.config")
     if path in data:
         return data[path]
 
     if "." not in path:
-        return data.get(path)
+        return _MISSING if path not in data else data[path]
```
SajidAlamQB:
```suggestion
        return _MISSING
```
nit: If we got here, path isn't in data (line 79 would've returned otherwise), so the ternary always picks `_MISSING`

SajidAlamQB (approved):
Awesome work @PragnyaKhandelwal this looks great! Just left a small nit.

PR #5546, opened 2026-05-08
Add http GET endpoint for accessing project snapshot

### LINKED ISSUE #5266: Introduce a Kedro project inspection layer (REST API)
### Description

This is part of #5262 

Users need a standard way to query a Kedro project’s structure (pipelines, nodes, datasets, parameters) without running pipelines or relying on tool-specific inspection logic.

Today, plugins like Kedro-Viz, Vscode-Kedro, Kedro-MLflow, and internal plugins each reimplement their own inspection by loading a `KedroSession` and traversing pipelines and the catalog. This leads to duplication, inconsistent results, and no unified API for remote inspection.

A lightweight, standalone inspection API would provide a single, reusable metadata surface for all tools—especially Kedro Service—and removes the need to depend on Kedro-Viz internals.

This work directly advances **Kedro #4363**, which proposes extracting and standardising Kedro-Viz’s inspection logic. By defining stable inspection endpoints here, we lay the foundation for Kedro-Viz to consume the API and drop its internal Pydantic models as part of modularisation.

---

### Context

This work aligns with two parallel (but independent) initiatives:

1. **[Kedro Service](#5262 )**  
   Remote deployments—such as APIs, orchestration tools, and agentic systems—need to query project metadata *before* choosing which pipeline or node to execute.  
   A stable inspection layer is a foundational requirement for the service architecture.

2. **[Kedro-Viz modularization](https://github.com/kedro-org/kedro-viz/issues/2265)**  Lightweight Kedro-viz 

---

### Why start with all Kedro-Viz inspection endpoints?

We do not yet know which inspection features external users will require, but Kedro-Viz already exercises the broadest set of inspection needs. By implementing all Viz-required endpoints:

- the API becomes immediately useful (Viz is a guaranteed consumer)  
- Kedro-Viz can shed duplicated logic and become lighter  
- Kedro Service gains a complete inspection surface  
- we establish a unified inspection contract for future tools  

We will also cross-check these endpoints against other Kedro plugins (Airflow, Docker, Telemetry) to ensure the inspection API generalises well beyond Kedro-Viz.

Additional endpoints can be added iteratively as new requirements emerge.

#### **Core Inspection Capabilities**

Provide static (non-execution) metadata about the project:

- **Project metadata:** root path, package name, environments, config paths  
- **Pipelines:** list pipelines, return full pipeline structure  
- **Nodes:** names, inputs, outputs, tags, namespace, source code location  
- **Datasets:** dataset definitions, types, metadata, layers, transcoding maps  
- **Parameters:** parameter keys, resolved values, definition files  
- **Modular pipelines:** namespace hierarchy, grouping  
- **Configuration lookup:** resolve catalog/parameter files per environment  

---

### Design Work Required

- Survey existing Kedro plugins (Airflow, Docker, Telemetry, etc.) to identify common inspection needs and ensure endpoint alignment across tools.
This validates that the API is suitable not only for Kedro-Viz but also for other external integrations.
- Determine how much of Kedro-Viz’s inspection logic can be **reused**, **extracted**, or **adapted** 
- Define response schemas (reuse Kedro-Viz Pydantic models where possible).  
- Align with discussions in **Kedro #4363** about centralizing inspection capability.  
- Ensure the API layer remains **lightweight**, **dependency-free**, and suitable for deployment.  
- Identify essential vs. optional features for the first release.

### Acceptance Criteria

- [ ] All inspection endpoints implemented
- [ ] Accurate representation of project structure, pipelines, nodes, datasets, and parameters
- [ ] Modular pipeline namespaces correctly exposed
- [ ] Pydantic response models defined and stable
- [ ] Compatible with Kedro-Viz schemas (but no Kedro-Viz dependency)
- [ ] Code reuse strategy documented (kedro-viz alignment)
- [ ] Unit and integration tests passing
- [ ] Acceptable performance on large projects (100+ nodes)
- [ ] API documentation includes examples

### LINKED ISSUE #5485: Add REST API endpoint for accessing project snapshot
## Description

Currently, there's no programmatic way to inspect a Kedro project's structure via an API. Users must write custom scripts or use the Python API directly. We need a REST endpoint that surfaces `get_project_snapshot` functionality to enable project introspection in production services and tooling.

## Context

This is a companion to the POST /run endpoint in the Kedro Service phase. The goal is to expose the inspection layer as a read-only API endpoint, giving clients visibility into pipelines, datasets, and parameters without executing anything. This enables tooling (dashboards, CI checks, IDE integrations) to query project structure over HTTP.

## Possible Implementation

**API Endpoint:**

- GET /snapshot that calls `get_project_snapshot`
- Accepts: project_path, optional env parameter (maps to the env argument)
- Returns: `ProjectSnapshot` - metadata, pipelines (with nodes), datasets, parameters ( may be validated against a pydantic response ?? )

## Acceptance Criteria
[ ] GET /snapshot returns a valid project snapshot
[ ] Response matches the structure of ProjectSnapshot (metadata, pipelines, datasets, parameters)
[ ] Optional env parameter is respected
[ ] Works without executing any pipeline nodes or loading data
[ ] Request/response models validated (pydantic?)
[ ] Tests passing

ravi-kumar-pilla:
## Description

Resolves #5485 

## Development notes

- Added GET `/snapshot` REST endpoint
- Added pydantic models for the `/snapshot` endpoint response
- Updated tests: Moved run endpoint tests to a different file, created conftest for reuse, added tests for snapshot endpoint
- Updated release notes

**NOTE:** Documentation will be implemented in a separate PR - https://github.com/kedro-org/kedro/pull/5558

**For testing -** 

- Navigate to your kedro project
- Install kedro from `main branch`:
- Follow this to start kedro server - https://docs.kedro.org/en/latest/extend/serving/
- Once started, access /snapshot endpoint - `curl http://127.0.0.1:8000/snapshot` 
- Verify the snapshot json

**Example for spaceflights project:**

```json
{
  "status": "success",
  "metadata": {
    "project_name": "My Project",
    "package_name": "my_project",
    "kedro_version": "1.0.0"
  },
  "pipelines": [
    {
      "name": "__default__",
      "nodes": [
        {
          "name": "split_data_node",
          "inputs": ["example_iris_data"],
          "outputs": ["X_train", "X_test"],
          "tags": [],
          "namespace": null
        }
      ],
      "inputs": ["example_iris_data"],
      "outputs": ["example_predictions"]
    }
  ],
  "datasets": {
    "example_iris_data": {
      "name": "example_iris_data",
      "type": "pandas.CSVDataset",
      "filepath": "data/01_raw/iris.csv"
    }
  },
  "parameters": ["example_learning_rate", "example_num_train_iter"],
  "error": null
}
```

## Checklist


kedro/server/http_server.py
```diff
@@ -45,7 +48,8 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
     """
     project_path = app.state.project_path
     logger.info("Bootstrapping Kedro project at: %s", project_path)
-    bootstrap_project(project_path)
+    metadata = bootstrap_project(project_path)
+    _seed_bootstrap_cache(project_path, metadata)
```
ravi-kumar-pilla: I could not get around this. I feel it is a hard coupling with the inspection layer but I do not want to bootstrap project even on the first call to `/snapshot` as it is already done on the server level. 

To avoid the coupling, the trade-off (small overhead) will be to allow duplication of bootstrapping project once i.e., during the first `/snapshot` call which will populate the `_bootstrapped` cache in inspection layer and avoids further duplication on multiple calls. 

Another solution is to add project metadata to `app.state` and pass it on to  `get_project_snapshot`.

docs/inspect/inspect-project.md
```diff
@@ -155,3 +155,50 @@ snapshot = get_project_snapshot(project_path = "/path/to/my_project", env="stagi
...
 This follows the same environment resolution rules as `kedro run --env staging`. See [Configuration basics](../configure/configuration_basics.md) for details on how environments work.
+
+## How to access the snapshot through the HTTP server
+
+When running the Kedro HTTP server, the same project snapshot is available over HTTP at `GET /snapshot`. The endpoint uses the environment configured at server startup and returns a JSON response with a `status` field (`"success"` or `"failure"`), the same `metadata`, `pipelines`, `datasets`, and `parameters` fields as `ProjectSnapshot`, and an `error` field on failure.
+
+Start the server:
+
+```bash
+kedro serve
```
ankatiyar: This command is `kedro server start` and once the docs PR is merged, can add the endpoint and cross link these pages https://github.com/kedro-org/kedro/pull/5547

docs/inspect/inspect-project.md
```diff
@@ -155,3 +155,50 @@ snapshot = get_project_snapshot(project_path = "/path/to/my_project", env="stagi
...
+```
+
+Fetch the snapshot:
+
+```bash
+curl http://localhost:4141/snapshot
```
ankatiyar: The default i've set on http server is `127.0.0.1:8000`

kedro/inspection/snapshot.py
```diff
@@ -134,7 +164,10 @@ def _build_project_snapshot(
             f"Invalid env value {env!r}: must contain only letters, digits, hyphens, and underscores."
         )
 
-    metadata = bootstrap_project(project_path)
+    with _bootstrap_lock:
+        if project_path not in _bootstrapped:
+            _bootstrapped[project_path] = bootstrap_project(project_path)
+        metadata = _bootstrapped[project_path]
     config_loader = _make_config_loader(project_path, env=env)
```
ankatiyar: QQ about this, do you not need the `conf_source`? For http server it is possible to pass a config source that is outside the project for example, a `.zip` file etc. When going through the http server, we are setting the `conf_source` but for getting the project snapshot it might be ignored and the default may be used.

kedro/server/http_server.py
ankatiyar: Prefer the second option, add it to `app.state` and pass it to `get_project_snapshot`

ravi-kumar-pilla: Thank you @ankatiyar ... I will update it once all the reviewers leave their suggestions on this.

kedro/inspection/snapshot.py
ravi-kumar-pilla: okay. For inspection this is how I get the conf_source - https://github.com/kedro-org/kedro/blob/main/kedro/inspection/helper.py#L34

Since for inspection we do not create a session, I constructed the config_loader this way - https://github.com/kedro-org/kedro/blob/main/kedro/inspection/helper.py#L20

I think this should work but if we want a custom conf_source that is not provided in settings, I should think of this. Let me know your thoughts. Thank you

docs/inspect/inspect-project.md
```diff
@@ -155,3 +155,50 @@ snapshot = get_project_snapshot(project_path = "/path/to/my_project", env="stagi
...
+  "datasets": {
+    "example_iris_data": {"name": "example_iris_data", "type": "pandas.CSVDataset", "filepath": "data/01_raw/iris.csv"}
+  },
+  "parameters": ["example_learning_rate", "example_num_train_iter"],
+  "error": null
+}
+```
+
+If the snapshot cannot be built (for example, due to a catalog error), the response still returns HTTP 200 with `"status": "failure"`. The `error` field contains the exception type, message, and stack trace.
+
```
ravi-kumar-pilla: I will link the server docs once the PR [here](https://github.com/kedro-org/kedro/pull/5547) is merged

kedro/server/models.py
```diff
@@ -116,3 +122,24 @@ class HealthResponse(BaseModel):
         default="healthy", description="Server health status."
     )
     kedro_version: str = Field(description="Kedro version.")
+
+
+class SnapshotResponse(BaseModel):
+    """Response model for the project snapshot endpoint."""
```
rashidakanchwala, line 128: I noticed this PR follows the same approach as RunResponse (#5521), which confused me a bit.

Do we want to have separate Success and Failure responses, with SnapshotResponse as a union of the two? If we take that route here, we would probably want to apply the same pattern to RunResponse as well, maybe we do both in follow up PRs.

I am not strongly attached to either approach....the current one does mean less code, so I wanted to get your thoughts, @ankatiyar @ravi-kumar-pilla.

kedro/server/http_server.py
rashidakanchwala: Yes agree with both. let's do add it to app.state

ElenaKhaustova (commented):
## Kedro PR Review

### Overview
- **Critical:** 1 | **Suggestions:** 5

- [x] `kedro/server/http_server.py:147` — `ErrorDetail` is constructed with a `traceback` keyword the model does not declare. With pydantic v2's default `extra="ignore"`, the field is silently dropped. Both lint (`mypy [call-arg]`) and unit-tests (`KeyError: 'traceback'` in `test_snapshot_returns_failure_status_on_exception`) are red on this. Either add a `traceback` field to `ErrorDetail` (mind the `/run` strict-equality assertion at `test_run_endpoint.py:194-198`) or, cleaner, define a `SnapshotErrorDetail` and use it only in `SnapshotResponse`.

- [x] **Cache machinery may become unnecessary once `app.state` is adopted** (already agreed in the open `_seed_bootstrap_cache` thread). Consider extending that change: replace `_bootstrapped`, `_bootstrap_lock`, `_seed_bootstrap_cache`, `_clear_bootstrap_cache`, the cross-package import of a private symbol, and the cache-clearing test fixtures with a single optional `metadata` parameter on `_build_project_snapshot`. The lifespan stores metadata on `app.state.project_metadata` and passes it through; the inspection module loses all module-level mutable state. Trade-off: programmatic callers that invoke `get_project_snapshot()` repeatedly would re-bootstrap each time, which is correct but slower for a niche pattern.
- [ ] **Re: discriminated-union response model (rashidakanchwala's question)** — agree the `SnapshotSuccess | SnapshotFailure` shape models the success/failure invariants more accurately than the current "all-Optional" design. Consistency with `RunResponse` is the strongest reason to keep the current shape in *this* PR; migrating both endpoints together in a follow-up is the right path.
- [ ] **Asymmetric error contracts** — `/run` excludes traceback (asserted in `test_run_endpoint.py:194-198`), `/snapshot` advertises it in the docs. Pick one error shape and apply it consistently across the server (easier under the discriminated-union refactor above).
- [x] `kedro/inspection/snapshot.py:37` — module comment refers to `clear_bootstrap_cache()` but the function is `_clear_bootstrap_cache()`. Falls away if the cache is removed; otherwise rename the function or fix the comment.
- [x] `tests/server/conftest.py:8-11` — the autouse `_patch_seed_bootstrap_cache` mock isn't consumed by any test, so nothing asserts the lifespan calls `_seed_bootstrap_cache(project_path, metadata)` after `bootstrap_project`. Falls away if the cache is removed; otherwise add one positive integration test using the returned mock.

### Notes
- Already raised in existing review threads (not duplicated here): per-request `conf_source` (ankatiyar), `app.state` plumbing (ankatiyar + rashidakanchwala), docs port `127.0.0.1:8000`, `kedro server start` command, cross-link to docs after #5547.
- `RELEASE.md` entry is in the right H2 section. The test split into `test_run_endpoint.py` + `test_snapshot_endpoint.py` and the `make_http_server` factory are nice cleanups.
- The "Updated the documentation" PR-checklist box is unticked even though docs are touched in this PR.

kedro/server/http_server.py
```diff
@@ -108,6 +113,41 @@ async def health() -> HealthResponse:
...
+                parameters=snapshot.parameters,
+            )
+        except Exception as exc:
+            logger.error("Snapshot request failed: %s", str(exc), exc_info=True)
+            return SnapshotResponse(
+                status="failure",
+                error=ErrorDetail(
+                    type=type(exc).__qualname__,
+                    message=str(exc),
+                    traceback=traceback.format_tb(exc.__traceback__),
```
ElenaKhaustova: **Critical:** `ErrorDetail` only declares `type` and `message` (`kedro/server/models.py:99-103`). With pydantic v2's default `extra="ignore"`, this `traceback` value is silently dropped from the response.

This is what's breaking CI right now:
- mypy: `kedro/server/http_server.py:144: error: Unexpected keyword argument "traceback" for "ErrorDetail" [call-arg]`
- pytest: `tests/server/test_snapshot_endpoint.py::TestSnapshotEndpoint::test_snapshot_returns_failure_status_on_exception` fails with `KeyError: 'traceback'`

The docs page advertises that the `error` field contains the traceback, so the intent is clearly to include it.

A few possible fixes:
1. Add an optional `traceback: list[str] | None = Field(default=None, ...)` field to `ErrorDetail`. Note this will break the existing `/run` failure test (`tests/server/test_run_endpoint.py:194-198`) which strict-equals the dict and asserts `"traceback" not in payload["error"]` — you'd need `response_model_exclude_none=True` on the routes or to update that test.
2. Define a separate `SnapshotErrorDetail` model that extends/duplicates `ErrorDetail` with the `traceback` field, used only by `SnapshotResponse`. Keeps `/run`'s contract unchanged.
3. Drop `traceback` from the snapshot response and update the docs accordingly.

Option 2 is the cleanest if you want to keep `/snapshot` and `/run` independently evolvable. (See also the discriminated-union note in the summary — that path makes per-endpoint failure shapes natural.)

kedro/inspection/snapshot.py
```diff
@@ -28,6 +29,35 @@
 
 
 _ENV_RE = re.compile(r"^[A-Za-z0-9_-]+$")
+_bootstrap_lock = threading.Lock()
+# Process-global cache mapping resolved project path → ProjectMetadata.
+# Designed for single-project, long-running server processes where the project
+# configuration does not change at runtime. Multiple project paths are supported
+# (each bootstraps independently), but there is no automatic eviction.
+# Call clear_bootstrap_cache() to force re-bootstrap after a configuration change.
+_bootstrapped: dict[Path, ProjectMetadata] = {}
```
ElenaKhaustova: **Suggestion (conceptual):** once `app.state` plumbing is adopted (already agreed in the open thread on `kedro/server/http_server.py:53`), this whole cache machinery may not need to exist.

The single problem the cache solves is "don't re-bootstrap on the first `/snapshot` call after lifespan already bootstrapped". You can solve that by passing metadata through directly:

```python
def _build_project_snapshot(
    project_path: str | Path,
    env: str | None = None,
    *,
    metadata: ProjectMetadata | None = None,
) -> ProjectSnapshot:
    project_path = Path(project_path).resolve()
    if metadata is None:
        metadata = bootstrap_project(project_path)
    ...
```

Then the lifespan stores metadata on `app.state.project_metadata` and the `/snapshot` route passes it in. That deletes:
- `_bootstrap_lock`, `_bootstrapped`, `_seed_bootstrap_cache`, `_clear_bootstrap_cache`
- The cross-package import of a private symbol from the server
- `mocker.patch.dict(_bootstrapped, {}, clear=True)` boilerplate in `tests/inspection/test_project_snapshot.py`
- The autouse `_patch_seed_bootstrap_cache` fixture in `tests/server/conftest.py`
- The whole `TestBootstrapCacheLifecycle` class

Trade-off: programmatic callers of `get_project_snapshot()` invoked repeatedly in the same process would re-bootstrap each call. That's correct (just slower) for a niche usage pattern, and `bootstrap_project` is idempotent.

Not blocking — just worth weighing alongside the agreed `app.state` change since the two combine well.

kedro/inspection/snapshot.py
```diff
@@ -28,6 +29,35 @@
 
 
 _ENV_RE = re.compile(r"^[A-Za-z0-9_-]+$")
+_bootstrap_lock = threading.Lock()
+# Process-global cache mapping resolved project path → ProjectMetadata.
+# Designed for single-project, long-running server processes where the project
+# configuration does not change at runtime. Multiple project paths are supported
+# (each bootstraps independently), but there is no automatic eviction.
+# Call clear_bootstrap_cache() to force re-bootstrap after a configuration change.
```
ElenaKhaustova: **Suggestion (nit):** this comment says `clear_bootstrap_cache()` but the function is named `_clear_bootstrap_cache()` (line 52). Either fix the comment or, if the function is meant to be a supported escape hatch, drop the leading underscore (and consider doing the same for `_seed_bootstrap_cache`, which is already imported by name from another package).

Falls away if the cache is removed per the suggestion above.

kedro/server/models.py
ElenaKhaustova, line 128: **Re: your union question** — agreed, the discriminated `SnapshotSuccess | SnapshotFailure` shape models the invariants more accurately than the current "all fields Optional" design (the OpenAPI schema currently can't express "data is present iff `status == 'success'`"). Consistency with `RunResponse` is the strongest reason to keep the current shape in *this* PR. Migrating both endpoints together in a follow-up sounds right.

One nice side-effect of the union shape: each endpoint's failure variant can carry its own fields (e.g. `/snapshot` failure includes `traceback`, `/run` failure does not), which would resolve the asymmetric-error-contract issue mentioned in the summary.

tests/server/conftest.py
```diff
@@ -0,0 +1,40 @@
...
+
+import pytest
+
+from kedro.server.http_server import create_http_server
+
+
+@pytest.fixture(autouse=True)
+def _patch_seed_bootstrap_cache(mocker):
+    """Prevent _seed_bootstrap_cache from mutating the inspection module's cache during server tests."""
+    return mocker.patch("kedro.server.http_server._seed_bootstrap_cache")
```
ElenaKhaustova: **Suggestion:** this autouse mock patches `_seed_bootstrap_cache` for every test in `tests/server/`, but no test takes the fixture as a parameter to assert on the returned `Mock`. As a result, nothing in this test suite verifies that the lifespan calls `_seed_bootstrap_cache(project_path, metadata)` after `bootstrap_project` — a swap of arguments wouldn't be caught here (only the inspection-side tests of `_seed_bootstrap_cache` itself would still pass).

A small positive integration test would lock down the wiring:

```python
def test_lifespan_seeds_bootstrap_cache(
    mocker, make_http_server, _patch_seed_bootstrap_cache
):
    fake_metadata = mocker.Mock()
    mocker.patch(
        "kedro.server.http_server.bootstrap_project",
        return_value=fake_metadata,
    )
    app = make_http_server()
    with TestClient(app):
        pass
    _patch_seed_bootstrap_cache.assert_called_once_with(
        app.state.project_path, fake_metadata
    )
```

Falls away if the cache infrastructure is removed.

kedro/server/models.py
ravi-kumar-pilla: As you said, I followed the RunResponse pattern which returns error as `null` incase of success. We can go ahead with the similar pattern unless we want to introduce 2 separate responses which will increase the model count. For a fairly minimal http server, current approach should be fine unless someone reports of the usage complexity.

kedro/inspection/snapshot.py
ravi-kumar-pilla: I will add conf_source to `get_snapshot` as we discussed offline to respect the customization present in the server

kedro/server/http_server.py
ravi-kumar-pilla: I will remove the _bootstrapped cache for now then. Thank you @rashidakanchwala @ankatiyar

kedro/server/models.py
ravi-kumar-pilla: Will follow up here - https://github.com/kedro-org/kedro/issues/5557

ravi-kumar-pilla:
Thanks @ankatiyar @rashidakanchwala @ElenaKhaustova for the reviews. I fixed all of the issues raised and separated doc changes as it was getting a bit heavy in the same PR. I will open the docs [PR](https://github.com/kedro-org/kedro/pull/5558) for review once this is merged. Please take a look one more time and let me know of any changes. Thank you 👍

kedro/inspection/snapshot.py
```diff
@@ -134,8 +142,9 @@ def _build_project_snapshot(
             f"Invalid env value {env!r}: must contain only letters, digits, hyphens, and underscores."
         )
 
-    metadata = bootstrap_project(project_path)
-    config_loader = _make_config_loader(project_path, env=env)
+    if metadata is None:
```
ankatiyar, line 167: There is a possibility that someone provides `project_path` and `metadata` both and in that case `metadata.project_path` might be different. Maybe we should accept just one of these args and error OR warning when both are present and say something about `project_path` being ignored.

kedro/server/models.py
```diff
@@ -97,6 +103,15 @@ class ErrorDetail(BaseModel):
     message: str = Field(description="Error message.")
 
 
+class SnapshotErrorDetail(ErrorDetail):
+    """Error detail for the snapshot endpoint, including a stack trace."""
+
+    traceback: list[str] | None = Field(
```
ankatiyar: Do you need a traceback on top of type and message? I had it in `ErrorDetail` and removed it because it was flagged as a security issue - the same might be true here. Maybe you can just re-use `ErrorDetail`

kedro/inspection/snapshot.py
ankatiyar: Based on what you decide to do here - only pass one of these from the http server here so error/warnings don't show up when it's used via server (only programatically if both are passed)

kedro/server/http_server.py
```diff
@@ -108,6 +112,43 @@ async def health() -> HealthResponse:
...
+        Returns:
+            `SnapshotResponse` with project metadata, pipelines, datasets,
+            and parameter keys, or error details on failure.
+        """
+        try:
+            snapshot = get_project_snapshot(
+                project_path=app.state.project_path,
+                env=app.state.default_env,
+                conf_source=app.state.default_conf_source,
+                metadata=app.state.metadata,
```
ankatiyar, line 129: related to comment in `snapshot.py` -> we should pass just one of `project_path` or `metadata` from here since you only need it for bootstrapping.

ankatiyar (approved):
Thanks Ravi! LGTM

kedro/inspection/snapshot.py
```diff
@@ -113,29 +114,58 @@ def _build_pipeline_snapshots(
...
+            skipped entirely.
 
     Returns:
         A fully populated ``ProjectSnapshot``.
     """
-    project_path = Path(project_path).resolve()
+    if metadata is not None:
+        if (
+            project_path is not None
+            and Path(project_path).resolve() != metadata.project_path
```
ElenaKhaustova: **Nit (edge case):** `bootstrap_project` resolves with `Path(...).expanduser().resolve()` (`kedro/framework/startup.py:154`), but the comparison here only does `.resolve()`. So a programmatic caller doing

```python
metadata = bootstrap_project("~/proj")
get_project_snapshot(project_path="~/proj", metadata=metadata)
```

hits a false-positive warning, because `Path("~/proj").resolve()` keeps the literal `~` (resolves to `<cwd>/~/proj`) while `metadata.project_path` is `/Users/<you>/proj`. Absolute and `./relative` paths are unaffected.

The HTTP server path doesn't trigger this (it only passes `metadata`), so it's strictly a programmatic-API concern. Trivial fix:

```suggestion
            and Path(project_path).expanduser().resolve() != metadata.project_path
```

(plus the same in the warning's f-string on the next lines so the message reads consistently). Not blocking.

ElenaKhaustova (commented):
## Re-review

All the findings from my [previous review](https://github.com/kedro-org/kedro/pull/5546#pullrequestreview-4274327204) are addressed.

One tiny edge case below — non-blocking. LGTM otherwise.

ElenaKhaustova (approved):
Thank you, @ravi-kumar-pilla 👍

PR #5547, opened 2026-05-08
Server docs

ankatiyar:
## Description

Resolve #5450 
Rendered page: https://kedro--5547.org.readthedocs.build/en/5547/extend/serving/

## Development notes

- API docs for the `kedro.server` module
- API docs for `kedro.framework.cli.server` command
- Docs page for the http server under "Extend" header

## Developer Certificate of Origin
We need all contributions to comply with the [Developer Certificate of Origin (DCO)](https://developercertificate.org/). All commits must be signed off by including a `Signed-off-by` line in the commit message. [See our wiki for guidance](https://github.com/kedro-org/kedro/wiki/Guidelines-for-contributing-developers/).

If your PR is blocked due to unsigned commits, then you must follow the instructions under "Rebase the branch" on the GitHub Checks page for your PR. This will retroactively add the sign-off to all unsigned commits and allow the DCO check to pass.

## Checklist

- [ ] Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [ ] Signed off each commit with a [Developer Certificate of Origin (DCO)](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository)
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [ ] Added tests to cover my changes
- [ ] Checked if this change will affect Kedro-Viz, and if so, communicated that with the Viz team

PR #5552, opened 2026-05-11
Security fixes for HTTP Server

ankatiyar:
## Description

Did a quick security review for the HTTP server code and it flagged some issues which are resolved in this:
- Check if the `runner` provided is a subclass of `AbstractRunner` to avoid arbitrary code execution
- Remove full traceback from `ErrorDetail` - it is still logged server side
- Remove `project_path` from health endpoint 
- Add a warning for `--reload` functionality of `uvicorn` as it's not fit for use in production
- Updated tests wherever necessary, added tests for the warning

## Checklist

### DIFF
--- a/kedro/framework/cli/server.py
+++ b/kedro/framework/cli/server.py
@@ -42,7 +42,7 @@ def server_cli() -> None:
     "--reload",
     is_flag=True,
     default=False,
-    help="Enable auto-reload for development. Server restarts on code changes.",
+    help="Enable auto-reload for development. Server restarts on code changes. Should not be used in production.",
 )
 @click.option(
     "--env",
--- a/kedro/framework/cli/server.py
+++ b/kedro/framework/cli/server.py
@@ -1,4 +1,5 @@
 import os
+import warnings
 
 import click
 
@@ -90,6 +91,14 @@ def server_start(  # noqa: PLR0913
         ) from exc
 
     project_path = metadata.project_path
+
+    if reload:
+        warnings.warn(
+            "--reload enables Uvicorn's auto-reload mode which is intended for "
+            "development only. Do not use it in production.",
+            UserWarning,
+            stacklevel=1,
+        )
 
     # Unset env and conf-source to avoid re-using them from previous runs if not provided in the current command
     if KEDRO_SERVER_ENV in os.environ:
--- a/kedro/server/http_server.py
+++ b/kedro/server/http_server.py
@@ -6,7 +6,6 @@ import logging
 import os
 import threading
 import time
-import traceback
 from contextlib import asynccontextmanager
 from typing import TYPE_CHECKING
 
@@ -17,6 +16,7 @@ from kedro.framework.project import settings
 from kedro.framework.session.service_session import KedroServiceSession
 from kedro.framework.startup import bootstrap_project
 from kedro.io.core import generate_timestamp
+from kedro.runner import AbstractRunner
 from kedro.server.models import (
     ErrorDetail,
     HealthResponse,
@@ -103,11 +103,9 @@ def create_http_server(
 
         Returns server status and Kedro version information.
         """
-        project_path = app.state.project_path
         return HealthResponse(
             status="healthy",
             kedro_version=kedro_version,
-            project_path=str(project_path),
         )
 
     @app.post("/run", response_model=RunResponse, tags=["pipeline"])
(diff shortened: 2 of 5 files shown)


merelcht (approved):
Thanks @ankatiyar ! Good call to run an extra security check on this, these changes seem indeed to be for the better 👍

kedro/server/http_server.py
```diff
@@ -162,6 +160,13 @@ def _execute_pipeline(
     try:
         runner_name = request.runner or "SequentialRunner"
         runner_class = load_obj(runner_name, "kedro.runner")
+        if not (
+            isinstance(runner_class, type) and issubclass(runner_class, AbstractRunner)
```
ElenaKhaustova, line 164: **Suggestion (defense-in-depth):** `load_obj(runner_name, "kedro.runner")` calls `importlib.import_module(...)` *before* this `AbstractRunner` check runs. So an attacker can still cause the server to import any module installed in its environment (and execute that module's top-level code) by passing e.g. `"some_module.foo"` as `runner` — the subclass check then rejects the call, but the import has already happened.

The new check correctly stops `runner_class(is_async=...)` from invoking arbitrary callables, which is the bigger risk. But for a more complete defense, consider one of:

1. Restricting `runner` to a bare class name resolved against `kedro.runner` (and an explicit allow-list for any out-of-tree runners).
2. Validating the `runner` string upfront (e.g. reject anything not matching a known module prefix) before calling `load_obj`.
3. At minimum, documenting on `RunRequest.runner` / in the server docs that the HTTP endpoint must only be exposed in a trusted network context, since module imports remain arbitrary.

Not a blocker — just flagging the residual surface.

kedro/framework/cli/server.py
```diff
@@ -91,6 +92,14 @@ def server_start(  # noqa: PLR0913
 
     project_path = metadata.project_path
 
+    if reload:
+        warnings.warn(
```
ElenaKhaustova, line 96: **Suggestion (visibility of the production-safety warning):**

`warnings.warn(..., UserWarning)` is shown by default but can be silenced by `PYTHONWARNINGS=ignore`, `python -W ignore`, or any custom warning filter. For a "do not use in production" message it's worth making it harder to miss. Two ideas:

- Also emit via `click.secho("Warning: --reload is intended for development only. Do not use in production.", fg="yellow", err=True)` to match the user-facing CLI messaging style elsewhere in `kedro/framework/cli/`.
- Or use `logger.warning(...)` so it surfaces regardless of the warning filter.

Keeping `warnings.warn` is fine and consistent with `kedro/framework/cli/project.py`, but combining it with one of the above would make the production caveat unconditional.

Related nit (not in the diff): the `--reload` `--help` text at `kedro/framework/cli/server.py:43-45` still reads `"Enable auto-reload for development. Server restarts on code changes."`. Consider also updating it to e.g. `"Enable auto-reload on code changes. Development only — do not use in production."` so the caveat is visible from `kedro server start --help` without having to invoke the command.

ElenaKhaustova (commented):
## Kedro PR Review

### Overview
- **Suggestions:** 3

- [ ] `kedro/server/http_server.py:L164` — `load_obj` imports the target module before the `AbstractRunner` check, so the import-time side effects of any installed module are still triggerable. Consider tightening the runner string validation.
- [ ] `kedro/framework/cli/server.py:L96` — `warnings.warn(..., UserWarning)` can be silenced via `PYTHONWARNINGS=ignore` / `-W ignore`. For a production-safety message, consider also surfacing it via `click.secho(..., fg="yellow", err=True)` or `logger.warning(...)`.
- [ ] `kedro/framework/cli/server.py:L43` — Mirror the production caveat in the `--reload` `--help` text so it's visible from `kedro server start --help` without invoking the command.

### Notes
- The four security fixes themselves are correct and well-tested. The new `_FakeRunner` is correctly an `AbstractRunner` subclass and the rejection paths (non-class and non-subclass) are both covered.
- Heads-up for cross-PR coordination: PR #5547 ("Server docs") adds `docs/extend/serving.md`, where the `/health` example currently shows `project_path`. When that PR rebases on top of this one, the example needs to drop `project_path`, and ideally the docs should mention the new runner validation and the production warning for `--reload`.

ElenaKhaustova (approved):

ankatiyar:
I added a note to the help string for `--reload` so it shows up with the `--help` text and a field validator for the `runner` part of the run request. It does protect against some cases but will update the docs accordingly with a note.
Double warnings might be annoying I think so leaving that out.

PR #5561, opened 2026-05-13
Security scan skill

rashidakanchwala:
### Description

This PR introduces an automated security scanning skill for Kedro that helps identify and prioritise real security risks in the codebase.

---

#### What’s included

- A new **security scan agent** that runs Semgrep on:
  - the full repository, or
  - only files changed in a PR  
- A **Kedro-specific ruleset** to catch meaningful issues (not just generic warnings)
- A **triage system** that classifies findings into clear, actionable categories
- A final **human-readable report** (in chat or posted to the PR)

---

#### Why this matters

Generic security scanners tend to produce a lot of noise, making it hard to know what actually needs fixing.

This skill improves signal by:
- Adding **rules tailored to Kedro’s architecture**
- Filtering and **classifying findings based on real impact**
- Producing a **clear, prioritised report** instead of raw scan output

---

#### How it works (high level)

1. **Scan**
   - Runs four Semgrep rulesets in parallel:
     - `p/security-audit`
     - `p/secrets`
     - `p/python`
     - custom Kedro rules

2. **Detect Kedro-specific risks**, such as:
   - Loading classes/plugins from config
   - Unsafe YAML parsing
   - Path traversal risks
   - Executable config patterns (e.g. logging config)

3. **Triage findings** into:
   - **Kedro vulnerability**  
   - **Developer responsibility**  
   - **Deployment/environment issue**  
   - **False positive**  

4. **Output a clean report** with only relevant, actionable issues

---

#### How to use

- `"run security scan on this PR"`
- `"run security scan on the full codebase"`

---

#### Files added

- `SKILL.md` — workflow and output format  
- `rules/kedro-security-patterns.yml` — Kedro-specific Semgrep rules  
- `references.md` — classification guide  

---

#### Notes for reviewers

- Focus on **behavior and outcomes**, not Semgrep internals  
- The key value is in:
  - Kedro-specific detection
  - Intelligent triage (reducing noise)


.agents/skills/security-scan/SKILL.md
```diff
@@ -0,0 +1,359 @@
...
+4. run Semgrep in parallel across all rulesets
+5. read and deduplicate findings
+6. triage and classify findings against the Kedro security model
+7. run manual review checks from `references/kedro-findings-triage.md`
+8. emit one final report
+
+## References
+
+Read these before triaging findings:
+- [../../docs/about/security_model.md](../../docs/about/security_model.md)
```
ElenaKhaustova: **Critical:** broken relative path. This `SKILL.md` lives at `.agents/skills/security-scan/`, so reaching `docs/about/security_model.md` needs **three** `../` (`../../../docs/about/security_model.md`), not two. The current link resolves to `.agents/docs/about/security_model.md`, which doesn't exist.

Same bug at line 249 ("Triage against the Kedro security model"). Both need updating so the agent following the skill can actually read the security model reference the triage logic depends on.

.agents/skills/kedro-security-review/SKILL.md
```diff
@@ -0,0 +1,359 @@
...
+```bash
+LOCAL_RULESET="$(pwd)/.agents/skills/security-scan/rules/kedro-security-patterns.yml"
+```
+
+Run these in parallel:
+
+```bash
+(
+  "${SEMGREP_CMD[@]}" scan --metrics=off \
+    --config p/security-audit \
```
ElenaKhaustova, line 240: **Suggestion:** add path exclusions to every Semgrep command — `--exclude tests/ --exclude docs/ --exclude features/ --exclude .agents/` (or commit a `.semgrepignore` at repo root).

The triage doc itself says findings under these paths are "almost always `false_positive_or_informational`". On a full-codebase Kedro run, scanning them anyway produces hundreds of guaranteed-false-positive findings that the agent then has to read and classify one-by-one. This is probably the single biggest noise + runtime win in the skill.

.agents/skills/kedro-security-review/SKILL.md
```diff
@@ -0,0 +1,359 @@
...
+
+### Full codebase mode
+
+Use when the user asks to scan the repo, codebase, project, or current branch
+without narrowing to a PR.
+
+Target:
+- the current repo root
+
+### PR mode
```
ElenaKhaustova, line 114: **Suggestion:** add a one-question scope prompt at the start — "scanning Kedro framework code or a Kedro project?".

The triage doc already encodes this distinction in multiple rules (e.g. `kedro-yaml-unsafe-load` → `candidate_kedro_vulnerability` in framework code, `project_developer_responsibility` in project code). Asking once up front lets the agent pre-classify a large fraction of findings instead of resolving the question per finding.

ElenaKhaustova (commented):
## Kedro PR Review

### Overview
- **Critical:** 1 | **Suggestions:** 8 (plus 3 design questions in Notes)

- [ ] `.agents/skills/security-scan/SKILL.md:32, :249` — broken relative path to the security model doc (needs `../../../`, not `../../`). See inline comment.

### Suggestions (should improve before merge)

**High-value content improvements**
- [ ] `.agents/skills/security-scan/SKILL.md:189-225` — add path exclusions for `tests/`, `docs/`, `features/`, `.agents/`.
- [ ] `.agents/skills/security-scan/SKILL.md:78-105` — add a one-question scope prompt ("framework code or a Kedro project?") to pre-classify findings.
- [ ] `.agents/skills/security-scan/SKILL.md:298-307` — define a "next step" per classification bucket so the report is actionable, not just descriptive.
- [ ] `.agents/skills/security-scan/rules/kedro-security-patterns.yml:1` — pin a minimum Semgrep version so taint-mode rules don't silently no-op on older installs.
- [ ] `.agents/skills/security-scan/references/kedro-findings-triage.md:185-186` — wording about `yaml.safe_load` being "excluded by the rule automatically" is misleading.

**Cleanup**
- [ ] `.agents/skills/security-scan/SKILL.md:54-57, :129-132, :232-235` — same `uvx` / `~/.cache/uv` permissions paragraph repeated 3×; consolidate.
- [ ] `.agents/skills/security-scan/SKILL.md:13` — filler line; the frontmatter `description` already says it.
- [ ] `.agents/skills/security-scan/references/kedro-findings-triage.md` (file-level) — sibling skills use a flat `reference.md`; consider matching for consistency.

### Notes

**Structure / Cursor + Copilot compatibility**
PR is purely additive — `kedro-babysit/` and `review-kedro-pr/` are untouched. The `SKILL.md` frontmatter (`name:` + `description:`) matches the convention of the other two skills, so dispatch works the same way in both Cursor and Copilot. The new skill correctly reuses the shared `.agents/scripts/post_github_review.sh` extracted in the base branch.

**Design questions worth discussing (not blocking)**

1. **Ruleset breadth — is `p/security-audit` + `p/python` worth their noise?** The triage doc itself says `p/security-audit` produces high-volume false positives on framework codebases and `p/python` "rarely indicates Kedro vulnerabilities". On a full-codebase Kedro scan that's potentially 100–300 findings the agent has to read and classify one-by-one, almost all to be discarded. Worth considering whether to keep them (defense in depth) or drop them in favour of `p/secrets` + the custom Kedro rules + the manual review checks (which are where the actual Kedro-specific signal lives).

2. **Missing custom rule for `subprocess.run(..., shell=True)` with config-derived args.** Kedro CLI scaffolding constructs subprocess calls; same data→executable class of risk as the existing rules. Currently uncovered.

3. **Intended use case + report semantics — what does the skill *recommend*, not just classify?** The report currently produces classifications and severity headers but stops short of a recommendation. Two related decisions:
   - **Cadence:** one-off audit or recurring? If recurring, no baseline / acknowledged-findings mechanism exists today, so repeat runs re-triage everything. Consider Semgrep's `--baseline-commit` or a `.semgrepignore` with rule-id-specific suppressions.
   - **Severity → action:** should an ERROR-severity `candidate_kedro_vulnerability` produce a different recommendation (e.g. "block release / draft Security Advisory") than a WARNING-severity `false_positive_or_informational`? Right now they're rendered the same way. Pinning down "what does the skill tell the user to do" makes the output actionable rather than informational.

**General review**
The triage system is the real value-add — separating "Kedro's responsibility" from "developer's" from "deployment's" is exactly what generic scanners get wrong, and the manual review checks (new callable-key conventions, env-var → execution paths, external string → path) catch a class of issue static analysis fundamentally can't. Solid foundation, the above changes would push it from "useful audit tool" to "shippable recurring process".

**API usage / RELEASE.md**: not applicable — no changes under `kedro/`.

.agents/skills/security-scan/SKILL.md
```diff
@@ -0,0 +1,359 @@
...
+- highest Semgrep severities present
+
+For each finding, include:
+- file
+- line
+- rule id
+- Semgrep severity
+- Kedro classification
+- one-sentence reasoning
+- suggested next step
```
ElenaKhaustova: **Suggestion:** define what "suggested next step" actually is per classification bucket. Right now the report bullet asks for it but the buckets don't say what it should be. A small table would help, e.g.:

- `candidate_kedro_vulnerability` → draft Security Advisory / SECURITY.md report
- `project_developer_responsibility` → note in report only
- `deployment_or_environment_issue` → flag for ops docs
- `false_positive_or_informational` → suppress / add to `.semgrepignore`
- `needs_manual_review` → open a follow-up

Without this, the report classifies but stops short of being actionable.

.agents/skills/kedro-security-review/rules/kedro-security-patterns.yml
```diff
@@ -0,0 +1,122 @@
+rules:
```
ElenaKhaustova, line 2: **Suggestion:** pin a minimum Semgrep version with a top-of-file comment (e.g. `# requires semgrep >= 1.x`).

Taint mode + `focus-metavariable` are version-sensitive. Without a pin, an older `semgrep` on a contributor's machine can silently skip rules without error, and there's no obvious way to detect the regression.

.agents/skills/security-scan/references/kedro-findings-triage.md
```diff
@@ -0,0 +1,194 @@
...
+Fires on `yaml.load()` without a safe Loader. Kedro should always use
+`yaml.safe_load()` or an explicit safe Loader in framework code, because YAML
+config files are user-controlled data.
+
+- If found in **Kedro framework code** (`kedro/` package): classify as
+  `candidate_kedro_vulnerability` — unsafe YAML deserialization on
+  user-controlled config can execute arbitrary Python objects.
+- If found in **project developer code**: classify as
+  `project_developer_responsibility`.
+- `yaml.safe_load`, `yaml.SafeLoader`, and `yaml.CSafeLoader`
```
ElenaKhaustova: **Suggestion:** wording here is slightly misleading. `yaml.safe_load` isn't "excluded by the rule automatically" — it's simply never matched in the first place, because the rule's pattern is `yaml.load(...)`, a different function. The `pattern-not` exclusions only cover `yaml.SafeLoader` / `yaml.CSafeLoader` as `Loader=` args.

Worth clarifying so readers don't infer an explicit exclusion that doesn't exist.

.agents/skills/kedro-security-review/SKILL.md
```diff
@@ -0,0 +1,359 @@
...
+  - `.agents/skills/security-scan/rules/kedro-security-patterns.yml`
+- Temporary working directory:
+  - create with `mktemp -d`
+  - delete at the end unless the user explicitly asks to keep artifacts
+
+Always use `--metrics=off`.
+
+## Runtime note
+
+If a `uvx` or `uv tool run` Semgrep command fails with a permissions error on a
```
ElenaKhaustova, line 73: **Suggestion:** this `uvx` / `~/.cache/uv` permissions guidance is repeated three times (here at lines 54-57, again at lines 129-132 in step 1, and again at lines 232-235 in step 4). Consolidate into a single paragraph in step 1 ("Resolve runtime") and have the others reference it.

.agents/skills/security-scan/SKILL.md
```diff
@@ -0,0 +1,359 @@
...
+  Run a Kedro security scan on the full codebase or just a pull request. This
+  skill always runs Semgrep first and then evaluates the findings against the
+  Kedro security model. Use when the user says things like "run security scan
+  on the full codebase" or "run security scan on this PR". Produce one final
+  report only, in chat by default or posted to GitHub when explicitly asked.
+---
+
+# Security Scan
+
+Use this as the user-facing Kedro security skill.
```
ElenaKhaustova: **Suggestion:** filler line — the frontmatter `description` field already says this skill is the user-facing Kedro security skill. Drop the line to cut noise.

.agents/skills/kedro-security-review/reference.md
```diff
@@ -0,0 +1,194 @@
+# Kedro findings triage
```
ElenaKhaustova, line 1: **Suggestion:** sibling skills use a single flat `reference.md` per skill (`kedro-babysit/reference.md`, `review-kedro-pr/reference.md`). Right now `references/` (plural folder) holds a single file. Either flatten to `reference.md` for consistency, or keep the folder only if more reference files are planned.

.agents/skills/security-scan/reference.md
```diff
@@ -0,0 +1,213 @@
...
+
+### `p/secrets`
+
+_Detects hardcoded credentials, tokens, and API keys in source files._
+
+Triage in order: (1) check the file path — `tests/`, `docs/`, `features/`,
+`.agents/` paths are almost always `false_positive_or_informational`;
+(2) check for a placeholder value (`"YOUR_API_KEY"`, `"xxx"`, `"<token>"`);
+(3) if the string looks like a real secret in production code, classify as
+`needs_manual_review`.
```
ElenaKhaustova: **Suggestion (stale after the new `--exclude` flags):** step (1) — "check the file path — `tests/`, `docs/`, `features/`, `.agents/` paths are almost always `false_positive_or_informational`" — is now unreachable, because `SEMGREP_EXCLUDE` in `SKILL.md` already filters those paths out before Semgrep emits findings.

Suggested fix: drop step (1), or replace it with something like *"the scan already excludes these paths via `--exclude`; if a finding still appears in one of them, treat it as a misconfiguration"*.

Minor related point: the `p/security-audit` blurb (line 99) says "Expect high volume" — that expectation was set pre-exclusions, so the triage tone may now overstate the noise level too.

.agents/skills/kedro-security-review/reference.md
```diff
@@ -0,0 +1,213 @@
...
+
+Use when deeper context is required before you can classify confidently.
+
+**Suggested next step:** open a follow-up issue or assign a human reviewer;
+do not auto-dismiss.
+
+## Manual review checks
+
+Run these after Semgrep, regardless of whether any rule fired.
+Semgrep only covers known patterns. These checks catch the unknown ones.
```
ElenaKhaustova, line 67: **Suggestion:** the **Manual review checks** (this section + the three checks below) tell the agent to grep / search across "the scanned files", but unlike the Semgrep phase, they don't apply `SEMGREP_EXCLUDE`. As written, manual checks #2 (env-var → execution) and #3 (path construction) will keep flagging hits in `tests/fixtures/`, `docs/examples/`, etc.

One sentence at the top of this section would keep both phases consistent, e.g.: *"Apply the same path exclusions as the Semgrep scan (`tests/`, `docs/`, `features/`, `.agents/`)."*

.agents/skills/security-scan/SKILL.md
```diff
@@ -0,0 +1,394 @@
...
+  report only, in chat by default or posted to GitHub when explicitly asked.
+---
+
+# Security Scan
+
+Two modes:
+- full codebase
+- PR-only
+
+Confirm scan scope first (see `## Scan scope`), then run the numbered Workflow
```
ElenaKhaustova: **Suggestion:** the skill now has three orthogonal axes the agent must pick before scanning:

| Axis | Values | Section |
|---|---|---|
| Target source | full codebase / PR | `## Mode selection` |
| Scan scope | Kedro framework / Kedro project | `## Scan scope` |
| Delivery | chat / post | `## Delivery modes` |

That's 2×2×2 = 8 valid configurations. Each section is fine in isolation, but they're three separate "mode"-flavoured sections with no top-level summary. An agent reading top-to-bottom might miss one (especially the new `Scan scope`, which is the easiest to forget on a re-run).

A 4-line preamble in this overview would make the decision tree explicit, e.g.:

```markdown
Before scanning, decide three things:
1. Target source — full codebase or PR (see `## Mode selection`)
2. Scan scope — Kedro framework or Kedro project (see `## Scan scope`)
3. Delivery — chat or post (see `## Delivery modes`)
```

ElenaKhaustova (commented):
## Re-review follow-up

Thanks for the fixes — the Critical and all 8 Suggestions from the previous review are resolved (broken docs path, path exclusions, framework/project scope prompt, next-step-per-bucket, Semgrep version pin, `yaml.safe_load` wording, sandbox guidance consolidation, filler line, flattened `reference.md`).

Fresh pass on the new state surfaced **3 internal-consistency issues introduced by this round of changes**. All small but worth a tidy-up before merge.

- [ ] `reference.md` p/secrets triage step (1) is now unreachable after adding `--exclude tests/ docs/ features/ .agents/`.
- [ ] Manual review checks in `reference.md` don't share the new exclusions — grep-based checks will keep flagging excluded paths.
- [ ] `SKILL.md` overview — three orthogonal "modes" (target / scope / delivery) are each documented but never summarised together.

See inline comments for details.

ElenaKhaustova:
Before merging, I would also love your take on the 3 design questions from the first review (please see them above):

1. Ruleset breadth — keeping `p/security-audit` + `p/python` for defence-in-depth, or open to dropping them?
2. `subprocess.run(..., shell=True)` rule — add now or follow-up PR?
3. Severity → action — per-bucket "next step" handles classification well, but severity stays flat (ERROR and WARNING `candidate_kedro_vulnerability` get the same recommendation). Intentional? And related: one-off audit or recurring (which would motivate `--baseline-commit`)?

They are not blocking, just want to make sure all this is intentional.

rashidakanchwala:
> Before merging, I would also love your take on the 3 design questions from the first review (please see them above):
> 
> 1. Ruleset breadth — keeping `p/security-audit` + `p/python` for defence-in-depth, or open to dropping them?

Keeping both for now. Without them, Semgrep doesn’t raise anything; with them, it does introduce quite a bit of noise. That said, the triage step is what makes the noise manageable as the agent reviews every finding and classifies it, so users only see the actionable issues.
I think it’s still worth keeping them for defence-in-depth. Also, p/python does catch unsafe pickle usage in framework code, which is genuinely relevant for Kedro. If they end up adding no meaningful signal in practice, we can always remove them later.

> 2. `subprocess.run(..., shell=True)` rule — add now or follow-up PR?
Added in this PR. New rule kedro-subprocess-shell-injection is in rules/kedro-security-patterns.yml, with matching triage guidance in reference.md. Severity is ERROR, and it skips string/list literals so it only fires on non-literal commands with shell=True.

> 3. Severity → action — per-bucket "next step" handles classification well, but severity stays flat (ERROR and WARNING `candidate_kedro_vulnerability` get the same recommendation). Intentional? And related: one-off audit or recurring (which would motivate `--baseline-commit`)?

Fixed. Inside candidate_kedro_vulnerability, severity now drives the recommendation:

ERROR → flag prominently, recommend fixing before merge.
WARNING → recommend investigating, doesn't need to block the merge.
The skill only reports — it doesn't enforce anything. Sticking with a one-off manual audit for now; --baseline-commit is a follow-up if/when this moves into CI.

.agents/skills/security-scan/SKILL.md
```diff
@@ -0,0 +1,438 @@
+---
+name: security-scan
```
ElenaKhaustova: **Critical — naming creates dangling cross-skill references.**

On `origin/main`, **both** existing skills already route security findings to a skill named `kedro-security-review`:

- `.agents/skills/review-kedro-pr/SKILL.md:171` — *"Security issues — handled by the separate `kedro-security-review` skill."*
- `.agents/skills/kedro-babysit/SKILL.md:171` — *"Security review — defers to a future `kedro-security-review` skill."*

Once this PR merges, both of those references will dangle (no skill named `kedro-security-review` exists), and the new skill won't be reachable through the routing paths the sibling skills already document.

The naming convention is also asymmetric:

| Skill | Name |
|---|---|
| `kedro-babysit` | `kedro-` prefix |
| `review-kedro-pr` | `kedro` infix |
| `security-scan` (new) | no `kedro` anywhere |

Two ways to fix:

- **(Preferred)** Rename this skill `security-scan` → `kedro-security-review` (matches the prefix convention and the name already referenced by both sibling skills). One rename inside this PR, no edits to other skills.
- Or keep `security-scan` and update both sibling skills to point to it — touches three files instead of one and expands PR scope.

The rename is `name:` in this frontmatter + the directory name `.agents/skills/security-scan/` → `.agents/skills/kedro-security-review/` + the absolute path reference at SKILL.md:59 (`.agents/skills/security-scan/rules/...`) + the `$(pwd)/...` reference at step 4 (line 231).

.agents/skills/security-scan/SKILL.md
```diff
@@ -0,0 +1,438 @@
...
+- highest Semgrep severities present
+
+**Only itemise actionable findings** in the Findings section:
+- `candidate_kedro_vulnerability`
+- `needs_manual_review`
+
+**Do not itemise** findings in these buckets — they are reflected in the
+classification summary counts and that is sufficient:
+- `false_positive_or_informational`
+- `project_developer_responsibility`
```
ElenaKhaustova: **Suggestion — "actionable" set is scope-blind, which hides project-scope findings.**

The `## Scan scope` section (lines 40–45) routes findings differently per scope:

> `kedro-yaml-unsafe-load` in `kedro/` is `candidate_kedro_vulnerability`; the same rule in project code is `project_developer_responsibility`.

And `reference.md`'s rule-by-rule triage repeats this pattern everywhere (`kedro-subprocess-shell-injection` step 3, `kedro-taint-string-param-to-path` step 1, `p/python` pickle guidance, etc.) — project-scope findings consistently land in `project_developer_responsibility`.

But this Reporting block always collapses `project_developer_responsibility` to a count. Putting the two together:

| Scope | Most common actionable bucket | Itemised? |
|---|---|---|
| Framework | `candidate_kedro_vulnerability` | ✅ Yes |
| Project | `project_developer_responsibility` | ❌ No — count only |

So a project owner running the skill on their pipeline gets a report saying "5 findings in `project_developer_responsibility`" with no file paths, no line numbers, no rule IDs, no suggested fixes. The whole reason they ran the skill is hidden behind a count. Same issue applies to the Post-mode rule at line 434 ("Do not post inline comments for ... non-actionable buckets").

Two possible fixes:

- **Make the actionable set scope-dependent:** in `Kedro project` scope, also itemise `project_developer_responsibility`. Something like:

  ```markdown
  **Actionable buckets** (itemise in Findings; post inline in PR mode):
  - Always: `candidate_kedro_vulnerability`, `needs_manual_review`
  - In Kedro-project scope: also `project_developer_responsibility`
    (this is the primary actionable bucket for the project owner)
  ```

- Or, if project scope is intentionally narrow (only meant to flag Kedro-vulnerability exposure through project code), say so explicitly in the `## Scan scope` section — something like *"in project scope, this skill only surfaces findings that may indicate a Kedro framework vulnerability triggered by project code; for general security audit of project code, use a standalone Semgrep run."*

Either resolves the inconsistency.

.agents/skills/kedro-security-review/SKILL.md
```diff
@@ -0,0 +1,444 @@
...
+
+Read these before triaging findings:
+- [../../../docs/about/security_model.md](../../../docs/about/security_model.md)
+- [reference.md](reference.md)
+
+## Scan scope
+
+Before scanning, ask once:
+
+> Are we scanning **Kedro framework code** (this repository's `kedro/` package)
```
ankatiyar: We don't have kedro/package - it's kedro/kedro I think

ankatiyar (approved):
Minor comment but LGTM

lrcouto (approved):
Tested it here, looks like it works well. The structure of the skill itself looks good as well!

PR #5583, opened 2026-05-29
Wire selective pipeline loading to CLI

### LINKED ISSUE #5549: Wire selective pipeline loading to CLI `--pipeline` usage
## Description

As an outcome to the spike work - https://github.com/kedro-org/kedro/issues/5406
Addresses - #5207 

Kedro already has selective pipeline loading support through find_pipelines(pipelines_to_find=...), introduced in PR #5401. When called with a specific pipeline name, for example ["data_science"], it only imports that pipeline directory instead of importing every pipeline in the project.

However, this capability is not fully wired into CLI commands that already accept a pipeline argument. As a result, commands such as kedro run --pipeline X, kedro registry describe X, or catalog/inspection commands may still trigger full project pipeline loading before the selected pipeline is used.

This is a problem for projects with multiple pipelines that have different dependency footprints. A user trying to run or inspect only one pipeline can still be blocked by missing dependencies from unrelated pipelines.

## Context

Most Kedro projects using the default registry pattern define pipelines like this:

```python
# my_project/pipeline_registry.py

def register_pipelines():
    return find_pipelines()
```

Today, find_pipelines() is called without a filter, so it loads everything.

The selective loading capability already exists: find_pipelines(pipelines_to_find=["data_science"]) imports only my_project/pipelines/data_science/. The missing piece is to pass the CLI-requested pipeline names into the project pipeline registry before the global pipelines object is first accessed.

This would make targeted CLI commands behave more efficiently and predictably:

```python
kedro run --pipeline data_science
kedro registry describe data_science
kedro catalog describe-datasets --pipeline data_science
kedro catalog resolve-patterns --pipeline data_science
```

### Step 1 — `set_requested()` on `_ProjectPipelines`

`_ProjectPipelines` stores which pipelines are needed before the first dict access. Invalidates the cache if the filter changes so a subsequent full run still loads everything.

```python
# kedro/framework/project/__init__.py

class _ProjectPipelines(MutableMapping):
    def __init__(self) -> None:
        self._pipelines_module: str | None = None
        self._is_data_loaded = False
        self._content: dict[str, Pipeline] = {}
        self._requested_pipelines: list[str] | None = None  # NEW

    def set_requested(self, pipeline_names: list[str] | None) -> None:
        """Store which pipelines are needed before first dict access.
        Invalidates the cache if the filter changes.
        """
        if self._requested_pipelines != pipeline_names:
            self._is_data_loaded = False
            self._content = {}
        self._requested_pipelines = pipeline_names
```

### Step 2 — `find_pipelines()` reads `_requested_pipelines` with CLI precedence

The CLI-set filter (via `set_requested()`) takes precedence over the explicit `pipelines_to_find` kwarg. When no CLI filter is active (`_requested_pipelines` is `None`), the explicit kwarg is used as before.

Since `find_pipelines()` and the `pipelines` global are in the **same file**, there is no circular import risk. **No changes to `_load_data()` are needed for Scenario 1.**

```python
# kedro/framework/project/__init__.py

def find_pipelines(raise_errors: bool = False, pipelines_to_find: list[str] | None = None):
    # CLI-set filter takes precedence; falls back to the explicit kwarg.
    requested_pipelines = (
        pipelines._requested_pipelines
        if pipelines._requested_pipelines is not None
        else pipelines_to_find
    )
    load_all = requested_pipelines is None or "__default__" in requested_pipelines
    pipeline_filter: set[str] | None = None if load_all else set(requested_pipelines)
    # ... rest of function unchanged
```

### Step 3 — CLI commands call `set_requested()` before pipeline access

`set_requested()` must be called **before** any dict access on `pipelines`. For session-based commands, this means before `_create_session()` — the first catalog or pipeline access inside the session triggers `_load_data()`.

#### Current status

| Command | File | Call
|---------|------|------|
| `kedro run --pipeline X` | `session.py:348` | `pipelines.set_requested(names)` | 
| `kedro registry describe X` | `registry.py:38` | `pipelines.set_requested([name])` |
| `kedro catalog describe-datasets -p X` | `catalog.py:56` | `pipelines.set_requested([pipeline])` | 
| `kedro catalog resolve-patterns -p X` | `catalog.py:104` | `pipelines.set_requested([pipeline])` | 
| `inspection: get_project_snapshot(pipelines=["X"])` | `snapshot.py:136` | requires new `pipeline_names` parameter first |

## Limitation — Custom `register_pipelines` Without `find_pipelines()`

This fix only works for projects using `find_pipelines()`. Projects with a hand-written `register_pipelines` receive no benefit — `find_pipelines()` is never called so `_requested_pipelines` is never read. This includes top-level imports in `pipeline_registry.py`:

```python
# pipeline_registry.py — these fire BEFORE register_pipelines() is called
from my_project.pipelines.reporting import create_pipeline as reporting_pipeline
from my_project.pipelines.data_science import create_pipeline as ds_pipeline

def register_pipelines():
    return {"reporting": reporting_pipeline(), "data_science": ds_pipeline()}
```

**Options for affected projects:**

1. **Switch to `find_pipelines()`** — the default layout since Kedro 0.18.3.

2. **Read `_requested_pipelines` manually** — gate local imports behind the stored filter:
   ```python
   from kedro.framework.project import pipelines as _pipelines

   def register_pipelines():
       requested = _pipelines._requested_pipelines  # None means load all
       result = {}
       if requested is None or "data_ingestion" in requested:
           from my_project.pipelines import data_ingestion
           result["data_ingestion"] = data_ingestion.create_pipeline()
       if requested is None or "reporting" in requested:
           from my_project.pipelines import reporting
           result["reporting"] = reporting.create_pipeline()
       return result
---


## Acceptance Criteria

- kedro run --pipeline X sets the requested pipeline filter before the session or global pipelines mapping is accessed.
- kedro registry describe X sets the requested pipeline filter before pipeline registry access.
- kedro catalog describe-datasets -p X sets the requested pipeline filter before pipeline/catalog access.
- kedro catalog resolve-patterns -p X sets the requested pipeline filter before pipeline/catalog access.
- find_pipelines() uses pipelines._requested_pipelines when set.
- find_pipelines(pipelines_to_find=...) continues to work as before when no CLI filter is set.
- A subsequent unfiltered run can still load all pipelines.
- No changes are required to _load_data() for this scenario.
- The limitation for custom register_pipelines() implementations is documented.

lrcouto:
## Description

https://github.com/kedro-org/kedro/issues/5549

Wires the existing `find_pipelines(pipelines_to_find=...)` selective loading capability (introduced in #5401) to the `kedro run --pipelines` CLI flag so that only the requested pipeline modules are imported instead of all of them.

The global pipelines object (`_ProjectPipelines`) is a lazy-loading singleton available throughout the framework. It now holds a `_requested_pipelines` filter, set via a new `set_requested(pipeline_names)` method. When the filter changes, the lazy-load cache is invalidated so the next dict access triggers a fresh load with the new scope.

`find_pipelines()` reads `pipelines._requested_pipelines` before consulting its own `pipelines_to_find` kwarg. The CLI-set filter takes precedence. When the filter is None, behavior is identical to before.

## Checklist


kedro/framework/cli/registry.py
ankatiyar, line 47: Will this not be a problem for listing the `existing_pipelines` here in the error?

kedro/framework/project/__init__.py
```diff
@@ -226,6 +227,17 @@ def _load_data(self) -> None:
         self._content = project_pipelines
         self._is_data_loaded = True
 
+    def set_requested(self, pipeline_names: list[str] | None) -> None:
+        """Store which pipelines should be loaded on the next dict access.
+
+        Invalidates the cache when the filter changes so that a subsequent
+        access with a different (or absent) filter re-runs ``_load_data``.
+        """
+        if self._requested_pipelines != pipeline_names:
```
ankatiyar: This should be a set comparison, order of the list might be a problem

kedro/framework/cli/catalog.py
```diff
@@ -42,7 +42,7 @@ def catalog() -> None:
     callback=split_string,
 )
 @click.pass_obj
-def describe_datasets(metadata: ProjectMetadata, pipeline: str, env: str) -> None:
+def describe_datasets(metadata: ProjectMetadata, pipeline: list[str], env: str) -> None:
```
ankatiyar, line 45: This might be a breaking change technically since it has `--pipeline` which takes a single string but now we're changing it to list of strings

lrcouto: Oh this annotation was wrong before, it's already been a list for a while. The input from `--pipeline` is converted to a list with a single item inside the `run()` function.

rashidakanchwala (commented):
## Kedro PR Review

### Overview
- **Critical:** 1 | **Suggestions:** 2 | **Nits:** 2

- [ ] `docs/` — No documentation update for the new selective loading behaviour or its limitation (projects using hand-written `register_pipelines` without `find_pipelines()` get no benefit). The checklist item is unchecked and `docs/` is absent from the diff.

- [ ] `tests/framework/project/test_pipeline_discovery.py` — The `"__default__"` edge case is untested. `"__default__"` has special semantics: `load_all = True` if `"__default__" in effective`, meaning `kedro run --pipeline __default__` should load everything. That branch is never explicitly tested with the new filter path.
- [ ] `features/run.feature` — You already have behave infrastructure that spins up a real project and runs CLI commands end-to-end, but no scenario exercises `kedro run --pipeline <specific_pipeline>` through the selective loading path. A new scenario here would give much stronger confidence than the unit tests alone, since it exercises the real `find_pipelines()` → disk import chain on an actual package.

### Nits
- [ ] `kedro/framework/project/__init__.py:L243` — `set_requested` stores the caller's list by reference. In practice current call sites are safe, but since this is a public method on the module-level singleton, a defensive copy is more robust: `self._requested_pipelines = list(pipeline_names) if pipeline_names is not None else None`
- [ ] `kedro/framework/cli/catalog.py:L56`, `kedro/framework/cli/registry.py:L38`, `kedro/framework/session/session.py:L345` — The real constraint is "must be called before `load_context()`" (which triggers the first `pipelines` dict access), not before session creation specifically. A brief comment at each call site would make the invariant explicit — especially since `catalog.py` places the call before `_create_session()` while `session.py` places it inside `run()` after session creation. Both are correct but the inconsistency can confuse future maintainers.

### Notes
- **Existing inline comments:** All three are already addressed in the code — no action needed.
- **General review:** Logic is correct. Cache invalidation, the `set_requested(None)` fallback in the registry error path, and the `configure()` reset are all sound.
- **API usage:** No violations. Pattern is consistent with the existing lazy-loading design.

kedro/framework/project/__init__.py
```diff
@@ -226,6 +227,21 @@ def _load_data(self) -> None:
...
+        access with a different (or absent) filter re-runs ``_load_data``.
+
+        Args:
+            pipeline_names: Names of the pipelines to load selectively, or
+                ``None`` to load all registered pipelines.
+        """
+        if set(self._requested_pipelines or []) != set(pipeline_names or []):
+            self._is_data_loaded = False
+            self._content = {}
+        self._requested_pipelines = pipeline_names
```
rashidakanchwala: **Nit:** `set_requested` stores the caller's list by reference. In practice, current call sites are safe (all pass fresh literals or lists that aren't mutated before the next dict access), but since this is a public method on the module-level singleton, a defensive copy would be more robust:
```python
self._requested_pipelines = list(pipeline_names) if pipeline_names is not None else None
```

rashidakanchwala (changes requested):

lrcouto:
Thanks for the review @rashidakanchwala . I've added a description of the feature on docs, added the suggested tests, and addressed the suggestion for adding comments as well.

docs/build/run_a_pipeline.md
```diff
@@ -194,6 +194,25 @@ kedro run --pipeline=my_pipeline
 
 Further information about `kedro run` can be found in the [Kedro CLI documentation](../getting-started/commands_reference.md#kedro-run).
 
+## Selective pipeline loading
+
+When you run `kedro run --pipeline <name>`, Kedro imports the pipeline modules required for that run and skips unrelated modules. This speeds up startup time in large projects.
```
ravi-kumar-pilla: [Nit] We might need to update all our docs with using `--pipelines` in a follow up ticket.

features/run.feature
```diff
@@ -77,6 +77,14 @@ Feature: Run Project
     Then I should get a successful exit code
     And the logs should show that 4 nodes were run
 
+  Scenario: Run kedro run with selective pipeline loading
+    Given I have prepared a config file
+    And I have run a non-interactive kedro new with starter "default"
+    When I execute the kedro command "run --pipeline data_engineering"
+    Then I should get a successful exit code
+    And the logs should show that 1 nodes were run
+    And the logs should show that "split_data_node" was run
```
ravi-kumar-pilla, line 95: I think this happens even now ? May be we need to uninstall some module corresponding to other pipeline, lets say `data_science` and the kedro run with the pipeline flag should still work fine, that would be a nice e2e test to have.

kedro/framework/cli/catalog.py
ravi-kumar-pilla: I think it works via the callback present at line 42. The pipeline is received as a `str` but the callback makes it to split, so may be we can revert this to what it was before.

kedro/framework/cli/catalog.py
```diff
@@ -52,10 +52,12 @@ def describe_datasets(metadata: ProjectMetadata, pipeline: str, env: str) -> Non
     - `factories`: Datasets resolved from dataset factory patterns.\n
     - `defaults`: Datasets that do not match any pattern or explicit definition.\n
     """
+    p = pipeline or None
```
ravi-kumar-pilla: [Nit] may be we can avoid another variable by adding an if condition ? - https://github.com/kedro-org/kedro/pull/5532/changes#diff-f35591d9d091de85d12ec14bc566825cb43aad97b25f2aedb744de44cfd9363c

```    
if pipeline:
        pipelines.set_requested(pipeline)
```

ravi-kumar-pilla: Ignore this, as I see similar pattern of using another variable in `resolve_patterns`

ravi-kumar-pilla:
Hi @lrcouto ,

Left some minor comments but overall looks pretty good. I will do testing e2e tomorrow. Thank you

kedro/framework/project/__init__.py
```diff
@@ -501,9 +520,14 @@ def find_pipelines(  # noqa: PLR0912, PLR0915
             "Call 'configure_project' first."
         )
 
-    # Determine if specific pipelines were requested
-    load_all = pipelines_to_find is None or "__default__" in pipelines_to_find
-    requested_pipelines: set[str] | None = None if load_all else set(pipelines_to_find)  # type: ignore[arg-type]
+    # CLI-set filter takes precedence; falls back to the explicit kwarg.
+    effective = (
```
merelcht: Nit: can we use a more explanatory name? It's not super clear what `effective` means here.

docs/build/run_a_pipeline.md
merelcht: Very good point! Let's at least only use `--pipelines` for the docs that are added in this PR and then as a follow up let's change any other outdated mentions.

lrcouto:
Changed references on docs from `kedro run --pipeline" to `kedro run --pipelines"

merelcht (approved):
Thanks @lrcouto ! ⭐

docs/build/run_a_pipeline.md
```diff
@@ -187,15 +187,34 @@ To run the pipeline by its name, you need to add your new pipeline to the `regis
...
 ```
 
 !!! note
-    If you specify `kedro run` without the `--pipeline` option, it runs the `__default__` pipeline from the dictionary returned by `register_pipelines()`.
+    If you specify `kedro run` without the `--pipelines` option, it runs the `__default__` pipeline from the dictionary returned by `register_pipelines()`.
 
 
 Further information about `kedro run` can be found in the [Kedro CLI documentation](../getting-started/commands_reference.md#kedro-run).
 
+## Selective pipeline loading
```
ankatiyar, line 205: There's already a PR for updating the docs with `--pipelines` here: https://github.com/kedro-org/kedro/pull/5599

I would prefer if we kept the `--pipelines` in the part of docs that we're adding in this PR instead of all the docs.

kedro/framework/cli/registry.py
ankatiyar: Any resolution on this?

kedro/framework/cli/registry.py
```diff
@@ -35,8 +35,12 @@ def describe_registered_pipeline(
     """Describe a registered pipeline by providing a pipeline name.
     Defaults to the `__default__` pipeline.
     """
+    # Must be called before pipelines.get(), which triggers the first pipelines dict access.
+    pipelines.set_requested([name])
     pipeline_obj = pipelines.get(name)
     if not pipeline_obj:
+        # Reset to None so keys() loads all pipelines for the error message.
+        pipelines.set_requested(None)
```
ankatiyar, line 43: This fix also needs to be applied in `session` and `service_session` so that the suggestions work.

kedro/framework/session/session.py
ankatiyar, line 354: Add `set_requested` here to `None` so that the `pipelines.keys()` is a full list of pipelines for a proper suggestion message. Same for service session

ElenaKhaustova (changes requested):
Thank you @lrcouto!

Changes look mostly good, but there's still a critical issue flagged by @ankatiyar.

Could you please also close resolved comments and reply to the rest, so it is clear for the reviewers what the status is.

kedro/framework/session/service_session.py
```diff
@@ -159,6 +159,8 @@ def run(  # noqa: PLR0913
         self._logger.info("Run ID: %s", run_id)
         save_version = run_id
 
+        # Must be called before load_context(), which triggers the first pipelines dict access.
+        pipelines.set_requested(pipeline_names or None)
```
ElenaKhaustova, line 163: `kedro/framework/session/session.py:L353` and `kedro/framework/session/service_session.py:L169` — Error-path suggestions are broken by the active filter. When a user runs `kedro run --pipelines nonexistent: set_requested(["nonexistent"])` →`find_pipelines()` only looks for a "nonexistent" module → `_conten`t ends up empty → `pipelines["nonexistent"]` raises `KeyError` → `get_close_matches(name, pipelines.keys())` gets an empty set → no suggestions shown.

`registry.py` already handles this correctly by calling `pipelines.set_requested(None)` before `pipelines.keys()` in the error branch. The same fix needs to be applied in both `session.py` and `service_session.py`. This was flagged by @ankatiyar before and is still unresolved.

features/run.feature
```diff
@@ -77,6 +77,23 @@ Feature: Run Project
     Then I should get a successful exit code
     And the logs should show that 4 nodes were run
 
+  Scenario: Run kedro run with selective pipeline loading
+    Given I have prepared a config file
+    And I have run a non-interactive kedro new with starter "default"
+    When I execute the kedro command "run --pipeline data_engineering"
```
ankatiyar:
```suggestion
    When I execute the kedro command "run --pipelines data_engineering"
```

features/run.feature
```diff
@@ -77,6 +77,23 @@ Feature: Run Project
...
+    Then I should get a successful exit code
+    And the logs should show that 1 nodes were run
+    And the logs should show that "split_data_node" was run
+
+  Scenario: Run kedro run with selective pipeline loading skips unrelated pipeline modules
+    Given I have prepared a config file
+    And I have run a non-interactive kedro new with starter "default"
+    And I have broken the "data_science" pipeline with an unimportable dependency
+    When I execute the kedro command "run --pipeline data_engineering"
```
ankatiyar:
```suggestion
    When I execute the kedro command "run --pipelines data_engineering"
```

ankatiyar (approved):
Last couple nits, sorry! LGTM @lrcouto

lrcouto:
Thank you for the reviews!

@ElenaKhaustova @rashidakanchwala I've implemented the requested changes, LMK if there's anything else!

ElenaKhaustova (approved):
Looks good now, thank you!

Could you please also fix one remaining `--pipeline` entry:

https://github.com/kedro-org/kedro/blob/selective-pipeline-loading/docs/build/pipeline_registry.md#L76

`kedro run --pipeline data_engineering --tags data_engineering` should be `kedro run --pipelines data_engineering --tags data_engineering`.

rashidakanchwala (approved):

PR #5608, opened 2026-06-18
fix(cli): reject project names that shadow stdlib modules or keywords

### LINKED ISSUE #5607: kedro new accepts project names that shadow Python stdlib modules / keywords, silently producing a broken project
## Description

`kedro new` validates the project name against a character-set regex only. It never checks whether the derived Python package name collides with a standard-library module or is a Python keyword. When it does (for example `email`, `json`, `string`, `csv`, or a keyword like `import`), the project is created with a success message but is broken from the start. `kedro run` crashes at pipeline-registry import, and `settings.py` is silently ignored.

> Raised on Slack; a maintainer suggested opening an issue.

## Context

A user building an email-processing pipeline will naturally run `kedro new --name=email`. Kedro prints "Congratulations! Your project has been created", then every `kedro run` dies with a `ModuleNotFoundError` that gives no hint the cause is the project name they chose minutes earlier. The failure is far removed from its cause in both time and message, which makes it hard to diagnose. `email`, `json`, `csv`, `string`, `code`, `io`, `time`, and `types` are all stdlib modules and all plausible project names.

## Steps to Reproduce

```
kedro new --name=email --tools=none --example=n
cd email
kedro run
```

Reproduces identically with `json` and any other stdlib top-level module name, and with Python keywords (`import`, `class`).

## Expected Result

`kedro new` should reject, or at minimum warn, when the derived package name is a Python keyword or shadows a stdlib module, at creation time. The user should be told immediately rather than via a cryptic runtime error.

## Actual Result

Two distinct failures.

**1. Hard crash.** `kedro run` fails even though `src/email/pipeline_registry.py` exists:

```
ModuleNotFoundError: No module named 'email.pipeline_registry'
```
```
.../kedro/framework/session/session.py:351 in run
.../kedro/framework/project/__init__.py:207 in _get_pipelines_registry_callable
.../importlib/__init__.py:90 in import_module
```

**2. Silent failure.** `settings.py` exists (1379 bytes) but is silently ignored:

```
WARNING  No 'settings.py' found, defaults will be used.
```

**Control (proves the name is the sole cause).** An identically created project with a non-colliding name imports its registry fine and only fails later with the expected empty-project error:

```
# kedro new --name=controltest  (same flags)
ValueError: Pipeline contains no nodes after applying all provided filters.
```

`controltest` gets past the exact line where `email` and `json` crash.

## Screenshots

**1. `kedro run` in the `email` project crashes at registry import with `ModuleNotFoundError: No module named 'email.pipeline_registry'`:**

Image

**2. The same command in a normally named project (`controltest`) gets past that line and fails only with the expected empty-project `ValueError`:**

Image

**3. `src/email/pipeline_registry.py` exists in the project. It is shadowed by the stdlib `email`, not missing:**

Image

## Root Cause

In `kedro/framework/cli/starters.py`, `_validate_input_with_regex_pattern`:

```python
"project_name": {
    "regex": r"^[\w -]{2,}$",
    "error_message": "... only alphanumeric symbols, spaces, underscores and hyphens ...",
}
```

This checks the character set only. The derived package name is never validated against `keyword.iskeyword()` or `sys.stdlib_module_names`.

Why stdlib wins: Kedro imports modules like `email` and `json` during its own startup, so by the time it builds the project package path, `sys.modules['email']` already points at the stdlib module. `importlib.import_module("email.pipeline_registry")` uses that cached parent, finds no `pipeline_registry` submodule, and fails. The project's own `src/email/` is shadowed and never gets a chance. The same shadowing makes `import email.settings` fail, which Kedro swallows into the misleading "No settings.py found" fallback.

## Proposed Fix

After deriving the package name in project-name validation, reject or warn if:

- `keyword.iskeyword(package_name)`, or
- `package_name in sys.stdlib_module_names`  (Python 3.10+)

with a clear message at creation time.

This matches how other packaging and scaffolding tools guard against the same class of problem:

- **Django** (`django-admin startproject`) validates the project name and explicitly rejects names that conflict with an existing Python module, failing with "conflicts with the name of an existing Python module and cannot be used as a project name. Please try another name." It also rejects Python keywords and non-identifiers, in `TemplateCommand.validate_name`.
- **npm** (via `validate-npm-package-name`) rejects package names that match Node.js core/builtin module names, so you cannot create a package called `http` or `events`.
- **cookiecutter** Python templates commonly ship a `hooks/pre_gen_project.py` that aborts generation if the module name is not a valid identifier or is a Python keyword.

Kedro is currently the outlier in accepting such a name and deferring the failure to runtime. Happy to open a PR.

## Your Environment

* Kedro version used (`kedro -V`): 1.4.0
* Python version used (`python -V`): 3.12.7 (also reproduces on 3.11)
* Operating system and version: Windows 11 (the mechanism is import-cache based, so OS-independent)

RudraDudhat2509:
## Description

Fixes #5607.

`kedro new` derives the Python package name from the project name but never checks whether it collides with a Python keyword or a standard library module. Names like `email`, `json`, `string`, or a keyword like `import` scaffold successfully, then break at runtime:

- `kedro run` fails at pipeline-registry import with a confusing `ModuleNotFoundError: No module named '<name>.pipeline_registry'`, because `importlib` resolves the stdlib module (already cached in `sys.modules` from Kedro's own startup) instead of the project package.
- `settings.py` is silently ignored ("No settings.py found" even though it exists), because the same shadowing makes `import <name>.settings` fail and Kedro falls back to defaults.

These names are plausible (an email pipeline named `email`, a CSV tool named `csv`), and the failure is far removed from its cause, making it hard to diagnose.

## Development notes

The package name is derived exactly as the template does (`project_name.strip().replace(" ", "_").replace("-", "_").lower()`) and validated at a single chokepoint after `_get_extra_context`, so all three input paths (the `--name` flag, a config file, and interactive prompts) are covered without duplication.

The check rejects only:
- Python keywords (`keyword.iskeyword`), and
- standard library top-level modules (`sys.stdlib_module_names`, available on all supported Pythons since Kedro requires >=3.10).

It deliberately does not block builtins (a package named `type` does not shadow the builtin) or soft keywords (`match`, `case` are valid identifiers), and names that merely *contain* a stdlib substring (e.g. `email-service` -> `email_service`) are still allowed.

This mirrors how other scaffolders guard package names: Django's `startproject` rejects names that conflict with an existing Python module, and npm rejects names matching Node core modules.

## Test plan

Added to `tests/framework/cli/starters/test_new_from_cli_flags.py`:
- `test_project_name_clashing_with_stdlib_or_keyword_is_rejected` — `email`, `json`, `string`, `JSON`, `" email "`, `import` are rejected with a clear message.
- `test_project_name_containing_but_not_equal_to_stdlib_is_allowed` — `email-service`, `my-json-app`, `import_data` are still allowed (no over-blocking).

The full `tests/framework/cli/starters/` suite passes, and `ruff check` / `ruff format` are clean on the changed files.

## Checklist

- [x] Read the [contributing guidelines](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md)
- [x] Signed off each commit with a [Developer Certificate of Origin (DCO)](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md#developer-certificate-of-origin)
- [x] Updated `RELEASE.md` with a description of my changes


RudraDudhat2509:
Here are some screenshots for your reference 
Ran against the patched code (python -m kedro from the repo root so it uses the fix, not the installed release):

1. A name whose package shadows a stdlib module is now rejected at creation time:


python -m kedro new --name=email --tools=none --example=n
'email' is an invalid value for project name. It would create a Python package named 'email', which clashes with a Python keyword or standard library module and would prevent the project from being imported. Please choose a different name.

image


2. A name that only contains a stdlib substring still works (no over-blocking):


python -m kedro new --name=my-email-app --tools=none --example=n
Creates my_email_app successfully.

image

kedro/framework/cli/starters.py
```diff
@@ -376,6 +400,10 @@ def new(  # noqa: PLR0913
         starter_alias=starter_alias,
     )
 
+    _validate_project_name_is_importable(
+        extra_context.get("project_name", "New Kedro Project")
```
ravi-kumar-pilla, line 411: **Suggestion:** Validation here only derives the package name from `project_name`. A config file can set `python_package: email` explicitly while `project_name` is something innocuous (e.g. `"My Email Tool"`), and cookiecutter will still use `email` — bypassing this check.

Consider validating the **actual** package name cookiecutter will use:

```python
package_name = extra_context.get("python_package") or _derive_package_name(
    extra_context.get("project_name", "New Kedro Project")
)
```

A small test in `test_new_from_config_file.py` with `python_package: email` would lock this in. Not blocking — most users hit this via `--name` or interactive prompts — but worth covering if you want full parity across input paths.

ravi-kumar-pilla (approved):
## Kedro PR Review

### Overview

Nice fix — this directly addresses the stdlib shadowing bug (e.g. `kedro new --name email` → confusing `ModuleNotFoundError` at `kedro run`). The chokepoint after `_get_extra_context` is the right place, the derivation matches `cookiecutter.json`, tests cover both rejection and non-over-blocking, and `RELEASE.md` is updated.

No issues found.

- [ ] `kedro/framework/cli/starters.py:L404` — validate explicit `python_package` from config files (see inline comment)
- [ ] `docs/create/new_project.md` — optional follow-up: note that stdlib/keyword names are rejected

### Notes
- **Fixes #5607:** Yes, for `--name`, interactive prompts, and config where the package is derived from `project_name`.
- **PR checklist:** All good (`RELEASE.md`, tests, DCO).
- **API usage:** N/A.

RudraDudhat2509:
Noted, will implement the suggestions in the morning !

RudraDudhat2509:
thanks @ravi-kumar-pilla, good catch. pushed a fix: now validates the package cookiecutter actually uses (`python_package` if set, else derived), so an explicit `python_package: email` gets caught even with an innocent `project_name`. added a config-file test for it. suite green, ruff clean.

RELEASE.md
```diff
@@ -9,11 +9,13 @@
 ## Bug fixes and other changes
 * Fixed Rich logging integration so node input/output brackets render correctly in console logs and dataset colour markup does not leak into plain log handlers.
 * Improved the `AbstractDataset.from_config()` error message for custom dataset classes that are still abstract, so it no longer suggests invalid constructor arguments when required dataset methods are missing.
+* Fixed `kedro new` accepting project names whose derived package name shadows a Python standard library module or is a Python keyword (e.g. `email`, `json`, `import`), which silently produced a broken, unimportable project. Such names are now rejected at creation time with a clear message.
```
ElenaKhaustova, line 12: **Suggestion (non-blocking):** Consider mirroring this in `docs/create/new_project.md` — the "Project name" section (~L18–29) already documents the allowed-character rule and the `project_name → python_package` derivation, so a one-line addition saying that names whose derived package shadows a Python stdlib module or matches a keyword (e.g. `email`, `json`, `import`) are rejected would round it out. Otherwise, users only learn the rule from the error at creation time.

ElenaKhaustova (commented):
Thank you @RudraDudhat2509!

Left one minor suggestion, but overall, it looks good to me 👍

ElenaKhaustova (approved):

RELEASE.md
RudraDudhat2509: Noted ! on it

PR #5612, opened 2026-06-20
fix: reject Python keywords as pipeline names in kedro pipeline create

### LINKED ISSUE #5610: kedro pipeline create: Python keywords accepted as pipeline names but produce unimportable modules
## Description

`kedro pipeline create` accepts Python reserved keywords (`for`, `if`, `while`, `import`, `return`, `class`, etc.) as pipeline names without any error. The validator `_assert_pkg_name_ok` in `kedro/framework/cli/pipeline.py` only checks character set and minimum length, it never calls `keyword.iskeyword()`.

This creates a pipeline directory on disk that cannot be imported using standard Python syntax, producing a `SyntaxError` at runtime.

## Steps to reproduce

Calling Kedro's actual validator directly:

```python
import keyword
from kedro.framework.cli.pipeline import _assert_pkg_name_ok

for name in ['for', 'if', 'while', 'import', 'return', 'class']:
    try:
        _assert_pkg_name_ok(name)
        result = 'PASSES (bug: keyword not rejected)'
    except Exception as e:
        result = f'BLOCKED: {e}'
    print(f'{name!r:12} keyword={keyword.iskeyword(name)}  {result}')
```

Output:
```
'for'        keyword=True  PASSES (bug: keyword not rejected)
'if'         keyword=True  PASSES (bug: keyword not rejected)
'while'      keyword=True  PASSES (bug: keyword not rejected)
'import'     keyword=True  PASSES (bug: keyword not rejected)
'return'     keyword=True  PASSES (bug: keyword not rejected)
'class'      keyword=True  PASSES (bug: keyword not rejected)
```

Running `kedro pipeline create for` succeeds and creates `src/<project>/pipelines/for/`. Any code that then tries to import it with standard syntax fails:

```python
from my_project.pipelines import for  # SyntaxError: invalid syntax
```

## Expected behaviour

`kedro pipeline create for` should fail immediately with a clear error, the same way invalid characters are rejected.

## Actual behaviour

The command succeeds, silently creating a pipeline module that cannot be imported with standard Python syntax.

## Context

- The same gap was recently fixed for `kedro new` in PR #5608, which added a `keyword.iskeyword()` check for project package names
- The fix here would reuse the existing validator logic from `starters.py` inside `_assert_pkg_name_ok` in `pipeline.py`, keeping the validation consistent across both commands

## Environment

```
OS: Windows 11
Python: 3.12.7
Kedro: latest main
```

RudraDudhat2509:
## Description

Closes #5610

`kedro pipeline create` accepted Python reserved keywords (`for`, `if`, `while`, `import`, `return`, `class`, etc.) as valid pipeline names. The existing `_assert_pkg_name_ok` validator only checked character set and minimum length, it never called `keyword.iskeyword()`. This allowed creating a pipeline directory on disk that cannot be imported with standard Python syntax.

## Changes

- Added `import keyword` to `kedro/framework/cli/pipeline.py`
- Added a `keyword.iskeyword()` check at the end of `_assert_pkg_name_ok`, raising `KedroCliError` with a clear message — reusing the same validation pattern introduced in PR #5608 for `kedro new`
- Added `for`, `if`, `import`, `return` to the existing `test_bad_pipeline_name` parametrize list

## What was tested

```
tests/framework/cli/pipeline/test_pipeline.py::TestPipelineCreateCommand::test_bad_pipeline_name
8 passed
```

Before the fix:
```
kedro pipeline create for  ->  Pipeline 'for' was successfully created.
```

After the fix:
```
kedro pipeline create for  ->  Error: 'for' is not a valid Python package name. 'for' is a Python keyword and cannot be used as a package name.
```

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -10,6 +10,7 @@
 * Fixed Rich logging integration so node input/output brackets render correctly in console logs and dataset colour markup does not leak into plain log handlers.
 * Improved the `AbstractDataset.from_config()` error message for custom dataset classes that are still abstract, so it no longer suggests invalid constructor arguments when required dataset methods are missing.
 * Fixed `kedro new` accepting project names whose derived package name shadows a Python standard library module or is a Python keyword (e.g. `email`, `json`, `import`), which silently produced a broken, unimportable project. Such names are now rejected at creation time with a clear message.
+* Fixed `kedro pipeline create` accepting Python keywords (e.g. `for`, `import`, `return`) as pipeline names. Such names are now rejected at creation time with a clear error message.
 ## Documentation changes
 * Documented hooks limitation when using `ParallelRunner`.
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -11,6 +11,7 @@
 * Improved the `AbstractDataset.from_config()` error message for custom dataset classes that are still abstract, so it no longer suggests invalid constructor arguments when required dataset methods are missing.
 * Fixed `kedro new` accepting project names whose derived package name shadows a Python standard library module or is a Python keyword (e.g. `email`, `json`, `import`), which silently produced a broken, unimportable project. Such names are now rejected at creation time with a clear message.
 * Fixed `kedro pipeline create` accepting Python keywords (e.g. `for`, `import`, `return`) as pipeline names. Such names are now rejected at creation time with a clear error message.
+
 ## Documentation changes
 * Documented hooks limitation when using `ParallelRunner`.
 
--- a/kedro/framework/cli/pipeline.py
+++ b/kedro/framework/cli/pipeline.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import keyword
 import re
 import shutil
 from pathlib import Path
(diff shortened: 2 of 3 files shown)


RudraDudhat2509:
image

image

merelcht (approved):
Thanks so much for this contribution @RudraDudhat2509 ! ⭐ 

Can you also add a note in the release notes about this change?

RudraDudhat2509:
sure! pushed the changes, thanks for the quick review

RELEASE.md
```diff
@@ -10,6 +10,7 @@
 * Fixed Rich logging integration so node input/output brackets render correctly in console logs and dataset colour markup does not leak into plain log handlers.
 * Improved the `AbstractDataset.from_config()` error message for custom dataset classes that are still abstract, so it no longer suggests invalid constructor arguments when required dataset methods are missing.
 * Fixed `kedro new` accepting project names whose derived package name shadows a Python standard library module or is a Python keyword (e.g. `email`, `json`, `import`), which silently produced a broken, unimportable project. Such names are now rejected at creation time with a clear message.
+* Fixed `kedro pipeline create` accepting Python keywords (e.g. `for`, `import`, `return`) as pipeline names. Such names are now rejected at creation time with a clear error message.
```
ElenaKhaustova, line 13: Missing blank line before `Documentation changes heading`

ElenaKhaustova (approved):
Thank you, @RudraDudhat2509, the changes look good to me!

Left one nit comment, otherwise it's ready to be merged.

RELEASE.md
RudraDudhat2509: noted, pushing the change shortly

PR #5613, opened 2026-06-20
fix: reject dunder names as pipeline names and skip them in find_pipelines

### LINKED ISSUE #5611: find_pipelines silently skips pipelines named __init__ or __main__
## Description

`find_pipelines()` iterates the `pipelines/` directory and skips `__pycache__` and hidden directories, but does not skip `__init__` or `__main__`. If a user runs `kedro pipeline create __init__`, the pipeline is created successfully on disk but is silently never discovered at runtime.

Both `__init__` and `__main__` pass `_assert_pkg_name_ok` (they start with `_`, are 2+ chars, contain only word characters), so the creation command raises no error.

## Steps to reproduce

Demonstrating the gap in the skip logic using Kedro's actual source:

```python
import inspect

src = inspect.getsource(find_pipelines)
for i, line in enumerate(src.splitlines()):
    if 'pycache' in line or 'startswith' in line or 'pipeline_name ==' in line:
        print(f'L{i+1}: {line}')
```

Output shows only two skip conditions:
```
L93:     if pipeline_name == "__pycache__":
L96:     if pipeline_name.startswith("."):
```

`__init__` and `__main__` are not guarded. When `find_pipelines` encounters an `__init__` directory, it calls:

```python
importlib.import_module(f"{PACKAGE_NAME}.pipelines.__init__")
```

This resolves to the `pipelines` package's own `__init__.py`, not the `__init__/` subdirectory pipeline. It finds no `create_pipeline` function there and silently drops the pipeline, leaving the user with a created pipeline that never runs and no error message.

## Expected behaviour

`kedro pipeline create __init__` (and `__main__`) should be rejected at creation time with a clear error. Alternatively, `find_pipelines` should explicitly skip these directory names the same way it skips `__pycache__`.

## Actual behaviour

Creation succeeds. The pipeline exists on disk but is silently never discovered or executed by `find_pipelines`.

## Context

- `__pycache__` is already explicitly skipped (L93 in `find_pipelines`)
- The fix should extend the skip list to include `__init__` and `__main__`, and/or add a guard in `_assert_pkg_name_ok` rejecting these names at creation time
- Reusing the existing validator functions from `starters.py` (per PR #5608 pattern) would keep validation consistent

## Environment

```
OS: Windows 11
Python: 3.12.7
Kedro: latest main
```

RudraDudhat2509:
## Description

Closes #5611

`kedro pipeline create __init__` (and `__main__`) succeeded silently, creating a pipeline that is never discovered at runtime. Two things were missing:

1. `_assert_pkg_name_ok` did not reject dunder names (`__init__`, `__main__`, etc.)
2. `find_pipelines` only skipped `__pycache__` when it encountered an `__init__` directory, `importlib.import_module("project.pipelines.__init__")` resolved to the pipelines package itself, not the subdirectory, so the pipeline was silently dropped with no error

## Changes

- `kedro/framework/cli/pipeline.py`: added a dunder check in `_assert_pkg_name_ok`  any name starting and ending with `__` is rejected at creation time with a clear error
- `kedro/framework/project/__init__.py`: added a dunder skip in `find_pipelines` alongside the existing `__pycache__` skip acts as a runtime safety net for projects that may already have dunder-named pipeline directories
- `tests/framework/cli/pipeline/test_pipeline.py`: added `__init__` and `__main__` to the existing `test_bad_pipeline_name` parametrize list

## What was tested

```
tests/framework/cli/pipeline/test_pipeline.py::TestPipelineCreateCommand::test_bad_pipeline_name
6 passed
```

Before the fix:
```
kedro pipeline create __init__  ->  Pipeline '__init__' was successfully created.
# pipeline exists on disk but find_pipelines silently never discovers it
```

After the fix:
```
kedro pipeline create __init__  ->  Error: '__init__' is not a valid Python package name.
'__init__' is a reserved dunder name and cannot be used as a pipeline name.
```

### DIFF
--- a/kedro/framework/cli/pipeline.py
+++ b/kedro/framework/cli/pipeline.py
@@ -66,6 +66,12 @@ def _assert_pkg_name_ok(pkg_name: str) -> None:
             base_message + " It must contain only letters, digits, and/or underscores."
         )
         raise KedroCliError(message)
+    if pkg_name.startswith("__") and pkg_name.endswith("__"):
+        message = (
+            base_message
+            + f" '{pkg_name}' is a reserved dunder name and cannot be used as a pipeline name."
+        )
+        raise KedroCliError(message)
 
 
 def _check_pipeline_name(ctx: click.Context, param: Any, value: str) -> str:
--- a/kedro/framework/project/__init__.py
+++ b/kedro/framework/project/__init__.py
@@ -556,6 +556,10 @@ def find_pipelines(  # noqa: PLR0912, PLR0915
         # Prevent imports of hidden directories/files
         if pipeline_name.startswith("."):
             continue
+        # Skip dunder directories — importing e.g. package.pipelines.__init__
+        # resolves to the pipelines package itself, not the subdirectory.
+        if pipeline_name.startswith("__") and pipeline_name.endswith("__"):
+            continue
 
         if requested_pipelines is not None and pipeline_name not in requested_pipelines:
             continue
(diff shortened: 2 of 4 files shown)


RudraDudhat2509:
image
image

PR #5623, opened 2026-06-23
Improve pipeline mapping error message

### LINKED ISSUE #3953: Improve error message in `ModularPipelineError` raised within `_validate_datasets_exist()`
## Description
A nice feature of modular pipelines is that if you make a mistake in mapping inputs/outputs to what nodes need, it raises a `ModularPipelineError` with a helpful message containing a set of catalog items you missed:

```
Failed to map datasets and/or parameters onto the nodes provided: <what you didn't map>...
Did you mean <those> instead?
```

However I think this error message can be even more useful if we detail the mismatch more. For example, to distinguish this:
- Am I **not supplying** something that one of the nodes needs to `pipeline` wrapper?
- Or am I supplying something to the `pipeline` wrapper **that none of the nodes need**?

As far as I understand, those 2 scenarios would now lead to the same error message. However it seems possible to make this distinction based on inputs that go to `_validate_datasets_exist()` function that's responsible for raising this exception.

## Context
I think it might improve developer experience while building modular pipelines.

## Possible Implementation
Instead of inferring a single list of mismatches:

```python
existing = {_strip_transcoding(ds) for ds in pipe.datasets()}
non_existent = (inputs | outputs | parameters) - existing
```

Infer it as two separate sets (that sum to `non_existent`) called something like `redundant_inputs` and `missing_inputs`. And configure the error message reflect the difference.

samiat4911:
## Description

Fixes #3953.

This PR clarifies the `PipelineError` raised when a dataset or parameter passed to `pipeline()` mapping arguments does not exist in the provided pipeline.

Previously, the error message only listed the unmapped name, which could be confusing in parameter remapping cases. The updated message now states that the supplied dataset or parameter does not exist in the pipeline, while preserving the existing close-match suggestion.
## Development notes

Changed `_validate_datasets_exist()` to include `does/do not exist in the pipeline` in the error message.

Added a regression test covering a parameter remapping typo similar to the issue example. The test verifies that the error reports the invalid parameter mapping and does not incorrectly report intermediate pipeline datasets such as `X_train`, `X_test`, `regressor`, or `metrics`.

## Checklist

### DIFF
--- a/kedro/pipeline/pipeline.py
+++ b/kedro/pipeline/pipeline.py
@@ -89,8 +89,12 @@ def _validate_datasets_exist(
     if non_existent:
         sorted_non_existent = sorted(non_existent)
         possible_matches = get_close_matches(sorted_non_existent, existing)
+        verb = "does" if len(sorted_non_existent) == 1 else "do"
 
-        error_msg = f"Failed to map datasets and/or parameters onto the nodes provided: {', '.join(sorted_non_existent)}"
+        error_msg = (
+            "Failed to map datasets and/or parameters onto the nodes provided: "
+            f"{', '.join(sorted_non_existent)} {verb} not exist in the pipeline."
+        )
         suggestions = (
             f" - did you mean one of these instead: {', '.join(possible_matches)}"
             if possible_matches
--- a/tests/pipeline/test_pipeline_namespacing.py
+++ b/tests/pipeline/test_pipeline_namespacing.py
@@ -472,6 +472,55 @@ class TestPipelineHelper:
         with pytest.raises(PipelineError, match=pattern):
             pipeline(raw_pipeline, parameters={"parameters": "some_yaml_dataset"})
 
+    def test_non_existent_parameter_mapping_does_not_report_intermediate_datasets(self):
+        def split_data(model_input_table, model_options):
+            return None
+
+        def train_model(X_train, y_train):
+            return None
+
+        def evaluate_model(regressor, X_test, y_test):
+            return None
+
+        raw_pipeline = pipeline(
+            [
+                node(
+                    split_data,
+                    ["model_input_table", "params:model_options"],
+                    ["X_train", "X_test", "y_train", "y_test"],
+                    name="split_data_node",
+                ),
+                node(
+                    train_model,
+                    ["X_train", "y_train"],
+                    "regressor",
+                    name="train_model_node",
+                ),
+                node(
+                    evaluate_model,
+                    ["regressor", "X_test", "y_test"],
+                    "metrics",
+                    name="evaluate_model_node",
+                ),
+            ]
+        )
+
+        with pytest.raises(PipelineError) as exc:
+            pipeline(
+                raw_pipeline,
+                parameters={"params:model_options_1": "params:model_options"},
+            )
+
+        error_message = str(exc.value)
+        assert "params:model_options_1 does not exist in the pipeline" in error_message
+        assert (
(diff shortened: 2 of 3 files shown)


merelcht (commented):
Thanks @samiat4911, I've tested this and it works well. I can verify this also works when you pass an argument to the Pipeline wrapper that does exist in the catalog and is used in another pipeline. 

Could you add this change to the release notes as well along with your name in the contributor section? 🙂

samiat4911:
> 

sure I can thanks for the feedback! :)

merelcht (approved):
Thanks for the contribution @samiat4911 ! ⭐️

kedro/pipeline/pipeline.py
```diff
@@ -89,8 +89,12 @@ def _validate_datasets_exist(
     if non_existent:
         sorted_non_existent = sorted(non_existent)
         possible_matches = get_close_matches(sorted_non_existent, existing)
+        verb = "does" if len(sorted_non_existent) == 1 else "do"
 
-        error_msg = f"Failed to map datasets and/or parameters onto the nodes provided: {', '.join(sorted_non_existent)}"
+        error_msg = (
+            "Failed to map datasets and/or parameters onto the nodes provided: "
+            f"{', '.join(sorted_non_existent)} {verb} not exist in the pipeline."
```
ElenaKhaustova: **Suggestion:** When multiple items don't exist, the message reads e.g. `bar, foo do not exist in the pipeline.` — grammatically correct, but the list-then-verb structure can be a little ambiguous about what "do not exist" applies to. A restructured form like:

```python
error_msg = (
    "Failed to map datasets and/or parameters onto the nodes provided: "
    f"the following {verb} not exist in the pipeline: {', '.join(sorted_non_existent)}."
)
```

would read more clearly for multi-item cases. Minor readability point — the current wording is acceptable too.

tests/pipeline/test_pipeline_namespacing.py
```diff
@@ -472,6 +472,55 @@ def test_non_existent_parameters_mapped(self):
         with pytest.raises(PipelineError, match=pattern):
             pipeline(raw_pipeline, parameters={"parameters": "some_yaml_dataset"})
```
ElenaKhaustova, line 480: **Suggestion:** The existing `test_non_existent_parameters_mapped` method above (lines 467 and 471) still uses patterns that don't include the new `does not exist in the pipeline` suffix. While the tests still pass (since `pytest.raises(match=...)` uses `re.search`), updating them to assert the full new message would make the test suite more consistent and serve as better documentation of the expected format. For example:

```python
pattern = r"Failed to map datasets and/or parameters onto the nodes provided: params:beta does not exist in the pipeline"
```

ElenaKhaustova (commented):
## Kedro PR Review

### Overview

No issues found.

- [ ] `kedro/pipeline/pipeline.py:L96` — Consider restructuring the message for clarity when multiple items are listed
- [ ] `tests/pipeline/test_pipeline_namespacing.py:L467,L471` — Update existing test patterns to assert the new message suffix for consistency

### Notes
- General review: Clean, focused change. The singular/plural handling is a nice touch. The new test is well-structured.
- PR checklist: RELEASE.md entry present in the correct section; contributor listed. No public API change, no docs or Viz impact.
- API usage: No issues.

samiat4911:
hello @ElenaKhaustova made an update pls review when you have the time thanks

ElenaKhaustova (approved):
Thank you, @samiat4911!

PR #5630, opened 2026-06-25
Fix logging filter class configuration

### LINKED ISSUE #5619: Custom logging Filter subclasses can't be configured via `logging.yml`
## Description

There is no working way to configure a custom `logging.Filter` subclass through the project `logging.yml`:

- The `()` factory key — the only key stdlib `logging.config.dictConfig` uses to instantiate a custom filter — is rejected with `ValueError: The '()' key is not allowed in logging configuration as it poses a security risk.`
- The `class:` key is accepted by Kedro's validator (`_validate_logging_class` explicitly whitelists `logging.Filter` subclasses), but stdlib `DictConfigurator.configure_filter` never reads class — it only handles `()` or name, otherwise falling back to a base `logging.Filter(name='')`. So the user's filter class is silently never instantiated.

## Context

We use a custom `logging.Filter` to deduplicate a noisy `UserWarning`. It worked before the `()`/`class` validation was added; afterwards the filter silently stopped applying with no error.

## Steps to Reproduce

```
import logging
import sys
import types

from kedro.framework.project import LOGGING


class MyFilter(logging.Filter):
    """Custom filter — easy to spot if it was actually attached."""

    def filter(self, record: logging.LogRecord) -> bool:
        return "KEEP" in record.getMessage()


# dictConfig resolves `class:` by import path, so register the class on a module.
demo_filters = types.ModuleType("demo_filters")
demo_filters.MyFilter = MyFilter
sys.modules["demo_filters"] = demo_filters

logging_config = {
    "version": 1,
    "disable_existing_loggers": False,
    "filters": {
        "my_filter": {"class": "demo_filters.MyFilter"},
    },
    "handlers": {
        "console": {
            "class": "logging.StreamHandler",
            "filters": ["my_filter"],
        },
    },
    "root": {"handlers": ["console"], "level": "INFO"},
}

# Reset logging so repeated runs behave the same.
logging.shutdown()
for logger in (logging.root, logging.getLogger("kedro")):
    logger.handlers.clear()
    logger.filters.clear()

LOGGING.configure(logging_config)

attached = [type(f) for h in logging.getLogger().handlers for f in h.filters]
print("Expected:", MyFilter)
print("Attached:", attached)
```

## Expected Result

Either:

1. `class:` instantiates the named `logging.Filter` subclass (since the validator already accepts it), or
2. a sanctioned, non-`()` way to register a custom filter class exists and is documented, or
3. Kedro raises/warns that custom filter classes are unsupported, instead of silently producing a no-op.

## Actual Result

`[<class 'logging.Filter'>]` — a base no-op filter. The custom `MyFilter` is never constructed, no error or warning is raised.

## Your Environment

- Kedro 1.4.0
- Python 3.13
- macOS / Linux (CI)

samiat4911:
# Fix logging filter class configuration

## Description
Fixes #5619.

Custom `logging.Filter` subclasses configured with `class:` in `logging.yml` were validated by Kedro but were not actually instantiated by Python's `logging.config.dictConfig()`. This meant a configuration such as:

```yaml
filters:
  my_filter:
    class: my_package.logging.MyFilter
```

silently attached a base no-op `logging.Filter` instead of the requested custom filter.

This PR keeps the existing security guardrail that rejects user-supplied `()` factory keys, while allowing validated `logging.Filter` subclasses configured with `class:` to be instantiated and applied.

## Development notes
Updated `_ProjectLogging` so it:

- resolves and validates logging classes through a shared helper;
- preserves the existing rejection of user-provided `()` keys;
- prepares a separate `dictConfig()` configuration where top-level filter definitions using a validated `logging.Filter` subclass are converted to an internal callable factory;
- keeps `LOGGING.data` in the original validated `class:` form, without storing internal `()` keys;
- preserves logging configurations that do not define a `filters` section.

Added regression tests that verify:

- a custom `logging.Filter` subclass configured with `class:` is attached and filters records;
- filtered records do not appear in the handler output;
- config preparation does not add an empty `filters` section when the user has not configured filters;
- non-filter classes are rejected when used inside the top-level `filters:` section;
- the existing `()` factory-key and non-logging-class security checks still pass.

Updated `docs/develop/logging.md` to explain how to configure custom filters with `class:` and why Kedro does not support the standard `()` factory key in `logging.yml`.

Tested with:

```powershell
.venv\Scripts\python.exe -m pytest --no-cov tests\framework\project\test_logging.py
pre-commit run ruff --files kedro\framework\project\__init__.py tests\framework\project\test_logging.py
pre-commit run ruff-format --files kedro\framework\project\__init__.py tests\framework\project\test_logging.py
pre-commit run trailing-whitespace --files docs\develop\logging.md kedro\framework\project\__init__.py tests\framework\project\test_logging.py
pre-commit run end-of-file-fixer --files docs\develop\logging.md kedro\framework\project\__init__.py tests\framework\project\test_logging.py
git diff --check
.venv\Scripts\mypy.exe kedro\framework\project\__init__.py --strict --allow-any-generics --no-warn-unused-ignores
```

Vale was not available in my local environment, so I used the Kedro docs manual style checklist for the docs update.

## Checklist

### DIFF
--- a/kedro/framework/project/__init__.py
+++ b/kedro/framework/project/__init__.py
@@ -361,7 +361,11 @@ class _ProjectLogging(UserDict):
         without allowing user-provided factories.
         """
         prepared_config = logging_config.copy()
-        filters = logging_config.get("filters", {})
+
+        if "filters" not in logging_config:
+            return prepared_config
+
+        filters = logging_config["filters"]
 
         if not isinstance(filters, dict):
             return prepared_config
--- a/kedro/framework/project/__init__.py
+++ b/kedro/framework/project/__init__.py
@@ -282,13 +282,16 @@ class _ProjectLogging(UserDict):
         self.configure(yaml.safe_load(logging_config))
         logger.info(msg)
 
-    def _validate_logging_class(self, class_path: str) -> None:
-        """Validate that a class referenced in logging configuration is a legitimate
-        logging class (i.e. a subclass of logging.Handler, logging.Formatter, or
-        logging.Filter).
+    def _resolve_logging_class(self, class_path: str) -> type[Any] | None:
+        """Resolve and validate that a class referenced in logging configuration is
+        a legitimate logging class (i.e. a subclass of logging.Handler,
+        logging.Formatter, or logging.Filter).
 
         Args:
             class_path: Dotted import path to the class, e.g. ``logging.StreamHandler``.
+
+        Returns:
+            Resolved class, or ``None`` for bare names resolved internally by logging.
 
         Raises:
             ValueError: If the class cannot be imported or is not a logging base class.
@@ -297,7 +300,7 @@ class _ProjectLogging(UserDict):
         module_path, _, class_name = class_path.rpartition(".")
         if not module_path:
             # Bare name (e.g. "StreamHandler") resolved internally by logging machinery.
-            return
+            return None
 
         try:
             module = importlib.import_module(module_path)
@@ -320,6 +323,14 @@ class _ProjectLogging(UserDict):
                 f"Got {type(cls).__name__!r}."
             )
 
+        return cls
+
+    def _validate_logging_class(self, class_path: str) -> None:
+        """Validate that a class referenced in logging configuration is a legitimate
+        logging class (i.e. a subclass of logging.Handler, logging.Formatter, or
+        logging.Filter)."""
+        self._resolve_logging_class(class_path)
+
     def _validate_logging_config(self, config: Any) -> Any:
         """Recursively check the logging configuration and raise an error if dangerous
         '()' factory keys are encountered or if any 'class' value is not a legitimate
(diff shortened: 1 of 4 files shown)


tests/framework/project/test_logging.py
```diff
@@ -461,6 +466,64 @@ def test_validate_logging_class_bare_name_passes():
...
+    drop_record = logging.LogRecord(
+        "test_logger", logging.INFO, __file__, 1, "DROP this message", (), None
+    )
+    keep_record = logging.LogRecord(
+        "test_logger", logging.INFO, __file__, 1, "KEEP this message", (), None
+    )
+    handler.handle(drop_record)
+    handler.handle(keep_record)
+
+    assert stream.getvalue() == "KEEP this message\n"
```
merelcht: To make the test even clearer I'd also assert that "DROP this message" does not appear in the logs.

merelcht (changes requested):
Thanks for having a stab at fixing this @samiat4911 ! A couple of comments from my side:

- Now we always add the "filters" key to the resolved config even if the user isn't passing a custom filter (see also the breaking unit tests). Can you make sure that doesn't happen?
- Can you also update the logging docs explaining how users should add a custom filter? Using `class` differs from the standard way of using `()`, which we deliberately don't want to use.

Thank you! ⭐

tests/framework/project/test_logging.py
samiat4911: hello @merelcht could you run the test again? I just made an update thanks :)

merelcht: I have run the tests. Our contributor guide also explains how you do this locally: https://github.com/kedro-org/kedro/wiki/Guidelines-for-contributing-developers that way you can make sure the tests pass before requesting reviews.

merelcht: You'll need to run the full test suite and not just the test files that have changed.

samiat4911: yeah I'm running a full test rn

samiat4911: @merelcht ran test locally and it passed could you run the ci tests?

merelcht (approved):
Thanks for addressing the comments @samiat4911 ! This looks good to me now 👍

kedro/framework/project/__init__.py
```diff
@@ -359,13 +370,59 @@ def _validate_logging_config(self, config: Any) -> Any:
...
+            return prepared_config
+
+        prepared_filters = filters.copy()
+        prepared_config["filters"] = prepared_filters
+
+        for filter_name, filter_config in prepared_filters.items():
+            if not isinstance(filter_config, dict) or "class" not in filter_config:
+                continue
+
+            class_path = filter_config["class"]
```
ElenaKhaustova, line 415: **Suggestion:** Each filter class is resolved twice: once during `_validate_logging_config` (which calls `_validate_logging_class` → `_resolve_logging_class` for every `class` key in the config), and then again here in `_prepare_logging_config`. This involves redundant `importlib.import_module` + `getattr` + `issubclass` calls.

This is harmless — it's a startup-only code path — but if you wanted to avoid it, `_prepare_logging_config` could skip re-resolution by importing the class inline (since `_validate_logging_config` has already guaranteed the class is a valid logging class).

tests/framework/project/test_logging.py
```diff
@@ -461,6 +466,125 @@ def test_validate_logging_class_bare_name_passes():
...
+
+    logging_instance = _ProjectLogging()
+    prepared_config = logging_instance._prepare_logging_config(logging_config)
+
+    assert prepared_config["filters"]["passthrough"] == filter_config
+
+
+def test_prepare_logging_config_ignores_bare_filter_class_name():
+    from kedro.framework.project import _ProjectLogging
+
```
ElenaKhaustova, line 618: **Suggestion:** Consider adding a test for a filter configured with `class` plus additional constructor parameters (e.g., `name`). The dict comprehension in `_prepare_logging_config` preserves non-`class` keys so they're passed as `**kwargs` via `dictConfig`'s `configure_custom`, but there's no test exercising that path.

For example, a config like `{"class": "...", "name": "custom_name"}` should result in a filter constructed with `name="custom_name"`. This would prevent regressions if the dict comprehension logic changes.

ElenaKhaustova (commented):
## Kedro PR Review

### Overview

No issues found.

- [ ] `kedro/framework/project/__init__.py` — Double resolution of filter classes
- [ ] `tests/framework/project/test_logging.py` — Missing test for filter with extra constructor parameters

### Notes
- **General review:** The approach is well-designed. The security model is correct — user-provided `()` factory keys are still rejected by `_validate_logging_config`, and the internal `()` conversion in `_prepare_logging_config` only happens for validated `logging.Filter` subclasses. The separation between validated config (stored in `self.data`) and prepared config (passed to `dictConfig`) is clean. The additional check in `_prepare_logging_config` that rejects non-Filter logging classes in the `filters` section is a good improvement — it catches misconfigurations that would previously result in a silent no-op filter.
- **PR checklist:** RELEASE.md entry present. Docs updated. Tests cover core functionality, edge cases, and security guardrails. No backward compatibility concerns.
- **API usage:** No Kedro public API changes. All modifications are in private methods of `_ProjectLogging`.

samiat4911:
hello @ElenaKhaustova I made an update from your feedback pls could you give it a review? thank you

ElenaKhaustova (approved):
Thank you, @samiat4911!

PR #5633, opened 2026-06-27
Fix custom logging filter class configuration

### LINKED ISSUE #5619: Custom logging Filter subclasses can't be configured via `logging.yml`

jeevan6996:
## Description
Fixes #5619.

Kedro validates `class` entries in logging config, but Python's stdlib `DictConfigurator.configure_filter()` ignores `class` for filters and only instantiates custom filters through the `()` factory key. Since Kedro blocks user-provided `()` for security, custom `logging.Filter` subclasses configured in `logging.yml` were silently replaced with a base no-op `logging.Filter`.

This PR keeps the user-facing config restricted to the validated `class` key, then internally converts top-level filter class configs to a safe callable only after validation confirms the target is an importable logging class. `LOGGING.data` remains the original validated config so subprocess reconfiguration can repeat the same safe preparation step.

## Development notes
- Added a regression test using an importable custom `logging.Filter` subclass.
- Added a release-note entry under Upcoming Release.

Validation performed:
- `python -m pytest tests/framework/project/test_logging.py -q --no-cov` -> 40 passed
- `ruff format kedro/framework/project/__init__.py tests/framework/project/test_logging.py` -> unchanged
- `ruff check kedro/framework/project/__init__.py` -> passed
- `git diff --check` -> passed

Notes:
- Running the focused pytest command without `--no-cov` executes all 40 tests successfully, then exits non-zero because the repo-wide coverage gate requires 100% total coverage for the entire package even when only this file is selected.
- A full `ruff check tests/framework/project/test_logging.py` with the latest downloaded Ruff reports existing local-import and long-line findings elsewhere in the file that are unrelated to this change; this PR avoids adding another local import by importing `_ProjectLogging` at module scope.

## Checklist

- [x] Signed off each commit with DCO
- [x] Opened this PR as a Draft Pull Request if it is work-in-progress
- [x] Added a description of this change in RELEASE.md

Docs note: no docs update included because this restores the existing documented `class`-based logging configuration behavior without adding a new public option.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -8,6 +8,7 @@
 
 ## Bug fixes and other changes
 * Fixed Rich logging integration so node input/output brackets render correctly in console logs and dataset colour markup does not leak into plain log handlers.
+* Fixed project logging configuration so custom `logging.Filter` subclasses can be configured with the validated `class` key without using the unsafe `()` factory key.
 * Improved the `AbstractDataset.from_config()` error message for custom dataset classes that are still abstract, so it no longer suggests invalid constructor arguments when required dataset methods are missing.
 * Fixed `kedro new` accepting project names whose derived package name shadows a Python standard library module or is a Python keyword (e.g. `email`, `json`, `import`), which silently produced a broken, unimportable project. Such names are now rejected at creation time with a clear message.
 * Fixed `kedro pipeline create` accepting Python keywords (e.g. `for`, `import`, `return`) as pipeline names. Such names are now rejected at creation time with a clear error message.
--- a/kedro/framework/project/__init__.py
+++ b/kedro/framework/project/__init__.py
@@ -301,7 +301,7 @@ class _ProjectLogging(UserDict):
         self.configure(yaml.safe_load(logging_config))
         logger.info(msg)
 
-    def _validate_logging_class(self, class_path: str) -> None:
+    def _validate_logging_class(self, class_path: str) -> type | None:
         """Validate that a class referenced in logging configuration is a legitimate
         logging class (i.e. a subclass of logging.Handler, logging.Formatter, or
         logging.Filter).
@@ -316,7 +316,7 @@ class _ProjectLogging(UserDict):
         module_path, _, class_name = class_path.rpartition(".")
         if not module_path:
             # Bare name (e.g. "StreamHandler") resolved internally by logging machinery.
-            return
+            return None
 
         try:
             module = importlib.import_module(module_path)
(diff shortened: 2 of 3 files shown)


merelcht:
Hi @jeevan6996, thanks for having a go at addressing this. However, there's already a PR in review that addresses this issue https://github.com/kedro-org/kedro/pull/5630 opened a couple of days earlier. So I'll close yours.

jeevan6996:
Thanks for the note, and sorry for the duplicate PR. I missed #5630 during my duplicate check before opening this. I’ll leave this closed so the earlier PR can proceed without competing work.

PR #5663, opened 2026-07-13
Add the node's function name to NodeSnapshot

### LINKED ISSUE #5653: Add the node's function name to NodeSnapshot in get_project_snapshot
## Description

The project inspection snapshot (`get_project_snapshot`, `kedro>=1.4`) exposes each node's `name`, `namespace`, `tags`, `inputs` and `outputs`, but not the underlying function's name. We would like `NodeSnapshot` to also carry the node's function name.

## Context

Kedro-Viz is moving its graph backend to read from `get_project_snapshot` instead of loading a live `KedroSession` (see kedro-org/kedro-viz#2265). Two things need the function name, and neither can be reliably recovered from the fields available today:

1. **Stable node IDs (avoids a breaking change).** Kedro-Viz's existing node IDs are a hash of the node's string form, which includes the function name. Because the snapshot does not expose the function name, Kedro-Viz cannot reconstruct the existing IDs and has had to switch to a new hashing scheme. That is a breaking change for anything keyed on node IDs (saved node metadata, run-status mapping, deep links to a node). If the function name is in the snapshot, Kedro-Viz can reconstruct the existing ID scheme and avoid the break.

2. **Display name.** For auto-named nodes, `name` is `func__<hash>`; for explicitly named nodes, the function name is not present in `name` at all. The function name gives a clean, reliable label for the UI without having to parse a hash suffix out of the name.

## Possible Implementation

Add a field to `NodeSnapshot`, for example `func_name: str`, populated from the node's function (`node.func.__name__`, or the readable function name Kedro already derives internally). The value is available at snapshot-build time, since the pipelines are already imported when the snapshot is assembled, so this should not require any extra loading.

## Possible Alternatives

- Expose a stable, uniform node ID directly on `NodeSnapshot`. This also solves the ID need, but it would be a *new* ID rather than the one currently in use, so it would not avoid the breaking change on its own.
- Leave it out and have Kedro-Viz import the project itself to read the functions. This reintroduces exactly the coupling to project loading that the inspection snapshot is meant to remove.

jitu5:
## Description

Fixes #5653.

`get_project_snapshot` exposes each node's `name`, `namespace`, `tags`, `inputs` and `outputs`, but not the name of the underlying function. Kedro-Viz needs the function name when reading the graph from the snapshot instead of a live `KedroSession` (kedro-org/kedro-viz#2265):

- to reconstruct its existing node IDs, which are a hash of the node's string form (which includes the function name), avoiding a breaking change for anything keyed on those IDs

## Development notes

- Added `func_name: str` to `NodeSnapshot`, populated in `_node_to_snapshot` from `Node._func_name`, the readable function name Kedro already derives internally (the same one `kedro registry describe` prints). The value is available at snapshot-build time, so no extra project loading is needed.
- Updated the inspection docs (`NodeSnapshot` field list and example) and the `GET /snapshot` example response in the serving docs.
- Tests: extended the `NodeSnapshot` instantiation and `_node_to_snapshot` tests, and the snapshot endpoint test now asserts `func_name` appears in the JSON payload.

**Note on the constructor:** `func_name` is added as a required field, second in `NodeSnapshot`, so it sits next to `name` in serialized output. This changes the signature for any code that constructs `NodeSnapshot` positionally, likely rare, since snapshots are produced by Kedro and all internal call sites use keyword arguments. Happy to make it keyword-only (`field(kw_only=True)`) or give it a default if preferred.

## Checklist

### DIFF
--- a/kedro/inspection/models.py
+++ b/kedro/inspection/models.py
@@ -65,7 +65,7 @@ class NodeSnapshot:
     """
 
     name: str
-    func_name: str
+    func_name: str = field(kw_only=True)
     namespace: str | None = None
     tags: list[str] = field(default_factory=list)
     inputs: list[str] = field(default_factory=list)
--- a/kedro/inspection/models.py
+++ b/kedro/inspection/models.py
@@ -57,6 +57,7 @@ class NodeSnapshot:
 
     Attributes:
         name: Fully-qualified node name (includes namespace prefix if present).
+        func_name: Readable name of the node's underlying function.
         namespace: Node namespace, or ``None`` if the node has no namespace.
         tags: Sorted list of tags assigned to the node.
         inputs: Ordered list of input dataset names.
@@ -64,6 +65,7 @@ class NodeSnapshot:
     """
 
     name: str
+    func_name: str
     namespace: str | None = None
     tags: list[str] = field(default_factory=list)
     inputs: list[str] = field(default_factory=list)
--- a/kedro/inspection/snapshot.py
+++ b/kedro/inspection/snapshot.py
@@ -79,7 +79,7 @@ def _node_to_snapshot(node: Node) -> NodeSnapshot:
     """
     return NodeSnapshot(
         name=node.name,
-        func_name=node._func_name,
+        func_name=node._func_name,  # Matches Node.__str__ and registry describe.
         namespace=node.namespace,
         tags=sorted(node.tags),
         inputs=node.inputs,
--- a/kedro/inspection/snapshot.py
+++ b/kedro/inspection/snapshot.py
@@ -79,6 +79,7 @@ def _node_to_snapshot(node: Node) -> NodeSnapshot:
     """
     return NodeSnapshot(
         name=node.name,
+        func_name=node._func_name,
         namespace=node.namespace,
         tags=sorted(node.tags),
         inputs=node.inputs,
--- a/tests/inspection/test_node_pipeline_snapshot.py
+++ b/tests/inspection/test_node_pipeline_snapshot.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+from functools import partial
+
 import pytest
 
 from kedro.inspection.models import NodeSnapshot, PipelineSnapshot
@@ -50,6 +52,11 @@ class TestNodeSnapshot:
         assert snapshot.inputs == []
         assert snapshot.outputs == []
 
+    def test_func_name_is_keyword_only(self):
+        snapshot = NodeSnapshot("my_node", "my_namespace", func_name="my_func")
+        assert snapshot.func_name == "my_func"
+        assert snapshot.namespace == "my_namespace"
+
 
 class TestNodeToSnapshot:
     def test_populates_all_fields(self, simple_node):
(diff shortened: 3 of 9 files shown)


kedro/inspection/models.py
```diff
@@ -57,13 +57,15 @@ class NodeSnapshot:
...
         name: Fully-qualified node name (includes namespace prefix if present).
+        func_name: Readable name of the node's underlying function.
         namespace: Node namespace, or ``None`` if the node has no namespace.
         tags: Sorted list of tags assigned to the node.
         inputs: Ordered list of input dataset names.
         outputs: Ordered list of output dataset names.
     """
 
     name: str
+    func_name: str
```
ElenaKhaustova: `func_name` inserted as a required positional field silently corrupts field values for any pre-existing positional `NodeSnapshot(...)` call (values shift by one field with no error). Should be `field(kw_only=True)` or given a default as you proposed.

kedro/inspection/snapshot.py
```diff
@@ -79,6 +79,7 @@ def _node_to_snapshot(node: Node) -> NodeSnapshot:
     """
     return NodeSnapshot(
         name=node.name,
+        func_name=node._func_name,
```
ElenaKhaustova: `node._func_name` is now called unconditionally for every node. For nodes built from a "naked" `functools.partial (no __name__)` that have an explicit `name=`, this is a new code path: previously `Node.name` short-circuits on `self._name` and never touches` _func_name`, so no warning fired. Now it will emit a `UserWarning ("...made from a 'partial' function...")` the first time such a node is snapshotted, surfacing via routine inspection rather than an explicit call like kedro registry describe.

Could you add a test covering this case (partial-function node with an explicit name) and confirm the warning is expected/acceptable here?

kedro/inspection/models.py
jitu5: I’ll make `func_name` required but keyword-only so existing positional fields cannot silently shift.

kedro/inspection/snapshot.py
jitu5: Yes, I’ll add a test for an explicitly named partial node that verifies both `func_name == "<partial>"` and the warning.

tests/inspection/test_node_pipeline_snapshot.py
```diff
@@ -42,18 +44,25 @@ def simple_pipeline(simple_node):
...
-        snapshot = NodeSnapshot(name="my_node")
+        snapshot = NodeSnapshot(name="my_node", func_name="my_func")
         assert snapshot.name == "my_node"
+        assert snapshot.func_name == "my_func"
         assert snapshot.namespace is None
         assert snapshot.tags == []
         assert snapshot.inputs == []
         assert snapshot.outputs == []
 
+    def test_func_name_is_keyword_only(self):
```
ravi-kumar-pilla, line 55: This test covers the happy path, but not the failure mode Elena was concerned about. Can we add a negative case to lock in `kw_only=True` behavior:

```python
with pytest.raises(TypeError, match="func_name"):
    NodeSnapshot("my_node", "my_namespace")
```

Without `func_name`, construction should fail loudly rather than silently mis-assign fields. That would make the regression explicit.

kedro/inspection/snapshot.py
ravi-kumar-pilla: Using `node._func_name` is fine here, it matches what `kedro registry describe` and `str(node)` use, and handles partials consistently. A brief inline comment noting that intentional reuse would help future readers who might wonder why a private property is accessed.

ElenaKhaustova (approved):
LGTM 👍

ravi-kumar-pilla (approved):
LGTM 💯

PR #5664, opened 2026-07-13
fix: deduplicate get_close_matches suggestions and support iterator targets

eeshsaxena:
## Description

`get_close_matches` (in `kedro/utils.py`) has two issues when it receives a **list** of inputs:

1. **Duplicate suggestions.** Matches are collected per input and concatenated, so a target matched by more than one input is repeated. This surfaces in the user-facing "did you mean one of these instead: ..." text of `PipelineError` (via `_validate_datasets_exist` in `kedro/pipeline/pipeline.py`). For example:

   ```python
   get_close_matches(["pipelin", "pipeine"], ["pipeline", "catalog"])
   # ['pipeline', 'pipeline']   <- 'pipeline' repeated
   ```

2. **`targets` iterable can be exhausted.** `targets` is typed as `Iterable[str]`, but `list(targets)` was called once *per input* inside the loop. A one-shot iterable (e.g. a generator) is consumed on the first input, so later inputs match against an empty list and silently return nothing.

## Development notes

- Materialize `targets = list(targets)` once, before iterating the inputs.
- Deduplicate matches while preserving order (first-seen), so shared matches appear only once.
- Unified the `str` and `list` input handling; single-string behaviour is unchanged (a single `difflib.get_close_matches` call already returns unique, score-ordered results).

Testing:
- Added two regression tests to `TestGetCloseMatches` in `tests/test_utils.py`: one asserting shared matches are deduplicated, one passing a generator as `targets`. Both fail on `main` and pass with this change; the existing string/list tests still pass (`pytest tests/test_utils.py -k GetCloseMatches` -> 4 passed).
- Added a `RELEASE.md` entry under "Bug fixes and other changes".

An AI assistant helped spot and implement this; I reviewed the change and ran the tests locally.

## Checklist

- [x] Signed off each commit with a Developer Certificate of Origin (DCO)
- [x] Added a description of this change in the `RELEASE.md` file

### DIFF
--- a/kedro/utils.py
+++ b/kedro/utils.py
@@ -279,19 +279,21 @@ def get_close_matches(
     Returns:
         List of close matches or empty list if no matches are found.
     """
-    _init_matches = []
-    if isinstance(input, str):
-        matches = difflib.get_close_matches(
-            input, list(targets), n=max_suggestions, cutoff=cutoff
-        )
-        _init_matches.extend(matches)
-    else:
-        for source_str in input:
-            _matches = difflib.get_close_matches(
-                source_str, list(targets), n=max_suggestions, cutoff=cutoff
-            )
-            _init_matches.extend(_matches)
-    return _init_matches[:max_suggestions]
+    # Materialize once: `targets` may be a one-shot iterable, and it is reused
+    # for every input string below.
+    targets = list(targets)
+    inputs = [input] if isinstance(input, str) else input
+
+    matches: list[str] = []
+    for source_str in inputs:
+        for match in difflib.get_close_matches(
+            source_str, targets, n=max_suggestions, cutoff=cutoff
+        ):
+            # Deduplicate while preserving order: different inputs can match the
+            # same target, which would otherwise repeat it in the suggestions.
+            if match not in matches:
+                matches.append(match)
+    return matches[:max_suggestions]
 
 
 def find_config_file(
--- a/kedro/utils.py
+++ b/kedro/utils.py
@@ -285,13 +285,16 @@ def get_close_matches(
     inputs = [input] if isinstance(input, str) else input
 
     matches: list[str] = []
+    seen: set[str] = set()
     for source_str in inputs:
         for match in difflib.get_close_matches(
             source_str, targets, n=max_suggestions, cutoff=cutoff
         ):
             # Deduplicate while preserving order: different inputs can match the
             # same target, which would otherwise repeat it in the suggestions.
-            if match not in matches:
+            # A `seen` set keeps the membership check O(1).
+            if match not in seen:
+                seen.add(match)
                 matches.append(match)
     return matches[:max_suggestions]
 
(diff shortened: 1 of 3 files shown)


tests/test_utils.py
```diff
@@ -35,6 +35,19 @@ def test_list_input_returns_list(self):
...
+    def test_list_input_deduplicates_shared_matches(self):
+        # Two misspellings that both resolve to the same target should suggest
+        # that target only once, not repeat it.
+        result = get_close_matches(["pipelin", "pipeine"], ["pipeline", "catalog"])
+        assert result == ["pipeline"]
+
+    def test_list_input_accepts_one_shot_iterator_targets(self):
+        # `targets` is typed as an Iterable; a generator must not be exhausted
+        # after matching the first input string.
+        targets = (name for name in ["data_science", "data_engineering"])
```
ankatiyar:
```suggestion
        targets = iter(["data_science", "data_engineering"])
```

ankatiyar (approved):
Thanks @eeshsaxena LGTM with a minor suggestion to make the test intent clearer

eeshsaxena:
Good call, updated to `iter([...])` so the one-shot intent reads directly. Thanks for the review!

kedro/utils.py
```diff
@@ -279,19 +279,21 @@ def get_close_matches(
...
+
+    matches: list[str] = []
+    for source_str in inputs:
+        for match in difflib.get_close_matches(
+            source_str, targets, n=max_suggestions, cutoff=cutoff
+        ):
+            # Deduplicate while preserving order: different inputs can match the
+            # same target, which would otherwise repeat it in the suggestions.
+            if match not in matches:
+                matches.append(match)
```
ElenaKhaustova, line 298: **Suggestion:** The `if match not in matches` check is a linear scan. For the typical use case (`max_suggestions=3`) this is perfectly fine, but a `seen` set alongside the list would make the O(1) intent explicit and future-proof the function:

```python
matches: list[str] = []
seen: set[str] = set()
for source_str in inputs:
    for match in difflib.get_close_matches(
        source_str, targets, n=max_suggestions, cutoff=cutoff
    ):
        if match not in seen:
            seen.add(match)
            matches.append(match)
return matches[:max_suggestions]
```

Completely optional — a true nit given current usage.

ElenaKhaustova (commented):
## Kedro PR Review

One minor suggestion below — otherwise LGTM.

ElenaKhaustova (commented):
Thank you, @eeshsaxena! I made a pass, and it looks good to me. Left one minor suggestion.

Could you please also fix DCO and add your name under the `Community contributions` section of the `RELEASE.md`?

eeshsaxena:
Thanks @ElenaKhaustova! Done:
- Applied the `seen`-set dedup so the membership check is O(1).
- Signed off the commits (DCO green now).
- Added myself under Community contributions in `RELEASE.md`.

ElenaKhaustova (approved):
Thank you, @eeshsaxena!

PR #5685, opened 2026-07-18
fix: keep IPv6 hosts intact when parsing cloud filepaths

eeshsaxena:
`_parse_filepath` stripped the port with `rsplit(':', 1)`, which also cut an IPv6 host like `[2001:db8::1]` at its own colons (e.g. `s3://[2001:db8::1]/bucket` parsed to path `[2001:db8:/bucket`). Strip only a trailing `:port` instead. Adds direct tests for `_parse_filepath` (it was only mocked before).

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -3,8 +3,11 @@
 * Added the node function name to project inspection snapshots as `NodeSnapshot.func_name`.
 
 ## Bug fixes and other changes
+* Fixed `_parse_filepath` truncating IPv6 host addresses (e.g. `s3://[2001:db8::1]:9000/bucket`) when stripping the port.
 ## Documentation changes
 ## Community contributions
+Many thanks to the following Kedroids for contributing PRs to this release:
+* [eeshsaxena](https://github.com/eeshsaxena)
 
 # Release 1.5.0
 ## Major features and improvements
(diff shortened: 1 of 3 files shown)


eeshsaxena:
Cleaned this branch up: rebased onto `main` so it now contains **only** the IPv6 `_parse_filepath` fix (an unrelated `get_close_matches` change had been bundled in - that lives in #5664). Also signed off the commit so DCO passes and added a RELEASE.md entry.

eeshsaxena:
Closing this for now - it's picked up conflicts with `main` and I'd rather resend it cleanly than leave a conflicting PR sitting open. The IPv6 issue itself still stands, so I'll reopen or resend once it's rebased.

ElenaKhaustova:
> Closing this for now - it's picked up conflicts with `main` and I'd rather resend it cleanly than leave a conflicting PR sitting open. The IPv6 issue itself still stands, so I'll reopen or resend once it's rebased.

@eeshsaxena, thank you for your interest in contributing to Kedro!

Would you mind opening an issue with a clear description of the problem before opening the PR with the solution?

This is what we require according to our contribution guidelines: https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md and the developer guide: https://github.com/kedro-org/kedro/wiki/Guidelines-for-contributing-developers.

PR #5687, opened 2026-07-21
Add serving mode distinction

ankatiyar:
## Description

`KedroServiceSession` is designed to support concurrent `run()` calls (e.g. from an HTTP server). However, the shared `pipelines` singleton was being mutated on every `run()` via `set_requested(pipeline_names)`, which is a data race under concurrent load — one thread's call can invalidate the cache seen by another.

This PR introduces **serving mode**: pipelines are preloaded eagerly at session creation, before any request threads start. Once loaded, `run()` skips the `set_requested()` write entirely.

## Development notes

- Added `serving_mode: bool = False` to `KedroServiceSession.create()`.
- `_preload_pipelines()` calls `set_requested(None)` then iterates the singleton to trigger `_load_data()` for all pipelines.
- `run()` guards its `set_requested()` call (and the error-path reset) behind `if not self._serving_mode`.
- The flag is set **after** a successful preload so a failure during loading leaves the session in normal CLI mode rather than an inconsistent state.
- Default behaviour is unchanged — `serving_mode=False` means existing CLI usage is unaffected.

## Checklist


merelcht (approved):
Thanks @ankatiyar, this change makes sense to me! Should we also specify this behaviour in the docs?

docs/extend/session.md
```diff
@@ -90,6 +90,7 @@ You can provide the following optional arguments in `KedroServiceSession.create(
 - `project_path`: Path to the project root directory
 - `env`: Environment for the `KedroContext`
 - `conf_source`: Optional argument to specify the configuration source for the `KedroContext`
+- `serving_mode`: A Boolean value, `False` by default. Set to `True` when the session will handle concurrent `run()` calls from multiple threads, such as behind a web server. When enabled, all registered pipelines are loaded eagerly during `create()`, and `run()` no longer mutates the shared pipeline registry to select which pipelines to load. This avoids race conditions that could otherwise occur when multiple threads call `run()` at the same time. Leave this `False` for single-threaded or CLI use, where loading only the requested pipelines keeps startup faster.
```
ElenaKhaustova: Should we trim this to a few sentences like other bullet points? Something like:

`Set to True when the session will handle concurrent run() calls; preloads all pipelines eagerly during create() to avoid race conditions.`

tests/framework/session/test_service_session.py
```diff
@@ -668,3 +668,57 @@ def test_multiple_runs(self, fake_project, mock_runner, mocker):
         assert first_run_id != second_run_id
         assert first_runtime_params == {"param1": "value1"}
         assert second_runtime_params == {"param2": "value2"}
+
```
ElenaKhaustova, line 672: Should we also test the concurrency race condition here as well? For example, add a threading stress test (`ThreadPoolExecuto`r firing N run() calls in parallel against a `serving_mode=True` session, asserting no exceptions)

ElenaKhaustova (approved):
Thank you, @ankatiyar! Changes look good to me 👍

Left a few minor suggestions.

docs/extend/session.md
ankatiyar: Done

tests/framework/session/test_service_session.py
ankatiyar: Done

Issue #5680, opened 2026-07-17
Programmatic API (validate_dataset / validate_catalog)


SajidAlamQB:
## Description

Public functions to validate catalog datasets on demand: `validate_dataset(catalog, name, ...)` and `validate_catalog(...)`, returning a structured `ValidationResult` rather than raising. Built for machine consumers (the VSCode extension, CI scripts).

**No CLI in this issue** — that is a separate decision issue.

## Context

Part of **[Implementation] Dataset Validation (KEP-10)** — parent issue #5677. Depends on the funnel sub-issue.

- **Design:** #5602
- **Tech design summary:** https://github.com/kedro-org/kedro/issues/5390#issuecomment-5001293477
- **Prototype:** PR #5666 / branch [`feat/kep10-dataset-validation`](https://github.com/kedro-org/kedro/tree/feat/kep10-dataset-validation)
- **Runnable demo:** [spaceflights-validation-demo](https://github.com/SajidAlamQB/spaceflights-validation-demo)

Implements the locked rename to `validate_dataset`. Review from the VSCode extension side is worth arranging on the PR — `ValidationResult` is the extension's integration contract.

## Possible Implementation

- [ ] Rename `validate_catalog_dataset` → `validate_dataset` across code and docs
- [ ] `ValidationResult` statuses (`passed` / `failed` / `skipped` / `errored`) with the `error_type` discriminator; JSON-safe `to_dict()`
- [ ] Explicit calls always validate (ignore global and per-dataset disable flags)
- [ ] Factory-pattern enumeration for `validate_catalog` (open TODO in prototype)
- [ ] API status-taxonomy tests

Issue #5679, opened 2026-07-17
DataCatalog validation funnel + settings wiring


SajidAlamQB:
## Description

Wire dataset validation into `DataCatalog.load()`/`save()` (roughly 25 lines on the hot path plus lifecycle handling) and add the framework opt-out.

**This is the PR that touches `kedro/io`** — flagged at tech design as needing the most review scrutiny.

## Context

Part of **[Implementation] Dataset Validation (KEP-10)** — parent issue #5677. Depends on the core package sub-issue.

- **Design:** #5602
- **Tech design summary:** https://github.com/kedro-org/kedro/issues/5390#issuecomment-5001293477
- **Prototype:** PR #5666 / branch [`feat/kep10-dataset-validation`](https://github.com/kedro-org/kedro/tree/feat/kep10-dataset-validation)
- **Runnable demo:** [spaceflights-validation-demo](https://github.com/SajidAlamQB/spaceflights-validation-demo)

**A catalog-side reviewer is explicitly requested for this one.**

## Possible Implementation

Productionise from the prototype branch:

- [ ] `VALIDATOR_KEY` reservation + defensive strip in `parse_dataset_definition`
- [ ] Spec capture in `_add_from_config` (cleaned copy, resolver store not mutated)
- [ ] `_maybe_validate` guard ladder in `load`/`save`; error enrichment with dataset name and mode; `severity: warn`; `skip_load_after_save`
- [ ] Lifecycle: `__setitem__` replacement clears specs while materialisation keeps them; `release()`; `to_config()` re-injection; pickling drops resolved cache
- [ ] `DATASET_VALIDATION` setting + `KEDRO_DATASET_VALIDATION` env var (env wins, read inside `kedro.io`); context propagation + custom-catalog warning
- [ ] Wire `preflight_check` at catalog build (exists in prototype, currently unwired)
- [ ] No-op-path micro-benchmark in `kedro_benchmarks` (defends the hot-path cost)
- [ ] Funnel, lifecycle and kill-switch tests

Issue #5677, opened 2026-07-17
[Parent] Dataset Validation Implementation


SajidAlamQB:
## Description

Implement catalog-native dataset validation per the accepted **KEP-10** (#5602): a `validator:` key on catalog entries, enforced in the `DataCatalog` load/save path, with a pluggable `Validator` protocol and Pandera as the reference backend.

## Context

Follows the exploration initiative #5390 and spike #5391. KEP-10 accepted with **+3 TSC votes**.

- **Tech design summary:** https://github.com/kedro-org/kedro/issues/5390#issuecomment-5001293477
- **Tech design recording:** [Zoom](https://zoom.us/rec/play/mTxLoJQNvMHAj2BfBhKYlwI-i_xxY2vyo9GMsopYMH2id6AZqH2sEO3U7zUeHCLgj0pZDLlHVFnNkVwz.4H1PpBBHxCjs_IrH?accessLevel=meeting&canPlayFromShare=true&from=share_recording_detail&continueMode=true&oldStyle=true&componentName=rec-play&originRequestUrl=https%3A%2F%2Fzoom.us%2Frec%2Fshare%2F7IQpcMkcoVSHNLM8rdkDXc6mX-iGPCSJlIRZgrRcBKSSdZr4QahVNi8P7HX9BWt-.YIxNpBXPdmOxtYTv)
- **Miro board:** https://miro.com/app/board/uXjVH7zKmY4=/
- **Working prototype:** PR #5666 / branch [`feat/kep10-dataset-validation`](https://github.com/kedro-org/kedro/tree/feat/kep10-dataset-validation), plus a public runnable demo: [spaceflights-validation-demo](https://github.com/SajidAlamQB/spaceflights-validation-demo)

## Decisions locked at tech design

- **Extras:** per-backend — `kedro[pandera-pandas]`, `kedro[pandera-polars]`, `kedro[pandera-pyspark]`
- **Public API name:** `validate_dataset`
- **Entry-point group:** deferred until a second backend exists (name reserved in docs)
- **CLI:** split out into its own decision issue (naming / semantics / existence still open)

## Tasks

- [x] **PR 1 — core:** `kedro.validation` package, Pandera adapter, catalog funnel, settings + decision sweeps (extras, `validate_dataset` rename, install-hint strings)
- [x] **PR 2 — programmatic API:** `validate_dataset` / `validate_catalog` (no CLI)
- [ ] **New issue:** CLI naming / semantics / whether it ships
[...]

PR #5692, opened 2026-07-24
Add core validation package with pluggable Validator protocol and Pandera adapter

### LINKED ISSUE #5678: `kedro.validation` core package + Pandera adapter
# Description

Add the `kedro.validation` core package for catalog-native dataset validation: the `Validator` protocol, validator spec parsing, resolution, structured errors, and the Pandera reference adapter.

**Purely additive — no existing kedro modules touched.** This is deliberately split from the catalog funnel work so the `kedro/io` diff can be reviewed separately (the smaller-PRs ask from tech design).

# Context

Part of **[Implementation] Dataset Validation (KEP-10)** — parent issue #5677.

- **Design:** #5602
- **Tech design summary:** https://github.com/kedro-org/kedro/issues/5390#issuecomment-5001293477
- **Prototype:** PR #5666 / branch [`feat/kep10-dataset-validation`](https://github.com/kedro-org/kedro/tree/feat/kep10-dataset-validation) — working implementation, reviewed and green
- **Runnable demo:** [spaceflights-validation-demo](https://github.com/SajidAlamQB/spaceflights-validation-demo)

Implements the locked decisions on per-backend extras and install-hint strings. **The prototype is the base to work from**

# Possible Implementation

Productionise from the prototype branch:

- [ ] `Validator` protocol, `ValidatorSpec` (shorthand + long form, strict key validation, YAML 1.1 `on:` key normalisation), `CheckFailure`, `DataValidationError` (bounded grouped rendering), `ValidationConfigurationError`
- [ ] `resolve_validator` (adapters → class instantiation → instance → callable; never protocol-check a class object) + `CallableValidator` + `preflight_check`
- [ ] Pandera adapter: modern namespace detection (`pandera.pandas`/`polars`/`pyspark`), grouped `failure_cases` → `CheckFailure` with capped examples, coercion, options forwarding; pyspark accessor branch kept but not advertised
- [ ] Decision sweep: per-backend extras in `pyproject.toml`, install-hint strings updated
- [ ] Unit tests: spec parsing, resolution ordering, adapter behaviour
- [ ] Tech design action item: verify `SchemaErrors.failure_cases` shape stability from pandera 0.24 to current (folds into the Pandera maintainer review)

SajidAlamQB:
## Description

First implementation dataset validation (#5678, parent #5677, design discussion #5602).

This PR adds the core building blocks for dataset validation as a new set of modules under kedro.validation. It is purely additive and nothing here is reachable from user code paths yet. The catalog integration (#5679) and the public API (#5680) come in follow up PRs.

What it adds:

- Validator protocol: validate(data) -> data, any raise counts as a validation failure
- ValidatorSpec: parses a validator declaration (shorthand string or long form dict) with strict key validation
- resolve_validator: turns a spec into a live validator, handling schema classes, custom classes, instances and plain callables
- DataValidationError with grouped, bounded error reporting (CheckFailure), and ValidationConfigurationError for bad declarations
- A Pandera adapter supporting pandera.pandas, pandera.polars and pandera.pyspark models and schema instances (pandas and Polars are the supported backends, pyspark handling exists but is not advertised)
- New optional extras: kedro[pandera-pandas], kedro[pandera-polars], kedro[pandera-pyspark]

## Development notes

- kedro/validation/core.py, kedro/validation/pandera_validator.py
- New tests: tests/validation/test_validator_spec.py, test_resolution.py, test_pandera_validator.py
- Modified: kedro/validation/__init__.py (exports only), pyproject.toml (new extras), RELEASE.md

> [!IMPORTANT]

## Checklist


kedro/validation/core.py
```diff
@@ -0,0 +1,466 @@
...
+        if not isinstance(class_path, str) or not class_path:
+            raise ValidationConfigurationError(
+                f"Invalid validator declaration for dataset '{ds_name}': "
+                f"'class' must be a non-empty class path string."
+            )
+
+        on = config.get("on", _VALID_MODES)
+        if isinstance(on, str):
+            on = (on,)
+        elif isinstance(on, list | tuple):
```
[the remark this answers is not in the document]
SajidAlamQB: Union types in isinstance are supported from Python 3.10 (PEP 604), which is kedro's floor, the tests exercising this pass on CI's 3.12.

kedro/validation/pandera_validator.py
```diff
@@ -0,0 +1,325 @@
...
+            group["examples"].append(record.get("failure_case"))
+        if group["index"] is None:
+            group["index"] = record.get("index")
+
+    failures = []
+    for (column, check), group in grouped.items():
+        target = f"column '{column}'" if column else "dataframe"
+        failures.append(
+            CheckFailure(
+                message=f"Check '{check}' failed for {target}",
```
ravi-kumar-pilla: **Suggestion:** When pandera `failure_cases` rows omit `check`, grouping produces messages like `Check 'None' failed for column 'x'`, which reads like a Kedro bug. Consider a fallback label when `check is None` (e.g. `"validation check"` or omit the quoted check name).

kedro/validation/core.py
```diff
@@ -0,0 +1,500 @@
...
+            )
+
+        if not isinstance(config, dict):
+            raise ValidationConfigurationError(
+                f"Invalid validator declaration for dataset '{ds_name}': "
+                f"expected a class path string or a dictionary, got "
+                f"{type(config).__name__} ({config!r})."
+            )
+
+        if True in config:
```
ravi-kumar-pilla: **Suggestion:** YAML 1.1 `on:` normalisation is good, but if a config ever contained both `True` (parsed `on`) and an explicit `'on'` key, the dict comprehension silently keeps whichever comes last. Consider detecting `{True, "on"} ⊆ config.keys()` and raising `ValidationConfigurationError` for the collision.

kedro/validation/core.py
```diff
@@ -0,0 +1,500 @@
...
+            # YAML 1.1 parses an unquoted `on` key as boolean True
+            # (the "Norway problem"). Normalise it back so users can write
+            # `on: [load, save]` without quoting the key.
+            config = {("on" if k is True else k): v for k, v in config.items()}
+
+        unknown_keys = sorted(set(config) - set(_ALLOWED_SPEC_KEYS))
```
ravi-kumar-pilla: **Suggestion:** `unknown_keys = sorted(set(config) - set(_ALLOWED_SPEC_KEYS))` can raise `TypeError` if YAML supplies non-string keys (e.g. numeric keys), because mixed types are not sortable. Use `sorted(..., key=repr)` or validate keys are strings first.

kedro/validation/core.py
```diff
@@ -0,0 +1,500 @@
...
+    ) -> None:
+        super().__init__(message)
+        self.message = message
+        self.dataset_name = dataset_name
+        self.mode = mode
+        self.validator = validator
+        self.failures: list[CheckFailure] = list(failures) if failures else []
+
+    @staticmethod
+    def _render_failure(failure: CheckFailure) -> str:
```
ravi-kumar-pilla: **Suggestion:** `DataValidationError.__str__` is documented as bounded, but `_render_failure()` can emit unbounded lines when the column-only branch uses raw `failure.message`, or when pyspark accessor errors embed large `details` dicts in `CheckFailure.message`. Apply the same truncation used for headers/examples to rendered labels, or truncate `CheckFailure.message` at creation time in the pyspark path.

kedro/validation/core.py
```diff
@@ -0,0 +1,500 @@
...
+        on: Modes the validator applies to; subset of `("load", "save")`.
+        severity: `"error"` (raise on failure) or `"warn"` (log only).
+        enabled: Whether the validator is active in the catalog funnel.
+        skip_load_after_save: Skip load-validation when the same catalog
+            instance already validated the dataset on save in this process.
+        options: Keyword options forwarded to the validator/adapter.
+    """
+
+    class_path: str
+    on: tuple = ("load", "save")
```
ravi-kumar-pilla: **Suggestion:** `ValidatorSpec` field annotations use bare `tuple` and `dict` while the rest of the module uses parameterized types. Prefer `tuple[str, ...]` and `dict[str, Any]` for consistency and mypy clarity.

ravi-kumar-pilla (commented):
## Kedro PR Review

### Overview
- **Critical:** 0 | **Suggestions:** 5

No issues found.

- [ ] `kedro/validation/pandera_validator.py:L175` — Avoid `Check 'None'` in failure messages when check name is missing
- [ ] `kedro/validation/core.py:L256` — Detect `True` / `'on'` key collision during YAML normalisation
- [ ] `kedro/validation/core.py:L262` — Guard `sorted(unknown_keys)` against non-string config keys
- [ ] `kedro/validation/core.py:L125` — Truncate long failure labels so `str(exc)` stays bounded (esp. pyspark path)
- [ ] `kedro/validation/core.py:L214` — Tighten `ValidatorSpec` field type hints (`tuple[str, ...]`, `dict[str, Any]`)

### Notes
- General review: Focused, well-tested groundwork; resolution order and runtime-checkable protocol footgun are handled correctly. Most earlier automated review feedback appears addressed in `c59ae8a`.
- PR checklist: RELEASE.md, linked issue, tests, and DCO are in good shape. Docs checkbox is marked done with no `docs/` changes — acceptable for this internal groundwork PR; public docs expected in follow-up.
- API usage: Uses existing Kedro patterns (`load_obj`, Google docstrings, optional extras). Purely additive; no backward-compatibility concerns.

kedro/validation/core.py
```diff
@@ -0,0 +1,500 @@
...
+_VALID_SEVERITIES = ("error", "warn")
+
+#: Maximum number of failure examples captured per check.
+MAX_FAILURE_EXAMPLES = 5
+#: Maximum number of failed checks rendered by `DataValidationError.__str__`.
+_MAX_RENDERED_FAILURES = 10
+#: Maximum length of a rendered failure example.
+_MAX_EXAMPLE_REPR_LEN = 40
+#: Maximum length of a rendered fallback message.
+_MAX_MESSAGE_LEN = 500
```
merelcht: Can a user change these defaults?

kedro/validation/core.py
```diff
@@ -0,0 +1,500 @@
...
+
+    class_path: str
+    on: tuple = ("load", "save")
+    severity: str = "error"
+    enabled: bool = True
+    skip_load_after_save: bool = False
+    options: dict = field(default_factory=dict)
+
+    @classmethod
+    def from_config(cls, ds_name: str, config: Any) -> ValidatorSpec:
```
merelcht: Maybe name this `from_dataset_config` to make it extra clear what type of config it's referring to.

kedro/validation/core.py
```diff
@@ -0,0 +1,500 @@
...
+            raise ValidationConfigurationError(
+                f"Invalid validator declaration for dataset '{ds_name}': "
+                f"expected a class path string or a dictionary, got "
+                f"{type(config).__name__} ({config!r})."
+            )
+
+        if True in config:
+            # YAML 1.1 parses an unquoted `on` key as boolean True
+            # (the "Norway problem"). Normalise it back so users can write
+            # `on: [load, save]` without quoting the key.
```
merelcht: Can we make this comment clearer? Maybe link to the issue? This "norway problem" mention isn't really helpful without a link.

kedro/validation/core.py
```diff
@@ -0,0 +1,500 @@
...
+                    f"got {type(config[flag_name]).__name__}."
+                )
+
+        return cls(
+            class_path=class_path,
+            on=on,
+            severity=severity,
+            enabled=config.get("enabled", True),
+            skip_load_after_save=config.get("skip_load_after_save", False),
+            options=dict(options),
```
merelcht, line 208: Why do you need to cast options in a dict here if you're already checking the type is dict above?

kedro/validation/core.py
```diff
@@ -0,0 +1,500 @@
...
+                    f"'{flag_name}' must be a boolean, "
+                    f"got {type(config[flag_name]).__name__}."
+                )
+
+        return cls(
+            class_path=class_path,
+            on=on,
+            severity=severity,
+            enabled=config.get("enabled", True),
+            skip_load_after_save=config.get("skip_load_after_save", False),
```
merelcht: Nit: reads a bit cleaner if the assignment is done above so it's consistent with the rest.

kedro/validation/core.py
```diff
@@ -0,0 +1,500 @@
...
+    def validate(self, data: Any) -> Any:
+        """Call the wrapped function; any raise is a validation failure."""
+        result = self._func(data)
+        return data if result is None else result
+
+
+def _import_error_hint(missing: str) -> str:
+    """Build a hint for a failed validator import.
+
+    `missing` is the module or attribute name the import machinery reported
```
merelcht: import machinery? 😅

kedro/validation/core.py
```diff
@@ -0,0 +1,500 @@
...
+                f"accept options {sorted(spec.options)}; declare a class "
+                f"path instead or drop the options."
+            )
+        return obj
+
+    if callable(obj):
+        if spec.options:
+            raise ValidationConfigurationError(
+                f"Validator '{spec.class_path}' is a plain callable and "
+                f"cannot accept options {sorted(spec.options)}."
```
merelcht: Can we specify what the user should do to resolve this like in the error above?

kedro/validation/pandera_validator.py
```diff
@@ -0,0 +1,325 @@
...
+        return PanderaValidator(obj, **options)
+    except TypeError as exc:
+        name = getattr(obj, "__name__", type(obj).__name__)
+        raise ValidationConfigurationError(
+            f"Unsupported option(s) for pandera validator '{name}': {exc}. "
+            f"Supported options: lazy, head, tail, sample, random_state."
+        ) from exc
+
+
+def _failure_cases_records(failure_cases: Any) -> list[dict]:
```
merelcht, line 158: How are pyspark dataframes handled? Same as pandas or polars? Might be worth clarifying in the comments already there.

kedro/validation/pandera_validator.py
```diff
@@ -0,0 +1,325 @@
...
+    """
+    records = _failure_cases_records(getattr(exc, "failure_cases", None))
+    if not records:
+        return [CheckFailure(message=str(exc))]
+
+    grouped: dict[tuple, dict[str, Any]] = {}
+    for record in records:
+        column = _clean_column(record.get("column"))
+        check = record.get("check")
+        check = str(check) if check is not None else None
```
merelcht: Can we give one of those variables a different name so it's clear which "check" value we're dealing with?

merelcht (commented):
I did an initial review and left some comments & suggestions. I haven't looked at the tests yet, but will do that for the second round :)

kedro/validation/pandera_validator.py
```diff
@@ -0,0 +1,325 @@
...
+            sample: Validate a random sample of `sample` rows.
+            random_state: Seed for `sample`.
+        """
+        self._schema = schema
+        self._lazy = lazy
+        self._head = head
+        self._tail = tail
+        self._sample = sample
+        self._random_state = random_state
+        self._lazyframe_warned = False
```
ElenaKhaustova, line 290: **Suggestion:** `_lazyframe_warned` is mutable instance state read and written in `_maybe_warn_lazyframe` without synchronisation. The `Validator` protocol docstring says implementations "may be called concurrently and must therefore be thread-safe or stateless." In a `ThreadRunner` scenario two threads could both enter the method before either sets the flag, producing duplicate warnings.

The real-world impact is minimal (worst case: a repeated log line), but since the protocol explicitly documents a thread-safety contract, consider either:
- (a) documenting that this particular race is harmless, or
- (b) using a `threading.Lock` or an atomic flag pattern to be strict about the contract.

kedro/validation/pandera_validator.py
```diff
@@ -0,0 +1,325 @@
...
+            self._lazyframe_warned = True
+            logger.warning(
+                "Validating a polars LazyFrame with pandera: validation depth "
+                "may be schema-only (data-level checks require collection)."
+            )
+
+    def _validate_pyspark(self, data: Any) -> Any:
+        """Validate a pyspark DataFrame; pandera's pyspark backend never raises."""
+        out = self._schema.validate(data)
+        errors = getattr(getattr(out, "pandera", None), "errors", None)
```
ElenaKhaustova: **Suggestion:** The double-`getattr` chain silently passes validation if pandera ever changes its pyspark accessor API. When `out` has no `pandera` attribute, `getattr(out, "pandera", None)` returns `None`, then `getattr(None, "errors", None)` returns `None`, `if errors:` is `False`, and the method returns `out` without raising.

Consider logging a warning when the accessor is absent so users aren't silently unprotected if the API shifts:

```python
pandera_accessor = getattr(out, "pandera", None)
if pandera_accessor is None:
    logger.warning(
        "Pyspark validation ran but the pandera accessor was not found "
        "on the output DataFrame; validation results cannot be inspected."
    )
    return out
errors = getattr(pandera_accessor, "errors", None)
```

kedro/validation/core.py
```diff
@@ -0,0 +1,500 @@
...
+        if self.skip_load_after_save:
+            config["skip_load_after_save"] = self.skip_load_after_save
+        if self.options:
+            config["options"] = dict(self.options)
+        if not config:
+            return self.class_path
+        return {"class": self.class_path, **config}
+
+
+class CallableValidator:
```
ElenaKhaustova, line 235: **Suggestion:** Neither `CallableValidator` nor `PanderaValidator` defines `__repr__`. The default `<CallableValidator object at 0x…>` isn't helpful when debugging (e.g. inspecting catalog state or stepping through `resolve_validator`). A minimal repr showing the wrapped function name or schema class path would improve the experience:

```python
# CallableValidator
def __repr__(self) -> str:
    return f"CallableValidator({self._func!r})"

# PanderaValidator
def __repr__(self) -> str:
    return f"PanderaValidator({self._validator_path()})"
```

kedro/validation/pandera_validator.py
SajidAlamQB: Fixed, a failure with no check name now reads `Schema validation failed for column 'x'`

kedro/validation/core.py
SajidAlamQB: Makes sense and it was worse than silent last-wins. I traced it end to end and the dropped key turns off save validation. Now raises telling you to keep one on key. 👍

kedro/validation/core.py
SajidAlamQB: Went with your key=repr suggestion

kedro/validation/core.py
SajidAlamQB: Truncating each part separately rather than the composed line, so a long column name can't crowd out the check.

kedro/validation/core.py
SajidAlamQB: No, they are internal to make sure errors can be readable. I will add _ to make it more clear

kedro/validation/pandera_validator.py
SajidAlamQB: Its a different path entirely as pandera's pyspark backend never raises, it reports on an accessor on the returned DataFrame, so pyspark never reaches `_failure_cases_records`. Only pandas and polars do.

kedro/validation/core.py
SajidAlamQB: The `isinstance` check confirms the type, but the copy is for aliasing, without it the spec holds a reference to the dict inside the catalog config. `ValidatorSpec` is frozen but a dict inside it isn't.

kedro/validation/pandera_validator.py
SajidAlamQB: Tried (a) I did try to make it fire and there was 0 duplicates across 180000 tries and 90 real ThreadRunner runs at up to 16 threads. It only reproduces purposefully with settrace. 

The flag only moves False to True and never affects a validation result, so the worst case really is one repeated log line. 

I would avoid lock as that would make `PanderaValidator` unpicklable, which it isn't normally. I added Docstring why the flag is unsynchronised.

kedro/validation/pandera_validator.py
SajidAlamQB: Good catch, split the states rather than warning on all of them so missing accessor now raises while `errors is None` warns and passes, since switching validation off is an option. Shoudl we reduce the raise to a warning if you think we rather not hard fail on a pandera version issue?

kedro/validation/pandera_validator.py
```diff
@@ -0,0 +1,374 @@
...
+    raises, so its failures are read from the accessor on the returned
+    DataFrame instead (see `PanderaValidator._validate_pyspark`).
+    """
+    if failure_cases is None:
+        return []
+    # pandas DataFrame
+    to_dict = getattr(failure_cases, "to_dict", None)
+    if to_dict is not None and hasattr(failure_cases, "columns"):
+        try:
+            return cast("list[dict]", to_dict(orient="records"))
```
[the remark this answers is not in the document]
SajidAlamQB: Not quite, typing.cast accepts a string as its first argument, that's the standard forward-reference form. mypy --strict and ruff both pass on this file as it stands, and kedro already uses the same form in kedro/ipython/__init__.py.

SajidAlamQB:
Thanks @ravi-kumar-pilla @merelcht @ElenaKhaustova  I've addressed everything.

Ready for a re-review

kedro/validation/core.py
merelcht: Makes sense, thanks for clarifying!

kedro/validation/pandera_validator.py
merelcht: Got it!

kedro/validation/core.py
```diff
@@ -0,0 +1,524 @@
...
+#: Maximum number of failed checks rendered by `DataValidationError.__str__`.
+_MAX_RENDERED_FAILURES = 10
+#: Maximum length of a rendered failure example.
+_MAX_EXAMPLE_REPR_LEN = 40
+#: Maximum length of a rendered check name or fallback message.
+_MAX_LABEL_LEN = 200
+#: Maximum length of a rendered column name.
+_MAX_COLUMN_LEN = 60
+#: Maximum length of a rendered fallback message.
+_MAX_MESSAGE_LEN = 500
```
merelcht: **Nit — constant docs.** Two issues here:

- `#:` Sphinx-style attribute comments don't appear anywhere else in `kedro/`; the project renders API docs with mkdocstrings, so these read as an import from another toolchain.
- `_MAX_LABEL_LEN` (line 44) and `_MAX_MESSAGE_LEN` (line 48) both document themselves as bounding "a rendered fallback message". One of the two descriptions is wrong.

kedro/validation/core.py
```diff
@@ -0,0 +1,524 @@
...
+
+def _truncate(value: Any, limit: int) -> str:
+    """Render `value` as a string of at most `limit` characters."""
+    text = str(value)
+    if len(text) > limit:
+        text = text[: limit - 3] + "..."
+    return text
+
+
+class DataValidationError(Exception):
```
merelcht: On the one hand it makes sense to have`DataValidationError` and `ValidationConfigurationError` here, but at the same time I wonder if they belong in `exceptions.py`.

kedro/validation/core.py
```diff
@@ -0,0 +1,524 @@
...
+        failure_examples: Sample of failing values (capped, default cap 5).
+        index: Index/location of the first failure case, if available.
+    """
+
+    message: str
+    check: str | None = None
+    column: str | None = None
+    failure_count: int = 1
+    failure_examples: list[Any] = field(default_factory=list)
+    index: Any | None = None
```
merelcht: `CheckFailure.index` is populated (in both `_failures_from_schema_errors` and `_failure_from_schema_error`) but never read — it isn't rendered by `__str__` and nothing else consumes it. Either render it or drop it until something needs it.

kedro/validation/core.py
```diff
@@ -0,0 +1,524 @@
...
+    def validate(self, data: Any) -> Any:
+        """Validate `data` and return the (possibly transformed) data.
+
+        ANY exception raised by this method is treated as a validation
+        failure. Implementations may be called concurrently and must therefore
+        be thread-safe or stateless.
+        """
+
+
+@dataclass(frozen=True)
```
merelcht: **`frozen=True` promises something it doesn't deliver** (same for `ValidatorSpec` at line 201). Both dataclasses hold mutable containers, so:

```python
>>> hash(CheckFailure(message="m"))
TypeError: unhashable type: 'list'

>>> s = ValidatorSpec.from_dataset_config("ds", {"class": "a.B", "options": {"lazy": True}})
>>> s.options["lazy"] = False   # succeeds
>>> s.to_config()
{'class': 'a.B', 'options': {'lazy': False}}
```

`frozen=True` here blocks only attribute rebinding while reading as an immutability guarantee, and it silently costs hashability. Given `test_options_copied_not_aliased` exists, real immutability was the intent — so either use `tuple` / `MappingProxyType` for the containers, or drop `frozen=True` so it doesn't over-promise.

kedro/validation/core.py
```diff
@@ -0,0 +1,524 @@
...
+        self.failures: list[CheckFailure] = list(failures) if failures else []
+
+    @staticmethod
+    def _render_failure(failure: CheckFailure) -> str:
+        """Render one grouped check failure as a single bounded line.
+
+        Each part is capped separately so that a long column name still leaves
+        room for the check that failed.
+        """
+        detail = _truncate(failure.check or failure.message, _MAX_LABEL_LEN)
```
merelcht: The per-part caps are the right idea, but the resulting bound is much looser than the tests claim. With examples present a single line reaches ~490 characters:

```python
f = CheckFailure(message="m", check="c"*5000, column="col"*5000,
                 failure_examples=["e"*500]*5)
max(len(l) for l in str(DataValidationError("boom", dataset_name="ds", failures=[f]*10)).splitlines())
# 491
```

`_MAX_COLUMN_LEN` (60) + `_MAX_LABEL_LEN` (200) + 5 x `_MAX_EXAMPLE_REPR_LEN` (40) + separators. That's still bounded, which is what matters — but see my comment on `test_every_rendered_line_is_bounded`, which asserts `< 300` and only passes because none of its cases sets `failure_examples`.

kedro/validation/core.py
```diff
@@ -0,0 +1,524 @@
...
+        return "\n".join(lines)
+
+
+class ValidationConfigurationError(Exception):
+    """Raised when a `validator:` declaration is invalid or unresolvable."""
+
+    pass
+
+
+def _parse_on_modes(value: Any, ds_name: str) -> tuple[str, ...]:
```
merelcht, line 60: **`on` isn't normalised.** Duplicates and ordering both survive parsing:

```python
>>> ValidatorSpec.from_dataset_config("ds", {"class": "x.Y", "on": ["load", "load"]}).to_config()
{'class': 'x.Y', 'on': ['load', 'load']}
```

and `on: [save, load]` is a different spec from `on: [load, save]` (the latter collapses to the shorthand string, the former doesn't). Dedupe and canonicalise the order here so `on` has one representation.

kedro/validation/core.py
```diff
@@ -0,0 +1,524 @@
...
+            on=on,
+            severity=severity,
+            enabled=enabled,
+            skip_load_after_save=skip_load_after_save,
+            # Copy so the spec never aliases the catalog configuration it was
+            # parsed from, and cannot mutate it later through `options`.
+            options=dict(options),
+        )
+
+    def to_config(self) -> str | dict[str, Any]:
```
merelcht, line 211: `to_config()` has no consumer anywhere in the codebase — it exists only to be round-trip tested. Same for `preflight_check` (line 485), `CheckFailure.index`, and the `on` / `severity` / `enabled` / `skip_load_after_save` fields, none of which anything acts on yet.

That's expected for a groundwork PR, but these are exported from `kedro.validation.__init__`, which commits to a public surface ahead of #5680 — the PR whose whole job is reviewing that surface. I'd keep them private (or unexported) and promote them in the PR that actually uses them.

kedro/validation/core.py
```diff
@@ -0,0 +1,524 @@
...
+    def validate(self, data: Any) -> Any:
+        """Call the wrapped function; any raise is a validation failure."""
+        result = self._func(data)
+        return data if result is None else result
+
+    def __repr__(self) -> str:
+        return f"{type(self).__name__}({self._func!r})"
+
+
+def _import_error_hint(missing: str) -> str:
```
merelcht, line 254: The pandera install hint is written twice — here and in `preflight_check` (line ~500) — with divergent wording and trailing punctuation ("`... for your backend)`" vs "`... for your backend).`"). Extract a single module-level constant so the two can't drift.

kedro/validation/core.py
```diff
@@ -0,0 +1,524 @@
...
+        if result is not None:
+            return result
+
+    if inspect.isclass(obj):
+        # NEVER isinstance-check the class object itself against the
+        # runtime-checkable Protocol: it matches any class merely defining
+        # a `validate` method and would return the uninstantiated class,
+        # silently dropping options.
+        try:
+            instance = obj(**spec.options)
```
merelcht, line 331: **Two problems in this block.**

**1. Only `TypeError` is caught, so unrelated constructor failures escape the funnel as raw backend exceptions:**

```python
>>> resolve_validator(ValidatorSpec(class_path="decimal.Decimal", options={"value": "not-a-number"}))
decimal.InvalidOperation: [<class 'decimal.ConversionSyntax'>]
```

Any validator whose `__init__` validates its own arguments — `ValueError`, a pydantic `ValidationError`, pandera's own errors — will surface with no mention of the dataset or the class path, which defeats the point of `ValidationConfigurationError`. Catch `Exception` and wrap.

**2. Every imported class is instantiated *before* it's checked for being a validator:**

```python
>>> resolve_validator(ValidatorSpec(class_path="pathlib.Path"))
ValidationConfigurationError: Validator 'pathlib.Path' was instantiated but does not provide a 'validate(data)' method.
```

A `Path()` was constructed on the way to that error. A typo'd class path can run an arbitrary, expensive or side-effecting `__init__`. The comment above correctly explains why you must not `isinstance`-check the class object against the `runtime_checkable` protocol — but a plain `callable(getattr(obj, "validate", None))` check on the class is safe and would let you reject before constructing.

kedro/validation/core.py
```diff
@@ -0,0 +1,524 @@
...
+        return CallableValidator(obj)
+
+    raise ValidationConfigurationError(
+        f"Validator '{spec.class_path}' resolved to "
+        f"{type(obj).__name__} ({obj!r}), which is not a validator class, "
+        f"a validator instance, a supported schema or a callable."
+    )
+
+
+def preflight_check(specs: dict[str, ValidatorSpec]) -> list[str]:
```
merelcht, line 372: See my note on `to_config` — `preflight_check` is likewise unused and publicly exported. Two smaller things if it stays:

- The local is named `warnings`, which shadows the stdlib module name for anyone reading the function body (no bug today, since `warnings` isn't imported here).
- Returning warning strings for a caller to emit, rather than emitting them, is a reasonable choice but leaves the contract undefined until #5679 lands. Worth a line in the docstring on who is expected to log these.

kedro/validation/pandera_validator.py
```diff
@@ -0,0 +1,374 @@
...
+
+    Raises:
+        ValidationConfigurationError: If `options` contains keys not
+            supported by `PanderaValidator`.
+    """
+    if not (_is_pandera_model(obj) or _is_pandera_schema_instance(obj)):
+        return None
+    try:
+        return PanderaValidator(obj, **options)
+    except TypeError as exc:
```
merelcht: **Bare `except TypeError` over the whole constructor call will mislabel failures.** Today `PanderaValidator.__init__` only assigns attributes, so this is latent — but any future `TypeError` raised inside it gets reported to the user as "Unsupported option(s)".

The related issue is line 110: `"Supported options: lazy, head, tail, sample, random_state"` is hardcoded and will silently drift from `PanderaValidator.__init__`'s signature. Both are fixed by the same change — diff `options.keys()` against `inspect.signature(PanderaValidator.__init__).parameters` up front, build the message from the signature, and drop the `try`/`except`.

kedro/validation/pandera_validator.py
```diff
@@ -0,0 +1,374 @@
...
+
+    def _validator_path(self) -> str:
+        """Dotted path of the schema class for error reporting."""
+        target = self._schema if inspect.isclass(self._schema) else type(self._schema)
+        return f"{target.__module__}.{target.__qualname__}"
+
+    def _is_pyspark_dataframe(self, data: Any) -> bool:
+        """Check whether `data` is a pyspark DataFrame (guarded imports)."""
+        try:
+            import pandera.pyspark  # noqa: F401
```
merelcht: **Per-`validate()` imports on the catalog hot path.** This runs on every call, as do `import polars` (line 281) and `import pandera.errors` (line 349). `sys.modules` lookups are cheap but not free, and this sits on every dataset load/save.

Worth noting the `_lazyframe_warned` latch doesn't help: it's only set when the data *is* a `LazyFrame`, so a pandas user pays the polars import lookup on every single load forever. Resolving the types once at module level (or via a memoised helper) keeps the optional-dependency behaviour while taking the work off the hot path.

kedro/validation/pandera_validator.py
```diff
@@ -0,0 +1,374 @@
...
+
+        The flag is not synchronised: concurrent callers can both pass the
+        guard, but it only ever moves `False` -> `True` and never affects a
+        validation result, so the worst case is a repeated log line.
+        """
+        if self._lazyframe_warned:
+            return
+        try:
+            import polars as pl
+        except ImportError:
```
merelcht: **This will break the coverage gate for anyone who has polars installed.** These two lines are only reachable when polars is *absent*, and CI is green today purely because polars isn't in the `test` extra. With polars in the environment:

```
kedro/validation/pandera_validator.py   150   2   99%   282-283
```

and `pyproject.toml` sets `fail_under = 100`, so `make test` fails locally for any contributor who has polars — or who installs the `kedro[pandera-polars]` extra this PR adds. Coverage shouldn't depend on which optional backends happen to be present.

Fix is `except ImportError:  # pragma: no cover`, matching the existing convention in this same package at `kedro/validation/utils.py:12`.

merelcht (commented):
Thanks @SajidAlamQB, I ran another review using Claud, which flagged some good things. They are added inline at the relevant points in the code.

pyproject.toml
```diff
@@ -100,6 +101,10 @@ benchmark = [
     "kedro[pydantic]",
 ]
 
+pandera-pandas = ["pandera[pandas]>=0.24"]
+pandera-polars = ["pandera[polars]>=0.24"]
+pandera-pyspark = ["pandera[pyspark]>=0.24"]
```
merelcht: **`pandera-pyspark` is advertised, which contradicts the PR description.** The description says pyspark "handling exists but is not advertised", but this is a published extra and it's named in two user-facing hint strings (`_import_error_hint` and `preflight_check`). Meanwhile the pyspark code path is only exercised against `SimpleNamespace` fakes — no real pyspark DataFrame is ever validated. Either drop the extra from this PR or own the support with a real test.

Separately: `>=0.24` here (presumably for the `pandera.pandas` namespace) is worth a brief comment, and it now disagrees with `docs/integrations-and-plugins/pandera.md`, which tells users `pandera[pandas]>=0.18.0`. That page also teaches the hook-based approach this feature is meant to replace — the PR checklist claims docs were updated, but nothing in `docs/` changed.

pyproject.toml
```diff
@@ -63,6 +63,7 @@ test = [
     "kedro-datasets",
     "mypy>=1,<3",
     "pandas>=2.0,<4.0",
+    "pandera[pandas]>=0.24",
```
merelcht, line 66: Adding `pandera[pandas]` to the core `test` extra means every contributor's dev install pulls pandera and its dependency chain. Probably the right call given the tests import it unconditionally — but note the consequence: there's now no environment in which the genuinely-pandera-absent paths run, so they're only covered by `sys.modules` injection.

tests/validation/test_pandera_validator.py
```diff
@@ -0,0 +1,570 @@
...
+        ).validate(df)
+        assert schema.kwargs == {
+            "lazy": False,
+            "head": 2,
+            "sample": 10,
+            "random_state": 42,
+        }
+
+
+class TestErrorRendering:
```
merelcht, line 187: **These tests are in the wrong file.** `TestErrorRendering`, `TestErrorRenderingEdgeCases` (line 249), and the two orphan module-level tests at the bottom (lines 562, 567) all test `DataValidationError.__str__` and `CheckFailure`, which live in `kedro/validation/core.py` — not in the pandera adapter. Someone changing `__str__` will not think to look in `test_pandera_validator.py`.

This is part of a broader layout issue: `tests/validation/` is otherwise one test file per module (`test_exceptions.py`, `test_model_factory.py`, `test_type_extractor.py`, `test_utils.py`), but `core.py` here is split across `test_validator_spec.py` and `test_resolution.py` with no `test_core.py` at all. Suggest a `test_core.py` that owns everything in `core.py`, leaving this file for the adapter.

tests/validation/test_pandera_validator.py
```diff
@@ -0,0 +1,570 @@
...
+    return pd.DataFrame({"id": [1, 2, 3], "company_rating": [0.0, 2.5, 10.0]})
+
+
+@pytest.fixture
+def invalid_df():
+    # two failed checks: id uniqueness (1 case pair) and rating >= 0 (2 cases)
+    return pd.DataFrame({"id": [10542, 10542, 3], "company_rating": [-0.5, -1.2, 1.0]})
+
+
+class TestDetectionHelpers:
```
merelcht, line 54: **Most of this class tests `issubclass`/`isinstance` rather than the adapter.** `_is_pandera_model(dict) is False`, `_is_pandera_model("not a class") is False`, `_is_pandera_schema_instance(pd.DataFrame()) is False` — these assert language semantics through a private helper, and the behaviour that actually matters is already covered by `TestPanderaAdapter.test_non_pandera_object_returns_none`, which parametrises over `dict`, `"string"`, `42`, `pd.DataFrame` at the public boundary.

I'd keep only the two cases that encode a real decision — model class detected, schema *class* not accepted as an instance — and drop the rest.

tests/validation/test_pandera_validator.py
```diff
@@ -0,0 +1,570 @@
...
+    def test_model_class_adapted(self):
+        validator = pandera_adapter(CompaniesSchema, {})
+        assert isinstance(validator, PanderaValidator)
+
+    def test_schema_instance_adapted(self):
+        schema = pa.DataFrameSchema({"id": pa.Column(int)})
+        validator = pandera_adapter(schema, {})
+        assert isinstance(validator, PanderaValidator)
+
+    def test_options_forwarded(self):
```
merelcht: This asserts five private attributes; `test_only_non_none_kwargs_forwarded_to_schema_validate` (line 179) already proves the same property behaviourally, by observing what reaches `schema.validate`. Keep the behavioural one and drop this.

tests/validation/test_pandera_validator.py
```diff
@@ -0,0 +1,570 @@
...
+        [
+            CheckFailure(message="m", check="c" * 5000, column="col"),
+            CheckFailure(message="m", check="c" * 5000),
+            CheckFailure(message="m" * 5000, column="col"),
+            CheckFailure(message="m" * 5000),
+            CheckFailure(message="m", check="c", column="col" * 5000),
+        ],
+        ids=["column+check", "check", "column+message", "message", "long-column"],
+    )
+    def test_every_rendered_line_is_bounded(self, failure):
```
merelcht: **The name overpromises and the bound is wrong.** None of the five parametrised cases sets `failure_examples`, so this only exercises the label path. With examples the same rendering produces lines of ~490 characters:

```python
f = CheckFailure(message="m", check="c"*5000, column="col"*5000,
                 failure_examples=["e"*500]*5)
# longest rendered line: 491
```

So "every rendered line is bounded < 300" isn't true — it's the label that's bounded. Either add a long-examples case and assert the real bound (~500), or rename to `test_rendered_label_is_bounded` so the guarantee matches the assertion.

tests/validation/test_pandera_validator.py
```diff
@@ -0,0 +1,570 @@
...
+
+
+class TestOptionalBackendDetection:
+    def test_is_pandera_model_without_pandera(self, monkeypatch):
+        import sys
+
+        monkeypatch.setitem(sys.modules, "pandera.api.dataframe.model", None)
+        assert _is_pandera_model(CompaniesSchema) is False
+
+    def test_schema_instance_detection_with_fake_backends(self, monkeypatch):
```
merelcht, line 331: **This test can't fail for the reason it implies.** It injects `FakePolarsSchema`/`FakePySparkSchema` into `sys.modules` and then asserts `isinstance(FakePolarsSchema(), ...)` is `True` — i.e. it verifies that the try/except/append scaffolding is shaped the way it's shaped. It would pass unchanged if the real `pandera.api.polars.container.DataFrameSchema` were never detected.

Also, `monkeypatch.setitem(sys.modules, name, None)` (used at lines 421 and 446 to force an `ImportError`) leans on a fairly obscure CPython behaviour and carries no comment explaining it. Worth one line, since the next reader will assume it's a no-op.

tests/validation/test_pandera_validator.py
```diff
@@ -0,0 +1,570 @@
...
+
+def test_failures_from_error_without_failure_cases():
+    from kedro.validation.pandera_validator import _failures_from_schema_errors
+
+    assert _failures_from_schema_errors(ValueError("plain")) == [
+        CheckFailure(message="plain")
+    ]
+
+
+def test_adapter_unsupported_option_is_config_error():
```
merelcht: This is a duplicate of `TestPanderaAdapter.test_unknown_option_raises_config_error` (line 98), and weaker — it only regex-matches `"Supported options"` where the other asserts the schema name and the full option list. Delete this one.

tests/validation/test_pandera_validator.py
```diff
@@ -0,0 +1,570 @@
...
+        assert _failure_cases_records(["bad"]) == [{"failure_case": "bad"}]
+        assert _failure_cases_records([{"a": 1}]) == [{"a": 1}]
+
+    def test_unknown_object_returns_empty(self):
+        from kedro.validation.pandera_validator import _failure_cases_records
+
+        assert _failure_cases_records(object()) == []
+
+
+class TestCleanCell:
```
merelcht, line 298: `test_variants` bundles four unrelated assertions under a name that says nothing about what's being checked; on failure you get "assert None is None"-style output with no indication of which case broke. Either parametrise (`None`, `nan`, `str`, `int`) or fold these into the `_failures_from_schema_errors` tests that already cover the nan-check case at line 324.

Same observation applies to `TestFailureCaseRecords` above: `test_none_returns_empty` and `test_unknown_object_returns_empty` are testing a private normaliser's guard clauses; the pandas-shape and polars-shape cases are the ones that encode a real decision.

tests/validation/test_pandera_validator.py
```diff
@@ -0,0 +1,570 @@
...
+        monkeypatch.setattr(
+            validator, "_schema", types.SimpleNamespace(validate=lambda data: out)
+        )
+
+        with caplog.at_level(logging.WARNING):
+            assert validator._validate_pyspark(object()) is out
+        assert "not validated" in caplog.text
+
+
+def test_header_includes_mode_when_set():
```
merelcht: These two module-level tests are duplicates of assertions made above:

- `test_header_includes_mode_when_set` is a strict subset of `test_enriched_error_renders_per_spec` (line 206), which already asserts `"Validation failed for dataset 'companies' on load"`.
- `test_excess_failures_render_hidden_note` (line 567) duplicates `test_rendering_caps_number_of_checks` (line 235), which already asserts the `... and N more check(s)` line.

Both can go. They also belong in a `test_core.py` rather than here.

tests/validation/test_resolution.py
```diff
@@ -0,0 +1,326 @@
...
+
+
+NOT_A_VALIDATOR = 42
+
+
+# --- Tests ------------------------------------------------------------------
+
+
+class TestPanderaResolution:
+    def test_dataframe_model_class_routes_through_adapter(self):
```
merelcht: `assert validator._schema is CompaniesSchema` reaches into a private attribute to prove routing. `test_resolved_pandera_validator_validates` (line 109) makes the same point behaviourally. If you want both, this one could assert on `repr(validator)` (which is public and already tested to name the schema) instead of `_schema`.

tests/validation/test_resolution.py
```diff
@@ -0,0 +1,326 @@
...
+    def test_non_validator_object_is_config_error(self):
+        spec = ValidatorSpec(class_path=f"{_MODULE}.NOT_A_VALIDATOR")
+        with pytest.raises(
+            ValidationConfigurationError, match="not a validator"
+        ) as exc_info:
+            resolve_validator(spec)
+        assert "int" in str(exc_info.value)
+
+
+class TestPreflightCheck:
```
merelcht: **Missing the one structural test that matters.** Nothing in the suite asserts that the resolved validators actually satisfy the `Validator` protocol:

```python
assert isinstance(resolve_validator(spec), Validator)   # PanderaValidator
assert isinstance(CallableValidator(fn), Validator)
```

That's the contract the catalog funnel in #5679 will depend on, and it's the only claim in this PR that's currently untested. Worth adding here (or in a `test_core.py`).

tests/validation/test_validator_spec.py
```diff
@@ -0,0 +1,255 @@
+"""Tests for `ValidatorSpec` parsing and serialisation."""
+
+from __future__ import annotations
+
+import pytest
+
+from kedro.validation import ValidationConfigurationError, ValidatorSpec
+
+
+class TestFromDatasetConfigShorthand:
```
merelcht: Single-test class that adds no grouping — this could just be a module-level `test_string_shorthand`, or live in `TestFromDatasetConfigLongForm`'s neighbour as a plain function.

ElenaKhaustova (approved):
LGTM 👍

kedro/validation/core.py
SajidAlamQB: I agree actually it makes more sense

kedro/validation/core.py
SajidAlamQB: These do have consumers in #5715: it uses to_config() for catalog re-injection and preflight_check in the context wiring, and the funnel gates on the spec fields. Kept them exported for that, happy to finalise the public surface in #5680.

kedro/validation/core.py
SajidAlamQB: Renamed the local to messages and added a line to the docstring saying nothing is logged here and the caller decides how to emit them. In #5715 that caller is the context wiring, which logs each warning at catalog creation.

tests/validation/test_pandera_validator.py
SajidAlamQB: Trimmed to the two cases you suggested, model class detected and schema class not accepted as an instance. The rejection cases are covered at the public boundary by test_non_pandera_object_returns_none.

tests/validation/test_pandera_validator.py
SajidAlamQB: I kept the test because it's what makes the polars/pyspark import branches run whatever is installed, but the comment now says that's its job rather than implying it proves real backend detection. Also added a line explaining the sys.modules None trick where it's first used.

kedro/validation/core.py
merelcht: Makes sense!

tests/validation/test_pandera_validator.py
merelcht: Sounds good!

merelcht (approved):
Thanks @SajidAlamQB, this looks great now ⭐

Issue #5655, opened 2026-07-08
Thread safety for pipelines in `KedroServiceSession`


ankatiyar:
## Description
Follow up from #5535 
The pipeline loading mechanism at the moment is not thread safe for concurrent runs due to the recent optimisation added for selective pipeline loading.

## Context
Full context here: https://github.com/kedro-org/kedro/issues/5535#issuecomment-4914222307

The central issue is `_ProjectPipelines`. It is a process-global singleton with a lazy-loading cache. Its job is to defer the import of user pipeline modules until they are first needed, which is valuable for CLI startup speed. But for a long-running server, this deferred-load mechanism becomes a liability.

`_ProjectPipelines` holds three pieces of mutable state:

- `_pipelines_module` — which module to call register_pipelines() on
- `_is_data_loaded` — whether the pipeline dict has been populated
- `_content` — the populated dict of pipeline name → Pipeline object
- `_requested_pipelines` — a filter hint for which pipelines to load
`KedroServiceSession.run()` calls `pipelines.set_requested(pipeline_names)` at the start of every run. This is an optimisation: tell the loader to only import the modules for the requested pipelines, avoiding the cost of importing everything. `set_requested()` does this by invalidating the cache when the filter changes:
```
def set_requested(self, pipeline_names):
    if set(self._requested_pipelines or []) != set(pipeline_names or []):
        self._is_data_loaded = False   # invalidate
        self._content = {}             # clear
    self._requested_pipelines = list(pipeline_names) if pipeline_names else None
```
None of these three assignments is atomic relative to the others, and the entire operation is not protected by any lock. With two concurrent runs requesting different pipelines, the sequence becomes:
```

Thread 1: set_requested(["ingest"])      → _is_data_loaded=False, _content={}
Thread 2: set_requested(["train"])       → _is_data_loaded=False, _content={}, filter="train"
Thread 1: pipelines["ingest"]           → _load_data() fires with filter="train"
                                           → loads "train" pipeline only
                                           → Thread 1 gets KeyError for "ingest"
```

Secondary concern is the `_hook_manager` is created once and shared across every run. 

## Possible Implementation

This optimisation works for CLI based workflows but for serving use cases we could just preload all the pipelines.
- `mode=CLI/serving` during session creation which allows for this optimisation when in CLI mode but for serving mode will load all pipelines 

## Possible Alternatives

- Incrementally build pipelines snapshot:
```
[...]

PR #5695, opened 2026-07-27
Fix thread safety in `_ProjectPipelines`

ankatiyar:
## Description

Part of #5655 

## Development notes

`_ProjectPipelines` (the global pipelines object) loads pipeline definitions lazily on first access. In multi-threaded environments — like Kedro's HTTP server or parallel test runners — concurrent access could cause two bugs:

- Double loading: two threads both saw _is_data_loaded = False and both called register_pipelines(), loading pipelines twice.
- Torn reads: one thread loaded the data, then a second thread cleared it via configure() or set_requested(), causing the first thread to operate on an empty dict.

**Fix**: add a `threading.RLock` and hold it across all state mutations and reads. The wrapper that proxies dict operations (`__getitem__`, `__len__`, etc.) now captures the content reference under the lock, so no other thread can clear it mid-operation.

`RLock` (reentrant) is used instead of a plain Lock so that the outer lock in the wrapper and the inner lock in `_load_data()` can nest from the same thread. 
**The trade-off**: if user `register_pipelines()` code tries to access `pipelines[...]` during loading, it will recurse infinitely rather than deadlock.

> [!IMPORTANT]

## Checklist


kedro/framework/project/__init__.py
```diff
@@ -204,28 +206,27 @@ def __init__(self) -> None:
...
-            return
-
-        register_pipelines = self._get_pipelines_registry_callable(
-            self._pipelines_module
-        )
-        project_pipelines = register_pipelines()
-
-        self._content = project_pipelines
-        self._is_data_loaded = True
+        with self._lock:
```
ElenaKhaustova, line 242: **Critical:** Holding `self._lock` across `_get_pipelines_registry_callable()` (which calls `importlib.import_module`) and `register_pipelines()` (arbitrary user code) creates a lock-order inversion against Python's per-module import locks. It is a hard hang with no traceback, and it is a regression from `main`.

Repro: a module whose body touches `pipelines` at import time, being imported by thread B, while thread A triggers the first load. Thread stacks at the hang:

```
Thread A:  inner (__init__.py:180, inside `with self._lock`)
             -> _load_data (:221)
               -> _get_pipelines_registry_callable (:215)
                 -> importlib.import_module  ->  BLOCKED on the module import lock

Thread B:  <module body>  (holding that module's import lock)
             -> inner (__init__.py:180)  ->  BLOCKED acquiring self._lock
```

Same script, swapping only this file:

```
source = main (no lock):   {'A_finished': True,  'B_finished': True,  'outcomes': ['B', 'A']}
source = this branch:      {'A_finished': False, 'B_finished': False, 'outcomes': []}
```

There is a genuine design tension here, so I would rather flag it than prescribe a fix: you cannot guarantee both 'loads exactly once' and 'never hold the lock across user imports' with a plain mutex. A load-in-progress `Condition` does not solve it either, because thread B would still wait on A while holding the import lock that A needs.

Worth weighing that the two failure modes are not equally bad, though. Duplicate loading is idempotent — `register_pipelines()` returns a fresh dict, so loading twice wastes time but yields a correct result, and that was the behaviour on `main`. A deadlock is unrecoverable. That argues for scoping the lock to the state mutations only and accepting rare duplicate loads, which would mean reframing the exactly-once test. Your call on the trade-off, but as it stands this swaps a performance wart for a hang.

kedro/framework/project/__init__.py
```diff
@@ -204,28 +206,27 @@ def __init__(self) -> None:
...
 
     @staticmethod
     def _get_pipelines_registry_callable(pipelines_module: str) -> Any:
         module_obj = importlib.import_module(pipelines_module)
         register_pipelines = getattr(module_obj, "register_pipelines")
         return register_pipelines
 
-    def _load_data(self) -> None:
+    def _load_data(self) -> dict[str, Pipeline]:
         """Lazily read pipelines defined in the pipelines registry module."""
```
ElenaKhaustova: **Suggestion:** The signature changed from `-> None` to `-> dict[str, Pipeline]`, so this docstring needs a Google-style `Returns:` section. Worth being explicit that it returns the (empty) `_content` in the not-configured case as well as after a successful load, since that is not obvious from the name.

kedro/framework/project/__init__.py
```diff
@@ -204,28 +206,27 @@ def __init__(self) -> None:
         self._is_data_loaded = False
         self._content: dict[str, Pipeline] = {}
         self._requested_pipelines: list[str] | None = None
+        # RLock: allows configure()/set_requested() re-entry from the same thread.
```
ElenaKhaustova: **Suggestion:** This comment explains a re-entrancy that never happens. `configure()` and `set_requested()` are leaf methods that take the lock exactly once and are never re-entered. The nesting that actually requires `RLock` is `inner` at line 180 taking the lock and then calling `_load_data()` at line 221, which takes it again on the same thread. Suggested rewording:

```python
# RLock because _load_data_wrapper takes the lock and then calls
# _load_data(), which takes it again on the same thread.
```

The second line is also slightly misleading, in that the constraint is not new. On `main`, `register_pipelines()` accessing `pipelines[...]` already recursed forever, since `_is_data_loaded` is only set after the call returns (I confirmed `RecursionError` on both `main` and this branch). Using `RLock` deliberately preserves that behaviour rather than turning it into a deadlock, which is the right call — it may read more clearly if the comment says so.

kedro/framework/project/__init__.py
```diff
@@ -176,8 +177,9 @@ def _load_data_wrapper(func: Any) -> Any:
     """
 
     def inner(self: Any, *args: Any, **kwargs: Any) -> Any:
-        self._load_data()
-        return func(self._content, *args, **kwargs)
+        with self._lock:
+            content = self._load_data()
+            return func(content, *args, **kwargs)
```
ElenaKhaustova, line 185: **Suggestion (low priority, latent rather than live):** Capturing `content` under the lock correctly protects against `configure()` and `set_requested()`, because those *rebind* `_content` rather than mutating it, so the captured object stays internally consistent. That is the insight that makes this approach work.

It does not cover in-place mutation, though. `__iter__`, `keys`, `values` and `items` return live views/iterators over the captured dict, and the lock is released before the caller iterates, so a concurrent `__setitem__` / `__delitem__` can break the reader. A Python-level `for` loop over the escaped view plus a mutating thread reproduces `RuntimeError: dictionary changed size during iteration` immediately.

To be clear about the limits of this: I could not find a reachable call site in Kedro today. I initially suspected `get_close_matches(name, pipelines.keys())` in `session.py`, but it does `targets = list(targets)`, and `list()` over a dict view runs entirely in C without releasing the GIL, so it cannot interleave (verified — zero errors under sustained hammering). So this is latent rather than a live bug, and it would become real on a free-threaded build. Snapshotting the view-producing methods would close it if you want the wrapper to fully deliver on the guarantee described in the PR body.

ElenaKhaustova (commented):
## Kedro PR Review

**Edited after posting:** the import-lock deadlock is downgraded from Critical to a Suggestion, and I've retracted part of my comment on the `RLock` note. Both are corrected in replies on the relevant threads; the counts below reflect the revision.

Nice, focused PR, and the core insight is right: `configure()` / `set_requested()` rebind `_content` rather than mutating it, which is exactly why capturing the reference under the lock gives readers a consistent snapshot. The exactly-once test is a good one and genuinely pins the double-loading fix.

I verified every finding by running the real code on this branch (Python 3.13, editable install in a worktree) and re-running the same scripts with only `kedro/framework/project/__init__.py` reverted to `main`, so the regression claims are measured rather than inferred.

**The one blocking issue:** two of the three new tests pass 20/20 against the un-fixed source, so they don't protect the fix. Only the exactly-once test actually fails without the lock.

**The most interesting non-blocking issue:** holding the lock across `importlib.import_module` and user `register_pipelines()` code deadlocks against Python's per-module import locks. Same script, both threads complete on `main` and both hang forever on this branch. It's a real regression, but every repro I could build was synthetic, so a docstring note about the constraint is probably the right response rather than a redesign.

I have skipped everything Copilot already raised (thread join timeouts, daemon threads, swallowed exceptions in the invalidator/reconfigurer threads, and the original comment about the lock being released before the dict operation, which later commits fixed).

### Overview
- **Critical:** 1 | **Suggestions:** 6

- [ ] `tests/framework/project/test_pipeline_registry.py:162,196` — both tests pass 20/20 against the un-fixed source; they assert only that nothing raised, which the buggy code also satisfies

- [ ] `kedro/framework/project/__init__.py:221` — lock held across `importlib.import_module` and user `register_pipelines()` can deadlock against per-module import locks. Verified regression (`main` completes, this branch hangs), but only reproducible synthetically — documenting the constraint is likely enough
- [ ] `kedro/framework/project/__init__.py:209` — the `RLock` comment omits the nesting that fires on every load (`inner`:180 -> `_load_data`:221). See my correction in that thread: the case it *does* describe is legitimate, contrary to what I first wrote
- [ ] `kedro/framework/project/__init__.py:220` — return type changed, docstring has no `Returns:` section (minor: private method, not in the API docs)
- [ ] `kedro/framework/project/__init__.py:182` — view-returning methods escape the lock (latent on CPython, real on free-threaded builds)
- [ ] `kedro/framework/project/__init__.py:241` — the `set_requested()` + read sequence in `KedroSession.run` / `ServiceSession.run` is still racy. Pre-existing, not a regression — but worth softening the description and opening a follow-up under #5655. Note the failure ratio I quoted comes from a model with a stubbed loader, not from a real concurrent session run
- [ ] Tick the PR checklist boxes — every underlying requirement is actually met

### Notes
- **General review:** single-purpose and easy to follow. Only the test-coverage point needs resolving before merge in my view.
- **PR checklist:** `RELEASE.md` entry present and correctly filed under `## Bug fixes and other changes`; no project docs needed (private class); no public API or backward-compatibility impact.
- **API usage:** no issues. No Kedro-Viz impact.

kedro/framework/project/__init__.py
```diff
@@ -237,22 +238,24 @@ def set_requested(self, pipeline_names: list[str] | None) -> None:
             pipeline_names: Names of the pipelines to load selectively, or
                 ``None`` to load all registered pipelines.
         """
-        if set(self._requested_pipelines or []) != set(pipeline_names or []):
-            self._is_data_loaded = False
-            self._content = {}
-        self._requested_pipelines = (
-            list(pipeline_names) if pipeline_names is not None else None
-        )
+        with self._lock:
```
ElenaKhaustova, line 262: **Suggestion (scope / PR description):** The lock makes each individual call atomic, but the pattern that actually matters in the multi-threaded server is a multi-step sequence over module-global state that no per-call lock can make atomic:

```python
pipelines.set_requested(pipeline_names or None)   # session.py:348, service_session.py:163
context = self.load_context(...)                  # slow
combined += pipelines[name]                       # session.py:355, service_session.py:169
```

Two concurrent runs with different `pipeline_names` interleave: A sets the filter to `['a']`, B overwrites it with `['b']` and invalidates the cache, then A's `pipelines['a']` reloads under B's filter and raises `ValueError: Failed to find the pipeline named 'a'`. Because `load_context()` sits between the two calls, the window is wide. Modelling it with a 5 ms gap:

```
successes=42  KeyErrors=78   (sample: `a: KeyError 'a'`, `b: KeyError 'b'`)
```

65% of concurrent runs fail. Relatedly, `find_pipelines()` reads `pipelines._requested_pipelines` with no lock at all (line 611).

This is pre-existing on `main`, so it is explicitly **not** a regression from this PR, and properly fixing it means getting the per-request filter out of global state — well beyond this PR's scope. My only ask is that the description not imply the HTTP server is now thread-safe, and that a follow-up issue be opened under #5655. Otherwise this reads as done while the server-facing bug stays open.

tests/framework/project/test_pipeline_registry.py
```diff
@@ -76,3 +78,119 @@ def test_configure_project_should_not_raise_for_unimportable_pipelines(
...
+    threads += [threading.Thread(target=invalidator, daemon=True) for _ in range(2)]
+    for t in threads:
+        t.start()
+    for t in threads:
+        t.join(timeout=5)
+
+    assert all(
+        not t.is_alive() for t in threads
+    ), "threads did not finish — possible deadlock"
+    assert not errors, errors
```
ElenaKhaustova, line 157: **Critical:** This test passes against the un-fixed implementation, so it does not protect the fix. Reverting only `kedro/framework/project/__init__.py` to `main` while keeping these tests:

```
test_load_data_fires_exactly_once_under_concurrent_first_load  FAILED at :127  (loader called 20 times)
test_set_requested_concurrent_with_reads_does_not_raise        PASSED  (20/20 runs)
test_configure_concurrent_with_reads_does_not_raise            PASSED  (20/20 runs)
```

The reason is that the torn-read bug does not raise — it silently returns a wrong value. On the old code `len(p)` over a cleared `_content` returns `0`, and nothing here asserts on the value, only on the absence of exceptions. These two tests would keep passing if the lock were deleted tomorrow.

Adding value assertions alone will not be enough, though: I could not provoke a torn read on CPython in ~200k iterations, because the window between `_load_data()` returning and the dict operation is two adjacent bytecodes. A test that reliably fails without the lock needs a forced interleaving rather than a thread race — for example, have the mock loader signal a second thread to call `set_requested()` and block until it has completed, so the ordering is deterministic.

Worth saying that `test_load_data_fires_exactly_once_under_concurrent_first_load` is a genuinely good test and does pin the double-loading fix. One deterministic test would be worth more than these two racy ones.

tests/framework/project/test_pipeline_registry.py
```diff
@@ -76,3 +78,119 @@ def test_configure_project_should_not_raise_for_unimportable_pipelines(
...
+    threads += [threading.Thread(target=reconfigurer, daemon=True)]
+    for t in threads:
+        t.start()
+    for t in threads:
+        t.join(timeout=5)
+
+    assert all(
+        not t.is_alive() for t in threads
+    ), "threads did not finish — possible deadlock"
+    assert not errors, errors
```
ElenaKhaustova, line 223: **Critical:** Same issue as `test_set_requested_concurrent_with_reads_does_not_raise` above — this test also passes 20/20 against the un-fixed source, because `configure()` clearing the cache makes `len(p)` return `0` rather than raising, and the only assertion here is that nothing raised. See the fuller note on that test for the suggested deterministic approach.

kedro/framework/project/__init__.py
ElenaKhaustova: Correction to my own comment: I was too absolute above. Saying this describes "a re-entrancy that never happens" is wrong.

If user code inside `register_pipelines()` calls `pipelines.configure(...)` or `pipelines.set_requested(...)`, that *is* `configure()` / `set_requested()` re-entry from the same thread while `_load_data` holds the lock, and `RLock` is exactly what stops it deadlocking where a plain `Lock` would. Nothing in Kedro itself does this, which is why I dismissed it, but user registry code plausibly could — so your comment is defensible as written.

The narrower version of my point, which I'd still make: the comment documents that secondary case while omitting the one that fires on every single load, i.e. `inner` (line 180) taking the lock and then calling `_load_data()` (line 221), which takes it again. Mentioning both would be ideal. Sorry for the overstatement.

kedro/framework/project/__init__.py
ElenaKhaustova: Downgrading this from **Critical** to a **Suggestion** — the classification was mine to get right and I pitched it too high.

The two things I actually measured stand: the lock-order inversion against per-module import locks is real, and it is a regression (the same script completes on `main` and hangs on this branch). What I could *not* do is find real-world Kedro code that triggers it. It requires a module somewhere in the import graph to touch `pipelines[...]` at import time, which is unusual, and my repro was synthetic. Blocking a merge on a synthetic repro isn't fair.

So I'd suggest the cheap option rather than the redesign I floated above: **document the constraint instead of restructuring the locking.** A line in the `_ProjectPipelines` docstring noting that `_load_data()` holds the lock across `importlib.import_module` and user `register_pipelines()` code, and that module-level access to `pipelines` during import can therefore deadlock, would be enough. That keeps the exactly-once guarantee your test pins, and leaves a breadcrumb for whoever next considers widening the critical section.

Please disregard my suggestion about scoping the lock to state mutations only — that would trade this away for duplicate loading, and it isn't warranted at this likelihood.

kedro/framework/project/__init__.py
ankatiyar: In line 270 onwards, the `__iter__`,`keys`, `values`, `items` now return snapshots instead of live views

kedro/framework/project/__init__.py
ankatiyar: Added a note to the docstring (line 202)

kedro/framework/project/__init__.py
ankatiyar: Is this still valid after the changes from #5687 Under "serving mode", all pipelines are pre-loaded to avoid this situation with selective loading per run

tests/framework/project/test_pipeline_registry.py
ankatiyar: Have updated the tests

tests/framework/project/test_pipeline_registry.py
```diff
@@ -76,3 +78,173 @@ def test_configure_project_should_not_raise_for_unimportable_pipelines(
...
+    original_load_data = p._load_data
+
+    def racing_load_data():
+        result = original_load_data()
+        p.set_requested(["pipe_a"])  # invalidates the cache "concurrently"
+        return result
+
+    p._load_data = racing_load_data
+
+    assert len(p) == 2
```
merelcht, line 187: How does this verify that the content loaded is not stale? I might just be confused by the docstring. Can we maybe make the test a bit more readable and have explanations inline instead of a big block at the top?

tests/framework/project/test_pipeline_registry.py
```diff
@@ -76,3 +78,173 @@ def test_configure_project_should_not_raise_for_unimportable_pipelines(
...
+    for t in threads:
+        t.join(timeout=5)
+
+    assert all(
+        not t.is_alive() for t in threads
+    ), "threads did not finish — possible deadlock"
+    assert not errors, errors
+
+
+def test_configure_racing_with_load_return_does_not_yield_stale_content():
```
merelcht: What's the difference between this test and test_set_requested_racing_with_load_return_does_not_yield_stale_content?

merelcht (commented):
Thanks @ankatiyar, this fix makes sense to me, but there's a trade-off I think we need to be clear about. While this does solve the issue the added lock could also cause a deadlock, which might be hard to detect on a long running process. 

In `_load_data()` (`kedro/framework/project/__init__.py:228-243`) we hold `self._lock`
while calling `importlib.import_module()` and while running the user's
`register_pipelines()`:

```python
with self._lock:
    register_pipelines = self._get_pipelines_registry_callable(...)  # imports a module
    self._content = register_pipelines()                            # runs user code
```

That means we hold our lock while blocking on a lock we don't control: CPython's
per-module import lock. Two threads can then deadlock.

Concrete case — a project module that touches pipelines at import time, which people
do write:

```python
# my_project/helpers.py  (imported by pipeline_registry.py)
from kedro.framework.project import pipelines
DEFAULT_NODES = pipelines["__default__"].nodes   # executes during import
```

I think our options are:
A. Accept it as-is. If we go this way, I'd suggest saying so explicitly in the docstring ("accessing pipelines at module import time is not supported") so it reads as a contract rather than a known bug, and note in the release entry that the fix is scoped to state races.

B. Narrow the critical section (single-flight). Use _lock only to elect a loader, then run the import and register_pipelines() with the lock released; other threads wait on a threading.Event instead of on _lock. That removes _lock from the cycle entirely — while the import runs, _lock is free. The catch is error handling: if `register_pipelines()` raises, the event must be set and the loader slot cleared in a finally, or waiters hang forever and we've reintroduced the bug in a new form. This is a lot less simple than the current solution, hence putting it up for discussion rather than just asking for it.

I lean toward A for this PR plus B as a follow-up if we see it in practice and users run into issues happy to go with whatever you prefer.

ankatiyar:
Hey @merelcht, apologies for the delayed response!
I've made the following changes in https://github.com/kedro-org/kedro/pull/5695/changes/1574ecf1046bc6ace4929319160fc8c89fff9b7a :
- Simplified the tests
- Expanded the note in the docstring to reflect the constraint. 


I'll make a follow up issue for the second option that you suggested so if users come across the deadlock situation, they can indicate that on the issue. 
I believe importing `pipelines` would be unusual in the helper files. Usually it's the other way round, helper functions will be imported in the pipelines. I can follow up with a review of documentation and see where we can add this note as well.

merelcht (approved):
Thanks @ankatiyar! 👍 Agree with the next steps here.

> I believe importing pipelines would be unusual in the helper files. Usually it's the other way round, helper functions will be imported in the pipelines. I can follow up with a review of documentation and see where we can add this note as well.

I think this is a fair assumption, but it's not impossible that people do it and I think this is also where we have a bit of a blind spot. In any case if this is going to cause issues for people, they'll likely report it and we'll know better how they import pipelines 🙂

tests/framework/project/test_pipeline_registry.py
```diff
@@ -25,15 +27,12 @@ def register_pipelines():
...
-):
-    # Reimport `pipelines` from `kedro.framework.project` to ensure that
-    # it was not set by a prior call to the `configure_project` function.
-    del sys.modules["kedro.framework.project"]
-    from kedro.framework.project import pipelines
-
-    assert pipelines == {}
+def test_pipelines_without_configure_project_is_empty():
+    # A freshly created, unconfigured `_ProjectPipelines` is empty. Using a new
+    # instance (rather than the `kedro.framework.project.pipelines` singleton)
```
ankatiyar, line 32: This change because CI was failing https://github.com/kedro-org/kedro/actions/runs/32125473788/job/95674958933
`test_pipelines_without_configure_project_is_empty` used `del sys.modules["kedro.framework.project"]` + reimport to get a "fresh" module, but never restored it — leaving an orphaned copy of the module in `sys.modules` with its own unconfigured `PACKAGE_NAME = None`. If anything later did a fresh `from kedro.framework.project import find_pipelines` in the same process, it could bind to that orphaned copy instead of the one `configure_project()` was actually called on, causing spurious `RuntimeError`s in unrelated tests (e.g. `test_registry.py::test_list_registered_pipelines`) depending on xdist scheduling.

Fixed by asserting on a plain `_ProjectPipelines()` instance instead — no module reloading, no global side effects.

SajidAlamQB (approved):
Thanks @ankatiyar I've had a look over it and makes sense to me!

PR #5673, opened 2026-07-16
Fix generated Sphinx docs build

Shizoqua:
## Description
Fixes #5592 by making docs generated with `kedro new --tools=docs` build API docs without requiring users to manually discover or run `sphinx-apidoc`.

## Development notes
- Added generated project `docs/Makefile` and `docs/make.bat`.
- Updated the generated docs `html` target to run `sphinx-apidoc` before `sphinx-build`, which creates `docs/source/modules.rst`.
- Kept the existing `modules` toctree reference in `docs/source/index.rst`.
- Updated generated Sphinx `conf.py` to include the project `src` path for autodoc imports.
- Added starter template test coverage and updated the expected docs-tool file count.

Tested with:
- `python -m pytest --no-cov tests/framework/cli/starters -q`  
  `299 passed`
- Generated a project with `kedro new --tools docs --example no`
- Ran generated `docs/make.bat html`
- Confirmed `docs/source/modules.rst` and `docs/build/html/index.html` were created.

## Developer Certificate of Origin
All commits are signed off with `Signed-off-by`.

## Checklist

- [ ] Read the [[contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md)](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [x] Signed off each commit with a [[Developer Certificate of Origin (DCO)](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository)](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-the-commit-signoff-policy-for-your-repository)
- [ ] Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ] Updated the documentation to reflect the code changes
- [ ] Added a description of this change in the [`[RELEASE.md](https://github.com/kedro-org/kedro/blob/main/RELEASE.md)`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [x] Added tests to cover my changes
- [x] Checked if this change will affect Kedro-Viz, and if so, communicated that with the Viz team

PR #5701, opened 2026-07-29
fix: ship Makefile + make.bat with docs tool and generate modules via apidoc

### LINKED ISSUE #5592: Fix the `docs` tool so generated Sphinx docs build correctly
## Problem

When a user selects the `docs` tool (`kedro new --tools=docs`), the generated `docs/source/index.rst` points to a `modules` page that is never created, nothing runs `sphinx-apidoc`, and no `Makefile` is shipped. Building the docs warns about the missing document and the API docs come out empty. The user has to discover and run `sphinx-apidoc` by hand, with nothing telling them to.

## Why

The old `kedro build-docs` command used to run `sphinx-apidoc` for users. It was removed in 0.19.0, but the Sphinx scaffold it drove was left behind unfinished, so the `modules` reference has been broken ever since.

## Proposed solution

- Ship a `docs/Makefile` (+ `make.bat`) whose `html` target runs `sphinx-apidoc` then `sphinx-build`, so the `modules` page is generated automatically and the toctree reference always resolves. This restores the convenience the removed `kedro build-docs` command used to provide, as standard Sphinx, no new CLI command.
- Keep the `modules` reference in `index.rst` (it now resolves once apidoc runs). Optionally add a one-line comment pointing users at `make html`.

## References

- `kedro build-docs` removed in 0.19.0 (`RELEASE.md`, PR #3065)

tungyhardDevOps:
Fixes #5592.

When kedro new --tools=docs is selected, the generated docs/source/index.rst references a modules page that is never created because sphinx-apidoc is never run. This PR ships a docs/Makefile and docs/make.bat whose html target runs sphinx-apidoc then sphinx-build, so the modules reference always resolves.

### Changes
- Added docs/Makefile with html target running sphinx-apidoc then sphinx-build
- Added docs/make.bat for Windows parity
- Added a one-line comment in docs/source/index.rst pointing users at make html
- Updated 	ests/framework/cli/starters/utils.py to expect 4 docs files instead of 2

### Testing
The existing starter tests (_get_expected_files) now account for the two new files.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -6,6 +6,7 @@
 * Fixed a `RecursionError` when initialising a session with dynaconf-backed settings by converting `settings.SESSION_STORE_ARGS` to a plain `dict` before deepcopying it.
 * Excluded `kedro_benchmarks` from the built wheel so benchmark tests are no longer shipped with the package.
 * Fixed `get_close_matches` returning duplicate suggestions when several inputs matched the same target, and being able to exhaust a one-shot iterable passed as `targets`.
+* Fixed the `docs` tool so that the shipped `Makefile` and `make.bat` run `sphinx-apidoc` before `sphinx-build`, ensuring the `modules` page is generated when building HTML documentation.
 
 ## Documentation changes
 ## Community contributions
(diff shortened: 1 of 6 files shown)


ElenaKhaustova:
Duplication of https://github.com/kedro-org/kedro/pull/5673

PR #5710, opened 2026-07-30
Add a Kedro user agent to S3 config filesystems

goanpeca:
> **Status: draft, awaiting maintainer input on #5711.**
> The PR template asks that every PR link a well-described issue and that the approach be discussed first, so #5711 opens that discussion. This PR is the reference implementation so the diff is concrete.

Relates to #5711.

---

## Description

`OmegaConfigLoader` can read configuration from remote storage, and for `s3://` paths `_initialise_filesystem_and_protocol` builds the filesystem with `fsspec.filesystem(protocol="s3")`. Those requests go out with the default botocore user agent, so nothing in the request identifies Kedro as the caller. When remote config loading misbehaves against Amazon S3 or another S3-compatible object store (throttling, permissions, unexpected request volume), server-side access logs give no way to tell Kedro's traffic apart from anything else using boto3 in the same environment.

This PR appends `kedro/<version>` to the S3 user agent, and documents how to point `s3://` paths at a non-Amazon S3-compatible endpoint.

There is no GitHub issue for this yet, which is why I am opening it as a draft rather than for review. Happy to open one, or raise it on Slack first, if you would prefer to agree the approach before this goes further.

## Development notes

Two commits.

`feat: add kedro user-agent to S3 config filesystem`

- When the resolved protocol is one of `s3`, `s3a` or `s3n`, pass `config_kwargs={"user_agent_extra": f"kedro/{__version__}"}` to `fsspec.filesystem()`. `s3fs` forwards `config_kwargs` to `botocore.config.Config`, and botocore *appends* `user_agent_extra` to the user agent rather than replacing it, so the standard `Boto3/... botocore/...` string is preserved.
- The kwarg is gated on a small `_S3_PROTOCOLS` tuple so it is not passed to filesystems that do not accept `config_kwargs` (HTTP, `gcs`, `abfs` and the rest of `CLOUD_PROTOCOLS`). Their construction is unchanged.
- No new dependency, no public API change, and nothing for users to configure.

`docs: document endpoint_url for S3-compatible object stores`

- Added a note to `docs/catalog-data/data_catalog_yaml_examples.md` explaining that `s3://` paths are handled by `s3fs` and target Amazon S3 by default, and that `client_kwargs.endpoint_url` in the credentials entry points them at another S3-compatible store. It links to the MinIO example already present further down the same page, which shows the syntax.

Tests: the two existing assertions in `tests/config/test_omegaconf_config.py` that check how the S3 filesystem is constructed (`TestOmegaConfigLoader` and `TestRemotePathHandling`) now expect the `config_kwargs` argument. No new test cases were added, since the behaviour under test is the construction call itself.

No `RELEASE.md` entry is included. The user-agent suffix is not part of Kedro's API surface and there is nothing users need to do differently, so I was unsure whether it counts as user facing. Happy to add a line under "Bug fixes and other changes", or wherever you think it belongs, if you would like it recorded.

## Checklist

- [ ] Linked to a relevant GitHub issue (none exists yet, see above)
- [ ] Added a description of this change in the [`RELEASE.md`](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file (happy to add one, see above)
- [ ] Added tests to cover my changes (no new tests; the existing S3 filesystem assertions were updated to cover the new `config_kwargs`)


goanpeca:
Opened #5711 to satisfy the template's linked-issue requirement and to discuss the approach before asking for review. Keeping this as a draft in the meantime; happy to close or reshape it based on your preference.

kedro/config/omegaconf_config.py
```diff
@@ -409,7 +413,11 @@ def _initialise_filesystem_and_protocol(
         # Create and return the appropriate filesystem
         if protocol in HTTP_PROTOCOLS or protocol in CLOUD_PROTOCOLS:
             # For HTTP and cloud storage protocols, create the appropriate filesystem
-            return fsspec.filesystem(protocol=protocol), protocol
+            fs_args: dict[str, Any] = {}
+            if protocol in _S3_PROTOCOLS:
+                # Identify Kedro in the S3 user agent, appended not replaced.
+                fs_args["config_kwargs"] = {"user_agent_extra": f"kedro/{__version__}"}
```
SajidAlamQB, line 419: Passing explicit `config_kwargs` replaces whole part of any `config_kwargs` a user set via fsspec config files or `FSSPEC_S3_*` env vars.

Since OmegaConfigLoader has no storage options parameter, that's the settings users have so it would appear as unexplained auth failures on remote config loading. 

Fix is probs one line merge instead of replace.

SajidAlamQB (commented):
Thanks @goanpeca just on scope this only tags config-loading traffic.

The real S3 volume comes from kedro-datasets 

The design question for [#5711](https://github.com/kedro-org/kedro/issues/5711) is whether the team wants a shared helper that kedro-datasets adopts too.

merelcht:
Hi @goanpeca, I'm closing this because we haven't heard from you in over a month here or on the issue you opened. Feel free to-reopen if you have time to address this again.

PR #5715, opened 2026-08-03
Enforce catalog validators in DataCatalog load and save

### LINKED ISSUE #5679: DataCatalog validation funnel + settings wiring

SajidAlamQB:
## Description

Second part of dataset validation (#5679, parent #5677, design discussion #5602).

This PR adds the core validation package from #5692 into the `DataCatalog`, so a dataset can declare a `validator` in `catalog.yml` and have it enforced on every `load` and `save`. This is the first PR where validation becomes reachable from user code paths. The public API (#5680) comes in a follow up PR.

> [!NOTE]
> Stacked on #5692 the base branch is `feat/dataset-validation-core-implementation`, so this diff shows only the catalog changes. It will retarget to `main` once #5692 merges.

What it adds:

- `validator` becomes a reserved key in catalog entries (shorthand string or long form dict), captured for both explicit entries and dataset factory patterns, and stripped before dataset creation so existing datasets are unaffected
- `DataCatalog.load()` and `save()` run the declared validator, with `on` (load/save), `severity` (`error` raises, `warn` logs and passes the data through) and `skip_load_after_save` honoured; validators resolve lazily on first use and are cached after that
- Specs follow the catalog lifecycle: dataset replacement clears them, `release()` drops cached state, `to_config()` re-injects the declaration, and a `validators` property exposes resolved validators
- A `DATASET_VALIDATION` setting in `settings.py` (default `True`) propagated by the context, and a `KEDRO_DATASET_VALIDATION` environment variable that wins in both directions as a kill switch
- A pre-flight check at catalog creation that logs a warning per validator whose package or class cannot be imported, instead of failing mid-run
- asv benchmarks for catalog build time and load overhead with and without validators

## Development notes

- Modified: kedro/io/core.py, kedro/io/data_catalog.py, kedro/framework/context/context.py, kedro/framework/project/__init__.py, kedro_benchmarks/benchmark_validation.py, RELEASE.md
- New tests: tests/validation/test_catalog_validation.py (funnel, lifecycle, kill switches), plus context tests for settings propagation and pre-flight warnings in tests/framework/context/test_context.py
- All new and changed lines are at 100% coverage; ruff and mypy --strict pass

> [!IMPORTANT]

## Checklist


kedro/framework/context/context.py
```diff
@@ -319,20 +321,6 @@ def _get_parameters(self) -> dict[str, Any]:
         params_dict: dict[str, Any] = {"parameters": params}
 
         def _add_param_to_params_dict(param_name: str, param_value: Any) -> None:
```
ElenaKhaustova, line 323: **Suggestion:** The docstring for `_add_param_to_params_dict` (14 lines, including a code example) was removed here. This looks unrelated to the validation feature — the base branch still has it. Worth reverting this deletion or moving it to a separate commit so the validation PR stays focused.

kedro/framework/context/context.py
```diff
@@ -362,6 +350,51 @@ def _get_config_credentials(self) -> dict[str, Any]:
...
+            conf_catalog: The raw catalog configuration, used to detect
+                declared `validator` keys on datasets.
+        """
+        from kedro.framework.project import settings
+
+        enabled = bool(getattr(settings, "DATASET_VALIDATION", True))
+
+        if hasattr(catalog, "validation_enabled"):
+            catalog.validation_enabled = enabled
+            specs = getattr(catalog, "_validator_specs", None)
```
ElenaKhaustova: **Suggestion:** This reaches into the catalog's private `_validator_specs` via `getattr`. If the attribute is renamed, the context silently stops running preflight checks (`getattr` returns `None`). Consider exposing a public property on `DataCatalog` for the raw specs — the existing `validators` property returns the serialised config form, but `preflight_check` needs the `ValidatorSpec` objects.

kedro/io/data_catalog.py
```diff
@@ -1055,9 +1134,107 @@ def load(self, ds_name: str, version: str | None = None) -> Any:
...
+                validator=spec.class_path,
+            )
+            wrapped.__cause__ = exc
+            if spec.severity == "warn":
+                self._logger.warning(str(wrapped))
+                return data
+            raise wrapped from exc
+
+        if mode == "save":
+            self._save_validated.add(ds_name)
```
ElenaKhaustova, line 1259: **Suggestion:** `_save_validated.add()` runs inside `_maybe_validate`, before `dataset.save()` succeeds. The `discard` in the `except DatasetError` handler at L1094 compensates, but only for `DatasetError` — a non-`DatasetError` exception from a custom dataset would leave the stale entry.

Moving the bookkeeping to the `save()` method *after* `dataset.save()` succeeds would be cleaner:

```python
validated = self._maybe_validate(ds_name, data, mode="save")
try:
    dataset.save(validated)
except DatasetError as e:
    raise DatasetError(f"{ds_name}: {e}") from e
spec = self._validator_specs.get(ds_name)
if spec is not None and spec.skip_load_after_save:
    self._save_validated.add(ds_name)
```

This eliminates the `discard`, gives `_maybe_validate` a single responsibility (just validation), and closes the theoretical gap.

ElenaKhaustova:
A couple of test/benchmark gaps worth considering:

**Test coverage:**
- No `save` + `severity: warn` test — the load path is tested (`test_severity_warn_logs_and_returns_original_data`), but the save path isn't. With warn on save, unvalidated data goes to storage, which is worth an explicit test to document that behaviour.
- No `on: ["load"]` only test — `test_on_save_only_skips_load_validation` covers `on: ["save"]`, but the inverse (validate on load, skip on save) has no symmetric test.

**Benchmarks:**
- Consider adding a save-path benchmark and a cold-resolution (first-load, no cached validator) benchmark alongside the existing load benchmarks — this feature runs on every `save()` too, and the lazy resolution cost is currently hidden by the pre-warmed cache in `setup()`.

kedro/io/data_catalog.py
SajidAlamQB: Went halfway with this one. Marking after a successful save based on the spec alone would mark even when validation didn't actually run (disabled, on: [load], or the env kill switch), and only the funnel knows whether it ran. So the mark stays in the funnel, but the cleanup now catches BaseException rather than only DatasetError, so a custom dataset raising anything, or even an interrupt, clears. Added a test with a RuntimeError-raising dataset.

kedro/io/data_catalog.py
```diff
@@ -1055,9 +1150,108 @@ def load(self, ds_name: str, version: str | None = None) -> Any:
         )
 
         try:
-            return dataset.load()
+            data = dataset.load()
         except DatasetError as e:
             raise DatasetError(f"{ds_name}: {e}") from e
+        return self._maybe_validate(ds_name, data, mode="load")
+
+    def _validation_enabled_effective(self) -> bool:
```
merelcht: Nit: the `effective` part of the name confused me at first. Maybe `_is_validation_enabled()` is clearer?

kedro/io/data_catalog.py
```diff
@@ -1055,9 +1150,108 @@ def load(self, ds_name: str, version: str | None = None) -> Any:
...
+        env_value = os.environ.get("KEDRO_DATASET_VALIDATION")
+        if env_value is not None:
+            normalised = env_value.strip().lower()
+            if normalised in ("0", "false", "off"):
+                return False
+            if normalised in ("1", "true", "on"):
+                return True
+        return self.validation_enabled
+
+    def _maybe_validate(self, ds_name: str, data: Any, mode: str) -> Any:
```
merelcht: To me it makes more sense to check whether validation needs to happen before calling this method. So this method is not "maybe validate" but "validate" and the check: 

```
if not self._validation_enabled_effective():
            return data
```

happens outside of here in the save or load. I feel that makes the whole flow easier to follow.

kedro_benchmarks/benchmark_validation.py
```diff
@@ -280,3 +283,68 @@ def time_context_params_scoped_validation(self, node_count, typed_params_per_nod
         self.context._validated_params_cache = None
         self.context._cached_validation_scope = None
         self.context.params
+
+
+def _passthrough_validator(data):
```
merelcht, line 288: What is this for?

kedro/io/data_catalog.py
```diff
@@ -213,6 +217,7 @@ def __init__(
         config_resolver: CatalogConfigResolver | None = None,
         load_versions: dict[str, str] | None = None,
         save_version: str | None = None,
+        validation_enabled: bool = True,
```
merelcht, line 220: Should `validation_enabled` also be passed in `from_config()` for the catalog?

kedro/io/data_catalog.py
```diff
@@ -1055,9 +1150,108 @@ def load(self, ds_name: str, version: str | None = None) -> Any:
...
+        The `KEDRO_DATASET_VALIDATION` environment variable, when set to a
+        recognised value, wins over the catalog-level `validation_enabled`
+        flag. Recognised values are `0`/`false`/`off` (disabled) and
+        `1`/`true`/`on` (enabled). Unset or unrecognised values fall
+        back to `self.validation_enabled`.
+
+        Returns:
+            True if validation should be applied, False otherwise.
+        """
+        env_value = os.environ.get("KEDRO_DATASET_VALIDATION")
```
merelcht, line 1180: This is now read and set at every `load`/`save` operation. Does that make sense or should we store it in a global? Is there a likely scenario where the env var changes between operations?

merelcht (commented):
Thanks @SajidAlamQB, this is looking really good! I left some questions/suggestions. And one more thing I feel is worth discussing is the error handling. 

Bottom line: DataValidationError / ValidationConfigurationError should probably subclass DatasetError

Kedro currently guarantees that anything failing during a dataset load/save surfaces as DatasetError. That's enforced deliberately in `_load_wrapper` / `_save_wrapper` (kedro/io/core.py), which catch every exception type and re-raise as `DatasetError`. with a comment saying it's by design because composed datasets can throw anything. 

Validation runs in DataCatalog.load/save, outside that funnel:

```python
validated = self._maybe_validate(ds_name, data, mode="save")   # data_catalog.py:1099
try:
    dataset.save(validated)      # only this is inside the wrapper
except DatasetError as e:
```

Since DataValidationError subclasses plain Exception it doesn't get caught there.

The practical impact: a user adds `validator:` to a catalog entry and existing error handling silently stops covering a case it used to cover.

```python
try:
    df = catalog.load("cars")
except DatasetError:          # caught missing files, bad CSVs, unregistered datasets
    df = fallback()           # no longer catches schema failures
```

They changed only catalog.yml and didn't change any Python code. Same applies to plugins and hooks that wrap catalog ops in except DatasetError. 

That said, there's a fair counter-argument: a schema violation isn't an I/O failure, and merging them means except DatasetError can't distinguish "couldn't read the file" from "the data is wrong." If a separate hierarchy is the deliberate choice, that works for me too, but then could we update the Raises: docstrings on load/save and add a RELEASE.md note, so users know to add a second except clause? The thing to avoid is the contract changing with nothing documenting it.

kedro/io/data_catalog.py
SajidAlamQB: Done, the check now happens in load and save. Kept the per-dataset gating inside since that part depends on the spec.

kedro_benchmarks/benchmark_validation.py
SajidAlamQB: It's the validator the benchmark entries point at. Does no work so we measure funnel overhead, not validation. Module-level because validators resolve by dotted path. I will add to the docstring.

kedro/io/data_catalog.py
SajidAlamQB: Deliberate, reading it live means the kill switch works mid-session without rebuilding the catalog the cost is one dict lookup

SajidAlamQB:
> Bottom line: DataValidationError / ValidationConfigurationError should probably subclass DatasetError

Went with subclassing DatasetError for both. Existing except DatasetError handling keeps working, and you can catch DataValidationError first if you want to treat schema failures differently. Change is on #5692 with a tests.

SajidAlamQB:
Thanks @merelcht, all addressed ready for another review

kedro/io/data_catalog.py
merelcht: Can you explain a bit more @SajidAlamQB ? Not sure I fully get what you mean with "the kill switch" 😅

kedro/io/data_catalog.py
merelcht: Should the name still me "maybe_validate" Or is it more guaranteed now? I just feel the maybe is a bit weird inside code. When going through a flow function names can help a lot with understanding what's happening, but here it's still ambiguous.

kedro/io/data_catalog.py
SajidAlamQB: I mean the `KEDRO_DATASET_VALIDATION` env var, its used to turn validation off without touching code or config.

SajidAlamQB: For kedro run it makes no difference, it's set before the process starts and never changes. Where it helps is notebooks and kedro ipython you can do something like `os.environ["KEDRO_DATASET_VALIDATION"] = "0"` and the next `catalog.load()` skips validation

merelcht: Ah yes that makes total sense 👍 Thanks for explaining!

kedro/io/data_catalog.py
SajidAlamQB: Fair, I renamed it to _validate.

merelcht (approved):
Thanks @SajidAlamQB ! This looks really good now 👍

ankatiyar (approved):
Thanks Sajid, looks good to me!

PR #5718, opened 2026-08-04
Add node source to node snapshot

### LINKED ISSUE #5668: Add node source location to the inspection snapshot
## Description

Add per-node source-location metadata to the snapshot returned by `kedro.inspection.get_project_snapshot(...)`, so a consumer can show a node's source code without importing/running the project or holding live `Node`/function objects.

Add one field to `NodeSnapshot`:

- `source: NodeSourceSnapshot | None` — the function's source **location** (file path + line range), or `None` when it cannot be determined.

## Context

Kedro-Viz is moving its graph backend onto the inspection snapshot (kedro-org/kedro-viz#2265). The target is that Viz builds the entire flowchart from a single `get_project_snapshot(...)` call, without bootstrapping and running the project and without keeping live function objects in memory.

One thing Viz still needs the live objects for is the node **Code** panel. Today it calls `inspect.getsource` on the live function object. In a snapshot-only backend there is no live function to inspect, so the source location has to travel with the snapshot. Viz would then read the file, slice the given line range to render the panel, and fall back to no code panel when `source` is `None`.

## Proposed API

```python
@dataclass
class NodeSourceSnapshot:
    """Source location metadata for a pipeline node's underlying function."""

    filepath: str        # project-relative path (only populated when the file is inside the project root)
    line_start: int      # 1-based line number of the first line of the definition
    line_end: int        # 1-based line number of the last line of the definition
```

Add to `NodeSnapshot`:

```python
    source: NodeSourceSnapshot | None = None   # None when the location cannot be resolved
```

## Behaviour

- `source` is `None` when the location cannot be resolved: lambdas, `functools.partial`, dynamically generated functions, built-ins, functions defined **outside the project root**, or files that cannot be read. Consumers must handle `None` (for example, render the node with no code panel).
- When `source` is present, `filepath` is always project-relative to the project root passed to `get_project_snapshot(...)`.
- `line_start` and `line_end` are 1-based and inclusive.

ravi-kumar-pilla:
## Description

Closes [#5668](https://github.com/kedro-org/kedro/issues/5668).

Adds per-node source location metadata to the inspection snapshot so consumers (e.g. Kedro-Viz) can render a node's Code panel from `get_project_snapshot(...)` without holding live function objects.

Each `NodeSnapshot` now includes `source: NodeSourceSnapshot | None` with:
- `filepath` : project-relative POSIX path when the source file is inside the project root
- `line_start` and `line_end` : 1-based inclusive line range of the function definition

`source` is `None` when the location cannot be resolved (e.g. `functools.partial`, lambdas, built-ins, unreadable source, or files outside the project root).

## Development notes

- Added `NodeSourceSnapshot` and `source` field on `NodeSnapshot` in `kedro/inspection/models.py`
- Added `_resolve_node_source()` in `kedro/inspection/snapshot.py`, wired through `_node_to_snapshot()` and `_build_pipeline_snapshots()`
- Uses `Path.is_relative_to()` for project containment and `.as_posix()` for cross-platform path strings
- Line range derived from `inspect.getsourcelines()`
- Tests added/updated in `tests/inspection/test_node_pipeline_snapshot.py` covering edge cases, project-relative paths, out-of-project functions, and line range alignment with `inspect`

## Checklist


tests/inspection/test_node_pipeline_snapshot.py
```diff
@@ -37,11 +45,54 @@ def namespaced_node():
...
+class TestNodeSourceSnapshot:
+    def test_instantiation(self):
+        source = NodeSourceSnapshot(
+            filepath="src/pkg/pipelines/nodes.py",
+            line_start=10,
+            line_end=15,
+        )
+        assert source.filepath == "src/pkg/pipelines/nodes.py"
+        assert source.line_start == 10
+        assert source.line_end == 15
```
merelcht: This test doesn't seem very useful since it's basically just testing that python class instantiation works.

kedro/inspection/snapshot.py
```diff
@@ -68,11 +71,54 @@ def _build_dataset_snapshots(
...
+    files outside the project root).
+    """
+    if (
+        isinstance(func, partial)
+        or inspect.isbuiltin(func)
+        or getattr(func, "__name__", None) == "<lambda>"
+    ):
+        return None
+
+    source_file = inspect.getsourcefile(func)
```
merelcht: Shouldn't this also go inside a try/except like below?

tests/inspection/test_node_pipeline_snapshot.py
```diff
@@ -61,27 +113,103 @@ def test_func_name_is_keyword_only(self):
...
+
+    def test_returns_none_when_source_file_cannot_be_resolved(
+        self, mocker, project_path
+    ):
+        mocker.patch(
+            "kedro.inspection.snapshot.inspect.getsourcefile", return_value=None
+        )
+        assert _resolve_node_source(_identity, project_path) is None
+
+    def test_returns_none_when_source_is_unreadable(self, mocker, project_path):
```
merelcht: Isn't this test pretty much the same as `test_returns_none_when_getsourcelines_fails` above?

kedro/inspection/snapshot.py
```diff
@@ -68,11 +71,54 @@ def _build_dataset_snapshots(
...
+    if source_file is None:
+        return None
+
+    resolved_path = Path(source_file).resolve()
+    resolved_project_path = project_path.resolve()
```
merelcht: Should this be called for every node? Or can we do resolution once per project and then just pass the resolved path here?

tests/inspection/test_node_pipeline_snapshot.py
merelcht: Instead, we should add a test to cover the  `TypeError` branch.

merelcht (commented):
Thanks @ravi-kumar-pilla ! Added my comments. Are there any docs that need to be updated too to add the source field?

kedro/inspection/snapshot.py
```diff
@@ -68,11 +71,54 @@ def _build_dataset_snapshots(
...
+    if not resolved_path.is_relative_to(resolved_project_path):
+        return None
+
+    filepath = resolved_path.relative_to(resolved_project_path).as_posix()
+
+    try:
+        source_lines, line_start = inspect.getsourcelines(func)
```
jitu5, line 113: Could we unwrap `func` once and use the same target for both the filepath and line lookup? `getsourcefile()` currently sees the wrapper, while `getsourcelines()` unwraps to the original function. If the decorator is defined in another file, this can pair one file with line numbers from another.

jitu5:
## Kedro PR Review

### Overview
- **Critical:** 2 | **Suggestions:** 2

- [x] `kedro/inspection/snapshot.py:L103` — Use the same resolved function for both the filepath and line range.
- [x] `RELEASE.md` — Add an Upcoming Release entry for the new node source metadata.

- [x] `kedro/inspection/snapshot.py:L97` — Align the outside-project behavior with #5668.
- [x] `kedro/inspection/models.py:L87` — Test serialization of populated source metadata through `/snapshot`.

### Notes
- The new model is additive and backward-compatible.
- The existing comments about exception handling, duplicate tests, and resolving the project path once are valid, so I have not repeated them.
- Kedro-Viz should be included when deciding the decorated-function and filepath contracts.

ravi-kumar-pilla:
> Thanks @ravi-kumar-pilla ! Added my comments. Are there any docs that need to be updated too to add the source field?

Thank you for the review. Yes I will update the docs too.

kedro/inspection/snapshot.py
ravi-kumar-pilla: Wrapped `inspect.getsourcefile()` in try/except (OSError, TypeError). Thank you

kedro/inspection/snapshot.py
ravi-kumar-pilla: Good catch !

merelcht (commented):
Thanks for addressing the earlier feedback @ravi-kumar-pilla ! Two things I noticed:

1. https://docs.kedro.org/en/1.5.0/extend/serving/#get-snapshot example is now stale. The `GET /snapshot` example response (~line 82-89) doesn't include the new `source` key on the node object. 

2. `test_returns_none_when_getsourcefile_raises` and `test_returns_none_when_getsourcelines_raises_type_error` cover the exception branches via mocks, but no test constructs a real callable-class instance (`class Foo: def __call__(self, x): ...`), which is the concrete, supported `Node` pattern that will trigger the `TypeError` for which we had not test before.

Everything else looks solid.

kedro/inspection/snapshot.py
```diff
@@ -68,11 +71,63 @@ def _build_dataset_snapshots(
...
+
+    Returns ``None`` when the location cannot be determined (for example,
+    lambdas, ``functools.partial``, built-ins, unreadable source files, or
+    files outside the project root).
+    """
+    if (
+        isinstance(func, partial)
+        or inspect.isbuiltin(func)
+        or getattr(func, "__name__", None) == "<lambda>"
+    ):
```
ElenaKhaustova, line 93: **Suggestion:** The `Returns` paragraph here doesn't follow the Google-style `Returns:` section used by every other function in this file. Minor consistency nit (private function, so no API-doc impact):

```python
    Returns:
        Source location metadata, or ``None`` when the location cannot be
        determined (for example, lambdas, ``functools.partial``, built-ins,
        unreadable source files, or files outside the project root).
```

ElenaKhaustova (approved):
Thank you, @ravi-kumar-pilla!

jitu5 (approved):

merelcht (approved):
Thanks @ravi-kumar-pilla ! LGTM 👍

Issue #5652, opened 2026-07-06
Design a unified allowlisting mechanism for Kedro components


ElenaKhaustova:
## Goal

Investigate and document a consistent approach to allowlisting external sources for config and code loading across Kedro, rather than solving it component-by-component.

## Context

Two separate efforts have tackled the same underlying problem independently:

- **`OmegaConfigLoader`** ([#5636](https://github.com/kedro-org/kedro/issues/5636)) — proposes `allowed_schemes` / `allowed_hosts` constructor params to restrict remote `conf_source` URLs.
- **HTTP Server runner validation** ([#5568](https://github.com/kedro-org/kedro/pull/5568), re-added in [#5576](https://github.com/kedro-org/kedro/pull/5576)) — validates runner modules against `settings.RUNNER_MODULES_WHITELIST` before importing. This is already live on `main`.

Both solve "restrict what Kedro will load from external sources at runtime" but with completely different APIs and mechanisms. The runner validation uses a `settings.py` entry (`RUNNER_MODULES_WHITELIST`), while the config loader proposal uses constructor parameters. The main concern is that we don't want to end up with N different allowlist patterns across N components.

## Scope

This spike covers allowlisting for **config sources and module/code loading**. These share the same threat model: an attacker who controls what Kedro loads can alter pipeline behavior or achieve code execution.

**Catalog/dataset allowlisting is explicitly out of scope.** Reasons:

- The catalog is defined in config. If config sources are locked down, an attacker cannot inject malicious dataset entries.
- The SSRF/RCE chain from the security report requires config control as the first step — blocking that kills the rest of the chain.
- Data source restriction is a platform/infrastructure concern best handled by IAM, credential scoping, and network segmentation rather than framework-level validation.
- Edge cases (e.g. multi-tenant platforms accepting user-submitted catalog definitions) are the application's responsibility, not Kedro's.

## Questions to answer

1. **What components need this?** Audit config and code loading paths — config sources, runner modules, plugin modules, hooks, templates, etc.
2. **Centralised vs per-component?** One global config (e.g. a `settings.py` entry) vs individual constructor params vs a hybrid.
3. **What granularity?** Scheme-only, scheme+host, URL prefix, module path prefix?
4. **Where does the config live?** `settings.py`, environment variable, dedicated security config, or constructor parameters?
5. **Should code loading and config loading share one mechanism or just a consistent API pattern?** They have the same threat model but different shapes (URLs vs module paths).
[...]

PR #5726, opened 2026-08-10
Validate Dataset modules for the Kedro HTTP server

### LINKED ISSUE #5706: (security) Security Vulnerability
A detailed security report has been sent separately by email, per SECURITY.md.

lrcouto:
## Description

Addresses a possible RCE vulnerability on catalog `type` resolution through non-trusted `runtime_params` on the HTTP server.

## Development notes

This PR was reworked after initial review. The first version added a `dataset_modules_whitelist` allowlist to `OmegaConfigLoader` based on https://github.com/kedro-org/kedro/pull/5568. We decided to not use an allowlist for this specific issue (an unified allowlist mechanism for Kedro is being currently discussed here: https://github.com/kedro-org/kedro/issues/5652)

For this current version, `OmegaConfigLoader` gains `restrict_runtime_params_type_selection: bool = False`. When set, any catalog `type` field actually resolved via `runtime_params` (at any nesting depth) is simply rejected.

Instead, `KedroServiceSession.run()` / `.load_context()` / `._get_config_loader()` gains a `trusted_runtime_params: bool = True` parameter. `trusted_runtime_params=False` is passed by `http_server.py`'s `_execute_pipeline`, since `request.params` is where untrusted and possibly malicious input is going to come from.

## Impact to existing projects

Only projects whose catalog uses the `runtime_params:` resolver to set a dataset's `type` field, and run that catalog through the Kedro HTTP server, are affected. For example:

```yaml
source_data:
  type: "${runtime_params:dataset.type}"
  filepath: "data/01_raw/companies.csv"
```

If the catalog does not contain this pattern, no changes need to be made.

If it does and it serves it over the HTTP server, `POST /run` requests that receive `type` from `params` will now fail with `InterpolationResolutionError`,
```

## Checklist

### CODE REVIEW

**Comment by ankatiyar** [comment]
Thanks @lrcouto! I think before we add allowlisting mechanism here, we should wait to finalise the protocol re: https://github.com/kedro-org/kedro/issues/5652
In either case, I think we don't need to complicate the solution with allow listing at all. Is there a way we can check if the `runtime_params` passed by the user is an executable string (similar to the logging config bug we solved) if it goes into the `type:` field in general. Maybe we can centralise some utility functions and use them in the `OmegaConfigLoader` to check when `catalog` is being resolved?

**Comment by lrcouto** [comment]
> Thanks @lrcouto! I think before we add allowlisting mechanism here, we should wait to finalise the protocol re: #5652 In either case, I think we don't need to complicate the solution with allow listing at all. Is there a way we can check if the `runtime_params` passed by the user is an executable string (similar to the logging config bug we solved) if it goes into the `type:` field in general. Maybe we can centralise some utility functions and use them in the `OmegaConfigLoader` to check when `catalog` is being resolved?

I'm thinking on how this could be done, because any dataset type entry is "executable" as in "running the module code", as it resolves to the dataset object. But not all of those have the possibility of being vulnerable to RCE. In the example of the PickleDataset, it'll load and arbitrary python object, which can be dangerous, but that same danger is not going to happen on a CSV dataset. The idea of using a whitelist would be to add that element of human oversight to the type selection before the dataset object itself is loaded.

So I'm considering blocking dataset selection via `runtime_params` only when the `runtime_params` themselves come from an untrusted source, i.e. from a request from outside. Lemme see if I can make that work.

**Comment by lrcouto** [comment]
Pushed a different solution. A little less flexible than the allowlist, but simpler.

File: docs/configure/how_to_use_templating.md [hunk 1]
```diff
@@ -157,6 +157,12 @@ companies:
 
 If the `folder` parameter is not passed through the CLI `--params` option with `kedro run`, the default value `'data/01_raw/'` is used for the `filepath`.
 
+!!! warning "Restricted for untrusted callers: `type` with `runtime_params`"
+
+    A catalog entry's `type` field selects the dataset class Kedro instantiates. Resolving `type` from `runtime_params` (for example, `type: "${runtime_params:dataset.type}"`) lets whoever supplies that parameter choose the class, including ones with side effects during load, such as `kedro_datasets.pickle.PickleDataset`.
+
+    This is safe for trusted, non-server use such as `kedro run --params`, since that caller already has as much control over the process as the parameter it's supplying. It is not safe when `runtime_params` comes from an untrusted caller, such as an HTTP request body — the [Kedro HTTP server](../extend/serving.md#dataset-type-security) always rejects a catalog `type` that resolves to an actual `runtime_params` value in that case, with no allowlist or setting to relax it. A `type` that falls back to a default (for example, a `globals` value) instead of an actual `runtime_params` value is unaffected either way.
```
ankatiyar: Could we make this a little less verbose - no need to mention the no allowlist or setting to relax etc

docs/extend/serving.md
```diff
@@ -211,6 +211,12 @@ Short names (for example, `SequentialRunner`) always resolve against `kedro.runn
 RUNNER_MODULES_WHITELIST = ["external_lib.runners"]
 ```
 
+#### Dataset type security
+
+A catalog entry's `type` field selects which dataset class Kedro imports and instantiates, so the HTTP server's `params` must never be able to choose it. The server always rejects a catalog `type` that resolves through the [`runtime_params` resolver](../configure/how_to_use_templating.md#how-to-override-configuration-with-runtime-parameters-with-the-omegaconfigloader) to an actual request-supplied value, at any nesting depth. There is no allowlist and no setting to relax this: any `POST /run` request whose `params` reach a catalog `type` field, however that value resolves, fails with an `InterpolationResolutionError`.
```
ankatiyar: Same here, it's quite long.

kedro/server/http_server.py
```diff
@@ -254,6 +254,10 @@ def _execute_pipeline(
             namespaces=request.namespaces,
             only_missing_outputs=request.only_missing_outputs,
             runtime_params=request.params,
+            # `request.params` comes straight from the HTTP request body, so it
+            # must never be allowed to select a catalog dataset's `type` --
+            # see kedro-org/kedro#5706.
+            trusted_runtime_params=False,
```
ankatiyar: I believe we don't need this as when http server is being used, it sets the `KedroServiceSession._serving_mode` as `True`. Maybe that can be used to trigger the restriction with `type` selection instead of adding another argument to `run()` which already has too many args

kedro/framework/session/service_session.py
```diff
@@ -196,8 +215,16 @@ def run(  # noqa: PLR0913
         namespaces: Iterable[str] | None = None,
         only_missing_outputs: bool = False,
         runtime_params: dict[str, Any] | None = None,
+        trusted_runtime_params: bool = True,
```
ankatiyar: Same comment as below - maybe we can use `serving_mode` field to determine if the restriction with config loader be applied or not instead of adding an arg here

kedro/config/omegaconf_config.py
```diff
@@ -377,13 +402,126 @@ def load_and_merge_dir_config(
...
+            else OmegaConf.to_container(
+                OmegaConf.unsafe_merge(*aggregate_config), resolve=True
+            )
         )
+
         return {
             k: v for k, v in merged_config_container.items() if not k.startswith("_")
         }
 
+    @typing.no_type_check
```
ankatiyar, line 414: Is this necessary here and on line 435?

ankatiyar (commented):
Thanks @lrcouto this solution looks better to me - left some suggestions
Also wondering if this can be done in a slightly cleaner way without adding too much code in `OmegaConfigLoader`. I don't have a clean solution yet myself, will think about it some more and get back to you about this.

kedro/config/omegaconf_config.py
lrcouto: Mypy complains about the big Union that returns from `OmegaConf.to_container`. It's the same reason why this decorator is being used for `load_and_merge_dir_config`, for example.

lrcouto:
Thank you @ankatiyar ! Applied the suggested changes.

docs/extend/serving.md
```diff
@@ -211,6 +211,12 @@ Short names (for example, `SequentialRunner`) always resolve against `kedro.runn
 RUNNER_MODULES_WHITELIST = ["external_lib.runners"]
 ```
 
+#### Dataset type security
```
merelcht, line 214: Should we add an example of what's accepted in a catalog entry vs what isn't here?

kedro/config/omegaconf_config.py
```diff
@@ -129,13 +139,28 @@ def __init__(  # noqa: PLR0913
...
+        )
         self.base_env = base_env or ""
         self.default_run_env = default_run_env or ""
         self.merge_strategy = merge_strategy or {}
         self._globals_oc: DictConfig | None = None
         self._runtime_params_oc: DictConfig | None = None
+        # Variable paths (e.g. "dataset.type") that an actual runtime_params
+        # value was found for during the most recent resolution pass -- as
+        # opposed to falling back to a resolver default. Used to guard against
+        # runtime_params selecting a catalog dataset's 'type'.
```
merelcht: Can we keep this as short as possible? Ideally we wouldn't need to add long comments in the code and the doc string + argument names should be self explanatory.

kedro/config/omegaconf_config.py
```diff
@@ -377,13 +402,126 @@ def load_and_merge_dir_config(
...
+        isn't something this can determine from the string alone (a safe
+        and a dangerous dataset type are equally valid-looking dotted
+        paths), so when the flag is set, any real hit is rejected
+        unconditionally rather than checked against a list of safe values.
+
+        Checked at every nesting depth, not just each top-level catalog entry:
+        wrapper datasets such as ``CachedDataset`` and ``PartitionedDataset``
+        accept a nested dataset config (e.g. a ``dataset:`` key) with its own
+        ``type``, which reaches the same ``AbstractDataset.from_config`` /
+        ``load_obj`` path as a top-level entry.
```
merelcht: Can we make this less verbose?

kedro/framework/session/service_session.py
```diff
@@ -152,19 +152,35 @@ def close(self) -> None:
...
+
+        Args:
+            runtime_params: Extra parameters passed to a Kedro run.
+
+        In serving mode, ``runtime_params`` originates from an untrusted caller
+        (an HTTP request body), so the config loader is forced to reject a
+        catalog dataset `type` driven by `runtime_params`, regardless of what a
+        project's own `CONFIG_LOADER_ARGS` says -- that restriction must not be
+        weakenable by project config for an untrusted caller. See
+        kedro-org/kedro#5706.
```
merelcht: I'd suggest either giving the full link to the issue here or not linking at all.

merelcht (commented):
The approach looks good to me. My main suggestion is to clean up the docstring/comments a bit, because they're quite verbose.

ankatiyar (approved):
Thanks @lrcouto the method looks good to me! I agree with merel's comments about making things less verbose!

kedro/framework/session/service_session.py
lrcouto: Changed it for the full link

lrcouto:
Thank you @merelcht ! Applied the suggested changes.

merelcht (approved):
Thanks Laura, this looks good to me now! 👍

Issue #5220, opened 2025-11-23
dask documentation for local runs


bparbhu:
## Description
I'm looking for a working example of a dask integration with kedro and also possibly ray integration with kedro using dask on ray

### Documentation page (if applicable)
https://docs.kedro.org/en/0.19.14/deployment/dask.html

## Context
I'm not sure the documentation shows a fully working pipeline with dask that can work

bparbhu:
Also this would be beneficial for people looking at running dask locally on a laptop for example

JonathanLoscalzo:
Yup, looking into this, and it isn't working. 
Some mistakes I've found

1. Dask Runner isn't being working. The issue found is related of how nodes get retrieved from node_futures dictionary. Some fix could be:
```python
        for node in nodes:
            dependencies = (
                node_futures[dependency] for dependency in node_dependencies[node]
            )
            node_futures[node] = client.submit(
                DaskRunner._run_node,
                node,
                catalog,
                self._is_async,
                run_id,
                *dependencies,
            )

# HERE IS THE FIX
        node_futures_inv = {v.key: k for k, v in node_futures.items()}
        for i, (node_future, node_result) in enumerate(
            as_completed(node_futures.values(), with_results=True)
        ):
            node = node_futures_inv[node_future.key]
            self._logger.info("Completed node: %s", node.name)
            self._logger.info("Completed %d out of %d tasks", i + 1, len(nodes))
```

2. Another problem is, the data it isn't being loaded correctly
```
Expected 2D array, got scalar array instead:
array=<object object at 0x132546060>.
Reshape your data either using array.reshape(-1, 1) if your data has a single feature or array.reshape(1, -1)
```
This is happening when the pipeline is being run with spaceflights examples, something related to the catalog I guess
`regressor.fit(X_train, y_train)`

For some reason, create_default_dataset isn't being called (IDK if it is related).

3. dependencies are not being used by DaskRunner  `_run_node` method
4. What is it the DaskDataset's purpose? 

It will be nice to have the same example with spaceflights, but working for dask.

JonathanLoscalzo:
I've followed the example from DaskRunner, using spaceflight pipelines tutorial. 
`kedro run --runner spaceflights.runner.DaskRunner --pipeline data_science`

lrcouto:
Thanks for opening the issue. It seems like the documentation itself needs to be reviewed.

lrcouto:
[...]

PR #5732, opened 2026-08-14
Fix DaskRunner example in Dask deployment docs

### LINKED ISSUE #5233: Code examples on Deploy/Dask not working
## Description
Community members have reported that the code examples on the Deploy/Dask page are not working out of the box. They are pretty old, and might be in need of a revision.

Issue where the problem was reported: https://github.com/kedro-org/kedro/issues/5220
The docs page in question: https://docs.kedro.org/en/stable/deploy/supported-platforms/dask/

kingjaiteh:
## Description

Fixes #5233. The `DaskRunner` example on `docs/deploy/supported-platforms/dask.md`
does not run: copied verbatim into a fresh project, it fails with
`AttributeError: 'dict' object has no attribute 'name'`. This also covers the
symptom reported in #5220.

The page was written against the pre-1.0.0 runner API and never updated. This PR
fixes only the two defects that stop the example running. I have left the rest
out deliberately, see below.

## Development notes

Two problems in `_run_node`, both from Kedro 1.0.0 breaking changes.

**1. Wrong positional arguments to `run()`.** The page had:

```python
runner = SequentialRunner()
return runner.run(Pipeline([node]), catalog, hook_manager, is_async, run_id)
```

Bound against the current
`run(pipeline, catalog, hook_manager=None, run_id=None, only_missing_outputs=False)`,
`is_async` lands in `run_id` and `run_id` lands in `only_missing_outputs`, where a
non-empty string is truthy, silently putting every node into only-missing-outputs
mode. `SequentialRunner()` is also built with no arguments, so the `is_async`
intent is dropped.

**2. `run()` no longer returns the node.** Since 1.0.0 it returns all pipeline
outputs. `_run` consumes the result through `as_completed(..., with_results=True)`
and reads `node.name`, which is where the `AttributeError` comes from. The
docstring already says `_run_node` returns "the node argument", so the fix
restores the documented contract.

The change:

```python
runner = SequentialRunner(is_async=is_async)
runner.run(Pipeline([node]), catalog, hook_manager, run_id)
return node
```

### Testing

Tested against `main`, with the runner re-extracted from the edited page rather
than hand-copied, so what is tested is what the page now says.

- Kedro installed `-e` from `main`, `distributed` 2026.7.1, Python 3.13, Windows 10
- Project from `kedro new --starter=spaceflights-pandas`
- Baseline: `kedro run --pipeline=data_processing` with the default runner passes

Before: exit 1, `AttributeError: 'dict' object has no attribute 'name'`, plus a
follow-on `ValueError: No clients found` from `DaskRunner.__del__` during teardown.

After: exit 0, all three nodes log by name

```
Completed node: preprocess_companies_node
Completed 1 out of 3 tasks
Completed node: preprocess_shuttles_node
Completed 2 out of 3 tasks
Completed node: create_model_input_table_node
Completed 3 out of 3 tasks
```

and `data/03_primary/model_input_table.parquet` is written. No teardown error.

No unit tests added: the change is to a code sample in a Markdown page, and there
is no harness that executes the deployment doc snippets.

### Deliberately not fixed here

Two further problems on the page, discussed in
https://github.com/kedro-org/kedro/issues/5233#issuecomment-5235082825:

- `create_default_dataset`, `run_only_missing`, `DataCatalog.list` and
  `DataCatalog.shallow_copy` were all removed in 1.0.0 but are still used by the
  page. Nothing reaches them in a run where every dataset is registered, which is
  why the example gets as far as it does. Removing `create_default_dataset`
  leaves `_DaskDataset` unwired, so the page's central idea needs a design answer
  before it can be rewritten honestly.
- The "Update CLI implementation" section asks the reader to hand-write a `cli.py`
  around `_instantiate_runner()`, which no longer exists. `--runner-params` looks
  like the intended replacement, and the section can probably be deleted, but I
  would rather have that confirmed than guess.

Happy to take both in follow-ups once there is a steer on the first one.

## Checklist

- [x] Signed off each commit with a DCO
- [x] Added a description of this change in the `RELEASE.md` file


RELEASE.md
```diff
@@ -14,6 +14,7 @@
 
 ## Documentation changes
 * Added a new "Vibe coding with skills" page documenting the `kedro-skills` plugin.
+* Fixed the `DaskRunner` example on the Dask deployment page, which passed `is_async` and `run_id` as the wrong positional arguments to `runner.run()` and returned its outputs instead of the node, so the documented example failed with `AttributeError: 'dict' object has no attribute 'name'`.
```
deepyaman:
```suggestion
* Fixed the `DaskRunner` example on the Dask deployment page.
```

deepyaman (approved):
Appreciate if you can shorten the release note (it's not important enough to most users warrant a whole explanation; noting that it's fixed is sufficient IMO). Otherwise LGTM!

SajidAlamQB (approved):
LGTM, thank you @kingjaiteh

Issue #4690, opened 2025-04-25
Prevent serialization/pickling errors when using cloud-based datasets (e.g., S3).


SajidAlamQB:
### Description

Related to: #2162 

Datasets backed by cloud storage (e.g., S3 via s3fs) often hold non-picklable internal references (e.g., `S3FileSystem._glob`), which cause `ParallelRunner` to fail during multiprocessing pickling.

Potential Fix:

- Add a validation layer in `ParallelRunner._validate_catalog()` that warns users about cloud datasets and suggests using `ThreadRunner`. Detect datasets with `_protocol` in ("s3", "gcs", "adl", etc.) and raise a warning with suggested alternatives (e.g., `ThreadRunner`).

- Update documentation with guidance and caveats for cloud-based storage and multiprocessing, explain why cloud datasets are will cause issues with `ParallelRunner`.

astrojuanlu:
Rather than looking at this from the Runners perspective, can we maybe do so from the Datasets perspective? https://github.com/kedro-org/kedro/issues/1936 

According to the original issue https://github.com/kedro-org/kedro/issues/2162 this has to do with fsspec. Custom datasets can of course introduce non-pickleable attributes, but our official ones are unpickleable because of fsspec, so it's something we can address.

PR #5734, opened 2026-08-18
Add warning on ParallelRunner about serialization/pickling errors

lrcouto:
https://github.com/kedro-org/kedro/issues/4690

Add warning on ParallelRunner about serialization/pickling errors when using cloud-based datasets

## What it does

Datasets backed by remote/cloud storage using `fsspec` frequently hold non-picklable internal states, like a cached filesystem client for example, and these do not play nice with `ParallelRunner`. It multiprocesses by running separate OS processes that don't share memory. Because these processes do not share memories, a `pipe` is needed to to communicate and share bytes between those processes. For this to happen, objects need to be serialized/pickled (`multiprocessing.reduction.ForkingPickler`), so no pickling, no multiprocessing.

This PR adds a warning to `ParallelRunner` in case one of these cloud protocols (s3, gcs, hdfs, etc) is being used, suggesting that they choose `ThreadRunner` instead. It uses threading instead of forking processes, so it doesn't have this same limitation.

## Why not change the Datasets themselves?

It would be possible to change the `self._fs = fsspec.filesystem(...)` line on the "offending" datasets with a different pattern that avoids this issue. The problem is that this change would have to be done on around 40 different datasets, which would have to be either one massive PR that alters a lot of files, or many separate PRs that alter few files each, and would widen the scope of the original issue. Something that we might want to address in the future.

## Checklist


merelcht (approved):
Thanks @lrcouto, this looks good to me 👍 

One thing I was wondering is if we could also suggest using `SequentialRunner`? Or is the assumption the user wants some sort of optimisation and that's why they picked `ParallelRunner` in the first place?

lrcouto:
> Thanks @lrcouto, this looks good to me 👍
> 
> One thing I was wondering is if we could also suggest using `SequentialRunner`? Or is the assumption the user wants some sort of optimisation and that's why they picked `ParallelRunner` in the first place?

I was working under the assumption that if user picks ParallelRunner it's because they want some sort of optimization, since SequentialRunner is the default option. But I don't see an issue in adding it to the suggestions.

kedro/io/data_catalog.py
```diff
@@ -38,6 +39,25 @@
...
+    "s3n",
+    "gcs",
+    "gs",
+    "adl",
+    "abfs",
+    "abfss",
+    "az",
+    "hdfs",
+    "oci",
+}
```
ravi-kumar-pilla: Hi @lrcouto ,

This looks fine but it will add to the maintenance. I would recommend having the other way i.e., 

Replace `_CLOUD_PROTOCOLS` with:

```python
protocol = getattr(dataset, "_protocol", None)
if protocol and protocol != "file":
    warnings.warn(...)
```

Kedro already maps local paths to `"file"` in `_parse_filepath` (`kedro/utils.py`). The warning only runs after pickling fails, so this is a targeted hint and not a scan of all cloud datasets.

No list to maintain, new fsspec backends (`lakefs`, `dbfs`, `hf`, …) are covered automatically.

lrcouto: That's a good point. I'll rework it a bit.

lrcouto:
@ravi-kumar-pilla thanks for the suggestion! Applied it now.

ravi-kumar-pilla (approved):
Thank you @lrcouto

PR #5735, opened 2026-08-18
Redact credentials from filepath for HTTP server

ankatiyar:
## Description

Closes a credential-leakage gap in the HTTP server layer: dataset filepaths and
third-party exception messages can carry presigned-URL signatures, tokens, and
`user:pass@` authority credentials. In the following places:
-  the `/snapshot` and
- `/run` endpoints (and dataset `repr()` / `DatasetError` messages) could surface
these values verbatim to clients and server logs.

## Development notes

- Introduced a shared `_redact_url_credentials` helper in `kedro/io/core.py`
  that parses URLs structurally with `urllib.parse.urlsplit` and redacts:
  - authority credentials (`user:pass@` → `<redacted>@`)
  - query strings (`?…` → `?<redacted>`)
  - fragments (`#…` → `#<redacted>`)
  
  Scheme, host, port, and path are preserved. Structural parsing means the
  helper does not depend on a curated list of "sensitive" query parameter
  names (e.g. `sig`, `token`, `X-Amz-Signature`) and works for any provider.

- Handles three input shapes:
  - a bare URL (`https://user:pass@host/f?sig=X`)
  - a scheme-less filepath that still carries `?`/`#` (Kedro strips
    `s3://`-style schemes before storing filepaths, so a query/fragment can
    appear without a scheme prefix)
  - free-form text with an embedded URL (e.g. a third-party exception
    message). 
  
  Non-string values and plain local paths are returned unchanged.

- Applied the helper at every known leak site:
  - `AbstractDataset._pretty_repr` → wraps each `_describe()` value via a new
    `_redact_repr_value` walker that recurses through `dict`, plain `list` /
    `tuple`, and `PurePath` (namedtuple subclasses like `Version` are left
    alone).
  - `AbstractVersionedDataset._get_save_path` and `.save()` error messages,
    which embed `self._filepath` directly.
  - `kedro/server/http_server.py` `/snapshot` and `/run` handlers: exception
    messages are redacted before being logged and before being returned in
    `ErrorDetail.message`.
  - `kedro/inspection/models.py` `DatasetSnapshot.from_config` now delegates
    to the shared helper (replacing a narrower local regex that only handled
    `user:pass@` in scheme-based URLs).

- Removed `exc_info=True` from the two HTTP-server `logger.error` calls
  because the traceback text embeds the original unredacted `str(exc)`.

## Checklist


kedro/server/http_server.py
```diff
@@ -140,12 +140,16 @@ def get_snapshot() -> SnapshotResponse:
                 parameters=snapshot.parameters,
             )
         except Exception as exc:
-            logger.error("Snapshot request failed: %s", str(exc), exc_info=True)
```
Huongg, line 143: if you removed the `exc_info=True`, does it make the debugging harder? Maybe we can redact the traceback content?

Huongg:
## Kedro PR Review

> Generated with review-kedro-pr skill.

- General review: Well-scoped and focused. Helper is cleanly factored and well-commented, docstrings are Google-style, and the list/tuple recursion guard that excludes namedtuples like Version is a nice touch.
- PR checklist: RELEASE.md entry present and in the correct ## Bug fixes and other changes section. No docs change needed (internal behavior). New tests reference existing fixtures/classes correctly.
- API usage: Correct — no base-class contracts touched; _pretty_repr / __repr__ / wrapper patterns used as intended.
- Scope check: All commits belong to this branch; no cross-PR contamination. .lycheeignore additions are test-support only.

---
Summary

- Critical: 1

---

## 🔴 Critical

### `kedro/io/core.py:75` — `user:pass@` credentials leak for HTTP(S) datasets (redaction misses scheme-stripped userinfo)

`_redact_single_url` detects authority credentials with `has_userinfo = "@" in parts.netloc`. But `urlsplit` only populates `netloc` when the string contains `//`. For HTTP(S) datasets, Kedro stores the filepath **with the scheme stripped but userinfo + query retained**, so the value reaching the redactor has no `//` — `urlsplit` puts the authority into `path`, leaves `netloc` empty, and `user:pass@` is never redacted.

Reproduced end-to-end with Kedro's own functions:

```python
>>> from kedro.io.core import get_protocol_and_path, _redact_url_credentials
>>> _, path = get_protocol_and_path("https://user:pass@host/file.csv?sig=SECRET")
>>> path
'user:pass@host/file.csv?sig=SECRET'          # scheme stripped, creds kept
>>> _redact_url_credentials(path)
'user:pass@host/file.csv?<redacted>'          # query redacted, user:pass@ LEAKS
```

Contrast with the scheme-carrying form, which is handled correctly: https://<redacted>@host/file.csv?<redacted>.

This is squarely in the PR's threat model: an HTTP(S) dataset using basic-auth credentials will leak them through the exact channels this PR fixes — dataset repr() → DatasetError messages → /run responses and server logs. The second-pass `_redact_url_credentials(str(exc))` in http_server.py also misses it, because the full exception message contains spaces and so takes neither the "://" branch nor the " " not in value branch. It likewise affects the versioned save() / _get_save_path error messages, which redact the scheme-less str(self._filepath).

`DatasetSnapshot.from_config` is not affected — it redacts the raw catalog config value, which still carries the scheme.

**Suggested fix**
 in the scheme-less branch of _redact_url_credentials, detect an @ in the authority position (the segment before the first /, ?, or #) and, when present, prefix // before parsing so urlsplit recognises the authority — then strip the // back off. Guard on the delimiter so a local filename that merely contains @ (e.g. report@2024.csv) is left untouched. Please add tests for the scheme-stripped HTTP forms (user:pass@host/file.csv?sig=x and userinfo-only user:pass@host/file.csv), since the current suite only exercises the scheme-carrying and query-only variants.

lrcouto:
Confirmed the same error from the bot reviewer here. A string like `"https://user:password@host/file.csv?sig=SECRET"` will end up as `"user:password@host/file.csv?<redacted>"`. The secret is redacted but the credentials are not.

Can be tested like:

```py
    https_uri = "https://user:password@host/file.csv?sig=SECRET"  # pragma: allowlist secret
    http_protocol, http_path = get_protocol_and_path(https_uri)
    print(f"get_protocol_and_path({https_uri!r}) -> protocol={http_protocol!r}, path={http_path!r}")
    redacted_http = _redact_url_credentials(http_path)
    print(f"_redact_url_credentials({http_path!r}) -> {redacted_http!r}")
    assert "password" in redacted_http, "expected credentials to leak"
    assert "<redacted>" in redacted_http, "expected the query string to still be masked"
```

ankatiyar:
@Huongg @lrcouto I've addressed the feedback
- Added back traceback with redacted filepath
- tightened the redaction

Huongg:
thanks @ankatiyar the change looks good to me, can you add some QA notes as a guidance on best way to test these changes please?

ankatiyar:
@Huongg 
You can check the following:
- Dataset `__repr__` is redacted
```
from kedro_datasets.pandas import CSVDataset
ds = CSVDataset(filepath="https://AKIA:SECRET@example.com/bucket/file.csv?X-Amz-Signature=deadbeef#frag")
print(repr(ds))
```

- `DatasetError` message is redacted
```
from kedro_datasets.pandas import CSVDataset
from kedro.io.core import Version
import pandas as pd

ds = CSVDataset(
    filepath="/tmp/versiontest/data.csv?X-Amz-Signature=deadbeef",
    version=Version(load=None, save="2024-01-01T00.00.00.000Z"),
)
ds.save(pd.DataFrame({"a": [1]}))
ds.save(pd.DataFrame({"a": [1]}))
```


For HTTP responses, you would need to use a Kedro project:
- `/snapshot` should redact filepaths with signature when added to `catalog`
- If `/snapshot` fails, it should redact the filepath in the error
- `DatasetError` in `/run` endpoint should also be redacted

lrcouto:
Looks solid!

The only thing I can think of if an exception that contains an URL without the `://`, but still with credentials or secrets in it, and that's part of a larger error message that contains spaces. Then it would fall into the `if " " in value` conditional and be returned whole.

TBH I don't even think this could happen outside of some extreme edge case and it's just a nitpick, so it looks good to me!

lrcouto (approved):

Huongg (approved):
thanks @ankatiyar, I have tested following the QA notes you left, confirmed the paths redact as expected, no credential is shown in reprs, error messages, or the snapshot response.

ankatiyar:
@lrcouto I'll merge this as is, might be rare that credentials are embedded in a scheme-less URL and we can fix it if it comes up. 🤔

PR #5747, opened 2026-08-28
[WIP] Pre-load OCL 

ankatiyar:
## Description

In serving mode, `KedroServiceSession.run()` currently builds a fresh `OmegaConfigLoader` on every request. Two problems fall out of that:

**1. Redundant work per request.** The loader's constructor re-parses `conf_source`, re-initialises the fsspec filesystem, re-registers OmegaConf resolvers, and re-loads `globals`. `__getitem__` then re-reads and re-parses YAML from disk for `credentials`, `catalog`, and `parameters`. `_get_catalog` then re-runs path conversion, re-loads credentials, and fires `after_catalog_created` again. Only `runtime_params` actually varies between requests — the rest is repeated work paid on every hit to the HTTP server.

**2. Concurrency hazards from OmegaConf's global resolver registry.** `OmegaConf.register_new_resolver` and `clear_resolver` mutate module-level state, so each new `OmegaConfigLoader` clobbers the previously registered `globals`, `runtime_params`, and `oc.env` resolvers. Under concurrent requests, resolvers registered by request A end up executing bound methods on request B's loader instance — so `_get_globals_value` reads B's `self._globals`, `_get_runtime_value` records hits in B's `_runtime_params_hits`, and so on. The most serious consequence is that `_guard_runtime_params_in_catalog_type` — the check that forbids `${runtime_params:...}` from selecting a dataset's `type` — can **fail-open**: it reads its own instance's `_runtime_params_hits`, but the hits were recorded on a different instance whose resolver happened to be registered at the moment. `oc.env` register/clear-inside-resolve is a similar race.

The root cause of both is treating the loader as request-scoped when the vast majority of what it does is request-independent.

## What

Split the loader's work into a **one-shot snapshot phase** that runs once on the session thread before any request threads start, and a **thin per-request phase** that only touches what actually depends on this request's `runtime_params`.

**`OmegaConfigLoader` refactor** (`kedro/config/omegaconf_config.py`). `__getitem__` was a single monolithic method doing directory discovery, YAML parsing, dedup checking, merging, resolving, and the security guard in one go. It is now a thin orchestrator over four helpers so the "read from disk" side can be reused without paying the "resolve with runtime_params" side, and vice versa:

- `_read_dir_config_files` — file discovery + YAML parse + duplicate-keys check for one directory. No merge, no resolve.
- `_merge_and_resolve` — takes pre-parsed per-file configs and produces the final dict using whatever `self.runtime_params` currently holds.
- `_read_raw_configs_for_key` — the "raw" half of `__getitem__`: base + env per-file configs plus paths and processed-files set.
- `_resolve_from_raw` — the "resolve" half: reproduces the base/env merge, missing-files check, and `UnsupportedInterpolationType`-for-globals translation. Calls the existing security guard unchanged.

`load_and_merge_dir_config` keeps its public signature (it's documented as reusable by subclasses) and now delegates to the two new helpers, so nothing in the CLI path changes.

**New `_serving_config.py`** (`kedro/framework/session/_serving_config.py`). Three pieces:

- `_ConfigSnapshot` — session-scoped, built **once**. Holds fully-resolved `credentials` and `globals`, raw per-file configs for `parameters` and `catalog`, a reference to the single persistent `OmegaConfigLoader`, and an `RLock`.
- `build_snapshot(persistent_loader)` — the preload step.
- `_ServingConfigLoader(AbstractConfigLoader)` — the per-request loader, cheap to construct.
  - Credentials and globals are lock-free deep-copied reads.
  - Parameters and catalog delegate to `persistent_loader._resolve_from_raw` under the snapshot lock, via a small `_swapped_runtime_params` context manager that sets the persistent loader's `runtime_params`, resets `_runtime_params_oc` and `_runtime_params_hits` for this request, then restores state on exit.
  - `restrict_runtime_params_type_selection` is exposed as a property mirroring the persistent loader's setting so tests / hooks that introspect it keep working.

**Session wiring** (`kedro/framework/session/service_session.py`). `_enable_serving_mode` now also calls `_preload_config`, which constructs one persistent `OmegaConfigLoader` (forced to `restrict_runtime_params_type_selection=True`) and hands it to `build_snapshot`. `_get_config_loader` returns a `_ServingConfigLoader` in serving mode, or the previous CLI `OmegaConfigLoader` otherwise. In serving mode, `runtime_params` is scoped to the request loader only; the persistent loader stays at `runtime_params=None`.

The security guard is preserved end-to-end: `_runtime_params_hits` is reset inside the snapshot lock for the current request and restored after, so concurrent requests can't cross-contaminate the hits set. `restrict_runtime_params_type_selection=True` remains forced in serving mode and cannot be weakened by project `CONFIG_LOADER_ARGS`. The CLI path is behaviourally unchanged.## Checklist


ankatiyar:
@copilot resolve the merge conflicts in this pull request

PR #5766, opened 2026-09-15
ci: validate branch input before checkout in unit-tests workflow

merelcht:
### Description

Fixes [code-scanning alerts #46 and #47](https://github.com/kedro-org/kedro/security/code-scanning/47) (CodeQL `actions/cache-poisoning/poisonable-step`).

### Context
`.github/workflows/unit-tests.yml` is a reusable workflow that accepts a `branch` string input, uses it as the `ref` for `actions/checkout`, then runs `astral-sh/setup-uv` (which caches) followed by steps that execute the checked-out code. This workflow is reachable from `nightly-build.yml`'s `schedule`/`workflow_dispatch` triggers, which get write access to the default-branch Actions cache scope — so CodeQL treats the `branch` input as attacker-controllable, even though today's only callers pass hardcoded literals (`main`/`develop`).

## Development notes
This adds a validation step before checkout that fails the job unless `inputs.branch` is empty, `main`, or `develop`, closing off the path for any future caller to smuggle an untrusted ref into this privileged, cache-writing job.

## Checklist
- [ ]  Read the [contributing](https://github.com/kedro-org/kedro/blob/main/CONTRIBUTING.md) guidelines
- [ ]  Linked to a relevant GitHub issue
- [x]  Signed off each commit with a [DCO](https://developercertificate.org/)
- [ ]  Opened this PR as a 'Draft Pull Request' if it is work-in-progress
- [ ]  Updated the documentation to reflect the code changes
- [ ]  Added a description of this change in the [RELEASE.md](https://github.com/kedro-org/kedro/blob/main/RELEASE.md) file
- [ ]  Added tests to cover my changes

Issue #47, opened 2019-07-08
SparkDataSet with a relative local file path doesn't work on jupyter notebook/lab


gotin:
## Description
SparkDataSet with a relative local file path doesn't work on jupyter notebook/lab
## Context
We can have a SparkDataSet entry which filepath has a relative local file path in catalog.yml which looks like this;

```yaml
something:
  type: kedro.contrib.io.pyspark.SparkDataSet
  file_format: parquet
  filepath: data/01_intermediate/something.parquet
  save_args:
    mode: overwrite
```

And when something.parquet is placed under <project_directory>/data/01_intermediate/something.parquet properly, kedro run successfully can load this parquet as long as the pipeline uses the 'something' data.

But on a jupyter notebook invoked by kedro jupyter notebook command, the following script doesn't load 'something' as expected.

```python
io.load('something')
```

Instead, it raises an exception looks like the following;
```python
Py4JJavaError: An error occurred while calling o25.load.
: org.apache.spark.sql.AnalysisException: Path does not exist: file:/Users/go_kojima/sample_kedro_project/notebooks/data/01_intermediate/something.parquet
```

Reading spark_data_set.py (of kedro) and readwriter.py (of pyspark), I think it is caused by spark.read.load implementation. And apparently spark.read.load tries to read the data which is located at /Users/go_kojima/sample_kedro_project/notebooks/data/01_intermediate/something.parquet mistakenly, somehow, spark.read.load tries to resolve a given relative filepath referring from the directory of the notebook.

## Steps to Reproduce
As I showed above, 
1. put some spark readable data on your local file system
2. put a corresponding data entry using a relative file path on catalog.yml
3. invoke jupyter notebook by using kedro jupyter notebook command at the root directory of the kedro project
4. execute io.load('something') on a jupyter notebook

## Expected Result
load a 'something' dataframe

## Actual Result
raises exceptions

```python
Py4JJavaError: An error occurred while calling o25.load.
: org.apache.spark.sql.AnalysisException: Path does not exist: file:/Users/go_kojima/sample_kedro_project/notebooks/data/01_intermediate/something.parquet
```

## Your Environment
Include as many relevant details about the environment in which you experienced the bug:

* Kedro version used (`pip show kedro` or `kedro -V`):
kedro, version 0.14.1
(anaconda3-2019.03)

* Python version used (`python -V`):
Python 3.7.3
(anaconda3-2019.03)

* Operating system and version:
macOS Mojave version 10.14.5

[...]

PR #5768, opened 2026-09-15
ci: remove untrusted branch ref from checkout to close cache-poisoning alerts

merelcht:
## Description

Follow-up to kedro-org/kedro#5766 and kedro-org/kedro#5757, which added a runtime allow-list ("Validate branch input") before checkout in `unit-tests.yml`/`lint.yml`/`e2e-tests.yml`/`detect-secrets.yml` to mitigate the CodeQL `actions/cache-poisoning/poisonable-step` alerts (#41-#47). 

Those alerts stayed open, because CodeQL's data-flow rule flags any `workflow_call` string input reaching `actions/checkout`'s `ref:` before a cache-writing step (`astral-sh/setup-uv@v7`) — a runtime shell check doesn't change that from a static-analysis standpoint, and (worse) interpolating `${{ inputs.branch }}` directly into a `run:` script is itself a script-injection anti-pattern.

+ it seems the syntax was incorrect as well and builds are failing on `main` and `develop`.

## Development notes

Fixes a GitHub Actions cache-poisoning vulnerability (code-scanning alert #47) by removing the branch input from all-checks.yml and its reusable sub-workflows (unit-tests.yml, e2e-tests.yml, lint.yml, detect-secrets.yml), instead of only validating it.

**Fix**: remove the branch input entirely so no workflow input ever reaches a checkout ref:. nightly-build.yml now:
1. Dispatches `all-checks.yml` directly on the target branch via `gh workflow run all-checks.yml --ref main|develop` (hardcoded, trusted strings — not user input).
2. Polls `gh run list` for the newly created run (filtering by branch + workflow_dispatch event + a createdAt timestamp taken just before dispatch, to avoid picking up a stale run).
3. Uses `gh run watch --exit-status` to block on and inherit that run's pass/fail result.

This eliminates the vulnerable data flow rather than papering over it — no code path exists anymore where a workflow input controls what gets checked out.

### Extra context
Root cause: `nightly-build.yml` runs on schedule/workflow_dispatch with default-branch cache and secret access. It called the reusable workflows with a branch input (main/develop) that was passed straight into actions/checkout's ref:. CodeQL flags this as a poisonable step: an attacker-influenced ref checked out in a privileged, cache-writing context can be used to poison the shared cache and achieve code execution in later privileged runs. 

A previous commit on this branch added a runtime case main|develop allowlist, but that doesn't satisfy CodeQL's static data-flow check — the sink (checkout ref: inputs.branch) is still reachable + fails the builds.

### Known limitation

The run-discovery loop matches by `createdAt >=` before rather than a unique correlation ID, so in the rare case of API listing lag or a concurrent manual dispatch of `all-checks.yml` on the same branch, it could time out or pick the wrong run. Low risk, and the failure mode is a loud nightly-job error rather than a silent skip or security issue.

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)


SajidAlamQB (approved):
Thank you @merelcht!

.github/workflows/nightly-build.yml
```diff
@@ -11,14 +11,68 @@ permissions:
...
+    permissions:
+      actions: write
+    steps:
+      - name: Trigger and wait for all-checks on main
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+          TARGET_BRANCH: main
+        run: |
+          set -euo pipefail
```
ankatiyar, line 24: This is flagged by AI: The `set -e` will apparently abort the whole script on the first transient `gh` failure instead of re-trying. `gh run list` can return non zeroes for reasons other than run not being visible yet eg.GH rate limit or network blip. 

I think it's fine, this has always been flaky so we can just re-run the thing but just flagging 🤔

ankatiyar: Possible solutions;

- prefix the assignment with || true and check emptiness (run_id=$(gh run list ... ) || true), or
- temporarily disable -e around the polling call, or
- wrap it in an if ! run_id=$(...); then continue; fi.

ankatiyar (approved):
Thanks!

.github/workflows/nightly-build.yml
merelcht: Thanks @ankatiyar ! I went with the first option.

